PR #157, opened 2012-03-21
Updated YouTube Tech

jhurliman:
I wrote a new YouTube tech following the latest VideoJS API conventions. It uses the YouTube iFrame embed method which automatically chooses between HTML5 or Flash when serving videos, staying true to the HTML5 nature of VideoJS :-).

I tested all of the basic functionality with the following platforms:
- Chrome 17.0.963.79 on OSX Lion
- Chrome 17.0.963.79 on Windows 7
- iPhone Safari on iOS 5.0
- iPad Safari on iOS 5.1

Testing on additional platforms and more API feature testing (such as swapping out the playing video with a src() call) is welcome.

### DIFF
--- a/src/player.js
+++ b/src/player.js
@@ -18,7 +18,8 @@ _V_.Player = _V_.Component.extend({
     tag.removeAttribute("controls");
 
     // Poster will be handled by a manual <img>
-    tag.removeAttribute("poster");
+    if (!options.nativeposter)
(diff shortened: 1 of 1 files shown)


kipotlea:
@heff  @jhurliman  : can videojs-youtube working perfectly with videojs-vast-plugin ?

PR #195, opened 2012-06-07
added composer.json for packagist.org

basz:


thijstriemstra:
@heff should I add this for my video.js plugins as well? I thought packagist.org was strictly for PHP packages..

basz:
Being six years more mature the I was six years ago I wouldn’t create this PR. Indeed packagist/composer is for php.

To manage dependecies other then PHP in a PHP project I would now simple use multiple managers.

Many of them allow for one to trigger the other which mean installing all dependencies is usually one command. 

It’s better then television!

PR #990, opened 2014-02-05
Add UMD support

peteschaffner:
This isn't ready for merging, but rather a place to start the conversation and move towards adding support for different module loaders.


ngaer:
@gkatsev @heff To solve problem with flash/js interaction we can use postMessage instead of global functions.

PR #1126, opened 2014-04-02
Live w/DVR CSS

tomjohnson916:
- Updated styles for a Live w DVR state.
- Player.addClass('vjs-live-dvr') to view.

### DIFF
--- a/src/js/control-bar/live-display.js
+++ b/src/js/control-bar/live-display.js
@@ -8,6 +8,8 @@
 vjs.LiveDisplay = vjs.Component.extend({
   init: function(player, options){
     vjs.Component.call(this, player, options);
+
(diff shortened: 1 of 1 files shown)


Surt:
Hello all,

Triying to show the seekBar during live (vod) hls without END.

I was looking through the issues until I found this referenced.
Any news about this? Since the new version on videojs where contrib-hls was merged I don't know how to proceed. Is there any way to show the seekBar during vod hls live?
I was triying the dvrSeekbar but it requires the videojs defined as global on window, and that means that I need to refactor a lot of my custom code.

Any help is appreciated. Thanks.

gkatsev:
I have some news about that. We're going to be bringing "live dvr" UI to Video.js. We're going to be working on it over the next few months.

Issue #942, opened 2014-01-15
Vertical Slider Not Working


nmatthew123:
I am trying to make my volumeBar vertical by using this code posted here : https://github.com/videojs/video.js/pull/477
This is how I am setting up my  video player 

``` javascript
var myPlayer = videojs("video-player", {
    'children': {
      'controlBar': {
        'children': {
          'volumeControl': {
            'children': {
              'volumeBar' : {
                vertical: true
              }
            }
          }
        }
      }
    }
});
```

For some reason my video menu bar disappers. Do you have any idea what is going on? I've been working on this for a day and Im not sure what I should do. I haven't worked with videojs before so I'm a complete noob.

nothernrace82:
Perfect solution, Thank you for sharing, mmcc !!!

maluklo:
Evaluate new skin video-js-6.2.4

https://github.com/maluklo/Skin-video-js-6.2.4

PR #1161, opened 2014-04-23
Adding styling and implementation fixes for vertical volume menu

jnwng:
Previous implementation wasn't completely correct; styling was
out-of-date. Addresses #942

example to play with here: http://codepen.io/jnwng/pen/peDhl

screen shot 2014-04-23 at 11 50 11 am
screen shot 2014-04-23 at 11 50 06 am
screen shot 2014-04-23 at 11 49 58 am

### DIFF
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -294,10 +294,6 @@ fonts to show/hide properly.
   margin: 1.1em auto 0;
 }
 
-.vjs-default-skin .vjs-volume-menu-button .vjs-menu-content {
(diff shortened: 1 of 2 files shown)


tneaves:
@gkatsev has this been simplified? Is there a place I can look?

gkatsev:
@tneaves yes! the default volume control the `volumeMenuButton` has an `inline: false` option which also sets `vertical: true` which makes the volume vertical. Please take a look at the latest v5 release.

PR #1209, opened 2014-05-14
Fixed the loadstart event listener order

### LINKED ISSUE #1193: When videojs is loaded using Modernizr the "vjs-has-started" class is not added
Test case: http://sandbox.saeidmohadjer.com/video_vjs_modernizr.html

The lack of "vjs-has-started" class causes video play button to remain visible and video control bar to remain hidden. When digging deeper I noticed that this is due to videojs missing the "loadstart" event which browser invokes and hence not setting "initFirstPlay" handler which triggers "firstplay" event that is responsible for setting "vjs-has-started" class on video container.

### LINKED ISSUE #1207: vjs-has-started not being added on init
The [fix](https://github.com/videojs/video.js/commit/95c29e684f405cb410a7e19dbf9668b5b46d349b#diff-3b0266ff1c037b289ec624ab25b0272eR76) for #1050 changes the order of event handler registrations slightly and can mean that `vjs-has-started` is not added to the player. Before https://github.com/videojs/video.js/commit/95c29e684f405cb410a7e19dbf9668b5b46d349b, the first play event triggered on the player would add `vjs-has-started`. After the change, `loadstart` has to be triggered before `play`. If the sources are specified through the player options object, `loadstart` occurs when the player invokes its prototype constructor, before the player `loadstart` handler is registered.

heff:
so it's attached before the HTML5 tech triggers it. fixes #1208 fixes #1207 fixes #1193

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -32,8 +32,10 @@ vjs.Html5 = vjs.MediaTechController.extend({
     // If the element source is already set, we may have missed the loadstart event, and want to trigger it.
     // We don't want to set the source again and interrupt playback.
     if (source && this.el_.currentSrc === source.src && this.el_.networkState > 0) {
-      player.trigger('loadstart');
-
+      // wait for the player to be ready so the player listeners are attached
+      player.ready(function(){
+        player.trigger('loadstart');
+      });
     // Otherwise set the source if one was provided.
     } else if (source) {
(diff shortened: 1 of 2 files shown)


githubzhaoqian:
Why don't flash mode support drag Multiple buffered regions or ?start=number? Or how to support  and call?

PR #1353, opened 2014-07-17
WIP: Working browserify. Not backwards compatible.

gkatsev:
DO NOT MERGE.

This is the work-in-progress for browserifying videojs. The `bundle.js` file is built using
```sh
browserify -d --standalone 'videojs' . -o bundle.js
```
from within the videojs directory. Can also just run `npm run build`.

This will create a `bundle.js` file which is a standalone (UMD) build of videojs. You can then load the bundle file directly in the browser together with the generated CSS and you have a working videojs player.
You can also consume videojs directly from another app/module that uses browserify by just `require('video.js')` and then when you run browserify it'll figure things out.

Unfortunately, this isn't backwards compatible right now. The various components aren't really exposed directly right now and if you add more components it won't find them on the videojs object since it excepts them on the `./components.js` module. Though, more modules could be added by requiring the components.js module (`require('video.js/components`)` and then adding the extra component to it.
Probably a better way is to create a function for registering components and techs.

I'm going to see whether I can make it backwards compatible. Also, this needs to be cleaned up because we don't need all the `var vjs = {}; /* ... */ module.exports = vjs` everywhere. That var is superfluous.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,17 +13,19 @@
   "homepage": "http://videojs.com",
   "author": "Steve Heffernan",
   "scripts": {
-    "test": "grunt test"
+    "test": "grunt test",
+    "build": "browserify -d --standalone 'videojs' . -o bundle.js"
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/videojs/video.js.git"
   },
-  "main": "./dist/video-js/video.js",
+  "main": "./src/js/video.js",
   "dependencies": {
     "videojs-swf": "4.4.2"
   },
   "devDependencies": {
+    "browserify": "^4.2.1",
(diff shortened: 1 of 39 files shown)


heff:
We should handle this first before any of the other 5.0 tasks. gkatsev/video.js#4 still needs to be merged, and then there will probably be a bunch of merge conflicts. Might even be easier to use this as a guide and start from scratch. @gkatsev do you think you'll have time to address this soon or should we open it to whoever can get to it?

mmcc:
I think we need to decide on #1812 before going any further.

mmcc:
Closing in lieu of #1976

PR #1742, opened 2014-12-12
Support of pseudo-streaming

ange007:
Necessary functions for pseudo-streaming.
https://github.com/videojs/video-js-swf/issues/13#issuecomment-66777407

### DIFF
--- a/src/js/media/flash.js
+++ b/src/js/media/flash.js
@@ -65,6 +65,23 @@ vjs.Flash = vjs.MediaTechController.extend({
         lastSeekTarget
     ;
 
+	// PseudoStreaming parameters
+	if( options.pseudoStreamStartParam ) {
+      if((source && source.type) || options.pseudoStreamStartParamType) {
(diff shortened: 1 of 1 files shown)


heff:
@tomjohnson916, if you do end up working on the swf for BC hack week, take this one into consideration.

ange007:
To release 5 versions it is planned to include this function?

yangshenys:
How is it going? @heff

ange007:
Chances of switching on of this function in the next versions aren't present?

githubzhaoqian:
I can't wait...... It's looking forward to it.Why not appear in version 5?

mmcc:
We've let this one linger for waay too long, so my apologies on that, but we're just not prepared to support or include such a feature in Videojs at this time. Sorry everyone, but thank you so much for the contributions @ange007.

degtyaryov:
My implementation works fine on production more than a year. From user complaints were not.

PR #1749, opened 2014-12-17
Captions

gkatsev:
This PR supercedes the other PRs (https://github.com/videojs/video.js/pull/1736)

* [x] fix up one last test
* [x] more tests
* [x] fixup settings menu styles (prettier/reworked should come as a separate PR)


package.json
```diff
@@ -21,7 +21,8 @@
   },
   "main": "./dist/video-js/video.js",
   "dependencies": {
-    "videojs-swf": "4.5.2"
+    "videojs-swf": "4.5.2",
+    "vtt.js": "git+https://github.com/gkatsev/vtt.js.git#custom-build"
```
heff: How is this going to get pulled into video.js?

marypearson: Well honestly I don¹t know, and I don¹t understand which is why I couldn¹t
get anything from your website.

The instructions I got from the creator of the theme were to install your
plugin and then to put:

[videojs mp4="http://longfarms.ca/long-content/uploads/birds_snow.mp4"
preload="auto" autoplay="true" loop="true" controls="false" height="330"
width="960" alt="video of birds with snow falling"]

Into GP Hooks section After Header.

If there is more than that then I guess this is the information I am
lacking.  Would really appreciate it if you could tell me what I¹m missing.

Thank you.

heff:
screen shot 2015-01-05 at 3 37 30 pm

On IE8, the captions get covered by the control bar, and since the background is not transparent, it's hard to tell you've even turned them on. With single line captions you never see them until the control bar fades out. It feels pretty awkward. We might need to go back to adjusting their height...

gkatsev:
@heff forgot to push the style change to move captions up. Should be up now.

package.json
gkatsev: This is what we should open a new issue about.

heff:
screen shot 2015-01-05 at 3 49 51 pm

In IE8, for some reason when I click the currently selected captions menu item, it toggles the highlight on the captions settings menu item.

heff:
Captions position looks much better now.

heff:
I still think it'd be nice to provide a demo window that shows what captions will look like as you change the setting. At least now you can pause when a caption is showing and see it change, but that's still not ideal I think. If no one else thinks it's worth adding to this then ok.

heff:
I'm getting broken tests in IE8. And strangely not in the minified tests.

screen shot 2015-01-05 at 4 04 54 pm

package.json
heff: @marypearson, this is a different comment thread then you were on before. You might want to ignore it.

@gkatsev If we merge this in and release it as is, what happens to people who just bump their version and are using captions?

src/js/event-emitter.js
```diff
@@ -0,0 +1,41 @@
+vjs.EventEmitter = function() {
+};
+
+vjs.EventEmitter.prototype.allowedEvents_ = {
+};
+
+vjs.EventEmitter.prototype.on = function(type, fn) {
+  var ael = this.addEventListener;
```
heff: We should have a comment here about why this is needed.

src/js/media/html5.js
```diff
@@ -26,6 +32,33 @@ vjs.Html5 = vjs.MediaTechController.extend({
       this.setSource(source);
     }
 
+    if (this.el_.hasChildNodes()) {
```
heff: Could this block be moved into createEl?

heff:
Do subtitles still work with this, and should we be styling them differently by default like we were before?

heff:
Aside from the comments this is looking really good. The biggest issue I see is handling the vtt.js code. I don't think we should merge this in until we have a story for that, because as-is this will break captions in Firefox and IE8 for everyone using captions.

My initial thoughts on that are we should just bundle vtt.js with video.js, and then if the file size is an issue  for people we can provide a custom build option that excludes it.

gkatsev:
subs definitely still work: .
I also don't see a problem with using the same default styles.
but I think chapters may or may not be working.

heff:
Ah, yeah we should check on chapters, since there was recent work to get
those working again.

gkatsev:
It shouldn't be hard to get it working, if it is broken but it's lower on priority list :)

heff:
Where is captions at now? Is everything here addressed and does it need any additional review?

gkatsev:
@heff there are still a few minor things left like the last few comments you had and IE8.

gkatsev:
@heff Here's the build task: 0a131f22b8b16af7a01dc89c63733b49b0c79103

src/js/event-emitter.js
gkatsev: fixed in 5c36d27

src/js/media/html5.js
gkatsev: Unsure. This is where this code has been for a long time. I just augmented it.

gkatsev: Actually, no, we currently can't move this piece down because it depends on the initialization of the 'featuresNativeTexttracks' and emulated text tracks which happen slightly after `createEl` gets called.

gkatsev:
Apparently, label-less captions added via `addRemoteTextTrack` get somehow swallowed by the captions button and not displayed.

gkatsev:
Fixed IE8 tests. Had to disable two tests on IE8 only because they weren't behaving properly. Also, I removed a usage of `isNaN` from slider which was causing a third test to fail in IE8.
The only thing needed now is fixing chapters which I'm looking at right now.

gkatsev:
Oh, looks like the reason label-less items weren't showing up was for a reason. Only thing I can think off is that we default language for display but even that isn't required. We could default to a "Unknown Captions" or something like ios does. Unsure whether not having then displayed at all is better than displayed with some default label.

heff:
Re label-less: I like falling back to srclang and then to "unknown". If they leave both off it's probably unintentional or that they don't understand what's needed. Showing "unknown" would be easier to debug than having them not show up at all. At least I can't think of any use case for hiding captions from the captions button.

gkatsev:
Would it be just "unknown" or "unknown {{kind}}"?

heff:
Kind seems unnecessary in the captions menu. "unknown language" might be more appropriate, but I think I'd stick with just unknown.

gkatsev:
This would apply to subtitles too. We can have it say "unknown subtitles" and "unknown captions". Just unknown is fine too.

heff:
Yeah, I still like unknown. When they are labeled it's typically just "English" not "English Captions"

gkatsev:
Right, makes sense. I think iOS does "unknown captions" because I think they have only one menu for all text tracks.

package.json
```diff
@@ -21,7 +21,8 @@
   },
   "main": "./dist/video-js/video.js",
   "dependencies": {
-    "videojs-swf": "4.5.3"
+    "videojs-swf": "4.5.3",
+    "vtt.js": "git+https://github.com/gkatsev/vtt.js.git#shim-build"
```
heff: Do the recent merges mean we can change this to the main lib yet?

gkatsev: Unfortunately, not yet. All the IE8 changes went in but the shim build stuff still need to get tweaked first.

src/css/video-js.less
```diff
@@ -964,33 +964,92 @@ body.vjs-full-window {
...
-  left: 1em;
-  right: 1em;
+  top: 0;
+  left: 0;
+  bottom: 3em;
+  right: 0;
+  pointer-events: none;
+}
+
+/* Increase font-size when fullscreen */
```
heff: Is there resource or discussion we can point to in the comment for why we're doing this? I actually don't  know why myself.

gkatsev: Hm... I think this should be removed, though, it's actually not doing anything right now because vttjs overrides the font-size manually. I think it got added when I merged some stuff from aster.

src/js/media/html5.js
```diff
@@ -384,6 +522,28 @@ vjs.Html5.canControlPlaybackRate = function(){
...
+
+  // Figure out native text track support
+  // If mode is a number, we cannot change it because it'll disappear from view.
+  // Browsers with numeric modes include IE10 and older (<=2013) samsung android models.
+  // Firefox isn't playing nice either with modifying the mode
+  supportsTextTracks = !!vjs.TEST_VID.textTracks;
+  if (supportsTextTracks && vjs.TEST_VID.textTracks.length > 0) {
+    supportsTextTracks = typeof vjs.TEST_VID.textTracks[0]['mode'] !== 'number';
+  }
+  if (supportsTextTracks && vjs.IS_FIREFOX) {
```
heff: It feels like a shame to disable native captions in firefox completely. I feel like we should submit a bug, or see if one already exists that we can track for this.

gkatsev: Yeah. I just didn't have time to figure out what was going on. Still not certain where the root of the problem is.

src/js/media/media.js
```diff
@@ -253,6 +261,139 @@ vjs.MediaTechController.prototype.setCurrentTime = function() {
   if (this.manualTimeUpdates) { this.player().trigger('timeupdate'); }
 };
 
+vjs.MediaTechController.prototype.initTextTrackListeners = function() {
+  var player = this.player_,
+      tracks,
+      textTrackListChanges = function() {
+        var textTrackDisplay = player.getChild('textTrackDisplay'),
```
heff: Was there a specific reason we couldn't let the textTrackDisplay handle updating itself instead of requiring the tech to update it?

gkatsev: Not certain. It's possible that this function can be moved to the text track display.

test/unit/tracks/tracks.js
```diff
@@ -0,0 +1,330 @@
...
+
+  vjs.IS_FIREFOX = true;
+
+  ok(!vjs.Html5.supportsNativeTextTracks(), 'if textTracks are available on video element, native text tracks are supported');
+
+  vjs.TEST_VID = oldTestVid;
+  vjs.IS_FIREFOX = oldIsFirefox;
+});
+
+test('if native text tracks are not supported, create a texttrackdisplay', function() {
```
heff: This appears to be a duplicate test of line 197

src/js/media/html5.js
heff: Ok, if that's gonna take work, can we add a TODO comment here for now so we can move forward?

src/js/media/media.js
heff: Same here as far as adding a TODO comment.

heff:
@gkatsev I'll work on some of the remaining cleanup tasks. 

I don't see a resolution to [the last chapters related question](https://github.com/gkatsev/video.js/commit/83a53afd56df2f85183f50b6b2d497a7cd4a39a5#commitcomment-9425855). Were you able to figure that out?

src/css/video-js.less
gkatsev: Removed.

src/js/media/html5.js
gkatsev: Done.

src/js/media/media.js
gkatsev: Done.

test/unit/tracks/tracks.js
gkatsev: Huh, so it is.

gkatsev:
@heff pushed TODOs. Would you want me to merge master into this branch so it's easier to merge this back into master?

heff:
Thanks for handling those! Yeah, if you have a minute to merge master you're in a better place to understand merge conflicts, but otherwise I can take care of it.

gkatsev:
@heff merged in.

PR #1777, opened 2014-12-31
Let the progress bar fill the space at 100%,fixes #1247

### LINKED ISSUE #1247: Progress bar can't be filled on 100%
At the end of video progress bar not in the end:
https://monosnap.com/image/yhAoC2yYAh7x95ikCMq2xxJokAJptc
https://monosnap.com/image/o9qxlbq17SLWd0O7JMt7hnqlo85CJN

moshoujingli:
I choosed the second way
>Allow the handle edges to go beyond the edges of the progress bar, i.e. the center of the handle is always locked to the mouse, even at 0 and 100%.

So the progress bar will fill the space as expected, and at the beginning or the end, the handler will beyond the edges of the progress bar like below.

### DIFF
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -462,7 +462,7 @@ fonts to show/hide properly.
 }
 
 .vjs-default-skin .vjs-seek-handle {
-  width: 1.5em;
+  width: 1em;
   height: 100%;
 }
 
--- a/src/js/slider.js
+++ b/src/js/slider.js
@@ -101,16 +101,8 @@ vjs.Slider.prototype.update = function(){
         // The width of the handle in percent of the containing box
         // In IE, widths may not be ready yet causing NaN
         handlePercent = (handleWidth) ? handleWidth / boxWidth : 0,
-
-        // Get the adjusted size of the box, considering that the handle's center never touches the left or right side.
-        // There is a margin of half the handle's width on both sides.
-        boxAdjustedPercent = 1 - handlePercent,
-
-        // Adjust the progress that we'll use to set widths to the new adjusted box width
-        adjustedProgress = progress * boxAdjustedPercent;
-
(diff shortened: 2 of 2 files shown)


heff:
@moshoujingli any chance we could get a live demo of this somewhere to try it out and get a feel for it?
@videojs/core-committers any thoughts on this design?

heff:
We talked about this a bit in IRC and the general consensus was that the handle poking out the side of the player would bug people.

There's a few things I think we need to do with this then
- Make the handle stop when it's left edge is at zero, and right edge is at 100%
- Find a shape that won't look weird with the new setup
- Separate the logic of the handle to clean this up a bit

moshoujingli:
I'm trying the second thing
> Find a shape that won't look weird with the new setup


Would you like to tell me what tool do you use to generate the four font files?I think I need to change the font for \e009 or append a new one with the shape like a normal square but I don't know if there's any easy way to do that.

heff:
@mmcc, would you want to make a PR for the controls updates you've been working on? I think it'd be better if @moshoujingli was participating in that rather than the current version.

mmcc:
I was going to open the PR for the whole branch pretty soon, but it might not be a bad idea to split that into its own PR anyway to keep things a little easier to digest. I'll do that tomorrow, but in the meantime, @moshoujingli, [these are the changes](https://github.com/mmcc/video.js/blob/base-styles/src/js/slider.js#L76-L151)

mmcc:
There's a PR open to break up the logic a bit: #1816. I think we should tackle that first, get it pulled in, then revisit how we want to handle this issue at that point.

In the new base theme we don't necessarily need to do this since we're not going with a 100% width bar, but I think we should still easily allow for the current default theme to work (or at least this style). I'm going to close this for now, but let's continue the conversation over on #1816.

PR #1781, opened 2015-01-06
fix multiple assignments in var statements

rphv:
fixed var statements in `player.js` with multiple assignments as per the [Code Style Guide](https://github.com/videojs/video.js/blob/master/CONTRIBUTING.md#code-style-guide).

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -800,10 +800,10 @@ vjs.Player.prototype.buffered = function(){
  * @return {Number} A decimal between 0 and 1 representing the percent
  */
 vjs.Player.prototype.bufferedPercent = function(){
-  var duration = this.duration(),
-      buffered = this.buffered(),
-      bufferedDuration = 0,
-      start, end;
+  var duration = this.duration();
+  var buffered = this.buffered();
+  var bufferedDuration = 0;
+  var start, end;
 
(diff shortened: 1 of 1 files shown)


heff:
Interesting, I'd forgotten that rule was in there. Thanks for the cleanup help. While that rule is there, if you look around we actually moved to using the single-var method with new code.

If @gkatsev likes multiple vars now, that's actually what I'd prefer now, but otherwise we don't need this to turn into a debate and we'll just keep going with what we're doing, and then do a big style cleanup with the next major version to make everything consistent.

mmcc:
So much +1 for multiple vars. I've actually done a bit of this already in a branch I've got open, so for the sake of merge conflicts, I think we should just update these as we touch classes while working on the next version, then do one last final cleanup before the release.

heff:
@gkatsev any objections? (Gary is the maintainer of the internal Brightcove javascript style guide, which the video.js style guide is closely tied to).

heff:
I'm going to close this issue until we've resolved #1812 (add ES6 transpiler). That will be a more natural point to discuss and change our var style, which this issue should wait for a decision on. (Assuming there isn't a consensus before then)

Thanks again for the contribution @rphv.

Issue #1500, opened 2014-09-11
tabbing at the keyboard, focus outlines, and button order


albell:
If you go to the homepage and play the vid, and then pause it, and start tabbing around, the focus moves in markup order, like this:

- play button
- progress control

…so far so good. But then it leaps to the far right and starts incrementing to the left:

- fullscreen
- volume
- mute
- captions

This feels like a friction point to me. The first time I did this, I had no idea where focus had leapt to. I had to do it over and over again, watching carefully. There are two separate issues here:

1) The focus-indicating `text-shadow` in the default skin is not a conventional focus indicator. It’s more of a light visual feedback for users whose eyes are already trained on the cursor. I realize mouse users and some designers sometimes find focus outlines uncool and cluttery, but I think it can be done both clearly and tastefully using conventional (dotted blue?) outlines. If a client violently hates the outlines, and an author can’t win that fight, then the author can turn them off, but clear, usable focus outlines should be incentivized in the default.

2) The visual leap across the player is unexpected, especially given the above.  Users expect tabbing to work consistently in one direction across a component. Obviously we don’t want to go setting tabindex on the elements. So one solution would be adding a wrapper for everything thats supposed to float right, float that wrapper right, and reverse the markup order of its children. Meh. Another slicker solution could be to switch the `float:right` elements to `display: inline-block`, and use `text-align:right`, and again reverse the markup order for those elements.

Thoughts?

gdkraus:
This appears to be very similar to how YouTube handles it. YouTube appears to
1. assume a user is a mouse user
2. have a listener for key presses anywhere in the player, and on a key press add the class 'ytp-keyboard-focus' to the player, which has the corresponding :focus CSS to show the onscreen focus
3. have another listener for any mouse clicks in the player, and on a mouse click it removes the class 'ytp-keyboard-focus'

mmcc:
Ok so just an update on this one, tab order has been immensely improved with the default theme for version 5. Instead of `float: right` we've moved to [flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) for modern browsers and table as a fallback. Thanks to this combo the ordering stays intact, so we can now list the elements to be created in the correct order in the control bar and tab ordering stays intact.

[...]

PR #1792, opened 2015-01-09
Feature/tab order

gdkraus:
This is a proposed solution to part of the discussion in #1500. This reorders the DOM elements so the tab order make more sense for keyboard users. It does this by creating two containers to store some of the UI elements in the control bar. The basic concept is this.
```
<div class="vjs-control-bar-right">
   <div class="vjs-control-bar-middle">
       Caption Menu
       Subtitle Menu
       Chapters Menu
   </div>
   Mute Control
   Volume Control
   Full Screen Control
</div> 
```
vjs-control-bar-right is float:right
vjs-control-bar-middle is float:left
all of the containing controls are float:left

The reason the inner vjs-control-bar-middle is needed is because those items are not loaded initially with all of the other controls, so I can't set when to load them in the DOM. Creating this area gives me a target where I can insert them at a later time.

Thoughts?

### DIFF
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -169,14 +169,14 @@ The default control bar that is a container for most of the controls.
 
 /* The right side of the control bar */
 .vjs-default-skin .vjs-control-bar .vjs-control-bar-right {
-    float:right;
+    float: right;
 }
 
 /* The middle section of the control bar 
 It's technically the content that is visually just to the left of the right-most controls.
 */
 .vjs-default-skin .vjs-control-bar .vjs-control-bar-middle {
-    float:left;
+    float: left;
 }
 
 /* Show the control bar only once the video has started playing */
@@ -749,7 +749,7 @@ easily in the skin designer. http://designer.videojs.com/
 --------------------------------------------------------------------------------
 */
 .vjs-default-skin .vjs-menu-button {
-  float: right;
+  float: left;
   cursor: pointer;
 }
 
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -167,6 +167,18 @@ The default control bar that is a container for most of the controls.
   .background-color-with-alpha(@control-bg-color, @control-bg-alpha);
 }
 
+/* The right side of the control bar */
+.vjs-default-skin .vjs-control-bar .vjs-control-bar-right {
+    float:right;
+}
+
+/* The middle section of the control bar 
+It's technically the content that is visually just to the left of the right-most controls.
+*/
+.vjs-default-skin .vjs-control-bar .vjs-control-bar-middle {
+    float:left;
+}
+
 /* Show the control bar only once the video has started playing */
 .vjs-default-skin.vjs-has-started .vjs-control-bar {
   display: block;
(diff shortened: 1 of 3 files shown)


gkatsev:
Re-ordering the elements, for better or worse, is a breaking change.

gdkraus:
Why is it a breaking change? I see the build failed, but I'm not sure why it's failing.

mmcc:
As always, thanks for the PR! This is a bit of a moot point to begin with since we've already addressed the component order to account for tab ordering in the new default skin for version 5, so I don't think we're going to pull anything drastic like this in in the meantime. I'm not a big fan of adding an additional container like you've got here, but again, I think we've addressed your issues moving forward with v5.

gkatsev:
@gdkraus it'll be a breaking chance because people rely (for better or worse) on the current ordering of elements as well as the general format of the HTML. This matters particularly for people who wrote and used custom videojs skins. As @mmcc said, we're going to be coming up with a new skin for videojs 5.0 which will have some other breaking changes in general as well. Hopefully, people who this affects can wait for the new skin.

It might be worth pulling this out into a videojs plugin that reorders the control bar elements based on a list configuration so tabbing through the elements works correctly.

gdkraus:
How is the solution implemented in version 5?

mmcc:
@gdkraus As far as ordering goes, the [control bar](https://github.com/videojs/video.js/blob/master/src/js/control-bar/control-bar.js#L13) elements are now described as an ordered array. There's still some thought that needs to go into how we want to appropriately handle arbitrary injection of menu items (such as via a plugin), but this solves the problem for a vast majority of users.

gdkraus:
@mmcc So with that reordering, will the UI elements appear in the DOM in the expected tab order?

mmcc:
Yes.

mmcc:
As I said, the remaining question mark is how we're going to appropriately handle menu items that get added later without borking the tab order. What we've got now improves things immensely, but it's not perfect when it comes to handling the awful world of arbitrary things :)

mmcc:
@gdkraus Ok, so after a little brainstorming for arbitrary icons, I think we've got a potential solution. Our thought is to essentially have a divider that will always sit in the control bar that you would use as your reference point when appending new icons to the dom. Rough sketch:

```
- Play
- Progress Bar
- Remaining Time
- **Separator Thingy**
- Captions
- Subtitles
- Volume
- Fullscreen
```

So, even if any of those elements were removed, as long as the plugin author or dev adding new icons adds things after that separator, they can know where it's getting added to the dom and the tab order remains intact

```
- Play
- Progress Bar
- **Separator Thingy**
- Awesome Plugin 1
- Awesome Plugin 2
- Random Btn Someone Added
- Volume
- Fullscreen
```

What do you think?

heff:
@gdkraus in the 5.0 skin we're no longer using floats to align the icons, we're using flex box (and display: table in ie8). That removes the issue of having to change the order of the html elements to get the buttons in the right order (i.e. fullscreen at the right).

And for what it's worth, I had the same idea of grouping the elements together in sub-divs. The flex box approach (not my idea) is magical.

PR #1793, opened 2015-01-09
ManualTimeUpdatesOff was not de-registering events

gkatsev:
`ManualTimeUpdatesOff` was trying to de-register `trackCurrentTime` and
`stopTrackingCurrentTime` against itself rather than the player, which is
where these events were registered against.

This manifests itself if you have flash-based tech that is loaded initially and then try to unload it and switch to an html5-based tech.

### DIFF
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -215,10 +215,12 @@ vjs.MediaTechController.prototype.manualTimeUpdatesOn = function(){
 };
 
 vjs.MediaTechController.prototype.manualTimeUpdatesOff = function(){
+  var player = this.player_;
+
   this.manualTimeUpdates = false;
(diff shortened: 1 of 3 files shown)


mmcc:
lgtm

heff:
@mmcc, want to attempt merging this in with contrib?

mmcc:
yep

mmcc:
^^ thanks, auth.

giakas:
this same issue exists for 'controlsenabled' and 'controlsdisabled' event.

we never have a deregister for these events. so if we load another tech as part of src(value) and then dispose, we are getting exception. we should deregister these events as well.

gkatsev:
@giakas yes we should. Would you be up for submitting a PR for that change?

PR #1796, opened 2015-01-12
Fix woff font. Closes #1775

### LINKED ISSUE #1775: vjs.woff error in FireFox (bad linegap)
I am getting the following error in my console in FireFox 34 with the latest version of video.js:
```
downloadable font: OS/2: bad linegap: -64 (font-family: "VideoJS" style:normal weight:normal stretch:normal src index:1)
source: http://vjs.zencdn.net/f/3/vjs.woff
```

mister-ben:
Re-exported woff font to avoid `bad linegap` error on Firefox.


heff:
I was able to confirm the error and that this PR fixes it. It's not great that it affects the icon size, but not sure what we're supposed to do to fix this otherwise. @mmcc, should we just merge this in or try to fix the icon size?

mmcc:
I think we should try to fix the icon size if we're going to merge this in, it would stink to break things for folks right before we push an entirely new version of the font / theme.

heff:
@mmcc is going to try exporting again and see if that fixes the size issue. If not we'll just merge this in.

gkatsev:
We shouldn't merge this in if it's going to break things.

heff:
@mister-ben, could we get a screen shot on how different the icons look with the fix? I don't think we're going to be able to pull this in until we figure out how to get the icons back to normal.

mister-ben:
before http://jsbin.com/axedog/

after http://jsbin.com/koticu/

However I think I've been able to fix it in a better way now by editing the metrics in FontForge. I'll do some more testing with that and update.

mister-ben:
Closing this in favour of #1870

PR #1797, opened 2015-01-12
Fixed type support checking for an empty src string

### LINKED ISSUE #1785: Setting src to an empty string throws an error.
Is there a good way to stop the current video from loading? `player.src('')` works but spews errors into the console, which is undesirable.

Why not dispose of the player? I want to continue playing the same video later, I just don't want it eating bandwidth while "paused".

heff:
fixes #1785

Can I get a review?

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -307,13 +307,13 @@ vjs.Html5.nativeSourceHandler = {};
  * @return {String}         'probably', 'maybe', or '' (empty string)
  */
 vjs.Html5.nativeSourceHandler.canHandleSource = function(source){
-  var ext;
+  var match, ext;
 
   function canPlayType(type){
     // IE9 on Windows 7 without MediaPlayer throws an error here
     // https://github.com/videojs/video.js/issues/519
     try {
-      return !!vjs.TEST_VID.canPlayType(type);
+      return vjs.TEST_VID.canPlayType(type);
     } catch(e) {
       return '';
     }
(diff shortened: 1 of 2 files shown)


src/js/media/html5.js
```diff
@@ -307,13 +307,13 @@ vjs.Html5.nativeSourceHandler = {};
  * @return {String}         'probably', 'maybe', or '' (empty string)
  */
 vjs.Html5.nativeSourceHandler.canHandleSource = function(source){
-  var ext;
+  var match, ext;
```
mmcc: I thought the plan was to go with explicit var declarations moving forward, or are instances like this an exception (no actual values set, just defined)?

src/js/media/html5.js
```diff
@@ -307,13 +307,13 @@ vjs.Html5.nativeSourceHandler = {};
...
 
   function canPlayType(type){
     // IE9 on Windows 7 without MediaPlayer throws an error here
     // https://github.com/videojs/video.js/issues/519
     try {
-      return !!vjs.TEST_VID.canPlayType(type);
+      return vjs.TEST_VID.canPlayType(type);
```
mmcc: Why'd we revert from the `!!`?

mmcc:
LGTM :thumbsup:

src/js/media/html5.js
heff: Waiting for @gkatsev's input to change what I'm using in practice.

src/js/media/html5.js
heff: When implementing the Flash source handler I decided to go with the maybe/probably/empty-string return values to be more similar to canPlayType. HTML5 needed to be updated to match that too.

heff:
Thanks!

PR #1803, opened 2015-01-14
(doc) Adds guide for removing players from the page

brycefisher:
Added a guide to explain how to safely remove players from the page.

Ran into a situation like this:
http://jsbin.com/gahususoju

I couldn't find this info anywhere in the guides or documentation, so thought I'd chip in :-)

### DIFF
--- a/docs/guides/removing-players.md
+++ b/docs/guides/removing-players.md
@@ -6,13 +6,17 @@ Sometimes, you want to remove players after page load (in single page apps or mo
 Call `.dispose()`
 -----------------
 
-Before removing the html associated with your videojs player from the page always call the player's `dispose()` method
+To remove the html associated with your videojs player from the page always call the player's [`dispose()`](https://github.com/videojs/video.js/blob/stable/docs/api/vjs.Player.md#dispose) method:
 
 ```javascript```
 var oldPlayer = document.getElementById('my-player');
-videojs(oldPlayer).dispose(); // Don't skip this step!
-oldPlayer.parentNode.removeChild(oldPlayer);
+videojs(oldPlayer).dispose(); 
 ```
+
+This method will:
+
+ 1) reset the internal state of videojs
+ 2) remove the player's dom from the page
 
 Showing / Hiding a Player
 -------------------------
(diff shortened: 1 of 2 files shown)


gkatsev:
LGTM. @mmcc?

docs/guides/removing-players.md
```diff
@@ -0,0 +1,38 @@
...
+```javascript```
+var oldPlayer = document.getElementById('my-player');
+videojs(oldPlayer).dispose(); // Don't skip this step!
+oldPlayer.parentNode.removeChild(oldPlayer);
+```
+
+Or...Use Unique Ids
+-------------------
+
+If you prefer not to call `dispose()` on a player, you can always create new players with unique ids for that page load.
```
mmcc: I'd probably suggest just removing this bit altogether. Any time you remove a player from the DOM (or hide it, or just don't need it anymore), it's a good idea to remove it, even if just from a resource management perspective.

docs/guides/removing-players.md
```diff
@@ -0,0 +1,38 @@
...
+
+Signs You Did It Wrong
+-------------------------
+
+If you encounter an console error in browser similar to this, you've probably forgotten to dispose() a player before removing it from the dom:
+
+```
+TypeError: this.el_.vjs_getProperty is not a function
+"VIDEOJS:" "Video.js: buffered unavailable on Hls playback technology element." TypeError: this.el_.vjs_getProperty is not a function
+Stack trace:
```
mmcc: Might want to explicitly mention this would happen when using the [contrib-hls](http://github.com/videojs/videojs-contrib-hls) plugin.

mmcc:
Looks great! The only other thing I'd suggest is adding a section about hiding / showing a player. For instance, if you've got a modal that a player pops up in, you should create and dispose the player when the player shows / hides. If the Flash tech is used and you try to hide it, things will go poorly because, Flash.

Thanks for the PR!

brycefisher:
Thanks for the extra info! I'll update the branch with your feedback.

brycefisher:
If this looks good to you @gkatsev and @mmcc, I'll squash this down into a single commit in a new PR

gkatsev:
Probably no need, since we squash it up during out merge process.

brycefisher:
okey-dokey!

brycefisher:
Woot! Thanks for accepting my PR @mmcc and @gkatsev

PR #1804, opened 2015-01-14
Removed unnecessary display:none; from text-track-display

mmcc:


heff:
Why's it unnecessary, and is this fixing a specific bug?

mmcc:
An issue was never opened, but the bug was discussed in [PR comments](https://github.com/videojs/video.js/commit/a67201f913e0ab3e61ba20690c3bdfa1dfbf0304#comments)

Since the track is actually added / removed from the dom for each cue point, it's redundant to also unhide it. I can't think of a situation that this would break, but I could be overlooking something. Either way, we need to make sure to resolve this before cutting a new release since this is currently broken in master.

heff:
LGTM!

PR #1809, opened 2015-01-18
Copy touch object for firstTouch

steverandy:
Using object reference will result in touchDistance value is always 0.
By copying touch object, firstTouch.pageX and firstTouch.pageY values
will not change during touchmove event.

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1043,7 +1043,7 @@ vjs.Component.prototype.emitTapEvents = function(){
   this.on('touchstart', function(event) {
     // If more than one finger, don't consider treating this as a click
     if (event.touches.length === 1) {
-      firstTouch = event.touches[0];
+      firstTouch = vjs.obj.copy(event.touches[0]);
       // Record start time so we can detect a tap vs. "touch and hold"
       touchStart = new Date().getTime();
       // Reset couldBeTap tracking
--- a/test/unit/component.js
+++ b/test/unit/component.js
@@ -462,13 +462,14 @@ test('should use a defined content el for appending children', function(){
 });
 
 test('should emit a tap event', function(){
-  expect(2);
+  expect(3);
 
   // Fake touch support. Real touch support isn't needed for this test.
   var origTouch = vjs.TOUCH_ENABLED;
   vjs.TOUCH_ENABLED = true;
 
   var comp = new vjs.Component(getFakePlayer());
+  var singleTouch = {};
 
   comp.emitTapEvents();
   comp.on('tap', function(){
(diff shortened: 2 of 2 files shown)


gkatsev:
Can you give a bit more background on when you're seeing the touches being zero when you don't copy the object? I've not had that issue before.

steverandy:
I find that tap is too sensitive on mobile. Scrolling on top of video.js player often triggers tap and play the video.

```
xdiff = event.touches[0].pageX - firstTouch.pageX;
ydiff = event.touches[0].pageY - firstTouch.pageY;
touchDistance = Math.sqrt(xdiff * xdiff + ydiff * ydiff);
```

On line 1063, `touchDistance` is always 0, because `firstTouch.pageX` and `firstTouch.pageY` values are equal to `event.touches[0].pageX` and `event.touches[0].pageY` respectively.

This makes scrolling always emits tap event, which is bad on small screen, because may often perform touch scroll on top of video.js player.

Further, I think we should also reduce `tapMovementThreshold` to 10 and `touchTime` limit to 200.

heff:
I'm not fully understanding the issue. @steverandy, could you describe the steps to reproduce the issue. Bonus points if you can record a video of it happening.

steverandy:
This is what I did to validate the issue:

1. Add the following lines to video.dev.js after line 2750, to log the values.

    ```
console.log("xdiff: " + xdiff);
console.log("ydiff: " + ydiff);
console.log("touchDistance: " + touchDistance);
    ```
2. Launch iOS simulator and Safari inspector
3. Do a quick scroll on top of video player

Actual result: xdiff, ydiff, and touchDistance values are always 0. Video starts playing.

Expected result: ydiff should be > 0, touchDistance should be > 0. Video should not start playing.

Video: https://dl.dropboxusercontent.com/u/2271268/simulator-result.mov

gkatsev:
Is this happening only on the simulator or on the ios device as well? Unfortunately, the simulator often does things differently.

steverandy:
Got the same result on device.

gkatsev:
Thanks for verifying!

heff:
Thanks for sticking with this @steverandy. The original issue now makes sense to me. Assuming the browser doesn't create a new touch object for touches[0] when the page X/Y changes, this would be a problem. Thanks for the fix.

@steverandy, we have some [exiting tests around touch events](https://github.com/videojs/video.js/blob/master/test/unit/component.js#L464). Do you think you could add some for this issue?

As a separate issue, I think it makes sense to reduce the touch more range to 10px. The current 22px referred to half the size of a standard iOS button. At the time HammerJS was using 10px but now they're [using 2px](https://github.com/hammerjs/hammer.js/blob/master/src/recognizers/tap.js#L35), so without knowing more myself it seem like we could at least go down to 10px.

steverandy:
Hi @heff, I have added two new tests to cover this issue.

heff:
This looks good to me. @videojs/core-committers, can we get a second review on this?

heff:
@gkatsev can you do a quick review of this one so we can get it pulled in for the next release?

dmlap:
Code-wise, this lgtm. Thanks for adding the additional test cases, @steverandy!

heff:
Thanks @steverandy!

steverandy:
Thanks! Hope to see a release soon.

PR #1814, opened 2015-01-20
Fixed problem with removing component

carpasse:
Component.id and Component.name are functions
we should either call the functions or pass component.id_ and component.name_


heff:
Good find, thanks! Looks good to me.

PR #1816, opened 2015-01-20
Slider updates - "separate" handle logic from the slider and allow for vertical or horizontal orientation

mmcc:
**NOTE**: This was originally done so I could unblock myself and continue working on the base styles, so there's almost certainly room for improvement. Going ahead and opening a PR to start the conversation, but this shouldn't be pulled in yet.

The crux of this is that it allows a slider to be either horizontal or vertical without needing to use a css transform. If the vertical option is set on the slider, it will update the elements height and do the necessary calculations against the Y axis, or vice versa if the slider is horizontal.

In the process of doing this, I broke the update logic into two parts so the slider handle update is handled in its own method. This is still pretty ugly, since we have to adjust the slider bar position in the slider handle method, but it helped a bit with readability (imo).

Thoughts?


src/js/slider.js
```diff
@@ -78,48 +81,73 @@ vjs.Slider.prototype.update = function(){
...
+  var barProgress = this.updateHandlePosition(progress);
 
-    var box = this.el_,
-        boxWidth = box.offsetWidth,
+  // Convert to a percentage for setting
+  var percentage = vjs.round(barProgress * 100, 2) + '%';
 
-        handleWidth = handle.el().offsetWidth,
+  // Set the new bar width or height
+  if (this.options()['vertical']) {
```
heff: I think this should check `this.vertical()` now instead of the option.

src/js/slider.js
```diff
@@ -78,48 +81,73 @@ vjs.Slider.prototype.update = function(){
...
-        // In IE, widths may not be ready yet causing NaN
-        handlePercent = (handleWidth) ? handleWidth / boxWidth : 0,
+/**
+* Update the handle position.
+*/
+vjs.Slider.prototype.updateHandlePosition = function(progress) {
+  var handle = this.handle;
+  if (!handle) return;
+
+  var vertical = this.options()['vertical'];
```
heff: `this.vertical()` here too

heff:
So in this PR we're still adjusting the bar location based on the handle, right?

mmcc:
@heff Yep, [that's correct](https://github.com/videojs/video.js/pull/1816/files#diff-d1c2556bd86fabaee2b77c943b1adcebR96).

mmcc:
ping on this one.

src/js/slider.js
```diff
@@ -78,11 +81,12 @@ vjs.Slider.prototype.update = function(){
   // If scrubbing, we could use a cached value to make the handle keep up with the user's mouse.
   // On HTML5 browsers scrubbing is really smooth, but some flash players are slow, so we might want to utilize this later.
   // var progress =  (this.player_.scrubbing) ? this.player_.getCache().currentTime / this.player_.duration() : this.player_.currentTime() / this.player_.duration();
+  var progress = this.getPercent();
+  var handle = this.handle;
```
heff: Doesn't look like this handle var is being used anymore.

heff:
looks good to me!

Issue #1813, opened 2015-01-20
URL matching bug


H1D:
This is a buggy regexp - https://github.com/videojs/video.js/blob/e8235c9022ae60e656695205948b6de61d440dd2/src/js/media/html5.js#L327

For urls like "https://example.com/video.sd.mp4?s=foo" match group `#1` will be `sd.mp4` instead of expected `mp4`

Example: https://www.debuggex.com/r/qksngI_8cqz6SyCz
![image](https://cloud.githubusercontent.com/assets/697625/5819034/95c1a0ec-a0cb-11e4-9111-42c1b0549bb1.png)

Is it an urgent bug? Should i use `contrib hotfix start`?

H1D:
fix is ready BTW https://github.com/H1D/video.js/commit/a24e79390b426322273df2913e808f1b46a3386f

mmcc:
Thanks for the quick PR! This looks good to me, go ahead and submit a PR whenever you're ready.

On another note, did you end up using contrib? Any thoughts there?

H1D:
@mmcc https://github.com/contrib/contrib is a fantastic idea! Dunno why no one created it before seems like an obvious idea.
Hope @heff will be able to make it as high quality OSS project as video.js is.

PR #1818, opened 2015-01-21
Match vimeo direct urls

H1D:
Sometimes video.js parse url to detect file type. There was a bug in parsing regexp. Details #1813 

Hope `stable` is the right branch to merge. `contrib` doesn't work for me due to this bug https://github.com/contrib/contrib/issues/11

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -324,7 +324,7 @@ vjs.Html5.nativeSourceHandler.canHandleSource = function(source){
     return canPlayType(source.type);
   } else if (source.src) {
     // If no type, fall back to checking 'video/[EXTENSION]'
-    match = source.src.match(/\.([^\/\?]+)(\?[^\/]+)?$/i);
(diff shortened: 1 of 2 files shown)


src/js/media/html5.js
```diff
@@ -324,7 +324,7 @@ vjs.Html5.nativeSourceHandler.canHandleSource = function(source){
     return canPlayType(source.type);
   } else if (source.src) {
     // If no type, fall back to checking 'video/[EXTENSION]'
-    match = source.src.match(/\.([^\/\?]+)(\?[^\/]+)?$/i);
+    match = source.src.match(/\.([^.\/\?]+)(\?[^\/]+)?$/i);
```
heff: I was thinking the period needed to be escaped, but [apparently not](http://stackoverflow.com/questions/10397968/escape-dot-in-a-regex-range).

heff:
looks good!

mmcc:
nice. looks good to me.

Issue #1810, opened 2015-01-18
Error event not documented


thijstriemstra:
The `error` event isn't documented in the player api docs.

mmcc:
Good catch. Any interest in submitting a PR? You can follow the example of one of the [other events](https://github.com/videojs/video.js/blob/master/src/js/player.js#L382) so it will show up in the generated documentation.

thijstriemstra:
Alright. Looking at the source you create a stub method for each event so you can document it in the docstring? PR will follow.

heff:
Closed by #1819

PR #1819, opened 2015-01-21
document the error event

thijstriemstra:
refs #1810

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -589,7 +589,7 @@ vjs.Player.prototype.onFullscreenChange = function() {
 };
 
 /**
- * Fired when an error occurred
+ * Fired when an error occurs
  * @event error
  */
 vjs.Player.prototype.onError;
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -588,6 +588,12 @@ vjs.Player.prototype.onFullscreenChange = function() {
   }
 };
 
+/**
+ * Fired when an error occurred
```
mmcc: This is probably a nitpick, but to be consistent with the other comments, let's go with "Fired when an error occurs"

mmcc:
:thumbsup: Aside from the minor comment, this looks good to me.

src/js/player.js
thijstriemstra: Pushed a fix.

mmcc:
lgtm!

heff:
looks good!

PR #1823, opened 2015-01-22
Update swf version to 4.5.3

bclwhitaker:
Let me know if there is something more I need to do, but we pushed this new version to stable branch of video-js-swf earlier this evening.  I am aware that it still needs to be pushed to the cdn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "main": "./dist/video-js/video.js",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


heff:
Tested it in IE8. Looks good to me. @dmlap confirms that it works for videojs-contrib-hls. I'll pull this in for the patch release.

bclwhitaker:
@heff thanks!

Issue #1817, opened 2015-01-21
Swiping video causes video to play on ipad iOS 8.1


smohadjer:
If I go to videojs.com and swipe video in an attempt to scroll the page the video starts playing! If I pause video and swipe again, video does not start playing. Why video plays on first swipe? How can I avoid video play during page swiping?

mmcc:
I can quasi-confirm this one. On both an iPad and iPhone it works fine if I scroll how I normally do, but if I shorten it to more of a quick, swiping motion then it will trigger a click event most of the time. We should probably decrease the length that's considered a swipe vs click?

smohadjer:
I can accept that a slow swipe be mistaken for a click, but a fast swipe being counted as click is certainly wrong.

mmcc:
Why would you expect an actual dragging motion to be mistaken for a click? I'm having to actively change how I scroll to reproduce this, but it seems like a reasonable movement so I agree that it's probably worth experimenting with a lower [tap movement threshold](https://github.com/mmcc/video.js/blob/master/src/js/component.js#L1041). Want to find a lower number that works for this movement and submit a PR?

smohadjer:
No, I can play video after it's stopped by tapping anywhere on the video on iPad, but I can't play it by swiping. This makes me wonder if the problem lies indeed in swipe logic because then swiping should continue to trigger a click after video is stopped and hence play it, same as a click does.

mmcc:
Well that's strange, I'll hold off opening an issue for that, then...For whatever reason click events on the player itself aren't working for me. Our iPad for testing just died, so I'll play with that more later today.

FWIW, I don't think that's related to the click / swipe confusion. A quick look around for the tap threshold in other projects shows anything from [2](http://hammerjs.github.io/recognizer-tap/) to 11, which even on the larger end is still half what our current tap movement threshold is.

I'll test the element-click issue more later today and open a separate issue if need be.

smohadjer:
I don't think reducing the tap threshold is the right way to better differentiate taps from swipes. Perhaps it would be better if we take velocity also into account. This article explains the problem well:

http://minuum.com/taps-and-swipes/

mmcc:
I think lowering the tap threshold would certainly improve things, but a PR taking velocity into account would certainly be welcome.

Jackipy:
I am hitting the same issue on iPhone iOS 8.1.2 as well.

[...]

PR #1830, opened 2015-01-27
Lower tap movement and time thresholds

mmcc:
Right now our tap movement threshold is much higher than most other libs. As @steverandy said, our touch time threshold is also a bit higher than other libs as well, so dropping that to be more inline with those.

This should resolve issues around swipes being confused with taps: #1817

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1039,7 +1039,7 @@ vjs.Component.prototype.emitTapEvents = function(){
 
   // Maximum movement allowed during a touch event to still be considered a tap
   // Other popular libs use anywhere from 2 (hammer.js) to 15, so 10 seems like a nice, round number.
-  tapMovementThreshold = 22;
+  tapMovementThreshold = 10;
 
   // The maximum length a touch can be while still being considered a tap
   touchTimeThreshold = 200;
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1031,14 +1031,18 @@ vjs.Component.prototype.onResize;
  */
 vjs.Component.prototype.emitTapEvents = function(){
   var touchStart, firstTouch, touchTime, couldBeTap, noTap,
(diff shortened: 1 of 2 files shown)


gkatsev:
LGTM but tests failed.

mmcc:
ha, the tests were passing for me after 30376cd because....I'd never actually committed the threshold change.

heff:
looks good!

Zae:
Is it possible this is still happening on paused videos? I'm seeing that I can't scroll over a paused videojs element without starting the video on ios. This also happens on the video element on the videojs.com site.

Zae:
By the way, it only happens if I touch the big white play button.

mmcc:
So you're saying you can't scroll only if you're touching the big play button? That seems like a pretty specific edge case to me...

Zae:
Scrolling works but the video starts to play after you let go. Seems to be an ios thing, because it happens on a 'naked' video element also. Was thinking maybe videojs can do something about it, but it's probably something very low on the prio list 😀

mmcc:
ahh iOS...it's the source of a few really bizarre bugs like this. Yep, this is definitely down there on the list of things we need to fix, but if you'd like to dig in we'd gladly accept a PR that helps fix it!

MaxLaumeister:
For me swipes are still being mistaken as taps on Chrome for Android in Video.js 5.4.6. My site's mobile view is very dense with videos in some places, so the user is just about guaranteed to put their finger on one of the videos as they scroll past them. When this happens, the video registers a tap and plays.

Here are steps to reproduce on the Video.js homepage using Chrome for Android: Wait until the page loads fully, then casually swipe up on the video to scroll down the page. Unless you make your swipe deliberately slow, the video will start to play as you scroll past it. A user who is just trying to scroll down the page will have videos start randomly playing, which is not ideal.

Here's a video (gif) of it happening on the Video.js homepage:

ezgif-2913097299

PR #1833, opened 2015-01-28
Extended Flash.nativeSourceHandler.canHandleSource to try to to guess the type from the src

carpasse:
We encountered a problem when we tried to run *player.setSrc('someVideoSource')* on IE 8 due to  *Flash.nativeSourceHandler.canHandleSource* not being able to deal with src objects that contains no type.


heff:
Ah, nice. I think we just [fixed this](https://github.com/videojs/video.js/pull/1797/files) in the HTML5 tech as well.

src/js/media/flash.js
```diff
@@ -233,15 +233,24 @@ vjs.Flash.nativeSourceHandler = {};
  * @param  {Object} source  The source object
  * @return {String}         'probably', 'maybe', or '' (empty string)
  */
-vjs.Flash.nativeSourceHandler.canHandleSource = function(source){
+vjs.Flash.nativeSourceHandler.canHandleSource = function (source) {
   var type;
 
-  if (!source.type) {
+  function guessMimeType(src) {
+    var match = src.match(/\.([^\.]+)$/);
```
heff: The [regex in the html5 version](https://github.com/videojs/video.js/blob/master/src/js/media/html5.js#L330) is a little more verbose. Would it be beneficial to use the same one here?

gkatsev: Sounds like we should be using the same function.

heff: Yeah, we could probably add a `getFileExtenson` method to lib.js.

dmlap: Prior art: http://nodejs.org/api/path.html#path_path_extname_p

heff: Can you find the underlying regex?

dmlap: [Sure can](https://github.com/iojs/io.js/blob/v1.x/lib/path.js#L379):

```js
const splitPathRe =
    /^(\/?|)([\s\S]*?)((?:\.{1,2}|[^\/]+?|)(\.[^.\/]*|))(?:[\/]*)$/;
var posix = {};


function posixSplitPath(filename) {
  return splitPathRe.exec(filename).slice(1);
}
```

dmlap:
The discussion ran away from us for a little while there. @carpasse would you be up for integrating the regexp change from my last comment so we can pull this one in?

carpasse:
Sure I will do that today

carpasse:
I have updated the pull request. 
I had quite some problems updating the branch (conflicts everywhere). Please review the modified files before accepting the pull request. It says I have modified 35 files. Most of them are from the dist folder that I had to delete to be able to finally merge.

@dmlap I have added vjs.getFileExtension method to libs and used in canHandleSource on both medias Html5 and Flash. I have extended the regexp that you proposed so that it ignores query parameters.
The test below did not pass with the original regex
```
  equal(vjs.getFileExtension('foo/.bar/test.video.flv?foo=bar'), 'flv');
```
I can revert it and use the original if you think that is unnecessary.

dmlap:
@carpasse filtering query params totally makes sense. This branch looks like it could use some rebasing before it gets pulled in to clean up all the merge commits but that's something we normally do when accepting patches. Thanks!

carpasse:
@dmlap I have having a lot of problems with my git repo and this pull request. I am going to close the close this pull request. Leave my fork in a right state and the do another pull request with fixes "#1833"
Sorry for the trouble

carpasse:
You can find the new pull request here https://github.com/videojs/video.js/pull/1974
This one created with a proper featureBranch

PR #1838, opened 2015-01-30
Feature/control bar separator

mmcc:
I have this in my updated base styles branch for 5.0, but I'm trying to pull components out of that one and submit separate PRs so that's not **too** unwieldy when it comes time for review. Basically this serves a few purposes:

1. Provide an injection point for plugins. Plugins can append any control bar elements they create after this separator and have reasonable assurance of where it will end up in the dom (which in my branch, is after the progress bar / time elements but before the playback rate menu button).
2. Allow for styling in different situations. Right now the base styles branch has the progress bar fill all available space between buttons on the left and right sides of the control bar. If one were to move the progress bar to where it is in the current layout, there would need to be styling work done to fill the empty void and keep the elements to the left / right. With the separator, the separator can be set to auto width in the absence of the control bar and nothing else needs to be changed.


heff:
Holding off on this until we lock down for 5.0.

heff:
Should we more formally name this? "button separator", "insert point", er something?

mmcc:
After conversation:

- Keep generalized separator
- Add specific separator name to default control bar separator instance (`.buttons-separator`)

mmcc:
@heff updated to reflect our most recent conversation.

mmcc:
I'd like to get this pulled in before I keep going on #1999, so can we get another pass? I decided to go with `spacer` instead of `separator` since it seems like it's more commonly used. The "insertion point" bit is still up for debate, but let's make a decision and finish this off.

src/js/control-bar/spacer-controls/insertion-point-spacer.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Spacer specifically meant to be used as an insertion point for new plugins, etc.
+ *
+ * @param {Player|Object} player
+ * @param {Obect=} options
+ */
+class InsertionPointSpacer extends Spacer {
+  buildCSSClass() {
+    return `vjs-insertion-point-spacer-control ${super.buildCSSClass}`;
```
heff: We don't actually need `-control` after every class. It started that way because I was expecting every control to inherit from a control class, but that didn't end up happening.

heff:
I don't see any problems with the code. 

What would the code look like when someone uses this to add a button? That might help with the naming.

heff:
CustomControlInsertPoint?

We should also add a method to the control bar so you can just do `player.controlBar.addCustomControl(something)`.

PR #1840, opened 2015-01-31
Added bower install information

kedar2a:
- As `videojs` in present in `bower` and bower being popular static files manager now-a-days, it's information has been added.
- Developers normally looks for bower install information to get the dependency as I was.
- Hope it would be useful.

### DIFF
--- a/docs/guides/setup.md
+++ b/docs/guides/setup.md
@@ -25,11 +25,17 @@ You can download the Video.js source and host it on your own servers, or use the
 ```
 
 
-### Bower install ###
-```html
-bower install --save videojs
+## Install via package manager
+
+### NPM
 ```
-Three important files (`video-js.min.css`, `video.js`, `video-js.swf`) to link in your html file are contained in the `dist` folder
+$ npm install --save video.js
+```
+
+### Bower
+```
+$ bower install --save video.js
+```
 
 
 ### Self Hosted. ###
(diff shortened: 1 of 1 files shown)


mmcc:
Thank you for the PR! We're going to encourage folks to use NPM / Browserify, but obviously different strokes and such. My concern with starting to add specific instructions like this is our README is going to end up looking like our root directory, full of a billion different manifest files for different managers ( that's an exaggeration, but we do have: `bower.json`, `package.json`, `component.json` and `composer.json`).

That's a wall of text to say, I think we should have instructions for using our "blessed" way (which will eventually be NPM / Browserify), then mention that it's available on these other platforms. At most, instructions for other package managers should probably just be one liners (ish) to avoid clutter.

So, for now, would you mind updating the PR to look something like this:

<pre>
## Install via package manager

### NPM
```
$ npm install --save video.js
```

### Browserify
```
$ browserify install --save video.js
```
</pre>

gkatsev:
There is no `browserify install`. It just uses npm.

mmcc:
that was me not being functional late in the second day of a conference.

gkatsev:
hehe, :D

kedar2a:
Hey, thank you @mmcc for the good suggestion and to keep the text clean and simple.
Updated file accordingly with the commit.

mmcc:
lgtm

Issue #1826, opened 2015-01-24
Rename zh.js to zh-CN.js


Globulopolis:
I think it's better to rename simplified Chinese translation file to zh-CN. Or not?

swordfeng:
Well, it's true... Use *zh_CN* is better,as is used in Linux.

heff:
@DevGavin, any feedback on this?

It depends on whatever are the most common values for the [html tag lang attribute or the navigator defined language value](https://github.com/videojs/video.js/blob/d5bbd80db989ac7a96abfafd461591d32cc44076/src/js/core.js#L125). I'm not sure how to determine that, but if both are common we could have a file for each.

mister-ben:
If the document/browser lang is `zh-CN` and we have a translation for the more generic `zh`, I think that should be used rather than the default.

DevGavin:
Sorry that I was not checking my email for the last few days. 

I think for the language of Chinese "zh", there are two major variations "zh-CN" and "zh-TW". Most people from the two places can understand both variations. I wouldn't worry about the UI confusion brought to people in one place by the language of the other. And I don't think there is a "generic" zh because difference also lies in the form of the characters. I would suggest having both zh-CN and zh-TW translations and name them in "language-countrycode" format. Are there browsers return "zh" only? If so, I would suggest to duplicate one of the two language files used more often. Do we have this kind of statistics?

heff:
It sounds like we're in agreement here. @Globulopolis, would you want to make a pull request for the change?

Globulopolis:
Unfortunately no @heff ( I can only report the issue. If someone can make the PR I would be greatly appreciated.

DevGavin:
@heff What is the agreement here? To rename the current zh to zh-CN and to create another zh-TW version? If yes, I can do this PR. But I would like to confirm with you guys first.

heff:
There's already a [zh-TW](https://github.com/videojs/video.js/blob/master/lang/zh.json), so I think we just need to rename the current zh to zh-CN.

heff:
Sorry, I linked the wrong file for zh-TW.
https://github.com/videojs/video.js/blob/master/lang/zh-TW.json

DevGavin:
Done

PR #1841, opened 2015-02-01
Feature/chs localization

DevGavin:
As discussed.
https://github.com/videojs/video.js/issues/1826#issuecomment-72195425


heff:
I think your branch was based on master prior to adding the zh.json, because this pull request only shows a file addition, not a file name change. I think I can still make this work though. Thanks for the help!!

heff:
Thanks @DevGavin!

PR #1847, opened 2015-02-05
Added @vjs-font-path variable

iSimonWeb:
It's a simple PR.
I just need to be able to change the font path since I'm using video-js via bower and the "font" directory of my project is not a subpath of "css".

Awesome Vanilla plugin anyway! Thank you! :+1:

### DIFF
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -73,13 +73,14 @@ e.g. <video class="video-js my-skin-name">
 The control icons are from a custom font. Each icon corresponds to a character
 (e.g. "\e001"). Font icons allow for easy scaling and coloring of icons.
 */
+@vjs-font-path: 'font';
 @font-face{
   font-family: 'VideoJS';
-  src: url('font/vjs.eot');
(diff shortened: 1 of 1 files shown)


heff:
Thanks for the help! So does this allow you to override the value of `vjs-font-path` then? Is there potentially a cleaner way we should be handling this for people who are including this via bower?

@mmcc thoughts?

mmcc:
@heff No, this doesn't allow you to override the value of `vjs-font-path`. As far as I know, this really just cuts down the number of places you need to type your new font-path (which is helpful).

I'm fine with pulling this in, but it's going to be fairly short lived.

iSimonWeb:
Thank you guys for the attention.
I've seen this approach in the Font-Awesome project
Here is the variable -> https://github.com/FortAwesome/Font-Awesome/blob/master/less/variables.less
And there is where are they using it -> https://github.com/FortAwesome/Font-Awesome/blob/master/less/path.less

Hope you'll appreciate :+1:

heff:
Thanks @iSimonWeb!

PR #1857, opened 2015-02-11
Implement vjs-ended class

### LINKED ISSUE #1675: Add a has-ended class name
We have has-started but not has-ended. Now that we've moved to using class states for showing/hiding the poster image, we don't have a good way to re-show the poster when the video ends. And I'm sure there will be a number of other good uses for it. Plus I mentioned that we *might* have it in a training the other day, but I was wrong. :)

See: http://stackoverflow.com/questions/27022730/video-js-display-poster-image-at-end-of-video

mister-ben:
Adds a `vjs-ended` class when playback ends, and removes it on a replay. Closes #1675.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -446,6 +446,7 @@ vjs.Player.prototype.onLoadedAllData;
  * @event play
  */
 vjs.Player.prototype.onPlay = function(){
+  this.removeClass('vjs-ended');
   this.removeClass('vjs-paused');
   this.addClass('vjs-playing');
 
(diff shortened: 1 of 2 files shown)


heff:
Whoa, with tests even, very nice! I think we need to reset it on loadstart also, because a new source doesn't have to autoplay.

mister-ben:
Ah yes of course it should reset then. Will add.

dmlap:
[The spec](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-ended), for reference.

heff:
lgtm! @dmlap or anyone else, can we get a second confirmation?

dmlap:
Looks good to me, too.

Issue #1505, opened 2014-09-16
Uncaught TypeError: Cannot read property 'vdata1410859035575' of null


pperliti:
I'm trying to use videojs-vast-plugin (https://github.com/theonion/videojs-vast-plugin) to play a preroll video (mp4 over HTML5) before the actual content (rtmp over Flash). 

VideoJS version is 4.8 (hosted locally). Tech order is ["html5","flash"].

The original video is a rtmp stream so the videojs Flash version is instanced during setup. 
Then videojs-vast-plugin downloads a VAST file that tells the plugin to start a simple video download (linear mode), so the plugin updates player's sources accordingly (the initial Flash player is then disposed), the player switches to HTML5 mode and then mp4 adv video is played. 
At then end of the preroll the videojs-vast-plugin restores an initial snapshot of the player (the HTML5 player is then disposed, as well as a new Flash player instance is finally created) and tries to play the (original) rtmp stream.

```
<video id="video-1" class="video-js vjs-default-skin vjs-big-play-centered">
  <source src="rtmp://myserver/video.mp4" type="rtmp/mp4"/>   
</video>

<script>
  var v1 = videojs("video-1",
                    { // options
                      controls: true,
                      preload: "none",
                      width: "640",
                      height: "360",
                      techOrder: ["html5","flash"],
                      plugins: {
                        ads: {
                          debug: true
                        },
                        vast: {
                          url: "http://vast-server/ad/sid=49&bt=n&fmt=fmt_testvideo_smtv"
                        }
                      }
                    },
                    function() {
                      console.log("The video " + this + " is ready");
                    });
</script>
 ```

The problem lies at the end of the preroll. When the HTML5 player is disposed (after the adv video successfully finished playing) and a new Flash player instance is created, it seems that some event handler is still alive for the very first Flash player instance created during the initial setup (and then an error is thrown): 

```javascript
Uncaught TypeError: Cannot read property 'vdata1410859035575' of null.
```

The error is throwed by the vjs.getData() function (see [1] in the following code) since is el[vjs.expando] (vjs.expando value is 'vdata1410859035575') is null:

```javascript
vjs.getData = function(el){
  var id = el[vjs.expando];  // <-- /* [1] */
  ...
};
```

Debugging the scripts I ended to face a strange behaviour. When the videojs-vast-plugin tries to restore the initial Flash Player (see [2] in the following code)

video-vast.js (videojs-vast-plugin)
```javascript
player.vast.tearDown = function() {
  ...
[...]

Issue #1485, opened 2014-09-04
Change techs to emit events instead of triggering them on the player


heff:
We currently have a mixed model where the player sets/gets properties on a tech while the tech triggers events on the player. It would be a big change (so tech 2.0/ vjs 5.0) but I think it would make the definition of a tech much cleaner if the player instead listened for tech events. Open to other thoughts on that.

Related: #863 #1194

Issue #1484, opened 2014-09-04
vjs.hasData Error Spam


Manbearpixel:
Developing a video playing website on AngularJS and using videoJS as the video player. The video player is contained within a controller, which gets destroyed and comes back upon a new video being loaded. In angular, I can run functions prior to a controller being destroyed, this is where I call videoObject.dispose() 

**Expectation:**
I expected `videoObject.dispose()` to handle all the necessary removal of information related to the video player.

**Reality:**
It seems that if a video is exited (controller is destroyed) prior to the video fully loading, an error is thrown repeatedly from `vjs.hasData` (see screenshot below). This error seems to be thrown every second (assuming an event is still active somewhere that wasn't cleared) I tried tracking down the `.dispose()` function and commenting out `this.trigger({ type: 'dispose', 'bubbles': false });` but that seems to have no effect. If I try checking that el is not null or wrapping it in a try/catch it still goes through but instead the error comes from `vjs.trigger` (see screenshot below) 

At this point I am unsure how to debug this and am curious if anyone else has encountered this problem. I will try to create a plunkr that might replicate this issue since the site is currently not live.

**vjs.hasData Screenshot**
![screen shot 2014-09-04 at 12 27 07 pm](https://cloud.githubusercontent.com/assets/6686750/4156473/d2c5b71c-3474-11e4-95b8-b4801d2b3501.jpg)

**vjs.trigger Screenshot**
![screen shot 2014-09-04 at 3 47 04 pm](https://cloud.githubusercontent.com/assets/6686750/4156476/d90a05c4-3474-11e4-9d46-7709a8935c9a.jpg)

**videoPlayer.dispose() call**
![screen shot 2014-09-04 at 3 51 51 pm](https://cloud.githubusercontent.com/assets/6686750/4156535/53a686ae-3475-11e4-9424-7fe30ce7fd41.jpg)

benjipott:
new on this issue

mmcc:
This shouldn't be an issue anymore, it should have been closed along with #1656 and #1642. @benjipott can you put together a reduced test case?

gkatsev:
Generally if you are seeing this, it means there's some other *unrelated* bug further up the chain.

saxena-gaurav:
this should have been fixed by #2125

PR #1859, opened 2015-02-11
Tech change handler spam

### LINKED ISSUE #882: Switching between sources with diffrenet technologies causes problem
In my case this is the techOrder:

```javascript
techOrder : ['html5', 'flash']
```

everything works fine while I'm using file types which are supported by Html5 tech, however as soon as I pass on a file type which is not supported by html5 and it requires  the tech to be switched to Flash everything start acting weird for example the play progress bar doesn't work anymore or the currrentTime always stays on zero.

My best guess here is that Flash player couldn't trigger events anymore.

benjipott:
Disables spam error handlers when changing tech
Fixes #882 ~~#1485~~ #1505 #1484

### DIFF
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -53,20 +53,14 @@ vjs.MediaTechController = vjs.Component.extend({
  * any controls will still keep the user active
  */
 vjs.MediaTechController.prototype.initControlsListeners = function(){
-  var player, activateControls;
+  var player;
 
   player = this.player();
 
-  activateControls = function(){
-    if (player.controls() && !player.usingNativeControls()) {
-      this.addControlsListeners();
-    }
-  };
-
   // Set up event listeners once the tech is ready and has an element to apply
   // listeners to
-  this.ready(activateControls);
-  this.on(player, 'controlsenabled', activateControls);
+  this.ready(this.activateControls);
+  this.on(player, 'controlsenabled', this.activateControls);
   this.on(player, 'controlsdisabled', this.removeControlsListeners);
 
   // if we're loading the playback object after it has started loading or playing the
@@ -79,6 +73,15 @@ vjs.MediaTechController.prototype.initControlsListeners = function(){
       this.player().trigger('loadstart');
     }
   });
+};
+
+vjs.MediaTechController.prototype.removeControlsListeners = function () {
+  var player;
+
+  player = this.player();
+
+  this.off(player, 'controlsenabled', this.activateControls);
+  this.off(player, 'controlsdisabled', this.removeControlsListeners);
 };
 
 vjs.MediaTechController.prototype.addControlsListeners = function(){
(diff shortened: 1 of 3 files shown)


src/js/media/media.js
```diff
@@ -81,6 +75,15 @@ vjs.MediaTechController.prototype.initControlsListeners = function(){
   });
 };
 
+vjs.MediaTechController.prototype.removeControlsListeners = function () {
```
heff: There's already [a removeControlsListeners](https://github.com/benjipott/video.js/blob/hotfix/switch-between-tech/src/js/media/media.js#L126). Should this be combined with that?

heff:
Thanks for the help! This looks like a good update to me. Any interested in writing a test for this? It doesn't appear we have any tests for this yet, but it would go in the [media tech tests file](https://github.com/videojs/video.js/blob/789c72945a37d637655fcb5dbb2f3cbf7013f5e3/test/unit/media.js).

src/js/media/media.js
heff: I think there's a problem with this. It will remove the removeControlsListeners and activateControls listeners when `controlsdisabled` is triggered, but what if `controlsenabled` is triggered again? I think this will break that use case.
e.g.
- enable controls
- disable controls
- enable controls again

heff:
Have you verified that this fixes the issue, and can you describe how this is fixing the problem?

src/js/media/media.js
```diff
@@ -239,7 +256,8 @@ vjs.MediaTechController.prototype.stopTrackingCurrentTime = function(){
   this.player().trigger('timeupdate');
 };
 
-vjs.MediaTechController.prototype.dispose = function() {
+vjs.MediaTechController.prototype.dispose = function(){
+  this.removeControlsListeners();
```
heff: This actuallyl shouldn't be needed. When you use `this.off(player, 'controlsenabled', this.activateControls);` it should automatically remove the activateControls listener from the player when this tech is disposed. It's [a feature of the off method](https://github.com/videojs/video.js/blob/master/src/js/component.js#L562).

vasklund:
Just wanted to chip in and say I've experienced several of the problems this pull is supposed to solve, and I think the root problem lies deeper. I've tried the best I could to explain my findings in short in a new issue at https://github.com/videojs/video.js/issues/1896, hopefully you can make sense of my description there.

The problems this pull solves are related to these event handlers (MediaTechController.initControlsListeners)
```
this.on(player, 'controlsenabled', activateControls);
this.on(player, 'controlsdisabled', this.removeControlsListeners);
```
but the real problem is that these handlers, among others, should have been automatically deregistred by the Component.on() logic in response to the dispose event.

mmcc:
I'm going to close this in lieu of @vasklund's investigation in #1896. We shouldn't pull in things to fix a symptom that we should solve upstream. Thanks for the PR, though @benjipott!

PR #1865, opened 2015-02-13
Disable native text tracks by default in html5

gkatsev:
Chrome, and possibly some other browsers, show captions by default, which is unwanted behavior.
Instead, on loadstart, change the mode of each captions or subtitles track to 'disabled'.
This is done in loadstart because trying to change data of text tracks
wholesale any earlier will be a no-op.

This has been tested with autoplaying videos in chrome.

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -59,6 +59,10 @@ vjs.Html5 = vjs.MediaTechController.extend({
       }
     }
 
+    if (this['featuresNativeTextTracks']) {
+      this.on('loadstart', vjs.bind(this, this.hideCaptions));
+    }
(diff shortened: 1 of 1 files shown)


heff:
lgtm!

heff:
Safari also shows captions by default, but not subtitles.

There's a [related Chrome bug for this](https://code.google.com/p/chromium/issues/detail?can=2&start=0&num=100&q=captions%20showing&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified&groupby=&sort=&id=388588).

PR #1868, opened 2015-02-16
Added background-color to vjs-poster to remove transparent borders around scaled poster image

tjenkinson:
Just found a minor issue that is when you have a cover art image that doesn't fit the player and it is scaled down, the player element is visible outside the scaled image.

This tweak sets the background-color property which means the player will always be completely covered if the poster does not fit exactly, or if the poster image cannot be loaded for some reason/whilst it's loading.

Here is what happens currently:
image

And here is what it looks like with the background-color property applied:
image

### DIFF
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -924,6 +924,7 @@ body.vjs-full-window {
   background-repeat: no-repeat;
   background-position: 50% 50%;
   background-size: contain;
+  background-color: #000000;
   cursor: pointer;
   margin: 0;
   padding: 0;


mmcc:
Interesting, thanks for the PR! I seem to remember some issue that went the other direction on this one, but it's totally escaping me. I'll look around a bit tomorrow, but otherwise this looks good to me.

mmcc:
I still have this nagging feeling that there's some edge case we're not thinking through, but I've looked through the issues a few times and don't see anything. I might be going crazy, so lgtm unless someone else disagrees.

heff:
This makes sense to me, however the poster is handled differently in older browsers that don't support background image, so we need to check if this works in IE8 too. @tjenkinson do you have IE8 available to test on?

gkatsev:
@tjenkinson IE8 emulation is not the same. You should grab a VM from http://modern.ie to play around with.

mmcc:
@tjenkinson fwiw, I have a VM and tried to load up your reduced test case. Looks like there's something unrelated to VJS causing IE8 issues :(

mmcc:
@tjenkinson have you had a chance to test this out yet in IE8?

gkatsev:
You need to install flash in the VMs. They don't come with it.

mmcc:
I would stick with [JSBin](http://jsbin.com/axedog/latest/edit), I think there are other issues in IE8 with our general site layout. Just remember you have to view the output in "standalone mode" so that Flash works (just remove "edit" from the URL).

FWIW, I'm astonished the site doesn't crash IE8 for you...

mmcc:
@tjenkinson Welp...this one totally fell off the radar there. Would you mind re-opening this against stable so we can pull it in with the next 4.12 patch release? I can pull this into the Sass in Master.

heff:
@mmcc this doesn't look like it was ever pulled into master. Can you confirm?

PR #1873, opened 2015-02-17
Improve localisation

mister-ben:
A couple of improvements to localisation:

* Try primary code before default string, e.g. if there's no `de-DE` translation, use `de` if available. 
* Treat language codes case insensitively. Safari uses `en-us` but Chrome and Firefox use `en-US`, these should be treated the same.

This should match how CSS applies styles based on lang attributes: http://jsbin.com/geviwa/edit

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -209,14 +209,26 @@ class Component {
     return Lib.createEl(tagName, attributes);
   }
 
+  /**
+   * Return a localised string if available
+   *
+   * @param  {String=} string  String to be localised
+   * @return {String}
+   */
   localize(string) {
-    let lang = this.player_.language();
+    let lang = ('' + this.player_.language()).toLowerCase();
+    let primaryCode = lang.split('-')[0];
     let languages = this.player_.languages();
 
-    if (languages && languages[lang] && languages[lang][string]) {
+    if (!languages) {
+      return string;
+    }
+    if (languages[lang] && languages[lang][string]) {
       return languages[lang][string];
     }
-
+    if (primaryCode !== lang && languages[primaryCode] && languages[primaryCode][string]) {
+      return languages[primaryCode][string];
+    }
     return string;
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -96,12 +96,7 @@ class Player extends Component {
     this.language_ = options['language'] || Options['language'];
 
     // Update Supported Languages
-    var languages = vjs.obj.copy(options['languages'] || vjs.options['languages']);
-    vjs.options['languages'] = {};
-    vjs.obj.each(languages, function(name, value) {
-      vjs.addLanguage(name, languages[name]);
-    });
-    this.languages_ = vjs.options['languages'];
+    this.languages_ = this.options['languages'];
 
     // Cache for video property values.
     this.cache_ = {};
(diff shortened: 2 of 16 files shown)


src/js/component.js
```diff
@@ -201,12 +201,25 @@ vjs.Component.prototype.createEl = function(tagName, attributes){
...
-  if (languages && languages[lang] && languages[lang][string]) {
+  var lang = ('' + this.player_.language()).toLowerCase();
+  var languages = this.player_.languages();
+  if (!languages) {
+    return string;
+  }
+  if (languages[lang] && languages[lang][string]) {
     return languages[lang][string];
   }
+  var primaryCode = lang.split('-')[0];
```
dmlap: Though I think no one likes it at this point, most of the rest of the code declares variables at the top of the scope.

src/js/player.js
```diff
@@ -49,7 +49,12 @@ vjs.Player = vjs.Component.extend({
     this.language_ = options['language'] || vjs.options['language'];
 
     // Update Supported Languages
-    this.languages_ = options['languages'] || vjs.options['languages'];
+    var languages = vjs.obj.copy(options['languages'] || vjs.options['languages']);
+    vjs.options['languages'] = {};
```
dmlap: Are you taking the player level language definitions and copying them out to the `window.videojs` level here? I'm not following. Maybe a comment above would help.

test/unit/component.js
```diff
@@ -590,3 +590,27 @@ test('should provide interval methods that automatically get cleared on componen
...
+        'Good': 'Awesome',
+        'Error': 'Problem'
+      }
+    }
+  });
+
+  var comp = new vjs.Component(player);
+
+  player.language('en-gb');
+  ok(comp.localize('Good') === 'Brilliant', 'Used subcode specific localization');
```
dmlap: You can use `strictEqual()` (for `===` equality) or `equal()` (for `==`) instead of `ok()` to get better error messages when these assertions fail.

src/js/player.js
mister-ben: Wrong solution to the problem. Or solution to the wrong problem. I need to rework that.

I've now realised that although `this.languages_` is either `options['languages']` or `vjs.options['languages'];`, `addLanguage()` always adds to `vjs.options['languages'];`. So a plugin which adds a language might not be able to add it to the player using it, but would add it to others on the page: http://jsbin.com/nomune/1/edit?html,console

dmlap: Ah, ok. There are two levels of options in video.js: page-level and player-level. I believe this code was originally intending to use player-level language settings if provided and fall back to the page-level language settings otherwise.

heff:
It sounds like there's still a little reworking to be done here, but the overall improvements sound good to me.

thijstriemstra:
+1!

heff:
@mister-ben, available to finish this one up? Let me know if you have any additional questions.

mister-ben:
I've been a bit swamped with other things, but do intend to revisit this soon.

heff:
Hey @mister-ben, still want to help finish this one up?

mister-ben:
Having another look now after getting it rebased.

mister-ben:
Re-done as #2177

PR #1874, opened 2015-02-17
Store tracks on player

gkatsev:
Previously, we were storing the text tracks on the techs. However, if you switched techs, and you were emulating captions, you would end up getting new text tracks. This meant that UI elements which registered against the initial tech's text tracks would no longer get any events.

### DIFF
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -345,13 +345,13 @@ vjs.MediaTechController.prototype.emulateTextTracks = function() {
 vjs.MediaTechController.prototype.textTracks_;
 
 vjs.MediaTechController.prototype.textTracks = function() {
-  this.textTracks_ = this.textTracks_ || new vjs.TextTrackList();
-  return this.textTracks_;
+  this.player_.textTracks_ = this.player_.textTracks_ || new vjs.TextTrackList();
+  return this.player_.textTracks_;
 };
 
 vjs.MediaTechController.prototype.remoteTextTracks = function() {
-  this.remoteTextTracks_ = this.remoteTextTracks_ || new vjs.TextTrackList();
-  return this.remoteTextTracks_;
+  this.player_.remoteTextTracks_ = this.player_.remoteTextTracks_ || new vjs.TextTrackList();
+  return this.player_.remoteTextTracks_;
 };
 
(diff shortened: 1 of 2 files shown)


src/js/media/media.js
```diff
@@ -345,13 +345,13 @@ vjs.MediaTechController.prototype.emulateTextTracks = function() {
 vjs.MediaTechController.prototype.textTracks_;
 
 vjs.MediaTechController.prototype.textTracks = function() {
-  this.textTracks_ = this.textTracks_ || new vjs.TextTrackList();
-  return this.textTracks_;
+  this.player_.textTracks_ = this.player_.textTracks_ || new vjs.TextTrackList();
```
dmlap: Should this logic be bumped into player.js?

heff:
The test turns off the native text tracks, which are the default for HTML5 video. What happens if you don't do that and switch to/from Flash?

gkatsev:
Yeah, I discussed this with @dmlap last night.
We definitely don't want to bump the texttracks methods out to the player because at least in native text tracks in html5, we don't want a direct reference to the tech element in the player.
However, this current fix would only cover emulated text tracks for html5 switching to another tech that also uses emulated text tracks. If you're switching between one tech that is using native tracks and then switching to another tech with non-native techs, you would lose the tracks.
One thing that's possible to do is to copy over the previous tracks over in that case but I'm not sure it's a good idea. It would depend on how frequent we think someone will be switching techs while playing the *same* video.

Perhaps in addition to this, we also want to have things like the caption buttons listen to loadstart and if the tracks changed, re-attach their listeners?

gkatsev:
However, regardless of the extra corner cases, I think this particular PR *is* good to go, as it does fix an issue. There's just possibly more things needed to fix.

gkatsev:
I'll open up a separate issue to discusse the native<->non-native tracks.

heff:
lgtm

@videojs/core-committers now that we're locked down for 5.0 make sure any fixes meant for patches are opened against the stable branch, because the master branch is about to get very different. The `contrib patch start` and `contrib patch submit` commands can make that process easier.

(@gkatsev I'll handle moving the open PRs)

gkatsev:
Thanks. I'll try and remember that for future patches :)

PR #1876, opened 2015-02-18
playbackRate takes a number

dmlap:
The playbackRate() function does not take a boolean, it takes a (double-precision) number.


mmcc:
I disagree!

lgtm.

heff:
lgtm

PR #1890, opened 2015-02-24
Loop through track elements for default attr

### LINKED ISSUE #1884: "default" attribute for text tracks not work
In 4.12.0 (and 4.12.1), default text track feature is not available. Subtitles (or captions) always off when the player is loaded.

gkatsev:
Use track's track attribute to refer to the text track itself
but use the track element to check the default attribute to decide
whether to disable or not to disable.

This should be mostly good to go. My home machine's chrome wasn't working quite right, so, someone (me) should double check that it works as expected.

Fixes #1884.


gkatsev:
Oops, opened up against master instead of stable.

PR #1897, opened 2015-02-25
Added new translations (Bosnian, Serbian, Croatian)

dn5:
Added new translations:
- Official language of Bosnian and Herzegovina (ba)
- Official language of Republic of Serbia (sr)
- Official language of Croatia (hr)

* All three languages are identical in translation process (they use same words)


mmcc:
Thanks! LGTM.

PR #1899, opened 2015-02-26
Add Danish translation

OleLaursen:
Here's a Danish translation of the texts.


heff:
Nice, thank you!

heff:
@OleLaursen this has been merged into master. Thanks for the help!

PR #1905, opened 2015-02-27
Fixed setting the source to an empty string

### LINKED ISSUE #1785: Setting src to an empty string throws an error.

heff:
Fixes #1785

### DIFF
--- a/test/unit/media.js
+++ b/test/unit/media.js
@@ -229,3 +229,25 @@ test('should add the source hanlder interface to a tech', function(){
   tech.dispose();
   ok(disposeCalled, 'the handler dispose method was called when the tech was disposed');
 });
+
+test('should handle unsupported sources with the source hanlder API', function(){
+  var mockPlayer = {
+    off: this.noop,
+    trigger: this.noop
+  };
+
+  // Define a new tech class
+  var Tech = videojs.MediaTechController.extend();
+  // Extend Tech with source handlers
+  vjs.MediaTechController.withSourceHandlers(Tech);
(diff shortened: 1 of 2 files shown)


test/unit/media.js
```diff
@@ -229,3 +229,26 @@ test('should add the source hanlder interface to a tech', function(){
...
+    trigger: this.noop
+  };
+
+  // Define a new tech class
+  var Tech = videojs.MediaTechController.extend();
+  // Extend Tech with source handlers
+  vjs.MediaTechController.withSourceHandlers(Tech);
+
+  Tech.nativeSourceHandler = {
+    handleSource: function(){ usedNative = true; }
```
mmcc: I don't love the weird indirection here. Why not just do the assertion right in this callback and set the number of expected assertions to 1?

mmcc:
Otherwise :thumbsup:

heff:
Cleaned up the test order

test/unit/media.js
```diff
@@ -229,3 +229,25 @@ test('should add the source hanlder interface to a tech', function(){
   tech.dispose();
   ok(disposeCalled, 'the handler dispose method was called when the tech was disposed');
 });
+
+test('should handle unsupported sources with the source hanlder API', function(){
```
gkatsev: handler*

Issue #1232, opened 2014-05-22
Expose utility functions


heff:
I'm opening this issue to discuss the idea of exposing more utility functions, e.g. vjs.on() and vjs.obj.each(). If we decide to expose more, we'll use this issue to discuss the right way to do that.

### Background
To date we've avoided exposing too many top-level utility functions because we're not trying to build and support a general use case utility library like jquery or underscore/lodash. Limiting the utility functions to just the capabilities we need in the internal video.js codebase helps keep them small, focused, and performant.

We've also never discouraged anyone from using jquery alongside video.js or even building plugins/components that require it. So "you can use jquery" has been one of the excuses for not exposing the internal utilities.

Many of these internal utility functions are already available when extending video.js components, e.g. Button.prototype.on(), but just not available to be used just on their own, e.g. vjs.on(any_element, 'event', fn).

The addition of plugins have encouraged a lot of new and interesting development on top of video.js specifically, and raises the question of whether or not it's still best to keep these functions private and require plugins to either use a lib like jquery/underscore or duplicate the same functionality manually when needed.

### Discuss
- Do the old reasons for keeping utility functions private still stand up? 
- Are there any other positives/negatives to either side that aren't mentioned here?

### UPDATE 2015-03-16
We're now in development of v5.0 and have removed Closure Compiler's aggressive variable mangling, as well as started to add ES6 modules + Browserify. So all properties of `videojs` will be publicly available (not mangled) but we have new options for creating internal utility modules that don't get exposed. The conversation is still essentially the same but the implementation will look different.

heff:
@videojs/core-committers we need to revisit this one now that we're in 5.0 development, and it applies to PR #1909.

Last comments in IRC were from @gkatsev and @dmlap about putting everything on the root object, `videojs.something`. Reason being it would force us to keep the API small and well defined.

When @mmcc and I were previously discussing we liked grouping functions under `videojs.util`. Partly to make it clear that videojs is not a utility library like lodash or jquery. @mmcc may be able to add other details.

With 5.0 we're making a big push improving the plugin authoring experience, so I want to make sure we're taking that into consideration with what we're doing here. For example do we add extra functions because of convenience even if they're not specifically used inside core? @dmlap and @gkatsev might actually be the best to speak to that. Would there be any utilities that would come in handy as a plugin author? Or what if you always had a reference to lodash and could expect those functions to always be there for plugins?

[...]

PR #1909, opened 2015-02-28
Added vjs.extend method

carpasse:
Added ```extend``` method to be able to easily merge options when you  create a plugin.

### DIFF
--- a/src/js/lib.js
+++ b/src/js/lib.js
@@ -849,3 +849,22 @@ vjs.arr.forEach = function(array, callback, thisArg) {
 
   return array;
 };
+
+vjs.extend = function (obj) {
+  var arg, i, k;
+  for (i = 1; i < arguments.length; i++) {
+    arg = arguments[i];
+    for (k in arg) {
+      if (arg.hasOwnProperty(k)) {
+        if (vjs.obj.isPlain(obj[k]) && vjs.obj.isPlain(arg[k])) {
+          obj[k] = vjs.extend({}, obj[k], arg[k]);
+        } else if (vjs.obj.isPlain(arg[k])) {
+          obj[k] = vjs.extend({}, arg[k]);
+        } else {
+          obj[k] = arg[k];
+        }
+      }
+    }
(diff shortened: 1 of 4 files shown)


heff:
@carpasse, FYI, there's a related conversation in #1232 if you have any thoughts.

heff:
Just commenting on the status of this. We're working on [a major refactor for 5.0](https://github.com/videojs/video.js/pull/1976). As part of that we're going to be reorganizing utils, and pulling in lodash to replace a lot of the custom functions. This may or may not be affected by that, but we need to wait to see how that plays out first. #1232 is where discussions around that will happen.

mmcc:
@carpasse So I think we've landed on pulling in [lodash](http://lodash.com) for whatever utilities we can, so this is a pretty prime usage. I think step 1 is to make use of your extensive tests to ensure that it's a drop in replacement (should be), and if all is well have you start the process of importing / exporting lodash methods.

@heff is currently working on splitting apart the lib stuff into different files, so you guys might want to coordinate on timing of this.

gkatsev:
Actually, for extend specifically, we should use `Object.assign`. There's a shim for it available on npm: https://www.npmjs.com/package/object.assign
It allows local usage of object.assign without shimming it globally:
```js
var assign = Object.assign || require('object.assign');
```
I think this is the best course of action, if possible. lodash has a lot of useful stuff but if the language has it (or is getting it), we should try and use it instead.

heff:
Ok, so as part of the 5.0 work we exported [videojs.mergeOptions](https://github.com/videojs/video.js/blob/fdd133906cc1a33ee81303989555cca843d41898/src/js/video.js#L253), and we'll be exporting `videojs.assign` also. @carpasse I think that covers everything here, but if it doesn't let me know and we'll reopen. I'm just trying to move quickly through the remaining 5.0 tasks so we can get it released.

carpasse:
@heff I thing it looks good. Thanks for the efforts

Issue #1895, opened 2015-02-25
Unable to `npm install` with git dependency vtt.js


silverwind:
I'm unable to install video.js's dependencies with the latest update that pulls in the git dependency of `vtt.js`. The issue seems to be something about git itself, which is at the latest version provided by Cygwin. Same installation works fine on OS X. Here's the error:

````
npm ERR! Command failed: git -c core.longpaths=true clone C:\Users\silverwind\AppData\Roaming\npm-cache\_git-remotes\https-gith
ub-com-gkatsev-vtt-js-git-9a867bb9 C:\Users\SILVER~1\AppData\Local\Temp\npm-3612-ea415d7b\git-cache-aa2cb630bc7b\3ba274b332de18
7f7269f277733021be3dd1f5bd
````

And my versions:
````
$ npm -v
2.5.1
$ git --version
git version 2.1.4
````

Would it maybe be possible to publish `vtt.js` to npm to work around this, maybe under a dist-tag?

heff:
I don't totally understand the error that's happening there, but why might publishing to npm fix this? @gkatsev any insight?

silverwind:
@heff because npm dependencies are just tarball downloads, no git involved.

heff:
Ah, that makes sense. I think we might just be waiting on mozilla/vtt.js#338 to be merged in before we can point directly at that project. @gkatsev will have a better idea of where that's at, and if it's worth publishing his own branch.

silverwind:
You could also depend on a tarball link from GitHub like this:

````js
  "dependencies": {
    "videojs-swf": "4.5.3",
    "vtt.js": "https://github.com/gkatsev/vtt.js/tarball/3ba274b332de187f7269f277733021be3dd1f5bd"
  },
````

mmcc:
Closing this based on the result of #1911

gkatsev:
Fixed by https://github.com/videojs/video.js/pull/2905

silverwind:
:+1:

PR #1911, opened 2015-02-28
use tarball dependency of vtt.js

silverwind:
This uses the tarball download instead of git to obtain the vtt.js fork, fixing my issue encountered in #1895.

Technically, there is no difference, except that npm apparently doesn't strip down the dependency as defined by `files` when downloaded as tarball, but that seems like a npm bug of its own.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "main": "./dist/video-js/video.js",
   "dependencies": {
     "videojs-swf": "4.5.3",
-    "vtt.js": "git+https://github.com/gkatsev/vtt.js.git#shim-build"
+    "vtt.js": "https://github.com/gkatsev/vtt.js/archive/shim-build.tar.gz"
   },
   "devDependencies": {
     "calcdeps": "~0.1.7",


heff:
Seems ok to me. @gkatsev?

gkatsev:
I don't have an issue per-se other than it's lame that cygwin ships a broken git implementation. Assuming that this is pointing at the correct branch of my fork, it's fine.

silverwind:
I'll dig deeper once time permits to get to the root of that Cygwin issue.

gkatsev:
Also, possibly an npm issue as well.

silverwind:
npm does execute the correct command at least, it's probably the shell or git itself that wrongly joins the `c:\\`-style paths when cloning from local to local repo.

gkatsev:
I'm going to close this one, especially since I don't want to point directly at a branch anymore and we have a new change for vttjs (https://github.com/videojs/video.js/pull/2327). Hopefully soon we'll be able to just use the official vttjs npm repo or I could always just publish a vjs-vtt-only fork onto npm.

silverwind:
Fine with me, given this is more of a personal issue. Looking forward to you using the official package soon.

gkatsev:
Fixed by https://github.com/videojs/video.js/pull/2905

PR #1916, opened 2015-03-02
Randomizing the Google Analytics calls to stay under the limit

heff:
GA is cutting us off, so we need to get this in soon.


mmcc:
What should we do about people using our GA code due to copying the CDN version?

heff:
Nothing we can do ourselves. I tried contacting one of the big sites doing it through a web form, so big hopes there. Also asked Google Analytics if there was anything we could do, but no word there yet.

PR #1924, opened 2015-03-04
Added `useractive` and `userinactive` events to API docs

tamlyn:
Had to dig through the source to find these. Don't know what to put for _defined in_ since these events are just triggered without any real definition.

### DIFF
--- a/docs/api/vjs.Player.md
+++ b/docs/api/vjs.Player.md
@@ -113,6 +113,8 @@ After an instance has been created it can be accessed globally using `Video('exa
   - [seeked](#seeked-event)
   - [seeking](#seeking-event)
   - [timeupdate](#timeupdate-event)
+  - [useractive](#useractive-event)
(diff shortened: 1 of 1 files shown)


gkatsev:
They definitely should be documented but I think our process for documentation is different.
@mmcc and @heff would know more.

mmcc:
Good catch, thanks for the contribution @tamlyn.

Our docs are auto generated from [JSDoc-style](http://usejsdoc.org/index.html) comments, so these additions would need to be added to the code somewhere. Since these events aren't explicitly defined yet, you can add some [dummy methods like these](https://github.com/videojs/video.js/blob/7bcb1d218c764d4d2729af0bfd589e1b62158b58/src/js/player.js#L405-L421) and just add the comments above them.

tamlyn:
Ah, OK. Have created new PR #1926.

PR #1927, opened 2015-03-05
Legend in pt-Br is wrong.

juniorlisboa:
Legend in pt-BR is wrong.
Progress is not "Progressão" in portugues, the correct word is "Progresso".


heff:
@juniorlisboa thanks!
@deedos could you review this change?

heff:
Ok, I think @deedos has had enough time. This can be merged in.

Issue #1928, opened 2015-03-05
Improve error for missing plugin


thijstriemstra:
When you try to load a plugin and forgot to include the actual .js file for that plugin, the following error is thrown:

```
TypeError: this[a] is not a function
http://cdnjs.cloudflare.com/ajax/libs/video.js/4.12.3/video.js
Line 59
```

Surely this can be handled and a more appropriate error message is shown.

gkatsev:
If I'm understanding the question correctly, there isn't much we can do for better handling for this, unfortunately. JS doesn't provide a mechanism for a "method missing" so we could print a better error. However, I know that v8 and chrome are updating this error output to actually print out the name of the function.
If I'm misunderstanding you, can you please provide a more detailed example of when you're getting the error?

thijstriemstra:
This should trigger the error:

```javascript
var player = videojs("myClip",
{
    controls: true,
    autoplay: true,
    loop: false,
    width: 600,
    height: 300,
    plugins: {
        myNonExistingPlugin: {
            foo: 'bar'
        }
    }
});
```

gkatsev:
Ok, cool. So, I was partially right.
We probably can do a better job when the plugins are included in the config to be initialized.
We cannot do a better job when you initialize the plugin yourself like so:
```js
var player = videojs('foo');
player.myNonExistingPlugin({
  foo: 'bar'
});
```

thijstriemstra:
Sounds good. Your use-case wouldn't work for me anyway because the plugin load order gets screwed up when you assign it dynamically like that.

gkatsev:
If you feel like opening a PR for your case, that would be awesome.

thijstriemstra:
I've implemented a fix and now trying to add a unit test. It was giving some issues with the `throws` keyword so I thought I upgrade both `grunt-contrib-jshint` and `jshint` itself in my video.js checkout. Guess it only made it worse because now I'm seeing:

```
$ npm install jshint
npm WARN package.json karma-ie-launcher@0.1.5 No README data
npm WARN package.json karma-phantomjs-launcher@0.1.4 No README data
npm WARN package.json karma-qunit@0.1.4 No README data
npm WARN package.json karma-sauce-launcher@0.2.10 No README data
npm WARN prefer global jshint@2.6.3 should be installed with -g
jshint@2.6.3 node_modules/jshint
├── strip-json-comments@1.0.2
├── underscore@1.6.0
├── exit@0.1.2
[...]

PR #1931, opened 2015-03-06
throw error when plugin is missing

thijstriemstra:
Improve error for missing plugin, refs #1928.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -99,7 +99,11 @@ vjs.Player = vjs.Component.extend({
 
     if (options['plugins']) {
       vjs.obj.each(options['plugins'], function(key, val){
-        this[key](val);
+        try {
+          this[key](val);
+        } catch (err) {
+          throw new Error('Unable to find plugin: ' + key);
+        }
       }, this);
     }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -97,12 +97,13 @@ vjs.Player = vjs.Component.extend({
     // Make player easily findable by ID
     vjs.players[this.id_] = this;
 
+    // scan plugins
     if (options['plugins']) {
       vjs.obj.each(options['plugins'], function(key, val){
-        try {
+        if (typeof this[key] === 'function') {
           this[key](val);
-        } catch (err) {
-          throw new Error('Unable to find plugin: ' + key);
+        } else {
+          vjs.log.error('Unable to find plugin:', key);
         }
       }, this);
     }
--- a/.jshintrc
+++ b/.jshintrc
@@ -37,6 +37,7 @@
         "ok",
         "QUnit",
         "raises",
+        "throws",
         "start",
         "stop",
         "strictEqual",
--- a/test/unit/plugins.js
+++ b/test/unit/plugins.js
@@ -163,17 +163,33 @@ test('Plugins should not get events after stopImmediatePropagation is called', f
   player.dispose();
 });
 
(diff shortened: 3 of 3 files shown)


src/js/player.js
```diff
@@ -99,7 +99,11 @@ vjs.Player = vjs.Component.extend({
 
     if (options['plugins']) {
       vjs.obj.each(options['plugins'], function(key, val){
-        this[key](val);
+        try {
```
gkatsev: rather than using try/catch, we can see whether `this[key]` is a function and then log an error. This way it continues working even if a plugin is missing but we still have a better error log.

thijstriemstra: I was trying to figure out what the right way to go was here but, in my case, my video.js player is just an empty shell without the plugin and I want to abort further loading. It sounds like logging it would not help me determine if the video.js player (and required plugin) is broken?

thijstriemstra:
Could anyone help me figure out why this build is dying?

src/js/player.js
thijstriemstra: So maybe log an error, fire an error event?

heff:
It looks like it doesn't like `window.throws`
https://travis-ci.org/videojs/video.js/builds/53280031#L466

src/js/player.js
gkatsev: Use [`videojs.log.error()`](https://github.com/videojs/video.js/blob/master/src/js/lib.js#L780-L785)

gkatsev:
`throws` needs to be added here https://github.com/videojs/video.js/blob/master/.jshintrc#L14-L45

gkatsev:
and then you should use `throws` without referencing it from `window`.

thijstriemstra:
Ok, thanks I'll check it out asap. I used the window reference because I saw this being used in text track tests, doesn't seem any other tests that check for a throw.

thijstriemstra:
It doesn't fix my local jshint 'issues' though, but I guess that's a different PR.

thijstriemstra:
Adding `throws` to jshint solved the problem in the jshint build step but not the `minify:tests` target:
```
test/unit/plugins.js:168: ERROR - Parse error. identifier is a reserved word
  throws(function() {
  ^
```
Since I'll just use `videojs.log.error`, and not be checking for a raised exception, I'm not going to bother looking into that error further.

thijstriemstra:
Why are the tests minified anyway? This causes unexpected results where the key name is minified it seems:

```
	Expected: nonExistingPlugin
	Actual: Yf
	    at /Users/foo/Sites/projects/video.js/node_modules/qunitjs/qunit/qunit.js:1540
	    at /Users/foo/Sites/projects/video.js/build/files/test.minified.video.js:344
	    at /Users/foo/Sites/projects/video.js/node_modules/qunitjs/qunit/qunit.js:1303
	    at /Users/foo/Sites/projects/video.js/node_modules/qunitjs/qunit/qunit.js:1463
	    at process (/Users/foo/Sites/projects/video.js/node_modules/qunitjs/qunit/qunit.js:1016)
	    at /Users/foo/Sites/projects/video.js/node_modules/qunitjs/qunit/qunit.js:186
```
I guess making it a string key in the test fixes it.

thijstriemstra:
Updated pull request and now a plugin that does not exist logs an error.

heff:
We don't need `throws` in the globals any more. But otherwise looks good to me.

Issue #1880, opened 2015-02-19
Deal with global css overrides


gkatsev:
There are issues, currently, specifically with the captions settings menu, where if it lives on a page with Foundation (or maybe other css frameworks), it will look all weird.
There are two solutions I can think of, in no particular order:
* Apply a css reset to the `.video-js` class and its descendents.
* Use the iframe trick from things like the facebook like button.

The css-reset thing is a bit of an arms race, though, if done right, it could last a long time.
The iframe trick is pretty interesting. It involves creating a programmatic iframe and putting all your elements in there. Having the iframe means that the css of the outer page doesn't bleed over into our styles but it means that it's more difficult for users to style it.

mmcc:
@heff and I were actually talking about a reset yesterday, so I'm leaning that direction. I don't love increasing specificity by chaining the same class, but that might be the best way to do a reliable reset in this case.

gkatsev:
Actually, I just thought up of another thing: web components and shadow DOM.
By default, CSS doesn't go into shadow DOM but there's now a selector that allows you to pierce shadow dom: `>>>`.

mmcc:
IE8. `/me cries`

heff:
What's the difference between the settings UI elements and the existing player UI elements where we're not running into these issues?

mmcc:
I can't speak to the settings UI elements, but for the new base theme, part of the issue is the fact that we're prepending the styles to the head. If I include the stylesheet after the reveal.js styles, things are fine, but if I let VJS automatically prepend, things get borked.

heff:
That's interesting. We definitely don't want it to matter what order you put those stylesheets in. Does that happen with the old skin too?

gkatsev:
I'm sure it does.

gkatsev:
I really wish that `scoped` stylesheets were actually a thing in more than 0.5% of all browsers.

heff:
I'd rather be sure it's not a regression before we make a decision like adding a CSS reset based on it. And that's still a different issue to me than the captions settings UI, and why they break while our existing UI elements don't.

We've had surprisingly few issues with the existing skin and specificity, so before we throw our hands in the air and add a generalized reset CSS, I think we should understand the problems more specifically.

gkatsev:
Just to be clear, not suggesting a general css reset but rather a specific css reset to make sure all our styles will work as *we* expected them to work regardless of what other css is on the page.

[...]

Issue #1499, opened 2014-09-11
Consider using <button> element


baloneysandwiches:
One of the bragging points of <a href="https://www.paypal-engineering.com/2014/09/05/introducing-an-accessible-html5-video-player/">Paypal’s new html5 player</a> is semantic markup. I think they have a point there. Best practice is to mark up a button as a `<button>`, and we’re currently not doing that. Given that we have proper ARIA roles, this is less of a concern. But `<button>`s get authors lots of accessibility goodness right out of the box, and I think they help to cover some edge cases, especially with screen readers. This would require a bit of an overhaul. Right now the menus are inside the button `<div>`s, and in the default skin they get hidden and shown with descendant selectors:

    .vjs-default-skin .vjs-menu-button:hover .vjs-menu {display: block;}

The idea would be to switch to markup like:

    <div class="subtitles-wrapper>
        <button class=“vis—menu-button”>Subtitles</button>
        <div class=“vjs-menu”></div>
    </div>

and switch to adjacent selectors in the CSS:

    .vjs-default-skin .vjs-menu-button:hover + .vjs-menu {display: block;}

Thoughts?

mmcc:
Done! Kind of, but aside from Menu buttons, vanilla buttons are now buttons.

Issue #841, opened 2013-11-21
Update aria settings with KLG's suggestions as appropriate


heff:
@karlgroves posted about how [div's aren't appropriate for buttons](http://www.karlgroves.com/2013/05/14/links-are-not-buttons-neither-are-divs-and-spans/). I emailed him the Video.js reasoning, and he made the following suggestions for updates to the ARIA settings.

I'd like to get a second set of eyes on these first, hopefully from @gdkraus who did the most recent work on this, and maybe @majornista or @jtangelder.

**Notes:  **
I don't think the comment "The 'display:none' here negates all the ARIA..." actually applies. We only add display:none on those buttons if there are no subtitles/captions/chapters tracks. If there were, those buttons would have display:block and would be showing.

```html
<div class="video span10 offset1">
	<div id="home_video" class="video-js vjs-default-skin vjs-paused vjs-controls-enabled vjs-user-inactive">
		<video class="vjs-tech" poster="/img/poster.jpg" preload="none" id="home_video_html5_api">
			
			<source type="video/mp4" src="http://vjs.zencdn.net/v/oceans.mp4"></source>
			<source type="video/webm" src="http://vjs.zencdn.net/v/oceans.webm"></source>

		</video>
		<div></div>
		
		<div class="vjs-poster" tabindex="-1" style="background-image: url(&quot;/img/poster.jpg&quot;);"></div>
		<div class="vjs-text-track-display"></div>
		
		<!--// KLG: May be good to add some text in here when the spinner is visible, or maybe mark the video as aria-busy='true' until loaded //-->
		<div class="vjs-loading-spinner"></div>
		
		<!--// KLG: use role of button. Toggle aria-pressed states on keydown & keyup. //-->
		<div class="vjs-big-play-button" aria-live="polite" tabindex="0" aria-label="play video">
			<span></span>
		</div>
		<div class="vjs-control-bar">
			<!--// KLG: use role of button. Toggle aria-pressed states on keydown & keyup. Get rid of aria-live //-->
			<div class="vjs-play-control vjs-control " aria-live="polite" tabindex="0">
				<div class="vjs-control-content">
					<span class="vjs-control-text">Play</span>
				</div>
			</div>
			
			<!--// KLG: Give role of timer. See http://www.w3.org/WAI/PF/aria/roles#timer //-->
			<div class="vjs-current-time vjs-time-controls vjs-control">
				<div class="vjs-current-time-display" aria-live="off">
[...]

PR #1932, opened 2015-03-06
Research Spike: Switching buttons to use the button element

heff:
For v5.0 we've been discussing switching to `<button>` instead of `<div>` for all the UI buttons. The general opinion among usability experts is that `<button>` is still better than `<div>` even with all the accessibility enhancements you can do (e.g. ARIA) because certain screen readers still won't support the enhanced divs. A [comment on css-tricks](https://css-tricks.com/use-button-element/#comment-1068889) confirms this for the Nuance Natural Speak app at least, though it would be great to have more specific examples if anyone knows of a resource or has details on specific screen readers to add here.

This PR has the very minimum required to switch to buttons so that they still look right in all browsers. Tested so far in IE8, IE11, Chrome, Firefox, and Safari Desktop.

This does not consider additional specificity that will be need to protect the buttons from being affected by external styles. The next steps will be to add additional property settings and a more specific scope. We at least need enough to be safe within a framework like Foundation that [adds a ton of styles to the base button element](https://github.com/zurb/foundation/blob/af84aa4153d5ee25197ec9f38b735072d80fd316/scss/foundation/components/_buttons.scss#L213).

Related issues: #841, #1499, #1880
Related contributors: @gdkraus, @OwenEdwards, @baloneysandwiches, @karlgroves

### DIFF
--- a/src/css/video-js.less
+++ b/src/css/video-js.less
@@ -64,8 +64,23 @@ replace all occurrences of 'vjs-default-skin' with a new name. Then add your new
 skin name to your video tag instead of the default skin.
 e.g. <video class="video-js my-skin-name">
 */
-.vjs-default-skin {
+.vjs-default-skin,
+.vjs-default-skin button {
   color: @main-font-color;
+}
+
+/* Button reset */
+button.vjs-control {
+  -webkit-appearance: none;
+  -moz-appearance:    none;
+  appearance:         none;
+
+  border: none;
+  background: none;
+  box-sizing: border-box;
+  font-size: 1em;
+
+  overflow: visible; /* IE8 */
 }
 
 /* Custom Icon Font
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -27,6 +27,8 @@ vjs.Button = vjs.Component.extend({
 vjs.Button.prototype.createEl = function(type, props){
   var el;
 
+  type = type || 'button';
+
   // Add standard Aria and Tabindex info
   props = vjs.obj.merge({
     className: this.buildCSSClass(),
@@ -36,6 +38,7 @@ vjs.Button.prototype.createEl = function(type, props){
   }, props);
 
   el = vjs.Component.prototype.createEl.call(this, type, props);
+  el.setAttribute('type', 'button');
 
   // if innerHTML hasn't been overridden (bigPlayButton), add content elements
   if (!props.innerHTML) {


baloneysandwiches:
This is great. From my limited experimentation, JAWS, Voiceover and (from hearsay) other latest generation readers do announce any element as a "button" if it's given that ARIA role. Normalizing markup is nice, but it seems like the big win with switching to `<button` markup has more to do with the spacebar versus the enter key as the presumed event trigger. <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role">MDN:</a>

> "Warning: Be careful when marking up links with the button role. Buttons are expected to be triggered using the Space key, while links are expected to be triggered through the Enter key. In other words, when links are used to behave like buttons, adding role="button" alone is not sufficient. It will also be necessary to add a key event handler that listens for the Space key in order to be consistent with native buttons."

gdkraus:
The biggest problem in the past was with Dragon Naturally Speaking not recognizing role="button" as a clickable element. DNS users can easily pull up a visual list of all of the clickable items so they can easily activate a clickable item anywhere on the page with voice commands. Because it couldn't recognize role="button" as clickable, the user had to repeatedly say "Tab" until the keyboard focus got to the play button in order to activate it. I believe the latest version of DNS fixes this. I can conform next week. I'm away at a conference right now.

I can do some screen reader tests next week too.

heff:
To play devil's advocate a bit then:
- we already support both spacebar and enter on our div buttons
- Dragon has apparently improved
- I don't see Dragon or Nuance in [this survey of readers](http://webaim.org/projects/screenreadersurvey5/#primary) (though they could be in the 5% of 'other')

If we can confirm that the rest of the readers there do support `role="button"` well, is this really a problem we need to fix? All things being equal we would anyway, but:
- We'll be opening the door for a lot more style conflicts from frameworks like Foundation
- We'll be breaking existing v4.0 skins more significantly (we're breaking them in minor ways for 5.0 either way)

baloneysandwiches:
Fair point. If everyone's on latest browsers/readers, with polyfilled behavior, then the difference is increasingly moot. There's still something slightly grating about it though, like having block-level spans and inline divs :crying_cat_face: 

(FWIW the <a href="https://gist.github.com/baloneysandwiches/d0f2b79d8b67da2cb0ea">default button styling in Foundation</a> isn't too crazy or high-specificity.)

dmlap:
lgtm

mmcc:
Looks like we're going to do this, but this PR isn't going to carry the baton across the finish line. Closing in lieu of #1989 and we'll revisit post-ES6.

PR #1939, opened 2015-03-09
Fixing some unremoved event handlers and hidden video tag for html5 tech

giakas:
issue 1: when Tech change to html5, video is hidden.
issue2: MediaController in videojs was not releasing some handlers during dispose.
issue3: Turn off Tracking simulated progress and timeupdates when Error occured

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -149,6 +149,12 @@ vjs.Html5.prototype.createEl = function(){
 
     vjs.insertFirst(el, player.el());
   }
+  else {
+    // need to make sure we add the video tag back in the UI, when we are changing tech to html5.
+    if (el.parentNode !== player.el()) {
+    vjs.insertFirst(el, player.el());
+    }
+  }
 
   // Update specific tag settings, in case they were overridden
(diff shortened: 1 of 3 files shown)


mmcc:
Thanks for the PR! Your tests aren't running at all because of the JSHint errors, but once they do, I think you're going to have more failures due to the `activateControls` changes. If you notice, that's actually defined within the scope of that function and not added to the `MediaTechController` prototype.

giakas:
Thanks Matthew,

Fixed the JSHint errors.
And ran grunt test and got no failures.

And the activateControls, I did create a new prototype method in MediaTechController so that it can be accessible in the media.js.

Do I need to create a new pull request,, or can I update the current pull request with the new commit to fix JShint erros.

Girish.

mmcc:
@giakas sorry for the delay, looks like the tests are passing fine now! This needs an actual review  beyond my initial, quick glance, so we'll ping this thread in the next few days.

ange007:
It is the best PR lately, solved at last problems which were often discussed here.

heff:
This was fixed by #2053 and the new Tech 2.0 work.

Issue #1586, opened 2014-10-15
Remove Closure Compiler aggressive variable mangling


heff:
We're currently using Google's Closure Compiler with [Advanced Optimizations](https://developers.google.com/closure/compiler/docs/compilation_levels) to get better minification of video.js, at least in comparison to uglifyjs using [all the compression options](http://davidwalsh.name/compress-uglify).

- CC w/ AO: **68k**
- UglifyJS: **89k** (21k or 30% bigger)

Part of getting to that level of minification is requiring that we whitelist any API property or function that should be accessible by name after minificaiton, using `exports` or `externs`. That in itself isn't necessarily a bad thing. It means we're deliberate about what we expose, and allows us to change major internal pieces of the player without worrying that we're breaking external code or plugins.

However as more plugins have been built on top of video.js we've had a ton of requests for exporting additional properties, many of which we just overlooked. There's also the issue of sub-classing video.js components, like `videojs.Button`. To allow sub-classing we can't simply `export` methods, which makes a copy of them. We have to `extern` them, which prevents them from ever being mangled. Moving to externs for all components would create a lot of extra work and would also reduce some of the closure compiler benefit. (I'm not exactly sure how much, but before we started more aggressively exporting/externing properties the uglify version was over 50% bigger.)

If we were only exposing the player API this might be a different story, but video.js is also part UI library and part utility library, and those other pieces can be just as valuable to plugin developers. Exposing those other pieces means there's going to be more things that are exported than are not.

Enough background... I'm hoping to get feedback on a few things.

1. **How significant is 21k?** It's 21k in a player that's delivered over a billion times a month including to regions with terrible bandwidth, so it's not nothing, but are there other ways we could/should be quantifying that?

2. **Are there other things we can do to improve minification in addition to or instead of UglifyJS?** Is there a way we could create a blacklist of object properties that get mangled, e.g. private vars?

3. **What are other things we should be considering?** We'll be exposing a lot of components and utilities for the first time, and once we do that it becomes a lot harder to change them in the future. (#1232 is related)

As a half answer to #2, I did some manual minification tests to see what different approaches could do for us if possible.
- Mangling private vars - savings of about 3k
- Reduce the verboseness of class method definitions, e.g. `component.prototype.method =` - savings of about 5k.

danrossi:
Confirming this is an issue, your entire public api for buttons and components is completely mangled

vjs.zencdn.net/4.11.4/video.js

[...]

PR #1940, opened 2015-03-09
Remove Closure Compiler

heff:
Replace Closure Compiler with UglifyJS and remove closure compiler specific files. Addresses #1586.

We could also remove the API step in the automated tests now that we're not mangling object properties, unless we think that could still be valuable. @videojs/core-committers?


tasks/build.js
```diff
@@ -16,7 +17,12 @@ module.exports = function(grunt) {
     combined = combined.replace(/GENERATED_CDN_VSN/g, grunt.vjsVersion.majorMinor);
```
gkatsev: when we get browserify, there's a transform that does this automatically for us.

heff:
[In the comments](https://github.com/videojs/video.js/issues/1586#issuecomment-61026460) of the related issue we discussed changing the private vars to use the underscore-first naming convention. I can wrap that into this one, but I wanted to give people a chance to respond in case there's significant reliance on the current private vars that we should be aware of.

gkatsev:
So much red!

Gruntfile.js
```diff
@@ -293,14 +324,25 @@ module.exports = function(grunt) {
       }
     },
     usebanner: {
```
gkatsev: is there a need for usebanner anymore? I believe uglify can do it for us.

gkatsev:
This keeps the current file naming scheme, right?

Gruntfile.js
heff: I saw that but wanted the banner in the non-minified version too, including version and copyright. Sounds like we could use browserify to do var replacement here too instead.

tasks/build.js
heff: Could we still support dropping the patch number like we're doing here?

heff:
It does keep the current file naming. That's a separate issue but I could easily wrap it into this one. That's one of those things I could see getting annoying when switching between 4.0 and 5.0 work, so we could also wait until closer to release. I could go either way.

Gruntfile.js
gkatsev: We can run the unminified code through uglify beautifier if we wanted to. But makes sense to keep banner.
Does uglify create sourcemaps? If so, this could mess up with that if it is run after uglify.

heff: It's run before uglify, and uses the `@license` tag which uglify will preserve.

gkatsev: Ah, cool.

heff:
I added a wiki page for tracking the details (what changed and why) of any API or process changes in 5.0. If we keep this up as we go it will be easier to put together communication around the 5.0 release.

mmcc:
lgtm

gkatsev:
:tada:

Grabilla:
http://lisperator.net/blog/should-you-switch-to-uglifyjs2/

Issue #1864, opened 2015-02-13
Captions: Weird setTimeout in Chapters' menu creation


gkatsev:
Investigate and figure out if we can have a better way of getting all the chapters and displaying it as soon as the chapters' cues are loaded in.

chemoish:
+1

gkatsev:
This has been fixed by @chemoish. Thanks @chemoish!

PR #1949, opened 2015-03-13
Add a 'loadeddata' event on TextTracks, which fixes issue #1864.

OwenEdwards:
Propose this useful feature for Text Track handling in general, which also fixes issue #1864.

Note that this needs broader testing, and a specific unit test for the fix; @gkatsev, you seem to be the expert on this? :-)

### DIFF
--- a/src/js/tracks/text-track-controls.js
+++ b/src/js/tracks/text-track-controls.js
@@ -488,22 +488,21 @@ vjs.ChaptersButton.prototype.createMenu = function(){
       track, chaptersTrack,
       items = this.items = [];
 
+  //TODO: Handle more than one chapters track
   for (; i < l; i++) {
     track = tracks[i];
     if (track['kind'] == this.kind_) {
-      if (!track.cues) {
-        track['mode'] = 'hidden';
-        /* jshint loopfunc:true */
-        // TODO see if we can figure out a better way of doing this https://github.com/videojs/video.js/issues/1864
-        window.setTimeout(vjs.bind(this, function() {
-          this.createMenu();
-        }), 100);
-        /* jshint loopfunc:false */
-      } else {
-        chaptersTrack = track;
-        break;
-      }
+      chaptersTrack = track;
+      break;
     }
+  }
+
+  if (chaptersTrack && !chaptersTrack.cues) {
+    chaptersTrack['mode'] = 'hidden';
+
+    chaptersTrack.addEventListener('loadeddata', vjs.bind(this, function() {
+      this.update();
+    }));
   }
 
   var menu = this.menu;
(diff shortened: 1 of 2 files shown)


gkatsev:
Awesome, thanks. I'll take a look over the weekend.

gkatsev:
I'm not sure I like that we're adding a `loadeddata` event, given that these are supposed to be very much in-line with the spec. Maybe we need to add an actual shim for a <track> element since that's what we can listen to for load events in native tracks.
And it would mean we could return an actual "track" object from `addRemoteTextTrack` that you could listen to events on.

OwenEdwards:
Oh, right, now I'm looking more closely at the WhatWG spec, in section [sourcing-out-of-band-text-tracks](https://html.spec.whatwg.org/multipage/embedded-content.html#sourcing-out-of-band-text-tracks), it mentions:

> If the fetching algorithm does not fail, and the file was successfully processed, then the final task that is queued by the networking task source, after it has finished parsing the data, must change the text track readiness state to loaded, and fire a simple event named load at the track element.

I don't see the TextTrack element has a [text-track-readiness-state](https://html.spec.whatwg.org/multipage/embedded-content.html#the-track-element:text-track-readiness-state), but you're right, really we should return an actual "track" object and fire a "load" event on that. I was working from the [HTML5 spec](http://www.w3.org/TR/html5/embedded-content-0.html#the-track-element), and this seemed reasonable!

gkatsev:
The Tracks that videojs currently has are actual [TextTracks](https://html.spec.whatwg.org/multipage/embedded-content.html#texttrack) as opposed to the [HTMLTrackElement](https://html.spec.whatwg.org/multipage/embedded-content.html#the-track-element:text-track-readiness-state) that you linked. I thought that only the former would be sufficient, but seems like even though it gives us 95% of what we need, the other 5% are pretty important too.

gkatsev:
Once we have the HTMLTrackElement shim thing, we probably would still need this event but it just shouldn't be officially exported (like `modechange`).

OwenEdwards:
I see that the VideoJS TextTracks are not an HTML TrackElement. It's the process of calling addRemoteTrack, calling createTrackHelper, and taking src as an option (which I don't see documented?) which blurs the line between the two. There's the XHR call, and we need to know when that's completed (I'm looking to use it elsewhere). Otherwise, I'd XHR it myself, but then I could only pass it to TextTracks by parsing it myself and feeding it in as Cues, right?

chemoish:
@OwenEdwards Did you ever encounter race conditions with chapters because of this timeout?

gkatsev:
I think that #2804 incorporates and supersedes this one. Closing.

PR #1440, opened 2014-08-21
Adding "fluid" mode

baloneysandwiches:
This is a first pass at the ideas discussed in #982.

Issue #983, opened 2014-02-04
Improve default width/height settings


heff:
We have a challenge when trying to mimic the video element with the player div. The video element defaults to 300x150 if no width/height attributes or CSS is applied, but a div defaults to 100% width and zero height, and doesn't respond to width/height attributes. In order to make the div act like the video element and not require width and height, we have to apply a default using CSS.

Currently we're adding the default using the element styles if no width/height is supplied (and no `auto` is used either as in #359). The problem is still that these styles can't be overridden easily, and the `auto` setting isn't valid for width/height.

Talking with @mmcc we may have come up with a better solution: Add the 300x150 default to the video-js class. This would allow it to be overridden more easily, either by custom user styles that follow video-js.css in the dom, or through element styles like it currently does when you do supply a width/height.

So the solution would be to kill the default here:
https://github.com/videojs/video.js/blob/v4.3.0/src/js/core.js#L86

And add it here:
https://github.com/videojs/video.js/blob/v4.3.0/src/css/video-js.less#L672

This would need to be tested in different situations.
- No width/height provided at all
- No width/height provided but set by CSS

gkatsev:
I believe this has been taken care off with the 5.0 update and the defaults and styles classes.

Issue #982, opened 2014-02-04
Responsive Sizing


heff:
Responsive sizing is when the player can dynamically change size to fit its container while maintaining the desired aspect ratio. This should be a feature built into video.js and we need to find out the right way to add it. (it's currently possible with additional work or outside libs)

### To finish
- [ ] Tests written  
- [ ] Feature implemented  
- [ ] Docs/guides updated  
- [ ] Example created (e.g. [jsbin](http://jsbin.com/axedog/9999/edit))  
- [ ] Pull request submitted  

Related issues: #771, #359, #384, #925, #983
Also see the [intrinsic ratio example](http://jsfiddle.net/alexbell/Zs5WB/) from @baloneysandwiches

Related Links:
http://alistapart.com/article/creating-intrinsic-ratios-for-video/

The intrinsic ratio method seems like the best approach. In order to make that happen, we need to allow the user to supply the ratio some how. Some potential options:

```
// set a flag and then we calculate the ratio from the width and height
videojs('id', { responsive: true, width: 1600, height: 900 }); // --> ratio = 16:9

// provide a ratio
videojs('id', { aspectRatio: '16:9' });
```
(specific naming TBD)

Video.js would essentially have 2 layout modes, fixed and responsive, and we would probably apply a different class of styles to the player div depending on which was needed.

### Additional considerations
- How this might affect poster sizing
- What happens if width('600') is called on the player when in responsive mode?
- Determining the aspect ratio after the metadata loads from videoWidth/videoHeight. (We don't support this yet with the fixed layout, so should probably be a separate issue.)

### User Stories Draft

Not sure if all of these are valid, but providing possible solutions for each based on the aspectRatio/autoHeight setting.
```
User wants the video to fill width, and specifies an aspect ratio to calculate height
  `videojs({ aspectRatio: '16:9' });`
    -> Add width: 100%, padding-bottom: (9/16)%

User wants the video to fill width, and auto adjust height to match source aspect ratio from metadata
  `videojs({ aspectRatio: 'auto' });`
    Before metadata:
      Use a default aspect ratio, like 2:1 or 16:9
      -> Add width: 100%, padding-bottom: (9/16)%
    After loadedmetadata
      -> update padding to match source aspect ratio

User wants the video to fill height, and specifies an aspect ratio to calculate width?
  I don't think this is a valid user story, or at least I can't think of a real world use case. Video dimensions are rarely taller than wide (though...iPhone videos). Browser windows have restricted widths not heights.

[...]

PR #1952, opened 2015-03-17
Fluid widths and aspect ratios with metadata defaults

heff:
Building on #1440 from @baloneysandwiches. (Would appreciate your feedback Alex.) This isn't completely finished and needs tests but wanted to get it out for review.

The big diff here is that I'm creating a style element and using that to add widths (user-set, metadata-provided, and defaults) and aspect ratios for fluid layouts. The goal was to let the video.js user override player dimensions with CSS the same way they can with the video element. Previously we were using inline styles to add width and height, and we were doing that with the default values, which meant you couldn't easily use CSS to size the player. (I'm able to make this change now because we're in 5.0 development)

I've included a screen shot of all the different user stories from #982 passing.

One question that's come out of this is whether or not we need the width/height functions on other components, since everything is getting sized in the main stylesheet. It would be nice to drop or at least simplify those. In this PR I'm overriding the player's width/height functions so now they work differently from other components.

Amazingly this works in IE8 with Flash, including the metadata widths. I'm just having an issue getting the poster image centered again.

responsive-examples

Related Issues: #983, #1440, #982

### DIFF
--- a/src/css/components/_layout.scss
+++ b/src/css/components/_layout.scss
@@ -1,6 +1,14 @@
 .video-js {
-  display: block;
+  /* inline-block is as close as we get to the video el's display:inline */
+  display: inline-block;
+  /* Make video.js videos align top when next to video elements */
+  vertical-align: top;
   box-sizing: border-box;
+
+  /* Default to the video element width/height. This will be overridden by
+   * the source width height unless changed elsewhere. */
+  width: 300px;
+  height: 150px;
 
   color: $primary-text;
   background-color: $primary-bg;
@@ -9,8 +17,6 @@
   /* Start with 10px for base font size so other dimensions can be em based and
   easily calculable. */
   font-size: $base-font-size;
-  /* Allow poster to be vertially aligned. */
-  vertical-align: middle;
 
   /* Provide some basic defaults for fonts */
   font-weight: normal;
@@ -35,6 +41,28 @@
 .video-js *:before,
 .video-js *:after {
   box-sizing: inherit;
+}
+
+/* Fill the width of the containing element and use padding to create the
+   desired aspect ratio. Default to 16x9 unless another ratio is given. */
+.video-js.vjs-fluid,
+.video-js.vjs-16-9 {
+  width: 100%;
+  max-width:100%;
+  height: 0;
+  padding-top: 56.25%;
+}
+
+.video-js.vjs-4-3 {
+  width: 100%;
+  max-width:100%;
+  height: 0;
+  padding-top: 75%;
+}
+
+.video-js.vjs-fill {
+  width: 100%;
+  height: 100%;
 }
 
 /* Playback technology elements expand to the width/height of the containing div
@@ -65,6 +93,8 @@ body.vjs-full-window {
   right: 0;
   width: 100% !important;
   height: 100% !important;
+  /* Undo any aspect ratio padding for fluid layouts */
+  padding-top: 0 !important;
 }
 .video-js.vjs-fullscreen.vjs-user-inactive {
   cursor: none;
--- a/src/css/components/_layout.scss
+++ b/src/css/components/_layout.scss
@@ -45,19 +45,20 @@
 
 /* Fill the width of the containing element and use padding to create the
    desired aspect ratio. Default to 16x9 unless another ratio is given. */
+@mixin apply-aspect-ratio($width, $height) {
+  width: 100%;
+  max-width: 100%;
+  height: 0;
+  padding-top: 100% * ($height/$width);
+}
+
 .video-js.vjs-fluid,
 .video-js.vjs-16-9 {
-  width: 100%;
-  max-width:100%;
-  height: 0;
-  padding-top: 56.25%;
+  @include apply-aspect-ratio(16, 9);
 }
 
 .video-js.vjs-4-3 {
-  width: 100%;
-  max-width:100%;
-  height: 0;
-  padding-top: 75%;
+  @include apply-aspect-ratio(4, 3);
 }
 
 .video-js.vjs-fill {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -268,7 +268,7 @@ class Player extends Component {
       // If an empty string is given, reset the dimension to be automatic
       this[property] = undefined;
     } else {
-      this[property] = parseInt(value, 10);
+      this[property] = parseFloat(value);
 
       if (isNaN(this[property])) {
         Lib.log.error(`Improper value "${value}" supplied for for ${dimension}`);
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -249,20 +249,32 @@ class Player extends Component {
     return el;
   }
 
-  width(width) {
-    if (width === undefined) {
-      return this.width_ || 0;
-    }
-    this.width_ = parseInt(width, 10);
-    this.updateStyleEl_();
-    return this;
+  width(value) {
+    return this.dimension('width', value);
   }
 
-  height(height) {
-    if (height === undefined) {
-      return this.height_ || 0;
+  height(value) {
+    return this.dimension('height', value);
+  }
+
+  dimension(dimension, value) {
+    let property = dimension + '_';
+
+    if (value === undefined) {
+      return this[property] || 0;
     }
-    this.height_ = parseInt(height, 10);
+
+    if (value === '') {
+      // If an empty string is given, reset the dimension to be automatic
+      this[property] = undefined;
+    } else {
+      this[property] = parseInt(value, 10);
+
+      if (isNaN(this[property])) {
+        Lib.log.error(`Improper value "${value}" supplied for for ${dimension}`);
+      }
+    }
+
     this.updateStyleEl_();
     return this;
   }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -222,10 +222,17 @@ class Player extends Component {
     // Default state of video is paused
     this.addClass('vjs-paused');
 
-    // Make box use width/height of tag, or rely on default implementation
-    // Enforce with CSS since width/height attrs don't work on divs
-    this.width(this.options_['width'], true); // (true) Skip resize listener on load
-    this.height(this.options_['height'], true);
+    // Add a style element in the player that we'll use to set the width/height
+    // of the player in a way that's still overrideable by CSS, just like the
+    // video element
+    this.styleEl_ = document.createElement('style');
+    el.appendChild(this.styleEl_);
+
+    // Pass in the width/height/aspectRatio options which will update the style el
+    this.width(this.options_['width']);
+    this.height(this.options_['height']);
+    this.fluid(this.options_['fluid']);
+    this.aspectRatio(this.options_['aspectRatio']);
 
     // Lib.insertFirst seems to cause the networkState to flicker from 3 to 2, so
     // keep track of the original for later so we can know if the source originally failed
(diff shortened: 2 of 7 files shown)


gkatsev:
Would be nice if the comments are made directly in this PR rather than on the commits themselves because it's easier to see the context and replies of everything.

baloneysandwiches:
Yeah, the integers-only requirement for display width seems like a spec bug to me. It's one of those things that look cleaner, but are actually inaccurate. Maybe because the spec was written before onslaught of high DPPX devices? Anyway, videos render at fractional values all the time. There really should be ways to get/set that, to and from the CSSOM.

A simple layout example: browsers currently render `border: 0.5px solid #000` as expected on retina+ devices. We're already there. The change happened in FF first, then webkit and IE caught up around three years ago:

http://trac.webkit.org/changeset/116009
http://blogs.msdn.com/b/ie/archive/2012/02/17/sub-pixel-rendering-and-the-css-object-model.aspx

You're right that this would have to be an across the board change in order to work. But in terms of page complexity, I actually think the pure-CSS use cases are simpler, and the JS dimension calculations are for more complex behaviors. In terms of a use case: I'm imagining something like a grid of X-by-Y videos, sized in percentages, with expected subpixel rendering on a 2x tablet. Once elements are getting substituted/replaced, JS animations are flying in, etc, you will want to be able to get and set the size of a given element with JS with subpixel precision. It can be hairpulling to troubleshoot rounding errors in layout, and it's pretty forbidding to rewrite all of a plugin's getters and setters to get a little border edge or unwanted overflow to go away.

If Chrome-style 13 digit CSS values offend your sensibilities (which I can totally understand), perhaps consider truncating to a small number of decimal places?

frapontillo:
I am really interested in having a responsive video player, just like the native one is. Are there any updates on this? What's blocking?

heff:
This PR is waiting on the 5.0 release to be fully available. Currently it's waiting on work for changing to ES6 modules. @mmcc we should probably just open a PR for the ES6 change now, so it's there to point to. I also need to write tests for this.

@baloneysandwiches, re CSS widths: I think the right way to do what you're describing is by setting styles on the player element.
```
myPlayer.el().style.width = '100.555px'
// or
document.querySelector('#video-js').style.width = '100.555px';
```
Using the style property is the only way to set fractional dimensions on the video element via javascript, you can't use videoEl.width, so by the "mimic the video element" guideline that's how you should set fractional dimensions for the video.js player too. At the same time I can't think of why anyone would complain that we're *not* rounding off their dimension values, so I don't see an issue with "improving" on the spec in this case to allow floats in the width/height properties/attributes (anyone disagree?). Are there any edge cases you can think of that we should be protecting against while accepting/setting float widths? My plan at this point would just be to switch parseInt with parseFloat.

heff:
Rebased this on master with the all the ES6 changes. All the sizing tests still work. I still need to write tests.

heff:
@videojs/core-committers can I get a review on this one? I'd like to get it merged in.

src/css/components/_layout.scss
```diff
@@ -37,6 +43,28 @@
...
 }
 
+/* Fill the width of the containing element and use padding to create the
+   desired aspect ratio. Default to 16x9 unless another ratio is given. */
+.video-js.vjs-fluid,
+.video-js.vjs-16-9 {
+  width: 100%;
+  max-width:100%;
+  height: 0;
+  padding-top: 56.25%;
```
mmcc: I don't think it's necessary, but it might actually make this more readable to refactor this into a mixin.

```scss
@mixin generate-player-sizing($ratio) {
  width: 100%;
  max-width: 100%;
  height: 0;
  padding-top: 100% * $ratio;
}

.video-js.vjs-fluid,
.video-js.vjs-16-9 {
  @include generate-player-sizing(9/16);
}

.video-js.vjs-4-3 {
  @include generate-player-sizing(3/4);
}
```

That mixin name sucks, but you get the picture.

heff: Cool, I like that.

src/js/player.js
```diff
@@ -242,6 +249,126 @@ class Player extends Component {
...
+  width(value) {
+    return this.dimension('width', value);
+  }
+
+  height(value) {
+    return this.dimension('height', value);
+  }
+
+  dimension(dimension, value) {
+    let property = dimension + '_';
```
mmcc: Can we call this something like `privateProperty`?

src/js/player.js
```diff
@@ -242,6 +249,126 @@ class Player extends Component {
...
+
+    if (value === undefined) {
+      return this[property] || 0;
+    }
+
+    if (value === '') {
+      // If an empty string is given, reset the dimension to be automatic
+      this[property] = undefined;
+    } else {
+      this[property] = parseFloat(value);
```
mmcc: We shouldn't set property to `NaN` (which would happen if the `isNaN` check is true below).

src/js/player.js
```diff
@@ -323,6 +450,7 @@ class Player extends Component {
     this.on(this.tech, 'ratechange', this.handleTechRateChange);
     this.on(this.tech, 'volumechange', this.handleTechVolumeChange);
     this.on(this.tech, 'texttrackchange', this.onTextTrackChange);
+	this.on(this.tech, 'loadedmetadata', this.updateStyleEl_);
```
mmcc: This is showing up weird in the diff for me. Is this a tab instead of spaces?

test/unit/player.js
```diff
@@ -152,31 +153,82 @@ test('should asynchronously fire error events during source selection', function
...
-  // Player container needs to have height in order to have height
-  // Don't want to mess with the fixture itself
-  container.appendChild(player.el());
-  container.style.height = '1000px';
-  ok(player.height() === 1000);
+  let rules;
 
-  player.dispose();
-});
+  function getStyleRules(){
```
mmcc: We've talked about how we have quite a few tests that should probably do more around checking styles. This seems useful and generic...any reason we wouldn't just toss this into the helper or something?

mmcc:
:+1: :tada: :tanabata_tree:

src/css/components/_layout.scss
```diff
@@ -1,7 +1,15 @@
...
+  /* inline-block is as close as we get to the video el's display:inline */
+  display: inline-block;
+  /* Make video.js videos align top when next to video elements */
+  vertical-align: top;
   box-sizing: border-box;
 
+  /* Default to the video element width/height. This will be overridden by
+   * the source width height unless changed elsewhere. */
+  width: 300px;
+  height: 150px;
```
dmlap: Maybe not for this PR, but a thought: should we default this to 16x9?

src/js/player.js
```diff
@@ -222,10 +222,17 @@ class Player extends Component {
...
 
-    // Make box use width/height of tag, or rely on default implementation
-    // Enforce with CSS since width/height attrs don't work on divs
-    this.width(this.options_['width'], true); // (true) Skip resize listener on load
-    this.height(this.options_['height'], true);
+    // Add a style element in the player that we'll use to set the width/height
+    // of the player in a way that's still overrideable by CSS, just like the
+    // video element
+    this.styleEl_ = document.createElement('style');
+    el.appendChild(this.styleEl_);
```
dmlap: This will come after the `head`, where most people will be setting their styles. So this sort of rule:
```css
.video-js {
  width: 100%;
}
```
would not override any rules in the base styles.

test/unit/player.js
```diff
@@ -152,31 +153,82 @@ test('should asynchronously fire error events during source selection', function
   Lib.log.error.restore();
 });
 
-test('should set the width and height of the player', function(){
-  var player = TestHelpers.makePlayer({ width: 123, height: '100%' });
+// test('should set the width and height of the player', function(){
```
dmlap: Remove or still needs fixing?

src/js/player.js
```diff
@@ -242,6 +249,126 @@ class Player extends Component {
...
+    if (this.width_ !== undefined) {
+      // Use any width that's been specifically set
+      width = this.width_;
+    } else if (this.height_ !== undefined) {
+      // Or calulate the width from the aspect ratio if a height has been set
+      width = this.height_ / ratioMultiplier;
+    } else {
+      // Or use the video's metadata, or use the video el's default of 300
+      width = this.videoWidth() || 300;
+    }
```
dmlap: What if the width of the player was set through CSS elsewhere in the document?

src/css/components/_layout.scss
heff: Yeah, this is a hard one for me. We've drawn a hard line around the html5 video spec and there's a lot of benefits in being strict about that. The 300x150 default appears to come from somewhere outside of the html5 video spec, though. The canvas element defaults to 300x150 too. We'd still be going against what browsers have implemented, so a vjs player next to a bare video element wouldn't line up. But I don't think there's any argument against 16:9 being better for video than 2:1. Plus I've already set the default aspect ratio in this to 16:9 when in fluid-mode.

Unless any of that changes your mind I'll make 16x9 the default.

src/js/player.js
heff: I'm not totally following. Are you saying adding styles like this makes it harder for people to override these styles with their own, or that adding styles like this won't override the base video.js styles?

test/unit/player.js
heff: Should be removed. Thanks.

src/js/player.js
heff: If you don't add a width/height to the player, these styles don't get added, so you only have to override the default dimensions in `.video-js`. If  you do set a width/height, this sets it using a new single classname.
```
.vjs_video_292-dimensions { width: 300px; height: 168.75px; }
```
If a user's goal is to set the dimensions of the player in CSS somewhere else, they just have to do *one* of the following:
- don't set a width/height or aspect ratio on the player
- use a stronger selector than a single classname, e.g. ID

I tried to set the lowest bar I could, and this gets pretty close to how the video element works. Trying to figure out what users have set elsewhere would get really complicated and I don't think would add a ton of value.

src/js/player.js
heff: @dmlap any other thoughts on this one or other comments on this PR? If you're good with this I'm gonna merge it in.

test/unit/player.js
heff: It would be cool if we could use this CSS AST other places, but right now this is the only place we add styles by creating a new style element. Other places we could only check if a class was added appropriately.

src/js/player.js
dmlap: Sorry I'm responding post-merge... I was saying this makes it harder for people to override the base styles with their own. I could be wrong but I think rules like the example above aren't uncommon.

src/css/components/_layout.scss
dmlap: I've been shifting towards 16:9 as a default for awhile now. I think I'm convinced it's a better default setting at this point.

src/js/player.js
heff: > I could be wrong but I think rules like the example above aren't uncommon.

I think you're right, but in 4.x we set the width/height with inline styles, so the example definitely wouldn't work with video.js today.

To ensure the example would work we'd have to inject the style el at the top of the head, and remove the default styles from the main stylesheet. Removing the default styles means the embed code wouldn't have any dimensions (at least in IE8) until the player loads. Maybe that's not a real issue?

gkatsev: Probably best to avoid major page reflows which would happen if the player/video don't have any dimensions until loaded.

PR #1963, opened 2015-03-20
Update vjs.Player.md

LagunaCompany:
Fix minor typo.


mmcc:
We really need to do a better job of documenting where these docs come from. These are all generated from comments in the code, so [this](https://github.com/videojs/video.js/blob/2c14c70ee80e7baa9dcd494d965423d6bccc93c7/src/js/component.js#L339) is where that change should be made. Almost all docs reference component, so you'll find this typo in just about every doc page :)

LagunaCompany:
Ah, I see! Thanks for the info :)

LagunaCompany:
New pull request to make this change in component.js. Issue #1964.

PR #1974, opened 2015-03-24
Extended canHandleSource in media files to consistently guess mime types. Fixes #1833

carpasse:
Added getFileExtension to libs. And used it on Html5 and Flash canHandleSource method to guess the mime type if necessary.

### DIFF
--- a/src/js/lib.js
+++ b/src/js/lib.js
@@ -849,3 +849,25 @@ vjs.arr.forEach = function(array, callback, thisArg) {
 
   return array;
 };
+
+/**
+ * Returns the extension of the passed file name. It will return an empty string if you pass an invalid path
+ *
+ * @param {String}    path    The fileName path like '/path/to/file.mp4'
+ * @returns {String}          The extension in lower case or an empty string if no extension could be found.
+ */
+vjs.getFileExtension = function(path) {
+  var splitPathRe;
+  var pathParts;
+
+  if(typeof path === 'string'){
+    splitPathRe = /^(\/?)([\s\S]*?)((?:\.{1,2}|[^\/]+?)(\.([^\.\/\?]+)))(?:[\/]*|[\?].*)$/i;
+    pathParts = splitPathRe.exec(path);
+
+    if (pathParts) {
+      return pathParts.pop().toLowerCase();
+    }
+  }
+
+  return '';
+};
--- a/src/js/media/flash.js
+++ b/src/js/media/flash.js
@@ -233,15 +233,24 @@ vjs.Flash.nativeSourceHandler = {};
  * @param  {Object} source  The source object
  * @return {String}         'probably', 'maybe', or '' (empty string)
  */
-vjs.Flash.nativeSourceHandler.canHandleSource = function(source){
+
+vjs.Flash.nativeSourceHandler.canHandleSource = function (source) {
   var type;
 
-  if (!source.type) {
+  function guessMimeType(src) {
+    var ext = vjs.getFileExtension(src);
+    if (ext) {
+      return 'video/' + ext;
+    }
     return '';
   }
 
-  // Strip code information from the type because we don't get that specific
-  type = source.type.replace(/;.*/,'').toLowerCase();
+  if (!source.type) {
+    type = guessMimeType(source.src);
+  } else {
+    // Strip code information from the type because we don't get that specific
+    type = source.type.replace(/;.*/, '').toLowerCase();
+  }
 
   if (type in vjs.Flash.formats) {
     return 'maybe';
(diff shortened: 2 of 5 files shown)


heff:
Awesome, thanks!

src/js/lib.js
```diff
@@ -849,3 +849,25 @@ vjs.arr.forEach = function(array, callback, thisArg) {
...
+ *
+ * @param {String}    path    The fileName path like '/path/to/file.mp4'
+ * @returns {String}          The extension in lower case or an empty string if no extension could be found.
+ */
+vjs.getFileExtension = function(path) {
+  var splitPathRe;
+  var pathParts;
+
+  if(typeof path === 'string'){
+    splitPathRe = /^(\/?)([\s\S]*?)((?:\.{1,2}|[^\/]+?)(\.([^\.\/\?]+)))(?:[\/]*|[\?].*)$/i;
```
heff: That's a pretty advanced regex. What does it add over the one that was in the html5 tech?

src/js/media/flash.js
```diff
@@ -233,15 +233,24 @@ vjs.Flash.nativeSourceHandler = {};
  * @param  {Object} source  The source object
  * @return {String}         'probably', 'maybe', or '' (empty string)
  */
-vjs.Flash.nativeSourceHandler.canHandleSource = function(source){
+
+vjs.Flash.nativeSourceHandler.canHandleSource = function (source) {
   var type;
 
-  if (!source.type) {
+  function guessMimeType(src) {
```
heff: Is it worth adding a function here if it's used once? It does scope the var so if you think it's just cleaner that's fine with me.

test/unit/lib.js
```diff
@@ -403,3 +403,22 @@ test('should loop through each element of an array', function() {
     }
   });
 });
+
+//getFileExtension tests
+test('should get the file extension of the passed path', function() {
```
heff: Can we add a test for a full URL? e.g. `http://...` I can see that it supports it but it'd be good to have a test for it.

heff:
Added  a few comments, but otherwise this looks good to me. :+1:

src/js/lib.js
gkatsev: It's the regex from [iojs](https://github.com/iojs/io.js/blob/4e9bf93e9c645f33446f7d47005868ffcb3d281e/lib/path.js#L394), which we talked about in the [other pr comments](https://github.com/videojs/video.js/pull/1833)

heff: Ah, right. Thanks.

src/js/media/flash.js
carpasse: I believe it clarifies the intention. I can remove it if you don't like it.

test/unit/lib.js
carpasse: I have added some tests cases with 'http://' is that enough??

src/js/media/flash.js
heff: Cool, I can see that. Works for me.

test/unit/lib.js
heff: I think that's good. If anyone else has ideas on URL formats we should be testing for speak up.

heff:
I think this is ready to go. Just need one more person to review and sign off on it. @videojs/core-committers

xy2015:
Sent from Windows Mail





From: Steve Heffernan
Sent: ‎Wednesday‎, ‎March‎ ‎25‎, ‎2015 ‎7‎:‎25‎ ‎AM
To: videojs/video.js





Added a few comments, but otherwise this looks good to me.  :+1:

carpasse:
@xy2015 I can not see you comments can you please add them again or point me to them??

test/unit/lib.js
dmlap: A protocol-relative URL (e.g. "//example.com/movie.mp4") would be nice but I don't think it needs to hold up merging this.

dmlap:
LGTM

dmlap:
@heff I hit the "merge" button by habit, sorry. I'll fix up the changelog tomorrow morning

heff:
No problem. Though @mmcc might not have fun rebasing the ES6 work again. :smile:

mmcc:
(╯°□°）╯︵ ┻━┻ 




—

dmlap:
@mmcc sorry!

mmcc:
Ok this is now pulled into ES6 branch, but I'm going to hold off on pushes until all the comments are addressed. I've rebased from master to clean up the commit history so I think we'll lose all the comments when I end up force pushing.

Issue #1907, opened 2015-02-28
docs say playback rate is a boolean?


parke:
From the docs for playbackRate:

"PARAMETERS: rate Boolean New playback rate to set."

source: https://github.com/videojs/video.js/blob/stable/docs/api/vjs.Player.md#playbackrate-rate-

I believe rate should be a Number, not a Boolean.

gkatsev:
Yep, the docs are wrong :)

mmcc:
Wasn't there a recent PR to fix this?



—

gkatsev:
Yes: https://github.com/videojs/video.js/pull/1876

heff:
The PR was against master, so unfortunately it wouldn't go into the stable docs until the next minor release, which wouldn't be until 5.0 now. Not quite sure the best way to handle this.

gkatsev:
Don't most people look at the docs in master?
If it's a huge concern, we probably should just cherry-pick the commit/PR.

parke:
I do not know where most people look.

This page:  http://www.videojs.com/

Links to here:  https://github.com/videojs/video.js/blob/stable/docs/index.md

Which eventually leads to here:  https://github.com/videojs/video.js/blob/stable/docs/api/vjs.Player.md

gkatsev:
In that case we should cherry-pick the commit/PR. Git will figure it out.

PR #1975, opened 2015-03-25
Fix playbackRate option parameter

TylerWilcock:
I'm assuming the parameter for playbackRate should be a number, not a boolean.


gkatsev:
It's already been fixed in the [code](https://github.com/videojs/video.js/pull/1876) itself. We need to re-generate the docs. Follow #1907 for updates.

Issue #1720, opened 2014-12-05
Too many places list required files


gkatsev:
* [`source-loader`](https://github.com/videojs/video.js/blob/master/build/source-loader.js) - this is mostly the canonical place
* [`gruntfile`](https://github.com/videojs/video.js/blob/master/Gruntfile.js) - this is mostly via very permissive globs
* [`test/index.html`](https://github.com/videojs/video.js/blob/master/test/index.html) - just the test files
* [`karma-conf`](https://github.com/videojs/video.js/blob/master/test/karma.conf.js) - source files plus a glob for unit tests

Issue #867, opened 2013-12-03
Remove JSON shim to avoid eval() CSP issues


paz:
(updated title and description to describe the enhancement)

IE8 supports JSON in IE8 standards mode (the default), but not other browser modes. In 5.0 we should be able to ask video.js users who need to support JSON in other modes or older browsers to supply their own JSON shim. This way we can avoid Content Security Policy issues that come up, even when the shim isn't used. 

This shouldn't be removed in 4.0 because of the potential to break existing installs, including those that are still making Video.js work with IE6 & 7.

**Original Title + Description**

### eval() in vjs.JSON.parse forces use of 'unsafe-eval' in Content Security Policy in order for VideoJS to work

eval() in vjs.JSON.parse forces use of 'unsafe-eval' in Content Security Policy in order for VideoJS to work. Is there any chance the JSON parser will move to a non-eval()ing parser? 

In vjs.JSON.parse on line 7024 in video.dev.js:
```
j = eval('(' + text + ')');
```

dconnolly:
I'm taking this up. :+1:

dconnolly:
Fixed by c62aa10637c1f967ef4633924b363dd827336f11

PR #1976, opened 2015-03-25
ES6ification

mmcc:
buckle up

This isn't ready to merge yet...not even a little. I currently have a hacky way I'm going about getting tests working (custom es6 build), so that needs to be cleaned up and then another cleanup pass (or two...or 10).

I've purposefully kept this really light in terms of ES6 features. Once we've got this merged, I propose we take another pass at implementing a few more things that can be easy wins for readability, such as string interpolation and fat arrow functions. For this PR, I've only implemented `import`, `let`, and the occasional `const`. Because `let` and `const` are block scoped, I've moved almost all declarations to the point of assignment (where applicable).

There are some things such as VjsLib and VjsEvents that should be changed to just Lib and Events, and Utils should be merged with Lib.

So, if folks wouldn't mind taking a quick glance at 2,983 additions and 2,453 deletions across 66 files, that'd be wonderful. :thumbsup:

---
Related: #1720


src/js/core.js
```diff
@@ -66,76 +69,26 @@ var vjs = function(id, options, ready){
...
-    'posterImage': {},
-    'loadingSpinner': {},
-    'textTrackDisplay': {},
-    'bigPlayButton': {},
-    'controlBar': {},
-    'errorDisplay': {},
-    'textTrackSettings': {}
-  },
-
-  'language': document.getElementsByTagName('html')[0].getAttribute('lang') || navigator.languages && navigator.languages[0] || navigator.userLanguage || navigator.language || 'en',
```
heff: Where did these language options go? They're not in options.js.

mmcc: I uh...well...

that's my bad

heff: lol. I added it back in my PR

src/js/button.js
```diff
@@ -1,3 +1,8 @@
+import Component from './component';
+import * as VjsLib from './lib';
```
heff: I'm regularly typing `import Lib from './lib'` and having to fix it. Can you remind me why this is better than just exporting a Lib object with the individual functions on it?

gkatsev: Preferably, you wouldn't necessarily just import everything but rather the specific functions you need.

heff:
@videojs/core-committers we're looking to get this one in quickly so we can start migrating the other open PRs. I know it's big but try to get any comments you have in now.

sandbox/index.html.example
```diff
@@ -4,19 +4,25 @@
   <meta charset="utf-8" />
   <title>Video.js Sandbox</title>
 
-  <link href="../build/files/video-js.css" rel="stylesheet" type="text/css">
+  <link href="../build/temp/video-js.css" rel="stylesheet" type="text/css">
 
   <!-- LOAD VIDEO.JS SOURCE FILES IN ORDER -->
-  <script src="../build/source-loader.js"></script>
+  <script src="../build/temp/video.js"></script>
```
dmlap: Will the sandbox require a build before being usable now?

src/js/big-play-button.js
```diff
@@ -8,16 +11,20 @@
  * @class
  * @constructor
  */
-vjs.BigPlayButton = vjs.Button.extend();
+var BigPlayButton = Button.extend();
```
dmlap: `const` or `let`?

src/js/button.js
```diff
@@ -7,13 +12,13 @@
  * @class
  * @constructor
  */
-vjs.Button = vjs.Component.extend({
+var Button = Component.extend({
   /**
    * @constructor
    * @inheritDoc
    */
   init: function(player, options){
```
dmlap: Are we planning on keeping the old inheritance system for the long run or is that coming later? If we do stick with the current system, I think the simplified method declaration syntax is nice:

```js
vjs.Component.extend({
  init(player, options) {
    // ...
  }
})
```

src/js/events.js
```diff
@@ -5,6 +5,120 @@
...
+/**
+ * Fix a native event to have standard property values
+ * @param  {Object} event Event object to fix
+ * @return {Object}
+ * @private
+ */
+var fixEvent = function(event) {
+
+  function returnTrue() { return true; }
+  function returnFalse() { return false; }
```
dmlap: `const returnTrue = () => true; const returnFalse = () => false`. Any reason these are defined within fixEvent instead of at the module level?

src/js/events.js
```diff
@@ -5,6 +5,120 @@
...
+    // Firefox returns false for event.hasOwnProperty('type') and other props
+    //  which makes copying more difficult.
+    // TODO: Probably best to create a whitelist of event props
+    for (var key in old) {
+      // Safari 6.0.3 warns you if you try to copy deprecated layerX/Y
+      // Chrome warns you if you try to copy deprecated keyboardEvent.keyLocation
+      if (key !== 'layerX' && key !== 'layerY' && key !== 'keyLocation') {
+        // Chrome 32+ warns if you try to copy deprecated returnValue, but
+        // we still want to if preventDefault isn't supported (IE8).
+        if (!(key == 'returnValue' && old.preventDefault)) {
```
dmlap: Probably should be `===`

src/js/lib.js
```diff
@@ -373,7 +359,7 @@ vjs.TEST_VID = vjs.createEl('video');
  * @constant
  * @private
  */
-vjs.USER_AGENT = navigator.userAgent;
+var USER_AGENT = navigator.userAgent;
```
dmlap: Why do we have an alias for this?

src/js/lib.js
```diff
@@ -381,21 +367,21 @@ vjs.USER_AGENT = navigator.userAgent;
...
  * @private
  */
-vjs.IS_IPHONE = (/iPhone/i).test(vjs.USER_AGENT);
-vjs.IS_IPAD = (/iPad/i).test(vjs.USER_AGENT);
-vjs.IS_IPOD = (/iPod/i).test(vjs.USER_AGENT);
-vjs.IS_IOS = vjs.IS_IPHONE || vjs.IS_IPAD || vjs.IS_IPOD;
+var IS_IPHONE = (/iPhone/i).test(USER_AGENT);
+var IS_IPAD = (/iPad/i).test(USER_AGENT);
+var IS_IPOD = (/iPod/i).test(USER_AGENT);
+var IS_IOS = IS_IPHONE || IS_IPAD || IS_IPOD;
```
dmlap: Is it possible to define these in the export statement?

src/js/lib.js
```diff
@@ -601,7 +587,7 @@ vjs.trim = function(str){
  * @return {Number}     Rounded number
  * @private
  */
-vjs.round = function(num, dec) {
+var round = function(num, dec) {
   if (!dec) { dec = 0; }
```
dmlap: You could use default arguments instead of the explicit `if (!dec)` here.

src/js/lib.js
```diff
@@ -726,16 +709,15 @@ vjs.parseUrl = function(url) {
...
   // convert args to an array to get array functions
-  argsArray = Array.prototype.slice.call(args);
+  let argsArray = Array.prototype.slice.call(args);
   // if there's no console then don't try to output messages
   // they will still be stored in vjs.log.history
   // Was setting these once outside of this function, but containing them
   // in the function makes it easier to test cases where console doesn't exist
-  noop = function(){};
-  console = window['console'] || {
+  let noop = function(){};
```
dmlap: It's kind of fancy but I've grown to like @gkatsev's use of `Function.prototype` for noops

src/js/button.js
```diff
@@ -1,3 +1,8 @@
+import Component from './component';
+import * as Lib from './lib';
```
gkatsev: As I mentioned before, we can just ask for sepecifc things from lib and events. For example:
```js
import { createEl, bind } from './lib.js';
import { on, off } from './events.js';
```

dmlap:
This is a *mega* PR but I can't think of a better way to make this transition. I don't think we should shoot for a total, perfect conversion before pulling this in given the size of the migration. I say get it merged sooner rather than later so we can get more eyes on it.

gkatsev:
Even if it gets merged in, we could always continue to comment here.

gkatsev:
I would like a bit more time to a cursory glance at this PR, though.

sandbox/index.html.example
heff: Unfortunately yes. Browserify had no other option at least. But I guess now it might depend on what Chrome supports in ES6.

mmcc: Right now Chrome supports a good deal of what we've already implemented, but I don't think any browsers support ES6 modules, so Browserify is still a hard requirement.

src/js/big-play-button.js
mmcc: I originally had `let` for pretty much everything, but for some reason, JS Hint was complaining about exporting them.

src/js/events.js
mmcc: They're currently only used within `fixEvent`, so I'm not sure why we'd want them at the module level. Are there plans to use them elsewhere?

src/js/button.js
mmcc: For the first run through the code base it was easier to simply import everything and try and get things working. My thought was that after we got tests passing we could go back and do cleanup like this, but I think @heff has other opinions about how we handle lib / imports.

mmcc:
@dmlap yeah...I've had the same struggle. I *hate* both creating and reviewing PRs this big, but this is, unfortunately, an all or nothing transition.

heff:
I mean, it's *only* 6000 changes. Just take your lunch hour and review 100 changes per minute.

src/js/button.js
heff: Yeah, looks like I'm out voted on this one. I do like a few additional steps we discussed.
- exporting each function when it's created instead of a giant object at the end of the file
- breaking up lib into smaller focused libs, e.g. dom functions in separate file

src/js/lib.js
mmcc: Is there a benefit to that style of noop? It seems borderline too clever to me, since it's kinda obtuse at a glance.

As an aside that shouldn't be taken into consideration *at all*, I did a quick search to see if there were other benefits I wasn't aware of and came across a [relevant JSPerf](http://jsperf.com/noop-vs-function-prototype). Clearly take with a few thousand massive grains of salt, but the orange is what we have (labeled "noop"), and red is what you're proposing (labeled "function prototype"). In everything except FF 34 where it blows everything out of the water by a mile (wat), the function prototype method is the slowest.

src/js/big-play-button.js
heff: See this commit for a list of the jshint reverts https://github.com/mmcc/video.js/commit/f82e4d93105b05292950cf1f8696cb8e8764998a

src/js/button.js
gkatsev: Yeah, we don't necessarily have to do this change now.
I'm 100% for breaking up the libs file into separate files.
I believe that you can export each function as needed. Something like:
```js
module 'lib' {
  export let foo = function() {}
  export let bar = function() {}
}

module 'main' {
  import {foo, bar} from 'lib'
}
```
(The `module` syntax is kind of how you can have multiple modules in a single file, but we don't need to worry about that. It's just for demonstration purposes.)

mmcc: So on the second note, what that would end up with is importing a collection of utilities (I think we recently said we wanted `Lib` -> `Util`). So, it would look vaguely like this (but with better names maybe):

```js
import { dom, browserCheck, kittens, puppies, bryanCranston } from './util';

browserCheck.IS_IE8;
```

src/js/button.js
heff: I think we should move to the [ES6 classes](http://javascriptplayground.com/blog/2014/07/introduction-to-es6-classes-tutorial/) instead of our custom CoreObject/Class implementation. They'll be more accessible and interchangeable as ES6 classes get more use, and our current implementation doesn't do anything special. It would change `init` to `constructor`, so we'd need to figure out what larger external implications that will have.

src/js/lib.js
heff: I think just because it's used often later in the file, but probably doesn't need to be exported.

src/js/lib.js
heff: @mmcc mentioned breaking these out into their own browser/device lib, and then I think they could all just be `export IS_IPHONE ...`

mmcc:
Also, just to reiterate, there are some other easy wins I think we can get from ES6, but I think they should be done in a different PR to minimize pain and suffering. This PR is just modules + let / const, the next PR can be default params and string interpolation. If we're going to move towards using classes and such, I think that should be its own PR.

src/js/button.js
gkatsev: util is a crutch.
Why not just `import IS_IE8 from './browsers.js'`?

mmcc: I think we submitted our two comments at roughly the same time, so I never saw yours until a refresh.

There's some organizational overhead that I think we need to think through as well. If we break lib up, we're going to end up with a lot of extra noise in the root, and "browsers.js" doesn't really say anything about what it is. "utils/browsers.js" tells me that this is a group of utility functions that have something to do with browsers.

src/js/lib.js
heff: If it's actually improving something that's one thing, but otherwise you're gonna send a bunch of contributors to Google to look up wtf it does. It's not explicit.

src/js/events.js
heff: @dmlap is just saying define them in the module scope instead of each time the fixEvent function is run. They could be moved out now.

src/js/lib.js
gkatsev: `export let IS_IPHONE` I believe.

src/js/lib.js
heff: Cool!

src/js/button.js
gkatsev: It could just as well just be named `browser-checks.js`. Having it be under `util` is serving no purpose.

src/js/lib.js
gkatsev: Also, this should be a `const` probably.

src/js/big-play-button.js
gkatsev: We should consider switching from jshint to eslint. It's support for ES6 and above is far superior currently.

sandbox/index.html.example
gkatsev: We should make sure we have a `watchify` npm script we can use.

src/js/button.js
mmcc: I would say grouping purely utility functions under a utility module serves a purpose. I'm fine with not going through a `util` module to access these, but I would prefer the broken apart `lib`/`util` functionality goes under its own folder.

heff: Sounds like we'd have a utils directory for the broken out libs, which will be individual modules that you'll import individually. That sounds good to me. We still need to nail down exactly how we're exposing these functions to plugin authors, which may not affect this exactly but could. #1232

src/js/events.js
mmcc: :thumbsup: got it

src/js/button.js
gkatsev: In the main module:
```js
import { extend, bind, foo } from './lib.js';
import { on, off, bar, baz } from './events.js';

export { extend, bind, on, off, foo, bar, baz };
```
or something like that. Don't know the exact syntax.

sandbox/index.html.example
heff: We have `grunt watch` now. Does that work?

src/js/big-play-button.js
heff: I actually tried switching to eslint but I think I ran into issues with es6 modules. If you want to see if you can get past that that'd be cool.

src/js/button.js
heff: Right, that assumes we've decided to drop everything on the main object, which I don't know that we've officially landed on that, and I feel like I've lost references to a few of the specific conversations we had around that.

sandbox/index.html.example
gkatsev: We could theoretically use `grunt watch` but we really do want `watchify`. `watchify` caches the build so rebuilds take minimal amounts of time. That way, you can `watchify` the build and then when you make a change and save you don't have to wait that long before you have to refresh the page.

heff: Nice, that's great.

gkatsev: I use watchify and karma's watch to have continuous tests running in [videojs-playlist](https://github.com/brightcove/videojs-playlist/blob/master/package.json#L6-L17)

src/js/lib.js
heff: Didn't mean to shut down this thread with my comment if that happened. I am interested to know if there's benefits. I think the issue is that `noop` is a pretty ambiguous var name itself, so `noop = Function.prototype`, doesn't give you any clues unless you know one or the other.

src/js/button.js
carpasse: I really like the Idea of moving to ES6 classes but I am worried about the amount of work it will take to migrate from 4.x.x to 5.x.x.  

Once this branch is finally merged I will migrate the player we are using at MailOnline to see how much effort it takes. 
Are you already writing a migration guide for 5.0.0 if not I can start writing one while I migrate our player.

gkatsev: @carpasse theoretically, you shouldn't have any issues with the es6-changes beyond whatever breaking API changes we're having in 5.0. We definitely should have a doc and we might even make a small migration library.

mmcc: First off, a migration guide would be incredibly valuable and probably way more useful coming from someone that's actually doing it, so... :100::thumbsup:

So the goal is that this shouldn't change *that* much on the outside. There will need to be some changes regarding how one extends a VJS class, but beyond that things should be fairly straight forward (famous last words).

carpasse: If you do change to es6 classes it will force developers to do bigger changes in their existing components/plugins besides the api changes. 

Don't get me wrong I am not saying that it should not be done. I am just wondering how much effort will it take to make everything work again build, test, demos, and so on.

gkatsev: Why would ES6 classes do that?
For one, we are going to be compiled this down to ES5 code, but even then, the new class syntax can be intermixed with what we currently have just the same. The new syntax is just some sugar.

carpasse: I maybe wrong but, since you proposed to use es6 classes that means that you will use constructor, extend and super to build your components. right? 
Wouldn't that change the way we create the components?? Even if we don't migrate to es6 the component?
For example this component below:
```
vjs.FullWidthPopup = vjs.Component.extend({
  /** @constructor */
  init: function (player, options) {
    vjs.Component.call(this, player, options);
    var closeButton = new vjs.CloseButton(player, options);
    this.addChild(closeButton);
    this.player_.on('CloseButton.pressed', vjs.bind(this, function (){
      this.trigger('FullWidthPopup.close');
    }));
  }
});

vjs.FullWidthPopup.prototype.createEl = function(type, props) {
  props = props || {};
  props.className = (props.className || '') + ' vjs-full-width-popup vjs-control';
  return vjs.Component.prototype.createEl.call(this, type || 'div', props);
};
```
How would you migrate it??

mmcc: We're still talking through this, so consider this pseudo-code at best:

```js
var FullWidthPopup = videojs.extend('Component', {
  /** @constructor */
  init: function (player, options) {
    // ...
  }
});

// You'll also need to register a component for it to be a first class component citizen
videojs.registerComponent('FullWidthPopup', FullWidthPopup);
```

gkatsev: Ah, I see. Yes, that will need to change. If you don't want to adopt the new syntax, you can just use what is currently available for inheritance:
```js
var FullWidthPopup = function(player, options) {
  vjs.Component.call(this, player, options);
  // ...
};
FullWidthPopup.prototype = Object.create(vjs.Component.prototype);
```

This is something that a migration plugin could plaster over with lots of helpful logs in the console.

heff: There's a [wiki page](https://github.com/videojs/video.js/wiki/5.0-Change-Details) started to track the 5.0 changes, and it's up to date with everything that's been merged into master so far. I think it makes sense to update it when things get merged in. (feel free to add to it)

Starting the migration plugin now and continually adding to it would probably best too, if anyone wants to kick that off.

heff: A note on this: In the ES6 classes PR #1993 I've only converted Component so far, and all tests pass after adding back `init` and `Class.extend`. So it is definitely possible to add support for the old inheritance API, we just need to decide if that should be included in the core lib (with deprecation warnings) or only in a migration plugin. I think I'd be tempted to add it to core, and then remove it in a later minor release.

gkatsev: If it's in core, it cannot be removed in a minor release as that it will be a breaking change and thus not a minor version update.

heff: That's fine, it can stay until the next major. The important question is whether or not we immediately break existing plugins that use this API.

mmcc: I'm torn on this one tbh. I feel like there's a band-aid aspect that would just feel better if we made the change now and were done with it, but to have done that right I think we would have needed to let people know we were deprecating it a while back. Either that, or be prepared to help existing plugins migrate.

Either option is fine, but given the time constraints, I think we'd be better to add it to core with a deprecation notice (in the dev version at least) and remove it in the next major.

gkatsev: I think that if we can make the migration plugin and have it come out in conjunction with the 5.0 release (what's the point otherwise?), then we could just pull the plug right now.

heff: I'd love to pull the plug and just move forward. Here's a few things I'm concerned about.
- Even if we make it really easy to migrate it still requires action from the plugin authors, and that could take a long time if it happens at all. Enough API breakages and our plugin ecosystem starts over from scratch with 5.0.
- 4.0 versions of the player are going to be around for a while for various reasons (3.0 is still around). Should plugin authors immediately break their support for those players, or find some way to support both on their own?

It'd probably be worth doing a review of the plugins and seeing which ones use this. All of the techs do at least, but that's not a ton.

gkatsev: Actually, with the plugin, it hopefully won't require action from plugin authors but rather vjs and plugin users. "Upgrade to 5.0, add migration plugin". We have a small amount of plugins right now that we could probably go through each one once 5.0 is in alpha or something with a migration plugin and open up issues on the progres.

src/js/lib.js
dmlap: I like the definition closer to the export, though `export let` looks crazy to my es5 eyes.

src/js/lib.js
dmlap: Srsly. Default arguments are slick.

src/js/lib.js
dmlap: I think it's totally reasonable to stick with the inline `function(){}` for readability sake. The performance argument (without backing evidence-- add _ten_ thousand massive grains of salt) is that the inline noop is a separate function instance, eats heap space, and could lead to a GC. It might not even show up in microbenchmarks like JSPerf because they probably don't trigger garbage collection during test runs but could be noticeable in a real app.

I don't want anybody headed to Google over that many "maybes". Let's stick with `function() {}` unless we have evidence a change is actually valuable.

dmlap:
@mmcc I'm with you on making this PR the minimal es6-ification and tacking the cool stuff in further updates. This is looking great!

mmcc:
@dmlap sweet! @heff has already done some work towards the classes stuff, and we can get started on other sugar after this is merged in.

Speaking of which, it'd be great if we can get this merged in tomorrow. Get any final comments in as soon as you can so other 5.0 work can be done off this branch.

eXon:
I'm seriously glad we are going in that direction. It's time to stop writing basic stuff like this and work on what really matters. I've been using ES6 with Babel in the past 3 months and it rocks. Big kudos @mmcc for doing that gigantic pull request.

I would love to see the for loops transformed into the for-of loop. Much much cleaner and easier to read. [You can read more about it here.](http://tc39wiki.calculist.org/es6/for-of/)

An example for  `src/js/component.js`

```javascript
for (var i = 0; i < children.length; i++) {
    let child = children[i];

    let name, opts;
    if (typeof child == 'string') {
        // ['myComponent']
        name = child;
        opts = {};
    } else {
        // [{ name: 'myComponent', otherOption: true }]
        name = child.name;
        opts = child;
    }

    handleAdd(name, opts);
}
```

Would become:

```javascript
for (let child of children) {
    let name, opts;
    if (typeof child == 'string') {
        // ['myComponent']
        name = child;
        opts = {};
    } else {
        // [{ name: 'myComponent', otherOption: true }]
        name = child.name;
        opts = child;
    }

    handleAdd(name, opts);
}
```

One other thing that would be interesting is also stop using our own event system. We should be using the EventEmitter inside the built-in `event` module. I'll create a separate issue to discuss this when this is merged.


The build looks fine to me:
```
Chrome 40.0.2214 (Windows 8.1): Executed 202 of 0 SUCCESS (2.188 secs / 2.124 secs)

=============================== Coverage summary ===============================

Statements   : 77.41% ( 2999/3874 )
Branches     : 63.2% ( 1130/1788 )
Functions    : 68.87% ( 489/710 )
Lines        : 77.14% ( 2923/3789 )
================================================================================


Done, without errors.
```

gkatsev:
As much as I like for...of, it's probably not worth it at this time for this project because it will incur a pretty big hit from various shims necessarily for iterators and symbols since we need IE8 support.

eXon:
@gkatsev If we use the loose mode, we can use for-of on arrays only, but it will work on IE8. I don't think IE8 supports `Symbol.iterator`. Here's how it transpile:

```
"use strict";

for (var _iterator = children, _isArray = Array.isArray(_iterator), _i = 0, _ite
rator = _isArray ? _iterator : _iterator[Symbol.iterator]();;) {
    var _ref;

    if (_isArray) {
        if (_i >= _iterator.length) break;
        _ref = _iterator[_i++];
    } else {
        _i = _iterator.next();
        if (_i.done) break;
        _ref = _i.value;
    }

    var child = _ref;

    var _name = undefined,
        opts = undefined;
    if (typeof child == "string") {
        // ['myComponent']
        _name = child;
        opts = {};
    } else {
        // [{ name: 'myComponent', otherOption: true }]
        _name = child.name;
        opts = child;
    }

    handleAdd(_name, opts);
}
```

gkatsev:
The issue is all the other stuff like that is added like the `Symbol.iterator` polyfill and all the other plumbing necessary.

heff:
> and all the other plumbing necessary

I only see the `Symbol.iterator` which @eXon is saying we can avoid by just using for-of for arrays. What other plumbing is necessary?

eXon:
We don't need any sham/shim/shmouz. It will only add a `if` in the code and the danger that someone don't know we shouldn't use for-of on anything else than native arrays.

gkatsev:
We do need the Symbol polyfill according to babel's docs: http://babeljs.io/docs/usage/caveats/

eXon:
@gkatsev the Symbol will never be used on native arrays if we use the loose mode + only on native arrays

gkatsev:
Ah, I see. Hm... yeah, it's possible that if we limit `for...of` to just arrays it will be fine.

Gruntfile.js
```diff
@@ -408,19 +319,59 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-github-releaser');
   grunt.loadNpmTasks('grunt-aws-s3');
   grunt.loadNpmTasks('grunt-contrib-uglify');
+  grunt.loadNpmTasks('grunt-browserify');
+  grunt.loadNpmTasks('grunt-coveralls');
+  grunt.loadNpmTasks('grunt-contrib-copy');
+  grunt.loadNpmTasks('grunt-contrib-concat');
+  grunt.loadNpmTasks('grunt-exorcise');
```
gkatsev: we really should just replace this with [load-grunt-tasks](https://www.npmjs.com/package/load-grunt-tasks)

Gruntfile.js
```diff
@@ -408,19 +319,59 @@ module.exports = function(grunt) {
...
+  grunt.registerTask('default', ['build', 'test']);
+
+  // Development watch task. Doing the minimum required.
+  grunt.registerTask('dev', ['jshint', 'less', 'browserify', 'karma:chrome']);
 
-  grunt.registerTask('dist', 'Creating distribution', ['dist-copy', 'zip:dist']);
+  // Tests.
+  // We want to run things a little differently if it's coming from Travis vs local
+  if (process.env.TRAVIS) {
+    grunt.registerTask('test', ['build', 'test-travis', 'coveralls']);
```
gkatsev: is there a way of checking the coverage locally?

package.json
```diff
@@ -20,38 +20,43 @@
...
+    "blanket": "^1.1.6",
+    "browserify-istanbul": "^0.2.1",
+    "browserify-versionify": "^1.0.4",
     "chg": "~0.2.0",
-    "contribflow": "~0.2.0",
-    "github": "~0.1.14",
     "grunt": "^0.4.4",
     "grunt-aws-s3": "^0.12.1",
-    "grunt-banner": "~0.2.0",
+    "grunt-browserify": "^3.5.0",
```
gkatsev: I feel like we should just use browserify directly here instead of via grunt-browserify so we can use whatever version of browserify we want. But this shouldn't hold up this PR and can be changed anytime.

src/js/big-play-button.js
gkatsev: We can do it in another PR.

src/js/button.js
```diff
@@ -78,6 +83,8 @@ vjs.Button.prototype.onKeyPress = function(event){
 };
 
 // Blur - Remove keyboard triggers
-vjs.Button.prototype.onBlur = function(){
-  vjs.off(document, 'keydown', vjs.bind(this, this.onKeyPress));
+Button.prototype.onBlur = function(){
+  Events.off(document, 'keydown', Lib.bind(this, this.onKeyPress));
 };
+
+export default Button;
```
gkatsev: this line could probably be moved up to where we declare and define button.
```js
export default var Button = function() {};
```

gkatsev: Actually, looks like you cannot do that with `default`.
```js
let Button= function() {};
export default Button;

// or
export let Button = function() {}
export default Button;
```

src/js/component.js
```diff
@@ -359,19 +365,20 @@ vjs.Component.prototype.getChild = function(name){
  * @return {vjs.Component} The child component (created by this process if a string was used)
  * @suppress {accessControls|checkRegExp|checkTypes|checkVars|const|constantProperty|deprecated|duplicate|es5Strict|fileoverviewTags|globalThis|invalidCasts|missingProperties|nonStandardJsDocs|strictModuleDepCheck|undefinedNames|undefinedVars|unknownDefines|uselessCode|visibility}
  */
-vjs.Component.prototype.addChild = function(child, options){
-  var component, componentClass, componentName;
-
-  // If child is a string, create new component with options
+Component.prototype.addChild = function(child, options){
+  let component, componentName;
```
gkatsev: we don't need `componentName` here if it is defined in the `if` and only used in the `if`.

Also, we should consider still using `var` for variables that are function scoped. But this is a style-guide type thing.

src/js/component.js
```diff
@@ -359,19 +365,20 @@ vjs.Component.prototype.getChild = function(name){
...
+  // If child is a string, create nt with options
   if (typeof child === 'string') {
-    componentName = child;
+    let componentName = child;
 
     // Make sure options is at least an empty object to protect against errors
-    options = options || {};
+    if (!options || options === true) {
```
gkatsev: default options!

src/js/component.js
```diff
@@ -359,19 +365,20 @@ vjs.Component.prototype.getChild = function(name){
...
+      options = {};
+    }
 
     // If no componentClass in options, assume componentClass is the name lowercased
     // (e.g. playButton)
-    componentClass = options['componentClass'] || vjs.capitalize(componentName);
+    let componentClassName = options['componentClass'] || Lib.capitalize(componentName);
```
gkatsev: could be changed back to regular dot operator assuming GCC was already removed.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -5,24 +8,26 @@
...
 
-vjs.FullscreenToggle.prototype.buttonText = 'Fullscreen';
+Component.registerComponent('FullscreenToggle', FullscreenToggle);
+
+FullscreenToggle.prototype.buttonText = 'Fullscreen';
 
-vjs.FullscreenToggle.prototype.buildCSSClass = function(){
-  return 'vjs-fullscreen-control ' + vjs.Button.prototype.buildCSSClass.call(this);
+FullscreenToggle.prototype.buildCSSClass = function(){
+  return 'vjs-fullscreen-control ' + Button.prototype.buildCSSClass.call(this);
```
gkatsev: We can use the template strings here:
```js
`vjs-fullscreen-control ${Button.prototype.buildCSSClass.call(this)}`
```
In *this* particular case it isn't *that* helpful since it just gets compiled down to what we have, but thought I'd point it out.

src/js/control-bar/live-display.js
```diff
@@ -5,18 +8,20 @@
...
+
+LiveDisplay.prototype.createEl = function(){
+  var el = Component.prototype.createEl.call(this, 'div', {
     className: 'vjs-live-controls vjs-control'
   });
 
-  this.contentEl_ = vjs.createEl('div', {
+  this.contentEl_ = Lib.createEl('div', {
     className: 'vjs-live-display',
     innerHTML: '<span class="vjs-control-text">' + this.localize('Stream Type') + '</span>' + this.localize('LIVE'),
```
gkatsev: a great usecase for the template strings
```js
`<span class="vjs-control-text">${this.localize('Stream Type')}</span>${this.localize('LIVE')}`
```

src/js/control-bar/mute-toggle.js
```diff
@@ -27,18 +31,18 @@ vjs.MuteToggle = vjs.Button.extend({
   }
 });
 
-vjs.MuteToggle.prototype.createEl = function(){
-  return vjs.Button.prototype.createEl.call(this, 'div', {
+MuteToggle.prototype.createEl = function(){
+  return Button.prototype.createEl.call(this, 'div', {
     className: 'vjs-mute-control vjs-control',
     innerHTML: '<div><span class="vjs-control-text">' + this.localize('Mute') + '</span></div>'
```
gkatsev: more template strings.

src/js/control-bar/progress-control.js
```diff
@@ -1,3 +1,7 @@
+import Component from '../component';
+import Slider, { SliderHandle } from '../slider';
```
gkatsev: I would like to actually split up the slider file to have a slider and a sliderhandle file. Same with this progress-control file and most of these controls files that have more than one component in them.

src/js/control-bar/progress-control.js
```diff
@@ -196,30 +208,35 @@ vjs.PlayProgressBar.prototype.createEl = function(){
...
     'aria-live': 'off'
   });
 };
 
-vjs.SeekHandle.prototype.updateContent = function() {
-  var time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
-  this.el_.innerHTML = '<span class="vjs-control-text">' + vjs.formatTime(time, this.player_.duration()) + '</span>';
+SeekHandle.prototype.updateContent = function() {
+  let time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
+  this.el_.innerHTML = '<span class="vjs-control-text">' + Lib.formatTime(time, this.player_.duration()) + '</span>';
```
gkatsev: more template strings.

src/js/control-bar/time-display.js
```diff
@@ -28,10 +33,10 @@ vjs.CurrentTimeDisplay.prototype.createEl = function(){
   return el;
 };
 
-vjs.CurrentTimeDisplay.prototype.updateContent = function(){
+CurrentTimeDisplay.prototype.updateContent = function(){
   // Allows for smooth scrubbing, when player can't keep up.
-  var time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
-  this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Current Time') + '</span> ' + vjs.formatTime(time, this.player_.duration());
+  let time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
+  this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Current Time') + '</span> ' + Lib.formatTime(time, this.player_.duration());
```
gkatsev: more template strings.

src/js/control-bar/time-display.js
```diff
@@ -69,10 +76,10 @@ vjs.DurationDisplay.prototype.createEl = function(){
   return el;
 };
 
-vjs.DurationDisplay.prototype.updateContent = function(){
-  var duration = this.player_.duration();
+DurationDisplay.prototype.updateContent = function(){
+  let duration = this.player_.duration();
   if (duration) {
-      this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Duration Time') + '</span> ' + vjs.formatTime(duration); // label the duration time for screen reader users
+    this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Duration Time') + '</span> ' + Lib.formatTime(duration); // label the duration time for screen reader users
```
gkatsev: more template strings.

src/js/control-bar/time-display.js
```diff
@@ -101,25 +110,27 @@ vjs.TimeDivider.prototype.createEl = function(){
...
+
+RemainingTimeDisplay.prototype.createEl = function(){
+  let el = Component.prototype.createEl.call(this, 'div', {
     className: 'vjs-remaining-time vjs-time-controls vjs-control'
   });
 
-  this.contentEl_ = vjs.createEl('div', {
+  this.contentEl_ = Lib.createEl('div', {
     className: 'vjs-remaining-time-display',
     innerHTML: '<span class="vjs-control-text">' + this.localize('Remaining Time') + '</span> ' + '-0:00', // label the remaining time for screen reader users
```
gkatsev: more template strings

src/js/control-bar/time-display.js
```diff
@@ -129,12 +140,15 @@ vjs.RemainingTimeDisplay.prototype.createEl = function(){
   return el;
 };
 
-vjs.RemainingTimeDisplay.prototype.updateContent = function(){
+RemainingTimeDisplay.prototype.updateContent = function(){
   if (this.player_.duration()) {
-    this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Remaining Time') + '</span> ' + '-'+ vjs.formatTime(this.player_.remainingTime());
+    this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Remaining Time') + '</span> ' + '-'+ Lib.formatTime(this.player_.remainingTime());
```
gkatsev: more template strings

src/js/control-bar/volume-menu-button.js
```diff
@@ -40,15 +47,19 @@ vjs.VolumeMenuButton.prototype.createMenu = function(){
...
+  MuteToggle.prototype.onClick.call(this);
+  MenuButton.prototype.onClick.call(this);
 };
 
-vjs.VolumeMenuButton.prototype.createEl = function(){
-  return vjs.Button.prototype.createEl.call(this, 'div', {
+VolumeMenuButton.prototype.createEl = function(){
+  return Button.prototype.createEl.call(this, 'div', {
     className: 'vjs-volume-menu-button vjs-menu-button vjs-control',
     innerHTML: '<div><span class="vjs-control-text">' + this.localize('Mute') + '</span></div>'
```
gkatsev: more template strings

src/js/core.js
```diff
@@ -36,22 +39,22 @@ var vjs = function(id, options, ready){
     }
 
     // If a player instance has already been created for this ID return it.
-    if (vjs.players[id]) {
+    if (Player.players[id]) {
 
       // If options or ready funtion are passed, warn
       if (options) {
-        vjs.log.warn ('Player "' + id + '" is already initialised. Options will not be applied.');
+        Lib.log.warn ('Player "' + id + '" is already initialised. Options will not be applied.');
```
gkatsev: extra whitespace after `log.warn`.

src/js/fullscreen-api.js
```diff
@@ -1,82 +1,81 @@
...
 
-  // determine the supported set of functions
-  for (i=0; i<apiMap.length; i++) {
-    // check for exitFullscreen function
-    if (apiMap[i][1] in document) {
-      browserApi = apiMap[i];
-      break;
-    }
+// determine the supported set of functions
+for (let i=0; i<apiMap.length; i++) {
```
gkatsev: whitespace around operators like `=` and `<`.

src/js/json.js
```diff
@@ -4,6 +4,10 @@
  * (Compiler doesn't like JSON not being declared)
  */
```
gkatsev: As part of 5.0, we should remove this file altogether. #867

src/js/lib.js
```diff
@@ -7,15 +11,13 @@ var hasOwnProp = Object.prototype.hasOwnProperty;
  * @return {Element}
  * @private
  */
-vjs.createEl = function(tagName, properties){
-  var el;
-
+var createEl = function(tagName, properties){
   tagName = tagName || 'div';
```
gkatsev: default arguments.

src/js/lib.js
```diff
@@ -547,7 +533,7 @@ vjs.el = function(id){
  * @return {String}         Time formatted as H:MM:SS or M:SS
  * @private
  */
-vjs.formatTime = function(seconds, guide) {
+var formatTime = function(seconds, guide) {
   // Default to using seconds as guide
   guide = guide || seconds;
```
gkatsev: default arguments
```js
let formatTime = (seconds, guide = seconds) => {
  ...
}
```

src/js/lib.js
```diff
@@ -654,12 +640,11 @@ vjs.setLocalStorage = function(key, value){
...
  */
-vjs.getAbsoluteURL = function(url){
-
+var getAbsoluteURL = function(url){
   // Check if absolute URL
   if (!url.match(/^https?:\/\//)) {
     // Convert to absolute URL. Flash hosted off-site needs an absolute URL.
-    url = vjs.createEl('div', {
+    url = createEl('div', {
       innerHTML: '<a href="'+url+'">x</a>'
```
gkatsev: template strings`

src/js/lib.js
```diff
@@ -667,26 +652,24 @@ vjs.getAbsoluteURL = function(url){
...
   // IE8 (and 9?) Fix
   // ie8 doesn't parse the URL correctly until the anchor is actually
   // added to the body, and an innerHTML is needed to trigger the parsing
-  addToBody = (a.host === '' && a.protocol !== 'file:');
+  let addToBody = (a.host === '' && a.protocol !== 'file:');
+  let div;
   if (addToBody) {
-    div = vjs.createEl('div');
+    div = createEl('div');
     div.innerHTML = '<a href="'+url+'"></a>';
```
gkatsev: template strings.

src/js/lib.js
gkatsev: Making decisions based on a jsperf is silly anyway :P

src/js/lib.js
```diff
@@ -807,21 +789,21 @@ vjs.findPosition = function(el) {
...
 
-  clientTop = docEl.clientTop || body.clientTop || 0;
-  scrollTop = window.pageYOffset || body.scrollTop;
-  top = box.top + scrollTop - clientTop;
+  const clientTop = docEl.clientTop || body.clientTop || 0;
+  const scrollTop = window.pageYOffset || body.scrollTop;
+  const top = box.top + scrollTop - clientTop;
 
   // Android sometimes returns slightly off decimal values, so need to round
   return {
```
gkatsev: if the values get rounded when set, we can just do
```js
return {left, top}
```

src/js/media/flash.js
```diff
@@ -12,56 +19,55 @@
...
  */
-vjs.Flash = vjs.MediaTechController.extend({
+var Flash = MediaTechController.extend({
   /** @constructor */
   init: function(player, options, ready){
-    vjs.MediaTechController.call(this, player, options, ready);
+    MediaTechController.call(this, player, options, ready);
 
-    var source = options['source'],
+    let source = options['source'];
```
gkatsev: We could also do `let {source, parentEl} = options;`

src/js/media/flash.js
```diff
@@ -391,11 +391,11 @@ vjs.Flash.getEmbedCode = function(swf, flashVars, params, attributes){
   }, params);
 
   // Create param tags string
-  vjs.obj.each(params, function(key, val){
+  Lib.obj.each(params, function(key, val){
     paramsString += '<param name="'+key+'" value="'+val+'" />';
```
gkatsev: template strings.

src/js/media/flash.js
```diff
@@ -355,35 +357,33 @@ vjs.Flash.version = function(){
...
 };
 
-vjs.Flash.getEmbedCode = function(swf, flashVars, params, attributes){
-
-  var objTag = '<object type="application/x-shockwave-flash" ',
-      flashVarsString = '',
-      paramsString = '',
-      attrsString = '';
+Flash.getEmbedCode = function(swf, flashVars, params, attributes){
+  const objTag = '<object type="application/x-shockwave-flash" ';
```
gkatsev: amazing usecase for template strings.

src/js/media/flash.js
```diff
@@ -406,9 +406,14 @@ vjs.Flash.getEmbedCode = function(swf, flashVars, params, attributes){
   }, attributes);
 
   // Create Attributes string
-  vjs.obj.each(attributes, function(key, val){
+  Lib.obj.each(attributes, function(key, val){
     attrsString += (key + '="' + val + '" ');
   });
 
   return objTag + attrsString + '>' + paramsString + '</object>';
```
gkatsev: template strings

src/js/player.js
```diff
@@ -33,23 +55,23 @@ vjs.Player = vjs.Component.extend({
...
-    options = vjs.obj.merge(this.getTagSettings(tag), options);
+    options = Lib.obj.merge(this.getTagSettings(tag), options);
 
     // Update Current Language
-    this.language_ = options['language'] || vjs.options['language'];
+    this.language_ = options['language'] || options['language'];
 
     // Update Supported Languages
-    this.languages_ = options['languages'] || vjs.options['languages'];
+    this.languages_ = options['languages'] || options['languages'];
```
gkatsev: oops, we have a shadowing issue here. The global and local options are called the same. We should change the global options to be called `opts` or something like that instead.

src/js/lib.js
mmcc: I think (for now) this needs to stay a var because of the JSHint issues @heff was running into. Perhaps switching to eslint will fix this?

src/js/player.js
```diff
@@ -632,14 +656,14 @@ vjs.Player.prototype.techGet = function(method){
     } catch(e) {
       // When building additional tech libs, an expected method may not be defined yet
       if (this.tech[method] === undefined) {
-        vjs.log('Video.js: ' + method + ' method not defined for '+this.techName+' playback technology.', e);
```
gkatsev: template strings

src/js/player.js
```diff
@@ -1464,7 +1490,7 @@ vjs.Player.prototype.error = function(err){
 
   // log the name of the error type and any message
   // ie8 just logs "[object object]" if you just log the error object
-  vjs.log.error('(CODE:'+this.error_.code+' '+vjs.MediaError.errorTypes[this.error_.code]+')', this.error_.message, this.error_);
+  Lib.log.error('(CODE:'+this.error_.code+' '+MediaError.errorTypes[this.error_.code]+')', this.error_.message, this.error_);
```
gkatsev: template strings

src/js/tracks/text-track-settings.js
```diff
@@ -1,284 +1,283 @@
...
-  function captionOptionsMenuTemplate() {
-    return '<div class="vjs-tracksettings">' +
-        '<div class="vjs-tracksettings-colors">' +
-          '<div class="vjs-fg-color vjs-tracksetting">' +
-              '<label class="vjs-label">Foreground</label>' +
+  target.selectedIndex = i;
+}
+
+function captionOptionsMenuTemplate() {
+  return '<div class="vjs-tracksettings">' +
```
gkatsev: this would also work great as a template string.

src/js/xhr.js
```diff
@@ -28,8 +32,8 @@
  * @param  {Function}       callback  The callback function
  * @returns {Object}                  The request
  */
-vjs.xhr = function(options, callback){
-  var XHR, request, urlInfo, winLoc, fileUrl, crossOrigin, abortTimeout, successHandler, errorHandler;
+var xhr = function(options, callback){
```
gkatsev: we should swap this file out for `xhr` from npm, but we can do that later, not even in the 5.0 but maybe in a 5.1 release, even.

gkatsev: We can also consider using `fetch` and the `fetch` polyfill.

test/index.html
```diff
@@ -1,97 +0,0 @@
-<!DOCTYPE HTML>
```
gkatsev: it would be nice to still have a test/index.html file we can use to load in the tests manually in the browser.

src/js/lib.js
gkatsev: It's fine for leaving it as const for now. switching to eslint can be another PR.

eXon:
Bad news if we want to use for-of: Array.isArray is not even IE8 compatible. The pollyfill is one line of code though.

```javascript
if (!Array.isArray) {
  Array.isArray = function(arg) {
    return Object.prototype.toString.call(arg) === '[object Array]';
  };
}
```

I don't think we need it at this point. I think it's worthless to talk about it at this point anyway.

gkatsev:
@eXon we're thinking of making es5-shim as requirement, so, might not be an issue.

src/js/lib.js
mmcc: I personally prefer the notion of having all the exports for a file defined in a single place and having it be consistent. It seems a bit like an anti-pattern to have it one way in the entire project, but export at definition for just things related to lib.

I think simply breaking up the files in the first place is going to mean we don't have the giaaant export statement, so perhaps this isn't as big of a deal.

Gruntfile.js
mmcc: Yes, every test run shows the coverage in the test output, and there's also a txt file break down generated in `test/coverage/coverage.txt`. Generating the HTML file unfortunately throws errors right now and I don't think it's worth the time to debug.

src/js/button.js
heff: Yeah, it seems like there's a pattern emerging (among ES6 modules at least) where imports are at the top and exports are at the bottom. I like that myself, at least for consistency.

src/js/button.js
heff: Babel 5.0 is out with some [export extensions](http://babeljs.io/blog/2015/03/31/5.0.0/#stage-1:-export-extensions) including `export * as ns from "mod";`. I can't tell yet if that helps this util exporting discussion.

src/js/control-bar/progress-control.js
mmcc: Agreed, I think we should move things like this to their own folders and split apart into multiple files.

#2010

src/js/component.js
mmcc: #2009

src/js/control-bar/fullscreen-toggle.js
mmcc: #2009

src/js/control-bar/live-display.js
mmcc: #2009

src/js/control-bar/mute-toggle.js
mmcc: #2009

src/js/control-bar/progress-control.js
mmcc: #2009

src/js/control-bar/time-display.js
mmcc: #2009

src/js/json.js
mmcc: #2010

src/js/control-bar/progress-control.js
heff: +1

src/js/control-bar/mute-toggle.js
heff: For some of these template string references we should really just clean up the logic and store the inner-most element as the `contentEl_`, so we can just do `this.contentEl_.innerHTML = this.localize('Mute')`. That would help improve [line 57](https://github.com/videojs/video.js/pull/1976/files#diff-c5678798938810795d9c25bc9e2236ecL57) of this file... :expressionless: 
```
if(this.el_.children[0].children[0].innerHTML!=this.localize('Unmute')){
```

src/js/xhr.js
heff: Sounds good to me if you want to open an issue for either. I'm assuming the XHR from npm wouldn't change the API then, yeah?

test/index.html
heff: This file had each test individually listed. Is there a good way to pull in all the test files without manually listing them? I guess we could always browserify them into one file, but we probably don't want to add that to the main watch task.

heff:
There's some open discussions on this still but nothing that seems to be blocking and can't be pulled out into its own issue, so merging this in. (phew!)

@mmcc has updated the [5.0 change wiki page](https://github.com/videojs/video.js/wiki/5.0-Change-Details), and started the [ES6 feature guide](https://github.com/videojs/video.js/wiki/ES6-Features-used-in-the-source) for contributors.

ES6 Classes (#1993) is the next big one. At this point no one appears to be against the idea, we just need a clean migration story, so I'm going to move forward with it and hopefully get it in quickly. If we want to be ready by the end of April we need to keep moving fast, so we could certainly use help knocking down more of the [5.0 issues](https://github.com/videojs/video.js/milestones/v5.0.0).

test/index.html
gkatsev: Yep, just have separate watch/build tasks that are specifically for tests and then include that bundled output. [Example](https://github.com/brightcove/videojs-playlist/blob/master/package.json#L12-L16).

gkatsev:
Awesome!

src/js/button.js
carpasse: I really like the plugin idea, that would keep the core code cleaner. When you update a major version you need to take a look to know what has changed and fix it. If by adding one extra plugin/script you can make make your code temporary work. It gives you the opportunity to start migrating all your code to es6 on separates steps instead of all at once.

src/js/component.js
carpasse: I know it has been discussed before but wouldn't it be better to do the opposite and always use let??

src/js/button.js
heff: Yeah, I like that. I just need to get more familiar with how the migrate plugin would work. Like would it be throwing deprecation warnings for older APIs or just silently make them available again?

carpasse: I would put a warning on every component/plugin registered the old way.

src/js/events.js
```diff
@@ -5,6 +5,120 @@
...
+    // The event occurred on this element
+    if (!event.target) {
+      event.target = event.srcElement || document;
+    }
+
+    // Handle which other element the event is related to
+    event.relatedTarget = event.fromElement === event.target ?
+      event.toElement :
+      event.fromElement;
+
```
carpasse: This is a bit unrelated and I should raise an issue but ...
Can we add a check to see if event.relatedTarget exists??
Firefox supports it https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget and it does not support event.fromElement so in Firefox event.relatedTarget is undefined 
:( We found this bug the other day.

src/js/component.js
gkatsev: Variables should definitely be defined in as small a scope as possible, i.e., inside blocks, and as close to where they are needed. Sometimes, you have function-level variables. Those you could define with either a let or a var but theoretically, it would be helpful to know whether that variable is long lived, i.e., for the full function. You could signal that by declaring the variable with a var instead of a let in that particular case. I'm not sure that it's worth it yet. Just throwing out some ideas.
If you have any thoughts on it, I'd be happy to hear them :)

src/js/events.js
heff: I don't see why not. Want to make a PR for it?

src/js/component.js
carpasse: I see the value of marking somehow long lived variables. But if you enforce the use of 'let' you are basically sending a very clear message. Variables should be block scoped whenever possible. People is used to 'var' and they will unconsciously use it whenever possible but with 'let' the will be forced to think on the visibility of the variable.  I also makes it easier. You don't have to think do I use let or var everytime you create a variable.

src/js/button.js
gkatsev: Yep, the plugin should be logging profusely to let users know what changed and what to do instead.
Use the [jquery-migrate](https://github.com/jquery/jquery-migrate/) plugin for inspiration.

src/js/events.js
carpasse: Sure, where should I make the pull request? on this branch? is that possible?

mmcc: This PR is merged now, so you can just make it against master.

carpasse: Ups, I did not realize it was already merged. I will create the PR tomorrow morning.

mischizzle: @mmcc @carpasse Created an issue https://github.com/videojs/video.js/issues/2024 and starting the pull request.

PR #1990, opened 2015-03-30
update pt-BR.json

deedos:
change 'progressão' to 'progresso"


mmcc:
Thanks for making the change! We just merged the other PR, so this one should be unnecessary now.

deedos:
yep.. such a small update.. i did not read the whole conversation .
cheers

PR #1993, opened 2015-03-31
ES6 Classes

heff:
<strike>This is built on #1976 so the commit to look at is https://github.com/videojs/video.js/commit/b7f6934d0cc4b6262f5c9ab6751e337a1259d300</strike>
An easier view of the changes here: https://github.com/mmcc/video.js/pull/7

So far I just converted Component to an ES6 class, and then re-added `init` and `extend` so that all the subclasses would still work (all tests are passing). After fully implementing this we would't have either of those methods anymore.

### Important notes
1. ES6 classes (through babelify) requires an Object.create shim. Before this we don't yet need to include the shim/sham combo, but after this we would.
2. This will remove the `MyComponent.extend()` static method, for all components, techs, etc., breaking that major API. We can't automatically add the static extend() method in the ES6 inheritance process like we're doing now for every sub class. We can manually add it for each component if we think that's important for backwards compatibility, maybe at least for techs. Otherwise we should introduce a `videojs.extends/videojs.inherits` for external projects that don't use ES6 classes. We can even just use Node's.

Please let me know any thoughts on this.


heff:
So apparently static methods are inherited in ES6 classes, so that makes the old `extend()` API really easy to add since it only needs to be added once.

Also I'm using the latest version of babelify but it's yelling at me for something I think it shouldn't.
```
createEl() {
  return super();
}

SyntaxError: slider/slider.js: Line 48: Direct super call is illegal in non-constructor, use super."createEl"() instead while parsing file
```

I instead have to do
```
createEl() {
  return super.createEl();
}
```

The first way works in the babel demo interface, so I'm not sure what's going on there.

gkatsev:
I think the spec changed to be the second example. I can find out for sure later.

heff:
The ES6 class work has been updated and all classes have been converted. All tests are passing so it could be merged in. I broke out any individual class into its own file, and some of the library got reorganized because of that.

This was mostly a syntax and organization change, with almost no functionality change. However, there's a limitation where you can't use `this` before `super` in a constructor, and I had to change when some variables are set. I was pretty careful about the order of things so I don't believe anything functionality should have changed around that.

However, I've realized we should be using `options` more to pass info to super classes, instead of setting values on the prototype. For example we have a bunch of `featuresX` properties on techs, and we really should be passing that through options.

I also renamed MediaTechController to Tech and the /media/ directory to /tech/ for clarity. But I also aliased Tech to MediaTechController so old techs won't break on that alone.

I have not converted over all static methods, and not sure if we need to.

Also, this and #2015...
1315301583_sudden_headon_collision

mmcc:
hide

src/js/big-play-button.js
```diff
@@ -11,20 +10,21 @@ import Button from './button';
...
+  }
 
-BigPlayButton.prototype.onClick = function(){
-  this.player_.play();
-};
+}
 
-export default BigPlayButton;
+Button.registerComponent('BigPlayButton', BigPlayButton);
+export default BigPlayButton;
```
mmcc: Am I reading this right that this is removing the trailing newline? I won't comment on every instance of this, but I feel as strongly about a trailing newline as I do about removing empty white space.

mmcc: Which is to say, very strongly. I'm pretty sure revolutions have been fought over the need for trailing newlines.

src/js/button.js
```diff
@@ -27,64 +24,65 @@ var Button = Component.extend({
...
+  createEl(type, props) {
+    // Add standard Aria and Tabindex info
+    props = Lib.obj.merge({
+      className: this.buildCSSClass(),
+      'role': 'button',
+      'aria-live': 'polite', // let the screen reader user know that the text of the button may change
+      tabIndex: 0
+    }, props);
+
+    let el = Component.prototype.createEl.call(this, type, props);
```
mmcc: Isn't `Component` `super` in this case?

src/js/button.js
```diff
@@ -27,64 +24,65 @@ var Button = Component.extend({
...
 
   // Click - Override with specific functionality for button
-Button.prototype.onClick = function(){};
+  onClick(){}
 
   // Focus - Add keyboard functionality to element
-Button.prototype.onFocus = function(){
-  Events.on(document, 'keydown', Lib.bind(this, this.onKeyPress));
-};
+  onFocus(){
```
mmcc: Nitpick: let's be consistent about spacing. `buildCSSClass() {` vs `onFocus(){`. I personally prefer the space, but let's just do it one way.

src/js/button.js
```diff
@@ -27,64 +24,65 @@ var Button = Component.extend({
...
-};
 
-// Blur - Remove keyboard triggers
-Button.prototype.onBlur = function(){
-  Events.off(document, 'keydown', Lib.bind(this, this.onKeyPress));
-};
+}
 
+
+Component.registerComponent('Button', Button);
```
mmcc: :+1: I like the idea of registering components and exports all at the bottom of each module.

mmcc:
This isn't a problem with the PR itself, obviously, but this diff is useless.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
+   *       ]
+   *     });
+   *
+   */
+  initChildren(){
+    let parent = this;
+    let parentOptions = parent.options();
+    let children = parentOptions['children'];
+
+    let handleAdd;
```
mmcc: Why is this defined outside of the `if` block if it's only used within? The only other path is the else, which could easily be changed to `Lib.obj.each(children, null);` and honestly be more readable.

mmcc: Wait a second, think the diff was screwing with me there, it doesn't look like there's *any* reason to define this outside of the `if` block.

src/js/big-play-button.js
heff: Ha, I guess that's what it is. And I guess Atom isn't configured to keep those for me. I'll update it.

src/js/button.js
heff: Yeah, I didn't do the most exhaustive search of old super method use cases, except for in the constructor. I guess I should do that.

src/js/button.js
heff: Yeah, I'm good with the space. I'd love to use escodegen to just rewrite the whole lib with the right formatting. Maybe after all of this is done.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
+   * you need to call `vjs.trigger(el, 'dispose')` on it to clean up
+   * references to it and allow the browser to garbage collect it.
+   *
+   * @param  {String|vjs.Component} first   The event type or other component
+   * @param  {Function|String}      second  The event handler or event type
+   * @param  {Function}             third   The event handler
+   * @return {vjs.Component}        self
+   */
+  on(first, second, third){
+    var target, type, fn, removeOnDispose, cleanRemover, thisComponent;
```
mmcc: Can we decide what we want to do about this from a style perspective? I, for one, **much** prefer to define a variable at the point of use or as close as possible to it, but I understand this is a bit of stylistic taste.

FWIW, I found *quite* a few unused variables when I did the first ES6 pass, and this style is inherently prone to that kind of thing.

src/js/button.js
heff: Cool. One interesting thing that came out of this is every component has an inherited `registerComponent` function, so we don't have to include Component just to register something. Though I couldn't decide if I should register the component with it's own register function or its super classes.
```
Button.registerComponent('PlayToggle', PlayToggle);
or
PlayToggle('PlayToggle', PlayToggle);
```

heff: Sorry, chopped off part of the example. It's fixed now.

```
Button.registerComponent('PlayToggle', PlayToggle);
or
PlayToggle.registerComponent('PlayToggle', PlayToggle);
```

heff:
Yeah, the diff is bad. I could have avoided moving files for the first pass, but even in the files I didn't move, the complete restructuring of the class makes it impossible to tell what exactly changed.

mmcc:
It's a lot better if you switch to the split diff, so that's what I'm using now.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
-    this.el_.parentNode.removeChild(this.el_);
-  }
+    // Remove all event listeners.
+    this.off();
 
-  Lib.removeData(this.el_);
-  this.el_ = null;
-};
+    // Remove element from DOM
+    if (this.el_.parentNode) {
```
mmcc: Should we use the `#el()` method instead of the "private" var when just reading this (throughout the module)?

src/js/control-bar/duration-display.js
```diff
@@ -0,0 +1,48 @@
...
+    // this might need to be changed to 'durationchange' instead of 'timeupdate' eventually,
+    // however the durationchange event fires before this.player_.duration() is set,
+    // so the value cannot be written out using this method.
+    // Once the order of durationchange and this.player_.duration() being set is figured out,
+    // this can be updated.
+    this.on(player, 'timeupdate', this.updateContent);
+  }
+
+  createEl() {
+    let el = Component.prototype.createEl.call(this, 'div', {
```
mmcc: `super.createEl`?

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -8,33 +7,25 @@ import Button from '../button';
...
-   */
-  init: function(player, options){
-    Button.call(this, player, options);
-  }
-});
+class FullscreenToggle extends Button {
 
-Component.registerComponent('FullscreenToggle', FullscreenToggle);
+  buildCSSClass() {
+    return 'vjs-fullscreen-control ' + Button.prototype.buildCSSClass.call(this);
```
mmcc: `super`

src/js/control-bar/mute-toggle.js
```diff
@@ -29,50 +29,51 @@ var MuteToggle = Button.extend({
...
-});
 
-MuteToggle.prototype.createEl = function(){
-  return Button.prototype.createEl.call(this, 'div', {
-    className: 'vjs-mute-control vjs-control',
-    innerHTML: '<div><span class="vjs-control-text">' + this.localize('Mute') + '</span></div>'
-  });
-};
+  createEl() {
+    return Button.prototype.createEl.call(this, 'div', {
```
mmcc: super

src/js/control-bar/mute-toggle.js
```diff
@@ -29,50 +29,51 @@ var MuteToggle = Button.extend({
...
+        level = 3;
 
-  if (vol === 0 || this.player_.muted()) {
-    level = 0;
-  } else if (vol < 0.33) {
-    level = 1;
-  } else if (vol < 0.67) {
-    level = 2;
-  }
+    if (vol === 0 || this.player_.muted()) {
```
mmcc: This might be outside of the scope of this PR, but perhaps this would be more readable as a switch statement.

src/js/control-bar/play-toggle.js
```diff
@@ -9,45 +8,45 @@ import * as Lib from '../lib';
...
-
-Component.registerComponent('PlayToggle', PlayToggle);
 
-PlayToggle.prototype.buttonText = 'Play';
-
-PlayToggle.prototype.buildCSSClass = function(){
-  return 'vjs-play-control ' + Button.prototype.buildCSSClass.call(this);
-};
+  buildCSSClass() {
+    return 'vjs-play-control ' + Button.prototype.buildCSSClass.call(this);
```
mmcc: super

src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
```diff
@@ -0,0 +1,108 @@
...
+
+    this.updateVisibility();
+    this.updateLabel();
+
+    this.on(player, 'loadstart', this.updateVisibility);
+    this.on(player, 'ratechange', this.updateLabel);
+  }
+
+  createEl() {
+    let el = MenuButton.prototype.createEl.call(this);
```
mmcc: super

src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  updateARIAAttributes() {
+    // Current playback rate
+    this.el().setAttribute('aria-valuenow', this.player().playbackRate());
+  }
+
+  onClick() {
+    // select next rate option
+    let currentRate = this.player().playbackRate();
```
mmcc: We definitely need to be more consistent. We use `this.player_` and `this.player()` seemingly at random.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -0,0 +1,93 @@
...
+class SeekBar extends Slider {
+
+  constructor(player, options){
+    super(player, options);
+    this.on(player, 'timeupdate', this.updateARIAAttributes);
+    player.ready(Lib.bind(this, this.updateARIAAttributes));
+  }
+
+  createEl() {
+    return Slider.prototype.createEl.call(this, 'div', {
```
mmcc: super - I'm starting to get pretty weak on these images

src/js/control-bar/mute-toggle.js
gkatsev: This doesn't translate to a switch since they're inequalities and not concrete values.
Also, switch is just a glorified if/else.
What about?
```js
level = (vol === 0)  ? 0 :
        (vol < 0.33) ? 1 :
        (vol < 0.67) ? 2 ;
```
or maybe
```js
if (vol === 0 || this.player_.muted()) {
  level 0;
} else
if (vol < 0.33) {
  level 1;
} else
if (vol < 0.67) {
  level = 2;
}
```
Or maybe
```js
if (vol === 0 || this.player_.muted()) {
  level = 0;                           } else
if (vol < 0.33) {
  level = 1;                           } else
if (vol < 0.67) {
  level = 2;                           }
```

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -0,0 +1,93 @@
...
+    this.player_.currentTime(this.player_.currentTime() + 5); // more quickly fast forward for keyboard-only users
+  }
+
+  stepBack() {
+    this.player_.currentTime(this.player_.currentTime() - 5); // more quickly rewind for keyboard-only users
+  }
+
+}
+
+SeekBar.prototype.options_ = {
```
mmcc: Any reason we're adding this to the prototype instead of defining it in the constructor?

gkatsev: Static properties like this can be shared across instances as opposed to being created each time in the constructor. Whether we want to bring it into the constructor will depend on what we want out of these options.

src/js/control-bar/text-track-controls/captions-button.js
```diff
@@ -0,0 +1,49 @@
...
+class CaptionsButton extends TextTrackButton {
+
+  constructor(player, options, ready){
+    super(player, options, ready);
+    this.el_.setAttribute('aria-label','Captions Menu');
+  }
+
+  update() {
+    let threshold = 2;
+    TextTrackButton.prototype.update.call(this);
```
mmcc: super...bowl?

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -0,0 +1,74 @@
...
+class VolumeBar extends Slider {
+
+  constructor(player, options){
+    super(player, options);
+    this.on(player, 'volumechange', this.updateARIAAttributes);
+    player.ready(Lib.bind(this, this.updateARIAAttributes));
+  }
+
+  createEl() {
+    return Slider.prototype.createEl.call(this, 'div', {
```
mmcc: super

src/js/control-bar/volume-control/volume-control.js
```diff
@@ -0,0 +1,47 @@
...
+      if (player.tech['featuresVolumeControl'] === false) {
+        this.addClass('vjs-hidden');
+      } else {
+        this.removeClass('vjs-hidden');
+      }
+    });
+  }
+
+  createEl() {
+    return Component.prototype.createEl.call(this, 'div', {
```
mmcc: super

src/js/poster-image.js
```diff
@@ -0,0 +1,101 @@
...
+    this.update();
+    player.on('posterchange', Lib.bind(this, this.update));
+  }
+
+  /**
+   * Clean up the poster image
+   */
+  dispose() {
+    this.player().off('posterchange', this.update);
+    Button.prototype.dispose.call(this);
```
mmcc: super

src/js/tech/html5.js
```diff
@@ -0,0 +1,682 @@
...
+      }
+    });
+
+    this.triggerReady();
+  }
+
+
+  dispose() {
+    Html5.disposeMediaElement(this.el_);
+    Tech.prototype.dispose.call(this);
```
mmcc: `super`

src/js/tech/html5.js
```diff
@@ -0,0 +1,682 @@
...
+    };
+
+    this.remoteTextTracks().addTrack_(track.track);
+
+    return track;
+  }
+
+  removeRemoteTextTrack(track) {
+    if (!this['featuresNativeTextTracks']) {
+      return Tech.prototype.removeRemoteTextTrack.call(this, track);
```
mmcc: `super`

src/js/tech/html5.js
```diff
@@ -0,0 +1,682 @@
...
+    if (!this['featuresNativeTextTracks']) {
+      return Tech.prototype.addTextTrack.call(this, kind, label, language);
+    }
+
+    return this.el_.addTextTrack(kind, label, language);
+  }
+
+  addRemoteTextTrack(options) {
+    if (!this['featuresNativeTextTracks']) {
+      return Tech.prototype.addRemoteTextTrack.call(this, options);
```
mmcc: `super`

src/js/tech/html5.js
```diff
@@ -0,0 +1,682 @@
...
+  textTracks() {
+    if (!this['featuresNativeTextTracks']) {
+      return Tech.prototype.textTracks.call(this);
+    }
+
+    return this.el_.textTracks;
+  }
+  addTextTrack(kind, label, language) {
+    if (!this['featuresNativeTextTracks']) {
+      return Tech.prototype.addTextTrack.call(this, kind, label, language);
```
mmcc: `super`

src/js/tech/html5.js
```diff
@@ -0,0 +1,682 @@
...
+
+  playbackRate() { return this.el_.playbackRate; }
+  setPlaybackRate(val) { this.el_.playbackRate = val; }
+
+  networkState() { return this.el_.networkState; }
+  readyState() { return this.el_.readyState; }
+
+  textTracks() {
+    if (!this['featuresNativeTextTracks']) {
+      return Tech.prototype.textTracks.call(this);
```
mmcc: `suuuper`

src/js/tech/tech.js
```diff
@@ -0,0 +1,536 @@
...
+    this.player().trigger('timeupdate');
+  }
+
+  dispose() {
+    // Turn off any manual progress or timeupdate tracking
+    if (this.manualProgress) { this.manualProgressOff(); }
+
+    if (this.manualTimeUpdates) { this.manualTimeUpdatesOff(); }
+
+    Component.prototype.dispose.call(this);
```
mmcc: `superduper`

src/js/tracks/text-track-display.js
```diff
@@ -0,0 +1,185 @@
...
+  toggleDisplay() {
+    if (this.player_.tech && this.player_.tech['featuresNativeTextTracks']) {
+      this.hide();
+    } else {
+      this.show();
+    }
+  }
+
+  createEl() {
+    return Component.prototype.createEl.call(this, 'div', {
```
mmcc: `super`

mmcc:
I'll take another pass tomorrow, but first pass showed a few things that can refactored to `super` calls.

Just a note not necessarily related to this PR: things seem a bit inconsistent about when we use things like `el_` and `player_` versus their respective getter methods.

src/js/button.js
gkatsev: Yep, spacing. It would be interesting to have a pre-commit hook that modifying the code styles based on output from node-jscs and jshint/eslint.
Also, I'm still debating whether we should have a space before `()` in the concise methods.
i.e.
```js
{
  foo() {}
}
// vs
{
  foo () {}
}
```

For function statements, we're supposed to put a space after the name and this sets it apart from where we call the function and where we declare the function. The same could be applied to the concise methods.

src/js/component.js
```diff
@@ -3,7 +3,7 @@
  *
  */
 
-import CoreObject from './core-object.js';
+// import CoreObject from './core-object.js';
```
gkatsev: remove if not needed.

src/js/component.js
```diff
@@ -38,14 +38,16 @@ import window from 'global/window';
...
-   */
-  init: function(player, options, ready){
-    this.player_ = player;
+class Component {
+
+  constructor(player, options, ready){
+
+    // The component might be the player itself and we can't pass `this` to super
+    if (!player && this.play) {
+      this.player_ = player = this;
```
gkatsev: Does the player not get the player argument?
I wonder if we can use default args here:
```js
class Component {
  constructor (player = this, options, ready) {
    this.player_ = player;
  }
}
```

src/js/component.js
```diff
@@ -65,14 +67,20 @@ var Component = CoreObject.extend({
     this.name_ = options['name'] || null;
 
     // Create element if one wasn't provided in options
-    this.el_ = options['el'] || this.createEl();
+    if (options['el']) {
```
gkatsev: `.el`

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
   }
-});
 
-/**
- * Dispose of the component and all child components
- */
-Component.prototype.dispose = function(){
-  this.trigger({ type: 'dispose', 'bubbles': false });
+  // Temp for ES6 class transition, remove before 5.0
+  init() {
```
gkatsev: when this PR is merged in, we probably want to start on the videojs-migrate plugin.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
-        this.children_[i].dispose();
+  /**
+   * Dispose of the component and all child components
+   */
+  dispose() {
+    this.trigger({ type: 'dispose', 'bubbles': false });
+
+    // Dispose all children.
+    if (this.children_) {
+      for (var i = this.children_.length - 1; i >= 0; i--) {
```
gkatsev: if we are making es5-shim a requirement, I would love to turn all for-loops into iterators.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
-Component.prototype.el = function(){
-  return this.el_;
-};
+  /**
+   * Return the component's DOM element where children are inserted.
+   * Will either be the same as el() or a new element defined in createEl().
+   *
+   * @return {Element}
+   */
+  contentEl(){
```
gkatsev: what's the difference between `contentEl` and `el`? Should be have both? Why not just insert children into `el`?

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
+   *       }
+   *     });
+   *
+   * @param {String|vjs.Component} child The class name or instance of a child to add
+   * @param {Object=} options Options, including options to be passed to children of the child.
+   * @return {vjs.Component} The child component (created by this process if a string was used)
+   * @suppress {accessControls|checkRegExp|checkTypes|checkVars|const|constantProperty|deprecated|duplicate|es5Strict|fileoverviewTags|globalThis|invalidCasts|missingProperties|nonStandardJsDocs|strictModuleDepCheck|undefinedNames|undefinedVars|unknownDefines|uselessCode|visibility}
+   */
+  addChild(child, options){
+    let component, componentName;
```
gkatsev: `componentName` shouldn't be defined out here.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
-  // If a name wasn't used to create the component, check if we can use the
-  // name function of the component
-  componentName = componentName || (component.name && component.name());
+    if (!component || !this.children_) return;
 
-  if (componentName) {
-    this.childNameIndex_[componentName] = component;
-  }
+    var childFound = false;
+    for (var i = this.children_.length - 1; i >= 0; i--) {
```
gkatsev: should be a `let`

src/js/component.js
gkatsev: I've been coming around on using the variables where needed with let. Also, only having on variable per declaration.
However, if possible, declaring all variables at the top of the scope/block is still preferable.

gkatsev: Also, it would be nice to move all this `on` stuff into the `EventEmitter` class we have and then have Component extend from it or something like that.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
+   * and document (both permanent), if you remove the element from the DOM
+   * you need to call `vjs.trigger(el, 'dispose')` on it to clean up
+   * references to it and allow the browser to garbage collect it.
+   *
+   * @param  {String|vjs.Component} first   The event type or other component
+   * @param  {Function|String}      second  The event handler or event type
+   * @param  {Function}             third   The event handler
+   * @return {vjs.Component}        self
+   */
+  on(first, second, third){
```
gkatsev: it would be really nice to name these better names

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
+  /**
+   * Bind a listener to the component's ready state
+   *
+   * Different from event listeners in that if the ready event has already happened
+   * it will trigger the function immediately.
+   *
+   * @param  {Function} fn Ready listener
+   * @return {vjs.Component}
+   */
+  ready(fn){
```
gkatsev: I had the thing below, but now I realize it isn't quite what we want, since we do only want to add the function to the queue only if it was given to us.
```js
class Component {
  ready (fn = function() {}) {
    if (this.isReady_) {
      fn.call(this);
    } else {
      this.readyQueue_ = (this.readyQueue_ || []);
      this.readyQueue_.push(fn);
    }
}
```

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
+   *
+   * @param  {String} widthOrHeight  'width' or 'height'
+   * @param  {Number|String=} num     New dimension
+   * @param  {Boolean=} skipListeners Skip resize event trigger
+   * @return {vjs.Component} The component if a dimension was set
+   * @return {Number|String} The dimension if nothing was set
+   * @private
+   */
+  dimension(widthOrHeight, num, skipListeners){
+    if (num !== undefined) {
```
gkatsev: `num` can be set to a default argument of `0`, then we only need to check if it's `null` or `NaN`.

src/js/component.js
```diff
@@ -82,1182 +90,1103 @@ var Component = CoreObject.extend({
...
-  this.addClass('vjs-hidden');
-  return this;
-};
+      // Check if using css width/height (% or px) and adjust
+      if ((''+num).indexOf('%') !== -1 || (''+num).indexOf('px') !== -1) {
+        this.el_.style[widthOrHeight] = num;
+      } else if (num === 'auto') {
+        this.el_.style[widthOrHeight] = '';
+      } else {
+        this.el_.style[widthOrHeight] = num+'px';
```
gkatsev: this should *not* be turned into a template string.

src/js/control-bar/current-time-display.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    el.appendChild(this.contentEl_);
+    return el;
+  }
+
+  updateContent() {
+    // Allows for smooth scrubbing, when player can't keep up.
+    let time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
+    this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Current Time') + '</span> ' + Lib.formatTime(time, this.player_.duration());
```
gkatsev: this will probably be fun to rebase with the template strings PR

src/js/control-bar/mute-toggle.js
gkatsev: I'm actually really liking this look for this particular usecase:
```js
level = (vol === 0)  ? 0 :
        (vol < 0.33) ? 1 :
        (vol < 0.67) ? 2 ;
```

src/js/control-bar/mute-toggle.js
```diff
@@ -29,50 +29,51 @@ var MuteToggle = Button.extend({
...
-      }
-  } else {
-      if(this.el_.children[0].children[0].innerHTML!=this.localize('Mute')){
-          this.el_.children[0].children[0].innerHTML = this.localize('Mute'); // change the button text to "Mute"
-      }
-  }
+    // Don't rewrite the button text if the actual text doesn't change.
+    // This causes unnecessary and confusing information for screen reader users.
+    // This check is needed because this function gets called every time the volume level is changed.
+    if(this.player_.muted()){
```
gkatsev:
```js
let toMute = this.player_.muted() ? 'Unmute' : 'Mute';
let localizedMute = this.localize(toMute);
if (this.el_.children[0].children[0].innerHTML !== localizedMute) {
  this.el_.children[0].children[0].innerHTML = localizedMute;
}
```

src/js/component.js
mmcc: [it's not](https://github.com/mmcc/video.js/blob/es6-defaults-interpolation/src/js/component.js#L981)

src/js/tracks/text-track.js
```diff
@@ -227,7 +226,7 @@ TextTrack.prototype.removeCue = function(removeCue) {
 /*
  * Downloading stuff happens below this point
  */
-let parseCues = function(srcContent, track) {
+var parseCues = function(srcContent, track) {
   if (typeof window['WebVTT'] !== 'function') {
```
gkatsev: why was this changed? We should move `parseCues` up in the file, if the TDZ is the issue.

src/js/big-play-button.js
mmcc: Looks like it's [enabled by default now](https://github.com/atom/atom/issues/399), which would be weird if you're not seeing it.

src/js/component.js
gkatsev: Yeah, not saying that it's an issue. Just that we should be careful around things like these because `num` is user inputtable and if it's carefully crafted, could cause issues with template strings since they have access to the variables.

mmcc: Ah good point. Noted for future reference.

heff:
I guess the punishment for having a crappy diff and forcing people to look at every line of code is you get to address every style problem from ever :) There's at least fewer than I'd expect.

src/js/component.js
heff: It made sense when it was vars. The if block used children which was defined with parentOptions which was defined with parent. I didn't actually touch this part but I've cleaned it up for you. :)

src/js/component.js
heff: The official Closure Library style is to use the private references everywhere internally so compiler can mangle them, and then export the methods for anything you want accessible externally. That's not always obvious for other contributors, and I also started to worry that people would start using the private properties externally after seeing internal code, but then on the other hand I think uglify has an option to mangle private props now if we wanted to keep that part of the file size savings.

Either way it's open for a change.

src/js/control-bar/mute-toggle.js
heff: That last one does look really clean. I think it depends if we want to be more strict in our linting about ternaries to avoid things like in the other PR.

(I'm not going to try to fix this one in this PR)

src/js/control-bar/progress-control/seek-bar.js
heff: Yeah, setting them on the prototype makes them globally overridable, which I assume is handy at the player level for setting global options across players, but less useful in other components.

I've started to think the same thing about moving the options to the constructor, and also potentially killing the `options()` accessor. It kind of creates two sources of truth. Not sure how much people are relying on that now though.

src/js/tech/html5.js
heff: I'm sad you gave up on the super images. :( In my email they got progressively larger as I scrolled down. :-P

src/js/component.js
heff: Yeah, I can't pass `this` (the player) as the first arg to the component constructor or I get yelled at for using `this` before the super().

We could default to this, but I also think it could be good to throw an error if there's no player arg when other components are created. If a component depends on the player API, it will throw random errors later.

src/js/component.js
heff: Yeah, definitely.

src/js/component.js
heff: By default they point to the same, but some elements use sub elements to create their appearance and you don't want the the children to go into the main element anymore. MenuButtons might be one of those but I'm not remembering exactly.

src/js/component.js
heff: You see why it's difficult though, yeah?

src/js/component.js
heff: Cool. I added the other cleanup item.

src/js/component.js
heff: But then we wouldn't know if dimension is being used as a getter.

src/js/control-bar/mute-toggle.js
heff: Much nicer. Updated.

src/js/tracks/text-track.js
heff: Yeah, can't use a let var before it's defined in the file. How about I make it `function parseCues()`? I don't like burying the constructor in its own file.

src/js/component.js
gkatsev: yes. What about: `first -> typeOrComponent`, `second -> `handlerOrType`, and `third -> handler`.

src/js/control-bar/mute-toggle.js
gkatsev: Ternaries are very useful sometimes, like here. They shouldn't definitely not be overused, but they shouldn't be outright discounted.
If we feel like the style here is more readable, we should go with it even if our current linting configuration may complain.

heff:
Updated with the super calls and cleanup changes mentioned. I've moved some not directly related discussions to their own issues. Otherwise the important bits here are done so going to merge this in so we can move forward.

PR #1997, opened 2015-04-01
Add playing event to docs.  Update play and playing event descriptions.

bclwhitaker:
It useful to know that there is an event that will fire when media is actually playing.  Since this is [required to be implemented by all techs per:](https://github.com/videojs/video.js/blob/3352346a74eca52328c14cad803e21587886bfbe/docs/guides/tech.md#required-events) I think its safe to include it here.

### DIFF
--- a/docs/api/vjs.Player.md
+++ b/docs/api/vjs.Player.md
@@ -109,6 +109,7 @@ After an instance has been created it can be accessed globally using `Video('exa
   - [loadstart](#loadstart-event)
   - [pause](#pause-event)
   - [play](#play-event)
+  - [playing](#playing)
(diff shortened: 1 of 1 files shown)


heff:
Thanks Lee! These docs are actually generated from the source, so the update would need to be made there. https://github.com/videojs/video.js/blob/6cc1b5191bf3db67ffd302ba5931d6c1c49461ec/src/js/player.js#L436

heff:
:pineapple:

heff:
@bclwhitaker do you want to try this again by adding it to the player.js file?

gkatsev:
Out of date.

PR #1999, opened 2015-04-02
New Base Theme

mmcc:
Opening the PR so we can start talking about it, but this needs a little more work before it can be pulled in. We should talk about things like:

1. Do we really want to do style injection by default. I personally love it, but there is a slightly annoying Flash of Unstyled Video Player™. 
2. Some things have changed in the VJS world since I started this, including the separator. That should be removed from this PR in lieu of #1838.
3. Any stylistic changes we want to make. I promise I'll only cry into my pillow a little if you have mean things to say about it.

I'm almost certain I blew away some necessary stuff in the Gruntfile when I was rebasing from master, but I didn't want to dig through the various tasks and mess with it. There are probably css-specific things from the build task specifically that got blown away. Sorry. :sweat:

Lastly, this needs to be tested a *ton* across all the browsers and devices. Seriously. I'm using `flexbox` with a `table` fallback, so it's not like anything could *possibly* go wrong :trollface:.

For working examples built on top of the skin:
- [Colors](http://mmcc.github.io/videojs-skin-colors/) - Simple color themes on top of the base styles
- [Twitchy](http://mmcc.github.io/videojs-skin-twitchy/) - Quickly hacked together in an afternoon to try and mimic the Twitch player


src/js/control-bar/time-display.js
```diff
@@ -94,7 +94,7 @@ vjs.TimeDivider = vjs.Component.extend({
 
 vjs.TimeDivider.prototype.createEl = function(){
   return vjs.Component.prototype.createEl.call(this, 'div', {
-    className: 'vjs-time-divider',
+    className: 'vjs-time-controls vjs-time-divider',
```
heff: `vis-time-controls` should be singular, yeah?

heff:
Another giant 5.0 PR, good luck everybody! :) Good stuff Matt.

For everyone, the commits has a decent list of the overall changes.

@mmcc, can you talk about some of the motivations behind all of this? I'm thinking specifically about breaking up the files, the variables file, and anything else related.

Also since every line of CSS was moved, it's hard to tell what styles changed. What can we do to start a skin migration guide?

.travis.yml
```diff
@@ -1,8 +1,10 @@
 language: node_js
 node_js:
-- 0.10
+  - 0.10
+before_install:
+  - gem install sass
```
heff: Is `gem install sass` a dependency for everyone or just for Travis?

gkatsev: Can we use `libsass` instead? I *really* don't want to need to have to deal with gems.

Gruntfile.js
```diff
@@ -97,11 +97,17 @@ module.exports = function(grunt) {
       tasks: 'dev'
     },
     connect: {
-      dev: {
+      preview: {
```
heff: `connect:preview` doesn't look like it's used anywhere else. What's it for?

src/css/_utilities.scss
```diff
@@ -0,0 +1,86 @@
...
+}
+
+@mixin animation($string: spin 1s infinite linear) {
+  -webkit-animation: $string;
+  -moz-animation: $string;
+  -o-animation: $string;
+  animation: $string;
+}
+
+@mixin context($old-context, $new-context) {
```
heff: I smell dark magic. Also the only use of this is commented out.

mmcc: Yeah, in fact I think you and I laughed about this dark magic. I think this was basically a way to prepend a class from within a mixin, but yes, this can go away.

.travis.yml
mmcc: Right now this PR is using grunt-contrib-sass, which uses [Ruby sass](https://github.com/gruntjs/grunt-contrib-sass/issues/38). However, we can try using [grunt-sass](https://github.com/sindresorhus/grunt-sass) instead.

gkatsev: Whatever doesn't require me to ever touch `gem`.

Gruntfile.js
mmcc: `grunt connect:preview` is the old way of doing things. `grunt connect:dev` is used in conjunction with watch so you can have a server and still get new sass hotness on saves.

.travis.yml
mmcc: :nail_care:

Gruntfile.js
heff: Is there a way we can get all the watch tasks and the server into one
command?

-- 
Sent from mobile

src/css/_variables.scss
```diff
@@ -0,0 +1,11 @@
+$primary-text: #fff;
+$secondary-text: #F4A460; // currently just used for visited links in errors and such.
+
+$primary-bg: #000;
```
gkatsev: we should do something like:
```sass
// colors.scss
$black: #000

//vars.scss
$primary-bg: $black

// other files
background-color: $primary-bg
```

that way, if we feel like changing our "black" to be `#3c3c3c` or something, we only change it in one place. The primary/stuff is already good the way it is :)

Gruntfile.js
mmcc: Yes, I just did it this way to preserve the old way so I wouldn't surprise people. We can definitely just combine everything.

ange007:
Colors skin Error
Linux Mint 17, Chrome 42.0.2311.90

src/css/_utilities.scss
```diff
@@ -0,0 +1,80 @@
...
+  -ms-flex: $value;
+  flex: $value;
+}
+
+@mixin order($value) {
+  -webkit-box-ordinal-group: $value;
+  -moz-box-ordinal-group: $value;
+  -ms-flex-order: $value;
+  -webkit-order: $value;
+  order: $value;
```
heff: Why is order red here?

mmcc: ¯\_(ツ)_/¯

src/css/components/_progress.scss
```diff
@@ -0,0 +1,52 @@
...
+  background: rgba(255, 255, 255, 0.2);
+}
+
+/* there are child elements of the load progress bar that represent the
+specific time ranges that have been buffered */
+.video-js .vjs-load-progress div {
+  background: rgba($secondary-bg, 0.1);
+}
+
+.video-js .vjs-slider-handle.vjs-seek-handle {
```
heff: Could we reorder the includes so this doesn't need the double class? It doesn't look like slider-handle currently sets these values anyway, but it probably should so you can use a default slider.

src/css/components/_control-bar.scss
```diff
@@ -0,0 +1,52 @@
...
+  visibility: visible;
+  opacity: 1;
+
+  $trans: visibility 0.1s, opacity 0.1s; // Var needed because of comma
+  @include transition($trans);
+}
+
+// Video has started playing AND user is inactive
+.video-js.vjs-has-started.vjs-user-inactive.vjs-playing .vjs-control-bar {
+  @include display-flex;
```
heff: I don't think `@include display-flex` is actually needed here again.

src/css/video-js.less
```diff
@@ -1,1237 +0,0 @@
...
-  display: none;
-}
-
-/* The control bar shouldn't show after an error */
-.vjs-default-skin.vjs-error .vjs-control-bar {
-  display: none;
-}
-
-/* Don't hide the control bar if it's audio */
-.vjs-audio.vjs-default-skin.vjs-has-started.vjs-user-inactive.vjs-playing .vjs-control-bar {
```
heff: These vjs-audio settings aren't added back anywhere.

src/css/components/_progress.scss
```diff
@@ -0,0 +1,52 @@
+.video-js .vjs-progress-control {
+  @include flex(auto);
+  @include display-flex(center);
+
+  @include transition(all 0.4s);
```
heff: If the progress bar isn't growing and shrinking anymore then this isn't needed.

src/css/components/_live.scss
```diff
@@ -0,0 +1,11 @@
+.video-js.vjs-live .vjs-time-control,
+.video-js.vjs-live .vjs-time-divider,
+.video-js.vjs-live .vjs-progress-control {
+  display: none;
+}
+
+.video-js .vjs-live-control, .video-js.vjs-no-flex .vjs-live-control {
```
heff: Is the `vjs-no-flex` duplication needed here?

heff:
All the `vjs-caption-settings` styles need to make it back into here

heff:
Here's the rest of the notes I took on changes, to be added to the skin migration guide.

- Added icon classes to be used externally e.g. vjs-icon-circle
- Removed the vis-default-skin class. New skins should just override the defaults. Most people have done that anyway.
- Dropped some IE6-specific fixes
- Switched the control bar to use flex-box in browsers that support it. Using table styles in older browsers.

### These probably aren't finished/correct

\* Dropped user-select from video.js.
\* SVG font is missing
\* Removed slider focus styles -- need to add better focus styles for all controls, specifically for accessibility

mmcc:
A few notes from Steve's initial comment that I never answered before.

>  can you talk about some of the motivations behind all of this? I'm thinking specifically about breaking up the files, the variables file, and anything else related.

At the most basic level, I've personally wanted to switch to Sass for a while. A lot of the original decision was motivated by being able to use Less in the browser, which is irrelevant these days. This skin is simpler in some ways, such as being able to let the handle overlap some of the progress bar to avoid the old issues with the diamond.

The new flexbox / table layout gives us the flexibility we had before in terms of being able to add new items to the control bar, without needing too much hackery (besides the whole table thing).

Re: breaking up the files, the old LESS file was *massive*, which meant doing anything at all in the file was a pain. `cmd+f`to find what you want, go get something else, remember approximately where you were and scroll back to it...just general pain. Now we have things broken up largely along the lines of each component having its own file, so you can quickly pop open the exact file you need to make a change to the progress bar or volume control.

Particularly with the ES6 / Browserify changes, I think the Sass updates just keep the skin in line with the rest of the project.

mmcc:
@ange007 your issue is fixed with 1c316

heff:
We're gonna party like it's this issue.

tumblr_mvg8c824z81qcvaxho6_250

hartman:
@mmcc I'm interested in the adaptive layout stuff (vjs-layout-tiny and friends) that were added with this. What was the though behind that ? It's not yet in use right ?

mmcc:
@hartman So the thought was we'd use JS to determine the width of the player on resize and add these classes programmatically. We decided not to do that (yet? maybe?), but I left them in for folks that wanted to use them on their own.

Anyway, the thought behind it was that VJS kinda sucked at smaller sizes. Once it got down to a certain point things started falling off the control bar and generally looked bad. Most of those problems should be fixed with the new layout since it resizes better, but at smaller sizes it still looks waaaay too busy imho. So, ultimately, this was an effort to make smaller players more usable.

Thoughts? Would you like to see this added programmatically or is it enough just to have the classes there?

hartman:
@mmcc Experimented a bit with doing this programatically. Took a lot more than I anticipated, though the code can be cleaned up a bit more. But, it works: http://jsfiddle.net/TheDJ/5raorqgy/

PR #2015, opened 2015-04-07
ES6 Default args and interpolation

mmcc:
As well as a some fat arrows thrown in for good measure. Feel free to shoot anything in here you don't like down and we can revert.

### DIFF
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -56,7 +56,7 @@ class Button extends Component {
 
   buildCSSClass() {
     // TODO: Change vjs-control to vjs-button?
-    return 'vjs-control ' + super.buildCSSClass();
+    return `vjs-control ${super.buildCSSClass()}`;
   }
 
   // Click - Override with specific functionality for button
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -7,7 +7,7 @@ import document from 'global/document';
 ================================================================================ */
 /**
  * Base class for all buttons
- * @param {vjs.Player|Object} player
+ * @param {Player|Object} player
  * @param {Object=} options
  * @class
  * @constructor
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -35,7 +35,6 @@ import window from 'global/window';
  * @param {Object=} options
  * @class
  * @constructor
- * @extends vjs.CoreObject
  */
 class Component {
 
@@ -131,7 +130,7 @@ class Component {
   /**
    * Return the component's player
    *
-   * @return {vjs.Player}
+   * @return {Player}
    */
   player() {
     return this.player_;
@@ -141,7 +140,7 @@ class Component {
    * Deep merge of options objects
    *
    * Whenever a property is an object on both options objects
-   * the two properties will be merged using vjs.obj.deepMerge.
+   * the two properties will be merged using Lib.obj.deepMerge.
    *
    * This is used for merging options for child components. We
    * want it to be easy to override individual options on a child
@@ -261,7 +260,7 @@ class Component {
   /**
    * Returns a child component with the provided ID
    *
-   * @return {vjs.Component}
+   * @return {Component}
    */
   getChildById(id){
     return this.childIndex_[id];
@@ -270,7 +269,7 @@ class Component {
   /**
    * Returns a child component with the provided name
    *
-   * @return {vjs.Component}
+   * @return {Component}
    */
   getChild(name){
     return this.childNameIndex_[name];
@@ -299,9 +298,9 @@ class Component {
    *       }
    *     });
    *
-   * @param {String|vjs.Component} child The class name or instance of a child to add
+   * @param {String|Component} child The class name or instance of a child to add
    * @param {Object=} options Options, including options to be passed to children of the child.
-   * @return {vjs.Component} The child component (created by this process if a string was used)
+   * @return {Component} The child component (created by this process if a string was used)
    * @suppress {accessControls|checkRegExp|checkTypes|checkVars|const|constantProperty|deprecated|duplicate|es5Strict|fileoverviewTags|globalThis|invalidCasts|missingProperties|nonStandardJsDocs|strictModuleDepCheck|undefinedNames|undefinedVars|unknownDefines|uselessCode|visibility}
    */
   addChild(child, options){
@@ -326,8 +325,6 @@ class Component {
 
       // Create a new object & element for this controls set
       // If there's no .player_, this is a player
-      // Closure Compiler throws an 'incomplete alias' warning if we use the vjs variable directly.
-      // Every class should be exported, so this should never be a problem here.
       let componentClass = Component.getComponent(componentClassName);
 
       component = new componentClass(this.player_ || this, options);
@@ -365,7 +362,7 @@ class Component {
    * Remove a child component from this component's list of children, and the
    * child component's element from this component's element
    *
-   * @param  {vjs.Component} component Component to remove
+   * @param  {Component} component Component to remove
    */
   removeChild(component){
     if (typeof component === 'string') {
@@ -443,7 +440,7 @@ class Component {
         }
 
         // Allow for disabling default components
-        // e.g. vjs.options['children']['posterImage'] = false
+        // e.g. options['children']['posterImage'] = false
         if (opts === false) return;
 
         // Create and add the child component.
@@ -505,20 +502,20 @@ class Component {
    *     myComponent.on(otherElement, 'eventName', myFunc);
    *     myComponent.on(otherComponent, 'eventName', myFunc);
    *
-   * The benefit of using this over `vjs.on(otherElement, 'eventName', myFunc)`
+   * The benefit of using this over `VjsEvents.on(otherElement, 'eventName', myFunc)`
    * and `otherComponent.on('eventName', myFunc)` is that this way the listeners
    * will be automatically cleaned up when either component is disposed.
    * It will also bind myComponent as the context of myFunc.
    *
    * **NOTE**: When using this on elements in the page other than window
    * and document (both permanent), if you remove the element from the DOM
-   * you need to call `vjs.trigger(el, 'dispose')` on it to clean up
+   * you need to call `myComponent.trigger(el, 'dispose')` on it to clean up
    * references to it and allow the browser to garbage collect it.
    *
-   * @param  {String|vjs.Component} first   The event type or other component
+   * @param  {String|Component} first   The event type or other component
    * @param  {Function|String}      second  The event handler or event type
    * @param  {Function}             third   The event handler
-   * @return {vjs.Component}        self
+   * @return {Component}        self
    */
   on(first, second, third){
     var target, type, fn, removeOnDispose, cleanRemover, thisComponent;
@@ -558,7 +555,7 @@ class Component {
         Events.on(target, 'dispose', cleanRemover);
 
       // Should be a component
-      // Not using `instanceof vjs.Component` because it makes mock players difficult
+      // Not using `instanceof Component` because it makes mock players difficult
       } else if (typeof first.on === 'function') {
         // Add the listener to the other component
         target.on(type, fn);
@@ -584,10 +581,10 @@ class Component {
    *     myComponent.off(otherElement, 'eventType', myFunc);
    *     myComponent.off(otherComponent, 'eventType', myFunc);
    *
-   * @param  {String=|vjs.Component}  first  The event type or other component
+   * @param  {String=|Component}  first  The event type or other component
    * @param  {Function=|String}       second The listener function or event type
    * @param  {Function=}              third  The listener for other component
-   * @return {vjs.Component}
+   * @return {Component}
    */
   off(first, second, third){
     var target, otherComponent, type, fn, otherEl;
@@ -629,10 +626,10 @@ class Component {
    *     myComponent.one(otherElement, 'eventName', myFunc);
    *     myComponent.one(otherComponent, 'eventName', myFunc);
    *
-   * @param  {String|vjs.Component}  first   The event type or other component
+   * @param  {String|Component}  first   The event type or other component
    * @param  {Function|String}       second  The listener function or event type
    * @param  {Function=}             third   The listener function for other component
-   * @return {vjs.Component}
+   * @return {Component}
    */
   one(first, second, third) {
     var target, type, fn, thisComponent, newFunc;
@@ -665,7 +662,7 @@ class Component {
    *     myComponent.trigger({'type':'eventName'});
    *
    * @param  {Event|Object|String} event  A string (the type) or an event object with a type attribute
-   * @return {vjs.Component}       self
+   * @return {Component}       self
    */
   trigger(event){
     Events.trigger(this.el_, event);
@@ -679,7 +676,7 @@ class Component {
    * it will trigger the function immediately.
    *
    * @param  {Function} fn Ready listener
-   * @return {vjs.Component}
+   * @return {Component}
    */
   ready(fn){
     if (fn) {
@@ -696,7 +693,7 @@ class Component {
   /**
    * Trigger the ready listeners
    *
-   * @return {vjs.Component}
+   * @return {Component}
    */
   triggerReady(){
     this.isReady_ = true;
@@ -721,7 +718,7 @@ class Component {
    * Check if a component's element has a CSS class name
    *
    * @param {String} classToCheck Classname to check
-   * @return {vjs.Component}
+   * @return {Component}
    */
   hasClass(classToCheck){
     return Lib.hasClass(this.el_, classToCheck);
@@ -731,7 +728,7 @@ class Component {
    * Add a CSS class name to the component's element
    *
    * @param {String} classToAdd Classname to add
-   * @return {vjs.Component}
+   * @return {Component}
    */
   addClass(classToAdd){
     Lib.addClass(this.el_, classToAdd);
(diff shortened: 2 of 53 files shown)


gkatsev:
[Something interesting you can do with destructuring](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=true&spec=false&playground=true&code=var getFileExtension %3D %28src%29 %3D> 'video%2Fmp4'%0Afunction src%28{ src %3D arguments[0]%2C type %3D getFileExtension%28arguments[0]%29 }%29 {%0A  console.log%28src%2C type%29%3B%0A}%0A%0Asrc%28'foo'%29%3B%0A%0Asrc%28{%0A  src%3A 'foo'%2C%0A  type%3A 'bar'%0A}%29):
```js
var getFileExtension = (src) => 'video/mp4' // or the real implementation

function src({ src = arguments[0], type = getFileExtension(arguments[0]) }) {
  console.log(src, type);
}

src('foo');
// => foo video/mp4

src({
  src: 'foo',
  type: 'bar'
})
```
// => foo bar

mmcc:
@gkatsev so I really like this, but fwiw I don't think this cleans up too much in the actual code base since this is spread across the techs. So far I've done [this](https://github.com/videojs/video.js/pull/2015/files#diff-3b0266ff1c037b289ec624ab25b0272eR1171), but that seems to be about as generalized as we can get.

src/js/tracks/text-track-settings.js
```diff
@@ -165,119 +165,123 @@ function setSelectedOption(target, value) {
   target.selectedIndex = i;
 }
 
+// I know what you're thinking: "This should be a tagged template!". I thought the same thing
+// when I first took a look. However,
```
gkatsev: However?

gkatsev:
Yeah, it's a fun experiment. Not sure how terribly applicable it is to our uses.

src/js/tracks/text-track-settings.js
mmcc: It's for suspense.

mmcc: Kidding, I think I decided to confirm that I was right and never went back and finished the comment.

src/js/component.js
```diff
@@ -360,21 +362,19 @@ Component.prototype.getChild = function(name){
...
   let component, componentName;
   // If child is a string, create nt with options
   if (typeof child === 'string') {
     let componentName = child;
 
-    // Make sure options is at least an empty object to protect against errors
-    if (!options || options === true) {
-      options = {};
-    }
+    // Apparently sometimes options is just true, so we need that to make sure we still convert options to an object.
```
heff: "Apparently" :-/

Yeah, you can do:
```
{
  children: {
    controlBar: true,
    posterImage: false
  }
}
```

The `false` bit is useful. I think the true might have just been for symmetry. Now that you should be defining children with arrays I think we can ditch the `true` option (but not `false`).

src/js/button.js
```diff
@@ -62,7 +62,7 @@ Button.prototype.createEl = function(type, props){
 
 Button.prototype.buildCSSClass = function(){
   // TODO: Change vjs-control to vjs-button?
-  return 'vjs-control ' + Component.prototype.buildCSSClass.call(this);
+  return `vjs-control ${Component.prototype.buildCSSClass.call(this)}`;
```
heff: super() will clean this up nicely

src/js/component.js
```diff
@@ -59,7 +59,7 @@ var Component = CoreObject.extend({
     // If there was no ID from the options, generate one
     if (!this.id_) {
       // Don't require the player ID function in the case of mock players
-      this.id_ = ((player.id && player.id()) || 'no_player') + '_component_' + Lib.guid++;
+      this.id_ = `${(player.id && player.id()) || 'no_player'}_component_${Lib.guid++}`;
```
heff: This one doesn't feel like an improvement

gkatsev:
```js
let id = player.id && player.id() || 'no_player;
this.id_ = `${id}_component_${Lib.guid++}`;
```

src/js/component.js
```diff
@@ -692,23 +688,21 @@ Component.prototype.off = function(first, second, third){
...
-    type = second;
-    fn = Lib.bind(this, third);
-    thisComponent = this;
+    let target = first;
+    let type = second;
+    let fn = Lib.bind(this, third);
+    let thisComponent = this;
 
-    newFunc = function(){
+    let newFunc = ()=> {
```
heff: I feel like the parens should have a space after them. Is that not right?

Also, this adds nothing except an unnecessary bind and syntax that the average contributor won't understand yet.

mmcc: heh this is actually exactly what I assumed I would have to revert.

I'm not sorry.

src/js/control-bar/mute-toggle.js
```diff
@@ -34,7 +34,7 @@ var MuteToggle = Button.extend({
 MuteToggle.prototype.createEl = function(){
   return Button.prototype.createEl.call(this, 'div', {
     className: 'vjs-mute-control vjs-control',
-    innerHTML: '<div><span class="vjs-control-text">' + this.localize('Mute') + '</span></div>'
+    innerHTML: `<div><span class="vjs-control-text">${this.localize('Mute')}</span></div>`
```
heff: See [this comment](https://github.com/videojs/video.js/pull/1976/files#r27762603)

src/js/control-bar/playback-rate-menu-button.js
```diff
@@ -98,7 +98,7 @@ PlaybackRateMenuButton.prototype.updateVisibility = function(){
  */
 PlaybackRateMenuButton.prototype.updateLabel = function(){
   if (this.playbackRateSupported()) {
-    this.labelEl_.innerHTML = this.player().playbackRate() + 'x';
+    this.labelEl_.innerHTML = `${this.player().playbackRate()}x`;
```
heff: I think this one got less readable.

src/js/component.js
gkatsev: idk, fat arrows are one of the biggest changes available.
We should just make sure we have prominent links to where people can find out more.

src/js/control-bar/progress-control.js
```diff
@@ -150,7 +150,7 @@ LoadProgressBar.prototype.update = function(){
   // get the percent width of a time compared to the total end
   let percentify = function (time, end){
     let percent = (time / end) || 0; // no NaN
-    return (percent * 100) + '%';
+    return `${percent * 100}%`;
```
heff: uhy

src/js/control-bar/time-display.js
```diff
@@ -36,12 +36,12 @@ CurrentTimeDisplay.prototype.createEl = function(){
 CurrentTimeDisplay.prototype.updateContent = function(){
   // Allows for smooth scrubbing, when player can't keep up.
   let time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
-  this.contentEl_.innerHTML = '<span class="vjs-control-text">' + this.localize('Current Time') + '</span> ' + Lib.formatTime(time, this.player_.duration());
+  this.contentEl_.innerHTML = `<span class="vjs-control-text">${this.localize('Current Time')}</span>${Lib.formatTime(time, this.player_.duration())}`;
```
heff: The `Lib.formatTime` feels like too much to be doing inside the string now. Should we pull the result of that into a var first?

src/js/control-bar/playback-rate-menu-button.js
mmcc: I went back and forth on this one, but I think we should err on the side of consistency. Once we get used to expecting string interpolation, my hope is that this will look normal.

src/js/control-bar/progress-control.js
mmcc: Consistency! Consistency I say!

src/js/events.js
```diff
@@ -330,6 +330,7 @@ var trigger = function(elem, event) {
    * for now just in case.
    */
   // // Added in addition to book. Book code was broke.
+  // TODO: figure out what these events are
```
heff: We can remove this commented out section.

src/js/lib.js
```diff
@@ -304,7 +301,7 @@ var isEmpty = function(obj) {
  * @private
  */
 var hasClass = function(element, classToCheck){
-  return ((' ' + element.className + ' ').indexOf(' ' + classToCheck + ' ') !== -1);
+  return ((` ${element.className} `).indexOf(` ${classToCheck} `) !== -1);
```
heff: The spaces here have a specific purpose that I think gets blurred when you add them in a template like this.

src/js/control-bar/time-display.js
gkatsev:
```js
let time = (this.player_.scrubbing) ? this.player_.getCache().currentTime : this.player_.currentTime();
let formattedTime = Lib.formatTime(time, this.player_.duration());
let localizedText = this.localize('Current Time');
this.contentEl_.innerHTML = `<span class="vjs-control-text">${localizedText}</span>${formattedTime}`;
```

src/js/lib.js
gkatsev: [template strings keep spacing.](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=true&spec=false&playground=true&code=let foo %3D 5%3B%0A%0Aconsole.log%28` %24{foo} `%29%3B)

src/js/lib.js
```diff
@@ -725,7 +719,7 @@ function _logType(type, args){
 
   if (type) {
     // add the type to the front of the message
-    argsArray.unshift(type.toUpperCase()+':');
+    argsArray.unshift(`${type.toUpperCase()}:`);
```
heff: Any of these where we're specifically adding on a single important character, I appreciate the verboseness of the old syntax.

src/js/lib.js
heff: I know they keep spacing but I don't it's as clear that we're adding spacing for a reason when you read it in template syntax.

src/js/player.js
```diff
@@ -1168,13 +1168,9 @@ Player.prototype.selectSource = function(sources){
  * @return {String} The current video source when getting
  * @return {String} The player when setting
  */
-Player.prototype.src = function(source){
+Player.prototype.src = function(source=this.techGet('src')){
```
heff: You can do this?? And it will execute each time the function is called, not just when the function is created?

src/js/lib.js
gkatsev: Add documentation for the method?

src/js/player.js
gkatsev: Yes!
Also, add spaces around the `=`.

src/js/tracks/text-track-settings.js
```diff
@@ -165,119 +165,123 @@ function setSelectedOption(target, value) {
...
-            '<option value="small-caps">Small Caps</option>' +
-          '</select>' +
-        '</div>' + // vjs-font-family
-      '</div>' +
-    '</div>' +
-    '<div class="vjs-tracksettings-controls">' +
-      '<button class="vjs-default-button">Defaults</button>' +
-      '<button class="vjs-done-button">Done</button>' +
-    '</div>';
+  let template = `<div class="vjs-tracksettings">
```
heff: It was all worth it for this. :)

src/js/player.js
heff: Cooooool

src/js/lib.js
mmcc: With the template syntax you can clearly see there's a space inside of the template, so I'm not sure I follow why one infers reasoning behind that space more than the other.

src/js/tracks/text-track-settings.js
mmcc: This was so wildly satisfying.

src/js/player.js
mmcc: We disagree here, I don't like spaces in default args for some reason.

But I realize I'm probably being weird there, so I'll defer.

src/js/lib.js
mmcc: And I like Gary's idea anyway...This is now going to be exported anyway, so a little documentation wouldn't hurt.

src/js/component.js
heff: Yeah, that would be cleaner.

src/js/component.js
heff: Yeah, I'm not saying all fat arrows are bad. I'm saying here I think it's overkill.

src/js/player.js
heff: There's elements of both that I like so I don't have a strong opinion here, but for other default args in this PR it has no spaces. We should be consistent either way.

src/js/control-bar/playback-rate-menu-button.js
heff: I don't think string templates were meant to replace every instance of string concatenation everywhere. They work beautifully in the examples like `some text ${var} some text`. i.e. "I need to drop a variable in the middle of and existing string of text". And multi-line strings.

Other cases are "I need to append text to a an existing value". `width + 'px'`. That's not 'modifying a template', that's creating a value.

Maybe put more succinctly, in one you're adding a value to a string, in the other you're adding a string to a value. So we could use that for the rule around consistency.

In this specific example, `${this.player().playbackRate()}x`, the most important thing I need to see there is that the `x` is getting appended to the value, but visually that gets lost in the junkiness of curly braces and parens. Whereas with `this.player().playbackRate() + 'x'`, it's the first thing I see.

mmcc: Fair enough, makes sense to me.

mmcc:
@heff I think this one's good to go unless you want to take another pass before one of us disappears into the rebase pit

src/js/component.js
```diff
@@ -1072,16 +1069,16 @@ Component.prototype.emitTapEvents = function(){
...
+      let ydiff = event.touches[0].pageY - firstTouch.pageY;
+      let touchDistance = Math.sqrt(xdiff * xdiff + ydiff * ydiff);
       if (touchDistance > tapMovementThreshold) {
         couldBeTap = false;
       }
     }
   });
 
-  noTap = function(){
+  let noTap = ()=> {
```
heff: I don't know how I missed these, but they're the same as the other example we reverted. I looked at how babel translates arrow functions and it at least doesn't do an extra `fn.bind()`, so that's good. I want to make sure I'm not just being a luddite here so if there's fault in my reasoning let me know. I think it's an important issue.

I still don't think the shortened syntax alone is benefit enough to warrant upping the javascript reading level for contributors. For most contributors JavaScript is probably still not their primary programming lang, and we'd be replacing one of the most most well known keywords with one of the least known that does the exact same thing (in this specific case).

Also, Babel makes named functions when translating arrow functions.
```
var noTap = function noTap() {}
```
That goes against the "Arrow functions are always anonymous" rule and could potentially [create issues in IE8](http://kiro.me/blog/nfe_dilemma.html). Though I'm not sure how likely that is and I wish we could use named functions more for stack traces.

Also if you have a text editor with a function shortcut, there's not much benefit left over for this specific case. `f` then `tab` quickly in Atom makes an empty function.

I do think there are some cool examples of using arrow functions, like:
```
let square = x => x * x;
```
I would have a harder time arguing against that, and it even seems more readable than the example above for some reason.

And using it for it's `this` binding.
```
var boundFn = ()=> {
  this.foo();
}
```

But in all the examples here we seem to just be using it for it's shorter anonymous function syntax, which if that's correct then for consistency there's over 500 other anonymous functions in the codebase we would need to update too. Am I missing another benefit in these examples?

src/js/component.js
```diff
@@ -568,36 +572,34 @@ Component.prototype.buildCSSClass = function(){
...
-    fn = Lib.bind(this, third);
-    thisComponent = this;
+    let target = first;
+    let type = second;
+    let fn = Lib.bind(this, third);
+    let thisComponent = this;
 
     // When this component is disposed, remove the listener from the other component
-    removeOnDispose = function(){
+    let removeOnDispose = ()=> {
```
heff: Here we could be using the arrow function it to replace `thisComponent`, assuming the IE8 issue mentioned above doesn't trump arrow functions all together. @dmlap, @gkatsev any thoughts on the IE8 named function issue? I've usually avoided using named functions when it's not needed, so can't speak to any experience running into the issue.

mmcc:
Trying to rebase from master to pull in #1993 was just uh...not gonna happen. Basically blew everything away and started from scratch, walking through the old diff to make sure I at least got all of the original changes.

src/js/control-bar/mute-toggle.js
```diff
@@ -64,9 +64,9 @@ class MuteToggle extends Button {
 
     /* TODO improve muted icon classes */
     for (var i = 0; i < 4; i++) {
-      Lib.removeClass(this.el_, 'vjs-vol-'+i);
+      Lib.removeClass(this.el_, `vjs-vol-${+i}`);
```
heff: `+i`?

heff:
Aside from the `+i` note this looks good to me! :+1:

src/js/control-bar/mute-toggle.js
mmcc: I prefer to be positive about things.

mmcc:
`+i` is fixed. Interesting that it actually outputs something like this:

```javascript
var bar = "foo bat " + +i;
```

Which is functionally the same. TIL.

mmcc:
woops, didn't mean to close this.

PR #2025, opened 2015-04-09
Checking on the event the relatedTarget exists. fixes #2024

### LINKED ISSUE #2024: event.relatedTarget is undefined on Firefox
In src/js/events.js we create the relatedTarget like this (line 55)

```
// Handle which other element the event is related to
event.relatedTarget = event.fromElement === event.target ?
  event.toElement :
  event.fromElement;
```

The problem is that Firefox does not support event.fromElement. Therefore the above code will overwrite the correct relatedTarget with undefined. 

Fix: Firefox does support relatedTarget (https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget). So it simply needs to check the existence of this property before setting it.  

note: This issue has already been discussed in ES6ification pull request (https://github.com/videojs/video.js/pull/1976/files#r27894428).

mischizzle:
- Checking the relatedTarget exists on the event,
- .idea to .gitignore, 
- added tests for the fix and a corresponding test helper to create mouseevent.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -16,11 +16,13 @@
         "_V_",
         "goog",
         "console",
+
         "require",
         "define",
         "module",
         "exports",
         "process",
+
         "q",
         "asyncTest",
         "deepEqual",
@@ -37,7 +39,6 @@
         "stop",
         "strictEqual",
         "test",
-        "sinon",
-        "MouseEvent"
+        "sinon"
     ]
 }
--- a/.jshintrc
+++ b/.jshintrc
@@ -16,13 +16,11 @@
         "_V_",
         "goog",
         "console",
-
         "require",
         "define",
         "module",
         "exports",
         "process",
-
         "q",
         "asyncTest",
         "deepEqual",
@@ -39,6 +37,7 @@
         "stop",
         "strictEqual",
         "test",
-        "sinon"
+        "sinon",
+        "MouseEvent"
     ]
 }
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -52,9 +52,11 @@ var fixEvent = function(event) {
     }
 
     // Handle which other element the event is related to
-    event.relatedTarget = event.fromElement === event.target ?
-      event.toElement :
-      event.fromElement;
+    if(!event.relatedTarget) {
+      event.relatedTarget = event.fromElement === event.target ?
+        event.toElement :
+        event.fromElement;
+    }
 
     // Stop the default browser action
     event.preventDefault = function () {
--- a/test/unit/events.js
+++ b/test/unit/events.js
@@ -1,6 +1,5 @@
 import * as Events from '../../src/js/events.js';
 import document from 'global/document';
-import TestHelpers from './test-helpers.js';
 
 q.module('Events');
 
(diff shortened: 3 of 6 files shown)


test/unit/events.js
```diff
@@ -201,3 +202,23 @@ test('should have a defaultPrevented property on an event that was prevent from
...
+
+  var el1 = document.createElement('div'),
+      el2 = document.createElement('div'),
+      relatedEl = document.createElement('div');
+
+  Events.on(el1, 'click', function(e){
+    equal(e.relatedTarget, relatedEl, 'relatedTarget is set for all browsers when referring element is set on the event');
+  });
+
+  Events.trigger(el1, TestHelpers.makeMouseEvent('click', relatedEl));
```
heff: If we're using Events.trigger I think we could just use a plain object here instead of creating a MouseEvent. We end up [rebuilding it as an object](https://github.com/videojs/video.js/blob/master/src/js/events.js#L31) either way. e.g. `{ type: 'click', relatedTarget: relatedEl }`

.jshintrc
```diff
@@ -16,13 +16,11 @@
         "_V_",
         "goog",
         "console",
-
```
heff: The line breaks here provided a bit of categorization since we can't use comments in JSON. It's kind of window vars, node vars, and testing vars.

heff:
Thanks for the help! Made a few comments but otherwise this looks good to me.

src/js/events.js
```diff
@@ -52,9 +52,11 @@ var fixEvent = function(event) {
     }
 
     // Handle which other element the event is related to
-    event.relatedTarget = event.fromElement === event.target ?
-      event.toElement :
-      event.fromElement;
+    if(!event.relatedTarget) {
+      event.relatedTarget = event.fromElement === event.target ?
+        event.toElement :
+        event.fromElement;
```
mmcc: This is a pre-existing condition and I think I noticed this in a few other places when I was checking the classes PR, but multi-line ternary operators....ಠ_ಠ

test/unit/events.js
mischizzle: It's definitely simpler; that's actually what I started with.. but then seems to defeat the purpose to test a property that's undefined in some browsers by setting it explicitly in trigger? ie testing it against a 'real' event. But if in the end it doesn't really matter, I'm happy to change it.

.jshintrc
mischizzle: No problem; sorry I'm a little ocd about these things! I'll revert

src/js/events.js
mischizzle: I did see the comments about preferring plain ol if/else over ternary. Alas it is quite the beauty </sarcasm>.. I'll attempt a facelift..

heff: Most of the events stuff was from Resig and still has his style, so all of that could be reformatted at some point, but don't feel like you need to in this PR. It's typically best to keep PRs and commits very specific.

test/unit/events.js
heff: It looks like makeMouseEvent is still manually setting relatedTarget on the created events, so I might be missing a detail but it seems like essentially the same thing.  I think to get a true real event we would have to use the element's `focus()` API or something like that to trigger a real event where the browser sets the related target. But that gets into integration test land, where I think a simple test to make sure we're not dropping relatedTarget in our code could be good enough.

src/js/events.js
mmcc: kick the can

Nah I'm kidding, @heff's right. this is a pre-existing condition and you shouldn't feel like you need to clean it up here, particularly if it's going to muddy up the diff on this PR. We should have a pre-5.0 cleanup where we actually pick a style guide and make the necessary changes.

mischizzle: ok left it as is; I'm happy with the minimum number pieces of flair.

test/unit/test-helpers.js
```diff
@@ -39,4 +39,4 @@ var TestHelpers = {
   }
 };
 
-export default TestHelpers;
\ No newline at end of file
+export default TestHelpers;
```
mischizzle: Not sure why it's diff'ing here; does my IDE need some kind of formatter on this line?

heff:
Got this merged in (finally). Thanks again!

mischizzle:
Hey no problem! Kind of forgot about it to be honest .. no conflicts from being stale, that i can see

PR #2038, opened 2015-04-15
Don't allow progress bar to overflow

bc-bbay:
If something goes wrong in a calculation and player.currentTime() is greater than player.duration(), lock the progressControl at 100%. This is a better user experience than overflowing progress off the right end of the control.


gkatsev:
LGTM, though, needs to be rebased against master.

gkatsev:
@videojs/core-committers would we want to backport this to stable?

bc-bbay:
Opened a new PR #2040 because master made a lot of big changes to the area and it was easier just to redo them than try and update the other branch.

PR #2040, opened 2015-04-15
Don't allow progress bar to overflow

bc-bbay:
If something goes wrong in a calculation and player.currentTime() is greater than player.duration(), lock the progressControl at 100%. This is a better user experience than overflowing progress off the right end of the control.


bc-bbay:
@gkatsev New PR with changes from #2038 and latest code from master

gkatsev:
Awesome, thanks.

heff:
LGTM!

Issue #1863, opened 2015-02-13
Captions: MediaTechController initializes some listeners for TextTrackDisplay


gkatsev:
Investigate and move the `initTextTrackListeners` into the TextTrackDisplay` directly.

gkatsev:
Fixed by #2057. Closing.

Issue #2008, opened 2015-04-03
VJS 5: Tech 2.0 proposal


eXon:
Ok I wrote this. Rewrote this. Now I re-re-re-re-rewrote this because having used the tech so much their is a lot of issues I want to address. Here is the first version that is worth discussing.

--

The data flow between the player and the tech could get some improvements.
The ideal situation is to achieve a perfect two-ways data binding between the player and the tech.

- When an event happens on the VJS player, the tech player should be notified.
- When an event happens on the tech player, the VJS player should be notified.

However, the tech should not care about keeping the state of anything.
He should not know what is the current volume or the current time.
He should only send the initial state at the player and then notify when a state change.
The VJS player is the only place where the current states should be known.

*(keep in mind this is just a way to express my idea in a simple way, not how things work)*
*(I don't know if we will keep the current EmitEmitter or [use the one in node.js](https://github.com/videojs/video.js/issues/2002) I don't care that much, this is only about the techs)*
```javascript
class YoutubeTech extends VjsTech {
  constructor(element, options) {
    super(element, options);

    element.src = 'https://www.youtube.com/embed/' + videoId;

    this.onVjsPlay = this.onVjsPlay.bind(this);
    this.onVjsPause = this.onVjsPause.bind(this);

    this.on('play', this.onVjsPlay);
    this.on('pause', this.onVjsPause);
    // ...

    // I'm not so sure with the emitToOthers name but it is pretty self-explanatory, I don't think their is something built-in
    this.emitToOthers('volume', initialValue);
    this.emitToOthers('poster', 'big-fat-cat.jpg');

    this.on('dispose', function() {
      this.removeListener('play', this.onVjsPlay);
      this.removeListener('pause', this.onVjsPause);
    });
  }

  onVjsPlay() {
    // Call play on the YouTube API
  }

  onVjsPause() {
    // Call pause on the YouTube API
  }

  onYoutubePlay() {
    this.emitToOthers('play');
  }

  onYoutubePause() {
    this.emitToOthers('pause');
  }
}

YoutubeTech.createElement = function() {
    var iframe = document.createElement('iframe');
    iframe.src = 'about:blank';
    return iframe;
};
```

Then the player would only need:
```javascript
// When a new event happens in the player, it call tech.emitToOthers(event)
// When a new event happens in the tech, it calls player.emit(event)
this.broadcast(tech);
```

Your first question is probably going to be: where is my player reference?
[...]

PR #2057, opened 2015-04-22
Tech 2.0

eXon:
I'm starting a pull request to make sure I'm working in the right direction and that everyone agrees with that. (https://github.com/videojs/video.js/issues/2008)

By the way I couldn't test Flash because `videojs.options` doesn't exists anymore so I can't set the swf path.

Let me know what you guys think.

## Roadmap
- [x] Trigger events on the tech instead of the player (https://github.com/videojs/video.js/issues/1485)
- [x] Player should be responsible for adding the DOM element of the tech on himself (using tech.el_)
- [x] Remove every dependencies of the player on
    - [x] tech (fix https://github.com/videojs/video.js/issues/1863)
    - [x] html5
    - [x] flash
- [x] Remove the player from the tech constructor signature
- [x] Fix unit tests


gkatsev:
You should be able to pass in the swf via an option to the player you're creating:
```js
videojs('foo', {
  flash: {
    swf: "http://google.com"
  }
});
```

Which reminds me, we probably want a way to change these defaults globally so you won't be required to specify it for each player.

src/js/tech/flash.js
```diff
@@ -321,7 +321,7 @@ Flash['checkReady'] = function(tech){
 // Trigger events from the swf on the player
 Flash['onEvent'] = function(swfID, eventName){
   let player = Lib.el(swfID)['player'];
-  player.trigger(eventName);
+  player.tech.trigger(eventName);
```
heff: We should store a reference to the tech on the swf object, probably [after it's created](https://github.com/videojs/video.js/blob/e1e69941a13d444525ccdf494d740eb1a3285320/src/js/tech/flash.js#L104). Then we don't need to reference the player here.

src/js/tech/html5.js
```diff
@@ -176,32 +174,6 @@ class Html5 extends Tech {
...
-    }
-  }
-
-  eventHandler(evt) {
-    // In the case of an error on the video element, set the error prop
-    // on the player and let the player handle triggering the event. On
-    // some platforms, error events fire that do not cause the error
-    // property on the video element to be set. See #1465 for an example.
-    if (evt.type == 'error' && this.error()) {
-      this.player().error(this.error().code);
```
heff: We're going to need something in the player to handle this now. Right now there's no equivalent listing for error events on the player and updating the error property.

heff:
Looking great so far!

eXon:
@gkatsev The option is now working. However, it seems like Flash is never sending me the onReady event. Any clue?

I'm on Ubuntu / Chrome and Flash works fine.

src/js/tech/html5.js
eXon: I'm not sure what it was doing before. Do we need to check for `evt.type == 'error'` on ALL events or just the error event?

eXon: I think I figured it out. Now the player listen for the error event on the tech the same way. This seems like a very good direction to take to abstract from the player.

src/js/tech/flash.js
```diff
@@ -26,9 +26,6 @@ class Flash extends Tech {
 
     let { source, parentEl } = options;
 
-    // Create a temporary element to be replaced by swf object
-    let placeHolder = this.el_ = Lib.createEl('div', { id: player.id() + '_temp_flash' });
```
eXon: I tried to guess why we were using this placeholder and I ended up looking at 2 years old commits. My only guess is that is was needed in the iframe mode. It seems to work perfectly fine without using it. @heff any clue?

src/js/tech/flash.js
heff: Yeah, I think the placeholder was actually a remnant of using swfobject.js, which required you to replace an existing element. #1340 got us a little closer but I think the placeholder could be refactored out now.

This specific line though is the swf's reference to the tech, not the tech's reference to the swf. We need the swf to have a reference because the Flash events get triggered globally and we need to figure out which tech they apply to.

src/js/tech/html5.js
heff: Yeah, awesome.

src/js/tech/flash.js
eXon: This is an out-dated diff (I think I misplaced my comment for the placeholder). It is now:
```javascript
let tech = Lib.el(swfID)['tech'];
tech.trigger(eventName);
```

and the tech is saved within the element instead of the player.

src/js/player.js
```diff
@@ -454,12 +512,95 @@ class Player extends Component {
...
+  onTechError() {
+    this.error(this.tech.error().code);
+    this.trigger('error');
+  }
+
+  /**
+   * Fires when the browser is intentionally not getting media data
+   * @event suspend
+   */
+  onTechSuspend() {
```
heff: For these simple ones we could probably loop though a list of the events and create these dynamically. Though it is kind of nice to have an excuse to add the event jsdoc tag.

src/js/tech/flash.js
```diff
@@ -101,7 +95,8 @@ class Flash extends Tech {
     // use stageclick events triggered from inside the SWF instead
     player.on('stageclick', player.reportUserActivity);
 
-    this.el_ = Flash.embed(options['swf'], placeHolder, flashVars, params, attributes);
+    this.el_ = Flash.embed(options['swf'], flashVars, params, attributes);
+    this.el_['tech'] = this;
```
heff: We should use dot notation here now that we're not using closure compiler. `this.el_.tech`

src/js/tech/html5.js
```diff
@@ -139,8 +137,6 @@ class Html5 extends Tech {
           el.appendChild(trackEl);
         }
       }
-
-      Lib.insertFirst(el, player.el());
```
heff: One thing we need to be aware of here is that the HTML5 tech only added it's element to the player if it wasn't reusing the embed code video tag that was [already in the player](https://github.com/videojs/video.js/blob/e1e69941a13d444525ccdf494d740eb1a3285320/src/js/player.js#L211). Now, the player will call insertFirst for the HTML5 tech element even if the tag is already there. I'm not sure what, if any impact that will have.

heff:
So it looks like we previously broke Flash. The swf is still told to fire the [`videojs.Flash.onReady` event](https://github.com/videojs/video.js/blob/master/src/js/tech/flash.js#L44), but we've [changed that to just `Flash.onReady`](https://github.com/videojs/video.js/blob/master/src/js/tech/flash.js#L285), and it's not longer available globally. We'll need to re-export that to the window some how.

In videojs/video-js-swf#143 we hardcoded these functions to prevent security issues. So either we'll need to expose window.videojs everywhere again, or create new global functions and release a new version of the swf.

gkatsev:
It would be awesome if we export a different flash-onready function per player on the window object :)

heff:
Is there a way we can do that without opening the security hole back up?

src/js/tech/flash.js
eXon: I agree it's ugly. I was going for consistency but I'll switch them while going through the tech files.

src/js/tech/html5.js
eXon: Yes I am aware of that. It only insert the element if it is not already in the player:

```javascript
// Add the tech element in the DOM if it was not already there
if (this.tech.el().parentNode != this.el()) {
  Lib.insertFirst(this.tech.el(), this.el());
}
```

heff: Oh, I missed that. Very nice!

gkatsev:
Oh, right, completely forgot about that. Hm...

src/js/player.js
```diff
@@ -276,6 +258,44 @@ class Player extends Component {
...
+    this.on(this.tech, 'ended', this.onTechEnded);
+    this.on(this.tech, 'seeking', this.onTechSeeking);
+    this.on(this.tech, 'seeked', this.onTechSeeked);
+    this.on(this.tech, 'ended', this.onTechEnded);
+    this.on(this.tech, 'play', this.onTechPlay);
+    this.on(this.tech, 'firstplay', this.onTechFirstPlay);
+    this.on(this.tech, 'pause', this.onTechPause);
+    this.on(this.tech, 'progress', this.onTechProgress);
+    this.on(this.tech, 'durationchange', this.onTechDurationChange);
+    this.on(this.tech, 'fullscreenchange', this.onTechFullscreenChange);
```
heff: fullscreenchange is actually a player event, not tech.

src/js/player.js
```diff
@@ -276,6 +258,44 @@ class Player extends Component {
...
+    this.on(this.tech, 'volumechange', this.onTechVolumeChange);
+    this.on(this.tech, 'texttrackchange', this.onTextTrackChange)
+
+    // Private events between the player and the tech
+    this.on(this.tech, 'useractive', this.onTechUserActive);
+    this.on(this.tech, 'userinactive', this.onTechUserInactive);
+
+    // Add the tech element in the DOM if it was not already there
+    // Make sure to not insert the original video element if using Html5
+    if (this.tech.el().parentNode != this.el() && (techName !== 'Html5' || !this.tag)) {
```
heff: Need another `=` here. `parentNode !== this.el`

src/js/player.js
```diff
@@ -454,12 +518,119 @@ class Player extends Component {
...
    */
-  onFullscreenChange() {
+  onTechFullscreenChange() {
     if (this.isFullscreen()) {
       this.addClass('vjs-fullscreen');
     } else {
       this.removeClass('vjs-fullscreen');
     }
+
+    this.trigger('fullscreenchange');
```
heff: This should be reverted since it's a player event.

src/js/player.js
```diff
@@ -454,12 +518,119 @@ class Player extends Component {
...
+  }
+
+  /**
+   * Fires when an error occurred during the loading of an audio/video
+   * @event error
+   */
+  onTechError() {
+    this.error(this.tech.error().code);
+    this.trigger('error');
```
heff: Calling `player.error(err)` already triggers the player error event, so this would be triggering a second. We should probably be more consistent with how we handle states like this, so we're adding css classes and triggering events the same way for all of them. We don't need to figure that out in this PR specifically though.

src/js/tech/tech.js
```diff
@@ -69,32 +70,46 @@ class Tech extends Component {
...
+    } else {
+      this.removeControlsListeners();
+    }
+  }
+
+  setNativeControls(bool) {
+    this.nativeControls_ = bool;
+  }
+
+  setUserActive(bool) {
```
heff: I think it might be better to leave all the userActive tracking to the player. We could have the tech still trigger click and tap events, and then the player will update it's userActive state when it sees those tech events.

src/js/tech/tech.js
```diff
@@ -179,6 +212,8 @@ class Tech extends Component {
   // Manually trigger progress events based on changes to the buffered amount
   // Many flash players and older HTML5 browsers don't send progress or progress-like events
   manualProgressOn() {
+    this.on('durationchange', this.onDurationChange);
```
heff: Why is durationchange tracking needed here?

src/js/tech/tech.js
```diff
@@ -205,7 +242,40 @@ class Tech extends Component {
...
       }
-    }, 500);
+    }), 500);
+  }
+
+  onDurationChange() {
+    duration_ = this.duration();
+  }
+
+  bufferedPercent() {
```
heff: If this is being duplicated from the player. If so we should probably create a utility for it so it can be shared.

Also it doesn't look like `duration_` is ever defined in this instance.

src/js/tech/tech.js
```diff
@@ -267,21 +333,15 @@ class Tech extends Component {
...
   // TODO: Consider looking at moving this into the text track display directly
   // https://github.com/videojs/video.js/issues/1863
   initTextTrackListeners() {
-    let player = this.player_;
-
-    let textTrackListChanges = function() {
-      let textTrackDisplay = player.getChild('textTrackDisplay');
-
-      if (textTrackDisplay) {
-        textTrackDisplay.updateDisplay();
```
heff: It looks like we're removing this but not replacing it elsewhere. We should probably do what the TODO above it says. 'Consider looking at moving this into the text track display directly'. ie. have the textTrackDisplay listen for player texttrackchange events.

src/js/tech/tech.js
```diff
@@ -339,13 +395,13 @@ class Tech extends Component {
    */
 
   textTracks() {
-    this.player_.textTracks_ = this.player_.textTracks_ || new TextTrackList();
-    return this.player_.textTracks_;
+    this.textTracks_ = this.textTracks_ || new TextTrackList();
```
heff: I think the tracks changes are going to be a little more complicated then just shifting things to the tech like this. @gkatsev should review this part.

src/js/tracks/text-track-display.js
```diff
@@ -32,6 +32,7 @@ class TextTrackDisplay extends Component {
     super(player, options, ready);
 
     player.on('loadstart', Lib.bind(this, this.toggleDisplay));
+    player.on('texttrackchange', Lib.bind(this, this.toggleDisplay));
```
heff: Ha, ignore my previous comment about the TODO. We can remove the TODO line with this change and it will close the related issue.

src/js/player.js
```diff
@@ -1207,6 +1378,8 @@ class Player extends Component {
       // Don't trigger a change event unless it actually changed
       if (this.controls_ !== bool) {
         this.controls_ = bool;
+        this.techCall('setControls', bool);
```
heff: Controls are complicated... Here we should check if `usingNativeControls()` returns true before passing the controls setting to the tech. Otherwise we'll end up with both the custom and native controls showing.

src/js/player.js
```diff
@@ -1240,6 +1413,8 @@ class Player extends Component {
       // Don't trigger a change event unless it actually changed
       if (this.usingNativeControls_ !== bool) {
```
heff: I think I'd actually like to remove this ability to switch between native and custom controls mid-playback. The decision to use native controls or not should be made when the tech is first initialized, and then not changed. That would simplify the controls situation a bunch.

The getter part of `player.usingNativeControls` is still needed though. It would just return the value from the tech.

src/js/tech/tech.js
eXon: We need it to compute `bufferedPercent`.

eXon:
We might also need to find a way to catch the mouse events even when the tech is an iframe (like YouTube). Out of scope of this though.

eXon:
No more player available in the tech :D. Now I can check the 100s of things I might have broken and the concerns of everyone about the changes.

src/js/button.js
```diff
@@ -70,7 +70,7 @@ class Button extends Component {
   // KeyPress (document level) - Trigger click when keys are pressed
   onKeyPress(event) {
     // Check for space bar (32) or enter (13) keys
-    if (event.which == 32 || event.which == 13) {
+    if (event.which === 32 || event.which === 13) {
```
heff: We should be careful about changes like this that aren't directly related and touch a lot of files. We've got a hack week going on right now and master is going to change a lot, so there's going to be a lot of merge conflicts.

src/js/player.js
```diff
@@ -138,6 +138,9 @@ class Player extends Component {
     this.userActive_ = true;
     this.reportUserActivity();
     this.listenForUserActivity();
+
+    this.on('fullscreenchange', this.onFullscreenChange);
+    this.on('stageclick', this.onStageClick);
```
heff: `stageclick` is Flash specific, so we should just capture that in the flash tech and fire a click.

src/js/player.js
```diff
@@ -274,7 +263,54 @@ class Player extends Component {
...
     // Initialize tech instance
     let techComponent = Component.getComponent(techName);
-    this.tech = new techComponent(this, techOptions);
+    this.tech = new techComponent(techOptions);
+
+    this.on(this.tech, 'ready', this.onTechReady);
+    this.on(this.tech, 'usenativecontrols', this.onTechUseNativeControls);
+
+    // firefox doesn't bubble mousemove events to parent. videojs/video-js-swf#37
+    // bugzilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=836786
```
heff: The firefox bug there says it's resolved, so we should test if we can remove this.

src/js/button.js
eXon: Yea sorry about that. I should have left the eqeqeq option in jslint for another pull request. I'm trying to fix the unit tests ASAP so we could merge quickly.

src/js/player.js
heff: This fullscreenchange listener could still be removed, now that you've added it to the player.

src/js/player.js
eXon: stageclick is a stinky name, I can change that

src/js/player.js
```diff
@@ -274,7 +263,54 @@ class Player extends Component {
...
+    this.on(this.tech, 'volumechange', this.onTechVolumeChange);
+    this.on(this.tech, 'texttrackchange', this.onTextTrackChange);
+
+    if (this.controls() && !this.usingNativeControls()) {
+      this.addTechControlsListeners();
+    }
+
+    // Add the tech element in the DOM if it was not already there
+    // Make sure to not insert the original video element if using Html5
+    if (this.tech.el().parentNode !== this.el() && (techName !== 'Html5' || !this.tag)) {
```
heff: Would the parentNode check be enough here? If the html5 tech is using the tag then the parentNode check should return false, or at least I can't think of a scenario where that wouldn't work.

eXon: The parentNode is null for the original video tag when we do the check at this point so we have to make sure to not add it twice. If I remove this check, you end up with two video tag. I don't even know why the parentNode is null but checking `player.tag` seems to do the trick.

src/js/player.js
```diff
@@ -287,11 +323,95 @@ class Player extends Component {
...
+      }
+    });
+
+    this.on(this.tech, 'touchend', function(event) {
+      // Stop the mouse events from also happening
+      event.preventDefault();
+    });
+
+    // Turn on component tap events
+    this.tech.emitTapEvents();
```
heff: We should probably just turn this on permanently in the Tech constructor. The player should already handle listening for it or not.

src/js/player.js
heff: Nevermind. Looks like we were just using it for `reportUserActivity` before, so I guess the player would need to listen for it.

src/js/player.js
```diff
@@ -463,6 +654,107 @@ class Player extends Component {
...
   /**
+   * native click events on the SWF aren't triggered on IE11, Win8.1RT
+   * use stageclick events triggered from inside the SWF instead
+   * @private
+   */
+  onStageClick() {
+    this.reportUserActivity();
+  }
+
+  onTechFullscreenChange() {
```
heff: This can be removed.

src/js/player.js
```diff
@@ -1203,14 +1495,25 @@ class Player extends Component {
       // Don't trigger a change event unless it actually changed
       if (this.controls_ !== bool) {
         this.controls_ = bool;
+
+        if (!this.usingNativeControls()) {
```
heff: I think this is backwards. We only want to set the controls on the tech if we *are* using native controls.

src/js/tech/flash.js
```diff
@@ -21,21 +21,13 @@ let navigator = window.navigator;
  */
 class Flash extends Tech {
 
-  constructor(player, options, ready){
-    super(player, options, ready);
+  constructor(options, ready){
+    super(options, ready);
```
heff: `super` is Component and would still require player as the first arg. In order to make this happen we would probably need to update Component to be a generic UI component class that doesn't require a player, and then create a PlayerComponent subclass that does. That will take some work, so we might want to save that for the end, or even a secondary PR after we finish everything else here.

src/js/tech/html5.js
```diff
@@ -105,29 +91,29 @@ class Html5 extends Tech {
         const clone = el.cloneNode(false);
         Html5.disposeMediaElement(el);
         el = clone;
-        player.tag = null;
+        // Is it really necessary?
+        //player.tag = null;
```
heff: I don't remember the specifics, but it is necessary. We should probably just null it when loading the first tech, or find a way to not store it in the first place.

src/js/tech/html5.js
```diff
@@ -275,14 +203,11 @@ class Html5 extends Tech {
 
     if ('webkitDisplayingFullscreen' in video) {
       this.one('webkitbeginfullscreen', function() {
-        this.player_.isFullscreen(true);
-
         this.one('webkitendfullscreen', function() {
-          this.player_.isFullscreen(false);
-          this.player_.trigger('fullscreenchange');
+          this.trigger('fullscreenchange');
```
heff: Ah, I see why the techFullscreenChange listeners were kept. These events won't really work though unless we're also setting isFullscreen to true or false somewhere also.

src/js/tech/tech.js
```diff
@@ -18,29 +18,30 @@ import document from 'global/document';
...
 class Tech extends Component {
 
-  constructor(player, options={}, ready=function(){}){
+  constructor(options={}, ready=function(){}){
+    options = options || {};
     // we don't want the tech to report user activity automatically.
     // This is done manually in addControlsListeners
     options.reportTouchActivity = false;
-    super(player, options, ready);
+    super(null, options, ready);
```
heff: Ok, I see what you did here. That works. :)

heff:
Looking pretty good @eXon. Does the player at least still play with all these changes?

eXon:
@heff surpisingly yes, even after those massive changes. The html5 tech is
playing fine but unit test are failing (working on that tonight) and the
unit tests for flash is working but i can't test it.

heff:
Nice, that's great.

eXon:
I still have 6 unit tests that I have troubles fixing. Anyone care to help a little bit?

```
Chrome 42.0.2311 (Linux) Player should get tag, source, and track settings FAILED
	tag player ref killed
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:3330:17 <- /var/git/eXon/video.js/test/unit/player.js:127:15)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:186:5
	
...WARN [web-server]: 404: /undefined
....WARN [web-server]: 404: /undefined
WARN [web-server]: 404: /undefined
.................................
Chrome 42.0.2311 (Linux) Text Track Controls should be displayed when text tracks list is not empty FAILED
	control is displayed
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4330:40 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:17:39)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
	textTracks contains one item
	Expected: 1
	Actual: 0
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4331:28 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:18:28)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
...
Chrome 42.0.2311 (Linux) Text Track Controls menu should contain "Settings", "Off" and one track FAILED
	menu contains three items
	Expected: 3
	Actual: 2
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4367:18 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:54:18)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
	Died on test #4     at Object.require./var/git/eXon/video.js/test/unit/tracks/text-track-controls.js.../../../src/js/control-bar/text-track-controls/text-track-menu-item.js (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4361:1 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:48:0)
	    at s (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:1:262 <- ../node_modules/karma-browserify/node_modules/browserify/node_modules/browser-pack/_prelude.js:1:0)
	    at /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:1:1: Cannot read property 'track' of undefined
	TypeError: Cannot read property 'track' of undefined
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4370:21 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:57:21)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
Chrome 42.0.2311 (Linux) Text Track Controls menu should update with addRemoteTextTrack FAILED
	menu does contain added track
	Expected: 4
	Actual: 3
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4380:47 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:67:47)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
	textTracks contains two items
	Expected: 2
	Actual: 1
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4381:28 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:68:28)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
Chrome 42.0.2311 (Linux) Text Track Controls menu should update with removeRemoteTextTrack FAILED
	menu does not contain removed track
	Expected: 3
	Actual: 2
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4391:47 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:78:47)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
	textTracks contains one item
	Expected: 1
	Actual: 0
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:4392:28 <- /var/git/eXon/video.js/test/unit/tracks/text-track-controls.js:79:28)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
...............................................
Chrome 42.0.2311 (Linux) Tracks when switching techs, we should not get a new text track FAILED
	the tracks are equal
	    at Object.<anonymous> (/tmp/3428dda50681355a7ad7c308cf2891df.browserify:5583:17 <- /var/git/eXon/video.js/test/unit/tracks/tracks.js:321:15)
	    at Object.Test.run (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1303:18)
	    at /var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1463:10
	    at process (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1016:24)
	    at next (/var/git/eXon/video.js/node_modules/qunitjs/qunit/qunit.js:1009:3)
	
...........
Chrome 42.0.2311 (Linux): Executed 206 of 0 (6 FAILED) (4.569 secs / 4.469 secs)

=============================== Coverage summary ===============================
Statements   : 76.97% ( 4659/6053 )
Branches     : 59.49% ( 1837/3088 )
Functions    : 77.82% ( 863/1109 )
Lines        : 80.15% ( 3545/4423 )
================================================================================
Warning: Task "karma:chrome" failed. Use --force to continue.

Aborted due to warnings.
```

src/js/tech/tech.js
gkatsev: Also, we don't actually want to shift it to the tech. This is because if you switch techs, you would lose the text tracks. We do not want that to happen.
There's probably something extra need to do to make text tracks work with techs that have their own tracks like the youtube tech, but for regular old tracks we want it on the player.

eXon: @gkatsev You are right, it is causing me a lot of trouble with the unit tests. I think using an event that the player listent to would be more appropriate though. Would it work?

dmlap:
@eXon you might want to consider merging in #2097 and #2098 so you can debug the tests in a browser. There's a way to get karma to keep its browser instances running but I don't recall it off-hand.

eXon:
Everything is done and fixed! :D now start the merge fest and then we can close this

eXon:
Everything is working perfectly now :) 

@heff let me know if I need to change anything quickly so I don't have to merge too much stuff (maybe someone funny would rename all the events function in the player...... just kidding ;)).

src/js/tech/tech.js
eXon: Instead what I did is when unloading a tech, I save the textTracks and give it back to the next tech with an option

eXon:
@dmlap you can enable watch and disable singleRun to do that, but I think
we could find a way to improve the browserify speed for unit tests because
it is really slow

src/js/tech/tech.js
gkatsev: Interesting. I guess that's another way of doing it :). I'll take a closer look at this tomorrow.

src/js/player.js
```diff
@@ -280,16 +266,24 @@ class Player extends Component {
...
+      this.triggerReady();
+    });
 
     // Grab tech-specific options from player options and add source and parent element to use.
-    var techOptions = Lib.obj.merge({ 'source': source, 'parentEl': this.el_ }, this.options_[techName.toLowerCase()]);
+    var techOptions = Lib.obj.merge({
+      'source': source,
+      'playerId': this.id(),
+      'textTracks': this.textTracks_
+    }, this.options_[techName.toLowerCase()]);
```
gkatsev: if tracks are added manually after the config and then the tech changes, would this pick up these changes?

eXon: @gkatsev Yes! It saves the `textTracks` just before unloading the tech and add it to the options for the next `loadTech`

gkatsev: I guess I just haven't gotten there yet :)

src/js/player.js
```diff
@@ -450,13 +602,66 @@ class Player extends Component {
...
+
+    this.trigger('ended');
   }
 
   /**
    * Fired when the duration of the media resource is first known or changed
    * @event durationchange
    */
-  handleDurationChange() {
+  handleTechDurationChange() {
```
gkatsev: speaking of which, do you think this change would allow us to update the control bar with the duration before we hit play? https://github.com/videojs/video.js/blob/v4.12.5/src/js/control-bar/time-display.js#L43-L55

src/js/tech/flash.js
gkatsev: I believe the placeholder used to be required for the flash fallback back in the day. I think it can be safely removed now.

src/js/player.js
eXon: I'm not sure. I guess we should check but I don't think

src/js/tech/tech.js
```diff
@@ -18,32 +18,35 @@ import document from 'global/document';
...
+    if (options.nativeCaptions === false || options.nativeTextTracks === false) {
+      this.featuresNativeTextTracks = false;
     }
 
-    if (!this['featuresNativeTextTracks']) {
+    if (!this.featuresNativeTextTracks) {
       this.emulateTextTracks();
     }
 
+    this.textTracks_ = options.textTracks;
```
gkatsev: Cool.

gkatsev: I wonder if it might be better to merge these? Though, I guess we shouldn't ever have `textTracks_` already set in the constructor.

src/js/player.js
gkatsev: Would be nice if the duration stuff you made in this PR help us fix that issue.

gkatsev:
Looks pretty good. I still wonder whether there's anything that needs to happen here so we can fix the duration issue in the time display: https://github.com/videojs/video.js/blob/v4.12.5/src/js/control-bar/time-display.js#L43-L55

src/js/control-bar/text-track-controls/text-track-button.js
```diff
@@ -62,4 +62,4 @@ class TextTrackButton extends MenuButton {
 }
 
 MenuButton.registerComponent('TextTrackButton', TextTrackButton);
-export default TextTrackButton;
```
heff: Why's the export removed here?

src/js/player.js
```diff
@@ -311,11 +360,82 @@ class Player extends Component {
     this.tech = false;
   }
 
+  addTechControlsListeners() {
+    let userWasActive;
```
heff: This isn't being used anywhere anymore

src/js/player.js
heff: Still need to address this one.

src/js/tech/flash.js
heff: Yeah, placeholder can be removed now.

heff:
I attempted to help by resolving merge conflicts in another PR to this one.

Otherwise there's a few remaining issues open that should be addressed, but I think we're close? We should definitely get this in ASAP, or at least start getting pieces of it merged.

src/js/control-bar/text-track-controls/text-track-button.js
eXon: Oops my bad, fixing this

src/js/player.js
eXon: I think you're right, fixing it

src/js/control-bar/text-track-controls/text-track-button.js
eXon: Oh! I haven't removed the line, only the EOF empty line (stupid Atom doing that all the time :/). You can check it's still there.

src/js/player.js
eXon: Their is no more time display in the new skin. I don't really know what else I can do.

src/js/tech/tech.js
eXon: Moving it before the other textTracks call should be enough.

eXon:
Hopefully I have addressed everything here and we can move on smaller PR now :)

eXon:
Big thanks for the merge by the way @heff, very helpful

gkatsev:
Looks like you need to rebase against master again, sorry. On the bright side, the flash tech should be working again :)

eXon:
Yay! I'll try tonight :)

eXon:
Flash is working perfectly now :) let's merge this

src/js/control-bar/text-track-controls/text-track-button.js
heff: Ok cool. Yeah I had to update my atom config to fix that recently.

src/js/player.js
```diff
@@ -280,16 +266,24 @@ class Player extends Component {
     // Turn off API access because we're loading a new tech that might load asynchronously
     this.isReady_ = false;
 
-    var techReady = function(){
-      this.player_.triggerReady();
-    };
+    var techReady = Lib.bind(this, function() {
```
heff: I think this is duplicating line 297 now? https://github.com/videojs/video.js/pull/2057/files#diff-3b0266ff1c037b289ec624ab25b0272eR297

src/js/player.js
```diff
@@ -311,11 +354,72 @@ class Player extends Component {
...
+
+    // If the controls were hidden we don't want that to change without a tap event
+    // so we'll check if the controls were already showing before reporting user
+    // activity
+    this.on(this.tech, 'touchstart', this.handleTechTouchStart);
+    this.on(this.tech, 'touchmove', this.handleTechTouchMove);
+    this.on(this.tech, 'touchend', this.handleTechTouchEnd);
+
+    // Turn on component tap events
+    this.tech.emitTapEvents();
```
heff: This should really be handled by the tech. Possibly turned on for all techs, unless there's a reason not to with players like youtube.

src/js/player.js
```diff
@@ -280,16 +266,24 @@ class Player extends Component {
...
+    var techReady = Lib.bind(this, function() {
+      this.triggerReady();
+    });
 
     // Grab tech-specific options from player options and add source and parent element to use.
-    var techOptions = Lib.obj.merge({ 'source': source, 'parentEl': this.el_ }, this.options_[techName.toLowerCase()]);
+    var techOptions = Lib.obj.merge({
+      'source': source,
+      'playerId': this.id(),
+      'textTracks': this.textTracks_
```
heff: We're going to need to pass in a lot more options here now, including `autoplay`, `preload`, `loop`, and `muted`. Basically all the html5 video attribute options.

src/js/tech/html5.js
```diff
@@ -105,29 +91,27 @@ class Html5 extends Tech {
...
-        if (!Lib.TOUCH_ENABLED || player.options()['nativeControlsForTouch'] !== true) {
+        let tagAttributes = this.options_.tag && Lib.getElementAttributes(this.options_.tag);
+        let attributes = VjsUtil.mergeOptions({}, tagAttributes);
+        if (!Lib.TOUCH_ENABLED || this.options_.nativeControlsForTouch !== true) {
           delete attributes.controls;
         }
 
         Lib.setElementAttributes(el,
           Lib.obj.merge(attributes, {
-            id: player.id() + '_html5_api',
```
heff: We should just have the player generate the full ID when it's building the tech.

src/js/tech/tech.js
```diff
@@ -18,29 +18,32 @@ import document from 'global/document';
  */
 class Tech extends Component {
 
-  constructor(player, options={}, ready=function(){}){
+  constructor(options={}, ready=function(){}){
+    options = options || {};
```
heff: This line can be removed now that there's a default arg.

src/js/tech/tech.js
```diff
@@ -204,7 +118,40 @@ class Tech extends Component {
...
       }
-    }, 500);
+    }), 500);
+  }
+
+  onDurationChange() {
+    this.duration_ = this.duration();
+  }
+
+  bufferedPercent() {
```
heff: This is still duplicated logic from the player I think. We should figure out the right way to share that. It could be just having the player ask the tech for buffered percent instead  of calculating itself.

src/js/tech/tech.js
```diff
@@ -213,36 +160,38 @@ class Tech extends Component {
...
 
-    this.on(player, 'play', this.trackCurrentTime);
-    this.on(player, 'pause', this.stopTrackingCurrentTime);
+    this.on('play', this.trackCurrentTime);
+    this.on('pause', this.stopTrackingCurrentTime);
     // timeupdate is also called by .currentTime whenever current time is set
 
-    // Watch for native timeupdate event
-    this.one('timeupdate', function(){
+    // Watch for native timeupdate event only
```
heff: We should just remove this bit now. It was originally added because at the time, not all browsers supported the timeupdate event.

test/unit/player.js
```diff
@@ -297,9 +297,9 @@ test('should set controls and trigger events', function() {
     ok(true, 'disabled fired once');
   });
   player.controls(false);
-  player.controls(true);
+  //player.controls(true);
```
heff: Why are these disabled now?

heff:
I want to go ahead and pull this in to prevent further merge conflicts, but there's a list of follow tasks we'll need to do to really finish it off. See #2126.

eXon:
Awesome :) I'll take a closer look at all this in the upcoming days

heff:
Woo!

heff:
@eXon do you think you could add any relevant notes for other tech developers to the 5.0 changes?

eXon:
Of course. You can add this to #2126

PR #2061, opened 2015-04-23
Added watchify task

eXon:
If I don't get a watchify task to work I'll start punching people in the face. Seriously, we need it to stay sane or we have to wait 30 secs everytime we want to just console.log something.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -277,6 +277,34 @@ module.exports = function(grunt) {
         }
       }
     },
+    watchify: {
+      options: {
+        debug: true,
+        standalone: 'videojs',
+        keepalive: true,
+        callback: function(bundle) {
+          bundle.transform(require('babelify').configure({
+            sourceMapRelative: './src/js'
+          }));
+
+          bundle.transform('browserify-versionify', {
+            placeholder: '__VERSION__',
+            version: pkg.version
+          });
+
+          bundle.transform('browserify-versionify', {
+            placeholder: '__VERSION_NO_PATCH__',
+            version: version.majorMinor
+          });
+
+          return bundle;
+        }
+      },
+      build: {
+        src: ['./src/js/video.js'],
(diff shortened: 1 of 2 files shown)


gkatsev:
I think it'll be better to use `grunt-browserify` that we already use to set up the watchify tasks.

gkatsev:
But yes, we definitely need watchify.

eXon:
@gkatsev We need to use browserify directly if we want to do that instead of grunt-browserify. I don't think it's a big deal to duplicate the options considering it doesn't work the same way. If you have an example on how to use watchify within grunt-browserify let me know.

gkatsev:
You only need to duplicate the regular grunt-browserify config and add the [`watch`](https://github.com/jmreidy/grunt-browserify#watch) option for it to use watchify. Let me know if that works.

eXon:
OMG didnt knew that option. i'm more used to gulp. I'll change it.

gkatsev:
Heh, no worries.
@heff @mmcc do you know why travis is failing here?

mmcc:
No clue, but re-running it fixed it.

eXon:
It looks like every PR I do the build fails. I'm doomed.

eXon:
I'm now using grunt-browserify to run watchify. However, here is the sad part: if we use browserify:watch with all other tasks in the watch, it doesn't work. 

I personally don't want to run everything when I save a file (especially that Karma is taking forever to open and close the browser). Using `grunt browserify:watch` now give us a way to develop and rebundle the javascript in a few seconds instead of over 30 secs. 

I would rather run everything once I'm done with what I'm doing.

gkatsev:
I think that sounds fine. `browserify:watch` is just for development and you should make sure to rerun `grunt build` and `grunt test` and what not before finishing up.

gkatsev:
Anyway, LGTM

heff:
lgtm

Issue #1980, opened 2015-03-26
Initialize the plugins before the components


carpasse:
Hi guys.

I use the plugins to get the data needed for the controls of the player. 
I usually use the plugins to get the data and have a separation between the business logic and the UI (Components) therefore I would like to propose that we initialize the plugins before we initialize the components.

What do you think??

carpasse:
Another possibility would be to enhance the build phase of a component to specify dependencies to a list of plugins

heff:
Yeah, I think initializing plugins before controls is a good idea. Some plugins need the controls to be there already when they're loaded, however those plugins could now wait for the `ready` event, whereas today there's no option for plugins that need to modify the controls ahead of time. 

I'd be interested to know @dmlap and @gkatsev's thoughts on this since they've had the most experience building plugins. For instance, if the ready event will work for that.

Confirming for 5.0 unless we hear differently. This would be a pretty big API breaking change for plugins so we would need to figure it out quickly if we want to get it in.

The dependencies issue is one we've been talking about off an on, and might be worth opening another issue for. We're running into it more and more. Basically, everything in the video.js ecosystem is a peer dependency, even video.js itself, and npm peer dependencies aren't so great, so it's been a challenge figuring out the best approach.

heff:
Closed by #2094

PR #2076, opened 2015-04-27
Load plugins before controls

bc-bbay:
For issue #1980


heff:
Awesome, thanks @bc-bbay. @carpasse want to take a look at this one?

I'm wondering if we should init plugins after the player element is created but before children are added.

mmcc:
Just a note, we definitely need to document that current plugins will need to wait for the `ready` event in order to see the same behavior they do today.

heff:
@bc-bbay, could we move it to between the el and the children? I think that makes more sense.
```
this.el_ = this.createEl();
// here
this.initChildren();
```

carpasse:
@heff and @bc-bbay that is exactly I proposed. It looks great.
Thanks

bc-bbay:
Made a new PR (accidentally pushed this change to my fork's master, and it was trying to sneak its way into other branches) with the relocated change.

PR #2077, opened 2015-04-27
Remove src/js/json.js, replace with safe-json-parse

### LINKED ISSUE #867: Remove JSON shim to avoid eval() CSP issues

dconnolly:
Use the tuple variant of safe-json-parse to shim JSON.parse where not
available without using eval() as json.js was before.

Fixes #867


gkatsev:
Sweet. Also, we probably want to have better handling of errors when they happen. Not sure if we want to take care of it in here or as part of another PR.

dconnolly:
Agreed on the error handling, this is just a simple swap out for now.

heff:
lgtm

heff:
Tests failed for a  non-obvious reason. Rerunning them.

Issue #2035, opened 2015-04-14
Determine API for externally subclassing components


heff:
With 5.0 we've switched to ES6 classes, but we still need a way for projects that aren't using ES6 to subclass components, e.g. techs.

The current proposal is `videojs.extends` to match the ES6 terminology. `videojs.inherits` could be an alias (if that makes sense) to match the Node naming.

heff:
Closed by #2078

PR #2078, opened 2015-04-28
Added and `extends` function for external subclassing

heff:
Addresses #2035 

It uses the babel _inherits under the hood, with the added benefit of defining additional methods. The external API will look like the following:

```js
 var Button = videojs.getComponent('Button');

 var MyButton = videojs.extends(Button, {
   constructor: function(player, options) {
     Button.call(this, player, options);
   },

  onClick: function() {
    // doSomething
  }
});
```

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -15,6 +15,8 @@ import * as setup from './setup';
 import Component from './component';
 import * as Lib from './lib';
 import * as Util from './util.js';
+import Player from './player';
+import extendsFn from './extends.js';
 
 if (typeof HTMLVideoElement === 'undefined') {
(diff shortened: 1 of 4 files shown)


src/js/video.js
```diff
@@ -39,6 +38,9 @@ videojs.util = Util;
 import Player from './player';
 videojs.players = Player.players;
 
+import extendsFn from './extends.js';
```
gkatsev: we should import everything at the top of the file.

heff: Updated. Everything else look ok?

gkatsev:
LGTM

PR #2081, opened 2015-04-28
Remove deprecated isDefaultPrevented.

### LINKED ISSUE #463: Switch event.isDefaultPrevented() to event.defaultPrevented to comply with spec
I think it'll be better if we follow the spec on this and change the method to know whether the event had it's default prevented to be the property `defaultPrevented` rather than requiring a method call of `isDefaultPrevented()`.

gkatsev:
Fix #463.


dconnolly:
:+1:

Issue #1444, opened 2014-08-22
Move to border-box as the default CSS setting for skin elements


heff:
We probably need to think through this a bit, but border-box can make certain things easier. It's not the default in most browsers though so it also might not be the best for making skinning easy, if most people are used to content-box.

@mmcc, @gkatsev

PR #2082, opened 2015-04-28
Switched to box-sizing:border-box for all player elements

heff:
This doesn't immediately break anything (I know of) because the controls
were originally designed to work in both border-box and content-box
environments. Extra elements were added to create internally padding.

This does open the door for some simplification of the HTML, though the extra
elements could still be useful for extra skin design options.

See #1444


dmlap:
LGTM

Issue #1204, opened 2014-05-14
Change the classname vjs-control to vjs-button


heff:
Originally 'control' was going to be the parent class of buttons and all other controls, but in practice only buttons use vjs-control. It would be a lot cleaner if we just vjs-button for button elements.

This would break existing skins so should wait for a major release. Though it's possible we could duplicate classes to help in the transition.

heff:
We discussed this a little on Friday, and the conclusion was that we need to switch to use the `<button>` element first, then decide what makes sense here classname wise. I think we should still have a `vis-button` class, and then decide if the `vjs-control` class is still necessary. It would be nice if the CSS class names matched the class inheritance.

dmlap:
Decided to add vjs-button instead of removing vjs-control

forbesjo:
Working on this.

heff:
Closed by #2084!

PR #2084, opened 2015-04-28
Added vjs-button class to button controls

forbesjo:
Fixe for issue #1204

### DIFF
--- a/src/css/components/_control.scss
+++ b/src/css/components/_control.scss
@@ -1,4 +1,5 @@
-.video-js .vjs-control {
+.video-js .vjs-control,
+.video-js .vjs-button {
   outline: none;
   position: relative;
   text-align: center;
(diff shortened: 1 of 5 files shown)


src/css/components/_control.scss
```diff
@@ -1,4 +1,5 @@
-.video-js .vjs-control {
+.video-js .vjs-control,
+.video-js .vjs-button {
```
heff: Actually, we can remove all of these vjs-button additions in the CSS. All of the elements already have the `vjs-control` class name, so all of those styles will already apply.

heff:
Made one comment but otherwise it looks good to me. The test failed for an unrelated reason so restarted them.

src/css/components/_control.scss
forbesjo: I'll remove them, I initially added them in case we remove vjs-control from
the buttons in the future.

heff: Cool, got it. That makes sense but we should do that in the same PR where we adjust the buttons.

heff:
Very nice. Lgtm!

PR #2088, opened 2015-04-29
Added scss files to dist folder

mmcc:
Also moved the `style-injection-task` to the build/tasks directory.

The cool thing about this is it allows someone with a Sass workflow to import our SCSS files directly into their own, so there doesn't need to be an additional build step. Dogfooding ftw.


gkatsev:
LGTM

heff:
@mmcc https://github.com/mmcc/video.js/pull/8

heff:
lgtm!

heff:
Also we've switched it to instead include the original source files in the npm package instead of copying them to dist.

PR #2095, opened 2015-04-29
Manual events cleanup

saxena-gaurav:


heff:
I there any way we could write a unit test for this that breaks in the old version and works with this new version?

I don't mean to drag this out but my one remaining issue with this is the addEventListeners naming is very generic and doesn't make it easy to understand what's happening. If I can see the error happening in code instead of a compiled demo I might be able to help clean it up some.

chemoish:
@heff this is from a while ago, any idea why it ended up being killed?

heff:
Sorry, I don't remember the details. There might have been some internal communication. Maybe @saxena-gaurav does.

chemoish:
The error we were tracking down ended up being in a plugin and there is an open pull request to address that.

https://github.com/benjipott/video.js-chromecast/pull/14

Issue #2034, opened 2015-04-14
Decide on a final var style


heff:
See https://github.com/videojs/video.js/pull/1993#discussion-diff-28028990

Now that we are using let in the codebase we can define variables where they're used, but there's still some open style questions that we should lock down.

--------------------------
TODO
* [ ] Update code to match videojs-standard
* [ ] Merge in videojs-standard checking
* [ ] Document videojs-standard and usage of the airbnb style guide.

dmlap:
I think [AirBnB's es6 style guide](https://github.com/airbnb/javascript/tree/es6) is really comprehensive and I don't hate a single thing in it (shocking!). Thoughts on adopting it wholesale for 5.0?

heff:
Yeah, I like the idea. It's grown a bit since the last time I looked at it so I need to review again, but I didn't see anything I couldn't live with. I think it's interesting they default to `const` for variables. [This issue](https://github.com/airbnb/javascript/tree/es6#references) has a few more details.

> - Why? This ensures that you can't reassign your references (mutation), which can lead to bugs and difficult to comprehend code.
> - If you must mutate references, use let instead of var.

I don't have any objections to that myself.

Whether it's AirBnB or another style guide, I love the idea of adopting an external public style guide, with zero exceptions. We could save a bunch of back and forth by shifting all responsibility and discussions around style to the external guide, not our projects.

gkatsev:
I generally like their style guide, though, some of the reasoning and terminology is weird.
There are two things I don't agree with, though:
* preferring function declarations over function expressions for their naming and hoisting - You can have named function expressions and relying on hoisting is bad.
* Not to use wildcard imports - I think this makes sense for modules that export an object or something concrete but for the "bag of functions" modules I think it's not as good. If no default export you can just do `import {foo, bar} from "./lib.js";` to just get what you need instead of just doing `import * as Lib from "./lib.js";` or `import Lib from "./lib.js";`

heff:
> relying on hoisting is bad

I can't find any discussions around this in the guide repo. Want to submit an issue and see what they respond with?

> but for the "bag of functions" modules I think it's not as good

It doesn't look like they're saying not to use `import {foo, bar} from "./lib.js";`, just literally the wildcard `import * as ...`. The example just above this note has `import { es6 } from './AirbnbStyleGuide';`.

gkatsev:
If they also talk about not using `import *` that's fine.

[...]

PR #2099, opened 2015-04-29
Add vjs standard

gkatsev:
So far, this adds in videojs-standard. `standard --format` is currently broken because it doesn't support the ES6 module syntax.

I'll be working towards making the changes to pass standard.
Addresses #2034.

TODO
* [ ] Remove jshint and others
* [ ] Update code
* [ ] ???
* [ ] Profit

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,6 +2,7 @@ module.exports = function(grunt) {
   var pkg = grunt.file.readJSON('package.json');
   var license = grunt.file.read('build/license-header.txt');
   var verParts = pkg.version.split('.');
+  var testBuildArray;
   var version = {
     full: pkg.version,
     major: verParts[0],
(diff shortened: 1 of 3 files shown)


heff:
@gkatsev how can we continue to make progress on this one?

forbesjo:
Would it be possible to do the same thing with just an eslint config?

gkatsev:
The idea is that you don't need to worry about copying the eslint config. You just `npm install -D videojs-standard` and then add it to your build.

forbesjo:
I guess I mean making a [sharable config](http://eslint.org/docs/developer-guide/shareable-configs.html).

gkatsev:
That still means that you would need to configure your app to use eslint. With stanard, you just install it and run it. No configurtion, no anything else.

PR #2103, opened 2015-04-30
Expose videojs.plugin()

### LINKED ISSUE #2102: 5.0.0-1 does not expose plugin() on videojs object.
```
> console.log(videojs);
2015-04-30 14:19:36.426 VM2029:2 videojs(id, options, ready)
undefined
> videojs.plugin('foo')
VM2031:2 Uncaught TypeError: videojs.plugin is not a function
```

This breaks existing plugins.

dconnolly:
So that plugins can be registered with the same api!  Also a unit test
module for the video.js entrypoint.

Fixes #2102

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -17,7 +17,7 @@ import * as Lib from './lib';
 import * as Util from './util.js';
 import Player from './player';
 import extendsFn from './extends.js';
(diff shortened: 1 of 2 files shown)


gkatsev:
LGTM

heff:
Very nice.

PR #2111, opened 2015-05-01
We want to call global setInterval not our own

gkatsev:
When we changed to concise methods, we ended up creating a new function
that shadowed the global setInterval, setTimeout, clearInterval, and
clearTimeout methods. A recent change in babel exposed this bug.


heff:
LGTM!

Issue #1998, opened 2015-04-02
Throw an error if tech isn't in options


mmcc:
If you try and unwittingly override all global options, you'll end up in a tech-less state and an unusable player (with none of the options you wanted). Right now this just silently fails, so we should consider throwing an error of some sort if there isn't a tech available in the options.

For example: 

```js
videojs.options = { width: 300, height: 150, controls: true, preload: auto };
videojs('sweet-vid'); // unusable player because all options have been overridden.
```

We should probably also actually document using global options correctly while we're at it.

gkatsev:
I feel like `videojs.options` shouldn't just be an object available to everything. It should be changed to be a setter/getter like components are.

forbesjo:
I'll look into this.

forbesjo:
Freeing this ticket up if someone else want to take it.

dmlap:
@gkatsev sounds like you're suggesting a more significant change. How about opening up a different issue to discuss?

gkatsev:
We want a `setGlobalOptions` and `getGlobalOptions` for setting the global options rather than just setting it on `videojs.options`. We can do the check in that function.

dmlap:
Got this one.

heff:
Closed by #2097

PR #2115, opened 2015-05-01
Add get and set global options methods

gkatsev:
The final part of #1998

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -45,7 +45,7 @@ videojs.extends = extendsFn;
 
 videojs.getGlobalOptions = () => options;
 videojs.setGlobalOptions = function(newOptions) {
-  Lib.obj.merge(options, newOptions);
+  Lib.util.mergeOptions(options, newOptions);
 };
 
 videojs.plugin = plugin;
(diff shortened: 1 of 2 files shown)


gkatsev:
Oh, I need to update the sandbox.

src/js/video.js
```diff
@@ -42,6 +43,11 @@ videojs.players = Player.players;
 
 videojs.extends = extendsFn;
 
+videojs.getGlobalOptions = () => options;
```
mmcc: If we're going to start using fat arrows, make sure to make a note in the ES6 page on the wiki.

mmcc:
:+1:

src/js/video.js
gkatsev: I'll add a note. I decided to use it here specifically because this function becomes so short as a fat arrow function.

heff:
LGTM. Can you also add a note to the 5.0 changes wiki? This is actually a pretty significant change considering it's used in most example embeds.

gkatsev:
I'll get to that soon.

PR #2127, opened 2015-05-06
Do not autoplay if no source defined

bc-bbay:


src/js/media/html5.js
```diff
@@ -76,7 +76,7 @@ vjs.Html5 = vjs.MediaTechController.extend({
     // In Chrome (15), if you have autoplay + a poster + no controls, the video gets hidden (but audio plays)
     // This fixes both issues. Need to wait for API, so it updates displays correctly
     player.ready(function(){
-      if (this.tag && this.options_['autoplay'] && this.paused()) {
+      if (this.src && this.tag && this.options_['autoplay'] && this.paused()) {
```
heff: `this.src` is a function (getter/setter) not a property, so this should be `this.src()`

bclwhitaker:
LGTM

dmlap:
Merged via 652a44026f12d47bc21e1a042a0a80e809c1a874

PR #2129, opened 2015-05-06
Unit test for dispose() when tech is changed.

saxena-gaurav:
add unit test to check dispose do not cause exception when tech is changed


heff:
Closing in favor of #2137

PR #2138, opened 2015-05-11
Added background-color to vjs-poster to remove transparent borders around scaled poster image

tjenkinson:
Same changes as pull request against master as per request. https://github.com/videojs/video.js/pull/1868#issuecomment-101025010


mmcc:
Thanks!

PR #2139, opened 2015-05-11
Break up utility functions into smaller libraries

heff:
Getting started on this one but interested in feedback early on.

I've started by just breaking up lib into libraries (e.g. dom.js), and maintaining the `Dom.createEl` library object property style. The next piece will be determining how to pull in individual functions instead of the entire library.

If you look at [lodash modularized](https://www.npmjs.com/browse/keyword/lodash-modularized) or [npm dom](https://github.com/npm-dom/domquery) the examples just have you name the function variable the same as simple function name.
```
var first = require('lodash.first');
```
I'm not super comfortable with that as a standard.
- It's not obvious what's internal vs. external to the file
- The function names get pretty generic and I think it could get confusing or even cause conflicts with other functions on the page.

An example on the second one is [Component createEl](https://github.com/videojs/video.js/blob/b65bad8f209d039d99f653a5c97485f652f07d02/src/js/component.js#L208). Babel creates named functions for class methods and I wonder if the createEl named function for the method would overwrite the createEl that's been imported.

My best thoughts are finding some common prefix for functions that are imported.
```
$createEl()
uCreateEl() // u for util
domCreateEl()
```
I'm not in love with any of those options but could live with any of them.

Anyone seen or used better options? Or are most people fine with the simple variable names.


heff:
Anyone know what might have changed with source maps recently? Travis is failing on them.

heff:
Fixed source maps error by locking down grunt-browserify version to 3.5.1 (it's now at 3.8.0). Probably not good long term.

(ping @forbesjo)

heff:
@videojs/core-committers can I get someone to glance at this? Now would be a good time to merge it in to prevent big merge conflicts. It's probably most valuable to look at the commit messages. The code itself didn't get changes all that much, just got moved around.

The next step is to switch the wildcard `import * as X` instances to pull in the specific functions. That could be another PR though.

src/js/utils/browser.js
```diff
@@ -0,0 +1,69 @@
+import document from 'global/document';
+import window from 'global/window';
+
+const USER_AGENT = window.navigator.userAgent;
```
mmcc: `export const USER_AGENT = ...`

src/js/button.js
```diff
@@ -1,7 +1,9 @@
 import Component from './component';
-import * as Lib from './lib';
+import * as Dom from './utils/dom.js';
 import * as Events from './events';
+import * as Fn from './utils/fn.js';
 import document from 'global/document';
+import assign from 'lodash.assign';
```
mmcc: `Object.assign`?

src/js/utils/to-title-case.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Uppercase the first letter of a string
+ * @param  {String} string String to be uppercased
+ * @return {String}
+ * @private
+ */
+const toTitleCase = function(string){
```
mmcc: `export default toTitleCase = ...`

src/js/utils/time-ranges.js
```diff
@@ -0,0 +1,21 @@
...
+ * Should create a fake TimeRange object
+ * Mimics an HTML5 time range instance, which has functions that
+ * return the start and end times for a range
+ * TimeRanges are returned by the buffered() method
+ * @param  {Number} start Start time in seconds
+ * @param  {Number} end   End time in seconds
+ * @return {Object}       Fake TimeRange object
+ * @private
+ */
+const createTimeRange = function(start, end){
```
mmcc: `export const createTimeRange = ...`

mmcc:
:+1: lgtm

src/js/button.js
heff: Babel doesn't appear to translate it to something that will work (and it's not ES5)

gkatsev: https://www.npmjs.com/package/object.assign

src/js/utils/browser.js
heff: USER_AGENT isn't used externally so I left it private.

src/js/button.js
heff: ah right, thanks!

heff: Switched to object.assign, not sure what the equivalent of `var assign = Object.assign || require('object.assign');` should look like with ES6 modules. Or if it matters to use the native one ever.

gkatsev: I guess we'd want to do something like:
```js
import objectAssign from 'object.assign';
let assign = Object.assign || objectAssign;
```
As it is now, it isn't bad, but we'll always be using the polyfill and never the native.

gkatsev:
Looking over this now. We can always make new PRs if there's anything we want to change.

As for the naming, we shouldn't do any special naming. From a functionality perspective, there shouldn't be a different between internal and external functions. However, I've seen styleguides that suggest (or require, eslint might be able to help with this, though, not sure) that the imports at the top of the file are grouped by where they're coming from.
I.e., the very top group of imports are built-ins, then come the group of imports from npm, and then come the imports that are local to the project.

package.json
```diff
@@ -24,8 +24,11 @@
   "style": "./dist/video-js.css",
   "dependencies": {
     "global": "^4.3.0",
+    "lodash.clonedeep": "^3.0.0",
+    "lodash.isplainobject": "^3.0.2",
+    "lodash.merge": "^3.2.1",
```
gkatsev: is this package used anymore?

src/js/button.js
```diff
@@ -1,7 +1,9 @@
 import Component from './component';
-import * as Lib from './lib';
-import * as Events from './events';
+import * as Dom from './utils/dom.js';
+import * as Events from './utils/events.js';
+import * as Fn from './utils/fn.js';
 import document from 'global/document';
+import assign from 'object.assign';
```
gkatsev: we probably should change this assign to be something like
```js
import objectAssign from 'object.assign';
let assign = Object.assign || objectAssign;
```
We could even have a small internal module that does this wrapping. This is because the npm module doesn't automatically change over to the built-in once available.

src/js/tracks/text-track-settings.js
```diff
@@ -122,7 +123,7 @@ class TextTrackSettings extends Component {
 
     let values = this.getValues();
     try {
-      if (!Lib.isEmpty(values)) {
+      if (Object.getOwnPropertyNames(values).length > 0) {
```
gkatsev: this is another one that would fail in IE8 unless we include es5-sham.

src/js/utils/browser.js
gkatsev: there are a few other uses of the user agent, should we change those to use `browser.USER_AGENT` instead, maybe?

src/js/xhr.js
```diff
@@ -1,5 +1,6 @@
-import * as VjsUtils from './util';
```
gkatsev: We totally should just get rid of this file and import `xhr` or another similar module from npm.

gkatsev:
Do we require es5-sham for other babelify things? If not, I think we should still make our own shims for `Object.create` and `Object.getOwnPropertyNames` so that only `es5-shim` is necessary.

Anyway, mostly minor stuff. Looks good!

src/js/button.js
gkatsev: Going to submit a PR to `object.assign` that will give us some way to grab the correct `assign` function.
Perhaps something like `import {assign} from 'object.assign'` as opposed to `import assign from 'object.assign'` but we won't need to do the above ourselves.

package.json
heff: Yeah, it's used in mergeOptions

src/js/button.js
heff: That'd be cool. I wonder though if it's even important that we use the native one ever, since we're including the module either way. I guess there could be performance benefits, but it also creates two different code paths that could (while unlikely) cause differences between browsers.

src/js/component.js
```diff
@@ -1181,7 +1190,7 @@ class Component {
     };
 
     // Inherit from this object's prototype
-    subObj.prototype = Lib.obj.create(this.prototype);
+    subObj.prototype = Object.create(this.prototype);
```
heff: It won't with the sham though, will it?

heff: If this will fail then so will the classes that get generated by babel, because it relies on Object.create too.

gkatsev: Yep, es5-sham should have it. 
Was wondering whether it was required so that maybe we could limit us to only have es5-shim.
https://github.com/videojs/video.js/pull/2139#issuecomment-103176907

src/js/xhr.js
heff: Yeah, I was thinking the same thing. We have some specific `file:///` url and `withCredentials` handling that we'll need to know how they translate or at least know we're losing them. But we should definitely make that happen before 5.0. Feel like making a PR for that?

src/js/button.js
gkatsev: The shim does actually check for issues and returns the shimmed version if so. So, there are benefits. But yeah, it's possible for the browser to super optimize things like `Object.assign`.

heff:
Yeah, so far the sham (as far as I know) is only needed for Object.create in the es6 classes. I'd also prefer to only need the shim, but I don't know how we'd make babel use our own Object.create shim.

gkatsev:
Yeah, if babel requires it, that's fine. We could theoretically make another transform that removes uses of `Object.create` and replaces it with a shim.

src/js/utils/url.js
```diff
@@ -0,0 +1,89 @@
+import document from 'global/document';
```
gkatsev: we should evaluate whether the built-in url module gives us everything we care about.

src/js/xhr.js
gkatsev: Probably a wrapper using the external module. Though, is `file://` support useful? `withCredential` support is available.

heff:
> make another transform that removes uses of Object.create and replaces it with a shim

Hmm...that's an interesting idea. I'm on the fence of whether or not it's worth it. We'd still need to bundle the es5 shim with the html5 video shim, so I can't tell if that would remove complication or add more. @mmcc @dmlap, any other opinions on whether we should go that route to avoid needing the sham?

src/js/xhr.js
heff: > Though, is file:// support useful

meh, probably not so much

dmlap:
I think I'd prefer a minimal shim for what we need if we can get away with it. I'm not an es5-sham expert though, so factor in this opinion is coming from a place of great ignorance.

PR #2142, opened 2015-05-12
Text descriptions track support

OwenEdwards:
Add support for text description tracks, and a demo in the sandbox.

By default, when a text description track is enabled, it is shown visually and also read out by a screen reader if the user has one. There is an option in the Captions/Descriptions Setting Menu to __hide__ the description track, so it is __only__ read by a screen reader. (Although note that the Settings Menu is not currently accessible by keyboard-only users, including screen reader users).

### DIFF
--- a/sandbox/descriptions.html.example
+++ b/sandbox/descriptions.html.example
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+  <title>Video.js Text Descriptions Example</title>
+
+  <link href="../build/files/video-js.css" rel="stylesheet" type="text/css">
+
+  <!-- LOAD VIDEO.JS SOURCE FILES IN ORDER -->
+  <script src="../build/source-loader.js"></script>
+
+</head>
+<body>
+  <p style="background-color:#eee; border: 1px solid #777; padding: 10px; font-size: .8em; line-height: 1.5em; font-family: Verdana, sans-serif;">This page demonstrates a text descriptions track (intended primarily for blind and visually impaired consumers of visual media)</p>
+
+  <!-- NOTE: we have to disable native Text Track support for the HTML5 tech,
+             since even HTML5 video players with native Text Track support
+             don't currently support 'description' text tracks in any
+             useful way! -->
+  <video id="example_video_1" class="video-js vjs-default-skin" controls preload="none" width="640" height="360"
+      data-setup='{ "html5" : { "nativeTextTracks" : false } }'>
+    <source src="https://archive.org/download/ElephantsDream/ed_hd.mp4" type="video/mp4">
+    <source src="https://archive.org/download/ElephantsDream/ed_hd.ogv" type="video/ogv">
+
+    <track kind="descriptions" src="elephantsdream.ad.vtt.example" srclang="en" label="English">
+
+    <p class="vjs-no-js">To view this video please enable JavaScript, and consider upgrading to a web browser that <a href="http://videojs.com/html5-video-support/" target="_blank">supports HTML5 video</a></p>
+  </video>
+</body>
+</html>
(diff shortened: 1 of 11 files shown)


OwenEdwards:
One thing that needs to be added is a proper icon in the icon font; see [Ryerson University's Universal Access Symbols](http://www.ryerson.ca/accessibility/tools/access_symbols.html), and the [ACB's Audio Description Project](http://www.acb.org/adp/ad.html#symbols)

gkatsev:
I assume the reason for the visible and hidden text track displays is because if we want to hide the descriptions track but still have it be read by a screen reader, we need to still write it to the DOM, but if we set a track's mode to `hidden`, it stops displaying it and only triggers events based on it.

sandbox/elephantsdream.ad.vtt.example
```diff
@@ -0,0 +1,278 @@
+WEBVTT
```
gkatsev: I think in the `build/` directory we have the sample files?

sandbox/descriptions.html.example
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
```
gkatsev: Not a comment for this PR, but can we just change all the sandbox files to drop the `.example` from them so they just work out of the box without requiring people to do anything? @heff

gkatsev:
This looks good to me, but it shouldn't be released as a patch update.

sandbox/elephantsdream.ad.vtt.example
OwenEdwards: I had them in the `build/demo-files` directory, but that didn't seem like the right place for them, particularly with that directory going away for V5. Sandbox seemed like the best place for now, but feel free to suggest somewhere else.

OwenEdwards:
Re visible vs. hidden text tracks, that's exactly it - they need to be rendered into the DOM in a way that is not visible to a sighted user, but is still read by a screen reader. See WebAIM's [Invisible Content Just for Screen Reader Users](http://webaim.org/techniques/css/invisiblecontent/).

sandbox/elephantsdream.ad.vtt.example
OwenEdwards: Actually, we should substitute [this file](http://html5videoguide.net/presentations/LCA_MM_Miniconf2013/elephant.audiodescriptions.vtt), which contains the attribution and license for this file from @silviapfeiffer.

heff: In 5.0/master this would go in the docs/examples/shared folder.

sandbox/descriptions.html.example
heff: If we did then changes made to them would get included in PRs, which would kind of defeat the point of the sandbox directory. The docs/examples directory is where we can have examples that work out of the box, but those should also probably be included in dist and use the dist files.

heff:
@OwenEdwards we've locked down the 4.x (stable) branch for only urgent changes, and are instead focusing on 5.0 master moving forward. Could we migrate this to be against master instead? Master looks a lot different now and I can help with anything that's not clear.

OwenEdwards:
I'm trying to port this over to 5.0, but VideoJS captions (as opposed to native browser captions) seem to be broken in v5.0.0-14. I'm using Firefox v37 on OS/X 10.10; the CC button appears, and the menu, but the captions do not appear when 'english' is selected. I *think* it's something to do with the captions settings (font etc) - I got them to appear a couple of times, but couldn't reproduce a sequence that makes them work consistently, and I don't have time to debug them fully. @gkatsev, any thoughts?

OwenEdwards:
I'll port this over to 5.0 once #2172 is fixed.

heff:
@OwenEdwards want to try porting this to 5.0 now?

gkatsev:
@heff he already has! https://github.com/videojs/video.js/pull/2328
Closing this on in favor of that one.

PR #2149, opened 2015-05-13
Clean up Sauce Labs browsers

forbesjo:
This change is to:
- Remove 'singleRun' duplicates
- Use latest default Sauce Labs browsers
- Not start Sauce Connect again, it should have started in .travis.yml before script


heff:
Awesome, thanks Joe. If we just left off the device and platform versions, would it use the latest ones? It'd be nice to not have to manually update this as versions increase.

forbesjo:
I simplified the browser capabilities some more, right now IE will default to Win7 - IE11 and Android will default to latest. Only iOS and Safari need versions or else they will default to iOS 5 and Safari for Windows.

I'm also not able to get the iOS and Android tests to run so I am skipping them for now.

heff:
Nice, thanks.

> they will default to iOS 5 and Safari for Windows.
:angry: 

> I'm also not able to get the iOS and Android tests to run so I am skipping them for now.
:disappointed: Yeah I saw Android pass once but Bootstrap skips them all together because the platform is flaky. I couldn't get iOS to run either. We should test on read devices to make sure there isn't a problem there.

forbesjo:
I bumped up the timeouts from 1min to 5 and got the Android tests to pass. The iOS tests at least start but seem to hang half way through. 

When I view the qunit page in the simulators/emulators on Sauce Labs they pass. Also just checked on real devices and they pass too

cvrebert:
The automated Sauce Labs test's screencast should give some indication of which individual testcase it's getting hung up on.

forbesjo:
Instead of dealing with karma being weird I decided to swap in grunt-saucelabs. The iOS and Android tests pass now but IE and Firefox on Linux are timing out. I'll see if I can figure out the issues.

heff:
> The iOS and Android tests pass now but IE and Firefox on Linux are timing out

e70639958fe1f60ffe3a17f643d2cad9 jpg

heff:
I just got an email that Sauce Connect 3 is being EOL'd for [Connect 4](http://sauceio.com/index.php/2014/02/meet-the-new-sauce-connect-4/). Are we already using 4?

Also I really hope their logo now includes this some how...
5103f3dhq0l _sy300_
(I want to feel like the kid on the left but right now Sauce makes me feel like the kid on the right)

forbesjo:
3.1 is being used in .travis.yml. We can have saucelabs-qunit run Sauce Connect 4 by setting tunneled to true. [Bootstrap](https://github.com/twbs/bootstrap) and [Winjs](https://github.com/winjs/winjs) do this.

heff:
Any guesses on if that will help with some of the issues we've seen?

forbesjo:
IE seems to be working now, I don't know if it is related to SC 4 though. Firefox on any platform refuses to work, I'll leave that for a future fix.

heff:
@forbesjo I asked @gkatsev to give a final review, then we'll pull in.

test/index.html
```diff
@@ -9,7 +9,33 @@
 <body>
   <div id="qunit"></div>
   <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
-
+  <script type="text/javascript">
```
gkatsev: What does this do exactly?

(Also, technically, the `type` isn't necessary in scripts anymore :)

gkatsev:
LGTM

test/index.html
forbesjo: This allows you to see the test results in Sauce Labs.
https://saucelabs.com/tests/68dcb9204fb94091996679fa2299f054

Should I just put this in globals-shim.js?

cvrebert:
Slightly tangential, but you might want to consider switching from Sauce Labs to a different Selenium provider. See https://github.com/theintern/intern/issues/394

forbesjo:
I also found Browserstack to be more stable. HLS videos seem to not work in Sauce Labs OS X but work fine in Browserstack so it might be something to look into.

heff:
How complicated would the switch be for us? @cvrebert have you tried out intern? Is that something we should look into using?

cvrebert:
I have not used intern personally, although it sounds promising.

PR #2168, opened 2015-05-19
Fixed all uses of registerComponent that relied on __proto__ inheritance

heff:
Didn't make the connection that the reason class method inheritance worked was the `__proto__` usage in babel... It breaks in IE8. Reverted those uses.


mmcc:
Assuming PAM doesn't yell at us, lgtm

PR #2173, opened 2015-05-19
incorrect spelling

stopspazzing:
Parallax was spelled incorrectly. This is a critical edit ofcourse.


heff:
Thanks! This has been removed in master and we're trying to avoid unnecessary patch releases on stable while we work on 5.0, but appreciate the attempt to help. If you're interested in helping out more check out the master branch.

PR #2177, opened 2015-05-19
Improved localisation

mister-ben:
Replaces #1873, which got messy after being stagnant so long...

* Try primary code before default string, e.g. if there's no `de-DE` translation, use `de` if available. 
* Treat language codes case insensitively. Safari uses `en-us` but Chrome and Firefox use `en-US`, these should be treated the same.
* Re-merge global and player languages so that `addLanguage()` is still effective for players that had languages set as a setup option. Player-local languages take precedence.

### DIFF
--- a/docs/guides/languages.md
+++ b/docs/guides/languages.md
@@ -37,7 +37,7 @@ Video.js uses key/value object dictionaries in JSON form. A sample dictionary fo
 ```
 
 Notes:
- 
+
 - The file name should always be in the format `XX.json`, where `XX` is the two letter value of the language reported to the browser (for options see the bottom of this document).
 - For automatic inclusion at build time, add your language file to the `/lang` directory (see 'Adding Languages to Video.js below').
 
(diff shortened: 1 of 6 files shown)


src/js/player.js
```diff
@@ -104,7 +104,17 @@ class Player extends Component {
...
-    this.languages_ = options['languages'] || Options['languages'];
+    if (options['languages']) {
+      // Normalise player option languages to lowercase
+      let languagesToLower = {};
+      Object.getOwnPropertyNames(options['languages']).forEach(function(value,index,array){
+        languagesToLower[value.toLowerCase()] = options['languages'][value];
+      });
+      this.languages_ = languagesToLower;
+    }
+    else {
```
dmlap: This `else` should be on the same line as the closing curly-brace of the preceding `if`.

docs/guides/languages.md
```diff
@@ -122,6 +122,24 @@ During a Video.js player instantiation you can force it to localize to a specifi
 </video>
 ```
 
+Determining Player Language
```
dmlap: Very nice addition.

dmlap:
I think there may be an additional whitespace issue in here (@gkatsev, how's that code formatting build step going?) but looks good to me.

src/js/player.js
```diff
@@ -104,7 +104,17 @@ class Player extends Component {
     this.language_ = options['language'] || Options['language'];
 
     // Update Supported Languages
-    this.languages_ = options['languages'] || Options['languages'];
+    if (options['languages']) {
+      // Normalise player option languages to lowercase
+      let languagesToLower = {};
+      Object.getOwnPropertyNames(options['languages']).forEach(function(value,index,array){
```
gkatsev: space after commas.

gkatsev:
It's coming along slowly. If someone wanted to run it locally, it's doable by installing `videojs-standard` as global and then running `standard` in the videojs folder. Should report all errors. Be aware it will also report warnings, but you can `echo $?` to find out the exit code to see whether it succeeded or failed. Though, error code won't help much until everything gets converted.

mister-ben:
Fixed those and other issues flagged by `videojs-standard`.

heff:
Suggested some changes here: https://github.com/mister-ben/video.js/pull/1

PR #2182, opened 2015-05-20
Clean up and documentation of src/js/video.js and DOM functions

heff:
- Standardized naming of dom functions in preparation of exporting on the `videojs` object.
- Cleaned up an documented a lot of the main video.js source file
- Fixed an issue with the new mergeOptions function. It no longer makes a clone of the first object, but does make new objects for empty keys that are being merged with, to match how lodash.merge works by default.


src/js/utils/dom.js
```diff
@@ -70,7 +68,7 @@ export const insertFirst = function(child, parent){
  * @type {Object}
  * @private
  */
-export const cache = {};
+const _data = {};
```
gkatsev: since this isn't exported, there's no need for the `_`.

gkatsev:
LGTM

src/js/utils/dom.js
heff: Cool. Is that the common practice? Add an underscore only to private vars that are going to be accessible externally?

gkatsev: Yep.

src/js/video.js
```diff
@@ -90,64 +84,235 @@ var videojs = function(id, options, ready){
...
+ * @return {Object} The resulting language dictionary object
  */
 videojs.addLanguage = function(code, data){
-  if(Options['languages'][code] !== undefined) {
-    Options['languages'][code] = mergeOptions(Options['languages'][code], data);
-  } else {
-    Options['languages'][code] = data;
-  }
-  return Options['languages'];
+  return merge(globalOptions.languages, { [code]: data })[code];
```
gkatsev: NICE

heff: tumblr_m2su7wfsrr1qhlnuu

Issue #1667, opened 2014-11-17
Consider making triggerReady async


gkatsev:
The internal function triggerReady should probably be made async itself, trigger the actual handlers should be sync.
There's already a place in the code-base where there are issues encountered from it: https://github.com/videojs/video.js/blob/caf44a4ad3a2b885bbff03fc9f092e231c57bbc3/src/js/tracks.js#L61-L69 and I've encountered the same issue with related tracks work.

heff:
@gkatsev if you want to respond to my comments here we could try to change this for 5.0. Otherwise it might wait awhile.

heff:
Had another thought on the performance issue here, and as long as the tech isn't reliant on the ready event, this shouldn't inhibit fast load. Confirming and adding to 5.0.

gkatsev:
Also, this would make the two techs work closer to each other.

dmlap:
Dissenting (but possibly unrealistic) opinion: HTML guarantees synchronous `<video>` readiness. I think we should be aiming for that level of simplicity for video.js users.

gkatsev:
If we can make flash load in synchronously, I'd be 100% for that. But whatever route we choose, it should be the same regardless of what tech is loaded.

heff:
Yeah wow, good point @dmlap, this is a conflict of probably our two biggest principles.
- The player should work just like the video element
- Techs should work the same as each other

If everything was synchronous then we wouldn't need a ready event. So I still think this is the right direction for this feature.

We still have #1728 and the web component direction where we might push harder on synchronicity, since you'll be working directly with an element. We already do, and can do more of a few things:
- Return default values for player properties before the tech is ready
- Queue method calls like `play`, that are async in nature, while the tech is loading

You can't do everything synchronously on the video element (e.g. set currentTime), so that helps.

PR #2188, opened 2015-05-22
Updated player.ready to always be async

heff:
Addresses #1667 

I've changed ready() and triggerReady() to always be async. However now a bunch of track tests are falling. 

@gkatsev, can you provide any insight here? I would guess it has to do with the text track display.
https://github.com/videojs/video.js/blob/1ff361a825400a2be803d6d4f0b72e0ecfbc8e6f/src/js/tracks/text-track-display.js#L41
```
Text Track Controls should be displayed when text tracks list is not empty FAILED
  control is displayed
  textTracks contains one item. Expected: 1 Actual: 0
```

If I move the addRemoteTextTrack part out of the ready listener the tests actually pass, but then other tracks tests fail.
```
Tracks TextTrackDisplay initializes tracks on player ready FAILED
  TypeError: Cannot read property 'tracks' of undefined
```

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -715,7 +715,8 @@ class Component {
   ready(fn) {
     if (fn) {
       if (this.isReady_) {
-        fn.call(this);
+        // Ensure function is always called asynchronously
+        this.setTimeout(fn, 1);
       } else {
         this.readyQueue_ = this.readyQueue_ || [];
         this.readyQueue_.push(fn);
@@ -732,20 +733,22 @@ class Component {
   triggerReady() {
     this.isReady_ = true;
 
-    let readyQueue = this.readyQueue_;
+    // Ensure ready is triggerd asynchronously
+    this.setTimeout(function(){
+      let readyQueue = this.readyQueue_;
 
-    if (readyQueue && readyQueue.length > 0) {
+      if (readyQueue && readyQueue.length > 0) {
+        readyQueue.forEach(function(fn){
+          fn.call(this);
+        }, this);
 
-      for (let i = 0; i < readyQueue.length; i++) {
-        readyQueue[i].call(this);
+        // Reset Ready Queue
+        this.readyQueue_ = [];
       }
 
-      // Reset Ready Queue
-      this.readyQueue_ = [];
-
-      // Allow for using event listeners also, in case you want to do something everytime a source is ready.
+      // Allow for using event listeners also
       this.trigger('ready');
-    }
+    }, 1);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


gkatsev:
It probably has to do with the workarounds I've done previously to get around the fact that html5 is sync and flash is async.

gkatsev:
Also, code-wise, LGTM

src/js/component.js
```diff
@@ -730,22 +731,30 @@ class Component {
...
+    this.setTimeout(function(){
+      let readyQueue = this.readyQueue_;
 
-    if (readyQueue && readyQueue.length > 0) {
+      if (readyQueue && readyQueue.length > 0) {
+        readyQueue.forEach(function(fn){
+          fn.call(this);
+        }, this);
 
-      for (let i = 0; i < readyQueue.length; i++) {
```
gkatsev: kill those for-loops with extreme prejudice! :D

gkatsev:
Also, didn't someone say that text tracks are generally broken in master, currently? Possibly related to that as well.

gkatsev:
It's possible that we can remove the `player.ready()` wrapped for tracks and then just remove the test that checks that tracks are created on `ready`. I think that was one of my workarounds for this.

gkatsev:
Ah, I know what the fix for the tests is. We need to actually trigger ready by ticking the clock forward.

gkatsev:
I have a PR against this that fixes the text track errors: https://github.com/heff/video.js/pull/9

heff:
Thanks! Nice use of the PR PR. :)

I tried that with one test but only ticked 1ms, which apparently isn't enough even though ready() waits 1ms. Guess it's safer to be more liberal with the time.

Tests are passing now. Attempting to remove the ready listener now.

heff:
Removed it and it's just the one test breaking now that tests specifically for that (like you said). Killing that test.

src/js/component.js
```diff
@@ -730,22 +731,30 @@ class Component {
    * @return {Component}
    */
   triggerReady() {
+
+    // // Ready should only be fired once
```
heff: @dmlap, I had this change in there but then decided to hold on it until we figured out how we wanted to handle tech changes. I forget where we landed, but also realized that the readyQueue does reset itself after ready is triggered, so ready listeners won't fire again on a tech change. That is, unless you listen for the ready event, not pass the function to ready(fn).

But basically the ready event was meant to let you add new ready listeners when the tech changes, it just may not be so clear that's the purpose. Maybe a `techloaded` event is the answer.

src/js/component.js
heff: :curly_loop: :gun: :scream_cat:

Issue #1913, opened 2015-02-28
Browser locks on player load while parsing all VTT files


ksheurs:
I'm finding on videos that have ~10 subtitles (various different languages) the entire browser locks while VideoJS goes through and parses all of them. I can see users trying to click to the play button and thinking the player is broken with no immediate playback/feedback.

Is it necessary to parse all these on load vs 1x1 as the user selects a particular language? If not, is there any way to do this in a non blocking fashion?

We are using VideoJS on a premium video service. I can't see putting this into production as is. If this the way it will behave for foreseeable future, is the only other option `video.novtt.js`?

ksheurs:
I'm finding this pre-fetching of all tracks + browser locking only happens on the Flash tech FYI. HTML5 lazy loads them per selection.

There is about ~10 seconds of no interaction ability with the player. I also get a few `VIDEOJS: ERROR: ParsingError` on Flash tech, but not HTML5. After all have loaded, I am able to play the video and all tracks work.

mmcc:
Do you have an example player that you can point us to so we can take a look? I assume so, but are you seeing this with *any* browser loading the Flash tech?

ksheurs:
Getting a demo together now @mmcc. Will report back soon. Thank you!

ksheurs:
Hi @mmcc. I've put together a reduced test case here: http://jsbin.com/semitutufo/1/. These are tracks for a different (feature length) film, but it demonstrates the bug.

You can see in Chrome (HTML5 tech) the player is responsive and tracks are loaded as needed.

In Firefox (Flash tech) the browser almost crashes as it tries to load and parse all the tracks on load.

mmcc:
Doesn't look like this is a Flash tech issue, seems to be a browser issue. Firefox doesn't need the Flash tech to play the MP4 source, but I'm still seeing the player go unresponsive for a moment.

mmcc:
By the way, thanks for the reduced test case, @ksheurs. Don't really have much else to say yet, but at least it's useful to know that this doesn't seem to be tech specific.

gkatsev:
It sounds like this issue has to do with the emulated text tracks which always run in Firefox right now because of a [known issue](https://github.com/videojs/video.js/issues/1862).
Even though we're making the XHRs for downloading all the files asynchronously, it might be that they all come back at once. There's probably some work that needs to happen to figure out specifically why the slow down happens since everything is done asynchronously, so, the browser should have time to get in between each request coming back.
[...]

PR #2192, opened 2015-05-22
Load text tracks on demand.

gkatsev:
This PR is also a place for discussion around how we want to handle tracks.
Currently, I've changed it so that tracks get loaded on demand, i.e., when you select them to display, and not all at once when the player is loaded. `default` tracks get loaded on player creation, still.
The menu should still be showing up because the TextTrack object is being created, just that the cues aren't being populated.
Also, this might be causing issues with chapters as it is right now, but could be fixed by loading those initially along with the default tracks.

This fixex #1913 and related issues.

@dmlap @heff

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -112,7 +112,7 @@ vjs.TextTrack = function(options) {
 	loadTrack(this.src_, this);
       }
 
-      if (mode === 'showing') {
+      if (mode !== 'disabled') {
         this.player_.on('timeupdate', timeupdateHandler);
       }
       this.trigger('modechange');
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -179,7 +179,7 @@ vjs.TextTrack = function(options) {
   if (options.src) {
     tt.loaded_ = true;
     tt.src_ = options.src;
-    if (options.default) {
+    if (options['default']) {
       loadTrack(options.src, tt);
     }
   } else {
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -107,6 +107,11 @@ vjs.TextTrack = function(options) {
         return;
       }
       mode = newMode;
+
+      if (mode !== 'disabled' && this['cues'].length === 0) {
+	loadTrack(this.src_, this);
+      }
+
       if (mode === 'showing') {
         this.player_.on('timeupdate', timeupdateHandler);
       }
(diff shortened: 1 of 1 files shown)


heff:
Not to dig into unrelated things, but does this line do something? https://github.com/videojs/video.js/blob/69a4884880dc3baaacd91ea23194b46dd7b48f53/src/js/tracks/text-track.js#L63

gkatsev:
It does. It causes the [`activeCues` getter](https://github.com/videojs/video.js/blob/69a4884880dc3baaacd91ea23194b46dd7b48f53/src/js/tracks/text-track.js#L129-L170) to trigger which swaps out the cues in and out of the activeCues array

src/js/tracks/text-track.js
```diff
@@ -107,6 +107,11 @@ vjs.TextTrack = function(options) {
         return;
       }
       mode = newMode;
+
+      if (mode !== 'disabled' && this['cues'].length === 0) {
+	loadTrack(this.src_, this);
+      }
+
       if (mode === 'showing') {
```
heff: This should include hidden tracks too so cuechange events fire for those also. Would that mess anything up?

gkatsev: It should.

heff:
Looks good to me. What are the open questions around how we want to handle tracks?

gkatsev:
Specifically, this is a change in functionality. Previously, all tracks and their cues were available after load and now they won't be available until they're selected. Possibly, we want to be able to load all the tracks sequentially starting with the default tracks (and the selected tracks) rather than only loading tracks on-demand.
Also, as I mentioned, chapters should be verified before pulling this in.

heff:
Ok, so it ads an asynchronous step for each track as opposed to all up front. Do we know how the video element handles this by default?

gkatsev:
I think it does load them all up as soon as possible, though, haven't tried it yet.

gkatsev:
Actually, looks like chrome doesn't have any of the cues available until the track is shown and chrome only uses either the "best match", top/first, or the default track. So, actually, this won't be too dissimilar.

heff:
Nice. That's good. In Chrome are the cues available synchronously after showing a track?

gkatsev:
Doesn't seem to be the case in chrome. Btw, here's a test page modified from #1913: http://output.jsbin.com/rubacujite/1/edit?html,output

gkatsev:
Ugh, this has turned into a much worse thing to fix than I thought :)
With the files from #1913 loading in more than one item on init causes the video player to more or less lock up. So, we should need to add some event to know when a particular track has finished and loop through the items. Basically, #1949.
I'll look into merging the two PRs. I'll probably make a new PR for it.

gkatsev:
I'm going to put this one on ice for a bit. The main issue with the really large text tracks from #1913 is that vttjs takes too long to parse (mozilla/vttjs#340). Hopefully, I'll have some time later to look into that but don't have it immediately.

gkatsev:
Going to close this PR. We can revisit it after the vttjs issue has been fixed.

Issue #2185, opened 2015-05-21
IE8 crashing after activity following inactivity


mmcc:
While testing #2132 in IE8, we noticed that user activity following no activity would cause the browser to crash and reload in compatibility mode. 

Very little debugging has been done so far, just wanted to open this as a tracking issue / reminder. With any luck, fixing this will also fix the inactivity timeout flicker we've been seeing.

heff:
This one may be fixed. Need someone to confirm whether or not they're seeing it.

mmcc:
Looks like it's fixed!

![Success](http://media.giphy.com/media/2XskdWAKNVppqYWF5W8/giphy.gif)

heff:
lol

PR #2206, opened 2015-05-28
Fix IE8 issues from 5.0 changes

### LINKED ISSUE #2184: timeupdate not firing for Flash tech
We noticed this while testing in IE8, but it's reproducible in modern browsers when using the Flash tech. Everything seems to work fine, including seeking via the play progress handle, but timeupdate never fires.

heff:
- Fixed an issue using lodash.isPlainObject on elements in IE8 (reverted to old isPlain)
- Turned on loose mode for Babel and added ES5 shim and sham to the examples
- Fixed the Flash tech's events by creating the object in component.createEl (otherwise a default 'div' was created) 

Fixes #2184
*Might fix* #2185 - I don't know why but I'm not seeing it at the moment

@eXon mind checking out the Flash tech bit?


eXon:
@heff LGTM, no wonder Flash was crashing everywhere if it wasn't using createEl, thanks for fixing it faster than me :)

heff:
Awesome, thanks @eXon. Pulling in.

Issue #2155, opened 2015-05-15
Keyboard accessibility of control bar menus (e.g. captions, subtitles, etc)


OwenEdwards:
In #1932, @heff wrote:

>we already support both spacebar and enter on our div buttons

That's true for buttons, but the way that menus are handled means that they are not keyboard accessible. Click propagation is prevented through the Menu component, but keyboard interaction is not, so, for example, pressing Enter on the captions button effectively clicks on the top menu item (often 'captions setting'), and pressing Enter on one of the other menu items doesn't behave correctly. (Actually clicking on the menu button, rather than a menu item, seems to select the top menu item, which is a weird behavior).

Menus ought to work like [W3C 'using ARIA menus'](http://www.w3.org/WAI/GL/wiki/Using_ARIA_menus), and the [OAA Accessibility Menubar example](http://oaa-accessibility.org/examplep/menubar1/). See also [Recommended WAI-ARIA implementation for navigation bar/menu on StackOverflow](http://stackoverflow.com/questions/12279113/recommended-wai-aria-implementation-for-navigation-bar-menu).

OwenEdwards:
Note that we're talking about several distinct menus here rather than a 'menubar'. The biggest change needed is to support keyboard navigation of the menu widget's items (Up arrow, Down arrow, Enter, Space, Escape).

The next step would be to pass extra information when adding menu items so that the items can be distinguished with ARIA (and, potentially, in the CSS/skin) as either the [title of the menu](https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L54) (which is a little redundant right now, since the menu title is handled as both the [button name](https://github.com/videojs/video.js/blob/master/src/js/button.js#L41) and also as a specific [ARIA label](https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/captions-button.js#L14)), as one of a set of effectively radio buttons (e.g. different languages of captions, of which only one can ever be 'select'ed), or as a [control which brings up another control](https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/captions-button.js#L41).

OwenEdwards:
Another great resource is http://hanshillen.github.io/jqtest/#goto_menubar. Also, see http://hanshillen.github.io/jqtest/#goto_button for a combination of buttons and a 'button group'; the video.js Control Bar is a similar mix of buttons and other control widgets.

gkatsev:
Fixed by #3033.

PR #2209, opened 2015-05-29
Buttons

mmcc:
This PR switches to using buttons instead of divs, as discussed in #1499. Just a note, this PR does *not* switch menu buttons over since that would require a substantial amount of refactoring from how things currently work. We plan on making that switch in a future minor release.

### DIFF
--- a/src/css/components/_big-play.scss
+++ b/src/css/components/_big-play.scss
@@ -23,12 +23,7 @@
   // Since the big play button doesn't inherit from vjs-control, we need to specify a bit more than
   // other buttons for the icon.
   &:before {
-    position: absolute;
-    top: 0;
-    left: 0;
-    width: 100%;
-    height: 100%;
-    text-align: center;
+    @extend %icon-default
   }
 }
 
--- a/src/css/components/_big-play.scss
+++ b/src/css/components/_big-play.scss
@@ -2,7 +2,7 @@
   font-size: 3em;
   line-height: 1.5em;
   height: 1.5em;
-  width: 3em; // Firefox bug
+  width: 3em; // Firefox bug: For some reason without width the icon wouldn't show up. Switched to using width and removed padding.
   display: block;
   z-index: 2;
   position: absolute;
--- a/src/css/components/_big-play.scss
+++ b/src/css/components/_big-play.scss
@@ -1,12 +1,14 @@
 .video-js .vjs-big-play-button {
   font-size: 3em;
-  line-height: 1em;
+  line-height: 1.5em;
+  height: 1.5em;
+  width: 3em; // Firefox bug
   display: block;
   z-index: 2;
   position: absolute;
   top: 10px;
   left: 10px;
-  padding: 0.2em 1em;
+  padding: 0;
   cursor: pointer;
   opacity: 1;
   border: 2px solid $primary-text;
@@ -17,6 +19,17 @@
   @include transition(all 0.4s);
 
   @extend .vjs-icon-play;
+
+  // Since the big play button doesn't inherit from vjs-control, we need to specify a bit more than
+  // other buttons for the icon.
+  &:before {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    text-align: center;
+  }
 }
 
 /* Hide if controls are disabled, the video is playing, or native controls are used. */
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -27,7 +27,7 @@ class Button extends Component {
     this.on('blur', this.handleBlur);
   }
 
-  createEl(type, props) {
+  createEl(type='button', props={}) {
     // Add standard Aria and Tabindex info
     props = assign({
       className: this.buildCSSClass(),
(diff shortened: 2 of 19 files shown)


src/css/components/_big-play.scss
```diff
@@ -1,12 +1,14 @@
 .video-js .vjs-big-play-button {
   font-size: 3em;
-  line-height: 1em;
+  line-height: 1.5em;
+  height: 1.5em;
+  width: 3em; // Firefox bug
```
heff: Add a more descriptive comment about the bug, please

src/css/components/_big-play.scss
```diff
@@ -17,6 +19,17 @@
   @include transition(all 0.4s);
 
   @extend .vjs-icon-play;
+
+  // Since the big play button doesn't inherit from vjs-control, we need to specify a bit more than
```
heff: We should think about how we can remove this duplication with a shared class or something.

src/js/button.js
```diff
@@ -38,24 +38,27 @@ class Button extends Component {
...
-      this.contentEl_.appendChild(this.controlText_);
-      el.appendChild(this.contentEl_);
-    }
+    this.controlText();
 
     return el;
   }
 
+  controlText(text) {
+    if (!text && (this.controlTextEl_ && this.controlTextEl_.innerHTML)) return this.controlText_ || 'Need Text';
```
heff: Note: Change this to a simple setter/getter and pass the text into the controlText() call above.

src/js/control-bar/mute-toggle.js
```diff
@@ -30,13 +30,6 @@ class MuteToggle extends Button {
     });
   }
 
-  createEl() {
-    return super.createEl('div', {
-      className: this.buildCSSClass(),
-      innerHTML: `<div><span class="vjs-control-text">${this.localize('Mute')}</span></div>`
```
heff: After removing this, I don't see anything that adds the control text on init. Update may not get called immediately.

src/js/menu/menu-button.js
```diff
@@ -76,9 +76,24 @@ class MenuButton extends Button {
...
   buildCSSClass() {
-    return `${this.className} vjs-menu-button ${super.buildCSSClass()}`;
+    return `vjs-menu-button ${super.buildCSSClass()}`;
+  }
+
+  controlText(text) {
+    if (!text) return super.controlText();
+
+    this.controlText_ = text;
+    this.el().children[0].children[0].innerHTML = this.controlText_;
```
heff: Can we clean this up at all?

heff:
Made some comments but otherwise lgtm.

src/js/menu/menu-button.js
mmcc: This whole thing can go away since we could just override `this.controlTextEl_` in `MenuButton` if we needed to change the location.

src/js/control-bar/mute-toggle.js
mmcc: I'm not sure if we cleared this up IRL, but `controlText_` is set to "Mute" on the `MuteToggle` class, which is set via `controlText()` in button.

heff:
@OwenEdwards any thoughts on how "menu buttons" should be treated? e.g. the captions menu. Should they be switched to buttons too or another element?

OwenEdwards:
Let's start by renaming 'menu button' to 'menu widget'! Semantically, this thing is a control widget with much more functionality than a button (but see below). Then, see my comments on #2155.

I want to clarify a point here though about the menus, which is about usability as much as accessibility:

Currently, the 'menu button' code is designed to work as a toggle button, which displays or hides the menu when [clicked](https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L99) or activated by the [keyboard (Space, Enter and Escape are supported)](https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L114).

However, in the CSS/skin, the [menu is displayed on hover](https://github.com/videojs/video.js/blob/master/src/css/components/_menu.scss#L31), so clicking on the menu button only toggles focus (highlight) onto / off the top item in the menu, which (to me at least) is confusing, because it's not actually *doing* anything to the video player, but there's a visual change.

When using the keyboard to control the 'menu button' (Tab to the control), pressing Enter or Space not only displays the menu, but also activates the top item on the menu, so the **keyboard access is broken**.

We need to make a decision on what we want clicking on the menu button to do for mouse/touch users, and whether we want that behavior to be changeable with a skin. Note that it's okay to have 'display menu on hover' and still maintain keyboard accessibility (see http://oaa-accessibility.org/examplep/menubar1/). Then, we need to fix keyboard access to menus, per #2155.

gkatsev:
The quick fix to making menu buttons accessible is making the menus show up on `:focus` and `:active` as well as on `:hover`. In addition, we'd need to make the next `<tab>` key go into menu items.

OwenEdwards:
`:focus` would be a quick fix to bring up the menu with keyboar navigation of the controls, but wouldn't resolve the issues of:

* What should clicking on or keyboard activating the main button of the menu do?

* Fixing how keyboard activating the main button of the menu also activates the top item in the menu

* Menu items should not be in the `<tab>` order, otherwise a long list of possible caption/subtitle languages would take a very long number of `<tab>` presses to get through to the next control widget. `<tab>` should go from control widget to control widget, `<enter>`, `<space>`, `<escape>`, and the arrow keys should interact with the widget.

PR #2216, opened 2015-05-29
increase progress holder / play handle size on hover

mmcc:
The seek bar is a bit skinny as it is, so this increases the size on hover.


heff:
Looks good to me. We discussed a potential change to `.vjs-progress-holder:hover > * { font-size: x }` but it's not required.

PR #2218, opened 2015-05-29
added IE10 flex support check

### LINKED ISSUE #2212: v5.0.0-25: Progress bar sits on top of control bar in IE10
All other versions of IE, Chrome, and FF do not exhibit this issue. This looks to be IE10 specifc

screen shot 2015-05-29 at 11 35 05 am 2

mmcc:
IE10 only implements the 2012 version of the Flex spec, which doesn't support `flex-basis`. Because of this, we were left in a weird middle ground where IE10 was getting the `no-flex` class but overriding the no-flex table styles with flex things. Closes #2212.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2248,10 +2248,13 @@ Player.prototype.handleError;
 Player.prototype.flexNotSupported_ = function() {
   var elem = document.createElement('i');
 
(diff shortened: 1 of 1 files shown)


heff:
lgtm!

PR #2219, opened 2015-05-30
One font-size transition

mmcc:
The grow transition was herky jerky in Chrome. Even though the handle is a bit too big for my tastes, it looks fine and overall it's better than the weirdness involved with two sizes.


heff:
lgtm!

PR #2220, opened 2015-05-30
added display: none; to fix the custom spacer in ie8

mmcc:
The spacer, as a table-cell, seemed to be defaulting to width: auto


heff:
lgtm

PR #2223, opened 2015-06-01
Move fonts out of core repo

mmcc:
There's now a font build pipeline: [videojs-font](http://github.com/videojs/font)


gkatsev:
You mean https://github.com/videojs/font?

gkatsev:
Ugh, the diff for the grunt.js file is horrible.

mmcc:
Ah yeah, forgot to revert the formatting change. I have my editor set to
auto format on save and meant to disable it before messing with core.

src/css/_variables.scss
```diff
@@ -4,7 +4,6 @@ $secondary-text: #F4A460; // currently just used for visited links in errors and
 $primary-bg: #000;
 $secondary-bg: lighten($primary-bg, 35%);
 
-$icon-font-family: 'VideoJS';
```
heff: Should we not still set this here anyway? How might someone override it with their own? If that's a real use case.

mmcc: If someone was overriding it with there own, they'd probably want to override all of the icons partial basically (at least that was my thought).

heff:
Should be good to go after the grunt file cleanup.

src/css/_variables.scss
heff: yeah, that makes sense

heff:
Looks good. Pulling in.

PR #2229, opened 2015-06-03
Deprecate options

mmcc:
- Added deprecation notice to `Component#options()`
- Switch to passing the parent video options to each child as `parentOptions`
- Use `parentOptions` from `this.options_` throughout tests / code base

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -186,6 +186,7 @@ class Component {
    * @return {Object}     A NEW object of this.options_ and obj merged
    */
   options(obj) {
+    log.warn('this.options() has been deprecated and will be moved to the constructor in 6.0');
     if (!obj) {
       return this.options_;
     }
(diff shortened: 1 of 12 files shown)


src/js/component.js
```diff
@@ -186,6 +186,8 @@ class Component {
    * @return {Object}     A NEW object of this.options_ and obj merged
    */
   options(obj) {
+    log.warn('this.options() has been deprecated and will be moved to the constructor in 6.0');
```
heff: We might want to consider bringing in [util-deprecate](https://www.npmjs.com/package/util-deprecate) so warnings are only logged once. Not required for this PR.

src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
```diff
@@ -76,11 +77,15 @@ class PlaybackRateMenuButton extends MenuButton {
     this.player().playbackRate(newRate);
   }
 
+  playbackRates() {
+    return (this.options_.playerOptions && this.options_.playerOptions['playbackRates']) || this.options_['playbackRates'];
```
heff: note: reverse this

src/js/player.js
```diff
@@ -151,6 +151,11 @@ class Player extends Component {
       }, this);
```
heff: Note: let's add player options to plugin options also, for the same reasons

heff:
A few small notes but this looks good to me.

PR #2230, opened 2015-06-03
Removed all the CDN logic

heff:
And moved it into videojs/cdn#3


package.json
```diff
@@ -27,6 +27,7 @@
     "lodash-compat": "^3.9.3",
     "object.assign": "^2.0.1",
     "safe-json-parse": "^4.0.0",
+    "videojs-font": "1.1.0",
```
mmcc: Did you mean to make this a hard requirement? We should relax this so it will pull the most recent version.

gkatsev: `^1.1.0`

heff: I originally locked it down as a convenience for the CDN stuff, but we talked offline and decided to lock it down for better reasons.
- We want to be able to call out specific font changes in the video.js change log, or at least have the option to.
- At the CDN level we can't use the auto-updating font because we also base64 encode fonts into the CSS, so the versions would get mismatched. Also the CDN has to point to a specific hosted version of the font, so it needs some way to know the specific font version that was used. It can't get that from a flexible version number.

heff: Also, should we be more careful with the compiled dependencies? Lodash I feel safe enough with, but for the others the carat feels pretty trusting of the other project authors and npm to not make any mistakes.

mmcc:
lgtm

mmcc:
(travis failure was `vjs_getproperty`)

PR #2232, opened 2015-06-03
Return currentSource_.src instead of blob-urls in html5 tech

imbcmdth:
The Flash tech already does this and it makes sense to return the URL of the resource instead of a blob-url if the SourceHandler uses MSE.

### DIFF
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -513,7 +513,7 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
 
     // Set currentSource_ asynchronously to simulate the media element's
     // asynchronous execution of the `resource selection algorithm`
-    setTimeout(function () {
+    this.setTimeout(function () {
       if (source && source.src !== '') {
         tech.currentSource_ = source;
       }
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -495,7 +495,6 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
    */
   Tech.prototype.setSource = function(source){
     var sh = Tech.selectSourceHandler(source);
-    var tech = this;
 
     if (!sh) {
       // Fall back to a native source hander when unsupported sources are
@@ -513,11 +512,11 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
 
     // Set currentSource_ asynchronously to simulate the media element's
     // asynchronous execution of the `resource selection algorithm`
-    this.setTimeout(function () {
+    this.setTimeout(vjs.bind(this, function () {
       if (source && source.src !== '') {
-        tech.currentSource_ = source;
+        this.currentSource_ = source;
       }
-    }, 0);
+    }), 0);
 
     this.sourceHandler_ = sh.handleSource(source, this);
     this.on('dispose', this.disposeSourceHandler);
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -495,6 +495,7 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
    */
   Tech.prototype.setSource = function(source){
     var sh = Tech.selectSourceHandler(source);
+    var tech = this;
 
     if (!sh) {
       // Fall back to a native source hander when unsupported sources are
@@ -510,7 +511,14 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
     this.disposeSourceHandler();
     this.off('dispose', this.disposeSourceHandler);
 
-    this.currentSource_ = source;
+    // Set currentSource_ asynchronously to simulate the media element's
+    // asynchronous execution of the `resource selection algorithm`
+    setTimeout(function () {
+      if (source && source.src !== '') {
+        tech.currentSource_ = source;
+      }
+    }, 0);
+
     this.sourceHandler_ = sh.handleSource(source, this);
     this.on('dispose', this.disposeSourceHandler);
 
--- a/test/unit/media.js
+++ b/test/unit/media.js
@@ -221,6 +221,10 @@ test('should add the source hanlder interface to a tech', function(){
 
   // Pass a source through the source handler process of a tech instance
   tech.setSource(sourceA);
+
+  // Increment clock since currentSource_ is set asynchronously
+  this.clock.tick(1);
+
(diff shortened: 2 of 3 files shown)


heff:
Awesome. 

I was worried we'd have a problem switching to other source handlers an not clearing out the previous currentSource_, but it looks like we [always set currentSrc_](https://github.com/videojs/video.js/blob/dac91a7b7bb92b283015da62a52a2f2f3a709c46/src/js/tech/tech.js#L423), so the `else` wouldn't only be hit when no source has been set.

This does create  a difference from the video element we might want to think about. The video element's currentSrc does't return the source that's "next". When you set `videoEl.src = 'something'` and *don't* call load(), `videoEl.currrentSrc` won't return `something`, whereas I think this new code will.
http://jsbin.com/jupuwituri/2/edit?js,console,output

It's weird, but I think that's the whole reason currentSrc exists instead of just relying on src. See any easy ways to make that work the same?

imbcmdth:
Gary and I came up with an interesting test: http://jsbin.com/sepajuyada/4/edit

Seems like the currentSrc property is always set but is done so asynchronously when the "resource selection algorithm" is executed.

heff:
Interesting, good find. I always assumed it waited for load(), but clearly not.

I remembered the other reason currentSrc exists, which is that you can set the source via the src attribute or the source child elements. vidEl.src should always return the src attribute value, which will be an empty string if the source was set with children. At the tech level we always set the source with the src property not children, so that's not really a problem for us.

So we could either assume we have more info than the video element in this case and return the set source as currentSrc immediately, or try to mimic the asynchronous side of it so it matches exactly with how the video element works, returning the previous source for any synchronous calls to currentSrc. Any other thoughts?

I added on to the jsbin to test some things. You can comment out the source child block to see how that works a little differently.
http://jsbin.com/wulomalopu/1/edit?js,console,output

Some other notes...

When you set an empty string as a source, you'll see there's some differences there. The src attribute will return the absolute URL of the page, while currentSrc will return an empty string in Firefox and Safari, or the previous source URL in Chrome. Go figure.

It looks like the loadstart event should fire [before currentSrc is updated](http://www.w3.org/html/wg/drafts/html/master/semantics.html#loading-the-media-resource:queue-a-task-4), but the browsers are either ignoring that or setting it in parallel before the listener gets to it.

imbcmdth:
I tried to capture the asynchronous behavior that you mentioned in a test and made some changes to media to emulate that behavior. I think I may have gone too far with the way I handle src when it is an empty string. It's probably best to just allow currentSrc_ to be set to an empty string as well since that seems to be the majority behavior.

src/js/media/media.js
```diff
@@ -510,7 +511,14 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
...
     this.off('dispose', this.disposeSourceHandler);
 
-    this.currentSource_ = source;
+    // Set currentSource_ asynchronously to simulate the media element's
+    // asynchronous execution of the `resource selection algorithm`
+    setTimeout(function () {
+      if (source && source.src !== '') {
+        tech.currentSource_ = source;
+      }
+    }, 0);
```
heff: Very nice. If you use `this.setTimeout` it will automatically kill the timeout when the tech is disposed. Just to clean up.

Adding async always makes me worried about edge cases we're not thinking of... I can't think of anything specific though so we could pull this in and see how it goes. @videojs/core-committers any final opinions on this (setting currentSrc asynchronously or not)?

test/unit/media.js
```diff
@@ -250,4 +254,54 @@ test('should handle unsupported sources with the source hanlder API', function()
...
+    handleSource: function(s, t){return {};}
+  };
+
+  Tech.registerSourceHandler(handler);
+
+  // Pass a source through the source handler process of a tech instance
+  tech.setSource(sourceA);
+
+  // Test that currentSource_ is not immediately specified
+  strictEqual(tech.currentSource_, undefined, 'sourceA was not stored immediately');
```
dmlap: Should we be testing `tech.currentSrc()` here instead? That seems closer to what we're trying to achieve.

src/js/media/media.js
dmlap: If this matches `<video>` behavior better, I think it's the way to go. +1 to `this.setTimeout`.

dmlap:
Two small comments but otherwise LGTM

test/unit/media.js
imbcmdth: The `currentSrc` function is only provided on html5 and flash techs. It isn't a base function on the MediaTechController. This is testing that MediaTechController does what it should to allow higher-level techs to operate as closely to the media element as possible.

dmlap: Hmm. Given all the state to implement `currentSrc()` lives in MediaTechController and both the html and flash tech have copy-pasted implementations, it seems like we should just pull the implementation up and make this a real test case.

imbcmdth: The current implementation is backwards compatible with existing techs. Moving anything to MediaController either doesn't buy us code savings or requires potentially breaking changes which we shouldn't be making in VJS4.

I agree that for VJS5 we should strongly consider moving much the `currentSrc` function to `Tech` and have each child-tech implement something like a `currentSrc_` function that `Tech#currentSrc` will look for and call if `currentSource_` is not defined.

src/js/media/media.js
```diff
@@ -510,7 +511,14 @@ vjs.MediaTechController.withSourceHandlers = function(Tech){
     this.disposeSourceHandler();
     this.off('dispose', this.disposeSourceHandler);
 
-    this.currentSource_ = source;
+    // Set currentSource_ asynchronously to simulate the media element's
+    // asynchronous execution of the `resource selection algorithm`
+    this.setTimeout(function () {
+      if (source && source.src !== '') {
+        tech.currentSource_ = source;
```
gkatsev: you should bind this method instead of using the `tech` variable.

test/unit/media.js
gkatsev: If it's a breaking change and we feel this is useful, we should consider it quickly so we can add it to 5.0 before it's too late. Though, really, now would be the time to just iterate through the major versions quickly, once it settles down, we should move more carefully.

gkatsev:
Minor comment, otherwise, LGTM.

gkatsev:
LGTM

gkatsev:
Closed via 20b46b9

PR #2254, opened 2015-06-10
Sped up tests by relying on watchify and splitting up tests

heff:
The main thing that's happening here is setting up parallel watch tasks so that we can use watchify's smart caching of browserify source files. We end with 3 watch tasks that run at the same time. We can't use grunt-contrib-watch for all of them, because it just starts browserify from scratch each time, instead of using watchify to cache the files, which is why our suite is so slow right now.

- grunt browserify:watch - rebuild build/temp/video.js with each change for sandbox manual tests
- grunt karma:watch - rerun tests with relevant changes for automated tests
- grunt watch - everything else that doesn't need to be smart. Sass and jshint.

The first two are different tasks because that automated tests only pull in the source files they need, not the full video.js. If we wanted to share the two we would need to rewrite the test suite to work off the fully compiled video.js, but I don't think that's the way we want to go.

Also temporarily killed API tests. I'm combining them with other tests so we only have one suite.

Would be interested to get @gkatsev's feedback since the parallel watch tasks idea came from him.
And also @dmlap's since this is reverting some of the work he did to compile the library first before running tests. Let me know if any of this isn't clear.


grunt.js
```diff
@@ -314,6 +287,18 @@ module.exports = function(grunt) {
...
+      // Run multiple watch tasks in parallel
+      // Needed so watchify can cache intelligently
+      watchAll: {
+        command: [
+          'grunt connect:dev',
+          'grunt watch',
+          'grunt browserify:watch',
+          'grunt karma:watch'
+        ].join('&')
+      }
```
mmcc: Is there a benefit of doing it this way over something like [grunt-concurrent](https://github.com/sindresorhus/grunt-concurrent)?

heff: Cooool. Didn't know about that. I'll try it. Originally had a non-grunt `wait` command in there too, but with this it's not needed.

mmcc: Heh, my bad, should have mentioned that yesterday. What we've got here makes sense, but while we're still in a Grunt world we should probably use grunt-concurrent. Also helps avoid any weirdness around zombie processes that this could potentially introduce.

test/karma.conf.js
```diff
@@ -1,71 +1,28 @@
...
-      '../build/temp/video-js.min.css',
-      '../build/temp/video.min.js',
-      'api/api.js'
+      '../build/temp/video-js.css',
+      '../test/globals-shim.js',
+      '../test/unit/**/*.js'
     ],
 
+    preprocessors: {
+      '../test/**/*.js': [ 'browserify' ]
```
gkatsev: is this thing actually working well? We've had trouble with it previously. Also, it makes it annoying to work with the standalone qunit file. At the very least, we'd want a separate watchify task to build out the test file that the standalone html page wants.

grunt.js
heff: It's been switched to grunt-concurrent now. Was able to back out the keep alive change for connect as well.

test/karma.conf.js
heff: I haven't seen any problems with the browserify preprocessor yet. What trouble did you have?

And yeah, just remembered the third piece in all this is the tests' browser UI. Bah. That would either need to be a third watch task, or share the file with karma like we were doing. The biggest problem I saw with how we were doing it was the line numbers in errors were useless. With this change the line numbers aren't perfect but they're better. Any idea what we were doing wrong before to lose the source map chain?

mmcc:
lgtm

dmlap:
I'm all for speeding up development builds. My main goal in the build work I did last month was to ensure it remained easy to open up the automated tests in your regular browser so you can debug issues properly. I'd be happy tracing back concatenated+babelified files to their sources as long as I don't have to attempt to fix a test issue solely through a stack trace on the command line. I love running automated tests through the command line but it's a terrible way to fix them when they break.

heff:
Ok, I think I have a solution for tests here. Not perfect but better. Could use a review, please!

Run `grunt dev`, it will...
- Build
- Start a server
- Concurrently watch multiple file sets
  - jsbin and sass watching
  - browsersify:watch, building the lib for sandbox and dist while using watchify caching for speed

Run `grunt test-ui`, it will watch and build test files for the Qunit UI, located at localhost:9999/test/. It requires `grunt dev` to be running the server already.

Run `grunt test-cli` it will run `karma:watch`, watching and auto running test files on the command line when they change.

Also...
- Fixed line number reporting (coverage only runs on Travis). In the Qunit UI, you have to use the try/catch option to see the source line numbers in the console. Qunit doesn't handle sourcemaps.
- Consolidated Travis test differences into karma.conf.js
- Removed iPad and Android from Sauce testing for now
- Updated Travis to use their new containers for better speed, and also added npm caching

heff:
I should also note, I tried combining ALL the watch tasks into one concurrent command, but it was a mess of log output. I think it probably makes more sense just to open a  new terminal for other tasks. I welcome any ideas for improving from here.

Also there's something wrong with our tests right now that causes them to run infinitely when there's an error in a commonly used file, like component. I can't figure it out, but it made the super-concurrent setup all the more complicated.

.travis.yml
```diff
@@ -22,3 +23,7 @@ env:
   - secure: GIbhjUJapvC70nIZVlhVyK+3KAD2TVKpiY/q412OO7V2izbvcM1tvU3LBoMZbROzrt5TT84tCoJDvHnrpL0OvxPwrzL5CUU7h4UTxhTOyQkEinbYAnWlW9wdrvtdczsEvANkFPqBZ53B3hVHZHMLOG8QRWaTBicF68vSHEJFqb4=
```
mmcc: This isn't your fault, but while you're editing this Travis yml file, please fix this indentation.

grunt.js
```diff
@@ -348,25 +353,17 @@ module.exports = function(grunt) {
...
+  grunt.registerTask('default', ['test']);
 
-  // Development watch task. Doing the minimum required.
-  grunt.registerTask('dev', ['jshint', 'sass', 'browserify:build', 'karma:chrome']);
+  grunt.registerTask('test', ['build', 'karma:defaults']);
 
-  // Skin development watch task.
-  grunt.registerTask('skin-dev', ['connect:dev', 'watch:skin']);
+  // Run while developing
+  grunt.registerTask('dev', ['build', 'connect:dev', 'concurrent:watchSandbox']);
```
mmcc: Does `connect:dev` actually work with this setup?

mmcc:
lgtm

.travis.yml
```diff
@@ -8,6 +8,7 @@ before_install:
 - export CHROME_BIN=chromium-browser
 - export DISPLAY=:99.0
 - sh -e /etc/init.d/xvfb start
+script: grunt test && grunt coveralls
```
gkatsev: Why not just keep the default and have `npm test` either get passed wither to run coveralls as an option or figure it out itself?

heff: Can you show me what you mean? Previously we had a conditional in the grunt file and it was creating a long chain to follow to figure out tests. This at least seemed like an improvement.

grunt.js
heff: Yeah, it works here. It has to be part of a chain of tasks or it just shuts down, in case we want another task that has keepalive: true.

.travis.yml
heff: fixed

.travis.yml
```diff
@@ -1,24 +1,28 @@
...
-  | bash; fi
+  - if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then curl https://gist.githubusercontent.com/santiycr/5139565/raw/sauce_connect_setup.sh | bash; fi
 before_install:
-- export CHROME_BIN=chromium-browser
-- export DISPLAY=:99.0
-- sh -e /etc/init.d/xvfb start
+  - export CHROME_BIN=chromium-browser
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+script: grunt test && grunt coveralls
```
gkatsev: `npm test` can receive a arguments in npm 2.0 and up. So, we can do `npm test -- --coverage` and then the script can see whether that argument was available.
Alternatively, the value for `npm test` could be something like `grunt test && if [ "$TRAVIS" = "true" ] grunt coveralls]` (or however you check for env variables.

grunt.js
```diff
@@ -219,72 +218,59 @@ module.exports = function(grunt) {
...
-            'test/globals-shim.js',
-            'test/unit/**/*.js'
-          ]
+      watch: {
+        options: {
+          watch: true,
+          keepAlive: true
         },
+        files: {
+          'build/temp/video.js': ['src/js/video.js']
```
gkatsev: why shouldn't `browserify:watch` watch the test files as well?

gkatsev: Or should `browserify:tests` also contain `watch:true` as an option?

grunt.js
gkatsev: can we get another task that runs `['build', 'connect:dev', 'concurrent:watchAll']`?

test/karma.conf.js
```diff
@@ -120,5 +87,72 @@ module.exports = function(config) {
...
+    },
+
+    safari_sl: {
+      singleRun: true,
+      base: 'SauceLabs',
+      browserName: 'safari',
+      platform: 'OS X 10.8'
+    },
+
+    ipad_sl: {
```
gkatsev: didn't we want to disable ipad for now because it's so flaky?

package.json
```diff
@@ -47,6 +47,7 @@
     "grunt-banner": "^0.3.1",
     "grunt-browserify": "3.5.1",
     "grunt-cli": "~0.1.0",
+    "grunt-concurrent": "^1.0.0",
```
gkatsev: not necessary for this PR, but we probably should look through our devDeps and make sure we're still using them all and if not, get rid of them.

gkatsev:
For the most part, looking good. I'd like to actually try it out locally before we pull it in, though.

grunt.js
heff: @gkatsev that's related to my previous note
> I should also note, I tried combining ALL the watch tasks into one concurrent command, but it was a mess of log output. I think it probably makes more sense just to open a new terminal for other tasks. I welcome any ideas for improving from here.

I'll add it in and you can try it and let me know what you think.

.travis.yml
heff: Moved it to package.json. We'll see if I did it right.

grunt.js
heff: `browserify:watch` is just for the build/dist video.js file. `browserify:tests` does have `watch:true`, and is specifically for building the file that the Qunit UI needs. If we want a single run script for Qunit UI we would need to add that.

test/karma.conf.js
heff: Yeah, it is disabled now because it's not being added to the browsers list in the `process.env.SAUCE_ACCESS_KEY` conditional. The custom launcher definition here doesn't do anything unless you use it.

package.json
heff: Yeah, we should. I'm sure there's a few.

heff:
@gkatsev I've added in your requests. Let me know when you've had a chance to try it out. I'd like to get this pulled in soon.

gkatsev:
I'll get to it really soon.

grunt.js
gkatsev: Oh, I've missed that `browser:tests` has the `watch:true` because of the diff.

grunt.js
```diff
@@ -314,6 +300,23 @@ module.exports = function(grunt) {
...
+    },
+    concurrent: {
+      options: {
+        logConcurrentOutput: true
+      },
+      // Run multiple watch tasks in parallel
+      // Needed so watchify can cache intelligently
+      watchAll: [
+        'watch',
+        'browserify:watch',
```
gkatsev: we should move `browserify:watch` and `browser:tests` to the first two items in this list.
Seems like as it currently is, `browserify:tests` ends  up running after `karma:watch` runs so, it takes some time extra before you can refresh your browser window.

heff: The order doesn't seem to have a significant impact on when they're run (or rather when they finish), from what I've seen. But I'll try it. I think we still want `watch` first though, because if you're developing the skin you don't want to wait for the rest, and also the tests depend on the skin being available.

grunt.js
heff: I've been using the side-by-side view of the diffs lately. Definitely seems to help for diffs like this.

grunt.js
heff: Also, this is why I separated them out into `test-ui` and `test-cli`, so you can choose which you're using. I like to let the cli version run and beep at me when there's test errors, so I'd personally prioritize `karma:watch` over the others.

heff:
@gkatsev does this mean anything to you? https://travis-ci.org/videojs/video.js/builds/71163633#L618

gkatsev:
Not sure. I can take a look tomorrow.

PR #2256, opened 2015-06-12
More fixes for async currentSrc behavior

imbcmdth:
Two main fixes here:

1. The `player.currentSrc` function didn't allow an empty string to be returned from techs - it would always use `src` in that case
2. In my testing it appears that `currentSrc` should always be updated by the time any other `loadstart` listener is triggered which wasn't always the case using only a setTimeout of 0.

The second bug has the potential to cause problems in plugins that rely on `currentSrc` being properly updated by the time `loadstart` is emitted. Since events are triggered synchronously, there are cases where the `setTimeout` is simply "too late" to match the video element's behavior.

These changes bring it much more in-line with what the video element does 

As an example, I ran the test @heff created http://jsbin.com/wulomalopu/1/edit?js,console,output (but modified to use video.js) and the output with these changes on Chrome 43 is:

```
"el created"
""
""
2
"source set"
"http://vjs.zencdn.net/v/oceans.mp4"
""
3
"async"
"http://vjs.zencdn.net/v/oceans.mp4"
"http://vjs.zencdn.net/v/oceans.mp4"
4
"source set again"
"http://vjs.zencdn.net/v/oceans.webm"
"http://vjs.zencdn.net/v/oceans.mp4"
5
"async again"
"http://vjs.zencdn.net/v/oceans.webm"
"http://vjs.zencdn.net/v/oceans.webm"
6
"empty source set"
"file:///Users/jrivera/projects/video.js/tester.html"
"http://vjs.zencdn.net/v/oceans.webm"
7
"loadstart"
"file:///Users/jrivera/projects/video.js/tester.html"
"http://vjs.zencdn.net/v/oceans.webm"
8
"async for empty source"
"file:///Users/jrivera/projects/video.js/tester.html"
"http://vjs.zencdn.net/v/oceans.webm"
```

..and that is very, very close to the test's output on Chrome.

### DIFF
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -36,7 +36,7 @@ vjs.MediaTechController = vjs.Component.extend({
       this.emulateTextTracks();
     }
 
-    this.on('loadstart', this.updateCurrentSource);
+    this.on('loadstart', this.updateCurrentSource_);
 
     this.initTextTrackListeners();
   }
@@ -178,7 +178,7 @@ vjs.MediaTechController.prototype.onTap = function(){
  * in a timeout to make sure it is set asynchronously before anything else
  * but before other loadstart handlers have had a chance to execute
  */
-vjs.MediaTechController.prototype.updateCurrentSource = function () {
+vjs.MediaTechController.prototype.updateCurrentSource_ = function () {
   // We could have been called with a 0-ms setTimeout OR via loadstart (which ever
   // happens first) so we should clear the timeout to be a good citizen
   this.clearTimeout(this.updateSourceTimer_);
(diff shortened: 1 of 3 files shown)


gkatsev:
The other option is to just make setting `currentSrc` synchronous.

src/js/media/media.js
```diff
@@ -168,6 +170,24 @@ vjs.MediaTechController.prototype.onTap = function(){
...
 
+/**
+ * Set currentSource_ asynchronously to simulate the media element's
+ * asynchronous execution of the `resource selection algorithm`
+ *
+ * currentSource_ is set either as the first loadstart event OR
+ * in a timeout to make sure it is set asynchronously before anything else
+ * but before other loadstart handlers have had a chance to execute
+ */
+vjs.MediaTechController.prototype.updateCurrentSource = function () {
```
dmlap: Should this method be `_`-suffixed?

dmlap:
@imbcmdth `updateCurrentSource` feels like it should be marked as internal. Otherwise, LGTM.

gkatsev:
Closed via 463ba4ea44cf4fcbe749ef9e99f1eadad139207d

PR #2264, opened 2015-06-15
More fixes for async currentSrc behavior

gkatsev:


heff:
So we talked about moving to a synchronous currentSrc. Does that mean we should close this one?

gkatsev:
Yes.

Issue #1904, opened 2015-02-27
No tracks in Safari


ksheurs:
I am trying to upgrade to 4.12.2 (from 4.10.2) and finding subtitles are not working in Safari. Below are the two issues I am seeing:

On Safari desktop (v 6.0.5) there are no specific JavaScript errors, they just seem to silently not load. I am able to reproduce on www.videojs.com.

On Safari iOS they properly show in the native menu (http://cl.ly/image/0M180Y0t2Y3n), but selecting them does nothing. The only way I've been able to get subtitles working on iOS is to not initialize VideoJS and rely on native video + `crossorigin="anonymous"` on the video tag.

gkatsev:
Hm... safari 6 I think has some issues with running vtt files natively. You also need to make sure that captions are enabled at the OS-level.
The crossorigin issue is something that we need to fix but yes, that is the workaround.

ksheurs:
Any idea why subtitles would be working in Safari for VideoJS version 4.10.2?

gkatsev:
We changed the way we're doing captions in 4.12. Unfortunately, we don't have a safari 6 desktop device to test with.

heff:
I can confirm that captions do not work in iOS on videojs.com. Shouldn't that work now?

gkatsev:
It fell through the crack. Opened a new issue for the work needed to be done: #1941

gkatsev:
This is because #1888 hasn't been implemented.

EZWrighter:
Wanted to add a note on this in Safari, I got captions working, but they display under the control bar by default.  Which normally wouldn't be an issue if the controlbar disappears on mouse out, but in my case I needed to have the controlbar present all the time.

heff:
Could we [style the shadow dom](
http://www.samdutton.com/track/shadowDOM.html) and make the captions move
up when the controls are showing?

EZWrighter:
Thanks for the idea Heff, but I don't think that will work on newer safari.  Which is the place it is having problems.  Tested safari 5 and the captions display in the right place.  Safari 6 they display under the control bar.  I know Safari removed a lot of the shadow dom functionality in newer releases.

EZWrighter:
Ok...I tested Safari 6 and got the captions above controls with this css:

`video::-webkit-media-text-track-container {
	bottom: 1.5em;
}`

makes it look a little higher in chrome on my machine, but an ok compromise.

Still need to check how this looks in Safari 5 though.  I will update after testing.

[...]

PR #2269, opened 2015-06-16
Crossorigin

### LINKED ISSUE #1888: Add crossorigin documentation and explanation and update examples
At least on iOS, we cannot use native text tracks with cross-domain tracks without applying the `crossorigin="anonymous"` property to the video element. We should apply that where appropriate.
Different browser react differently to that attribute, so, we definitely don't want to apply it by default. We want to apply it to a targeted range of browsers.

The takeaway from #2269:
* [x] We need to document the crossorigin property that when tracks are crossorigin and require CORS the video file itself will also require CORS (at least, as far as we understand it)
* [x] Document that users will need to be in charge of adding or removing the crossorigin attribute on the video element
* [x] We should print a warning in the console when we detect that captions wouldn't be loaded because of CORS
* [x] Update the sandbox example and the website to correctly use the crossorigin attribute

gkatsev:
This tries to manage the lifecycle of the `crossorigin` attribute on the video element. We don't want it available when we're setting the source but we want to make sure it's available before text tracks get loaded. So, we add it back conditionally on `loadstart` or when `addRemoteTextTrack` is called.
It's removed and re-applied when the source of the video is changed. This behavior is only applicable to the Html tech and the Flash tech is unchanged.

Fixes #1888, #1904.

Things left to do:
* [x] Tests
* [ ] Forward port it to master/5.0

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -325,7 +325,11 @@ vjs.Html5.prototype.src = function(src) {
 };
 
 vjs.Html5.prototype.setSrc = function(src) {
+  this.toggleCrossorigin({crossorigin: null});
   this.el_.src = src;
+  setTimeout(vjs.bind(this, function() {
+    this.toggleCrossorigin({crossorigin: 'anonymous'});
+  }));
 };
 
 vjs.Html5.prototype.load = function(){ this.el_.load(); };
@@ -447,6 +451,35 @@ vjs.Html5.prototype.removeRemoteTextTrack = function(track) {
       tracks[i]['parentNode']['removeChild'](tracks[i]);
       break;
     }
+  }
+};
+
+/*
+ * This toggles the crossorigin attribute on the element.
+ * If no options object is specified it will toggle between presence and absence of the crossorigin attribute.
+ * The input to the function is an options object with a property called `crossorigin`.
+ * The value can be either 'anonymous' or 'use-credentials' as seen here <https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes>.
+ * Falsey values will remove the attribute, truthy values other than the two above strings will be interpreted as default (or 'anonymous').
+ */
+vjs.Html5.prototype.toggleCrossorigin = function(opt) {
+  var el = this.el(),
+      remove = false,
+      value = 'anonymous';
+
+  if (!opt) {
+    remove = el.hasAttribute('crossorigin');
+  } else if (opt.crossorigin === 'use-credentials') {
+    value = 'use-credentials';
+  } else if (opt.crossorigin) {
+    value = 'anonymous';
+  } else {
+    remove = true;
+  }
+
+  if (remove) {
+    el.removeAttribute('crossorigin');
+  } else {
+    el.setAttribute('crossorigin', value);
   }
 };
 
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -477,7 +477,7 @@ vjs.Html5.prototype.addCrossorigin = function(value) {
 vjs.Html5.prototype.enableCrossorigin_ = function(value) {
   var attrVal = value;
 
-  if (value !== 'anonymous' || value !== 'use-credentials') {
+  if (value !== 'anonymous' && value !== 'use-credentials') {
     attrVal = 'anonymous';
   }
 
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -20,6 +20,9 @@ vjs.Html5 = vjs.MediaTechController.extend({
 
     vjs.MediaTechController.call(this, player, options, ready);
 
+    // If we need a crossorigin attribute, make sure it gets added by loadstart
+    this.on('loadstart', vjs.bind(this, this.addCrossorigin));
+
     this.setupTriggers();
 
     var source = options['source'];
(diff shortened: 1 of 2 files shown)


gkatsev:
Ugh, phantom.

heff:
Man...what a mess of a situation. Thanks for tackling it. It's really annoying that the tracks rely on the video element's cross origin attribute, not the track element's. Should we push that back to the spec? The track/vtt spec is getting revitalized right now so it would be good timing.

Instead of juggling source and track loading, is there anyway we can encapsulated this in tracks. Can we know when we're adding the tracks to the video element or calling load on them, and add the crossorigin attribute just for that moment and remove it when done?

As is it looks clean to me but there's also enough asynchronisity in the juggling that based on previous experiences, I feel like we can guarantee this will bite us some unexpected way. Makes me cautious about releasing it as a patch.

src/js/media/html5.js
```diff
@@ -408,6 +415,8 @@ vjs.Html5.prototype.addRemoteTextTrack = function(options) {
 
   this.el().appendChild(track);
 
+  this.addCrossorigin();
```
heff: We need this in addition to the `loadstart` listener?

gkatsev: Yep, if you add a remote text track after a video has loaded, we'd want to add the `crossorigin` attribute to video element before adding in the track as we don't have the `loadstart` opportunity for it.

heff: Ah, nvm. In github it looked like this was in `load()`.

gkatsev:
Yep, this is a fun one :)

I think it kind of makes sense that `crossorigin` belongs on the video element. The messy part comes from it also being applied to the video itself rather than just tracks. So, if you don't have CORS headers for the `mp4` file, it'll fail; most people don't have CORS headers for `mp4` files.

> Instead of juggling source and track loading, is there anyway we can encapsulated this in tracks. Can we know when we're adding the tracks to the video element or calling load on them, and add the crossorigin attribute just for that moment and remove it when done?

This is what I was striving for and why there's the `this.addCrossorigin` in the `addRemoteTextTrack` method in the `html5` tech. The main issue with asynchronicity is in the `setSrc` method. I can't re-add the `crossorigin` attribute directly because `currentSrc` gets set asynchronously and we have to wait for it to get set.

As for having a `crossorigin` JIT, that would be nice, though, I'm not certain it works; it'll be easy to find out but there's a more pressing concern with it. If users would only interact with tracks through the menu, we could do this: add the `crossorigin` upon selection and remove it after the track loads. But if the mode is changed via the API, we won't know about it until it is too late (we heard the `change` event) and it errors out; the only way to recover from these errors right now is to load in a new source and re-load the track.

------------------------------------
One option we have with this if we don't feel comfortable releasing it as part of 4.x is to just have it in 5.0. It'll suck if we don't have it in 4 but hopefully 5.0 isn't too far out.

heff:
> I think it kind of makes sense that crossorigin belongs on the video element. The messy part comes from it also being applied to the video itself rather than just tracks. So, if you don't have CORS headers for the mp4 file, it'll fail; most people don't have CORS headers for mp4 files.

Yeah, see I would expect it to apply to the video file first and secondarily the tracks. It's an all or nothing situation now. Which appears to just be how the spec is written.

Is the real answer then that we need to educate people on these restrictions? Should we really be trying to be smarter than the video element with this?

Could we say, if you want to use cross-origin tracks
- You need to include the crossorigin attribute in the video.js embed code
- Your video sources files have to be CORS enabled also

gkatsev:
It sounds like having requiring both the video and the tracks to both have CORS is a security feature (https://html.spec.whatwg.org/multipage/embedded-content.html#security-and-privacy-considerations). So, actually, the behavior that we're relying on could be considered a bug rather than a feature.
> Without this restriction, an attacker could trick a user running within a corporate network into visiting a site that attempts to load a video from a previously leaked location on the corporation's intranet. If such a video included confidential plans for a new product, then being able to read the subtitles would present a serious confidentiality breach.

relevant example from the above link. So, it does sound that if tracks are to be used, the user of videojs should enable CORS on both the text tracks and the video itself, if they're not being hosted from the same domain as the site they are being shown on.

So, the correct answer is to educate people on the issue and have them enable CORS for both the video and track files. Then, they could either add the `crossorigin` attribute themselves or we could add it for them if we detect that tracks are used and it's a different domain than the page its embedded on. However, would we want to make it easier for people who have yet to turn on CORS for the video files?

@dmlap do you have any thoughts on the subject?

dmlap:
I think we should try to expose equivalent functionality to what the video element exposes and educate users on how to setup their embed codes correctly. This is a confusing area of the spec. I think the amount of discussion we've had about this issues indicates we're not yet experts enough to provide a nicer API without risking some use-case we didn't understand.

heff:
Cool, I think we're on the same page then. We should throw a helpful warning when we detect this situation, and not try to be smarter than that. At least not yet. And add all this info to the tracks guide.

To be clear, I also don't think we should try to auto-add crossorigin when tracks need it. It'll be clearer to throw a warning that the tracks (and source) need CORS when we first detect it (or rely on the native error), rather than have the tracks automatically work and later sources throw a security error when it they didn't before adding a track.

gkatsev:
So, the takeaway from this:
* We need to document the crossorigin property that when tracks are crossorigin and require CORS the video file itself will also require CORS (at least, as far as we understand it)
* Document that users will need to be in charge of adding or removing the crossorigin attribute on the video element
* We should print a warning in the console when we detect that captions wouldn't be loaded because of CORS
* Update the sandbox example and the website to correctly use the crossorigin attribute

Does that sound like the correct TODO here?

heff:
Sounds right to me.

gkatsev:
Closing and updating #1888 with the TODO from my last post.

PR #2270, opened 2015-06-17
Master

mboles:
Added JSDoc comments to files at the root of the src/js directory

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -8,8 +8,8 @@ import Component from './component.js';
  * big play button is done via CSS and player states.
  * @param {Player|Object} player
  * @param {Object=} options
- * @class
  * @constructor
+ * @class BigPlayButton
  */
 class BigPlayButton extends Button {
 
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -1,3 +1,6 @@
+/**
+ * @file Needed for full path retrieval
+ */
 import Button from './button.js';
 import Component from './component.js';
 
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -6,17 +6,27 @@ import Component from './component.js';
 /**
  * Initial play button. Shows before the video has played. The hiding of the
  * big play button is done via CSS and player states.
- * @param {Player|Object} player
- * @param {Object=} options
- * @constructor
+ *
+ * @extends Button
  * @class BigPlayButton
  */
 class BigPlayButton extends Button {
 
+  /**
+  * Allow sub components to stack CSS class names
+  *
+  * @return {String} The constructed class name
+  * @method buildCSSClass
+  */
   buildCSSClass() {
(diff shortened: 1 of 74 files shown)


heff:
Wow, this looks like a ton of work. Thanks Matt!

src/js/big-play-button.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @file Needed for full path retrieval
```
heff: Saying this each time isn't ideal. Can you explain why it's needed, and should we try to put real file descriptions for each of these? It could be event as generic as the file name.

src/js/extends.js
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
heff: Does this change just make sure it's private?

src/js/player.js
```diff
@@ -2212,51 +2522,58 @@ Player.prototype.options_ = globalOptions;
...
  * Fired when the user is inactive, e.g. a short delay after the last mouse move or control interaction
+ *
  * @event userinactive
  */
 Player.prototype.handleUserInactive;
 
 /**
- * Fired when the current playback position has changed
- *
+ * Fired when the current playback position has changed *
```
heff: You've got an extra asterisk on the end of this one

src/js/poster-image.js
```diff
@@ -72,8 +79,11 @@ class PosterImage extends Button {
   }
 
   /**
-   * Set the poster source depending on the display method
-   */
+  * Set the poster source depending on the display method
```
heff: The alignment of this one looks like it needs to be one space to the right. Not sure if that actually affects parsing.

src/js/slider/slider.js
```diff
@@ -181,13 +232,22 @@ class Slider extends Component {
   /**
    * Listener for click events on slider, used to prevent clicks
    *   from bubbling up to parent elements like button menus.
-   * @param  {Object} event Event object
-   */
+  *
```
heff: The alignment of the asterisks here gets offset

heff:
I noticed that all the jsdoc comments are now all formatted with no space infront of the asterisks after the first line.
```
/**
*
*/
```
We really want them to be standard jsdoc tags, which do have a space.
```
/**
 *
 */
```
Would you be able to make that update? Otherwise this looks good to merge in.

heff:
Merged in. Thanks @mboles!

src/js/big-play-button.js
```diff
@@ -1,22 +1,41 @@
+/**
+ * @file big-play-button.js
+ */
 import Button from './button.js';
 import Component from './component.js';
 
-/* Big Play Button
-================================================================================ */
+/ * Big Play Button
```
heff: @mboles FYI, the find replace wasn't 100% safe. It added a comment-breaking space in a few places like this. I cleaned them up, but just a heads up in case you use that again.

PR #2272, opened 2015-06-17
Allow viewing in-band metadata tracks from shimmed texttracklist

### LINKED ISSUE #2266: In-band metadata tracks should be visible when emulated text tracks are active
If javascript-emulated text tracks are being used, `player.textTracks()` should return at least all the tracks that are returned from `video.textTracks`.

## Steps to Reproduce
1. Create a player and force emulated text tracks in Safari
1. Load a stream with an embedded metadata track
1. Call `player.textTracks()`

The returned TextTrackList should include a metadata text track but it will be empty.

gkatsev:
Fixed #2266.

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -63,6 +63,7 @@ vjs.Html5 = vjs.MediaTechController.extend({
       this.on('loadstart', vjs.bind(this, this.hideCaptions));
     } else {
       // allow us to view the in-band metadata tracks from our shimmed TextTrackList
+      this.on('loadstart', vjs.bind(this, this.removeInBandMetadataTracks));
       this.el().textTracks.addEventListener('addtrack', vjs.bind(this, this.addInBandMetadataTracks));
     }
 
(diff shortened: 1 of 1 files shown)


src/js/media/html5.js
```diff
@@ -61,6 +61,9 @@ vjs.Html5 = vjs.MediaTechController.extend({
 
     if (this['featuresNativeTextTracks']) {
       this.on('loadstart', vjs.bind(this, this.hideCaptions));
+    } else {
+      // allow us to view the in-band metadata tracks from our shimmed TextTrackList
+      this.on('loadedmetadata', vjs.bind(this, this.addInBandMetadataTracks));
```
dmlap: I'm not sure that inband tracks will show up by `loadedmetadata`. Did you find this in the spec or discover it experimentally?

gkatsev: Experimentally.

gkatsev: Though, this reminds me. I probably need to remove the old ones from the list on source change.

dmlap: An in-band metadata track might not be evident until a decent amount of buffering has occurred. If there were a way to make the track registration a bit more robust for that case, that would be great.

gkatsev: I'll take a look whether the `addtrack` event fires for it. Might listen to both or something.

gkatsev:
This should handle multiple in-band metadata tracks better and should be able to clear them out properly between video loads.
I'm not sure how easy it would be to test this.
Also, this would also need to be ported to master as well.

src/js/media/html5.js
```diff
@@ -184,6 +193,39 @@ vjs.Html5.prototype.hideCaptions = function() {
   }
 };
 
+vjs.Html5.prototype.addInBandMetadataTracks = function(event) {
+  var tracks = this.textTracks(),
+      newTrack = event.track || null,
+      track,
+      i = 0;
+
+  for (i = 0; i < tracks.length; i++) {
```
dmlap: This shouldn't be necessary unless `addtrack` fired twice, right?

dmlap:
We should have a test for this.

src/js/media/html5.js
gkatsev: Yeah, I'm not certain it'll be necessarily useful. Even if you re-apply the same video you get a new track.
This was kind of left-over from when I had the removal of tracks in the same method as additions and so far I left it for "extra security" but I'm not certain it actually buys us much.

gkatsev:
I'll try and come up with some tests.

gkatsev:
This will become obsolete as part of the work to fix #2367. I'm using this general framework for that fix anyway. Given that we've decided not to include it in stable, I'm going to close this.

PR #2293, opened 2015-06-29
Add the vjs-big-play-centered class back

mmcc:


heff:
Looks good to me. Also to note, this switches the control bar variables to use the double-dash style for properties. I believe @mmcc said that's a standard sass style.

mmcc:
[Reference](https://medium.com/@fat/mediums-css-is-actually-pretty-fucking-good-b8e2a6c78b06#ff4b). I've seen it in a few other places, but over time I've started following more and more of what's outlined in that post, so at this point I'm almost willing to say we should just wholesale adopt it.

heff:
I haven't read all of that but don't have any immediate objections.

PR #2299, opened 2015-07-01
Fix inactivityTimeout declaration. Fixes #2298.

### LINKED ISSUE #2298: Inactivity check contains incorrect variable declaration leading to flickering behavior
Inactivity check is not behaving correctly as it "flickers" between states multiple times when the player is about to go to *inactive* state. The reason for this is incorrect variable declaration of `inactivityTimeout` variable in [player.js#1923](https://github.com/videojs/video.js/blob/master/src/js/player.js#L1923-L1934):

    let activityCheck = this.setInterval(function() {
      let inactivityTimeout;
      // ...
      this.clearTimeout(inactivityTimeout);
      // ...
      inactivityTimeout = this.setTimeout(...)

Variable `inactivityTimeout` must be declared outside of the inner function to work properly (otherwise the `clearTimeout` is simply no-op).

pavelhoral:
For some reason `contrib feature submit` (as described in CONTRIBUTING.md) did not work so I am submitting this manually via GitHub web.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1919,9 +1919,8 @@ class Player extends Component {
     // `this.reportUserActivity` simply sets this.userActivity_ to true, which
     // then gets picked up by this loop
     // http://ejohn.org/blog/learning-from-twitter/
+    let inactivityTimeout;
     let activityCheck = this.setInterval(function() {
-      let inactivityTimeout;
-
       // Check to see if mouse/touch activity has happened
       if (this.userActivity_) {
         // Reset the activity tracker


mmcc:
:+1: Looks good to me.

heff:
lgtm

Thanks for trying out contrib. We need to fix the contributing guide and that process.

heff:
Merged in, thanks! What error did you see with contrib feature submit?

pavelhoral:
I did few steps wrong at first (e.g. I didn't checkout the feature branch). However in the end everything works... just the pull request was not created. I get no error from `contrib feature submit`. After asking pull request details and GH credentials it looked like it is doing something, but ended up making nothing (the only output was something like `-> undefined`, which was probably indication that something failed).

pavelhoral:
@heff It is already reported as https://github.com/contrib/contrib/issues/22 .

heff:
Ah, got it. Thanks!

PR #2304, opened 2015-07-03
Added 'rtmp/mp3' to the list of supported formats

stepan-romankov:
Corresponding changes was done for flash tech https://github.com/slidepresenter/video-js-swf/commit/50f3e9b427c2079508ee1cd0926d6c2705487f7e

### DIFF
--- a/src/js/tech/flash-rtmp.js
+++ b/src/js/tech/flash-rtmp.js
@@ -1,7 +1,8 @@
 function FlashRtmpDecorator(Flash) {
   Flash.streamingFormats = {
     'rtmp/mp4': 'MP4',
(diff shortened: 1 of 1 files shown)


gkatsev:
If you can make a PR for your videojs-swf changes and rebase this against latest master we'll pull this in.
However, I've been able to play an rtmp mp3 stream in the current version just fine.

Agnostic:
Hi @stepan-romankov , do you know how can I get the duration of an rtmp stream? I'm using a build of this branch but when I call `player.duration()` the value is always 0, I was looking at the `.as` files but I don't understand much =/

gkatsev:
We can re-open if there are updates.

PR #2308, opened 2015-07-06
Enable simple class extensions without any constructor

### LINKED ISSUE #2303: Can not use `videojs.extends` without a constructor function
Would be nice if we can use `videojs.extends` without a constructor function. In that case it can provide implicit constructor which would just call parent constructor:

    var Component = videojs.getComponent('Component');
    var Foo = videojs.extends(Component, { 
        // constructor: function() {
        //     Component.apply(this, arguments);
        // },
        createEl: function() {
        }
    });

Currently this does not work as the `videojs.extends` expects the second parameter to either be constructor or object with a `constructor` property.

pavelhoral:
Allows `videojs.extends` to properly handle creating a subclass without any specified constructor. Previous implementation just threw child properties/methods away. Fixes #2303.

### DIFF
--- a/src/js/extends.js
+++ b/src/js/extends.js
@@ -47,8 +47,10 @@ const extendsFn = function(superClass, subClassMethods={}) {
   };
   let methods = {};
 
-  if (subClassMethods.constructor !== Object.prototype.constructor) {
-    subClass = subClassMethods.constructor;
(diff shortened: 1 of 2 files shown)


heff:
Looks good to me, thanks! And thanks for the test.

heff:
Thanks again!

PR #2328, opened 2015-07-09
Support text descriptions tracks in v5.

OwenEdwards:
Add support for text descriptions tracks into video.js v5. Adds a control for 'audio description' tracks, and allows the tracks to be displayed and styled by the same control as closed captions. Also allows the description track to be hidden, but still read by a screen reader, for the use case where a user has a screen reader but doesn't want the text overlaying the video.

An example HTML file is included as `sandbox/descriptions.html.example`; the associated descriptions track, chapters track, and several captions tracks are in `docs/examples/shared/elephantsdream`.


src/js/tracks/text-track-settings.js
```diff
@@ -282,6 +288,13 @@ function captionOptionsMenuTemplate() {
             <option value="small-caps">Small Caps</option>
           </select>
         </div> <!-- vjs-font-family -->
+        <div class="vjs-descriptions-playback vjs-tracksetting">
+          <label class="vjs-label">Descriptions</label>
+          <select>
+            <option value="">Visible</option>
+            <option value="screenReaderOnly">Hidden (for screen reader users)</option>
```
gkatsev: Which one is supposed to be the default? Visible or hidden?

gkatsev:
Looks like there are some test failures.

src/js/tracks/text-track-display.js
```diff
@@ -103,7 +121,20 @@ class TextTrackDisplay extends Component {
       cues.push(track['activeCues'][i]);
     }
 
-    window['WebVTT']['processCues'](window, track['activeCues'], this.el_);
+    if (track['kind'] === 'descriptions') {
+
+      if ((overrides.descriptionsPlayback) && (overrides.descriptionsPlayback === 'screenReaderOnly')) {
```
gkatsev: Currently, if you enable both descriptions and captions at the same time, they would overlap each other.

src/js/tracks/text-track-settings.js
OwenEdwards: The default is Visible, set in the [default button handler](https://github.com/ICTASolutions/video.js/blob/descriptions-track-support-for-v5/src/js/tracks/text-track-settings.js#L34). Its value is "", following the pattern of the [Foreground Color selector](https://github.com/ICTASolutions/video.js/blob/descriptions-track-support-for-v5/src/js/tracks/text-track-settings.js#L188). The settings control could be a simple checkbox, but this allows for future expansion to other modes of display. Defaulting to Visible makes the most sense, since most users (those without a screen reader) will see the descriptions if they turn them on, and users with a screen reader in general won't be distracted by visible text.

gkatsev: Makes sense.

src/js/tracks/text-track-display.js
OwenEdwards: That's technically true, although:
* It's a pretty extreme use case to want captions _and_ descriptions.
* The author of the descriptions ought to avoid them clashing with captions along the timeline, specifically because they want the descriptions to be read out in pauses in the dialog/soundtrack.

Handling this situation in a more sophisticated way will get complicated quickly. What happens if you turn on CC _and_ Subtitles?

gkatsev: It looks like it isn't possible right now, so... :D

gkatsev: In which case, this shouldn't be a huge issue, though, we probably would want to think about whether we should allow having captions and subtitles on at the same time.

sandbox/descriptions.html.example
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
```
heff: Could we set this up as `docs/examples/descriptions/index.html` instead of sandbox? That's the new, more appropriate place.

src/js/control-bar/text-track-controls/descriptions-button.js
```diff
@@ -0,0 +1,25 @@
+import CaptionsButton from './captions-button.js';
+
+/**
+ * The button component for toggling and selecting descriptions
+ *
+ * @constructor
+ */
+class DescriptionsButton extends CaptionsButton {
```
heff: Is there a reason this needs subclass CaptionsButton instead of TextTrackButton?

If I were to guess it would be because there doesn't appear to be any handling of native vs non-native tracks in TextTrackButton. @gkatsev is there a reason why that's in CaptionsButton and not TextTrackButton?

src/js/tracks/text-track-display.js
```diff
@@ -63,14 +64,31 @@ class TextTrackDisplay extends Component {
   }
 
   createEl() {
-    return super.createEl('div', {
+    var parentEl = super.createEl('div', {
+    });
```
heff: Kill the extra newline in the middle of the object here, for cleanliness.

src/js/tracks/text-track-display.js
heff: Can vtt.js handle multiple tracks?

heff:
Thanks Owen! Looks pretty good to me. A few notes to address, and we need to get the tests passing. It's immediately obvious to me what the problem is.

src/js/tracks/text-track-display.js
gkatsev: The way we have it implemented right now, we don't support having multiple tracks showing.
A solution would be to just merge all the active cues we want to display into one array and give that to vttjs. But this would require some refactor.

src/js/control-bar/text-track-controls/descriptions-button.js
gkatsev: Subclassing from CaptionsButton would automatically get us a "descriptions settings" menu item.

heff: Interesting. Do captions and descriptions have independent styles or do they share them?

gkatsev: I think they'll end up sharing it, which isn't a huge issue.

OwenEdwards: As @gkatsev says, subclassing CaptionsButton is to get the "descriptions settings" menu item. I guess it could be refactored into a TextTrackWithSettingsButton, which is a subclass of TextTrackButton, and of which CaptionsButton and DescriptionsButton are a subclass?

In this implementation, Captions and descriptions share the same style options. This was a compromise between not styling the descriptions at all (since the main use case for descriptions is to render them in a non-visual way with a screen reader), and adding the complexity of a whole separate set of styling options.

The default of having descriptions rendered visibly is mostly so that people without a screen reader can get an idea what descriptions are, and don't get confused by a control button which would seem to do nothing if the descriptions were rendered invisibly by default.

Since descriptions are similar to captions, it makes sense to style them as captions. A more sophisticated option would style them differently from captions and allow visual presentation of both descriptions and captions at the same time, but watching both at once isn't a realistic use case; the amount of work to implement that option doesn't make sense.

src/js/tracks/text-track-display.js
OwenEdwards: I think we should split "showing multiple text tracks at once" out as an issue separate from descriptions. It looks like Chrome natively allows simultaneous display of a subtitles track and a captions track, and that video.js doesn't match that behavior for vtt.js subtitles/captions.

sandbox/descriptions.html.example
OwenEdwards: @heff I can do this, although it took me a while to realize that HTML in the docs/examples directory (including the existing `simple-embed/index.html`) don't work in that directory, only once they've been copied into the dist directory, because of the need for the built video.js files. Might it be worth having a note to that effect somewhere, since the HTML files in the sandbox work where they are?

OwenEdwards:
I reworked this to address the issues raised by @gkatsev and @heff, and to include a clearer description in the PR. That is at #2343, so I'm closing this PR, but I'll open a new issue about displaying multiple text tracks simultaneously.

Issue #2335, opened 2015-07-10
5.0 (rc) broken in IE8


nickygerritsen:
It seems that at least the video.js 5.0 RC's are broken in IE8 with the videojs-ie8 part.

It does play video, but as soon as the player controls "autohide" and you hover over the video, IE8 reloads the page in compatibility mode after an error.

After some investigation, it seems this is because of at least two reasons:

* vtt.js has some "trailing comma's" in object literals. They have fixed this, but not in the "shim-build" branch video.js is using. See https://github.com/mozilla/vtt.js/commit/db56965c978c0fd5a2993085a454f2fcc5934816 for the fix.
* The other problem seems to be related to babel. I get an "Accessors not supported!" exception and this appears somewhere in the babel code. I could not really find what goes wrong exactly, as the IE8 devtools are horribly slow and tend to crash

For a test case, see: http://jsbin.com/bihipoporo/edit?html,output
Note that this example uses "local" .js and .css files as I don't know of there are any 5.0 CDN versions yet.

heff:
Thanks for digging in to that. We'll get the vttjs update in today. You can use rawgit.com and the tagged releases to load 5.0 versions into jsbin.

I'm not quite sure about the babel issue...

gkatsev:
I have a PR out to fix the vttjs issue in master: https://github.com/videojs/video.js/pull/2331

gkatsev:
I think we might be misconfiguring `babelify` to loose mode. Looking into it.

gkatsev:
Well, looks like the main build is working fine for me. The test and watch builds don't specify loose mode so I'll add that but otherwise, I'm not seeing any issues.

nickygerritsen:
I'll try on Monday if I can reproduce it then. If so I'll create a jsbin that is problematic.

gkatsev:
I'm still seeing an error in ie8 but it doesn't seem to be affecting anything...
It's saying `Object doesn't support this property or method` on this line (in `buffered` method):
```js
    return _utilsTimeRangesJs.createTimeRange(0, this.el_.vjs_getProperty('buffered'));
```

gkatsev:
And I'm no longer seeing that error anymore...

heff:
#2316 probably fixed the getProperty error

nickygerritsen:
I manually applied that PR indeed because I was getting that same error before.

gkatsev:
Ah, yes, that would do it since I rebased to make PR #2337.

[...]

PR #2337, opened 2015-07-10
Unify 'loose' option for all babelify transforms

gkatsev:


gkatsev:
Should help out with #2335

grunt.js
```diff
@@ -235,7 +235,7 @@ module.exports = function(grunt) {
           transform: [
             require('babelify').configure({
               sourceMapRelative: './src/js',
-              loose: 'all'
+              loose: ['all']
```
heff: ...really?

gkatsev: I don't think this was actually an issue. It seems to work just fine. However, it seems to expect an array, so, I've changed it to be that.

heff:
If you don't mind, I'll just change this in #2254 to avoid conflicts.

gkatsev:
I'm ok with that.

PR #2338, opened 2015-07-10
Add poster to the options the tech are receiving

eXon:
The only way I can know if I must set the default YouTube poster is by checking if the option poster is already set. However, that information is not sent to the tech right now.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -503,7 +503,7 @@ class Player extends Component {
       'preload': this.options_.preload,
       'loop': this.options_.loop,
       'muted': this.options_.muted,
-      'poster': this.options_.poster
+      'poster': this.poster()
     }, this.options_[techName.toLowerCase()]);
 
     if (this.tag) {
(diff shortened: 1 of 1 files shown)


heff:
Makes sense to me. Does the video.js poster not cover up the youtube poster?

eXon:
The tech needs to set the YouTube poster url. If we use the one displayed by the iframe, we have a big red ugly play button (really weird with the vjs poster). When we set a VJS poster, it hides the whole iframe until you click play.

src/js/player.js
```diff
@@ -502,7 +502,8 @@ class Player extends Component {
       'autoplay': this.options_.autoplay,
       'preload': this.options_.preload,
       'loop': this.options_.loop,
-      'muted': this.options_.muted
+      'muted': this.options_.muted,
+      'poster': this.options_.poster
```
heff: @eXon could you change this to use `this.poster()`? That would make sure that if the poster changes, later techs will get the new poster.

heff:
Merged in, thanks! I think we still have a little to figure out around the poster in #2339.

PR #2339, opened 2015-07-10
Listen for posterchange on tech

eXon:
When someone is using a YouTube video and doesn't have a poster, we need to put the YouTube default poster. However, this is an async process. We need to check if the highres version exists. If it doesn't we use the low res.

To make that possible, the player has to listen for posterchange on the tech and trigger it again on himself so that the poster can set himself correctly.


heff:
I'm not understanding why the player needs to get involved in this process. If the youtube player is handling the poster, it seems like the process of deciding which poster to use should be in a black box as far as the player is concerned.

eXon:
Even though `poster()` returns something in the tech, the getter is actually never called. How can the player know when to get the new poster to update it?

heff:
Right now the poster attribute is really only used by the [PosterImage](https://github.com/videojs/video.js/blob/dfd481037e19c2af546ccbc706b7fb42db1c7125/src/js/poster-image.js) component. 

We assume:
- The poster source is being supplied by the player user (not the tech)
- The PosterImage component of the player is the primary display of the poster

While the poster is passed to the tech, it's more of a fallback, and I think it's rare that the tech's poster would actually show up whenever the user does supply a poster.

We currently [don't ever assume the tech has the source of the poster](https://github.com/videojs/video.js/blob/dfd481037e19c2af546ccbc706b7fb42db1c7125/src/js/player.js#L1884), however we do still let the [tech know](https://github.com/videojs/video.js/blob/dfd481037e19c2af546ccbc706b7fb42db1c7125/src/js/player.js#L1896), but again, it's more of a fallback.

The `posterchange` event is triggered so the PosterImage component [can know when the user changes the poster](https://github.com/videojs/video.js/blob/dfd481037e19c2af546ccbc706b7fb42db1c7125/src/js/poster-image.js#L24), get the new source from player.poster(), and [change the PosterImage source](https://github.com/videojs/video.js/blob/dfd481037e19c2af546ccbc706b7fb42db1c7125/src/js/poster-image.js#L68).

If the tech triggers a `posterchange`, the PosterImage will look for a new source and get whatever was previously stored in `player.poster_`, as set by the user.

Can the youtube tech provide a image source (jpeg, etc.) that the PosterImage component could read and display?

Do we need to treat the poster similarly to controls and tracks, where there's native (tech) vs custom (player) supplied versions of each, and you can choose between them some how?

eXon:
If the user set a poster on VJS, we should use it no matter what of course. However, if none is provided, we need to set the default YouTube poster otherwise we will see a big red play button in the middle with it.

All the YouTube tech needs is a fallback poster that can be used if the user didn't provided one. Is there any way to do that right now?

heff:
> otherwise we will see a big red play button in the middle with it

Ok, so you're saying when you load a YouTube video with the YouTube player, it provides a URL to a poster image (jpeg or whatever) that we can use in the video.js player's PosterImage component, so that it covers up the YouTube play button. Is that right?

eXon:
It will cover the whole iframe yep. Only way to get rid of that ugly button we don't want twice (VJS already got one).

heff:
Ok cool, so this is specifically what I was looking to understand.
https://github.com/eXon/videojs-youtube/blob/358cd9666eb041f58f96b05bcee27c753b7acb05/src/Youtube.js#L490

The youtube player does have a poster url we could point to if the user doesn't provide one. So this is yet another twist in the player/tech relationship. It's kind of like the native/custom controls and native/custom tracks, except that in this case we would always be using the custom poster and the source would either be coming from the user or the tech.

I think the next step here would be to update the poster function to get the poster from the tech if there's no user supplied poster.

```
poster(src) {
    if (src === undefined) {
      return this.poster_ || this.techCall('poster') || '';
    }
  ...
```

Then ideally the `posterchange` from the tech would only be fired on the player if there was no user supplied poster. That seems to get a little more complicated.

Another note on this is that this means you can't ever disable the poster with this setup. I'm not sure why a user would want to do that though.

pam:

Commit: 1086989500bddb3845fa5f3ab4101190d7aacf53
Build details: https://travis-ci.org/pam/video.js/builds/81285962

pam:

Commit: 57e53b0a04de7c171357e606b883277784499a16
Build details: https://travis-ci.org/pam/video.js/builds/81287330

eXon:
@heff This should fix the default poster for any tech.

- When the tech start loading, it ask for the default poster.
- When the tech is ready, it asks again for the default poster in case it changed (if their is a high res version it will)
- When the source changes, it asks for the new default poster

None of this is done if their is a specified poster.

pam:

Commit: 647354a7deb21945d36a72e96bb1f25713ac564c
Build details: https://travis-ci.org/pam/video.js/builds/81288809

eXon:
Not sure why it fails on Travis. They all passed locally.

pam:

Commit: 2f341b13643616440fa169e530dc7cc200912678
Build details: https://travis-ci.org/pam/video.js/builds/81293406

pam:

Commit: 810d61993bcde6583a0eb6c3162d3f90dc24ab54
Build details: https://travis-ci.org/pam/video.js/builds/81295303

pam:

Commit: e4a671308d03ea1e5b5a79bae7ab808f93dca047
Build details: https://travis-ci.org/pam/video.js/builds/81296925

pam:

Commit: 48ca3a0d13f0f63d0ebc6e98ee157a654a7e492a
Build details: https://travis-ci.org/pam/video.js/builds/81298545

eXon:
Had to switch a few unit tests to use HTML5 because poster() on the Flash version isn't working without mocking a few things.

eXon:
@heff What's the plan on the VJS5 release? This PR is critical for the YouTube tech to work correctly.

heff:
@eXon Trying to get it out tomorrow. Reviewing this one now. If I see any changes I'll make a note but try to change them myself.

test/unit/player.test.js
```diff
@@ -67,7 +67,7 @@ test('should accept options from multiple sources and override in correct order'
   videojs.options.attr = 1;
 
   let tag0 = TestHelpers.makeTag();
-  let player0 = new Player(tag0);
+  let player0 = new Player(tag0, { techOrder: ['html5'] });
```
heff: We have a 'techFaker' tech that I think could be used here instead of html5. It would be technically safer on IE8.

heff:
Made suggestions here: https://github.com/eXon/video.js/pull/1

eXon:
I think it's even better to be able to tell VJS that the default poster has changed. Thank you for the help! :)

pam:

Commit: 9d0453a69aad2aea0d1dbbe11a89a4fa6cb2f09f
Build details: https://travis-ci.org/pam/video.js/builds/82594222

PR #2342, opened 2015-07-11
Send the video.js language to the tech

eXon:
I want to set the YouTube player (or any other iframe tech) in the same language than video.js. To do so, it needs to know the language option.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -503,7 +503,7 @@ class Player extends Component {
       'preload': this.options_.preload,
       'loop': this.options_.loop,
       'muted': this.options_.muted,
-      'language': this.options_.language
+      'language': this.language()
     }, this.options_[techName.toLowerCase()]);
 
     if (this.tag) {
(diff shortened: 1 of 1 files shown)


heff:
lgtm.

Just interested, what does it look like to set the language of the underlying player?

src/js/player.js
```diff
@@ -502,7 +502,8 @@ class Player extends Component {
       'autoplay': this.options_.autoplay,
       'preload': this.options_.preload,
       'loop': this.options_.loop,
-      'muted': this.options_.muted
+      'muted': this.options_.muted,
+      'language': this.options_.language
```
heff: You might want to use the player's language() getter here. It would probably provide some more consistency.

eXon: `language()` returns the player itself instead of the language. Is it normal?

heff: It should only return the player if an argument is passed to the function.
https://github.com/videojs/video.js/blob/dfd481037e19c2af546ccbc706b7fb42db1c7125/src/js/player.js#L2407

eXon:
According to YouTube:

"The interface language is used for tooltips in the player and also affects the default caption track. Note that YouTube might select a different caption track language for a particular user based on the user's individual language preferences and the availability of caption tracks."

heff:
Merged in, thanks!

PR #2343, opened 2015-07-13
Add support for text Descriptions tracks to v5.0.0 rc.13.

OwenEdwards:
Adds a control bar button, a settings option to render the Descriptions visibly or hidden, and a demo of Descriptions in `docs/examples/descriptions/`.

By default, when Descriptions are enabled they are shown visibly like Captions, and are styled by the same settings control as Captions. In the settings control (which can be accessed from either the Captions or Descriptions menu), there is an option 'Descriptions', which defaults to `Visible` but can be set to `Hidden (for screen reader users)`. If Descriptions are enabled they are always available to a screen reader (if the user has one); changing the setting to `Hidden` simply prevents the text from overlaying the video, if the user prefers a purely auditory presentation of Descriptions.

Closes #2328.


pgrodrigues:
Hello, I need to implement a WCAG 2.0 level A compliant video player for a website. What is the current state of video.js regarding the use of descriptions?

Thanks in advance.

OwenEdwards:
Closing this PR - I'll add a new one which fixes the conflicts that this has with newer V5.x video.js.

PR #2348, opened 2015-07-13
Fixes so that the source handler mixin adds the expected API after minification

imbcmdth:
Issue:

In video.js, the `withSourceHandlers` mixin was getting all the names of the functions it used in the SourceHandlers (`canHandleSource`, `handleSource`, and `dispose`) mangled by Closure Compiler.

As a result, it worked only with native source handlers that were compiled together with video.js but not any source handler that was registered later. 

In addition, the `nativeSourceHandler` property was also mangled which meant that any Techs not compiled with video.js would not be able to *fall-back* to their `nativeSourceHandler` as a last ditch attempt to play the source if no SourceHandler was found that matched the type.

Finally, the tests were written expecting any SourceHandlers that were registered to use the same mangling as video.js instead of testing that the SourceHandler interface worked as expected.

Changes:

**Source**
- No longer explicitly export `registerSourceHandler` since it is properly mixed-in unmangled
- Mixin code now calls functions `canHandleSource`, `dispose`, and `handleSource` explicitly instead of by their mangled names
- A tech's `nativeSourceHandler` property is no longer mangled in the code and is called explicitly unmangled

**Tests**
- `withSourceHandlers` and `registerSourceHandler` are used explicitly instead of using the mangled names
- `canHandleSource` and `handleSource` are quoted to avoid mangling

### DIFF
--- a/src/js/exports.js
+++ b/src/js/exports.js
@@ -183,7 +183,6 @@ goog.exportProperty(vjs.Html5.prototype, 'exitFullScreen', vjs.Html5.prototype.e
 goog.exportProperty(vjs.Html5.prototype, 'playbackRate', vjs.Html5.prototype.playbackRate);
 goog.exportProperty(vjs.Html5.prototype, 'setPlaybackRate', vjs.Html5.prototype.setPlaybackRate);
 // Source Handler Functions
-goog.exportProperty(vjs.Html5, 'registerSourceHandler', vjs.Html5.registerSourceHandler);
 goog.exportProperty(vjs.Html5, 'selectSourceHandler', vjs.Html5.selectSourceHandler);
 goog.exportProperty(vjs.Html5.prototype, 'setSource', vjs.Html5.prototype.setSource);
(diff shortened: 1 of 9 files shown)


heff:
Wow, fun... Really glad we're moving away from GCC.

Should we just say source handlers is a 5.0 feature? If there continues to be updates to it it's gonna be a pain to change in two places.

Otherwise lgtm.

imbcmdth:
@heff If you are asking whether or not this PR should be merged in, I think yes. It is simply fixing the SourceHandler API that was being destroyed by minification - there are not functional changes. 

If you are asking whether we should consider source handlers an *unsupported* or *beta* features on vjs4, then I agree we shouldn't spend time adding or fixing them if any bugs or strange behaviors are discovered. Instead we should direct people wishing to write SHs to 5.

gkatsev:
I'll be merging this in today unless there are objections. @heff?

heff:
None from me.

gkatsev:
Closed by d2de00cae1827f5c81b3d32ebbf321a55f7aa0ab

PR #2349, opened 2015-07-13
Update CONTRIBUTING.md

aseidlitz:
`grunt && grunt test` doesn't produce dist/ folder. Have to explicitly run `grunt dis`. Tested on Mac OS X.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -146,6 +146,7 @@ Build a local copy of video.js and run tests
 ```bash
 grunt
 grunt test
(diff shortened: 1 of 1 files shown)


heff:
Thanks for the help! We do need to do another pass on the contributing guide for 5.0. I don't think this change is necessary though. The dist command is only for generating the official distribution for a rebase. The grunt command should still build the files (under /build/temp/) that contributors need to develop a feature or fix a bug. Let me know if you have any additional thoughts on that. Closing for now.

PR #2353, opened 2015-07-14
Fix inconsistent use of single and double quotes in example

nikolas:


docs/guides/tracks.md
```diff
@@ -22,10 +22,10 @@ Once you have your WebVTT file created, you can add it to Video.js using the tra
 ```html
 <video id="example_video_1" class="video-js vjs-default-skin"  
   controls preload="auto" width="640" height="264"  
-  data-setup='{"example_option":true}'>  
- <source src="http://video-js.zencoder.com/oceans-clip.mp4" type='video/mp4' />  
- <source src="http://video-js.zencoder.com/oceans-clip.webm" type='video/webm' />  
- <source src="http://video-js.zencoder.com/oceans-clip.ogv" type='video/ogg' />  
+  data-setup="{'example_option':true}">
```
gkatsev: given that the value given to `data-setup` is JSON, this cannot have single quotes around the key which is why single quotes were used for the whole value.
The changes to the `type` attribute below are good, though.

heff:
Single quotes are used for type because often the type includes extra codec information that includes double quotes.
```
type='video/mp4; codec="avc1.42E01E, mp4a.40.2"'
```
I don't know if double quotes are required in this case, but it's the standard practice to format it like this. 

Thanks for trying to help though. Any other improvements you find are welcome.

PR #2359, opened 2015-07-15
Create a dummy flash object to test whether flash is being blocked or not and use that in isSupported calc.

bclwhitaker:


heff:
Is this in reference to Firefox/Safari's click-to-enable, or other blocking?

bclwhitaker:
Yes, that's what spurred it on.  With this  code in place, the player would handle that blocking and fallback to HTML5 if the techorder was changed to prefer Flash first (or HLS if using contrib-hls).

In my testing on browserstack this correctly reported in all IE versions. FIrefox, Safari, Chrome also reported properly using my local browsers.

FF 39, Safari 8.0.6, Chrome  43.0.2357.134

bclwhitaker:
This does not detect the Safari power-save.  It does detect the Firefox blocking though.

heff:
Ok, so in Safari Flash would still load and show the power save button to load Flash, yeah?

Can you enable Flash in Firefox after it's been blocked (by clicking a warning or something), or is it permanently blocked? Just wondering if we'd be preventing users from ever loading Flash in Firefox if they're ok with enabling it after it's been blocked.

gkatsev:
Firefox "blocks" it in that it asks the user what they want to do while notifying them with a big red text that there's issues. A user can still continue and use it if they so choose.
It looks like

heff:
Thanks for the demo. So is this what we want? With this change, if you were
to use video.js with an FLV file, it wouldn't even give you this option. It
would just error.

src/js/media/flash.js
```diff
@@ -367,6 +367,18 @@ vjs.Flash.version = function(){
   return version.split(',');
 };
 
+vjs.Flash.isBlocked = function() {
+  var o = document.createElement('object');
```
dmlap: Please no single-letter variables.

bclwhitaker:
@gkatsev is correct that is what it looks like when you don't have a poster.  But if you do, that huge warning isn't present over the video player but you do get a bar at the top -- perhaps that is enough?:
image

FWIW. I found that loading from localhost you don't even get the bar. My initial tests were all from localhost and why I was a bit more concerned cause the player is just unresponsive and the only idication is the red lego block in the address bar:
image

So, I guess we could probably say these notifications are enough and can close this.  Not sure if there is anything we could do to make sure the messaging was always visible when blocked.  I think that would be ideal.

bclwhitaker:
Disregard localhost thing... it appears that messaging bar does not return on sites where you have already answered the notification dialog to continue blocking or allow.

All this to say, I guess I agree that the current messaging is enough.

heff:
@bclwhitaker Thanks for the added info. After some back and forth the general consensus in the witmtn standup this morning was to just let the Firefox warning work as it does. The poster covering the main message is a new detail, but I agree that the current messaging is probably still enough.

I'll close this but I imagine we'll continue to get additional feedback from users and we can continue to discuss this.

PR #2362, opened 2015-07-16
Make videojs Compatible with ReactJS

TissueFluid:
## Problem
when using ReactJS along with videojs, there will be two elements that has the same `data-reactid` attribute. But ReactJS only allows unique `data-reactid`.

## Solution
As is shown in `src/player.js`, at function `createEl()`:  videojs would create a `div` element outside each `video` element. And it copies most of the attributes of the `video` element to the `div`, which is depicted as the code below.

```javascript
if (attr === 'class') {
    el.className = attrs[attr];
} else {
    el.setAttribute(attr, attrs[attr]);
}
```

As we all know, ReactJS only allows unique `data-reactid` attribute. Thus, in the code above, we should do some check to make their `data-reactid` different.

```javascript
if (attr === 'class') {
    el.className = attrs[attr];
} else if (attr === 'data-reactid') {
    el.setAttribute(attr, attrs[attr] + '_foo');
} else {
    el.setAttribute(attr, attrs[attr]);
}
```


mmcc:
Thanks for the PR!

My gut instinct is that this doesn't belong in core because it's a workaround for a specific framework. Especially given the number of client-side frameworks out there (and growing), I cringe at the thought of having logic specific to each one sprinkled throughout the code base.

Perhaps this is a better broken out as a plugin? Is there other React-specific logic that could help the VJS experience that could be added to such a plugin?

Closing this because of the reasons listed above, but let's definitely continue the conversation in the comments. Thanks again.

dmlap:
Alternative proposal I've been thinking about for awhile: allow a `<div>` to initialize videojs instead of the normal `<video>`. React users could mark the subtree at the `<div>` as "unsafeHTML" (or whatever React's term is) and React would ignore subtree modifications. This would have the side-benefit of getting rid of the "flash of unstyled video player" while the video.js CSS loads.

mmcc:
@dmlap and I just talked this over in the IRC channel and I think we came to a potentially happy medium.

Our idea is to allow for instantiation via the `<video>` tag or `<video-js>` tag (the name of the tag may change, but same idea). The API would be just like the current `<video>` tag in every way, except that as @dmlap said, it would allow React users to ignore changes within the entire div since we would be able to nest everything beneath the custom element without "breaking out of our box" like we have to now when we create the wrapper div.

Make sense? @heff? It's not a web component, but it's like..1/10th of the way there :)

heff:
Yeah, seems like a decent step towards a web component, especially if it helps React. We could just make the leap and make it a web component. It would mean you could do document.createElement('video-js') or just use `<video-js></video-js>` in the page and it would automatically kick off the player initialization without needing our setup script. But that would also require the webcomponent.js and IE10+. We should probably wait to make the jump, but also try to make sure we don't get in our own way later.

Either way it's gonna take some effort to duplicate the video element API on the video-js element, so unless we're going to jump right into that, we might want a shorter term solution.

dmlap:
@heff: while I'm all for going the web component route, I think Matt's proposal offers some pretty nice benefits without requiring all the changes necessary to properly componentize everything. If we treat `<video-js>` just like the player div gets treated today and add the bit of code necessary to create a video element when one isn't already available, we make ourselves much more React-friendly and get rid of the unstyled player problem. It seems worth considering on its own merits to me.

heff:
Yeah, that makes sense to me. I read the "The API would be just like the current <video> tag in every way" line and assumed it included mimicking the video element API, but realizing now that's not the case.

+1

LeoZeda:
have this problem,   "video.js": "^5.10.8",

PR #2366, opened 2015-07-16
Update native controls functionality

### LINKED ISSUE #2365: using custom controls and poster on iphones
Loading up the player on iphones will currently display our custom controls and our custom poster element. This is an issue because on some iOS versions, our big play button will not work. It seems like on iOS 8, it might work under some circumstances, but I'm not certain it will work across the board.
The best solution is probably to just hide our poster and controls on iphones.
Thoughts, @videojs/core-committers

gkatsev:
Fixes #2365.
Also, wrap the trigger in a setTimeout. This is because otherwise, the
player will never be able to hear this event since it happens in the
constructor of the tech and the player adds listeners after the tech has
been created.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -73,8 +73,10 @@ class Html5 extends Tech {
     // Our goal should be to get the custom controls on mobile solid everywhere
     // so we can remove this all together. Right now this will block custom
     // controls on touch enabled laptops like the Chrome Pixel
-    if (browser.TOUCH_ENABLED && options.nativeControlsForTouch === true) {
-      this.trigger('usenativecontrols');
+    if ((browser.TOUCH_ENABLED && options.nativeControlsForTouch === true) || browser.IS_IPHONE) {
+      this.setTimeout(function() {
+        this.trigger('usenativecontrols');
+      }, 0);
     }
 
     this.triggerReady();


src/js/tech/html5.js
```diff
@@ -73,8 +73,10 @@ class Html5 extends Tech {
     // Our goal should be to get the custom controls on mobile solid everywhere
     // so we can remove this all together. Right now this will block custom
     // controls on touch enabled laptops like the Chrome Pixel
-    if (browser.TOUCH_ENABLED && options.nativeControlsForTouch === true) {
-      this.trigger('usenativecontrols');
+    if ((browser.TOUCH_ENABLED && options.nativeControlsForTouch === true) || browser.IS_IPHONE) {
+      this.setTimeout(function() {
```
gkatsev: I'm not certain this is the best way of doing it.
One other way I was thinking is moving the `usingNativeControls` method/info into the tech from the player.

pam:

Commit: a75fabb1a81b4d768e130b1350d156576412d989
Build details: https://travis-ci.org/pam/video.js/builds/71307444

src/js/tech/html5.js
heff: Got it, so we create the tech and then add the usenativecontrols listener.

"Native controls" to the player is really just controls: true/false to the tech. In player.loadTech could we check `player.techGet('controls')` and if it's true assume we're using native controls? Does that work in our favor with the iPhone controls-always-true issue?

heff:
Overall it seems like a bummer to revert the custom controls on iPhone, even though they don't do anything after play(). If it is breaking places I think it's fine to change this, but I think it'd also be good to have a bit more formal reduced test case and matrix of versions where it breaks so we know when we can undo this. On my iOS 8 phone I haven't been able to get it to break yet on videojs.com.

gkatsev:
iOS 7 definitely has issues around it. It seems like iOS8 doesn't even though it still says that controls are enabled while the controls attribute is disabled. On iOS8 I can press either our big play button or the middle of the video which ends up tapping apple's big play button.
While we *could* just disable it for iphones older than iOS8, I'm inclined to disable it for all iphones because we can't be sure that it will work in all scenarios and the way that iphones work is so magical anyway. [@dmlap agrees.](https://github.com/videojs/video.js/issues/2365#issuecomment-122146197).
This doesn't mean we can't revisit this decision at a later date when we're more sure of all the parameters and usecases around it.

heff:
Yeah, that's ok with me. There's still the implementation comment to discuss.

gkatsev:
Oops didn't mean to close this one.

src/js/tech/html5.js
dmlap: If just set `controls = true` here instead of the extra indirection of the event, I'm in favor.

dmlap:
Other than the question about how to toggle controls, LGTM.

src/js/tech/html5.js
gkatsev: Are you saying to move the logic out of Tech and in here?

heff: I'm saying in Player in `loadTech`, after the tech loads or after tech ready, something like this:
```
if (player.techGet('controls') === true) {
  this.usingNativeControls(true);
}
```

Instead of throwing the delayed event.

gkatsev: That makes sense, I'll update it.

gkatsev:
Closing in favor of a new rebased PR.

PR #2373, opened 2015-07-17
Added query parameters to parseUrl()

forbesjo:
This enhancement is to add query parameters to the object returned by parseUrl(). I think this would be useful to avoid duplicate query parameter code.

### DIFF
--- a/src/js/lib.js
+++ b/src/js/lib.js
@@ -691,7 +691,7 @@ vjs.parseUrl = function(url) {
       var parameters = {}, urlParameters, urlParameter, i , name, value;
 
       if (url.indexOf('?') === -1) {
-        return undefined;
+        return parameters;
       }
 
       urlParameters = url.split('?');
@@ -746,10 +746,7 @@ vjs.parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
-  parameters = getQueryParameters(url);
-  if (parameters) {
-    details.parameters = parameters;
-  }
+  details.parameters = getQueryParameters(url);
 
   return details;
 };
--- a/src/js/lib.js
+++ b/src/js/lib.js
@@ -686,7 +686,25 @@ vjs.getAbsoluteURL = function(url){
  * @return {Object}     An object of url details
  */
 vjs.parseUrl = function(url) {
-  var div, a, addToBody, props, details;
+  var div, a, addToBody, props, details,
+    getQueryParameters = function(url) {
+      var parameters = {}, urlParameters, name, value;
+
+      urlParameters = url.split('?');
+      if (urlParameters.length !== 2) {
+        return undefined;
+      }
+
+      urlParameters = urlParameters[1].split('&');
+      urlParameters.map(function(urlParameter) {
+        urlParameter = urlParameter.split('=');
+        name = urlParameter[0];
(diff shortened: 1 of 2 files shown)


src/js/lib.js
```diff
@@ -686,7 +686,25 @@ vjs.getAbsoluteURL = function(url){
...
+    getQueryParameters = function(url) {
+      var parameters = {}, urlParameters, name, value;
+
+      urlParameters = url.split('?');
+      if (urlParameters.length !== 2) {
+        return undefined;
+      }
+
+      urlParameters = urlParameters[1].split('&');
+      urlParameters.map(function(urlParameter) {
```
gkatsev: `map` isn't available in IE8.

src/js/lib.js
```diff
@@ -727,6 +745,10 @@ vjs.parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
+  if(getQueryParameters(url)){
```
gkatsev: We should cache the return value from this, so, we don't process it twice.

gkatsev:
Why not have the `parseUrl` and `getQueryParameters` always return an object? Whether it's empty or not. This way, end users don't necessarily have to do an existence check for whether the object from `parseUrl` has a `parameters` property and only need to worry about whether they have the query param they really care for.
For example, this:
```js
var lang = videojs.parseUrl('http://example.com/video.mp4?language=en').parameters.language;
if (lang) {
  // do stuff with lang
}
```
Compared to this:
```js
var url = videojs.parseUrl('http://example.com/video.mp4?language=en').;
if (url.parameters && url.parameters.language) {
  // do stuff with url.parameters.language
}
```

dmlap:
Browsers already do a thorough job of parsing `window.location`. Why not just parse the query parameters and leave it up to the native functionality for the rest?

gkatsev:
Browsers don't parse query parameters. They leave it as a giant string on `window.location.search`, you still need to find your specific parameters from it.

heff:
Does this need to be against stable and go out as a patch release, or can we switch it over to 5.x?

gkatsev:
We have a bug that it would be nice if this is available. For 5.0, we could also use the [`qs`](https://www.npmjs.com/package/qs) module or just pass [`true` to node's `url.parse` as the second argument](https://nodejs.org/api/url.html#url_url_parse_urlstr_parsequerystring_slashesdenotehost) so it will parse the query params for us.

heff:
Yeah, the `url.parse` route was what I was thinking too, for 5.0.

If there's a related bug in 4.x it'd be good to have a little more description about it here.

dmlap:
I feel like a full URL parser is heading towards overkill. I know we're trying to provide more out-of-the-box for 5.0 but `window.location` really does the job for most non-query-param use-cases IMO.

forbesjo:
In the BC player there's an incorrect regex for the language parameter, I thought that adding this enhancement instead of fixing the regex for that particular parameter would be more useful. 

If `url.parse` is going to be used for 5.0 then we can ignore this PR, I'll just use the modified regex and use the output from urlParse() when we upgrade to 5.0.

forbesjo:
Closing, better to not reinvent the wheel.

PR #2375, opened 2015-07-17
Feature/hide poster on autoplay

marguinbc:
Adding a new CSS class vjs-autoplay which will be applied to hide vjs-big-play-button and vjs-poster when player.options.autoplay(true). This is to prevent a flash of the poster image before preroll ads load and also during normal autoplay operation.


mmcc:
This PR is trying to merge master into stable, which we're not ready to do yet. If you want to submit a PR against 4.x, please create your branch from stable and create the PR against stable. If you'd like to make the PR for 5.0, please create the branch from master and submit against master.

mmcc:
(thanks for the PR, of course, just need to make sure we don't pull in things we aren't ready for yet)

Issue #2383, opened 2015-07-20
Component.extend() does not call subclass constructor


dmlap:
1. Create a subclass of Tech using `Tech.extend()` with some fancy constructor logic
1. Create a new instance of your subclass

Expected: your fancy constructor logic is run
Actual: Component's constructor is the only constructor that ran.

heff:
Extending a component [changed](https://github.com/videojs/video.js/wiki/5.0-Change-Details#switched-to-es6-classes-updated-how-you-subclass-components). Use the `videojs.extends()` function now.

We should add a warning to `Component.extend`. It could be possible to shim the old way in 5.0 if it's worth it, but wouldn't be very pretty.

dmlap:
The base tech tests are [using the deprecated way](https://github.com/videojs/video.js/blob/master/test/unit/tech/tech.test.js#L105), then.

heff:
Yeah, that needs to be updated. And I believe that should be failing in IE8.

misteroneill:
@dmlap To be clear, how were you adding "fancy constructor logic" via `Component.extend` - do you mean passing an `init` method to `extend` (in the old style)?

heff:
@misteroneill that would be my guess what he means. In the old way it's the only way to do that.

misteroneill:
After discussing in person, it seems his concern was related to doing `Tech.extend();` instead of `videojs.extends(Tech)`.

Making `Component.extend` work like `videojs.extends` would be a change to a deprecated function; so, we thought the best way forward was to make sure uses of `Component.extend` were removed and a deprecation warning was added.

heff:
Sounds good. And just to be clear, Tech.extend (and extend for any other component besides Component) aren't just deprecated, they no longer work.

PR #2390, opened 2015-07-21
More Explicitly Deprecate Component.extend

misteroneill:
This addresses #2383 by making sure `Component.extend` uses are removed from the source.

- The `Component.extend` method still exists for backward compatibility.
- Adds a deprecation warning to that method.
- Removes a commented-out line from that method.
- Component documentation is updated to reflect the new style (`videojs.extends()`).
- Fixes a typo.

### DIFF
--- a/test/unit/tech/tech.test.js
+++ b/test/unit/tech/tech.test.js
@@ -1,6 +1,6 @@
 var noop = function() {}, clock, oldTextTracks;
 
-import videojs from '../../../src/js/video.js';
+import extendsFn from '../../../src/js/extends.js';
 import Tech from '../../../src/js/tech/tech.js';
 import { createTimeRange } from '../../../src/js/utils/time-ranges.js';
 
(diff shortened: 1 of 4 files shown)


test/unit/tech/tech.test.js
```diff
@@ -1,5 +1,6 @@
 var noop = function() {}, clock, oldTextTracks;
 
+import videojs from '../../../src/js/video.js';
```
gkatsev: over here since we don't need all of videojs, we can import just the `extendsFn` itself: https://github.com/videojs/video.js/blob/dd9930b6c68a1016475b35d1f05c4421f83649bb/src/js/video.js#L18

gkatsev:
Small comment, LGTM otherwise.

heff:
I agree with @gkatsev's comment, but otherwise looks good to me.

test/unit/tech/tech.test.js
misteroneill: :+1:

gkatsev:
LGTM

heff:
@gkatsev feel free to merge in.

heff:
Thanks @misteroneill!

Issue #2367, opened 2015-07-16
Cannot addTextTracks after switching off of the HTML tech


dmlap:
#### Steps to Reproduce
1. Create a player with a text track
1. Load it on a browser that will use the HTML tech
1. Load a video that will force a switch to the flash tech
1. Call player.addTextTrack()

An error will be thrown because the flash tech will attempt to use the native text track list created by the HTML tech.

gkatsev:
What should we do with the available text tracks? Delete them or recreate them? I guess this does end up falling in with #2369

heff:
Maybe when we unload a tech we should record all the values of the existing tracks, so they can be added back as native or custom tracks when the next tech is loaded? That would seem to at least match how the video element works. Tracks don't disappear on a source change. You could add all the tracks first before adding a source.

gkatsev:
It kind of is doing so. I guess it needs to do it differently. The issue here is that we're literally grabbing the text track list element from the previous tech and trying to use it in the next tech which doesn't work if you're going from html to flash.

heff:
Yeah, sounds like we need to record the values (all attributes and cues), not keep the tracks themselves around.

gkatsev:
Do we want to preserve the in-band metadata tracks?
I guess I'll just preserve everything in my first pass and we can decide on what to keep or not later in the PR.

heff:
Ultimately I don't think we should. If the user didn't add them then they wouldn't know to remove them when changing the source.

gkatsev:
So, I just realized what we're doing here isn't what we actually want to do.
TextTrackList has several events listeners that can be added to it. If we copy it over between techs, we lose that event listener. Natively, you would still have the same listener applied after you change sources.
One thing I can think of for this is to always use our emulated text track list so that we always get events but have it proxy to the native text track list object if available and clean itself out on tech change.

gkatsev:
And this actually would apply to the text tracks themselves, too.

heff:
> One thing I can think of for this is to always use our emulated text track list so that we always get events 

Yeah, I wondering if we were going to need to go that direction. It sounds ok to me, but I defer to your judgement.

> And this actually would apply to the text tracks themselves, too

You're saying we would always need to use our emulated tracks? Why's that?

gkatsev:
tl;dr: it depends on how disruptive we want to be to users while still using as much native functionality as possible.

-------------------------------------------------------
[...]

PR #2391, opened 2015-07-21
Always emulated text track list

gkatsev:
Fixes https://github.com/videojs/video.js/issues/2367

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -67,6 +67,10 @@ class Html5 extends Tech {
 
     if (this.featuresNativeTextTracks) {
       this.on('loadstart', Fn.bind(this, this.hideCaptions));
+
+      this.handleTextTrackChange_ = Fn.bind(this, this.handleTextTrackChange);
+      this.handleTextTrackAdd_ = Fn.bind(this, this.handleTextTrackAdd);
+      this.handleTextTrackRemove_ = Fn.bind(this, this.handleTextTrackRemove);
       this.proxyNativeTextTracks_();
     }
 
@@ -88,13 +92,18 @@ class Html5 extends Tech {
    */
   dispose() {
     let tt = this.el().textTracks;
-
-    tt.removeEventListener('change', Fn.bind(this, this.handleTextTrackChange));
-    tt.removeEventListener('addtrack', Fn.bind(this, this.handleTextTrackAdd));
-    tt.removeEventListener('removetrack', Fn.bind(this, this.handleTextTrackRemove));
+    tt.removeEventListener('change', this.handleTextTrackChange_);
+    tt.removeEventListener('addtrack', this.handleTextTrackAdd_);
+    // events get triggered asynchronously and we want to get all the `removetrack` events
+    // so we need to remove the `removetrack` handler asynchronously.
+    setTimeout(() => {
+      tt.removeEventListener('removetrack', this.handleTextTrackRemove_);
+    }, 0);
 
     Html5.disposeMediaElement(this.el_);
     super.dispose();
+
+
   }
 
   /**
@@ -192,9 +201,9 @@ class Html5 extends Tech {
   proxyNativeTextTracks_() {
     let tt = this.el().textTracks;
 
-    tt.addEventListener('change', Fn.bind(this, this.handleTextTrackChange));
-    tt.addEventListener('addtrack', Fn.bind(this, this.handleTextTrackAdd));
-    tt.addEventListener('removetrack', Fn.bind(this, this.handleTextTrackRemove));
+    tt.addEventListener('change', this.handleTextTrackChange_);
+    tt.addEventListener('addtrack', this.handleTextTrackAdd_);
+    tt.addEventListener('removetrack', this.handleTextTrackRemove_);
   }
 
   handleTextTrackChange(e) {
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -94,16 +94,15 @@ class Html5 extends Tech {
     let tt = this.el().textTracks;
     tt.removeEventListener('change', this.handleTextTrackChange_);
     tt.removeEventListener('addtrack', this.handleTextTrackAdd_);
+
+    Html5.disposeMediaElement(this.el_);
+    super.dispose();
+
     // events get triggered asynchronously and we want to get all the `removetrack` events
     // so we need to remove the `removetrack` handler asynchronously.
     setTimeout(() => {
       tt.removeEventListener('removetrack', this.handleTextTrackRemove_);
     }, 0);
-
-    Html5.disposeMediaElement(this.el_);
-    super.dispose();
-
-
   }
 
   /**
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -721,12 +721,12 @@ class Html5 extends Tech {
 
     this.remoteTextTracks().removeTrack_(track);
 
-    tracks = this.el()['querySelectorAll']('track');
+    tracks = this.el().querySelectorAll('track');
 
-    for (i = 0; i < tracks.length; i++) {
-      if (tracks[i] === track || tracks[i]['track'] === track) {
-        tracks[i]['parentNode']['removeChild'](tracks[i]);
-        break;
+    i = tracks.length;
+    while (i--) {
+      if (track === tracks[i] || track === tracks[i].track) {
+        this.el().removeChild(tracks[i]);
       }
     }
   }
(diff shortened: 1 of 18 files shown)


pam:

Commit: f198faabffa1f50f07fe8b7ce7ad4b3710a91e61
Build details: https://travis-ci.org/pam/video.js/builds/72025560

gkatsev:
I'll tackle tests next but it's working for me when switching from html to flash and back.

src/js/tech/html5.js
```diff
@@ -86,6 +87,12 @@ class Html5 extends Tech {
    * @method dispose
    */
   dispose() {
+    let tt = this.el().textTracks;
+
+    tt.removeEventListener('change', Fn.bind(this, this.handleTextTrackChange));
```
heff: If `el().textTracks` is a native list then I don't know if the Fn.bind is going to do what you want here. The remove-by-same-id only works with our internal event listeners. The native removeEventListener still looks for the same function.

src/js/tracks/text-track-list-converter.js
```diff
@@ -0,0 +1,47 @@
...
+        endTime: cue.endTime,
+        text: cue.text,
+        id: cue.id
+      };
+    }),
+    src: track.src
+  };
+};
+
+let textTracksToJson = function(tech) {
```
heff: The naming of `textTracksToJson` compared to `ttToJson ` could be a little more different.

src/js/tech/html5.js
gkatsev: Yep, you're right. I keep getting more and more output when switching techs.

src/js/tech/tech.js
```diff
@@ -238,6 +238,14 @@ class Tech extends Component {
    * @method dispose
    */
   dispose() {
+    // clear out text tracks because we can't reuse them between techs
```
heff: Does this mean we're not maintaining tracks between tech/source changes?

heff:
One question around tech changes but other than that looks really good to me.

src/js/tech/tech.js
gkatsev: No, we cannot maintain them. If we're switching from Html to Flash, we're moving from track elements to emulated text tracks. And then if if we're going from Flash back to Html, we would want to clear out our emulated text tracks and use the native track elements.

If you're just switching sources but staying in the same tech, you should keep the tracks.

heff: Ok so help me understand what this means for users. What are caveats of using text tracks that we need to tell them?

gkatsev: It would mean that users would need to re-add `cuechange` events to the text tracks each time the tech changes. Also, events that are added to the cues themselves would need to be re-added, though, we don't currently support that for the emulated cues.

src/js/tech/html5.js
gkatsev: I have a fix for this locally but for some odd reason it only wants to work when I'm setting through the code for removing the old text tracks...

src/js/tech/tech.js
heff: So the text track list (player.textTracks) will survive. We'll remove the old text tracks (which will trigger change events) and add new text tracks (which will also trigger change events). Users should listen for those events to handle re-adding listeners to the new tracks. Is that right?

That doesn't sound too bad.

gkatsev: Yep, sounds right.

src/js/tech/html5.js
gkatsev: So, the issue seems that if we remove the event listener in the same event tick as removing the track, we don't actually get the `removetrack` event. Looking at the [spec](https://html.spec.whatwg.org/multipage/embedded-content.html#sourcing-out-of-band-text-tracks:event-media-removetrack), it seems like the `removetrack` events are queued to fire when tracks are removed, so, it makes sense for us to miss them.

gkatsev: I believe the solution in this case really is a setTimeout around the removal of tracks. Either that or create a remove handler that gets added on `addtrack` and bound to a particular track which then gets removed when that particular track gets removed. But I think that solution might be too clever.

heff: Could we queue our own removetrack event in that case? Seems like handling that asynchronously could get messy when switching to the new tech.

gkatsev: I just realized we could probably just clean out our emulated text track list manually. However, if we remove the event listener asynchronously, it works fine too. Or, we can have the `removetrack` handler be smart and get added on `addtrack` and removed on `removetrack` if there are no longer any text tracks (which is what I tried to describe above).

heff: I like the idea of something simple and less async, but whatever you think
works best.

-- 
Sent from mobile

gkatsev:
I updated things. I currently see a weird issue where when you switch back to the html5 tech, to get the non-current-locale text tracks to show, you need to select it twice. Not sure what is going on there.

pam:

Commit: 6d75a257bd2a0c603d250d774f7f8152f8e8cfb9
Build details: https://travis-ci.org/pam/video.js/builds/72875915

gkatsev:
I decided to not try and do anything asynchronous and just clear out our emulated text track list manually in dispose since nothing should be coming along across the tech switch from html5->flash

pam:

Commit: 486dc454fa306c6fa6fe8c665ac9df7a0f34ce58
Build details: https://travis-ci.org/pam/video.js/builds/72882669

test/unit/tech/html5.test.js
```diff
@@ -178,3 +178,48 @@ test('native source handler canHandleSource', function(){
   // Reset test video canPlayType
   Html5.TEST_VID.canPlayType = origCPT;
 });
+
+if (Html5.supportsNativeTextTracks()) {
```
gkatsev: these only run if we are using and support native text tracks, so, I'm only running the tests in those browsers as well because it'll fail otherwise.

test/unit/utils/dom.test.js
```diff
@@ -2,6 +2,8 @@ import document from 'global/document';
 import * as Dom from '../../../src/js/utils/dom.js';
 import TestHelpers from '../test-helpers.js';
 
+q.module('dom');
```
gkatsev: makes the output make more sense. Since I updated qunit to 1.18 there was an error that looked like it was in text tracks but was actually in url.

test/unit/utils/url.test.js
```diff
@@ -30,9 +32,11 @@ test('should strip port from hosts using http or https', function() {
   };
 
   url = Url.parseUrl('/domain/relative/url');
-  ok(!(/.*:80$/).test(url.host), ':80 is not appended to the host');
 
   document.createElement = origDocCreate;
+
+  ok(!(/.*:80$/).test(url.host), ':80 is not appended to the host');
```
gkatsev: our overriding of createElement here broke qunit. So, I moved the `ok` to after we reset `createElement` back to the original value.

src/js/tech/html5.js
```diff
@@ -86,6 +91,22 @@ class Html5 extends Tech {
...
+
+    // remove native event listeners
+    tt.removeEventListener('change', this.handleTextTrackChange_);
+    tt.removeEventListener('addtrack', this.handleTextTrackAdd_);
+    tt.removeEventListener('removetrack', this.handleTextTrackRemove_);
+
+    // clearout the emulated text track list.
+    let i = emulatedTt.length;
+
+    while (i--) {
```
gkatsev: I decided to not do any async shenanigans. Even though, those async shenanigans might be more technically correct, this one is good enough and much simpler.

src/js/tracks/text-track-list-converter.js
gkatsev: Done. Also, writing tests for this.

gkatsev:
I think I may have fixed the issue that text tracks have had with safari in this PR.

test/unit/tech/flash.test.js
```diff
@@ -70,6 +70,7 @@ test('dispose removes the object element even before ready fires', function() {
   mockFlash.off = noop;
   mockFlash.trigger = noop;
   mockFlash.el_ = {};
+  mockFlash.textTracks = () => ([]);
```
gkatsev: I think this is what was causing the issues in safari that we're seeing in master and elsewhere.

heff: Please can we fix this? How do we fix this?

heff: Oh, you did. Sweet. :)

gkatsev:
Actually, looks like selecting the captions twice is caused by what #2412 fixes.

src/js/tracks/text-track-list-converter.js
```diff
@@ -0,0 +1,47 @@
+let trackToJson = function(track) {
```
dmlap: This function and file could use an explanatory comment.

dmlap:
Other than my documentation comment, LGTM

heff:
2nd dmlap. Lgtm!

dmlap:
Superseded by #2425

PR #2392, opened 2015-07-22
Add sync option to ready()

### LINKED ISSUE #2326: Setting the src is not synchronous
Even when the tech is ready, setting the source doesn't take effect at the tech level for one tick because it's [wrapped in a call to ready()](https://github.com/videojs/video.js/blob/26a78627f1f231dbbe0fa1f7441bfc1b27893be5/src/js/player.js#L1521)

heff:
Should fix #2326

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -585,7 +585,7 @@ class Component {
    * @param  {String|Component} first   The event type or other component
    * @param  {Function|String}      second  The event handler or event type
    * @param  {Function}             third   The event handler
-   * @return {Component} 
+   * @return {Component}
    * @method on
    */
   on(first, second, third) {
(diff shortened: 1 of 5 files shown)


pam:

Commit: 44a6d8d8c4a48ec06688a5f61eab9ab178bdfa51
Build details: https://travis-ci.org/pam/video.js/builds/72042662

dmlap:
Nice. How did you look for `ready()` calls that needed to be converted to the sync form?

heff:
`ready\([^\)]`

I only added sync for the ones that needed it or seemed like they would be better with it.

heff:
Sorry, that could be more detailed. There were obvious ones where we were just waiting to send the call because the tech wasn't ready, but ideally they'd be synchronous. Then I also added sync to the ones that seemed like they could be player loading speed related. Otherwise if they looked like just needed to happen some time after the ready event, I left them alone.

Issue #2311, opened 2015-07-07
Restore some globals


dmlap:
It seems like they just save and retrieve the component from a private variable. Any background on why this change was made? Besides making it a bit more verbose to reference things like the techs it also makes enumerating available components seem like something that is discouraged.

gkatsev:
One of the main benefits is that it allows us to better encapsulate the internal modules and not have circular dependencies.
Just having users of videojs attach variables to the `videojs` object shouldn't be encouraged or supported.
It's possible that something like `listComponents()` might be useful.

dmlap:
I think you dodged the question :) Why shouldn't it be encouraged or supported?

heff:
When switching to modules this was our major circular dependency. Component children were added by name off the videojs object. In Component, `this.addChild(videojs[childName])`. And videojs required Component to export it as `videojs.Component`. That made structuring files as CommonJS modules very difficult.

The solution was to make it so Component didn't look for other components on `videojs` and instead moved component registration to itself. It really is much cleaner this way.

It doesn't have to be a function. We could have other components just add themselves to Component.components, and the export components, `videojs.components -> Component.components`, the same way we're currently exporting getComponents. The benefit I see of having it as a function is future flexibility. We can change the internal workings of component registration down the line without changing the API, we could throw warning messages during registration, throw deprecation messages when someone gets a deprecated component. None of that is required though. @gkatsev may have additional benefits in mind.

heff:
I'm gonna close this since it's a question but we can continue to discuss and reopen if a change is decided on.

gkatsev:
Also, knowing when a component registered with us could be beneficial. Don't really have anything else in mind right now but we could add other functionality around it at a later date without making it a breaking change. If Users just add it directly to the `videojs` object we don't know.

Also, this falls under the "don't modify objects you don't own" mantra we should be striving for. Users of videojs don't own the videojs object, so, they shouldn't be adding random properties to it themselves.

dmlap:
[...]

PR #2395, opened 2015-07-22
Restore options/players Globals

misteroneill:
Addresses #2311.

This restores the `options` and `players` properties of `videojs`.

Browsers with `Proxy` support (Firefox and IE Edge at the moment) will additionally log a warning on both get/set operations only through these specific properties. That is, modifying `videojs.players` will log a deprecation warning where internal modifications to `Player.players` will not.

Both cases - using `Proxy` and not - are tested.

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -99,17 +99,18 @@ var videojs = function(id, options, ready){
  * @private
  * @param {String} name The key name for the deprecated property.
  * @param {Object} target The target object.
+ * @param {Object} messages String messages to display from a Proxy, mapping to get/set operations.
  * @return {Object} A Proxy if supported or the `target` argument.
  */
-const createDeprecationProxy = (name, target) => {
+const createDeprecationProxy = (name, target, messages) => {
   if (typeof Proxy === 'function') {
     videojs[name] = new Proxy(target, {
       get: function(obj, key) {
-        log.warn(`access to videojs.${name} is deprecated; getting ${key}`);
+        log.warn(messages.get);
         return obj[key];
       },
       set: function(obj, key, value) {
-        log.warn(`modification of videojs.${name} is deprecated; setting ${key}`);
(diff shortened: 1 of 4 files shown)


src/js/video.js
```diff
@@ -89,6 +89,35 @@ var videojs = function(id, options, ready){
...
+ * supported. Browsers that support Proxy will log a deprecation warning for
+ * directly accessing or modifying a deprecated object.
+ *
+ * @private
+ * @param {String} name The key name for the deprecated property.
+ * @param {Object} target The target object.
+ * @param {Object} messages String messages to display from a Proxy, mapping to get/set operations.
+ * @return {Object} A Proxy if supported or the `target` argument.
+ */
+const createDeprecationProxy = (name, target, messages) => {
```
gkatsev: can we have this return either the proxy or just the target?
Then below in your usage we could do
```js
videojs.players = createDeprecationProxy(Players.players, {
  get: 'access to videojs.players is deprecated; use videojs.getPlayers instead',
  set: 'modification of videojs.players is deprecated'
});
```

Also, we should consider putting this function into its own file in `src/utils/` so that others could possibly use it as well.

misteroneill: :+1: on both.

The `name` argument is a holdover from the original implementation, which generated the log messages. Since that doesn't happen anymore, it's superfluous and your suggestion is better!

misteroneill:
Moved `createDeprecationProxy` into a module and changed its signature a bit (removing the `name` argument).

gkatsev:
LGTM

mmcc:
@pam retry

mmcc:
@pam retry

heff:
Nit-picky but I'd prefer the warning messages to have a capital first letter.

Looks good to me.

misteroneill:
@heff Sounds good (and done).

PR #2398, opened 2015-07-22
Fix potential triggerReady infinite-loop

imbcmdth:
If the `readyQueue_` is not reset *before* processing, a `triggerReady` later on in the call graph potentially results in an infinite loop.


heff:
At first I was wondering if this would affect ready handlers that were added during the ready loop, but since we set isReady to true first it won't.

LGTM

imbcmdth:
@heff That nuance is precisely what threw my off when looking over vjs5's implementation!

PR #2401, opened 2015-07-23
Reset seekable function when no source handler in tech

nickygerritsen:
The `seekable` function was only reset back to its original implementation when a source handler was set.

However, the seekable function is overwritten even if no source handler is set in `setSource`.

This resulted in a bug in the Flash tech when calling `player.src()` more than once during the lifetime of the player.

This should fix this issue.

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -537,6 +537,17 @@ Tech.withSourceHandlers = function(_Tech){
     return '';
   };
 
+  let originalSeekable = _Tech.prototype.seekable;
+
+  // when a source handler is registered, prefer its implementation of
+  // seekable when present.
+  _Tech.prototype.seekable = function() {
+    if (this.sourceHandler_ && this.sourceHandler_.seekable) {
+      return this.sourceHandler_.seekable();
+    }
+    return originalSeekable.call(this);
+  };
+
    /*
     * Create a function for setting the source using a source object
     * and source handlers.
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Should we set the original seekable back to `undefined` or `null` again? I don't think it really hurts if we don't, but it does clean up the state better.

heff:
Good catch.

Actually, I don't think we need to be resetting seekable when a source handler is disposed. After a second look at that code, `setSource` on a source-hander-tech is always used, so we're always just resetting seekable to the same value. We could instead be overriding seekable when the source handler functions are initially added.

In `Tech.withSourceHandlers`...
```js
let originalSeekable = _Tech.prototype.seekable;

// when a source handler is registered, prefer its implementation of
// seekable when present.
_Tech.prototype.seekable = function() {
  if (this.sourceHandler_ && this.sourceHandler_.seekable) {
    return this.sourceHandler_.seekable();
  }
  return originalSeekable.call(this);
};
```

@dmlap see any reason why we shouldn't do that?

dmlap:
I did have to fiddle around with the location of that setter a bit but can't think of a reason it wouldn't work in `withSourceHandlers`.

nickygerritsen:
Should I update this PR to put the code in `Tech.withSourceHandlers` or will one of you guys change this :)?

heff:
It'll get merged in faster if you can do it, but otherwise another contributor will get to it at some point.

nickygerritsen:
Ok I'll change it tomorrow. Will probably be a new PR as I accidentally did this one on my own master branch which is not really handy

nickygerritsen:
OK this should be the fix then :)

heff:
LGTM! @dmlap confirm?

-- 
Sent from mobile

dmlap:
Yep. Thanks @nickygerritsen!

Issue #2402, opened 2015-07-23
Backward Compatibility Audit


misteroneill:
I'm doing a run-through of the 4.x documentation and objects to look for properties and methods that are missing from the 5.x work or whose fate is otherwise unclear or undocumented in the wiki. A lot of this is up for discussion, but I wanted to get it in one place where we can all see it.

### Globals

Global | Resolution
--- | ---
`vjs` | _Document_

### `videojs` Properties

Property | Resolution
--- | ---
`JSON` | _Document_
`TOUCH_ENABLED` | _Restore_
`cache` | _Ignore_
`media` | _Ignore_
`options` | _Restore_ #2395 
`players` | _Restore_ #2395 
`util` | _Document_ No longer needed.

### `videojs` Methods

Method | Resolution
--- | ---
`bind` | _Document_ Wiki says removed, but was restored.
`createTimeRange` | _Document_ Wiki says removed, but was restored.
`parseUrl` | _Restore_
`round` | _Restore_
`trim` | _Restore_
`util.mergeOptions` | _Document_ Now top-level.

### Classes

Class | Resolution
--- | ---
`CoreObject` | _Document_
`EventEmitter` | _Restore_ as `EventTarget`
`MediaTechController` | _Document_

### Components

Component | Resolution
--- | ---
`SeekHandle` | _Document_ Moved to CSS
`SliderHandle` | _Document_ Moved to CSS
`VolumeHandle` | _Document_ Moved to CSS

### Other

Object | Resolution
--- | ---
`Component#init` | _Remove_ Shim and deprecate.

### Player

Property/Method | Resolution
--- | ---
`cancelFullScreen` | _Remove_ Had been deprecated in 4.x.
`getTagSettings` | _Document_ Now static.
`isFullScreen` | _Remove_ Had been deprecated in 4.x.
`onDurationChange` | _Document_ `handleTechDurationChange`
`onEnded` | _Document_ `handleTechEnded`
`onError` | _Document_ `handleTechError`
`onFirstPlay` | _Document_ `handleTechFirstPlay`
`onFullscreenChange` | _Document_ `handleFullscreenChange` / `handleTechFullscreenChange`
`onLoadStart` | _Document_ `handleTechLoadStart`
`onLoadedAllData` | _Document_ `handleLoadedAllData`
`onLoadedMetaData` | _Document_ `handleTechLoadedMetaData`
`onLoadedData` | _Document_ `handleTechLoadedData`
[...]

PR #2406, opened 2015-07-24
Restorations and Removals

### LINKED ISSUE #2405: Restore/Remove videojs Properties/Methods
Per discussion in #2402, this ticket covers restoring certain properties and methods for 5.0.

### Restore
- [x] `TOUCH_ENABLED`
- [x] `parseUrl`
- [x] `round`
- [x] `EventEmitter` as `EventTarget`

### Remove
- [x] `Player#cancelFullScreen`
- [x] `Player#isFullScreen`
- [x] `Player#requestFullScreen`

### Remove, Shim, Deprecate
- [x] `Component#init`

misteroneill:
Closes #2405 by restoring several properties/methods for 5.0. Follows on from the discussion in #2402.


src/js/video.js
```diff
@@ -389,6 +402,32 @@ videojs.createTimeRange = createTimeRange;
...
+ * @method roundFloat
+ */
+videojs.round = roundFloat;
+
+/**
+ * Event target class.
+ *
+ * @type {Function}
+ */
+videojs.EventTarget = EventEmitter;
```
gkatsev: if we are changing `EventEmitter` to `EventTarget`, I'd prefer to just rename it internally as well and not just on export.

heff:
Very nice. Thanks for cleaning up the old deprecations. 

+1 on Gary's comment

I think we should also deprecate round and export roundFloat by that name.

gkatsev:
Is `roundFloat` really necessary? Of the usecases I've seen in the codebase, 95% of it is solved by the built in `Number#toFixed`.

heff:
Ha, no, probably not. I doubt anyone's using it anyway.

-- 
Sent from mobile

gkatsev:
I vote for not exporting it and considering excising it from the codebase.

heff:
Yeah works for me.


-- 
Sent from mobile

misteroneill:
Sounds good. Wasn't sure about renaming `EventEmitter`, but I'll take care of that. Also, I'll remove `roundFloat`/`round` entirely and switch any uses to `Number#toFixed`.

src/js/extends.js
```diff
@@ -49,6 +51,11 @@ const extendsFn = function(superClass, subClassMethods={}) {
   let methods = {};
 
   if (typeof subClassMethods === 'object') {
+    if (typeof subClassMethods.init === 'function') {
+      log.warn('Constructor logic via init() is deprecated; please use constructor() instead.');
+      subClassMethods.constructor = subClassMethods.init;
+      delete subClassMethods.init;
```
misteroneill: I wasn't 100% sure `delete`ing the `init` was the best idea here. Had a minor concern that users inspecting the prototype chain could get thrown off, but perhaps the deprecation warning is enough.

gkatsev:
LGTM

misteroneill:
BTW, I updated [the wiki](https://github.com/videojs/video.js/wiki/5.0-Change-Details), too. Apparently you don't have to be an org member to change wiki pages... who knew!

gkatsev:
Yep, by default wikis are open to everyone. for better or worse.

misteroneill:
At least the wiki is just a git repo, so it's easy enough to roll back. :smile:

src/js/extends.js
heff: I'm not sure on this one either, but we may want to consider leaving it in case older external code points to the init directly for some reason. Probably better they get the warning than a broken function.

src/js/utils/dom.js
```diff
@@ -329,7 +328,7 @@ export function findElPosition(el) {
 
   // Android sometimes returns slightly off decimal values, so need to round
   return {
-    left: roundFloat(left),
```
heff: I think these should these just be Math.round() now. Back when `roundFloat` was just `round` it was used for most rounding.

src/js/video.js
```diff
@@ -183,7 +185,7 @@ videojs.getComponent = Component.getComponent;
  */
 videojs.registerComponent = Component.registerComponent;
 
-/*
+/**
```
heff: The additional asterisk was left out intentionally by @mboles, in order to make the comment private. We should use the `@private` tag instead, at least until we figure out how our new jsdoc process will handle this type of object.

heff:
LGTM!

EventTarget stills feels a little awkward to me exported directly on `videojs`. If someone can think of a namespace it and others like it could live under I'd be interested in that. But no need to figure that out for this PR.

gkatsev:
We could register `EventTarget` as a component so it's available through `getComponent`, though, it isn't technically a component.

src/js/utils/dom.js
misteroneill: Ah, I see. Instead of `toFixed` with no arguments, use `Math.round`. Makes sense.

src/js/extends.js
misteroneill: Good enough reason for me. I'll remove this line.

misteroneill:
@heff Addressed the issues you raised (except `EventTarget` being on the `videojs` namespace).

heff:
Great! I think it's good to pull in.

-- 
Sent from mobile

PR #2411, opened 2015-07-28
Fix time display by using scrubbing as a function

nickygerritsen:
`this.player_.scrubbing` is a function and thus this call should have `()` added. Currently this will always return the cached current time

### DIFF
--- a/src/js/control-bar/time-controls/current-time-display.js
+++ b/src/js/control-bar/time-controls/current-time-display.js
@@ -43,13 +43,13 @@ class CurrentTimeDisplay extends Component {
   }
 
   /**
-   * Update current time display 
+   * Update current time display
    *
    * @method updateContent
    */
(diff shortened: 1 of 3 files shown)


heff:
Oh man, thanks for catching that.

I think we should be fixing that in remainingTime too, though it looks like we have more updating to do in that component. Would you mind fixing that comment either way? https://github.com/videojs/video.js/blob/57152ae62fff461bfb5aedef4593e13317b26f1b/src/js/control-bar/time-controls/remaining-time-display.js#L58

nickygerritsen:
Found another spot where scrubbing was used without calling the function in comments, so modified that one too

heff:
Awesome, thanks! Looks good to me.

@videojs/core-committers can one other person confirm and pull in?


-- 
Sent from mobile

PR #2412, opened 2015-07-28
Fix track onload

### LINKED ISSUE #1914: Subtitles default does not work
I'm very sorry, I have not worked default subtitle.

```
 <track kind="captions" src="demo.captions.vtt" srclang="en" label="English" default></track><!-- Tracks need an ending tag thanks to IE9 -->
    <track kind="subtitles" src="demo.captions.vtt" srclang="en" label="English" default></track><!-- Tracks need an ending tag thanks to IE9 -->
```

gkatsev:
I believe this fixes #1914. Also, this makes it so that users can actually select native text tracks on the first try.

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -61,6 +61,23 @@ vjs.Html5 = vjs.MediaTechController.extend({
 
     if (this['featuresNativeTextTracks']) {
       this.on('loadstart', vjs.bind(this, this.hideCaptions));
+
+      // native captions load on demand. Sometimes, some of the tracks get loaded in by default
+      // by the user agent. We don't want the user agent to load stuff by default
+      // (unless it's a `default` track) so we add an `onload` handler to `disable` the tracks.
+      // However, if a user chooses the track, we want to let it through.
+      // `playing` is our arbitrary cutoff for user-agent vs user events.
+      this.one('playing', function() {
+        this.textTracks().addEventListener('change', vjs.bind(this, function() {
+          var tracks = this.el().querySelectorAll('track'),
+              i = 0;
+          for (; i < tracks.length; i++) {
+            if (tracks[i]['track']['mode'] === 'showing') {
+              tracks[i]['onload'] = null;
+            }
+          }
+        }));
+      })
     }
 
     // Determine if native controls should be used
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -59,6 +59,27 @@ vjs.Html5 = vjs.MediaTechController.extend({
       }
     }
 
+    if (this['featuresNativeTextTracks']) {
+      this.on('loadstart', vjs.bind(this, this.hideCaptions));
+
+      // native captions load on demand. Sometimes, some of the tracks get loaded in by default
+      // by the user agent. We don't want the user agent to load stuff by default
+      // (unless it's a `default` track) so we add an `onload` handler to `disable` the tracks.
+      // However, if a user chooses the track, we want to let it through.
+      // `playing` is our arbitrary cutoff for user-agent vs user events.
+      this.one('playing', function() {
+        this.textTracks().addEventListener('change', vjs.bind(this, function() {
+          var tracks = this.el().querySelectorAll('track'),
+              i = 0;
+          for (; i < tracks.length; i++) {
+            if (tracks[i]['track']['mode'] === 'showing') {
+              tracks[i]['onload'] = null;
+            }
+          }
+        }));
+      });
+    }
+
     // Determine if native controls should be used
     // Our goal should be to get the custom controls on mobile solid everywhere
     // so we can remove this all together. Right now this will block custom
@@ -159,6 +180,25 @@ vjs.Html5.prototype.createEl = function(){
 
   return el;
   // jenniisawesome = true;
+};
+
+
+vjs.Html5.prototype.hideCaptions = function() {
+  var tracks = this.el_.querySelectorAll('track'),
+      track,
+      i = tracks.length,
+      kinds = {
+        'captions': 1,
+        'subtitles': 1
+      };
+
+  while (i--) {
+    track = tracks[i].track;
+    if ((track && track['kind'] in kinds) &&
+        (!tracks[i]['default'])) {
+      track.mode = 'disabled';
+    }
+  }
 };
 
 // Make video events trigger player events
@@ -412,9 +452,11 @@ vjs.Html5.prototype.addRemoteTextTrack = function(options) {
 
   track['onload'] = function() {
     var tt = track['track'];
-    if (track.readyState >= 2) {
+    if (track.readyState >= 2 && !track.hasAttribute('default')) {
       if (tt['kind'] === 'metadata' && tt['mode'] !== 'hidden') {
         tt['mode'] = 'hidden';
+      } else if (tt['kind'] !== 'metadata' && tt['mode'] !== 'disabled') {
+        tt['mode'] = 'disabled';
       }
       track['onload'] = null;
     }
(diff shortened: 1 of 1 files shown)


imbcmdth:
LGTM :hamburger:

imbcmdth:
@gkatsev Is this something that will have to be fixed in vjs5 as well?

misteroneill:
@imbcmdth Yes, there's a PR for that! #2413

src/js/media/html5.js
```diff
@@ -61,6 +61,23 @@ vjs.Html5 = vjs.MediaTechController.extend({
 
     if (this['featuresNativeTextTracks']) {
       this.on('loadstart', vjs.bind(this, this.hideCaptions));
+
+      // native captions load on demand. Sometimes, some of the tracks get loaded in by default
+      // by the user agent. We don't want the user agent to load stuff by default
```
dmlap: Why don't we want user agents to load unless the default attribute is present?

misteroneill: Certain browsers/players will display captions for the first track it encounters if there is no track element with the `default` attribute - without any interaction from the user. Merely having track elements will cause captions to show.

dmlap: Could we fix this by setting the track mode to `hidden` explicitly when creating text tracks?

gkatsev: Nope, setting the mode to anything before the track has finished loading does nothing.

heff:
Referenced in #1865, but there's [a chromium bug open for this issue](https://code.google.com/p/chromium/issues/detail?can=2&start=0&num=100&q=captions%20showing&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified&groupby=&sort=&id=388588).

Just reiterating that showing any captions that exist by default is not *really* how the browsers should work.

misteroneill:
IIRC, it was happening in iOS as well, which prompted this PR originally (the fix for the ticket mentioned above was a happy accident).

I can dig into this today and do a quick pass to see which browsers have the bad behavior.

misteroneill:
This PR originally added another layer of workarounds to fix broken behavior in Safari on iPhone that was reported internally. But that issue was caused by the workaround from #1865. Upon further investigation, it seems the initial workaround is no longer needed - Chrome does not appear to be playing a track by default when no `default` attribute is present.

Both @dmlap and I agree that allowing clients to keep their default behavior is preferable to adding workarounds for workarounds, so I've changed this to simply remove the workaround from #1865.

Safari on iPhone still has some interesting and perhaps unintuitive behavior. It appears that it will choose a caption for the user if the user previously chose that caption language. If "Off" was previously selected, it will be going forward. I couldn't find a way to clear this apparent preference.

I did notice one strange thing with Safari on iPhone. The `TextTracksList` instance that gets created appears to magically gain an extra `TextTrack` captions object with an empty `"language"`... _without a call to_ `addTrack_()`! This appears to be the cause of iOS's display of an "Unknown CC" option in the CC menu. I wasn't able to track down the cause of this today, but wanted to note it for future investigation.

misteroneill:
Additional discovery. Going to _Settings > Accessibility > Subtitles & Captioning_ and disabling _Closed Captions + SDH_ appears to turn off the language selection caching behavior. Videos will then default to "Auto (Recommended)".

dmlap:
Deserves a test pass but lgtm

src/js/media/html5.js
```diff
@@ -438,8 +415,6 @@ vjs.Html5.prototype.addRemoteTextTrack = function(options) {
     if (track.readyState >= 2) {
```
heff: If we're going for default behavior now would we not just kill this whole function, or do we still need to force metadata tracks to hidden?

heff:
Just confirmed the latest Chrome doesn't show tracks by default unless they have a default attribute (woo!), and Firefox works the same way.

Safari on desktop and mobile still shows tracks by default. They make it easy enough to turn off the tracks in the native UI, but it's still not idea. Would anyone up for submitting a webkit bug to mimic the chrome bug that was fixed?

If we revert this fix I think we should test how the captions menu performs in the different browsers. Whether or not it shows the right track as 'on' depending on the browser's behavior.

Do we have any idea how this works in IE?

One other comment inline but otherwise this looks good to me.

src/js/media/html5.js
misteroneill: I was hesitant to delete this bit because I didn't totally understand its intent. I'd like @gkatsev to review tomorrow when he's back.

misteroneill:
I checked IE9 and it seems to behave the same as Chrome/Firefox. I'll check others today. I didn't see the same behavior you're describing with Safari on desktop at all - it behaved like Chrome/Firefox/IE9.

iOS seems to be the only odd case for me, but its behavior doesn't seem broken - it should only choose a text track if that language has been chosen previously in the native player. Turning off the captions accessibility feature resets it (though choosing a caption again seems to cause the defaulting behavior).

heff:
Just checked Desktop Safari 8.0.7 on a coworker's machine who doesn't use safari, and captions showed by default, using a plain video tag and no default attribute on the captions. Could you have already selected 'off' instead of 'auto' some time previously?

On iOS 8.3 it seems to also always show captions by default for me. I tried toggling Closed Captions + SDH but it didn't change anything. I'm upgrading to 8.4 to check that. Otherwise we might need to get a fresh phone somehow to know for sure. Basically I think it's worth submitting a webkit bug if on a new phone, captions show by default on a plain video tag when no default attribute is used.

misteroneill:
Ok, I was finally able to reproduce in Safari and in Chrome. I don't have access to iOS at home and the simulator is crashing when I try to play a video.

I'm thinking this is a difference between local text tracks and remote text tracks. The issue that was reported internally was for a video with remote text tracks. Now that I'm looking at a video with local text tracks (the default index.html in the VideoJS sandbox for both 4 and 5), I'm seeing the behavior you're describing - and the Chrome bug mentioned earlier.

heff:
By remote do you mean cross-domain, and that local tracks work differently from cross-domain tracks??

misteroneill:
I meant the distinction in code between `addTextTrack` and `addRemoteTextTrack`. I'm not entirely clear on the difference - or if I'm right on this - but the original changes in this PR were in response to looking at a Brightcove player pulling a video from the catalog. Upon looking at the default sandbox player (which does not), the default behavior is totally different across browsers. I'm starting to think I may just need to pop off my commit on this branch and go back to the workaround behavior, but I'll talk to @gkatsev today about that.

misteroneill:
I've created [a wiki page](https://github.com/videojs/video.js/wiki/Text-Tracks-Browser-Behavior) to start documenting the various handling of text tracks / captions in different browsers because - as @gkatsev has pointed out to me - there are some idiosyncratic behaviors there.

misteroneill:
Rolled back the rollback. :roller_coaster:

gkatsev:
So... how bad would it be if we just removed all this code, made the browser just do whatever it does and opened a bug against webkit? @videojs/core-committers

dmlap:
@gkatsev +1 I'm in favor of letting the browser do what it does naturally.

gkatsev:
This does mean that until webkit will get it fixed, we're going to get captions showing up by default in some cases (particualrly, iOS ans Safari), even if the `default` attribute wasn't set.

misteroneill:
If you want to do that, just revert my revert: 5ec859ce37b904f589e076a31f87e104b2589fd2 :)

gkatsev:
@misteroneill well, we'd be able to delete some more stuff than just that commit. :)

gkatsev:
I'm going to close this in favor of a new PR that removes all the codes.

gkatsev:
Opened a bug against webkit: https://bugs.webkit.org/show_bug.cgi?id=147951

PR #2420, opened 2015-07-31
Restore formatTime

misteroneill:
Several core plugins use this method and there is at least one external plugin that uses it (found via GitHub issues).

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -398,8 +398,6 @@ videojs.log = log;
 videojs.createTimeRange = createTimeRange;
 
 /**
- * @file format-time.js
- *
  * Format seconds as a time string, H:MM:SS or M:SS
  * Supplying a guide (in seconds) will force a number of leading zeros
  * to cover the length of the guide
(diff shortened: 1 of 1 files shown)


dmlap:
lgtm

src/js/video.js
```diff
@@ -397,6 +398,20 @@ videojs.log = log;
 videojs.createTimeRange = createTimeRange;
 
 /**
+ * @file format-time.js
```
heff: `@file`?

heff: Ah, I see. copy/paste. The `@file` isn't needed here, and could probably be moved into it's own comment block in format-time.js to avoid future issues like this.

heff:
Need to remove the `@file` tag but otherwise looks good to me.

misteroneill:
`@file` removed. Whoops!

dmlap:
seconded!

mmcc:
:thumbsup:

PR #2421, opened 2015-08-03
Restore Html5.Events

misteroneill:
There are a number of plugins that use the `Events` property of the `Html5` tech. This restores that; though, it's only one solution to that issue. If another solution is preferred, let's discuss.


dmlap:
lgtm

heff:
lgtm

PR #2423, opened 2015-08-03
Update package.json

paladox:


mmcc:
These changes don't make a difference, all the version bumps you've done are already encompassed by either the carat or tilde.

PR #2441, opened 2015-08-06
Allow customizing the port used by grunt connect

misteroneill:
Does what it says on the box.

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -68,7 +68,8 @@ module.exports = function(grunt) {
       dev: {
         options: {
           port: Number(process.env.VJS_CONNECT_PORT) || 9999,
-          livereload: true
+          livereload: true,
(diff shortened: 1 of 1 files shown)


misteroneill:
For some reasoning here, it was getting tedious to switch between `stable` and `master`, so I have a second clone now. This allows both to run `grunt connect` simultaneously.

gkatsev:
I think connect also has an option to try and figure out the next open port. Could we enable that option as well?

gkatsev:
https://github.com/gruntjs/grunt-contrib-connect#useavailableport

misteroneill:
I suppose that works, too. Why not both?

gkatsev:
Yep, both is what I meant :)

heff:
lgtm!

PR #2447, opened 2015-08-07
Restore novtt version of VideoJS

misteroneill:
The novtt version of VideoJS was not being created (though there was a `copy` task for it). This restores it and moves the generation of banners out of the browserify process so that the license banner can be programmatically generated depending on the presence of vtt.js. Relevant changes are:

- Lodash is pulled into the Gruntfile, but it was already a dependency.
- License banner templates are processed by customized processor functions, which manually render.

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -1,14 +1,32 @@
 module.exports = function(grunt) {
   require('time-grunt')(grunt);
 
+  let _ = require('lodash-compat');
   let pkg = grunt.file.readJSON('package.json');
   let license = grunt.file.read('build/license-header.txt');
+  let bannerCommonData = _.pick(pkg, ['version', 'copyright']);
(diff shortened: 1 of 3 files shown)


dmlap:
lgtm

gkatsev:
LGTM

grunt.js
```diff
@@ -293,6 +311,13 @@ module.exports = function(grunt) {
       }
     },
     concat: {
+      novtt: {
+        options: {
+          separator: '\n'
+        },
+        src: ['build/temp/video.js'],
```
heff: We're not actually concatenating anything here, but works for me. :)

heff:
fwiw, I don't think the separate license header for vtt is really necessary, but nicely done either way.

PR #2448, opened 2015-08-07
Pass vtt.js option to tech

misteroneill:
This is needed by the `emulateTextTracks` method when vtt.js is not bundled with video.js, so it can load vtt.js on demand from a custom location.


gkatsev:
LGTM

dmlap:
lgtm

PR #2456, opened 2015-08-10
Expose USER_AGENT on browser object.

misteroneill:
This was previously exposed in un-GCC compiled versions of video.js and is used in a few places internally.

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -4,7 +4,7 @@
 import document from 'global/document';
 import window from 'global/window';
 
-const USER_AGENT = window.navigator.userAgent;
+export const USER_AGENT = window.navigator.userAgent;
 
 /*
  * Device is an iPhone


src/js/utils/browser.js
```diff
@@ -4,7 +4,7 @@
 import document from 'global/document';
 import window from 'global/window';
 
-const USER_AGENT = window.navigator.userAgent;
+export const USER_AGENT = window.navigator.userAgent;
```
dmlap: External users would have to be using browserify to get at this, right?

misteroneill: Nope, [the `browser` object](https://github.com/videojs/video.js/blob/master/src/js/video.js#L219) includes all the exported constants here.

dmlap: Ah, ok. Any background on where this was used? It's not obvious to me why an alias to `userAgent` is helpful to outside code.

dmlap:
@misteroneill and I feel like aliasing `userAgent` isn't worth restoring after further discussion.

PR #2460, opened 2015-08-11
Turkish typo fix

ogun:


mmcc:
Thanks! This *could* be made against stable as well, but I don't know if this necessitates a patch release.

PR #2462, opened 2015-08-11
add safari user agent check

saxena-gaurav:


heff:
Could you add some background as to why this is needed?

Issue #2443, opened 2015-08-06
Accessing/Manipulating multiple player sources


chemoish:
Has there been any thought about being able to access any/all sources provided to the player?

https://github.com/videojs/video.js/blob/master/docs/api/vjs.Player.md#src-source-

``` js
myPlayer.src([
  { type: "video/mp4", src: "http://www.example.com/path/to/video.mp4" },
  { type: "video/webm", src: "http://www.example.com/path/to/video.webm" },
  { type: "video/ogg", src: "http://www.example.com/path/to/video.ogv" }
]);
```

I am guessing for the typical case it doesn't matter which it plays, as long as it plays.

The use case might be...

``` js
myPlayer.src([
  { type: "application/dash+xml", src: "480.mpd" },
  { type: "application/dash+xml", src: "720.mpd" },
  { type: "application/dash+xml", src: "1080.mpd" }
]);
```

This may be more easily achieved if your utilizing `source` tags, but for src's with larger configuration/protection data. I don't believe you have the luxury.
---

The end goal might be to create a resolution changer.

The closest thing I found, but didn't really work was `vjs('player').player().options().sources // []`
---

Similarly, with playroll (`videojs-contrib-ads`), you cannot hot swap sources with protected data, because there is no way of obtaining sources that are dynamically attached.

…**i think**…

heff:
If you're just looking for a file-based resolution changer, I think there's still one in the plugins listing. I'm not sure exactly how it pulls this off, but that type of functionality makes the most sense as a plugin. However we could consider adding some way of accessing the original sources provided. The options() path would only work if you used source elements to provide the multiple sources, instead of the src() api.

chemoish:
There is 2 one that is new and one that isn't supported anymore.

https://github.com/vidcaster/video-js-resolutions/blob/master/video-js-resolutions.js#L424 (old)
https://github.com/dominic-p/videojs-resolution-selector/blob/master/video-quality-selector.js#L185 (new)

Both players rely on `options()`—which still has the issue of interacting with dynamic sources (same with pre/mid/postrolls—pretty much any src swap).

It would be great if sources can have the same/similar interaction API, whether or not they are inline or not. 

It *seems* like tracks are already handled for this case (albeit kind of weird):

```js
videojs('player').player().options().tracks

// vs.

videojs('player').textTracks
```

---

Thinking of solutions…

Additionally, I noticed that:

https://github.com/videojs/video.js/blob/master/src/js/player.js#L1675

```js
[...]

PR #2467, opened 2015-08-11
Add the ability to obtain the current source object when dynamical…

chemoish:
…ly setting the src.

This is needed for source retrival for two use cases:
- preroll
- resolution switch (kind of)

Currently there is no way of accessing the source object of any DRM specified content. Plugin access only has API methods to `currentSrc` and `currentType`.
This will allow for `currentSource` which will return a valid source handled source object.

Partially addresses https://github.com/videojs/video.js/issues/2443

### DIFF
--- a/test/unit/media.js
+++ b/test/unit/media.js
@@ -222,6 +222,7 @@ test('should add the source hanlder interface to a tech', function(){
   // Pass a source through the source handler process of a tech instance
   tech.setSource(sourceA);
   strictEqual(tech.currentSource_, sourceA, 'sourceA was handled and stored');
+  strictEqual(tech.currentSource(), sourceA, 'sourceA was retrievable after being handled and stored.');
   ok(tech.sourceHandler_.dispose, 'the handlerOne state instance was stored');
 
   // Check that the handler dipose method works
--- a/src/js/media/media.js
+++ b/src/js/media/media.js
@@ -332,6 +332,14 @@ vjs.MediaTechController.prototype.emulateTextTracks = function() {
 };
 
 /**
+ * Returns the current source object that was sent to a source handler.
+ * @returns {Object} current source object
+ */
+vjs.MediaTechController.prototype.currentSource = function(){
+  return this.currentSource_;
+};
+
+/**
  * Provide default methods for text tracks.
  *
  * Html5 tech overrides these.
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+@chemoish add the ability to obtain the current source when dynamically setting the src.
```
gkatsev: This actually gets added automatically during our merge process.

src/js/media/flash.js
```diff
@@ -153,6 +153,10 @@ vjs.Flash.prototype['currentSrc'] = function(){
   }
 };
 
+vjs.Flash.prototype['currentSource'] = function(){
```
gkatsev: why not add it to the parent of these techs. That way, you can have it only in one place.

gkatsev:
Thanks for the PR. Seems like this might be useful.

src/js/media/flash.js
chemoish: Makes sense. I was trying to decipher the best way to inject it and these were the only places I saw access `currentSource_`. Didn't think to check that they were extending the `media.js` file (Was too focused on the Tech part).

Ty.

CHANGELOG.md
chemoish: Sounds good.

chemoish:
I hope so. Should I create a separate PR for 5?

test/unit/media.js
```diff
@@ -222,6 +222,7 @@ test('should add the source hanlder interface to a tech', function(){
   // Pass a source through the source handler process of a tech instance
   tech.setSource(sourceA);
   strictEqual(tech.currentSource_, sourceA, 'sourceA was handled and stored');
+  strictEqual(tech.currentSource(), sourceA, 'sourceA was retrievable after being handled and stored.');
```
gkatsev: sweet. Was just going to ask to add a test and then noticed this :)

heff:
So currentSource at the tech level is really just used for [source handler techs](https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L584). Both Flash and HTML5 use source handlers, but for external techs like Youtube, this will stop working.

I made some comments on #2443. I like the idea but I think we need a little different approach than this PR to get at what you're really going for. Hopefully the comments make that clear.

chemoish:
I am not familiar with the other techs beyond the naive use case I am trying to tackle, but I will take it to the #2443 and try to provide a more robust solution.

Issue #2436, opened 2015-08-04
Track metadata issues with video.js#4.12.12


chemoish:
This works on `Video.dev.js#4.12.12`.

`Video.js#4.12.12` throws the following:

![screen shot 2015-08-04 at 3 03 02 pm](https://cloud.githubusercontent.com/assets/549319/9073933/f797b260-3ab9-11e5-9808-c321aadb4166.png)

Maybe:
```js
track.track['kind'] // from

track['track']['kind'] // to?
```

heff:
@gkatsev is this related to the one already being fixed or is it new? At least sounds related to #2412.

@chemoish if you want to try moving to 5.0, we're not mangling anymore so shouldn't be an issue. http://jsbin.com/sofiwu/edit?html,console,output

gkatsev:
Yeah, it's a silly mistake. We could probably fix it as part of #2412 (yay gcc...)

gkatsev:
This will be removed as part of #2468 and #2475.

gkatsev:
This should be fixed in stable now. https://github.com/gkatsev/video.js/commit/a02f0befbce7808aeb0d84f3a3cea817949a1cb2

chemoish:
@gkatsev I am receiving an undefined track after this update.

Can you verify?

![screen shot 2015-08-28 at 2 10 07 pm](https://cloud.githubusercontent.com/assets/549319/9557165/9347dad2-4d8e-11e5-9099-7ed4ca8d1a0f.png)

https://github.com/videojs/video.js/blob/stable/src/js/tracks/text-track-list.js#L78

chemoish:
Moving this to a new issue.

PR #2468, opened 2015-08-12
Remove a bunch of code that tried to disable tracks by default

gkatsev:
I think that's all of it (plus probably some tests that are forthcoming).
I'll open it against master soon.

### DIFF
--- a/src/js/media/html5.js
+++ b/src/js/media/html5.js
@@ -59,10 +59,6 @@ vjs.Html5 = vjs.MediaTechController.extend({
       }
     }
 
-    if (this['featuresNativeTextTracks']) {
-      this.on('loadstart', vjs.bind(this, this.hideCaptions));
-    }
-
     // Determine if native controls should be used
     // Our goal should be to get the custom controls on mobile solid everywhere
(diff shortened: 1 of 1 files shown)


gkatsev:
Nope, looks like there aren't any tests for this that need to be removed.

heff:
Have we confirmed if this issue is fixed in Chrome? I got the impression it might at least be fixed in Canary. And fwiw, the captions still showed for me in desktop safari, even with the fix we're removing here. So if Chrome is fixed, not a huge impact I don't think. 

lgtm

gkatsev:
I'll see exactly where it is fixed, it'll be good to document. However, I still think this is the right move, for better or worse.

gkatsev:
Also fixes #2436.

heff:
still lgtm

forbesjo:
:+1:

gkatsev:
Looks like stable chrome still shows tracks by default.

gkatsev:
It's merged. https://github.com/gkatsev/video.js/commit/a02f0befbce7808aeb0d84f3a3cea817949a1cb2

PR #2469, opened 2015-08-12
Fix Vertical Slider

misteroneill:
- Updates how a slider determines whether it's vertical to use the `vertical()` method instead of referring to the option value. Conceivably the `vertical()` method could be called to update the `vertical_` property which would get it out of sync with the option value.
- Removed the unused `handle`-related code.
- Some minor formatting changes.


mmcc:
:thumbsup: lgtm

heff:
lgtm

PR #2473, opened 2015-08-13
Add tracks to techOptions

misteroneill:
There was an issue on the iPhone were we weren't getting tracks passed properly.

This was discovered by @gkatsev and myself during another marathon "WTF is iOS even doing with captions/subtitles" session, but was lost in one of the closed PRs related to those issues.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -499,6 +499,7 @@ class Player extends Component {
       'playerId': this.id(),
       'techId': `${this.id()}_${techName}_api`,
       'textTracks': this.textTracks_,
+      'tracks': this.options_.tracks,
       'autoplay': this.options_.autoplay,
       'preload': this.options_.preload,
       'loop': this.options_.loop,


gkatsev:
LGTM

heff:
Why do we need tracks and textTracks_? This seems like it might be a half fix because it will keep passing the tracks from the original options even if tracks have changed at some point.

heff:
If it is a half fix that actually fixes something immediate, that's cool with me, just wanted to note the potential future issue

gkatsev:
`tracks` is where videojs stores the original track elements from the embed code, where-as `textTracks` is the actual TextTrackList/current state of tracks.
Not having `tracks` means that in a lot of cases, iOS wouldn't receive any tracks on load.

heff:
Got it. What happens if both exist?

gkatsev:
`tracks` is only really ever used during init, so, I don't think there should be any issues with having both.

heff:
Ok...just want to make sure we're keeping the tech switching scenario in mind. The html5 tech may not be the first tech to play a source.

heff:
Otherwise lgtm!

gkatsev:
Hm... yeah, I guess I should double check whether further tech changes will cause any issues. It's conceivable we could get double tracks.

misteroneill:
I tried switching between techs a few times and never got duplicated tracks.

I did notice that if I had a track playing and switched techs, that it would no longer appear (though it was still selected in the CC flyout.

gkatsev:
I am getting double tracks when switching techs. I'll have to update this PR before merging.

gkatsev:
after looking at this some more, it seems like not exactly the solution we want. Going to close this an open a new PR.

PR #2477, opened 2015-08-13
Changed all component child lists to arrays instead of objects

heff:
Udpated the docs to match. This is a necessary change because object key order is not consistent in Chrome and children have a specific order. Somehow this hasn't actually affect us yet. But also there's not an easy way to insert a child into a specific spot in the order if it's an object.

The biggest issue here is that **player** children has been changed to an array. This would break anywhere that previously used
```
// any control bar or control bar child options under the children key. Same for other player children.
playerOptions = { children: { controlBar: { ... } } };
// Add a child to the children object
playerOptions = { children: { newChild: {} } };
```
Child options will no longer merge when they're under the children key, because children is now an array by default that will simply be overwritten by the new children object when options are merged.

You instead do:
```
playerOptions = { controlBar: { ... } };
// I'm not sure if adding children through options is real use case, but it wouldn't work with the array
// you can still use the addChild API, which I think is the primary one
player.addChild('newChild');
// or the prototype options
Player.prototype.options_.children.push('newChild'); // though this API has changed anyway
```

I know this is important but also don't have a good grasp of the significance, in case anyone else has any additional input. If we think it's really significant we can look into a patch or reconsider.

### DIFF
--- a/docs/guides/options.md
+++ b/docs/guides/options.md
@@ -121,21 +121,6 @@ is a child of `controlBar`, you can just set that component to false:
 
 ```javascript
 var player = videojs('video-id', {
-  children: {
-    controlBar: {
-      children: {
-        muteToggle: false
-      }
-    }
-  }
-});
-```
-
-All the children can start getting a little verbose, so to simplify things, you can also set options for child components directly on the parent options.
-This is functionally the exact same as the above, for instance.
-
-```javascript
-var player = videojs('video-id', {
   controlBar: {
     muteToggle: false
   }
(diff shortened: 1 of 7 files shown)


pam:

Commit: e499f1a024bd31d1f3f77689b1190615e01a83d3
Build details: https://travis-ci.org/pam/video.js/builds/75504974

heff:
@dmlap @gkatsev, this is one you guys should probably review.

dmlap:
Is using `children` object just deprecated or is the intent to completely remove it? I understand the motivation for pushing towards an array but it's a pretty pervasive pattern.

heff:
I don't feel a strong need to remove it completely, but it will probably stop being used either way if the internal components aren't using it. There won't be any value left in using the object approach.

heff:
@dmlap see #2534 for emphasis on why this is needed.

heff:
This one got lost for a bit but needs to go in before 5.0. Can I get another set of eyes?

heff:
Since no one else has anything to say on this one I'm going to pull it in on Monday.

mmcc:
lgtm

PR #2479, opened 2015-08-14
@chemoish add the ability to obtain current source object + current s…

chemoish:
…ource objects—when using `<source>` or `src()`.

This is needed for source retrival for two use cases:
- preroll
- resolution switch (kind of)

Currently there is no way of accessing the source object for any DRM specified content. Plugin access only has API methods to `currentSrc` and `currentType`.

This will allow for `currentSource` and `currentSources` to return unverified source object(s).

---

## Additional Testing

I manually ran sandbox test cases for the following (so I can verify dash tech integration more easily):

```html
<video>
  <source src="http://video-js.zencoder.com/oceans-clip.mp4" type='video/mp4'>
  <source src="http://video-js.zencoder.com/oceans-clip.webm" type='video/webm'>
  <source src="http://video-js.zencoder.com/oceans-clip.ogv" type='video/ogg'>
</video>

<!-- src(): mp4, webm, ogv -->
<video></video>

<video>
  <source src="http://rdmedia.bbc.co.uk/dash/ondemand/testcard/1/client_manifest-events.mpd" type="application/dash+xml">
</video>

<!-- src(): mpd -->
<video></video>
```

#### Screenshots

Here are the outputs from the `src()` examples.

screen shot 2015-08-13 at 6 38 18 pm

screen shot 2015-08-13 at 6 39 12 pm


chemoish:
Fixes https://github.com/videojs/video.js/issues/2443

src/js/player.js
```diff
@@ -1223,6 +1228,22 @@ vjs.Player.prototype.load = function(){
...
+vjs.Player.prototype.currentSources = function(){
+  return this.cache_.sources || [this.currentSource()];
+};
+
+/**
+ * Returns the current source object.
+ * @return {Object} The current source object
+ */
+vjs.Player.prototype.currentSource = function(){
+  return this.cache_.source || {};
```
heff: The empty object could have `{ src: this.currentSrc() }`

heff:
Thanks for putting this together! So far it looks pretty good. I need to find some time to dig into it more (and others are welcome to as well).

It think the issue I could see happening (which you may or may not have already accounted for):

1. We should be clearing the cache whenever a new source is set
```
// first set a list of sources
player.src([source1, source2]); 
// then set just a URL
player.src('blah.mp4');

// This should return `[{ src: 'blah.mp4'  }]` not the first list of sources
player.currentSources();
```

2. src() is a bit of a recursive function. In some cases if you pass it a list of sources, it will call src() again with the url of one of those sources. That would make it difficult to clear and update the cache appropriately.

src/js/player.js
chemoish: Will fix.

chemoish:
Ya, I'll give it another look on Monday once I fix the other issue. The recursive `src()` is a little bit tricky and the situation you accounted for is a known one (but not accounted).

The test cases I ran show that both inline `<source>`'s and `src([...])` populate the `currentSources()` correctly, but calling `src()` a second time will not **necessarily** clear `player.cache_.sources`.

I am not exactly sure which direction is the correct one, but I am also leaning toward keeping both methods 'nsync, the user may want the inline `<source>`'s to contain the `currentSources()`—Though in that case they might want to resort to `player().options().sources`.

---

In the preroll use case…

```js
let original_sources = player.currentSources();

player.src('preroll.mp4'); // current sources would then be [{ src: 'preroll.mp4' }]

// on preroll end

player.src(original_sources);
```

heff:
The preroll use case is a pretty clear one, thanks for adding that.

heff:
@chemoish I haven't forgotten about this one, but just haven't had more time to spend on it because of getting v5.0 out.

chemoish:
:+1: I rather have 1.5 sooner.

chemoish:
@heff How shall we proceed?

This was pointing to stable, prior. Should I recreate this PR and have it point to master?
Or should I fix my branch to point to stable (Accidentally merged master and got the 4 extra commits that are currently sitting on master—even though this is still pointing to stable)

heff:
If you can make it again master now that would be best.

PR #2481, opened 2015-08-14
Fix texttrack handling in IE10

nickygerritsen:
Apparantly IE10 does have `textTracks` on the video element, but does not support `addEventListener` and `removeEventListener` on it.

This should fix that problem.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -95,7 +95,7 @@ class Html5 extends Tech {
     let emulatedTt = this.textTracks();
 
     // remove native event listeners
-    if (tt) {
+    if (tt && tt.removeEventListener) {
       tt.removeEventListener('change', this.handleTextTrackChange_);
       tt.removeEventListener('addtrack', this.handleTextTrackAdd_);
       tt.removeEventListener('removetrack', this.handleTextTrackRemove_);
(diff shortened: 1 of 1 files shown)


gkatsev:
haha, I had that in my code but removed it. Does IE10 allow for us to listen for events on the `textTracks`? Or does it also not support it? In which case, we should disable it like we did IE10.

nickygerritsen:
IE10 gives `undefined` for `tt.addEventListener` (and also for the remove). So I think we should remove it

gkatsev:
is `onremovetrack` or `onaddtrack` available on the element?

nickygerritsen:
Oh just left work so I can only check much later

heff:
Bah...I told @gkatsev to remove the check. My bad!

nickygerritsen:
I tried to figure out if `onaddtrack` and `onremovetrack` are there, but both in IE and in Chrome if I output `vid.textTracks.onaddtrack` in the console I get nothing, so not really sure how to find out if this exists.

gkatsev:
If IE outputs nothing a falsey value for `'onremovetrack' in vid.textTracks`, then we should already be disabling native functionality (https://github.com/videojs/video.js/blob/master/src/js/tech/html5.js#L896-L898). Though, I guess we still need this because we might still be trying to bind and unbind the handlers.

heff:
lgtm

gkatsev:
Thanks.

PR #2482, opened 2015-08-14
Default dimensions

gkatsev:
Remove default dimensions from the stylesheet. Add them back in when videojs loads.
Add move the dimensions stylsheet into the head of the document.


pam:

Commit: c705b599ee55818a892371bbf27e20c7b2481294
Build details: https://travis-ci.org/pam/video.js/builds/75635241

gkatsev:
heff [13:34] 
@gkatsev: the stylesheet changes are pretty cool, and seem powerful, but also feel like overkill for this use case. What are the benefits of switching from just overwriting cssText.

gkatsev [14:11]
@heff: I think that it we should be using the DOM APIs that are given to us instead of using cssText/innerHTML for just anything. The extra code allows us to simplify the usage of adding the css in js and making it use cssText instead of the API won't really save us anything.

heff [14:21] 
It would save us a pretty sizable chunk of code. I think the DOM API argument would matter to me if we were modifying single rules, but we’re still clearing out the entire sheet and rewriting it, just now we’re doing it in two steps instead of one, with a lot of extra code. Doesn’t seem the most optimal.

gkatsev [14:22] 
I think
```[
      ['.' + idClass,
        ['width', width + 'px'],
        ['height', height + 'px']
      ],
      ['.' + idClass + '.vjs-fluid',
        ['padding-top', (ratioMultiplier*100) + '%']
      ]
    ])```

gkatsev [14:23]
I think that is much cleaner than 
```

gkatsev [14:23]
```var css = `.${idClass} { width: ${width}px; height: ${height}px; }`;
    // Add the aspect ratio CSS for when using a fluid layout
    css += `.${idClass}.vjs-fluid { padding-top: ${ratioMultiplier  * 100}%; }`;
```

gkatsev [14:23]
and most of the extra code goes into allowing that

heff [14:24] 
Maybe we should get another opinion on that because I can read the second more easily.

gkatsev [14:37] 
also, in SVT, we use an `addCssRule` type function for adding our fullscreen/fluid-iframe styles

gkatsev [14:43] 
@here thoughts on my PR and discussion above? (https://github.com/videojs/video.js/pull/2482)

GitHub
Default dimensions by gkatsev · Pull Request #2482 · videojs/video.js · GitHub
Remove default dimensions from the stylesheet. Add them back in when videojs loads. Add move the dimensions stylsheet into the head of the document.

pat [14:45] 
in this case, i find the second more readable. though, i do have some sympathy for the desire to use the DOM APIs rather than more blunt tools

pat [14:46]
hard to say. no strong opinion on this from me :simple_smile:

mmcc [15:14] 
the first example looks an attempt to write lisp with square brackets

gkatsev [15:14] 
more like css

mmcc [15:15] 
eh?

mmcc [15:15]
I don't know if I'm missing something, but I think the second looks more like css

mmcc [15:15]
just with interpolation happening

mmcc [15:18]
@gkatsev: Like @pat, I agree with you on why it's a good idea, but the second option is more readable

gkatsev [15:18] 
T__T

mmcc [15:18] 
lol I'm not saying that's the end of the conversation by any means

gkatsev [15:18] 
I can get rid of that and still be using insertRule

gkatsev [15:19]
`stylesheet.addCssRule(selector, properties);` (edited)

gkatsev [15:19]
```
properties = `width: ${width}px; height: ${height}px;`
```

mmcc [15:21] 
that seems clean to me. Any reason you don't like that?

gkatsev [15:22] 
1) the array method was available on MDN. 2) I like the array method because it will generate the properties string for you, you don't need to think about it. Just list the names and values

mmcc [15:24] 
got it

gkatsev [15:24] 
also, I'm a fan of lisp, so... :stuck_out_tongue:

mmcc [15:24] 
I don't have really strong feelings either way, but even knowing exactly what's going in that array syntax I still struggle to follow

gkatsev [15:25] 
idk, the array syntax looks pretty much like css but with extra brackets in there

mmcc [15:25] 
squints really hard

pat [15:26] 
that's a fair argument - that you're writing js, not css, so why should it look like css
gkatsev 
[15:26] 
Added Untitled in #general  
```js
['.video-js',
  ['width', '300px'],
  ['height', '150px']
]
```
```css
.video-js {
  width: 300px;
  height: 150px;
}
```
135b JavaScript/JSON • Edit • New window • View raw • Add comment

mmcc [15:26] 
@pat: because ultimately you're writing CSS

mmcc [15:27]
the end result is what other devs are having to parse there, which is the CSS output that the code will create

gkatsev [15:27] 
don't those two look pretty similar?

mmcc [15:28] 
@gkatsev: similar in the same way that ruby looks similar to javascript

mmcc [15:28] 
by that I mean, I can look at those to things and functionally know that they do similar things

mmcc [15:28]
if I know to look that they do similar things

mmcc [15:29]
if that makes sense

pat [15:29] 
the js style is harder to break because its structure is enforced vs. an interpolated string which is prone to eval-ish problems (bad syntax, etc)

gkatsev:
So, proposal:
remove code that supported the nested array structure for generating the css and instead just have it accept strings.
So, instead of:
```js
    stylesheet.addCssRules(this.styleEl_.sheet, [
      ['.' + idClass,
        ['width', width + 'px'],
        ['height', height + 'px']
      ],
      ['.' + idClass + '.vjs-fluid',
        ['padding-top', (ratioMultiplier*100) + '%']
      ]
    ]);
```
You would do:
```js
stylesheet.addCssRule(this.styleEl.sheet, '.' + idClass, `width: ${width}px; height: ${height}px;`);
stylesheet.addCssRule(this.styleEl.sheet, `.${idClass}.vjs-fluid`, `padding-top: ${ratioMultiplier * 100}%`);
```
It would still use `insertRule` behind the scenes. Basically, only thing that will be left is the `addRule` method.

heff:
Can you test that this direction works in IE8? `styleEl.sheet` doesn't exist, but I think you can use `styleEl.styleSheet`. Would be good to confirm the rest of it works.

If this lightens the added code I'll concede the direction. Still feels like we're using a spoon instead of a shovel to dig the same hole. It'd be nice if we could point to some other difference besides one DOM API is more right than the other, because that still feels debatable.

gkatsev:
I'll take a look at the `.sheet` thing in IE8. Though, we already use it elsewhere and I don't think we've had issues.

I'll think about it some more but generally I much prefer using an actual API, if available, over just setting the inner text of an element.

pam:

Commit: 8d7d48612bda09bef32f86724aa3717ff06afc34
Build details: https://travis-ci.org/pam/video.js/builds/75959270

pam:

Commit: 26e55c3b49d6e803314f4c862e76af5aac76b0ae
Build details: https://travis-ci.org/pam/video.js/builds/75959348

test/unit/player.test.js
```diff
@@ -172,7 +172,6 @@ test('should set the width, height, and aspect ratio via a css class', function(
   };
 
   // Initial state
-  ok(player.styleEl_.parentNode === player.el(), 'player has a style element');
```
gkatsev: we're no longer adding this element to the player. The tests below should cover whether the styles are applied correctly or not.

src/js/player.js
```diff
@@ -448,17 +451,16 @@ class Player extends Component {
     // Ensure the right class is still on the player for the style element
     this.addClass(idClass);
 
-    // Create the width/height CSS
-    var css = `.${idClass} { width: ${width}px; height: ${height}px; }`;
-    // Add the aspect ratio CSS for when using a fluid layout
-    css += `.${idClass}.vjs-fluid { padding-top: ${ratioMultiplier  * 100}%; }`;
+    stylesheet.setTextContent(this.styleEl_, `
```
gkatsev: I'm using one template string here to add both the previous css selectors to the style element.

gkatsev:
This one is good to go, I think.

pam:

Commit: 01191ed55f2b24ed683716bdef52728053383e7f
Build details: https://travis-ci.org/pam/video.js/builds/75970038

gkatsev:
The test failure is a timeout.

src/js/player.js
```diff
@@ -274,8 +275,10 @@ class Player extends Component {
     // Add a style element in the player that we'll use to set the width/height
     // of the player in a way that's still overrideable by CSS, just like the
     // video element
-    this.styleEl_ = document.createElement('style');
-    el.appendChild(this.styleEl_);
+    this.styleEl_ = stylesheet.getStyleElement('vjs-styles-dimensions');
+    let defaultsStyleEl = document.querySelector('.vjs-styles-defaults');
+    let head = document.querySelector('head');
+    head.insertBefore(this.styleEl_, defaultsStyleEl ? defaultsStyleEl.nextSibling : head.firstChild);
```
heff: Should we worry about scenarios where `defaultsStyleEl` is the only thing in the head. Or if there is no head for some reason?

src/js/player.js
heff: Beautiful.

src/js/player.js
gkatsev: I don't think it's worth worrying about if there's no head. I think looking for the `defaultsStyleEl` and prepending to HEAD otherwise, is quick and easy enough to do.

src/js/utils/stylesheet.js
```diff
@@ -0,0 +1,16 @@
+import document from 'global/document';
+
+export let getStyleElement = function(className) {
```
heff: This should probably be named `createStyleElement`, otherwise it looks like you're getting an existing style el with the supplied class name.

gkatsev: Make sense. I guess it could possibly look to see if a style element with the same class name exists and return that and create one, otherwise, or something, but I don't think that's that useful currently.

src/js/player.js
heff: Sounds good.

src/js/utils/stylesheet.js
heff: Yeah, since each player needs its own style element I don't think we need a getter. Which reminds me, should we have something that cleans up the style el when the player is disposed? Previously I think it was just relying on the style el being removed with the player div.

gkatsev: probably. I can add that in along with this rename.

gkatsev: Updated.

heff:
lgtm!

pam:

Commit: c020041ec1b5347db563ef1693bd4da2b9ca347f
Build details: https://travis-ci.org/pam/video.js/builds/75992694

gkatsev:
@pam retry

pam:

Commit: 58088baad40b5a6add8322ec6a1ab9e025c1141c
Build details: https://travis-ci.org/pam/video.js/builds/75993403

PR #2492, opened 2015-08-19
Test using BrowserStack

forbesjo:
This PR is to run the unit tests using BrowserStack because BrowserStack seems faster/more stable than Sauce Labs.

`BROWSER_STACK_USERNAME` and `BROWSER_STACK_ACCESS_KEY` need to be exported in Travis.
There also needs to be a mention of BrowserStack in the README or website to use the open source subscription.


heff:
Seems straight forward enough. Have you tried IE8 with this at all?

> There also needs to be a mention of BrowserStack in the README

Want to add something to the readme as part of this?

forbesjo:
IE's are good to go and added the mention

forbesjo:
Here's a demo build https://travis-ci.org/forbesjo/vjs-browserstack-demo/builds/76509552

mmcc:
The timeouts in that demo build make me sad.

heff:
It's weird because all the browsers seem to have run the tests except for IE9. I wonder if we fixed those vjs_getProp errors if the suite would pass, without IE9 that is.

forbesjo:
IE9 crashes when I try to view the qunit test page, when I skip that browser I don't see any disconnects. See https://travis-ci.org/forbesjo/vjs-browserstack-demo/builds/76531261

heff:
Strange that removing one browser stopped the disconnects in the rest. Also
we really need to clean up all those VIDEOJS error logs.

I'm working on adding the browser stack credentials to travis.

PR #2494, opened 2015-08-19
Fixup tracks some more

gkatsev:


pam:

Commit: 91d71c00d297978280de27fd6b199d4d777038ec
Build details: https://travis-ci.org/pam/video.js/builds/76361049

src/js/tech/html5.js
```diff
@@ -127,7 +127,7 @@ class Html5 extends Tech {
 
       // If the original tag is still there, clone and remove it.
       if (el) {
-        const clone = el.cloneNode(false);
+        const clone = el.cloneNode(true);
```
mmcc: Why is a deep clone necessary here?

gkatsev: Otherwise, we won't get tracks available on something like ios.
Previously, we'd grab all the tracks from the element and store them on the `options` object (which still happens) but then we need to pass it to the tech. However, if we pass through both that option and the `textTracks` object, if you swap between techs, you'll start getting duplicated tracks. So, instead, I can remove the need to pass through the `tracks` option to the tech and just clone the node with its sub nodes, since the `tracks` option should only have been used the first time the player was initialized.

gkatsev: Hopefully, that makes sense.

mmcc: Yep, that makes sense.

mmcc:
lgtm

src/js/control-bar/text-track-controls/text-track-button.js
```diff
@@ -20,6 +20,7 @@ class TextTrackButton extends MenuButton {
   constructor(player, options){
     super(player, options);
 
+    this.update();
```
gkatsev: this forces the buttons to update in case the video element has tracks from before we are able to register the handlers to update the buttons.

heff:
Got some travis errors

pam:

Commit: e585c5c9687dde55c03cbf8bfe1af05a4031e43d
Build details: https://travis-ci.org/pam/video.js/builds/76376526

heff:
lgtm

PR #2496, opened 2015-08-21
Fixed tracks test by switching to a fake tech

heff:
See #2492 and errors in https://travis-ci.org/forbesjo/vjs-browserstack-demo/builds/76509552

@gkatsev would this approach still be testing what we need it to, or are we testing nothing now?

### DIFF
--- a/test/unit/tracks/tracks.test.js
+++ b/test/unit/tracks/tracks.test.js
@@ -11,6 +11,7 @@ import Component from '../../../src/js/component.js';
 import * as browser from '../../../src/js/utils/browser.js';
 import TestHelpers from '../test-helpers.js';
 import document from 'global/document';
+import TechFaker from '../tech/tech-faker.js';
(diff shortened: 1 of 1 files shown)


gkatsev:
I guess the question is whether it would unload the old tech faker and re-load it like switching between a real tech would be? Maybe we can switch from the `Html5` tech to the `TechFaker` tech?

heff:
Yeah, there is a check to see if the same tech type is already loaded, but that happens in sourceList_, before loadTech. `loadTech` should always unload/load, so we should be ok there, as long as there's nothing special it's expecting the tech to do.

I'd like to avoid using even the HTML5 tech in tests that run on all platforms.

gkatsev:
Ok. That's fine by me then. I just wasn't certain whether calling `loadTech` with the same tech would actually be testing what we want.

PR #2497, opened 2015-08-21
Increase BrowserStack polling

forbesjo:
This is to increase the polling timeout so that the API rate limit of 120 requests/min is not hit as easily (default is 1000ms).


heff:
Looks good to me. Are there any potential downsides of this we should look out for?

I'll merge in a minute.

forbesjo:
I haven't seen it behave noticeably different
https://travis-ci.org/forbesjo/vjs-browserstack-demo/builds/76560552

PR #2498, opened 2015-08-21
Remove coveralls

forbesjo:
It looks like there's code coverage for each of the karma browsers and the grunt coveralls task. This PR is to remove the karma code coverage from the build. This also removes some extra code.


forbesjo:
Closing this, I misunderstood how the coverage process worked.

Issue #2277, opened 2015-06-19
5.0-rc.2 - no video playback in Android 4.1 stock browser


meleyal:
Happens on Android 4.1.2 using stock browser ("Internet" app).

1. Open test page
2. Click play button

Expected: Video starts playing
Actual: No video playback / black screen
Test case: http://jsbin.com/halokodoxo/1/edit?output

![screenshot_2015-06-19-10-45-14](https://cloud.githubusercontent.com/assets/15045/8250762/0f439f58-1677-11e5-9765-6b5d0aa1a4e7.png)

Maybe related: #2044, #1986, #975

moay:
Confirm. Also happens in rc58, but not only in stock browser - for me it's also chrome on android. The video will just not load, the poster image stays there.

moay:
I checked the logs I get on chrome, here is what I get:

    VIDEOJS: ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported. MediaError {code: 4, message: "The media could not be loaded, either because the …rk failed or because the format is not supported."}_logType 
    
    @ log.js:79116.log.error 
    @ log.js:31error 
    @ player.js:2026handleTechError 
    @ player.js:1019ret 
    @ fn.js:35ret     
    @ fn.js:35112.data.dispatcher 
    @ events.js:72

Works on iOS devices.

heff:
@forbesjo have you done any testing on Android with 5.0 yet?

moay:
This test was done on VJS 5. :-) Was just saying, that the displayed error cannot be true because the media and the network both are the same for a working setup with VJS4. Sorry, I hope it's easier to understand now?

gkatsev:
What device is it? Samsung?
What I could imagine is that you'd need to include the es5-shim for it to work but don't have specific proof.

forbesjo:
I can confirm that the issue still appears with the latest rc using 4.1.2 stock browser. 5.0 Android looks ok.

moay:
Yes, its a Samsung. But its running Android 5.0 (Cyanogenmod) and Chrome.

moay:
es5-shim doesn't change anything. ;-)

gkatsev:
Seeing this in logcat:
```sh
[...]

PR #2499, opened 2015-08-21
Native controls updates

gkatsev:
Previously, native controls were enabled by firing an event from the tech to the player. However, the player only bound events on the tech after the tech may have fired the event. Instead, the player will ask the tech whether controls are enabled or not. If the controls are enabled, it means we should be using native controls.
Also in this PR: when on iphones or native android browser, we should be using native controls.

### DIFF
--- a/src/css/components/_control-bar.scss
+++ b/src/css/components/_control-bar.scss
@@ -29,9 +29,9 @@
   @include transition($trans);
 }
 
-.video-js.vjs-controls-disabled .vjs-control-bar,
-.video-js.vjs-using-native-controls .vjs-control-bar,
-.video-js.vjs-error .vjs-control-bar {
+.video-js.video-js.video-js.vjs-controls-disabled .vjs-control-bar,
+.video-js.video-js.video-js.vjs-using-native-controls .vjs-control-bar,
+.video-js.video-js.video-js.vjs-error .vjs-control-bar {
   display: none;
 }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1953,6 +1953,7 @@ class Player extends Component {
             */
           this.trigger('usingnativecontrols');
         } else {
+          this.addTechControlsListeners();
           this.removeClass('vjs-using-native-controls');
           this.techCall('setControls', false);
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -501,6 +501,7 @@ class Player extends Component {
 
     // Grab tech-specific options from player options and add source and parent element to use.
     var techOptions = assign({
+      'nativeControlsForTouch': this.options_.nativeControlsForTouch,
       'source': source,
       'playerId': this.id(),
       'techId': `${this.id()}_${techName}_api`,
@@ -534,7 +535,6 @@ class Player extends Component {
     textTrackConverter.jsonToTextTracks(this.textTracksJson_ || [], this.tech);
 
     this.on(this.tech, 'ready', this.handleTechReady);
-    this.on(this.tech, 'usenativecontrols', this.handleTechUseNativeControls);
 
     // Listen to every HTML5 events and trigger them back on the player for the plugins
     this.on(this.tech, 'loadstart', this.handleTechLoadStart);
@@ -563,6 +563,8 @@ class Player extends Component {
     this.on(this.tech, 'volumechange', this.handleTechVolumeChange);
     this.on(this.tech, 'texttrackchange', this.onTextTrackChange);
     this.on(this.tech, 'loadedmetadata', this.updateStyleEl_);
+
+    this.usingNativeControls(this.techGet('controls'));
 
     if (this.controls() && !this.usingNativeControls()) {
       this.addTechControlsListeners();
@@ -663,16 +665,6 @@ class Player extends Component {
       delete this.tag.poster; // Chrome Fix. Fixed in Chrome v16.
       this.play();
     }
-  }
-
-  /**
-   * Fired when the native controls are used
-   *
-   * @private
-   * @method handleTechUseNativeControls
-   */
-  handleTechUseNativeControls() {
-    this.usingNativeControls(true);
   }
 
   /**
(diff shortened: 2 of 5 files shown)


src/js/player.js
```diff
@@ -501,6 +501,7 @@ class Player extends Component {
 
     // Grab tech-specific options from player options and add source and parent element to use.
     var techOptions = assign({
+      'nativeControlsForTouch': this.options_.nativeControlsForTouch,
```
gkatsev: pass the option to the tech

src/js/player.js
```diff
@@ -534,7 +535,6 @@ class Player extends Component {
     textTrackConverter.jsonToTextTracks(this.textTracksJson_ || [], this.tech);
 
     this.on(this.tech, 'ready', this.handleTechReady);
-    this.on(this.tech, 'usenativecontrols', this.handleTechUseNativeControls);
```
gkatsev: no longer needed because we assume if `techGet('controls')` is true, we are using native controls.

src/js/player.js
```diff
@@ -666,16 +668,6 @@ class Player extends Component {
   }
 
   /**
-   * Fired when the native controls are used
-   *
-   * @private
-   * @method handleTechUseNativeControls
-   */
-  handleTechUseNativeControls() {
```
gkatsev: no longer needed

src/js/player.js
```diff
@@ -1947,7 +1939,9 @@ class Player extends Component {
       if (this.usingNativeControls_ !== bool) {
         this.usingNativeControls_ = bool;
         if (bool) {
+          this.removeTechControlsListeners();
```
gkatsev: make sure to remove these listeners if they got added, otherwise, it will mess things up.
Our custom controls will show up unnecessarily, and also do weird stuff like prevent the video from playing correctly (#2277)

src/js/player.js
```diff
@@ -1947,7 +1939,9 @@ class Player extends Component {
...
           this.addClass('vjs-using-native-controls');
+          this.techCall('setControls', true);
```
gkatsev: Make sure we have controls

src/js/player.js
```diff
@@ -1960,6 +1954,7 @@ class Player extends Component {
           this.trigger('usingnativecontrols');
         } else {
           this.removeClass('vjs-using-native-controls');
+          this.techCall('setControls', false);
```
gkatsev: Remove controls if they are no longer needed

src/js/tech/html5.js
```diff
@@ -77,7 +77,7 @@ class Html5 extends Tech {
     // so we can remove this all together. Right now this will block custom
     // controls on touch enabled laptops like the Chrome Pixel
     if (browser.TOUCH_ENABLED && options.nativeControlsForTouch === true) {
-      this.trigger('usenativecontrols');
+      this.setControls(true);
```
gkatsev: make sure controls are set.

src/js/player.js
heff: This wouldn't hurt as a code comment if you feel like adding it.

dmlap: Do we need to re-add the listeners if we swap back to non-native controls?

pam:

Commit: 1af60898c42b85ed199fa660f64bb4ef18ad63f8
Build details: https://travis-ci.org/pam/video.js/builds/76673812

heff:
lgtm! 

`usingNativeControls` wasn't meant to actually let you change the state, in case we want to simplify that and say you can only set native controls through player/tech options. It seems fine as it is though otherwise.

heff:
Could you add a description for this issue for people who come back to it later?

src/js/player.js
gkatsev: I guess we should also re-add them below.

gkatsev:
Yep, I'll definitely update the description. Trying to figure out the best way to force the native controls on by default only on the native android browsers (because they don't work right with our controls).
I'll double check to see whether we need to do the `techCall('setControls', true);` in the `usingNativeControls` function.

pam:

Commit: 28509f4da6e4a12b21a6abc2497bd7d85216ca88
Build details: https://travis-ci.org/pam/video.js/builds/76678315

src/js/player.js
gkatsev: Added.

src/css/components/_control-bar.scss
```diff
@@ -29,9 +29,9 @@
   @include transition($trans);
 }
 
-.video-js.vjs-controls-disabled .vjs-control-bar,
-.video-js.vjs-using-native-controls .vjs-control-bar,
-.video-js.vjs-error .vjs-control-bar {
+.video-js.video-js.video-js.vjs-controls-disabled .vjs-control-bar,
+.video-js.video-js.video-js.vjs-using-native-controls .vjs-control-bar,
+.video-js.video-js.video-js.vjs-error .vjs-control-bar {
```
gkatsev: this is necessarily because otherwise, the `.vjs-no-flex` selector below ends up being more specific than this and sets the control bar's display to table. This avoids using `!important` or really strong css selectors.

gkatsev:
@heff updated.

pam:

Commit: 693ae2eb199ecd84805b4f5a261911cc6f445e27
Build details: https://travis-ci.org/pam/video.js/builds/76687501

pam:

Commit: d9c026375e99d461dc137bba544fc0279998fdde
Build details: https://travis-ci.org/pam/video.js/builds/76688952

src/css/components/_control-bar.scss
heff: I get the limitation here, but I don't think we should say this is an ok thing to do in the stylesheet. It's gonna result in uglier and uglier selectors, and I think we can do better.

- I'd prefer `important` here over a triple class name. It is after all important that the controls be hidden when the controls are disabled.

- We could try lowering some of the selector strengths and using strategic ordering over increasing class names.

- We could hide the controls in additional ways to `display`, so you have to do more than just display:table to override it.

mmcc: Another option to fix this specific issue would be to add a `.vjs-flex` class. Since that's the default behavior it could basically be a noop with no specific styles associated with it (because they're the default), but it would solve specificity problems in situations like this.

PR #2513, opened 2015-08-25
Fix console log errors in tests

heff:
Even though tests pass we still have a bunch of errors in the test logs from xhr events and browsers trying to load fake images/tracks. This fixes most of them.

See the individual commits.
- Fixed a console error in dom tests from loading a track source
- Switched to non-fetching poster urls in tests to prevent errors
- Stubbed XHR for TextTrack tests to prevent log errors

Still seeing two errors in IE10 that I need to find and clean up.


gkatsev:
Cool. LGTM

heff:
@gkatsev, FYI, there's some async happening somewhere in tracks that prevents stubbing XHR from working, at least in this test: https://github.com/videojs/video.js/blob/master/test/unit/tracks/tracks.test.js#L119
XHR isn't called until the test is already over. Disposing the player fixed the console errors though.

PR #2515, opened 2015-08-25
handle error getter on flash tech better

gkatsev:
The swf doesn't have [an error getter](https://github.com/videojs/video-js-swf/blob/master/src/VideoJS.as#L213-L299) and seems like all the info we have about the error is available in the `onError` method. So, instead of trying to call `vjs_getProperty` to the swf (which will never work or return a valid value), we'll store the error that we got there and return it in the error getter.

One thing I'm not certain off is when to clear out our reference to the error. Seems like when the error getter gets called is a good place to do so because it gets called from the player and player stores its own reference to the error which it clears out as necessary.

### DIFF
--- a/src/js/tech/flash.js
+++ b/src/js/tech/flash.js
@@ -456,12 +456,9 @@ Flash.onError = function(swfID, err){
 
   if (err === 'srcnotfound') {
     errorObj.code = 4;
-    tech.trigger('error', errorObj);
-
-  // errors we haven't categorized into the media errors
-  } else {
-    tech.trigger('error', errorObj);
   }
+
+  tech.trigger('error', errorObj);
 };
 
 // Flash Version Check
--- a/src/js/tech/flash.js
+++ b/src/js/tech/flash.js
@@ -268,6 +268,16 @@ class Flash extends Tech {
   }
 
   /**
+   * Get the last error object
+   *
+   * @return {Object}
+   * @method error
+   */
+  error() {
+    return this.error_ || {};
+  }
+
+  /**
    * Get fullscreen support -
    * Flash does not allow fullscreen through javascript
    * so always returns false
@@ -297,7 +307,7 @@ class Flash extends Tech {
 // Create setters and getters for attributes
 const _api = Flash.prototype;
 const _readWrite = 'rtmpConnection,rtmpStream,preload,defaultPlaybackRate,playbackRate,autoplay,loop,mediaGroup,controller,controls,volume,muted,defaultMuted'.split(',');
-const _readOnly = 'error,networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoTracks,audioTracks,videoWidth,videoHeight'.split(',');
+const _readOnly = 'networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoTracks,audioTracks,videoWidth,videoHeight'.split(',');
 
 function _createSetter(attr){
   var attrUpper = attr.charAt(0).toUpperCase() + attr.slice(1);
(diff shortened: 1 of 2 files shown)


src/js/tech/flash.js
```diff
@@ -297,7 +307,7 @@ class Flash extends Tech {
 // Create setters and getters for attributes
 const _api = Flash.prototype;
 const _readWrite = 'rtmpConnection,rtmpStream,preload,defaultPlaybackRate,playbackRate,autoplay,loop,mediaGroup,controller,controls,volume,muted,defaultMuted'.split(',');
-const _readOnly = 'error,networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoTracks,audioTracks,videoWidth,videoHeight'.split(',');
+const _readOnly = 'networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoTracks,audioTracks,videoWidth,videoHeight'.split(',');
```
gkatsev: remove `error`.

src/js/tech/flash.js
```diff
@@ -438,13 +448,19 @@ Flash.onEvent = function(swfID, eventName){
 Flash.onError = function(swfID, err){
   const tech = Dom.getEl(swfID).tech;
   const msg = 'FLASH: '+err;
+  const errorObj = {
+    message: msg
+  };
+
+  tech.error_ = errorObj;
```
gkatsev: Store the error object for usage above.

src/js/tech/flash.js
```diff
@@ -268,6 +268,16 @@ class Flash extends Tech {
   }
 
   /**
+   * Get the last error object
+   *
+   * @return {Object}
+   * @method error
+   */
+  error() {
+    return this.error_ || {};
```
gkatsev: Return the error or a black object.
This sounds like where I want to reset `this.error_`.

pam:

Commit: ff85e1249de721707f626a352ad478e59efc8006
Build details: https://travis-ci.org/pam/video.js/builds/77196947

src/js/tech/flash.js
heff: > or a black object

Sounds ominous. :)

If possible we should return `null` for no error to match the video element.

gkatsev: Heh, oops. Blank object.
If we do, (I don't oppose it), we'd want to fix the call site (https://github.com/videojs/video.js/blob/master/src/js/player.js#L1000-L1002) which assumes an object always.

src/js/tech/flash.js
heff: We could also make tech.error a setter that handles this part, but either way is fine.

src/js/tech/flash.js
```diff
@@ -438,13 +448,19 @@ Flash.onEvent = function(swfID, eventName){
...
 
   if (err === 'srcnotfound') {
-    tech.trigger('error', { code: 4, message: msg });
+    errorObj.code = 4;
+    tech.trigger('error', errorObj);
 
   // errors we haven't categorized into the media errors
   } else {
-    tech.trigger('error', msg);
+    tech.trigger('error', errorObj);
```
heff: `tech.trigger('error', errorObj);` is now the same in both places, so it could be moved outside the conditional.

src/js/tech/flash.js
heff: Technically the player event handler would only be called when an error did exist, but if we want to be save we can add a check for null. The `tech.error()` getter might be called outside of an error event.

heff:
A few notes but otherwise lgtm.

If we can figure out why the player and tech were continuing to operate after the tests, with little more effort, it might save us some pains down the road. My best guess is something isn't being disposed properly in the tests.

pam:

Commit: d663dbf46ac086c6766ce890a83bb77d7b085bdd
Build details: https://travis-ci.org/pam/video.js/builds/77204288

heff:
Does #2517 replace this one?

gkatsev:
Assuming we like #2517, yes.

gkatsev:
#2517 got merged. Closing.

PR #2519, opened 2015-08-26
Reset to 0 on play if the video has ended

bc-bbay:
The player now controls this behavior, not the SWF.


dmlap:
LGTM

heff:
Looks good to me.

When did this change in the swf? Can we note it specifically?

Has anyone run the test on IE8, since savage hasn't run here?

dmlap:
@heff here's the SWF change: https://github.com/videojs/video-js-swf/pull/175. It's in 4.7.3-- currently we have 4.7.2 on the CDN.

PR #2521, opened 2015-08-27
Add vjs-autoplay class when autoplaying

marguinbc:
Class allows player to hide poster, big play button when autoplay enabled

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -314,6 +314,11 @@ vjs.Player.prototype.loadTech = function(techName, source){
 
   var techReady = function(){
     this.player_.triggerReady();
+
+    // add vjs-autoplay to hide poster and big-play-button if autoplay:true
+    if (this.options_['autoplay']) {
+      this.addClass('vjs-autoplay');
+    }
   };
 
   // Grab tech-specific options from player options and add source and parent element to use.
--- a/src/js/poster.js
+++ b/src/js/poster.js
@@ -91,6 +91,11 @@ vjs.PosterImage.prototype.setSrc = function(url){
  */
 vjs.PosterImage.prototype.onClick = function(){
   // We don't want a click to trigger playback when controls are disabled
-  // but CSS should be hiding the poster to prevent that from happening
-  this.player_.play();
+  if (this.controls()) {
+    if (this.player_.paused())  {
+      this.player_.play();
+    } else {
+      this.player_.pause();
+    }
+  }
 };


src/js/player.js
```diff
@@ -314,6 +314,11 @@ vjs.Player.prototype.loadTech = function(techName, source){
 
   var techReady = function(){
     this.player_.triggerReady();
+
+    // add vjs-autoplay to hide poster and big-play-button if autoplay:true
+    if (this.options_['autoplay']) {
+      this.addClass('vjs-autoplay');
```
dmlap: Two questions:
- Why not just use an attribute selector to match when the `autoplay` attribute is actually present on the player div?
- If we do go this route, don't we need some equivalent code that removes the autoplay class when the attribute is removed?

marguinbc: 1) Because I just reused this comparison from another part of the code....
2) I was thinking we might need to remove it but then thought autoplay would either be set or not set.  I can't think of a use case where autoplay would be initially on but then removed.  If we do need to add/remove then I could move the code to the autoplay getter/setter instead.

dmlap: If it's possible to make this change purely through CSS, that would be preferable.

src/js/poster.js
```diff
@@ -91,6 +91,11 @@ vjs.PosterImage.prototype.setSrc = function(url){
  */
 vjs.PosterImage.prototype.onClick = function(){
   // We don't want a click to trigger playback when controls are disabled
-  // but CSS should be hiding the poster to prevent that from happening
```
dmlap: Is CSS not hiding the poster?

src/js/player.js
marguinbc: Attribute selector will work but when the poster is set as an attribute, there is still a brief flash of the poster image.

marguinbc: Poster flashes even with HTML5 video element so I think the best way to handle the poster is not to have one if you are using autoplay.

dmlap: That sounds reasonable to me. If your video takes a long time to start playing, you might even want the poster to show up.

heff:
The `vjs-has-started` class should already be hiding both of these components when the video plays automatically. Can you give more detail around why this is needed?

marguinbc:
@heff  The poster is still flashing for a split second before the content starts as you can see in this JSBin sample:  http://jsbin.com/jizobuxumi/edit?html,css

The poster appears somewhere between firstplay, play and waiting...  if I set a breakpoint on waiting then I see the poster image. 

The classes attached to the player at this point are: video-js vjs-default-skin vjs-controls-enabled vjs-playing vjs-has-started vjs-waiting

This is the issue I am trying to resolve.

heff:
I'd like to be at least a little bit conservative with the state classes we add to the player because they can add to some complexity in the css. So forgive my pushback but...

1. Is this really a problem that people complain about and is worth adding complexity for?
2. If you don't want the poster to show can you just remove the poster source?

When I remove video.js from your example and just look at the raw video element, the poster also flashes. So it's not just a video.js thing.
http://jsbin.com/davaqivotu/edit?html,output

marguinbc:
@heff No problem on the push back.  It is welcomed and appreciated.  To address your questions:

1) Yes, publishers are complaining about it, at least in the context of seeing the poster flash before pre-roll ads as opposed to regular content.
2) Removing the poster source is looking like the best/only option. 

Correct. It also happens in the vanilla HTML5 video element which appears to reinforce #2.

Not sure if there are other options at this point.

heff:
The playlist use case clarifies the issue for me, thanks. An alternative approach could be to have the playlist add a class of something like 'vjs-playlist-playing', or something that represents the playlist state, and then have that css class hide the poster and the play button.

gkatsev:
Was this a fix for the 4.x version of the player?
Has this been fixed in 5.x yet? Do we still need it in 4.x?

marguinbc:
@gkatsev This was for the 4.x player. I have not tested 5.x to see if the poster flash still occurs.  The issue is that the poster is shown for a split second while vjs-ad-loading is applied because we only remove the poster once vjs-ad-playing is applied.  This makes sense for a non-autoplay player but if the intent is to immediately play the video or preroll, the poster flash is distracting.

gkatsev:
Thanks @marguinbc. It would be good to figure out whether this needs to be fixed in 5.x.
Also, we need to decide whether we *need* it for 4.x or whether just fixing it in 5 and having people migrate is good enough.
Also, I'm closing this PR because stable is now for the 5.x releases. We'll need to figure out what to do for 4.x if we need to do a patch for it.

Issue #181, opened 2012-05-02
Feature Request: isPlaying() method


dominic-p:
Hello, in adding some custom event handlers I'm finding it necessary to determine if the video is currently playing or not. This is possible (albeit convoluted) with the HTML5 tech, but I have no idea how to go about it with the flash tech.

It might be nice to have an abstracted `isPlaying()` method that simply returns `true` or `false` based on if the video is playing or not.

shaharmor:
Btw, `paused()` doesn't correctly reflect `isPlaying()` because if the player is stuck on buffering/downloading the movie, `pasued()` will be false, while the video is not really playing.

oleynikd:
@shaharmor 100% agree! I'm also looking for `isPlaying()` method cuz `paused() == false` only tells that it is not paused...

framerate:
Was there ever a solution on this? I'm getting the same instance state and it's quite annoying.

misteroneill:
My $0.02 here is to listen to various events to track that state. Something like:

```js
var isPlaying = false;

player.on(['waiting', 'pause'], function() {
  isPlaying = false;
});

player.on('playing', function() {
  isPlaying = true;
});
```

The `playing` event definitely sounds like what you'd look for to determine that playback is beginning:

> Playback is ready to start after having been paused or delayed due to lack of media data.

Issue #133, opened 2012-01-19
Volume slider handle is offset too far on init


markisadesignerd:
Looks like when the player is initialized, the volume slider handle (vjs-volume-handle) is offset 100% from the left, which visually places it outside of its parent. When you interact with it for the first time, it snaps to the proper location (left:80%)

PR #2531, opened 2015-08-31
Updated the swf to v4.7.5

heff:
Related: videojs/video.js#181, videojs/video.js#133


dmlap:
lgtm

heff:
Merged, but not auto-closed because it's not going into master.

PR #2532, opened 2015-09-01
Removed the passing of playerOptions to plugins

### LINKED ISSUE #2510: Unable to initialize plugins via options
Looks like at some point we lost the ability to initialize a plugin via the `plugins` options object. Using the plugins example in sandbox:

### Works fine
```javascript
progressed = function(options) {
  this.on('progress', function(e) {
    console.log(this.bufferedPercent());
  });
};

// register the plugin
videojs.plugin('progressed', progressed);

// initialize it
vid1 = videojs('vid1');
vid1.progressed();
```

### Sadness
```javascript
var progressed = function(options) {
    this.on('progress', function(e) {
    console.log(this.bufferedPercent());
  });
};

videojs.plugin('progressed', progressed);

var vid1 = videojs('vid1', { plugins: { progressed: true }});
// => Uncaught TypeError: Cannot assign to read only property 'playerOptions' of true
```

heff:
...because it caused an error, and it actually creates an inconsistent
plugin init process, when considering plugins that are initialized
outside of the player init.

fixes #2510 

This was originally introduced in #2229, [by my request](https://github.com/videojs/video.js/pull/2229/files#r31776014). Seemed like a good idea at the time but unless anyone thinks it would be very valuable I think we should remove it for the consistency reason mentioned.

Also changed the Plugin var name in tests for clarity, before deciding just to remove the feature.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -157,7 +157,6 @@ class Player extends Component {
       let plugins = options.plugins;
 
       Object.getOwnPropertyNames(plugins).forEach(function(name){
-        plugins[name].playerOptions = playerOptionsCopy;
         if (typeof this[name] === 'function') {
           this[name](plugins[name]);
         } else {
--- a/test/unit/plugins.test.js
+++ b/test/unit/plugins.test.js
@@ -1,4 +1,4 @@
-import Plugin from '../../src/js/plugins.js';
+import registerPlugin from '../../src/js/plugins.js';
 import Player from '../../src/js/player.js';
 import TestHelpers from './test-helpers.js';
 import window from 'global/window';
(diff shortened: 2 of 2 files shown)


src/js/player.js
```diff
@@ -157,7 +157,6 @@ class Player extends Component {
       let plugins = options.plugins;
 
       Object.getOwnPropertyNames(plugins).forEach(function(name){
-        plugins[name].playerOptions = playerOptionsCopy;
```
heff: This is the only interesting change.

gkatsev:
Sounds good to me.

mmcc:
lgtm as well

PR #2533, opened 2015-09-01
Added a default swf url

### LINKED ISSUE #2488: SWF Default in 5.0
In 4.x, the location of the SWF defaulted to a Zencoder CDN URL. This default does not exist in `master` currently, but it would be good to have a default of some kind.

heff:
fixes #2488 

> The swf URL needs to be relative to the page (not video.js)
so this probably won't work for most, but it's the best we can do.
The CDN auto-adds a better swf URL for that specific version.
Otherwise you just need to set your swf location

### DIFF
--- a/src/js/tech/flash.js
+++ b/src/js/tech/flash.js
@@ -13,7 +13,6 @@ import FlashRtmpDecorator from './flash-rtmp';
 import Component from '../component';
 import window from 'global/window';
 import assign from 'object.assign';
-import pkg from '../../../package.json';
 
 let navigator = window.navigator;
 /**
@@ -76,7 +75,7 @@ class Flash extends Tech {
     // The CDN also auto-adds a swf URL for that specific version.
     if (!options.swf) {
       let protocol = ('https:' === window.location.protocol ? 'https://' : 'http://');
-      options.swf = protocol+'vjs.zencdn.net/swf/'+pkg.dependencies['videojs-swf']+'/video-js.swf';
+      options.swf = protocol+'vjs.zencdn.net/swf/__SWF_VERSION__/video-js.swf';
     }
 
     // Generate ID for swf object
--- a/src/js/tech/flash.js
+++ b/src/js/tech/flash.js
@@ -28,6 +28,12 @@ class Flash extends Tech {
   constructor(options, ready){
     super(options, ready);
 
+    // The swf URL needs to be relative to the page (not video.js)
+    // so this probably won't work for most, but it's the best we can do.
+    // The CDN auto-adds a better swf URL for that specific version.
+    // Otherwise you just need to set your swf location.
+    options.swf = options.swf || 'video-js.swf';
+
     // Set the source when ready
     if (options.source) {
       this.ready(function(){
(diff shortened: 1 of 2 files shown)


gkatsev:
Could we just default to the cdn url in the code?

heff:
If we think it's worth it we could go back to that. I was trying to completely separate out the CDN logic from the core codebase.

moay:
Just a comment from outside the box. ;-) I think it would be great to have the CDN as default fallback. Should be noted in the docs that this is relying on the CDN, but I think that most of the users would prefer having a default fallback which is already hosted over setting up their own one.

src/js/tech/flash.js
```diff
@@ -13,6 +13,7 @@ import FlashRtmpDecorator from './flash-rtmp';
 import Component from '../component';
 import window from 'global/window';
 import assign from 'object.assign';
+import pkg from '../../../package.json';
```
heff: @gkatsev can you check my sanity on this approach?

src/js/tech/flash.js
```diff
@@ -69,6 +70,15 @@ class Flash extends Tech {
   createEl() {
     let options = this.options_;
 
+    // If video.js is hosted locally you should also set the location
+    // for the hosted swf, which should be relative to the page (not video.js)
+    // Otherwise this adds a CDN url.
+    // The CDN also auto-adds a swf URL for that specific version.
+    if (!options.swf) {
+      let protocol = ('https:' === window.location.protocol ? 'https://' : 'http://');
+      options.swf = protocol+'vjs.zencdn.net/swf/'+pkg.dependencies['videojs-swf']+'/video-js.swf';
```
gkatsev: will the version string of the swf always be just the version? Could we at some point switch to using `^` or anything that could cause this to break?

src/js/tech/flash.js
gkatsev: Also, this does mean we include the whole package.json inside of the output. We probably want to use something like https://www.npmjs.com/package/browserify-versionify instead.

src/js/tech/flash.js
heff: I'd rather be more explicit with compiled dependencies. I made a note on the [cdn repo](https://github.com/videojs/cdn) "Patch releases should never change, including hosted dependencies". I don't want any risk of breaking patch-specific versions on the CDN.

src/js/tech/flash.js
heff: yeah, I was hoping to avoid another one of those... but it is better

heff:
Updated with @gkatsev's suggestions and tested. Appears to be working as it was before.

src/js/tech/flash.js
```diff
@@ -69,6 +69,15 @@ class Flash extends Tech {
   createEl() {
     let options = this.options_;
 
+    // If video.js is hosted locally you should also set the location
+    // for the hosted swf, which should be relative to the page (not video.js)
+    // Otherwise this adds a CDN url.
+    // The CDN also auto-adds a swf URL for that specific version.
+    if (!options.swf) {
+      let protocol = ('https:' === window.location.protocol ? 'https://' : 'http://');
```
gkatsev: why not just use a protocol relative url?

heff: Thanks. Was copy/pasting from an old version. Fixed.

gkatsev:
LGTM

PR #2536, opened 2015-09-01
Add the chinese readme translation

awesomes-cn:
awesomes.cn is an open source for  github web repos, we offer translation for repo's readme and improve this to let our chinese users use the repo easily.In addition, we want more more developer join us to enrich the chinese readmes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+[![中文说明](http://awesomes.oss-cn-beijing.aliyuncs.com/readme.png)](http://www.awesomes.cn/repo/videojs/video-js)
 ## Please Note!
 The master branch is now the development branch for 5.0 and should be considered unstable until the first 5.0 release. If you're looking for the most recent stable release, please refer to the [stable branch](https://github.com/videojs/video.js/tree/stable).
 


heff:
That's pretty cool. When we update the readme, will this version auto-update or will it need to be manually updated?

Is there a way we could include the translated readme in the repo?

awesomes-cn:
Our background task will always sync your repo in GitHub,when Detect the original readme changed,our team will update the chinese version manually. we are a free and open project(https://github.com/awesomes-cn/awesomes). we offer an open API,you can get the chinese version with `http://www.awesomes.cn/api/readme?owner=videojs&name=video.js` and the `format` param is mean to the format of the return content,this param is optional,default value is `markdown` ,another value is `html`.

gkatsev:
This PR is out of date. Also, given that this displays a badge, we should show it next to the other badges we have see https://github.com/videojs/video.js/blame/8d0de0fa51f08dea40748a9911ac992c889fc34d/README.md#L3
I'll close this PR for now but if it gets rebased, we can reopen, or a new PR can be made.

PR #2538, opened 2015-09-01
Added iOS and Android to unit tests

### LINKED ISSUE #2529: Are there functional tests for mobile platforms?
Hey, apologies if I missed it, but I just reading contributing.md, cloned and ran the tests. I noticed the config in the Gruntfile seems to indicate these are the covered platforms:

```
      // these are run locally on local browsers
      dev: { browsers: ['Chrome', 'Firefox', 'Safari'] },
      chromecanary: { browsers: ['ChromeCanary'] },
      chrome:       { browsers: ['Chrome'] },
      firefox:      { browsers: ['Firefox'] },
      safari:       { browsers: ['Safari'] },
      ie:           { browsers: ['IE'] },
      phantomjs:    { browsers: ['PhantomJS'] },

      // this only runs on PRs from the mainrepo on BrowserStack
      browserstack: { browsers: ['chrome_bs'] },
      chrome_bs:    { browsers: ['chrome_bs'] },
      firefox_bs:   { browsers: ['firefox_bs'] },
      safari_bs:    { browsers: ['safari_bs'] },
      ie11_bs:      { browsers: ['ie11_bs'] },
      ie10_bs:      { browsers: ['ie10_bs'] },
      ie9_bs:       { browsers: ['ie9_bs'] },
      ie8_bs:       { browsers: ['ie8_bs'] }
```

Now that you have BrowserStack, are there plans to cover Mobile Safari and the Android browsers? Or have you already somewhere else?

forbesjo:
This PR is to add Android and iOS as browsers for the unit tests. For some reason the BrowserStack tunnel isn't working correctly when using a 5.0 Android VM so I have targeted 4.4. 

I have also increased the timeouts to handle all this testing. All the browser VMs start simultaneously which means that if a VM doesn't start its testing before the capture timeout or no activity timeout is hit then it will give a warning, restart the browser VM and fail the build in the end. To avoid this the timeouts need to be bumped up to account for the total testing time. Polling timeout was also bumped up to avoid hitting the API rate limit.
 
This closes #2512 and closes #2529

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -156,7 +156,9 @@ module.exports = function(grunt) {
       ie11_bs:      { browsers: ['ie11_bs'] },
       ie10_bs:      { browsers: ['ie10_bs'] },
       ie9_bs:       { browsers: ['ie9_bs'] },
-      ie8_bs:       { browsers: ['ie8_bs'] }
+      ie8_bs:       { browsers: ['ie8_bs'] },
+      android_bs:   { browsers: ['android_bs'] },
+      ios_bs:       { browsers: ['ios_bs'] }
     },
     vjsdocs: {
       all: {
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -59,12 +59,12 @@ module.exports = function(config) {
     runnerPort: 9100,
     colors: true,
     logLevel: config.LOG_INFO,
-    captureTimeout: 60000,
-    browserNoActivityTimeout: 60000,
+    captureTimeout: 300000,
+    browserNoActivityTimeout: 300000,
 
     browserStack: {
       name: process.env.TRAVIS_BUILD_NUMBER + process.env.TRAVIS_BRANCH,
-      pollingTimeout: 10000
+      pollingTimeout: 30000
     },
     customLaunchers: getCustomLaunchers(),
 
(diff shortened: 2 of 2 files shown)


gkatsev:
It would be great if we can figure out how we can allow the browser to restart a set number of time before having it fail the build.

This PR LGTM, though.

forbesjo:
I haven't found that the tests are flaky enough in BrowserStack to require actual retries. The BrowserStack plugin retries 3 times by default but gives a warning each time it needs to retry which is what causes the task to fail without --force.

gkatsev:
Ah, so we need to make grunt not fail on a warning.

forbesjo:
We could use [grunt-force-task](https://www.npmjs.com/package/grunt-force-task) to keep going through warnings.

gkatsev:
Sounds interesting but seems like that will make errors not fail either then, no?

gkatsev:
Maybe this will help? https://www.npmjs.com/package/grunt-passfail#options-ignorewarn

forbesjo:
Looks like you're right about force covering up failures ([failed build](https://travis-ci.org/forbesjo/vjs-browserstack-demo/builds/78471592)). 

grunt-passfail makes me a little worried because the github link is broken. I think we could handle the retries in another PR. They shouldn't happen with the way the timeouts are setup in this PR.

gkatsev:
Yes, the retries thing should definitely not hold up this PR.

gkatsev:
Looks like it's on github but the url is wrong: https://github.com/sfrdmn/grunt-passfail

heff:
Should we wait to figure this out or pull this in now?

gkatsev:
I think we should pull this PR in. We can figure out how not to fail on warnings another time.

PR #2543, opened 2015-09-02
Allow passing body to XHR request

fvieudrin:
For POST requests for example, it allows to pass a body content to the request as the XHR library can do (option.body) : https://github.com/Raynos/xhr/blob/master/README.md

### DIFF
--- a/src/js/xhr.js
+++ b/src/js/xhr.js
@@ -127,6 +127,18 @@ vjs.xhr = function(options, callback){
   try {
     // Third arg is async, or ignored by XDomainRequest
     request.open(options.method || 'GET', options.uri, true);
+
+    var headers = request.headers = options.headers || {},
+      key;
+
+    if (request.setRequestHeader) {
+      for (key in headers) {
+        if (headers.hasOwnProperty(key)) {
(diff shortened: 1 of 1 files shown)


heff:
What would you be posting with the XHR lib built into video.js? Internally we only use it for GETs.

Also in master I believe we're still trying to change to raynos/XHR internally. #2321

@dmlap it'd be great to look back into that before 5.0 goes out.

fvieudrin:
I developed a plugin to log video errors, and I have to post many information in json to my web service. I already use this vjs.xhr() for GET calls, so I want to use the same lib for POST.

heff:
Thanks for the background, that makes sense. I'd like to get the rest of #2318 figured out before we move forward with this one.

It looks like Raynos/xhr#89 will finish it off, if you want to +1 it.

gkatsev:
We're now using [raynos/xhr](/raynos/xhr), so, I think this is covered by that?

gkatsev:
Closing this as fixed since we're using the XHR module. If it doesn't suffice, please open an issue. Thanks.

PR #2547, opened 2015-09-03
Added a default data attribute to fix the progress handle display in IE8

### LINKED ISSUE #2449: 5.0: In IE8, current time on scrubber hover is not correctly displayed
When hovering over the scrubber control in IE8 the current time is not being displayed, instead there's a little bar.

Here's a gif
vjs5hoverie8

When I disable and re-enable the css rule it does work properly

heff:
fixes #2449

The data attribute needed a default value (or just needed to exist from the start), otherwise IE8 would render it too small initially and never readjust to fit the time.


gkatsev:
LGTM

Issue #2507, opened 2015-08-24
Finalize default control bar components (and their options)


heff:
- Do we show current time in the bar or just on hover
- Vertical/horizontal (inline) volume
- Spinner icon (because Steve doesn't like that it's Picasa)

Followup
- Do we remove the other (hidden) volume components from the codebase

heff:
@mmcc, @gkatsev, @dmlap, I agree with the following statements. Please give me an agree/disagree for each.

- The controls that are currently shown by default are good and are in the best order
- We should continue to show just the remaining time, with the current time on hover
- We should remove the independent volume and mute controls from the control bar and rely on just the volume menu, which we've made work with at least the two primary volume designs

Default player controls:
![screen shot 2015-09-02 at 3 20 07 pm](https://cloud.githubusercontent.com/assets/166/9646192/23bc0124-5186-11e5-94d6-088760420462.png)

With hover states on:
![screen shot 2015-09-02 at 3 24 42 pm](https://cloud.githubusercontent.com/assets/166/9646256/c78bd1d0-5186-11e5-9880-79be5c8fbafd.png)

What the player looks like with all control bar components 'turned on'

![screen shot 2015-09-02 at 2 49 00 pm](https://cloud.githubusercontent.com/assets/166/9645645/d99d36d4-5181-11e5-89d0-be43e12cf474.png)

### Notes
- The volume slider needs to be fixed if we keep it, hoping we don't.
- The live control won't ever show now and needs the css to show it added back

And FYI, here's the CSS that was needed to show the ones that aren't showing by default. This uses the minimum strength of selectors required to show the control, in case we want the optional controls any easier to show. I expect to get requests on how to show the other time controls, to match the old way.

```css
    /* Hidden by default, selectively shown. 
       At this point this is how a skin designer would show them. */
    .video-js .vjs-current-time { display: block; }
    .video-js .vjs-time-divider { display: block; }
    .video-js .vjs-duration { display: block; }
    .video-js .vjs-mute-control { display: block; }
    .video-js .vjs-volume-control { display: block; }

    /* Could be broken. Can't find what's supposed to show it now */
    .video-js .vjs-live-control { display: block; }

    /* Programmatically hidden with vjs-hidden when no tracks */
    .video-js .vjs-chapters-button { display: block !important; }
    .video-js .vjs-subtitles-button { display: block !important; }
    .video-js .vjs-captions-button { display: block !important; }

    /* Hidden with state classes */
[...]

PR #2551, opened 2015-09-03
Fixed displaying the LIVE control bar item

heff:
...and also the layout in IE8 when live. Both are currently broken in the new skin.

Related to #2507

### DIFF
--- a/src/css/components/_live.scss
+++ b/src/css/components/_live.scss
@@ -1,11 +1,14 @@
-.video-js.vjs-live .vjs-time-control,
-.video-js.vjs-live .vjs-time-divider,
-.video-js.vjs-live .vjs-progress-control {
-  display: none;
-}
-
+// We are assuming there is no progress bar and using the live display
+// to fill in the middle space. Live+DVR will need to adjust this.
 .video-js .vjs-live-control {
-  display: none;
+  @include display-flex(flex-start);
+  @include flex(auto);
   font-size: 1em;
   line-height: 3em;
+}
+
+.vjs-no-flex .vjs-live-control {
+  display: table-cell;
+  width: auto;
+  text-align: left;
 }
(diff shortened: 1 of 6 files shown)


mmcc:
lgtm

src/css/components/_live.scss
```diff
@@ -1,11 +1,22 @@
...
+.vjs-live-control {
+  display: block;
+  font-size: 1em;
+  line-height: 3em;
 }
 
-.video-js .vjs-live-control {
+.vjs-live .vjs-time-control,
+.vjs-live .vjs-time-divider,
+.vjs-live .vjs-progress-control {
```
mmcc: In keeping with the `vjs-hidden` stuff we were talking about, should we add the class to these in `updateShowing`?

src/css/components/_live.scss
```diff
@@ -1,11 +1,22 @@
...
   display: none;
-  font-size: 1em;
-  line-height: 3em;
+}
+
+// Use the control spacer to fill in the middle space
+.vjs-live .vjs-custom-control-spacer {
+  @include display-flex();
```
mmcc: This can be removed because of the `updateShowing` method

src/css/components/_live.scss
```diff
@@ -1,11 +1,22 @@
...
+  @include flex(auto);
+}
+
+.vjs-no-flex.vjs-live .vjs-custom-control-spacer {
+  display: table-cell;
```
mmcc: Same with this

src/css/components/_live.scss
heff: If you mean have the live component manage the visibility of the other components, I think that would become a mess. I did consider having each of those components hide itself when the player is live, but I'm on the fence and not sure where to draw the line of when we use CSS states and when we use javascript for hiding. I'm also not sure how Live+DVR will affect these, so for now I think I'd lean towards leaving them like this.

src/css/components/_live.scss
heff: `updateShowing` doesn't control the visibility of the spacer, and the spacer is hidden by default so we still need this.

PR #2553, opened 2015-09-04
Made a few volume skin changes

heff:
Related to #2507

- Removed the non-default volume controls from the control bar
- Made the volume default to inline
- Started fixing the inline volume
  - Fixed hovering out of the slider while sliding (it would close on you)
  - Extended the hover area
  - There were a bunch of issues around tab focusing on the volume

Still need to figure out transitions on tab navigation. Looks a little wonky. Need @mmcc to check it out.

UPDATE: Also there seems to be something off with Safari's tab navigation now

### DIFF
--- a/src/css/components/menu/_menu-inline.scss
+++ b/src/css/components/menu/_menu-inline.scss
@@ -44,12 +44,14 @@
 }
 
 // Hover state
-.video-js .vjs-menu-button-inline:hover {
+.video-js .vjs-menu-button-inline:hover,
+.video-js .vjs-menu-button-inline.vjs-slider-active {
   // This width is currently specific to the inline volume bar.
-  width: 10em;
+  width: 12em;
 }
 
 .video-js .vjs-menu-button.vjs-menu-button-inline:hover .vjs-menu,
+.video-js .vjs-menu-button-inline.vjs-slider-active .vjs-menu,
 .video-js .vjs-menu-button-inline .vjs-menu.vjs-lock-showing {
   display: block;
   opacity: 1;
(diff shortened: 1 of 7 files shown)


src/css/components/menu/_menu-inline.scss
```diff
@@ -44,12 +44,14 @@
 }
 
 // Hover state
-.video-js .vjs-menu-button-inline:hover {
+.video-js .vjs-menu-button-inline:hover,
+.video-js .vjs-menu-button-inline.vjs-slider-active {
```
gkatsev: why not just use `:focus`?

heff: The menu button isn't actually in focus here, the internal slider is. I have the slider firing an event when it's active so the menu button can catch it add this class to itself.

heff:
All the volume in all their glory.

screen shot 2015-09-04 at 5 12 37 pm

Surprisingly working in IE8 too. I forced the inline volume to just be always showing in IE8. It might be possible to make it work the same but not sure it's worth the effort.

I also did a pass on reducing the selector strength on the involved components here and was able to clean up a bunch.

Could use another review.

src/js/control-bar/control-bar.js
```diff
@@ -46,7 +46,7 @@ ControlBar.prototype.options_ = {
   loadEvent: 'play',
   children: [
     'playToggle',
-    'volumeMenuButton',
+    { name: 'volumeMenuButton', inline: true },
```
heff: I guess I should switch the default of the volume menu button to inline

gkatsev:
I like the removal of specificity in the selectors. Seems like good changes.
LGTM.

heff:
Awesome, thanks!

PR #2554, opened 2015-09-04
Update package.json

paladox:
Since jshint was having problems wih 2.9.0 instead update by version since from now on the grunt jshint is updating verision everytime jshint is released to prevent having the problem again.


paladox:
Thanks.

PR #2555, opened 2015-09-04
Update package.json

paladox:


mmcc:
Thanks! Any reason there are 2 PRs to update the same file?

paladox:
Because if I update multiple in one patch it slows it down  but if I do it in multiple patches only one patch is slowed if the change is breaking or going to cause problems.

gkatsev:
Merging, thanks. In the future, it would be helpful if the name of the PR was more meaningful to the contents.

paladox:
Thanks and sorry.

gkatsev:
No worries.

PR #2557, opened 2015-09-04
Have userActive() run instead of just setting internal var. Fixes #2550

### LINKED ISSUE #2550: `vjs-user-active` or `vjs-user-inactive` aren't added by default
Only until the first triggered event happens does one of these get added.
`vjs-user-active` should probably be added by default on first play since at that time the user is definitely active.

gkatsev:


pam:

Commit: 7f392398266562e7aed3e6d6d7eb207c464558d0
Build details: https://travis-ci.org/pam/video.js/builds/78802069

heff:
Makes sense to me. Seems like by setting the internal property it might have been preventing the activity interval from changing the state. Specifically because userAcitve() checks ifs changed before adding the classes/triggering the event.

LGTM

PR #2558, opened 2015-09-04
Skip copying webkitMovementX/Y

forbesjo:
webkitMovementX/Y seem to be deprecated in Chrome 45+, this fix skips those keys.
Fixes #2548

PR #2559, opened 2015-09-04
Skip copying webkitMovementX/Y (stable version)

forbesjo:
The fix in #2558 but for stable


dmlap:
lgtm

heff:
Merged into stable

PR #2568, opened 2015-09-07
Update link to true destination to save redirect

hubdotcom:


gkatsev:
We probably should just make the site work with either one and not do a redirect. `www` is to 2002.
@heff ?

heff:
It's actually an issue with hosting a domain on s3, or at least was when we did the last site. It requires/required a subdomain. It's worth verifying because I agree non-www is better.

hubdotcom:
Does require Amazon DNS but a potential solution https://aws.amazon.com/blogs/aws/root-domain-website-hosting-for-amazon-s3/

gkatsev:
This has been fixed.

PR #2571, opened 2015-09-09
Sass variable might be defined globally

aksakalli:
Variables should be defined with `!default` syntax in case someone want to define customize it:
http://stackoverflow.com/questions/32253344/gulp-dependency-looking-in-wrong-folder/32475297

Similar Sass projects like `bootstrap` always use this syntax.

### DIFF
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -1,7 +1,7 @@
 @import "variables";
 @import "utilities";
 
-$icon-font-path: 'font';
+$icon-font-path: 'font' !default;
 @import "../../node_modules/videojs-font/scss/icons";
 
 @import "components/layout";


mmcc:
So the reasoning behind not doing this originally is that we didn't necessarily want to encourage people to override the SCSS when creating custom themes. We wanted to think of the base theme as more like the styling of the default video element that could be built on top of with ease.

I *think* we might have gone away from that a little, but if we're going to do it we should go full-on and do this for all the defined variables.

gkatsev:
A better question, though, is why that variable is being defined outside of the vars file?

mmcc:
I noticed that the other day and looked through chat to see if there was a justification. Not sure what the reasoning was at the time, but it should probably join its other var friends.

gkatsev:
I think we do actually want to pull this change in. And probably the other variables as well.

gkatsev:
Thanks for the PR. Ended up getting this fix in via #2574.

PR #2573, opened 2015-09-09
Make background color have opacity instead of everything.

gkatsev:
This is because otherwise, it could be hard to read the dropdowns and labels for the captions settings dialog.
`rgba()` is a sass function that takes a color and an opacity and outputs the rgba value for.
I kept the old background-color directive as a fallback for browsers that don't support rgba


mmcc:
:+1:

PR #2582, opened 2015-09-11
Added a new loading spinner

heff:
It looks like this: http://codepen.io/heff/pen/QjjZVW

Good?

### DIFF
--- a/grunt.js
+++ b/grunt.js
@@ -74,13 +74,13 @@ module.exports = function(grunt) {
     },
     dist: {},
     watch: {
-      jshint: {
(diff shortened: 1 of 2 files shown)


gkatsev:
It looks kind of weird when both the heads reach the top.

heff:
I think I see what you mean, and I think it could be because the animation didn't cover from 0 to 100%. Does it look any better now?
http://codepen.io/heff/pen/QjjZVW

Globulopolis:
@heff yes, much better :)

heff:
Updated so the two things don't show through each other at the top.
http://codepen.io/heff/pen/ZbQePZ

PR #2584, opened 2015-09-12
Created a private vars scss file

heff:
...to be clear about which variables should be modified,
at least at this point.

### DIFF
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -1,4 +1,5 @@
 @import "variables";
+@import "private-variables";
 @import "utilities";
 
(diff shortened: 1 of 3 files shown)


src/css/video-js.scss
```diff
@@ -1,4 +1,5 @@
 @import "variables";
+@import "private-variables";
```
gkatsev: what about just importing it in `variables.scss`?

heff: That'd be fine with me. Is that better for some reason?

gkatsev: No idea. @mmcc thoughts?

PR #2589, opened 2015-09-15
Added an 'attributes' arg to createEl()

### LINKED ISSUE #2176: vjs.obj.merge and vjs.createEl don't handle HTML attributes (which are case-insensitive) correctly
[HTML attributes are case-insensitive](http://www.w3.org/TR/html-markup/documents.html#case-insensitivity). JavaScript and vjs.obj.merge are case sensitive, so a merged object may contain two variables which are different only because of their case. vjs.createEl simply iterates through the properties object, applying them to the new element in the order it finds them; the el.setAttribute method it uses is case insensitive, but **order matters**.

So if this code were used:

````js
vjs.MenuItem.prototype.createEl = function(type, props){
  return vjs.Button.prototype.createEl.call(this, 'li', vjs.obj.merge({
    className: 'vjs-menu-item',
    role: 'menuitem',
    tabindex: -1,
    innerHTML: this.localize(this.options_['label'])
  }, props));
};

....

vjs.Button.prototype.createEl = function(type, props){
  var el;

  // Add standard Aria and Tabindex info
  props = vjs.obj.merge({
    className: this.buildCSSClass(),
    'role': 'button',
    'aria-live': 'polite', // let the screen reader user know that the text of the button may change
    tabIndex: 0
  }, props);

  el = vjs.Component.prototype.createEl.call(this, type, props);
````
the button element's `tabIndex` would not be set correctly.

(Note that this problem doesn't seem to have surfaced in the existing code, but it has the potential to; I discovered it while doing some development/enhancement. The attribute `tabIndex` is already used in both lower-case and camel-case in the existing code, so the potential to cause a problem is there).

### PR DESCRIPTION
**Pull request by heff**
fixes #2176

Properties and attributes need to be handled separately in createEl(). This adds a third arg for attributes and a deprecation warning for the cases were we expected attributes to come through the properties object.

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -5,6 +5,7 @@ import document from 'global/document';
 import window from 'global/window';
 import  * as Guid from './guid.js';
 import log from './log.js';
+import tsml from 'tsml';
 
 /**
  * Shorthand for document.getElementById()
@@ -40,9 +41,9 @@ export function createEl(tagName='div', properties={}, attributes={}){
     // We originally were accepting both properties and attributes in the
     // same object, but that doesn't work so well.
     if (propName.indexOf('aria-') !== -1 || propName === 'role' || propName === 'type') {
-      log.warn(`Setting attributes in the second argument of createEl()
+      log.warn(tsml(`Setting attributes in the second argument of createEl()
                 has been deprecated. Use the third argument instead.
-                createEl(type, properties, attributes). Attempting to set ${propName} to ${val}.`);
+                createEl(type, properties, attributes). Attempting to set ${propName} to ${val}.`));
       el.setAttribute(propName, val);
     } else {
       el[propName] = val;
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -4,6 +4,7 @@
 import document from 'global/document';
 import window from 'global/window';
 import  * as Guid from './guid.js';
+import log from './log.js';
 
 /**
  * Shorthand for document.getElementById()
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: src/js/utils/dom.js [hunk 1]
```diff
@@ -29,25 +30,28 @@ export function getEl(id){
...
-      } else {
-       el[propName] = val;
-      }
+    let val = properties[propName];
+
+    // See #2176
+    // We originally were accepting both properties and attributes in the
+    // same object, but that doesn't work so well.
+    if (propName.indexOf('aria-') !== -1 || propName === 'role' || propName === 'type') {
+      log.warn(`Setting attributes in the second argument of createEl()
```
gkatsev: does this have a bunch of extra whitespace in it? Might want to grab https://www.npmjs.com/package/tsml

gkatsev:
LGTM, we probably want to mention this in the wiki for the 5.0 transition.

heff:
Yeah, good point. I'll add it.

src/js/utils/dom.js
heff: Cool. Pulled in tsml.

src/js/utils/dom.js
```diff
@@ -29,25 +31,28 @@ export function getEl(id){
  * @return {Element}
  * @function createEl
  */
-export function createEl(tagName='div', properties={}){
+export function createEl(tagName='div', properties={}, attributes={}){
```
gkatsev: the jsdocs for this were never added.
Also, we should probably document exactly what the different between the two would be.

PR #2590, opened 2015-09-15
Switched player.tech to tech_, to promote it as private

### LINKED ISSUE #2060: Make player.tech a private property
player.tech should have always been a private property, but now that we're not mangling vars it's especially important to call that out.

Just want to make sure this happens for 5.0.

heff:
closes #2060

Just changes player.tech to player.tech_ wherever it's used, because it's important to promote that the tech is a private reference that plugin authors shouldn't be using. This was mangled in 4.x, so should be ok to change.

### DIFF
--- a/src/js/control-bar/mute-toggle.js
+++ b/src/js/control-bar/mute-toggle.js
@@ -21,14 +21,14 @@ class MuteToggle extends Button {
     this.on(player, 'volumechange', this.update);
 
     // hide mute toggle if the current tech doesn't support volume control
-    if (player.tech && player.tech['featuresVolumeControl'] === false) {
+    if (player.tech_ && player.tech_['featuresVolumeControl'] === false) {
       this.addClass('vjs-hidden');
     }
 
     this.on(player, 'loadstart', function() {
       this.update(); // We need to update the button to account for a default muted state.
 
-      if (player.tech['featuresVolumeControl'] === false) {
+      if (player.tech_['featuresVolumeControl'] === false) {
         this.addClass('vjs-hidden');
       } else {
         this.removeClass('vjs-hidden');
--- a/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
+++ b/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
@@ -125,8 +125,8 @@ class PlaybackRateMenuButton extends MenuButton {
    * @method playbackRateSupported
    */
   playbackRateSupported() {
-    return this.player().tech
-      && this.player().tech['featuresPlaybackRate']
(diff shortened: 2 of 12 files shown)


gkatsev:
Sounds good to me.

moay:
Just one question concerning this. How do I access a "non standard" tech function in a component with the tech being private?

I want to access a function of a custom tech which is not a default function like `volume()` or `src()`. I built a component which can access the player internally like this: `this.player_`, but if I try to access the function like this: `this.player_.myFunction()` it doesn't work. Neither does `this.player_.tech.myFunction()` (still on rc82). 

How to eiter publish the function from within the tech or to access the function via some other way from the component?

Thanks. :-)

heff:
How are you adding a function to the player so that
this.player_.myFunction() isn't working? That should work, and is ok to do
because it's a private var of `this`.

If you have a specific feature you want to access on a tech, you should
submit and issue about it so we can find a way to make it work the same
across all techs the same way. If we start adding tech-specific functions,
it breaks the model where any plugin can work with any tech.

moay:
Here is the basic tech (**very** reduced) setup:

```javascript
(function() {
    var Tech = videojs.getComponent('Tech');
    var MyTech = videojs.extends(Tech, {
        myFunction: function(){
            return something;
        }
    });
    videojs.registerComponent('MyTech', MyTech);
})();
```

Now in my Component (simple button, ES6), I can't access this function via `this.player().myFunction()`. In rc82, I can do `this.player().tech.myFunction()`, but if I get the purpose of the tech being made private, this is not how it's supposed to be used. Maybe I'm just not clever enough... ;-) Is there some other way to publish a method from the tech?

gkatsev:
In 5.0, the tech no longer has a reference to the player. But the player should still have `tech_` available. And just because it's private and discourage to use, doesn't mean that you shouldn't necessarily use it. It depends on the particular usecase.
Also, for [contrib-hls](/videojs/videojs-contrib-hls) we ended up adding a reference to the player with the name of the tech for us to use. `player.hls` would reference the hls tech. Though, we're transitioning away from a tech for it and to source handlers.

moay:
Okay, thanks. :)

heff:
Please don't use tech_ until you've at least submitted an issue about why you needed to use it and what the player is lacking. I haven't seen a usecase yet where we shouldn't be looking into how to expose the information at the player level, in a way that's consistent between techs (even if it's just reporting that the info/feature isn't supported).

The consistency between techs is actually growing in importance for video.js users (and users of video.js plugins), so we need to keep being religious about that.

mmcc:
lgtm

Issue #1676, opened 2014-11-19
LoadedAllData event fails to fire after Flash file played


ds00424:
In the below sample code below (insert your favorite flv and mp4 files), the flv will load and play, and then the mp4 will load but not play.  It seems the LoadedAllData event fails to fire after the flv is loaded/played.

Heff points to a bug.   See:
http://stackoverflow.com/questions/26985646/video-js-loadedalldata-firing-only-once

Here is the jist of that post (just in case):
This line of code turns the events off as soon as the buffer reaches 100%, but then never resets itself on a later source load.
https://github.com/videojs/video.js/blob/ce18a9af740eb2a01f2b0efe2d10299de32818ce/src/js/media/media.js#L192
It should restart these events when the loadstart event is fired again by the flash player, signaling a new source has been loaded.

My sample code:
```html
<html>
  <head>
    <link type="text/css" href="/js/video.js/video-js.css" rel="stylesheet" />
    <script type="text/javascript" src="/js/jquery/jquery-2.1.1.min.js"></script>
    <script type="text/javascript" src="/js/video.js/video.js"></script>

    <script>
        videojs.options.flash.swf = "/js/video.js/video-js.swf"
        $(document).ready(function () {
            var playIt = function playVid() {
                alert("Event");
                this.play();
            }

            videojs("video1", {}, function(){
                this.on('loadedalldata', playIt);
                this.on('ended', function() {
                    this.src({ type: "video/mp4", src: "/files/test.mp4" });
                    this.load();
                });
                this.src({ type: "video/flv", src: "/files/barsandtone.flv" });
                this.load();
            });
        });
    </script>
  </head>
  <body>
    <div id="div1">
        <video id="video1" class="video-js vjs-default-skin"></video>
    </div>
  </body>
</html>
```

billyrennekamp:
any updates on this issue? running into it myself while using the videojs-contrib-ads plugin

heff:
I think we need to deprecate the `loadedalldata` event for 5.0. It's a made up event (not from the video element) that has less use these days for a few reasons:
- For most cases, you should use the `canplaythrough` event instead of `loadedalldata`.
- Browsers manage their buffer more carefully now, they don't just load the whole video at once
- It doesn't apply to live events

Any objections to doing that?

mmcc:
Kill it with fire.

PR #2591, opened 2015-09-15
Removed the loadedalldata event

heff:
Want to get this out before 5.0. Reasoning from #1676. 

> I think we need to deprecate the loadedalldata event for 5.0. It's a made up event (not from the video element) that has less use these days for a few reasons:
- For most cases, you should use the canplaythrough event instead of loadedalldata.
- Browsers manage their buffer more carefully now, they don't just load the whole video at once
- It doesn't apply to live events


gkatsev:
Should we deprecate the other vjs-only events as well? Like `firstplay`?

heff:
I think I tried that once and got push back because it's actually useful. Don't we use it in ads or somewhere externally?

mmcc:
lgtm

PR #2594, opened 2015-09-15
Updated #2318 - Use raynos/xhr - rebased and fixed test errors

heff:
closes #2318

The only major change from #2318 is I moved sinon XHR stubbing to the cover all tests, because with raynos/xhr we have to stub it before the xhr module is ever loaded.


src/js/video.js
```diff
@@ -484,6 +453,29 @@ videojs.off = Events.off;
...
+ *       uri: "/foo",
+ *       headers: {
+ *         "Content-Type": "application/json"
+ *       }
+ *     }, function (err, resp, body) {
+ *       // check resp.statusCode
+ *     });
+ *
+ * Check out the [full
+ * documentation](https://github.com/Raynos/xhr/blob/master/README.md)
```
gkatsev: https://github.com/Raynos/xhr/blob/v2.1.0/README.md would be safer but not a huge issue.

gkatsev:
`Showing 8 changed files with 36 additions and 211 deletions. `
Woot. LGTM

src/js/video.js
heff: Good call

heff:
Thanks!

naugtur:
In xhr 2.0+ (and sone earlier versions) there are multiple ways you can stub or replace `xmlhttprequest` any time. Sorry I didn't notice what you did earlier. Adding a single line after the one with sinon stub would suffice.
I can help you get that done if you wish to go back to your previous stubbing location.

Anyway, I'm glad to see more adoption :)

heff:
Thanks for the note @naugtur (and for the help in getting this done!). It looked like the other ways to stub xhr required us to change the app code to allow passing an xhr object in. That didn't seem ideal, and the global stub is actually kind of nice because it removes the issue from any future tests. If there's other ways to stub that I'm missing though I'd be interested to know.

naugtur:
I'd suggest using the other option from the docs (at the end of README.md) and doing just 
```js
xhr.XMLHttpRequest = MockXMLHttpRequest
```

Whenever you do that, from then on, `xhr` will use `MockXMLHttpRequest` even if it was required earlier in some other module.

heff:
Ah great! Missed that option. Thanks!

PR #2604, opened 2015-09-16
Extend createTimeRange to enable creation of TimeRanges with multiple ranges

imbcmdth:
No checks are performed to ensure that the resulting TimeRanges-like object is "normalized" as the spec requires.

Supports not passing an argument to `start` and `end` and returns the first range for backward  compatibility.

Throws an error with Chrome's error message if `start` or `end` is passed an index that is outside the supported ranges.

### DIFF
--- a/src/js/utils/time-ranges.js
+++ b/src/js/utils/time-ranges.js
@@ -1,3 +1,5 @@
+import log from './log.js';
+
 /**
  * @file time-ranges.js
  *
@@ -43,6 +45,7 @@ function createTimeRangesObj(ranges){
 
 function getRange(fnName, valueIndex, ranges, rangeIndex){
   if (rangeIndex === undefined) {
+    log.warn(`DEPRECATED: Function '${fnName}' on 'TimeRanges' called without an index argument.`);
     rangeIndex = 0;
   }
   rangeCheck(fnName, rangeIndex, ranges.length - 1);
--- a/src/js/utils/time-ranges.js
+++ b/src/js/utils/time-ranges.js
@@ -6,14 +6,24 @@
  * return the start and end times for a range
  * TimeRanges are returned by the buffered() method
  *
- * @param  {Number} start Start time in seconds
- * @param  {Number} end   End time in seconds
- * @return {Object}       Fake TimeRange object
+ * @param  {(Number|Array)} Start of a single range or an array of ranges
+ * @param  {Number} End of a single range
  * @private
- * @method createTimeRange
+ * @method createTimeRanges
  */
-export function createTimeRange(start, end){
-  if (start === undefined && end === undefined) {
+export function createTimeRanges(start, end){
+  if (Array.isArray(start)) {
+    return createTimeRangesObj(start);
+  } else if (start === undefined || end === undefined) {
+    return createTimeRangesObj();
+  }
+  return createTimeRangesObj([[start, end]]);
+}
+
+export { createTimeRanges as createTimeRange };
+
+function createTimeRangesObj(ranges){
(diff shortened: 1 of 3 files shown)


src/js/utils/time-ranges.js
```diff
@@ -25,8 +35,22 @@ export function createTimeRange(start, end){
...
-    start: function() { return start; },
-    end: function() { return end; }
+    length: ranges.length,
+    start: getRange.bind(null, 'start', 0, ranges),
+    end: getRange.bind(null, 'end', 1, ranges)
   };
 }
+
+function getRange(fnName, valueIndex, ranges, rangeIndex){
+  if (rangeIndex === undefined) {
```
heff: No args here actually throws an error with native timeranges. If we want them to be interchangeable we probably shouldn't let people do this.

heff: Just saw the note about backward compatibility. Now I'm on the fence.

pam:

Commit: 46117a15d8212d73b5657b25c8ee22ca22e98e04
Build details: https://travis-ci.org/pam/video.js/builds/80735336

heff:
Looks good so far. We'll need to update the video.js file and export both naming versions on the videojs object also. And now that I see it here we could probably just do the aliasing in the video.js file, instead of in the time-ranges.js file, but up to you.

pam:

Commit: 7c3e1e0a8a18029424f3f54868967e406691fe93
Build details: https://travis-ci.org/pam/video.js/builds/80734903

imbcmdth:
That would require changes everywhere `createTimeRange` is currently used (about half a dozen or so files.) I was trying to make as minor a change as possible so close to the milestone but I can do that if you deem it as necessary.

src/js/utils/time-ranges.js
imbcmdth: We could add a deprecation warning to the no-index case.

heff:
> That would require changes everywhere

Exporting it on the videojs object would require that? I just mean having both videojs.createTimeRange and videojs.createTimeRanges
https://github.com/videojs/video.js/blob/master/src/js/video.js#L382

src/js/utils/time-ranges.js
heff: That would be ok with me. Do we actually use it in that way anywhere? If we are then either way we'd have to update those uses.

imbcmdth:
I meant that:
> we cold probably just do the aliasing in the video.js file, instead of in the time-ranges.js file

Would require changes everywhere that createTimeRange is used via an import statement.

pam:
Ahh, I see. Ok.

>

-- 
Sent from mobile

imbcmdth:
@heff Added the new `createTimeRanges` function and alias to the videojs object.

pam:

Commit: d7b7baa84520dac21194517e8339241bf75d9bc8
Build details: https://travis-ci.org/pam/video.js/builds/80937622

src/js/utils/time-ranges.js
heff: This one's still unresolved. Are we punting on it?

pam:

Commit: eefb775e5f4f915bd85afbfa7a8c3dbb1b6bbf4e
Build details: https://travis-ci.org/pam/video.js/builds/81466427

src/js/utils/time-ranges.js
imbcmdth: @heff I added the deprecation warning, none of our plugins use the no-index style but we can't be sure that no one else has. I'd rather we provide a clear indication that we intend to remove that "feature" in the near future.

heff:
lgtm!

PR #2605, opened 2015-09-17
Fix the double loadstart event and fix it good

### LINKED ISSUE #2588: loadstart getting triggered twice
This may be related to #1194

heff:
fixes #2588 @pat can you confirm?

Took me a day and half to fully wrap my brain around this. The code comments give some insight into the complexity.

With these changes you can now init a player on a video element of any loading state and the player will emit all the loading events, from loadstart to canplaythrough, to catch up to the current state.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -36,6 +36,8 @@ class Html5 extends Tech {
     // anyway so the error gets fired.
     if (source && (this.el_.currentSrc !== source.src || (options.tag && options.tag.initNetworkState_ === 3))) {
       this.setSource(source);
+    } else {
+      this.handleLateInit_(this.el_);
     }
 
     if (this.el_.hasChildNodes()) {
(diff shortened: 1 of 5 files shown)


src/js/tech/html5.js
```diff
@@ -167,6 +169,83 @@ class Html5 extends Tech {
...
+      // about to fire, and either way the player hasn't seen it yet.
+      // We don't want to fire loadstart prematurely here and cause a
+      // double loadstart so we'll wait and see if it happens between now
+      // and the next loop, and fire it if not.
+      // HOWEVER, we also want to make sure it fires before loadedmetadata
+      // which could also happen between now and the next loop, so we'll
+      // watch for that also.
+      let loadstartFired = false;
+      let setLoadstartFired = function() { loadstartFired = true; };
+      this.on('loadstart', setLoadstartFired);
```
gkatsev: maybe `this.one`?
Also, probably would've been a lot simpler with promises :P

src/js/tech/html5.js
```diff
@@ -167,6 +169,83 @@ class Html5 extends Tech {
...
+
+      let triggerLoadstart = function() {
+        // We did miss the original loadstart. Make sure the player
+        // sees loadstart before loadedmetadata
+        this.trigger('loadstart');
```
gkatsev: should we check to see if `loadstartFired` is true here? Or would the ready handler below would've fired before this would've hit and thus the handler would have been removed so it doesn't actually get triggered?

heff: Yeah, I think both. I don't think there's a case where both loadstart and loadedmetadata would be queued and fired in this limbo timeframe, but better to be safe.

src/js/tech/html5.js
heff: `one` seemed unnecessary since we're always removing it on ready anyway. I guess it felt better to clean it up more manually, since not cleaning it up always would have really bad side effects.

heff:
@misteroneill would you be able to verify this?

(Sorry @pat, end of a long day)

pat:
@heff no worries, it happens :)

misteroneill:
After attempting to verify: if you bind a listener _before_ the player is ready, both "loadstart" and "ready" are fired twice. If you bind a listener _on_ ready, both are fired once.

heff:
Thanks @misteroneill, these should be fixed now, if you get another second to try it. We had double ready triggers in the player. Not exactly sure how that happened.

misteroneill:
@heff Yeah, looking good now. :+1:

heff:
Thanks @misteroneill!

PR #2619, opened 2015-09-21
Make sure we only get one styles-defaults element.

### LINKED ISSUE #2606: make sure only one `style-defaults` element is ever added to the DOM
If you require videojs in a plugin but also include videojs separately as a global, you end up with two `style-defaults` elements. We want to make sure:
* `style-defaults` only ever gets added once
* make sure that we don't have any other side-effects of having multiple videojs's.

The particular usecase is to require videojs directly so a plugin could, for example, require videojs itself to get a component to extend.

gkatsev:
We do so by querySelecting it and if it is null, then we create one, otherwise, we assume it is set up correctly.
Fixes #2606

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -97,14 +97,17 @@ var videojs = function(id, options, ready){
 };
 
 // Add default styles
-let style = stylesheet.createStyleElement('vjs-styles-defaults');
-let head = document.querySelector('head');
-head.insertBefore(style, head.firstChild);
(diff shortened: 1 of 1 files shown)


pam:

Commit: 098fe2a8498712e320dc397048b96badae0c959b
Build details: https://travis-ci.org/pam/video.js/builds/81431120

heff:
lgtm!

PR #2621, opened 2015-09-21
Restore API tests

forbesjo:
This PR is to add the API tests back to the testing suite.


heff:
I actually removed these from the pipeline because I was hoping we could just merge them with the `videojs` and `player` tests, instead of having two separate setups. If that seems like too much work we can pull this in.

forbesjo:
I'll see if I can merge it in.

forbesjo:
Merged the tests together, the `extends` test wont fail this time but we already know it's an issue.

test/karma.conf.js
```diff
@@ -14,7 +14,8 @@ module.exports = function(config) {
       '../build/temp/ie8/videojs-ie8.min.js',
       '../test/globals-shim.js',
       '../test/unit/**/*.js',
-      { pattern: '../src/**/*.js', watched: true, included: false, served: false }
```
heff: This was in there because it causes karma to watch the source files for changes (but not serve or run them). I don't believe that was happening otherwise. Was there a need to remove it?

heff:
So I wasn't very clear, but what I meant was moving the individual tests out of api.js and into `video.test.js `, `player.test.js`, and `component.test.js` or wherever they best belong. If we require the compiled video.min.js in tests again we're going back to our super slow testing.

heff:
To be more clear, I don't think there's a need for a separate api.js test suite. I think we can cover the same thing through testing the objects in our normal test suite. The api.js suite was mostly valuable when we were mangling variables.

test/karma.conf.js
forbesjo: Sorry, had a brain fart and was reading it as being included in the tests.

forbesjo:
That makes more sense, I'll work on putting the tests in the appropriate modules.

forbesjo:
Closing this to revisit later

PR #2622, opened 2015-09-22
Remove phantomjs dependency

forbesjo:
This PR is to remove phantomjs as a dependency for testing because I believe that it is rarely used. Pull Requests test against Firefox on Travis or BrowserStack browsers and local tests use Chrome by default.


gkatsev:
LGTM

heff:
lgtm

PR #2625, opened 2015-09-22
Re-expose TextTrack

### LINKED ISSUE #2623: Re-expose TextTrack class
This is used by at least one plugin; so, after conversation with @gkatsev, we decided it would be best if it were re-exposed as a component (despite not truly being a component).

misteroneill:
Fixes #2623

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -1,7 +1,6 @@
 /**
  * @file text-track.js
  */
-import Component from '../component.js';
 import TextTrackCueList from './text-track-cue-list';
 import * as Fn from '../utils/fn.js';
 import * as Guid from '../utils/guid.js';
@@ -298,5 +297,4 @@ var indexOf = function(searchElement, fromIndex) {
   return -1;
 };
 
-Component.registerComponent('TextTrack', TextTrack);
 export default TextTrack;
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -1,6 +1,7 @@
 /**
  * @file text-track.js
  */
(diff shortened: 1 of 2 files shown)


pam:

Commit: c7dd29bc6de68228dec4098b490fe4df7c1f234b
Build details: https://travis-ci.org/pam/video.js/builds/81602820

gkatsev:
LGTM. Though, still feels a bit weird to register it as a component.
Also, do we want to expose the other text track definitions as well? I guess those can wait till 5.1 if needed.

heff:
Yeah, components are a pretty specific concept in video.js with expected functions and such, so I don't think it's safe to register TextTrack as one, at least as it is. And maybe more importantly we don't want to set a trend of registering non-components this way.

I think if we want this done quickly we should just export TextTrack onto the videojs object.

misteroneill:
I agree that it's weird as a component. Would it be preferable to simply expose it as `videojs.TextTrack`? Ultimately, it might make sense to have a getter/setter for non-component classes in the future.

dmlap:
+1 to `videojs.TextTrack`. I don't think we should add getters/setters for this sort of thing. It's just a class, we don't need a dynamic lookup mechanism for it.

misteroneill:
I thought it might make sense for consistency with `getComponent`/`registerComponent`, but that's neither here nor there.

I'll make the change to expose it as a property of `videojs` instead.

heff:
lgtm

pam:

Commit: e220c10e1a962492e15013e538e2cf39f4dceb17
Build details: https://travis-ci.org/pam/video.js/builds/81609899

PR #2627, opened 2015-09-22
Removed timeranges no-index tests because they are invalid now

heff:
They were throwing deprecation warnings in the test logs.

screen shot 2015-09-22 at 11 32 52 am


gkatsev:
LGTM

PR #2636, opened 2015-09-24
Use button state for icon

misteroneill:
Instead of using the player state to decide which icon a play control should display, this keeps that state in the button. This allows videojs-contrib-ads to keep working.

Fixes an internal Brightcove issue.


heff:
lgtm!

gkatsev:
LGTM

PR #2638, opened 2015-09-25
This adds touch and move listeners to the parent of volumeBar.

gkatsev:
This is useful because it allows us to more easily be able to interact with the volume bar. Instead of just being able to click or touch on the 4 pixel high bar, we can click or touch on the whole height of the container where the bar is located in.

### DIFF
--- a/src/js/control-bar/volume-control/volume-bar.js
+++ b/src/js/control-bar/volume-control/volume-bar.js
@@ -22,7 +22,6 @@ class VolumeBar extends Slider {
     super(player, options);
     this.on(player, 'volumechange', this.updateARIAAttributes);
     player.ready(Fn.bind(this, this.updateARIAAttributes));
-    player.ready(() => this.on(this.el().parentNode, ['mousedown', 'touchdown'], this.handleMouseDown));
   }
 
   /**
@@ -50,19 +49,6 @@ class VolumeBar extends Slider {
     }
 
     this.player_.volume(this.calculateDistance(event));
-  }
-
-  handleMouseDown(event) {
-    super.handleMouseDown(event);
-
-    this.on(this.el().parentNode, ['mousemove', 'touchmove'], this.handleMouseMove);
-    this.on(this.el().parentNode, ['mouseup', 'touchend'], this.handleMouseUp);
-  }
-
-  handleMouseUp(event) {
-    super.handleMouseUp(event);
-
-    this.off(this.el().parentNode, ['mousemove', 'touchmove'], this.handleMouseMove);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


pam:

Commit: e29c47cb485c87560448212268526dc2636f943e
Build details: https://travis-ci.org/pam/video.js/builds/82188478

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -22,6 +22,7 @@ class VolumeBar extends Slider {
     super(player, options);
     this.on(player, 'volumechange', this.updateARIAAttributes);
     player.ready(Fn.bind(this, this.updateARIAAttributes));
+    player.ready(() => this.on(this.el().parentNode, ['mousedown', 'touchdown'], this.handleMouseDown));
```
heff: Can we assume that this is safe for anything that parentNode could be, assuming that volume bar could be used independently of the volume menu?

gkatsev: Not sure. I could also move that into the volume menu button instead. And it's not quite safe as it is because some tests are failing because of this.
Either way, this is not a blocker for 5.0.

dmlap: @gkatsev sounds like we should move it into the parent component.

gkatsev: Ok, I'll move it to the volume menu button.

dmlap:
LGTM, though we'll have to be careful that we never allow volume menu buttons to be created without an associated volume bar.

gkatsev:
It gets created inside of `createMenu` which is also where the other methods are referenced in. So, unless someone overrides `createMenu`, we shouldn't have a problem. Though, even in that case, it might be ok because then the other methods won't get called.

gkatsev:
I guess I should wrap them in `if (this.volumeBar)` just in case, maybe?

src/js/control-bar/volume-menu-button.js
```diff
@@ -65,6 +67,12 @@ class VolumeMenuButton extends MenuButton {
     this.on(this.volumeBar, ['sliderinactive', 'blur'], function(){
       this.removeClass('vjs-slider-active');
     });
+
+    // we show the volumeBar on focus for keyboard accessibility
+    // when using the mouse, you don't want it to stay up, so, we blur it manually
+    this.on('mouseout', function() {
+      this.el().blur();
```
dmlap: mouseout could occur if they're using keyboard navigation but they also run their mouse over the volume bar.

gkatsev: I think that may be ok. Or at least, something that we would need to live with.
Unless we want to (if we are even able to) know whether we entered into this via a keyboard or mouse and then only blur on mouseout if we entered with the mouse.

gkatsev: I think I'm going to remove this and open a new issue for this to get a proper fix.

gkatsev: Opened a new issue: https://github.com/videojs/video.js/issues/2918

dmlap:
LGTM

PR #2639, opened 2015-09-25
Remove z-index on big play button

misteroneill:
This was interfering with some internal BCOV plugins and doesn't seem to be necessary in video.js proper.


gkatsev:
LGTM

PR #2664, opened 2015-10-01
add asyncSrc load feature (asynchronous load source on play button click)

astr0junk:
add ability to initialize video tag without sources and load them after click on play button

Example:

```javascript

 var options = {
      asyncSrc: function (callback) { // your function which for async request
        return $.ajax({ //jquery ajax for example
          url: '%urlToSrc%',
          type: "get"
        }).done(function (resp) {
          //get type and (or) url from response data

          return callback({type: resp.type, src: resp.url}); 
        })
      }
    }

    videojs('vid1', options);

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.idea
 build/files/*
 build/temp/*
 docs/api/*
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks for the PR but I don't think that we want to add such a feature to videojs quite yet.
You can get a similar affect by listening to the `play` event on an empty video and then putting in the source then.
I think that if you would want to make it easier to use, you could turn this into a videojs plugin and we can help you do that.

Issue #2033, opened 2015-04-14
Move default options into the constructor


heff:
In https://github.com/videojs/video.js/pull/1993#discussion-diff-28035717 we discussed moving options into the constructor instead of on the prototype. Secondarily removing the `this.options()` accessor.

For the player we use the prototype options as global options for the player, but this could be done by providing a global options object that overrides the default when the player is initialized.

I'm not sure what impact removing the `options()` accessor would have, but right now it creates another source of truth, which isn't great. This would force us to be more specific about providing the info from options that we need.

@dmlap @gkatsev, are there places externally where we're relying on the options() getter?

d7omee:
H

heff:
Instead of attacking this completely in 5.0 I want to start the process by deprecating the use of `component.options()` and fixing the uses of it over time before removing it in 6.0. I think we use it a lot, especially in plugins, so trying to fix everything for 5.0 would be too much.

dmlap:
People are definitely using the options() getter externally but I support deprecating it to avoid confusion over options versus properties. Pushing actual removal to 6.0 seems like a good move.

heff:
#2229 checked the first box for deprecating `options()`. I added a check list item for moving prototype options into the constructor. Pushing this to 5.1.

PR #2668, opened 2015-10-02
Modal Dialog

misteroneill:
This PR introduces two new components to video.js: `CloseButton` and `ModalDialog` with tests.

The general idea is to strike a balance between assumptions of how a modal should work - e.g., you want a "close" button or you only want to define the contents once - and flexibility.

At Brightcove, there are five plugins (both open and closed source) which individually define some kind of a modal. Every one in slightly and subtly different; so, part of the goal is that this can work for all those while anticipating potential needs of other plugin authors.

Much of the flexibility comes from adding hooks to important lifecycle events: fill, empty, open, close. This allows things like event binding to happen when users want - for example, it might make sense to bind events on `"modalfill"` so we are sure the DOM is in place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "object.assign": "^4.0.1",
     "safe-json-parse": "^4.0.0",
     "tsml": "1.0.1",
-    "videojs-font": "git+https://github.com/videojs/font.git#182aa4c2ee46cca7ca00155c4cdbd89acbad5a65",
+    "videojs-font": "1.4.0",
     "videojs-ie8": "1.1.0",
     "videojs-swf": "5.0.0-rc1",
     "vtt.js": "git+https://github.com/gkatsev/vtt.js.git#vjs-v0.12.1",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "object.assign": "^4.0.1",
     "safe-json-parse": "^4.0.0",
     "tsml": "1.0.1",
-    "videojs-font": "1.3.0",
+    "videojs-font": "git+https://github.com/videojs/font.git#182aa4c2ee46cca7ca00155c4cdbd89acbad5a65",
     "videojs-ie8": "1.1.0",
     "videojs-swf": "5.0.0-rc1",
     "vtt.js": "git+https://github.com/gkatsev/vtt.js.git#vjs-v0.12.1",
--- a/src/css/_utilities.scss
+++ b/src/css/_utilities.scss
@@ -105,6 +105,6 @@
 }
 
 %icon-default {
-  @extend %fill-space;
+  @extend %overlay-space;
   text-align: center;
 }
--- a/src/css/_utilities.scss
+++ b/src/css/_utilities.scss
@@ -96,7 +96,7 @@
   order: $value;
 }
 
-%overlay-space {
+%fill-parent {
   position: absolute;
   top: 0;
   left: 0;
@@ -105,6 +105,6 @@
 }
 
 %icon-default {
-  @extend %overlay-space;
+  @extend %fill-parent;
   text-align: center;
 }
--- a/src/css/_utilities.scss
+++ b/src/css/_utilities.scss
@@ -1,3 +1,5 @@
+@import "utilities/linear-gradient";
+
 @mixin background-color-with-alpha($color, $alpha) {
   background-color: $color;
   background-color: rgba($color, $alpha);
@@ -94,11 +96,15 @@
   order: $value;
 }
 
-%icon-default {
+%overlay-space {
   position: absolute;
   top: 0;
   left: 0;
   width: 100%;
   height: 100%;
+}
+
+%icon-default {
+  @extend %fill-space;
   text-align: center;
 }
--- a/src/css/components/_error.scss
+++ b/src/css/components/_error.scss
@@ -1,4 +1,23 @@
-.video-js .vjs-error-display .vjs-modal-dialog-content {
+.vjs-error .vjs-error-display .vjs-modal-dialog-content {
   font-size: 1.4em;
   text-align: center;
+}
+
+.vjs-error .vjs-error-display:before {
+  color: #fff;
+  content: 'X';
+  font-family: $text-font-family;
+  font-size: 4em;
+  left: 0;
+
+  // In order to center the play icon vertically we need to set the line height
+  // to the same as the button height
+  line-height: 1;
+  margin-top: -0.5em;
+  position: absolute;
+  text-shadow: 0.05em 0.05em 0.1em #000;
+  text-align: center; // Needed for IE8
+  top: 50%;
+  vertical-align: middle;
+  width: 100%;
 }
--- a/src/css/components/_error.scss
+++ b/src/css/components/_error.scss
@@ -1,48 +1,4 @@
-.vjs-error-display {
-  display: none;
-}
-
-.vjs-error .vjs-error-display {
-  display: block;
-  position: absolute;
-  left: 0;
-  top: 0;
-  width: 100%;
-  height: 100%;
-}
-
-.vjs-error .vjs-error-display:before {
-  content: 'X';
-  font-family: $text-font-family;
-  font-size: 4em;
-  color: #fff;
-  /* In order to center the play icon vertically we need to set the line height
-  to the same as the button height */
-  line-height: 1;
-  text-shadow: 0.05em 0.05em 0.1em #000;
-  text-align: center /* Needed for IE8 */;
-  vertical-align: middle;
-
-  position: absolute;
-  left: 0;
-  top: 50%;
-  margin-top: -0.5em;
-  width: 100%;
-}
-
-.vjs-error-display div {
-  position: absolute;
-  bottom: 1em;
-  right: 0;
-  left: 0;
+.video-js .vjs-error-display .vjs-modal-dialog-content {
   font-size: 1.4em;
   text-align: center;
-  padding: 3px;
-
-  @include background-color-with-alpha(#000, 0.5);
-}
-
-.vjs-error-display a,
-.vjs-error-display a:visited {
-  color: #66A8CC;
 }
--- a/src/css/components/_modal-dialog.scss
+++ b/src/css/components/_modal-dialog.scss
@@ -7,27 +7,3 @@
   @extend %fill-parent;
   z-index: 1;
 }
-
-// Ultimately, this would be moved to its own component partial, but
-// for now it's unclear what styles would be generic and which specific
-// to this context!
-.vjs-modal-dialog .vjs-close {
-  cursor: pointer;
-  font-size: 1.2em; // 12px
-  height: auto;
-  padding: 0.75em 1em; // 9px 12px
-  position: absolute;
-  right: 0;
-  top: 0;
-  width: auto;
-  z-index: 2;
-}
-
-// Reset inherited control-text styles.
-.vjs-modal-dialog .vjs-close .vjs-control-text {
-  clip: auto;
-  height: auto;
-  margin: 0;
-  position: static;
-  width: auto;
-}
--- a/src/css/components/_modal-dialog.scss
+++ b/src/css/components/_modal-dialog.scss
@@ -1,26 +1,9 @@
-%fill-player {
-  height: 100%;
-  left: 0;
-  position: absolute;
-  top: 0;
-  width: 100%;
-}
-
 .video-js .vjs-modal-dialog {
-  @extend %fill-player;
-
-  $gradient-start: rgba(0, 0, 0, 0.8);
-  $gradient-stop: rgba(255, 255, 255, 0);
-  background: $gradient-start;
-  background: -moz-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
-  background: -webkit-gradient(left top, left bottom, color-stop(0%, $gradient-start), color-stop(100%, $gradient-stop));
-  background: -webkit-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
-  background: -o-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
-  background: -ms-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
-  background: linear-gradient(top bottom, $gradient-start 0%, $gradient-stop 100%);
+  @extend %overlay-space;
+  @include linear-gradient(180deg, rgba(0, 0, 0, 0.8), rgba(255, 255, 255, 0));
 
   .vjs-modal-dialog-content {
-    @extend %fill-player;
+    @extend %overlay-space;
     z-index: 1;
   }
 
--- a/src/css/components/_modal-dialog.scss
+++ b/src/css/components/_modal-dialog.scss
@@ -5,5 +5,8 @@
 
 .vjs-modal-dialog .vjs-modal-dialog-content {
   @extend %fill-parent;
+
+  font-size: 1.2em; // 12px
+  padding: 2.5em; // 30px
   z-index: 1;
 }
--- a/src/css/components/_modal-dialog.scss
+++ b/src/css/components/_modal-dialog.scss
@@ -1,34 +1,33 @@
 .video-js .vjs-modal-dialog {
-  @extend %overlay-space;
+  @extend %fill-parent;
   @include linear-gradient(180deg, rgba(0, 0, 0, 0.8), rgba(255, 255, 255, 0));
+}
 
-  .vjs-modal-dialog-content {
-    @extend %overlay-space;
-    z-index: 1;
-  }
+.vjs-modal-dialog .vjs-modal-dialog-content {
+  @extend %fill-parent;
+  z-index: 1;
+}
 
-  // Ultimately, this would be moved to its own component partial, but
-  // for now it's unclear what styles would be generic and which specific
-  // to this context!
-  .vjs-close {
-    cursor: pointer;
-    font-size: 1.2em; // 12px
-    height: auto;
-    padding: 0.75em 1em; // 9px 12px
-    position: absolute;
-    right: 0;
-    top: 0;
-    width: auto;
-    z-index: 2;
+// Ultimately, this would be moved to its own component partial, but
+// for now it's unclear what styles would be generic and which specific
+// to this context!
+.vjs-modal-dialog .vjs-close {
+  cursor: pointer;
+  font-size: 1.2em; // 12px
+  height: auto;
+  padding: 0.75em 1em; // 9px 12px
+  position: absolute;
+  right: 0;
+  top: 0;
+  width: auto;
+  z-index: 2;
+}
 
-    .vjs-control-text {
-
-      // Reset inherited control-text styles.
-      clip: auto;
-      height: auto;
-      margin: 0;
-      position: static;
-      width: auto;
-    }
-  }
+// Reset inherited control-text styles.
+.vjs-modal-dialog .vjs-close .vjs-control-text {
+  clip: auto;
+  height: auto;
+  margin: 0;
+  position: static;
+  width: auto;
 }
--- a/src/css/components/_modal-dialog.scss
+++ b/src/css/components/_modal-dialog.scss
@@ -1,0 +1,51 @@
+%fill-player {
+  height: 100%;
+  left: 0;
+  position: absolute;
+  top: 0;
+  width: 100%;
+}
+
+.video-js .vjs-modal-dialog {
+  @extend %fill-player;
+
+  $gradient-start: rgba(0, 0, 0, 0.8);
+  $gradient-stop: rgba(255, 255, 255, 0);
+  background: $gradient-start;
+  background: -moz-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
+  background: -webkit-gradient(left top, left bottom, color-stop(0%, $gradient-start), color-stop(100%, $gradient-stop));
+  background: -webkit-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
+  background: -o-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
+  background: -ms-linear-gradient(top, $gradient-start 0%, $gradient-stop 100%);
+  background: linear-gradient(top bottom, $gradient-start 0%, $gradient-stop 100%);
+
+  .vjs-modal-dialog-content {
+    @extend %fill-player;
+    z-index: 1;
+  }
+
+  // Ultimately, this would be moved to its own component partial, but
+  // for now it's unclear what styles would be generic and which specific
+  // to this context!
+  .vjs-close {
+    cursor: pointer;
+    font-size: 1.2em; // 12px
+    height: auto;
+    padding: 0.75em 1em; // 9px 12px
+    position: absolute;
+    right: 0;
+    top: 0;
+    width: auto;
+    z-index: 2;
+
+    .vjs-control-text {
+
+      // Reset inherited control-text styles.
+      clip: auto;
+      height: auto;
+      margin: 0;
+      position: static;
+      width: auto;
+    }
+  }
+}
--- a/src/css/components/_modal-dialog.scss
+++ b/src/css/components/_modal-dialog.scss
@@ -7,6 +7,7 @@
   @extend %fill-parent;
 
   font-size: 1.2em; // 12px
-  padding: 2.5em; // 30px
+  line-height: 1.5; // 18px
+  padding: 30px;
   z-index: 1;
 }
--- a/src/js/close-button.js
+++ b/src/js/close-button.js
@@ -16,7 +16,7 @@ class CloseButton extends Button {
   }
 
   buildCSSClass() {
-    return `vjs-close ${super.buildCSSClass()}`;
+    return `vjs-close-button ${super.buildCSSClass()}`;
   }
 
   handleClick() {
--- a/src/js/close-button.js
+++ b/src/js/close-button.js
@@ -1,0 +1,28 @@
+import Button from './button';
+import Component from './component';
+
+/**
+ * The `CloseButton` component is a button which fires a "close" event
+ * when it is activated.
+ *
+ * @extends Button
+ * @class CloseButton
+ */
+class CloseButton extends Button {
+
+  constructor(player, options) {
+    super(player, options);
+    this.controlText(options && options.controlText || this.localize('Close'));
+  }
+
+  buildCSSClass() {
+    return `vjs-close ${super.buildCSSClass()}`;
+  }
+
+  handleClick() {
+    this.trigger({type: 'close', bubbles: false});
+  }
+}
+
+Component.registerComponent('CloseButton', CloseButton);
+export default CloseButton;
--- a/src/js/error-display.js
+++ b/src/js/error-display.js
@@ -46,14 +46,12 @@ class ErrorDisplay extends ModalDialog {
    */
   content() {
     let error = this.player().error();
-    this.content_ = error ? this.localize(error.message) : null;
-    return this.content_;
+    return error ? this.localize(error.message) : '';
   }
 }
 
 ErrorDisplay.prototype.options_ = mergeOptions(ModalDialog.prototype.options_, {
   fillAlways: true,
-  slug: 'error-display',
   uncloseable: true
 });
 
--- a/src/js/error-display.js
+++ b/src/js/error-display.js
@@ -27,9 +27,12 @@ class ErrorDisplay extends ModalDialog {
   }
 
   /**
-   * Build the modal's CSS class.
+   * Include the old class for backward-compatibility.
+   *
+   * This can be removed in 6.0.
    *
    * @method buildCSSClass
+   * @deprecated
    * @return {String}
    */
   buildCSSClass() {
@@ -39,22 +42,18 @@ class ErrorDisplay extends ModalDialog {
   /**
    * Generates the modal content based on the player error.
    *
-   * @return {Element|Null}
+   * @return {String|Null}
    */
   content() {
     let error = this.player().error();
-    if (error) {
-      this.content_ = Dom.createEl('span');
-      this.content_.textContent = this.localize(error.message);
-    } else {
-      this.content_ = null;
-    }
+    this.content_ = error ? this.localize(error.message) : null;
     return this.content_;
   }
 }
 
 ErrorDisplay.prototype.options_ = mergeOptions(ModalDialog.prototype.options_, {
   fillAlways: true,
+  slug: 'error-display',
   uncloseable: true
 });
 
--- a/src/js/error-display.js
+++ b/src/js/error-display.js
@@ -2,53 +2,61 @@
  * @file error-display.js
  */
 import Component from './component';
-import  * as Dom from './utils/dom.js';
+import ModalDialog from './modal-dialog';
+
+import * as Dom from './utils/dom';
+import mergeOptions from './utils/merge-options';
 
 /**
- * Display that an error has occurred making the video unplayable
+ * Display that an error has occurred making the video unplayable.
  *
- * @param {Object} player  Main Player
- * @param {Object=} options Object of option names and values
- * @extends Component
+ * @extends ModalDialog
  * @class ErrorDisplay
  */
-class ErrorDisplay extends Component {
+class ErrorDisplay extends ModalDialog {
 
+  /**
+   * Constructor for error display modal.
+   *
+   * @param  {Player} player
+   * @param  {Object} [options]
+   */
   constructor(player, options) {
     super(player, options);
-
-    this.update();
-    this.on(player, 'error', this.update);
+    this.on(player, 'error', this.open);
   }
 
   /**
-   * Create the component's DOM element
+   * Build the modal's CSS class.
    *
-   * @return {Element}
-   * @method createEl
+   * @method buildCSSClass
+   * @return {String}
    */
-  createEl() {
-    var el = super.createEl('div', {
-      className: 'vjs-error-display'
-    });
-
-    this.contentEl_ = Dom.createEl('div');
-    el.appendChild(this.contentEl_);
-
-    return el;
+  buildCSSClass() {
+    return `vjs-error-display ${super.buildCSSClass()}`;
   }
 
   /**
-   * Update the error message in localized language
+   * Generates the modal content based on the player error.
    *
-   * @method update
+   * @return {Element|Null}
    */
-  update() {
-    if (this.player().error()) {
-      this.contentEl_.innerHTML = this.localize(this.player().error().message);
+  content() {
+    let error = this.player().error();
+    if (error) {
+      this.content_ = Dom.createEl('span');
+      this.content_.textContent = this.localize(error.message);
+    } else {
+      this.content_ = null;
     }
+    return this.content_;
   }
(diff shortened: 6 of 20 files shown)


misteroneill:
This is a WIP, but I'm opening a PR to start gathering feedback and spark discussion as to whether or not this belongs in video.js proper.

pam:

Commit: 369f3d86c6a886ed2b54e2cbfd0d41ede8d8ea74
Build details: https://travis-ci.org/pam/video.js/builds/83376989

pam:

Commit: a7d18b45223b141b929b6096ae22a82a7d5d3d7c
Build details: https://travis-ci.org/pam/video.js/builds/83376481

pam:

Commit: d2988f2bfd8550e2bfda3f27aa27d407e770d556
Build details: https://travis-ci.org/pam/video.js/builds/83392382

pam:

Commit: 3a08ded21093aed904f3b3e7bb258ef65805563f
Build details: https://travis-ci.org/pam/video.js/builds/83391230

src/css/components/_modal-dialog.scss
```diff
@@ -0,0 +1,51 @@
...
+  top: 0;
+  width: 100%;
+}
+
+.video-js .vjs-modal-dialog {
+  @extend %fill-player;
+
+  $gradient-start: rgba(0, 0, 0, 0.8);
+  $gradient-stop: rgba(255, 255, 255, 0);
+  background: $gradient-start;
```
gkatsev: this should move into a mixin.

misteroneill: I agree. :+1:

heff:
This touches on our strategy for UI components in plugins moving forward, which doesn't necessarily need to be solved here but will impact it.

I think we need to strongly define how plugins should be accessing UI components to use/extend them. It's totally inefficient for plugins to be pulling in entire copies of video.js as a dependency to get at a component. One step forward might be to break out UI components into their own repos/projects, which is something we've talked about either way, but even then it seems really inefficient to have a plugin compile Component along with itself, when another copy of Component will already be in the page with Video.js.

It sounds like peer dependencies are [taking a big step forward in npm@3](http://blog.npmjs.org/post/110924823920/npm-weekly-5). Are they a viable option yet?

As far as this issue is concerned, if we go the route of having copies of components compiled with plugins, then I don't think it makes sense to pull this into core. If only because we might be moving all other components out.

If we go the route of saying plugins should use the copies included in video.js then a modal does seem like a root-level UI component, and we could probably justify it just because we could use it for the internal error display. If it weren't for that we'd be deciding if Video.js should include UI components just for what it uses internally, or if it should include additional general purpose UI components for the sake of plugins. I think I would lean towards the former in that case.

gkatsev:
> It's totally inefficient for plugins to be pulling in entire copies of video.js as a dependency to get at a component.

This is only really an issue if npm plus a packing system like browserify is *not* used. npm will auto-dedupe packages and if the end-user uses videojs and plugins from npm, you only include one version of video.js. 

> One step forward might be to break out UI components into their own repos/projects

One solution is to release each component into a separate repo. Then plugins can require just what they need. Basically, we can have a concept of component-plugins and have all the built-in videojs components be those themselves. This probably would need a pretty major rework to the API and how things work. But might be a good thing in the end. Maybe this plus redefined plugin API can be The Next Big Thing™ for videojs.

> it seems really inefficient to have a plugin compile Component along with itself, when another copy of Component will already be in the page with Video.js.

Ultimately, we can't really support everything. For this, we might need to find the happy path (I recommend browserify+npm) and support that path fully. That doesn't mean we shouldn't support other ways too, but it could mean that payloads end up slightly larger because there isn't a good way of doing dedupes and stuff like that if we're supporting globals for everything. However, I guess plugins can be smarter and have different builds. One browserify build that requires everything and assumed the end-user also uses browserify (or webpack). One global build that assumes none of the dependencies are available and/or one global build that assumes all the dependencies will be available in the global scope.

> It sounds like peer dependencies are taking a big step forward in npm@3. Are they a viable option yet?

Yes, seems like in npm 3, peer dependencies are finally not a hassle.

> we could probably justify it just because we could use it for the internal error display

I think we should update the internal error display to use this modal dialog.

misteroneill:
It seems like the big concern here is that builds for plugins will have video.js bundled with them. I agree that should be avoided! I think @gkatsev is right that npm/browserify solves this problem. Although, I do wonder how it would work with multiple versions of video.js in its dependency tree (e.g. if plugin-a depends on video.js@5.1.x and plugin-b depends on video.js@5.0.x)?

I agree with both of you that the internal error display could be updated to use this; it justifies the inclusion in video.js proper in my mind. One of my main concerns was having UI components in video.js that were only there for plugin authors. The flip side of that introduced another concern: having a plugin whose only purpose was to provide blank-slate UI components. Sort of a meta-plugin. Perhaps that's not a bad thing. Any opinions on that?

pam:

Commit: a9c21fdeb3cabf07843a7d3752989590806210a6
Build details: https://travis-ci.org/pam/video.js/builds/83708323

heff:
npm de-duping is great but it's not a complete answer unfortunately 
- most people apparently still [don't use any module bundler](http://ashleynolan.co.uk/blog/frontend-tooling-survey-2015-results#q6-javascript-module-bundlers), and only 17% said browserify
- most plugins in the wiki today don't have commonJS or requireJS support, including videojs-youtube and google's IMA3

If we were a full app framework we might be able to force stricter development pipelines on plugin authors and users, but video.js is a widget that's going to be dropped into a lot of existing pipelines, and using the independently compiled versions of plugins is gonna be the norm for a while.

If peer dependencies are better now it looks to me like they could support the best of both worlds. Are there still reasons why we wouldn't want to go that route?

pam:

Commit: 5058f8bae9496b83ad1a8961c2a5df59a002d42a
Build details: https://travis-ci.org/pam/video.js/builds/83708501

misteroneill:
I don't think anyone was suggesting forcing people into an npm/browserify workflow. Maybe I'm not clear on the problem. Plugins already depend on video.js (implicitly as a global); so, I'm not sure how this work would change that or how calling `videojs.getComponent('ModalDialog')` introduces issues that `videojs.plugin('foo', foo)` doesn't?

pam:

Commit: c1cfd051d158b05cdfd81966a5febf565a596657
Build details: https://travis-ci.org/pam/video.js/builds/83746275

pam:

Commit: 0fb1999bbff661a1b479ea922db8c545ba76dfbc
Build details: https://travis-ci.org/pam/video.js/builds/83765023

pam:

Commit: 83a68f5b98941888d153c614e14c0f68e0702b09
Build details: https://travis-ci.org/pam/video.js/builds/83767966

heff:
Check out [videojs-dock](https://github.com/brightcove/videojs-dock/blob/master/index.js#L1) as an example. It doesn't rely on Video.js as a global, and if you were to compile that for users looking to include it as a standalone js file next to video.js (*not* using browserify), it would include another copy of Video.js.

I may have jumped ahead, but that's what I assumed we meant by the npm workflow that doesn't rely on peer dependencies.

gkatsev:
It actually does currently rely on videojs as a global. I use browserify-shim to shim it in so the code looks like it can be required. Also, not sure why I didn't use es6 module syntax there.

heff:
Oh wow, that's cool. I stand corrected.

src/js/modal-dialog.js
```diff
@@ -0,0 +1,386 @@
...
+   * @param  {Object} [options]
+   * @param  {Mixed} [options.content=undefined]
+   *         Provide customized content for this modal.
+   *
+   * @param  {Boolean} [options.disposeOnClose=false]
+   *         If `true`, the modal can only be opened once. It will be
+   *         disposed as soon as it's closed, which is useful for one-off
+   *         modals.
+   *
+   * @param  {Boolean} [options.fillAlways=false]
```
misteroneill: Part of me wonders if this should be the default case and the "fill once" case should be the special case. The reason I went this way is that "fill once" is more conservative (fewer DOM changes, etc).

pam:

Commit: ffc410b9c8c0c7174f80d361e817aa260442da37
Build details: https://travis-ci.org/pam/video.js/builds/83782570

gkatsev:
Also, what I was kind of suggesting is that we target global and browserify packaging. This should hopefully be relatively simple to do by using browserify to write the plugins. Then with various plugins like browserify-shim, we can produce two builds of the plugin. One of the builds is the "happy path" I was describing above which uses browserify and depends on video.js and that is what you get when you `require` the plugin. Otherwise, you get a build that assumes that videojs/deps are available in the global namespace and can be shimed using browserify-shim. Don't have any concrete examples of this, but I believe it should be possible to do relatively easily. I can give it a shot with the videojs-dock plugin when I get back from my conference.

pam:

Commit: d4792b8c6f523b3ee13e5beb5a81ee5a7659cf07
Build details: https://travis-ci.org/pam/video.js/builds/83951157

pam:

Commit: 1f366f63ffc354bcbf5efd0078c48a56b6b2411d
Build details: https://travis-ci.org/pam/video.js/builds/83950845

pam:

Commit: 50c5241d927ee59066ac43e38cdf87524aee96f7
Build details: https://travis-ci.org/pam/video.js/builds/83975010

chemoish:
+1 @gkatsev, interested in seeing that.

I currently have to shim video.js with webpack as well. However, video.js itself could also benefit from module loader packaging.

pam:

Commit: d44816fc6edddc131165fa3535346926bcaa582e
Build details: https://travis-ci.org/pam/video.js/builds/84140951

pam:

Commit: 6fcfd90dfb3cc7aaeb3b64d7e34796309c0abbbf
Build details: https://travis-ci.org/pam/video.js/builds/84142059

heff:
But is there still a reason why we wouldn't just use peer dependencies now? They're at least trying to support plugin systems specifically, and if we could then there'd be one way for plugins to interact with video.js instead of two, no double documentation etc.

src/css/_utilities.scss
```diff
@@ -94,11 +96,15 @@
   order: $value;
 }
 
-%icon-default {
+%overlay-space {
```
heff: 'overlay-space' feels a little vague to me. Would 'fill-parent' be an accurate description?

src/css/components/_modal-dialog.scss
```diff
@@ -0,0 +1,34 @@
...
+    font-size: 1.2em; // 12px
+    height: auto;
+    padding: 0.75em 1em; // 9px 12px
+    position: absolute;
+    right: 0;
+    top: 0;
+    width: auto;
+    z-index: 2;
+
+    .vjs-control-text {
```
heff: We've avoided using nesting like this as much as possible up to this point. I can't state all the reasoning behind it but let's not start adding it until we've made an intentional choice around it.

src/css/_utilities.scss
misteroneill: Heh. It was originally fill-parent. I'll go back to that. :+1:

src/js/error-display.js
```diff
@@ -2,53 +2,60 @@
...
-   * @method update
+   * @return {String|Null}
    */
-  update() {
-    if (this.player().error()) {
-      this.contentEl_.innerHTML = this.localize(this.player().error().message);
-    }
+  content() {
+    let error = this.player().error();
+    this.content_ = error ? this.localize(error.message) : null;
```
heff: `error` is going to be an object, so we don't want content set to that. Previously we were using `this.player().error().message`.

src/css/components/_modal-dialog.scss
misteroneill: Sure. If I had to hazard a guess, it's probably to avoid overly-specific selectors.

src/js/error-display.js
heff: Nvm, was reading too quickly.

src/js/error-display.js
```diff
@@ -2,53 +2,60 @@
...
+    return this.content_;
   }
 }
 
+ErrorDisplay.prototype.options_ = mergeOptions(ModalDialog.prototype.options_, {
```
heff: Does mergeOptions not already happen as part of the inherited init process?

misteroneill: My reading was that it merges passed-in options with prototype options, but it doesn't recurse up through the prototype chain.

src/js/modal-dialog.js
```diff
@@ -0,0 +1,408 @@
...
+ * @class ModalDialog
+ */
+class ModalDialog extends Component {
+
+  /**
+   * Constructor for modals.
+   *
+   * @param  {Player} player
+   * @param  {Object} [options]
+   * @param  {Mixed} [options.content=undefined]
```
heff: I haven't tried this way of documenting options but I like it. We should make sure our jsdoc pipeline supports it.

src/js/modal-dialog.js
heff: At first glance Modal feels really complicated for what I think of as a box that displays text. I feel like it'd be nice if we could simplify it but I also don't understand all the use cases going into it. Do you think you could document the different use cases you're trying to solve for with the different options here?

src/js/error-display.js
heff: Yeah, that's interesting. It seems like that's how it should work though, right? That would be a benefit of #2033 that I hadn't really seen before.

heff:
We should do a hangout to talk through this one, if you're up for it.

src/js/modal-dialog.js
misteroneill: Yeah, I can elaborate on it more in the PR description.

misteroneill:
Absolutely. Maybe tomorrow? Let me know what works for you.

heff:
My day's clear (Pacific time). If there's a good time for you go ahead and make an event, and we can open it up to @gkatsev and anyone else that wants to participate.

src/js/modal-dialog.js
misteroneill: Another candidate for change would be to rename the `open`/`close` methods to `show`/`hide` (overriding/expanding on those methods of `Component`).

misteroneill:
@gkatsev I know you're on vacation/at a conference, but I'll add you to the invite on the off chance you want to jump back into video.js land for an hour. :wink:

gkatsev:
I'll pass. Though, I think anything related to plugins should happen in a separate conversation as it doesn't directly relate to modals, especially if they're in core. :)

src/js/video.js
```diff
@@ -29,6 +29,9 @@ import xhr from 'xhr';
 import Html5 from './tech/html5.js';
 import Flash from './tech/flash.js';
 
+import CloseButton from './close-button';
+import ModalDialog from './modal-dialog';
```
heff: Can be removed.

misteroneill:
Hangout take-aways:

- Dispose on close by default.
- Open it by default? Jon thinks if you're constructing, maybe not, but from a factory function.
- Lose SCSS nesting.
- Simplify `ErrorDisplay#content`. Doesn't need to set `content_`.
- Remove `slug` in favor of a component-level option in a separate PR.
- Remove `fillImmediately` and `openImmediately`.
- `Player#createModal()` factory for simple, `alert`-style use-cases.
- Thinking of the player as a "mini-DOM"/window.
- Remove `import`s from `src/js/video.js`; no longer needed.
- Exposing DOM helpers from video.js?

pam:

Commit: 28e7bd6bd3adbcb2a386e2c97f126518cd2528a4
Build details: https://travis-ci.org/pam/video.js/builds/84399873

pam:

Commit: 411bf5097cdb328130a31a420bda174200964103
Build details: https://travis-ci.org/pam/video.js/builds/84418542

pam:

Commit: c62c8f5b10167f67a721f5317a6a0e31aefec11d
Build details: https://travis-ci.org/pam/video.js/builds/84419577

misteroneill:
This is ready for another review, despite @pam's insistence on choking on IE8 tests...

misteroneill:
Alright. That should make these tests pass in IE8.

pam:

Commit: 3002c0ea1684b0b47250992b6095836725c89f27
Build details: https://travis-ci.org/pam/video.js/builds/84567971

gkatsev:
There's a lot of errors being logged. Would be nice to clean those up since @heff spent so much time doing so :)

Seems like the tests passed, though, the browser timed out.

@pam retry

pam:

Commit: edaf768d551e76732b71e6f4d0d7d3e71f773f68
Build details: https://travis-ci.org/pam/video.js/builds/85339423

pam:

Commit: edaf768d551e76732b71e6f4d0d7d3e71f773f68
Build details: https://travis-ci.org/pam/video.js/builds/85339423

pam:

Commit: edaf768d551e76732b71e6f4d0d7d3e71f773f68
Build details: https://travis-ci.org/pam/video.js/builds/85339423

pam:

Commit: 2f6623f976eeb5c4b0461a7c0db68010a8216a7e
Build details: https://travis-ci.org/pam/video.js/builds/87495464

src/css/components/_modal-dialog.scss
```diff
@@ -0,0 +1,33 @@
...
+
+.vjs-modal-dialog .vjs-modal-dialog-content {
+  @extend %fill-parent;
+  z-index: 1;
+}
+
+// Ultimately, this would be moved to its own component partial, but
+// for now it's unclear what styles would be generic and which specific
+// to this context!
+.vjs-modal-dialog .vjs-close {
```
gkatsev: can we move the close button styles into its own thing? Also, inherit from the close-button font icon thing as appropriate?

src/js/error-display.js
```diff
@@ -2,53 +2,58 @@
...
-    var el = super.createEl('div', {
-      className: 'vjs-error-display'
-    });
-
-    this.contentEl_ = Dom.createEl('div');
-    el.appendChild(this.contentEl_);
-
-    return el;
+  buildCSSClass() {
+    return `vjs-error-display ${super.buildCSSClass()}`;
```
gkatsev: completely not relevant to this PR but wouldn't it be awesome if Component did the \``class ${super.buildCSSClass}$`\` for us if we just returned `'class'` from this function?

src/js/modal-dialog.js
gkatsev: That's something we should verify. @heff @rcrooks Is there documentation about how the docs get built? [videojs/docs](/videojs/docs) is a bit sparse.

gkatsev: Related to the docs stuff, we should look into [esdoc](https://esdoc.org/) http://jonathancreamer.com/document-es6-with-esdoc/

src/css/components/_modal-dialog.scss
misteroneill: :+1:

src/js/modal-dialog.js
```diff
@@ -0,0 +1,379 @@
...
+   * Opens the modal.
+   *
+   * @method open
+   * @return {ModalDialog}
+   */
+  open() {
+    if (!this.opened_) {
+      let player = this.player();
+
+      this.trigger('beforemodalopen');
```
gkatsev: we should figure out how to document events.

src/js/error-display.js
misteroneill: Possibly. My concern with that sort of implicit behavior is that it's hard to override if you _don't_ want it.

src/js/modal-dialog.js
```diff
@@ -0,0 +1,379 @@
...
+
+      if (this.wasPlaying_) {
+        player.pause();
+      }
+
+      if (this.closeable()) {
+        Events.on(document, 'keydown', Fn.bind(this, this.handleKeyPress));
+      }
+
+      player.controls(false);
```
gkatsev: Should we hide the controls? Or should we make it configurable? @dmlap?

src/js/modal-dialog.js
```diff
@@ -0,0 +1,379 @@
...
+   * @method opened
+   * @param  {Boolean} [value]
+   *         If given as a Boolean, it will open (`true`) or close (`false`)
+   *         the modal.
+   *
+   * @return {Boolean|ModalDialog}
+   *         Returns `this` only if `value` was passed.
+   */
+  opened(value) {
+    if (typeof value !== 'undefined') {
```
gkatsev: so, `opened(null)` will open the modal. Is that expected?

src/js/modal-dialog.js
```diff
@@ -0,0 +1,379 @@
...
+      // If the player was playing, pause it and take note of its previously
+      // playing state.
+      this.wasPlaying_ = !player.paused();
+
+      if (this.wasPlaying_) {
+        player.pause();
+      }
+
+      if (this.closeable()) {
+        Events.on(document, 'keydown', Fn.bind(this, this.handleKeyPress));
```
gkatsev: I think you can also use `this.on(document, ...)` which should then handle the removal of the handler when this component gets disposed.

src/js/error-display.js
misteroneill: Could be a separate method.
```js
mixinCSSClass() {
  return 'vjs-error-display'; // gets combined with superclass classNames
}

buildCSSClass() {
  return 'vjs-error-display'; // overrides, identical behavior to today
}
```
Of course, it doesn't look like `buildCSSClass` gets implicitly called anywhere (like in `Component#createEl`), so maybe it would make sense to auto-inherit like that.

src/js/player.js
```diff
@@ -2480,6 +2481,38 @@ class Player extends Component {
...
+    options.content = content || '';
+
+    let modal = new ModalDialog(player, options);
+
+    player.addChild(modal);
+    modal.on('dispose', function() {
+      player.removeChild(modal);
+    });
+
+    return modal.open();
```
gkatsev: should we open the modal here or just return it for the user to open?

test/index.html
```diff
@@ -10,7 +10,7 @@
   <div id="qunit"></div>
   <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
 
-  <script src="../build/temp/ie8/videojs-ie8.min.js"></script>
+  <script src="../build/temp/ie8/videojs-ie8.js"></script>
```
gkatsev: we probably should wrap this in the IE8 conditional, but that isn't necessary for this Pr.

src/js/error-display.js
gkatsev: Yeah, maybe a separate method. Anyway, completely separate from this PR :)

src/css/components/_modal-dialog.scss
misteroneill: Hmm. Looks like the cancel icon is not released in videojs/font yet. It is [available in master](https://github.com/videojs/font/commit/182aa4c2ee46cca7ca00155c4cdbd89acbad5a65#diff-3709d0d9b83e562f3bfdb9483dd92c9f), though.

src/css/_utilities.scss
```diff
@@ -94,11 +96,15 @@
   order: $value;
```
gkatsev: any ideas why github could be complaining about this property name?

src/js/modal-dialog.js
misteroneill: Good question. Perhaps `typeof value === 'boolean'`?

src/css/components/_modal-dialog.scss
gkatsev: Ah, I can probably look into releasing that. Go with an `x` or `✕` (U+2715) for now.

src/js/close-button.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+class CloseButton extends Button {
+
+  constructor(player, options) {
+    super(player, options);
+    this.controlText(options && options.controlText || this.localize('Close'));
+  }
+
+  buildCSSClass() {
+    return `vjs-close ${super.buildCSSClass()}`;
```
gkatsev: while the close button should definitely be a button I'm a bit concerned with the fact that it gets the `vjs-control` class set on it. It currently implies a control bar control and possibly used for that places. Possibly not really an issue but wanted to call it out here.

test/unit/modal-dialog.test.js
```diff
@@ -0,0 +1,441 @@
...
+    'hidden'
+  ];
+
+  assert.expect(4 + classes.length);
+  assert.strictEqual(this.el.tagName.toLowerCase(), 'div', 'el is a <div>');
+  assert.strictEqual(this.el.tabIndex, -1, 'el has -1 tabindex');
+  assert.strictEqual(this.el.getAttribute('aria-role'), 'dialog', 'el has aria-role="dialog"');
+  assert.strictEqual(this.el.getAttribute('aria-label'), '', 'el has aria-role="" by default');
+
+  classes.forEach(function(s) {
```
gkatsev: almost seems like this should be turned into a helper function.

src/js/player.js
gkatsev: I guess it makes sense for it to open by default.

src/css/_utilities.scss
misteroneill: Nope, I noticed that. It's standard, though: https://drafts.csswg.org/css-flexbox-1/#order-property

src/js/modal-dialog.js
gkatsev: Maybe just check for truthyness?

src/js/player.js
misteroneill: Yeah, this came out of that big discussion we had a few weeks ago about this feature. The intent of this `createModal` method is to provide the most direct "I just want a modal right now" implementation.

src/js/modal-dialog.js
misteroneill: Except we don't necessarily want to rely on disposal. The modal shouldn't be listening to keydown if it's closed (but still in the DOM). That said, I still think `this.on` is better here than `import`ing `Events`.

src/js/modal-dialog.js
misteroneill: Well, this is following the get/set overloaded method pattern; so, testing for truthiness won't work (because `false` should set the internal value).

I think it should either check that it's not `undefined` here or that it is a boolean.

Or rethink this API approach. I'm not married to it by any means.

test/index.html
misteroneill: Maybe I'll put together a separate PR wrapping all references to the IE8 script with a conditional comment.

src/js/modal-dialog.js
misteroneill: As discussed, fixed to check for bool explicitly.

package.json
```diff
@@ -28,7 +28,7 @@
     "object.assign": "^4.0.1",
     "safe-json-parse": "^4.0.0",
     "tsml": "1.0.1",
-    "videojs-font": "1.3.0",
+    "videojs-font": "git+https://github.com/videojs/font.git#182aa4c2ee46cca7ca00155c4cdbd89acbad5a65",
```
gkatsev: I released 1.4.0 on npm. Check it out. https://github.com/videojs/font/commit/57161e1988df1b4ac86d527b59382db08050b2d9
If it works for you, I'll release it to the CDN tomorrow.

OwenEdwards:
A couple of points about adding a modal dialog in video.js:

1. Will the Captions Settings dialog move over to being a subclass of this modal dialog?

As I'm looking at accessibility of the menu controls, I'm realizing that I need to file an issue about the Captions Settings dialog not being fully accessible, and this might be a good way to address that; which leads nicely to:

2. Accessibility - modal dialogs have a number of requirements, best laid out in:

[Modal Dialog Widget - WAI-ARIA 1.0 Authoring Practices](http://www.w3.org/TR/wai-aria-practices/#dialog_modal)

and:

[Using the dialog role - Accessibility | MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role)

Being able to dismiss the dialog with the Esc key is a great start. Moving, trapping and returning keyboard focus are also going to be needed, as well as labeling/ titling.

How can I help make this happen? Would it make sense to let this PR complete and then file a separate issue and PR?

gkatsev:
@OwenEdwards 

1. The Captions settings dialog is kind of in a weird state because how how it's created and managed. Particularly since it kind of needs to be not fully modal (the currently displayed captions need to be showing for FCC compliance as a live example of the settings). Perhaps we could add the needs of the captions setting dialog into the modal. This probably should be a separate PR because we need to get this PR in soon.
2. We definitely should make the modal dialog as accessible as we can. We'll see if there's easy stuff we can figure out ourselves and and add it into this PR but otherwise, separate PRs would be awesome.

This comment is already helping! But issues and PRs and direction and testing once this PR lands would be amazing and most welcome. Thanks!

OwenEdwards:
@gkatsev see #2746 for some more specific points to think about in terms of modal dialog accessibility. Let's discuss using something from this modal dialog work for the Captions Settings dialog on that issue thread?

I'm happy to help!

package.json
misteroneill: :+1:

misteroneill:
I've got a number of tasks and small improvements to make after some discussion with @gkatsev, but I'll add some accessibility improvements here as well (thanks @OwenEdwards).

### Improvements
- [ ] As was originally the case, don't allow HTML strings as content for modals. Strings should be allowed, but only via `createTextNode`/`textContent`.
- [ ] Abstract `normalizeContent_` method and make it recursive (once) - this could be useful in multiple places.

### Accessibility
- [ ] Set `role="document"` on modal content.
- [ ] Use `aria-hidden` attributes appropriately.
- [ ] Capture focus within modal when it is open.
- [ ] Add a modal description option.

gkatsev:
That sounds reasonable to do for this PR and then we can add whatever else is needed afterwards.

misteroneill:
@gkatsev Alright, this is ready for further review - particularly the last 4 commits. Note particularly the changes to `utils/dom`, which is where I landed with the content normalization. This should come in useful for all sorts of XSS-conscious content injection.

gkatsev:
@pam retry

src/css/components/_modal-dialog.scss
```diff
@@ -0,0 +1,13 @@
...
+  @extend %fill-parent;
+  @include linear-gradient(180deg, rgba(0, 0, 0, 0.8), rgba(255, 255, 255, 0));
+}
+
+.vjs-modal-dialog .vjs-modal-dialog-content {
+  @extend %fill-parent;
+
+  font-size: 1.2em; // 12px
+  line-height: 1.5; // 18px
+  padding: 30px;
```
gkatsev: should probably be in `em` as well.

src/js/utils/dom.js
```diff
@@ -369,3 +385,115 @@ export function getPointerPosition(el, event) {
...
+    }
+
+    if (isEl(value) || isTextNode(value)) {
+      return value;
+    }
+
+    if (typeof value === 'string' && /\S/.test(value)) {
+      return document.createTextNode(value);
+    }
+  }).filter(value => value);
```
gkatsev: Why the filter here?

src/css/components/_modal-dialog.scss
misteroneill: I thought the same thing originally, but increasing the padding as the font-size increases looks bad.

src/js/utils/dom.js
misteroneill: The `map` will only return nodes, so this removes things that cannot be injected as child nodes.

src/css/components/_modal-dialog.scss
gkatsev: Makes sense. Also, `rem` is even less supported than `em`.

gkatsev:
Makes sense to fix the focus stuff later. I guess what we would want is to capture it only if the focus is already inside the player area but not otherwise? Something to think about.

src/js/modal-dialog.js
```diff
@@ -0,0 +1,372 @@
...
+        player.play();
+      }
+
+      if (this.closeable()) {
+        this.off(document, 'keydown', Fn.bind(this, this.handleKeyPress));
+      }
+
+      player.controls(true);
+      this.hide();
+      this.el().setAttribute('aria-hidden', 'true');
```
gkatsev: I was wondering about what value this attribute should get since boolean attributes in HTML are checked via presence rather than value so it's customary to set their value to their name but seems like ARIA specifically looks for the value `true`.
Just wanted to call that out for others that care :)

gkatsev:
I wonder why pam hasn't run on this after the rebase...

misteroneill:
I think she gave up. :smiling_imp:

src/css/components/_error.scss
```diff
@@ -1,48 +1,4 @@
...
-  display: block;
-  position: absolute;
-  left: 0;
-  top: 0;
-  width: 100%;
-  height: 100%;
-}
-
-.vjs-error .vjs-error-display:before {
-  content: 'X';
```
gkatsev: was this removed on purpose?
I would prefer to not change the look of the error display *too* much. Or at least, we still need something to make it noticeable as an error display and not just an uncloseable modal dialog.

misteroneill: Yeah. It looked totally broken with that style - it duplicates a lot of the style of the modal dialog.

I don't know what the "X" was about (I was guessing a "close" icon/button).

gkatsev: I think just to signify an error.

misteroneill: Ah, then I'll bring that back.

gkatsev:
LGTM

src/js/modal-dialog.js
OwenEdwards: Good point @gkatsev!

ARIA doesn't follow HTML syntax, in fact it's a language with its own syntax which can be used within HTML (but also within other languages). The fact that HTML booleans are either false through their absence or true through presence which can be denoted as, e.g. `selected="selected"` is [an SGML vs XML thing](http://stackoverflow.com/questions/7089584/html-why-boolean-attributes-do-not-have-boolean-value), whereas ARIA values may be [booleans _or_ true, false or undefined](http://www.w3.org/TR/wai-aria/states_and_properties#propcharacteristic_value) (as well as true/false/mixed!), so the syntax was defined differently.

Setting aria-* attributes otherwise is a syntax violation, but it's a tricky one to catch since HTML validators generally look at static pages rather than what the JavaScript is doing.

sethborg:
I know this isn't the place for this but there's no where else to turn. Would someone mind posting some example code of how to use this new function? I can't find a single example on the internet.

misteroneill:
@sethborg No worries. There are [ModalDialog API docs](http://docs.videojs.com/docs/api/modal-dialog.html), but I agree that a more prose-based document would be helpful. I'll add an issue to create one. The absolute easiest way to create a modal is the `createModal` method of a `Player`:

```js
// Get a reference to your player.
var player = videojs.getPlayers().yourPlayerId;

player.createModal('Hello, world!');
```

Note that if you want to populate it with HTML - and you probably do - you can pass a DOM element reference or a function that creates one:

```
player.createModal(function () {
  var content = document.createElement('h1');
  content.textContent = 'Hello, world!';
  return content;
});
```

Beyond this, the `ModalDialog` component exists as something that can be used to create more complex modals.

sethborg:
Oh this creates a modal dialog on the video player area. I was under the impression this allowed me to create a lightbox over the whole webpage with the video player in it. The API should clear that up.

PR #2680, opened 2015-10-06
Players know which plugins they're using

misteroneill:
It seems like it would be useful for players to know which plugins they're using.

My only concern with this approach is that the `Player.prototype[name]` function is no longer a direct reference to the `init` function passed to `videojs.plugin`. This required one assertion to be changed. This may or may not qualify as a backward incompatibility; seems debatable.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -152,6 +152,9 @@ class Player extends Component {
     // an infinite loop.
     let playerOptionsCopy = mergeOptions(this.options_);
 
+    // Create an object to track which plugins are active on a player
+    this.plugins_ = {};
+
     // Load plugins
     if (options.plugins) {
       let plugins = options.plugins;
@@ -2477,6 +2480,16 @@ class Player extends Component {
     }
 
     return options;
+  }
+
+  /**
+   * Whether or not this player is using a given plugin.
+   *
+   * @param  {String} name The name of a plugin
+   * @return {Boolean}
+   */
+  usingPlugin(name) {
+    return !!this.plugins_[name];
   }
 
   /**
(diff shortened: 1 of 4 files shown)


pam:

Commit: 465203e8fb5151954a5c8dd27b0b0c2b317be6ee
Build details: https://travis-ci.org/pam/video.js/builds/83971058

gkatsev:
Yeah, I'd be worried that this would cause an issue. However, we could probably release it as a minor (this since adds a feature) and things break, we'll revert it in a patch.

misteroneill:
Yeah, it's additive, but if anyone is depending on this behavior:

```js
videojs.plugin('myplugin', function myplugin() {});
videojs.getComponent('Player').prototype.myplugin === myplugin;
```

That would break. If there's any concern, I won't be in the least bit offended if we close this PR and I delete my branch. It can wait for 6.0 if need be. :)

chemoish:
+1, for some type of implementation

thijstriemstra:
Have you tested it with any plugins out there?

pam:

Commit: f2d4338008f7b29367889a5af7500d141b63e624
Build details: https://travis-ci.org/pam/video.js/builds/87495444

gkatsev:
I'm still not certain that I would be ok with pushing this out in a minor release.
We could probably update it to just have a list of plugins that are available as a first step and maybe do this change for whenever the next major release happens?

misteroneill:
Yeah, I tend to agree. A list of plugins won't tell us which plugins have been applied to a player, but I could definitely see that being useful. You're right that it's a borderline-major change.

misteroneill:
Going to close this. It's easy enough to bring into a future release.

For the time being, if you want this functionality in your usage of video.js, you could pretty easily monkey-patch this into your website/app.

PR #2692, opened 2015-10-09
Wait until the tech's element has been added to the DOM before calling emulateTextTracks

imbcmdth:
This is the same change in #2690 rebased onto the stable branch.


heff:
Merged manually

Issue #2193, opened 2015-05-22
Captions on bottom on Chrome 43


TearsDontFalls:
Since a few days and a chrome auto update, the **captions** of all Video.js players are **always at bottom:0**

Tested+affected: Windows 7 + Chrome 43
Tested+unaffected: Chrome 42 on Android, Windows 7 + Firefox 38

Example Screenshot of http://www.videojs.com/ :
![image](https://cloud.githubusercontent.com/assets/1379201/7774936/2304f9f2-00af-11e5-8de3-9646ed29c4c5.png)

heff:
Hmm, that's annoying. Did the shadow dom class names change? I was pretty
sure we had CSS pushing the captions up, but I'm not finding it.

TearsDontFalls:
On Chrome its using native tracks(so also no caption settings), so something is bugging there i think.

Found a quick workaround for now: Set these settings to false https://github.com/videojs/video.js/blob/5d550ffada237e937cf31ef344f3719112163ea6/src/js/tech/tech.js#L44 and it works as it isnt using native track support.

heff:
Yeah, you can disable native captions, but long term I think we want to rely on Chrome's native captions because people can set the captions settings once and use them across sites. I thought we had CSS that pushed the captions up using the shadow dom.

ksheurs:
Not sure if this is recommended, but this has been working for us:

```css
video::-webkit-media-text-track-container {
  bottom: 0.75em;
}
```

mmcc:
Yep, seems reasonable to me, just kinda stinks that one of the only references I could find to this shadow selector was [Sam Dutton's website](http://www.samdutton.com/track/shadowDOM.html).

gkatsev:
If you enable the shadow DOM in Chrome dev tools, you can see the selectors you could use.

mmcc:
Someone want to put together a PR for this? @ksheurs?

EZWrighter:
This affects Safari as well: https://github.com/videojs/video.js/issues/1904

gkatsev:
It seems like `::-webkit-media-text-track-container` on safari is doing weird stuff. Like, the caption shows up in the appropriate place (say `2em` up) but then moves up another `2em`s...

chemoish:
+1

PR #2702, opened 2015-10-15
Fix captions sticking to bottom for webkit browsers.

chemoish:
This PR is to start a conversation about addressing https://github.com/videojs/video.js/issues/2193.

screen shot 2015-10-14 at 7 15 54 pm

screen shot 2015-10-14 at 7 57 50 pm

Native (Chrome):

screen shot 2015-10-14 at 7 17 16 pm

screen shot 2015-10-14 at 7 17 02 pm

Resources:
- http://www.samdutton.com/track/shadowDOM.html
- http://advprog.blogspot.co.uk/2013/07/styling-html-media-inner-workings.html
- Did not check webkit/videojs implementation yet

Browsers:
- Tested in OSX Chrome, Safari (Works), FF, IE (Not so much—to varying degrees)

Compatibility:
- http://caniuse.com/#feat=transforms2d

---

Evaluations:

1. `px`, `rem`, `em`?
1. `bottom`, `top`, `translateY`?
1. `-webkit-media-text-track-container`, `-webkit-media-text-track-display`?

---

#### 1. px

`px` must be used to keep interface consistency. However, this doesn't follow the em standard that is currently used in the player.

```css
/* will make controls larger due to em usage */
.video-js
{
  font-size: 20px;
}
```

The problem with using `em` is that the *browser?* will dynamically modify the `font-size` on the `-webkit-media-text-track-container`, based on player's width.  The larger the width, the larger the `font-size`, the larger the `bottom` offset will be.

```css
/* will look fine on large screens, but poor/poorer on smaller screens */
video::-webkit-media-text-track-container {
  bottom: 0.75em;
}
```

As mentioned https://github.com/videojs/video.js/issues/2193#issuecomment-105519643.

#### 2. & 3. translateY

`bottom` can only be used on the `-webkit-media-text-track-container` since it conflicts with styles found in `-webkit-media-text-track-display` [WebVTT implementation of `line` (WebVTT) modifies `top` (css)]. However, this will cause placement issues if you want to utilize `line` (WebVTT), as they both will stack/conflict.

`translateY` seems to be the best option. `translateY` gives the most predictable offset and it doesn't conflict with any other embedded styles.

The best part about it is that, IF, `line` (WebVTT) is configured, it will override any `translateY` styles set on `-webkit-media-text-track-display`.

```js
WEBVTT

00:00.700 --> 00:04.110
Captions describe all relevant audio for the hearing impaired.
[ Heroic music playing for a seagull ]
```

```css
/* will offset the text 45px from the bottom of the player */
video::-webkit-media-text-track-display {
    transform: transateY(-45px);
}
```

```js
WEBVTT

00:00.700 --> 00:04.110 line:50%
Captions describe all relevant audio for the hearing impaired.
[ Heroic music playing for a seagull ]
```

```css
/* will be ignored and overridden by `-webkit-media-text-track-display`'s `top`' to force the captions to be vertically centered */
video::-webkit-media-text-track-display {
    transform: transateY(-45px);
}
```

This will allow users to benefit from both.

---

Other stuff:

- Positioning captions based on user activity or inactivity is possible (Control bar visibility).
- No idea where the CSS belongs.
- `@mixin` can be removed if `npm i compass-mixins` was favored.

### DIFF
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -1,3 +1,4 @@
+/* Emulated tracks */
 .vjs-text-track-display {
   position: absolute;
   bottom: 3em;
@@ -25,6 +26,12 @@
 .vjs-captions { color: #fc6 /* Captions are yellow */; }
 .vjs-tt-cue { display: block; }
 
+/* Native tracks */
 video::-webkit-media-text-track-display {
-  @include transform(translateY(-45px));
+  @include transform(translateY(-3em));
+}
+
+/* Move captions down when controls aren't being shown */
+.video-js.vjs-user-inactive.vjs-playing video::-webkit-media-text-track-display {
+  @include transform(translateY(-1.5em));
 }
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -24,3 +24,7 @@
 .vjs-subtitles { color: #fff /* Subtitles are white */; }
 .vjs-captions { color: #fc6 /* Captions are yellow */; }
 .vjs-tt-cue { display: block; }
+
+video::-webkit-media-text-track-display {
+  @include transform(translateY(-45px));
+}
--- a/src/css/_utilities.scss
+++ b/src/css/_utilities.scss
@@ -3,6 +3,14 @@
   background-color: rgba($color, $alpha);
 }
 
+@mixin transform($transform) {
+  -moz-transform: $transform;
+  -ms-transform: $transform;
+  -o-transform: $transform;
+  -webkit-transform: $transform;
+  transform: $transform;
+}
+
 @mixin transition($string: $transition--default) {
   -webkit-transition: $string;
   -moz-transition: $string;


gkatsev:
Thanks for the PR and the detailed description!
The last time I tested this in safari, safari did weird stuff with this by first rendering the result in the correct place but then jumping it up by that same amount. It was very weird. I wonder whether using translate might not have this issue or maybe using pixel values fixes it but that is definitely something that should be looked into before merging.

chemoish:
I believe this is the case with modifying the container with bottom.
Translate did not have the double repaint issue.

I will thoroughly test tomorrow.

chemoish:
Safari has a double repaint issue with modifying the position of the `-webkit-media-text-track-container`.

`translateY` has no issues when placed on either `-webkit-media-text-track-container` or `-webkit-media-text-track-display` (Though benefits come with putting it on the display).

Safari does have a intermittent (rare) issue initializing/disposing captions, where the font-size quadruples* and the caption balloons, but seems more apparent with `bottom`.

src/css/components/_text-track.scss
```diff
@@ -24,3 +24,7 @@
 .vjs-subtitles { color: #fff /* Subtitles are white */; }
 .vjs-captions { color: #fc6 /* Captions are yellow */; }
 .vjs-tt-cue { display: block; }
+
+video::-webkit-media-text-track-display {
+  @include transform(translateY(-45px));
```
gkatsev: This should probably be `4em` rather than `45px`. `4.5em` is a bit much, I think.
Also, can we add this value as a variable that defaults to `4em` to the [vars](https://github.com/videojs/video.js/blob/master/src/css/_variables.scss) file? That way, a user could more easily change where the captions are if they have a custom control bar and are including the sass files.

gkatsev:
Also, thanks for this PR. Glad you were able to figure out how to get safari to behave (which I also verified). It looks good to me, but had a comment.

src/css/components/_text-track.scss
gkatsev: Also, we should make this align with where we stick the emulated captions.

chemoish: Will do. I'll retest `em`, but expect to have the captions bleed into the video depending on the `font-size` set by the browser width.

Also, should the captions work like native `-webkit`, where they adjust for the control bar, or should they stay at a fixed (relative) position.

> Also, we should make this align with where we stick the emulated captions.

Don't follow.

gkatsev: We should make the emulated captions and native captions work as close to each other as possible. The emulated captions already are aligned to leave room for the control bar, so, we should make sure that the native captions are similar. If the emulated captions go up/down with the control bar, we should do the same for the native captions as well.
I tried using `em` locally and it didn't actually change. But that is kind of what we want because the control bar resizes itself based on the font size :)

chemoish:
I initially didn't follow the whole emulated notion, but I figured it out ><.

Added some comments to make it more explicit.

So its *closer*, but won't be perfect. Emulated tracks modify the font size of the `.vjs-text-track-display` element, a child element. This causes the positioning to be based on the the font-size of the `.video-js` element.

Native tracks have their positioning and font-size tied to the same element, so the positioning is much more noticeable and, of course, varies depending on the width of the viewport.

---

Control bar (Chrome 3em/1.5em, FF 3em/1em):

screen shot 2015-10-28 at 5 50 59 pm

No control bar (Chrome 3em/1.5em, FF 3em/1em):

screen shot 2015-10-28 at 5 51 36 pm

---

`^` is the reason why its such a different experience on huge viewports, but looks pretty similar on small ones.

gkatsev:
Awesome. I'll try it out tomorrow and probably merge it in.

gkatsev:
LGTM

dmlap:
Awesome! Thanks again, @chemoish!

PR #2708, opened 2015-10-16
Use all installed browsers for unit tests

forbesjo:
This PR uses the karma-detect-browsers to launch all available browsers when using the karma:defaults task on a local environment.

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -133,7 +133,7 @@ module.exports = function(grunt) {
 
       defaults: {
         detectBrowsers: {
-          enabled: true,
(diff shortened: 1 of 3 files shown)


gkatsev:
Does this happen only for local tests or for browserstack as well?

forbesjo:
This should only apply to local tests using the `karma:default` task. If the tests are being run in Travis it will disable detectBrowsers and fall back to the browsers array which is Firefox on PRs and BrowserStack on Pam/non-PRs.

gkatsev:
Cool, thanks!

gkatsev:
LGTM

PR #2723, opened 2015-10-21
'contrib release' should only pull changes from 'upstream' and not push things to 'origin'

gkatsev:


contrib.json
```diff
@@ -54,19 +54,44 @@
         [ "npm install", "Download dependencies"]
       ]
     },
+    "upstream": {
```
heff: It feels off to have this as `contrib update upstream master` since we're not 'updating' upstream master, we're fetching/pulling/getting upstream master and updating local master vs remote master. Should we have `update local master` and `update remote master`?

It's also just two commands if we want to just not have a separate task for it.

gkatsev: We can. So, would `update local` just to the `pull` and `update remote` also do the `push`?

heff: Yeah, that makes sense to me.

gkatsev: Updated.

heff:
lgtm

PR #2754, opened 2015-10-28
Expose DOM Helpers

misteroneill:
Building on the `ModalDialog` work...

- Adds a couple small DOM convenience methods - `$` and `$$` (essentially, `querySelector` and `querySelectorAll` respectively).
- Adds a `toggleElClass` function.
- Exposes a bunch of useful DOM methods on the `videojs` function.
- Rewrites the `className` manipulation functions to use `classList` if it's available (which is true of all supported browsers except IE8).

_Note:_ The methods are generally exposed without the `El` part (i.e. `Dom.hasElClass` is exposed as `videojs.hasClass`). The reason for this is parity with popular JS libraries (jQuery, etc) that people are familiar with.

### DIFF
--- a/src/js/tracks/text-track-list-converter.js
+++ b/src/js/tracks/text-track-list-converter.js
@@ -44,6 +44,8 @@ let trackToJson_ = function(track) {
  * @function textTracksToJson
  */
 let textTracksToJson = function(tech) {
+
+  // Cannot use $$ here because it is not an instance of Tech
   let trackEls = tech.el().querySelectorAll('track');
 
   let trackObjs = Array.prototype.map.call(trackEls, (t) => t.track);
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -627,6 +627,20 @@ videojs.addClass = Dom.addElClass;
 videojs.removeClass = Dom.removeElClass;
 
 /**
+ * Adds or removes a CSS class name on an element depending on an optional
+ * condition or the presence/absence of the class name.
+ *
+ * @method toggleElClass
+ * @param  {Element} element
+ * @param  {String} classToToggle
+ * @param  {Boolean|Function} [predicate]
+ *         Can be a function that returns a Boolean. If `true`, the class
+ *         will be added; if `false`, the class will be removed. If not
+ *         given, the class will be added if not present and vice versa.
+ */
+videojs.toggleClass = Dom.toggleElClass;
+
+/**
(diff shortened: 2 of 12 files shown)


src/js/tracks/text-track-list-converter.js
```diff
@@ -44,6 +44,8 @@ let trackToJson_ = function(track) {
  * @function textTracksToJson
  */
 let textTracksToJson = function(tech) {
+
+  // Cannot use $$ here because it is not an instance of Tech
   let trackEls = tech.el().querySelectorAll('track');
```
gkatsev: isn't `tech` here a Tech and thus it should have `$$`?

misteroneill: It wasn't when I was working on this PR (hence the comment).

src/js/video.js
```diff
@@ -501,22 +501,185 @@ videojs.xhr = xhr;
...
+ *
+ * @param  {Element|String} [context=document]
+ *         A DOM element within which to query. Can also be a selector
+ *         string in which case the first matching element will be used
+ *         as context. If missing (or no element matches selector), falls
+ *         back to `document`.
+ *
+ * @return {Element|null}
+ */
+videojs.$ = Dom.$;
```
gkatsev: I'm worried that exposing `$` and `$$` will make people assume that we either have jquery or are jquery compatible. We also don't want people do be reliant on videojs for that kind of functionality.
I think the other functions are totally fine, though.

misteroneill: That's fair and makes a ton of sense. To be clear: do you want to keep them internally and on components or ditch them entirely? I'm fine with either - they are the least useful part of this PR, I think.

gkatsev: Keeping them internal is fine, since doing `.el().querySelector` seems very common.
So, just remove them externally.

misteroneill: And keeping them on `Component.prototype`, right? The only concern there is they are sort-of-exposed (e.g. `player.$('selector')`), I guess.

gkatsev: Yeah, I guess them leaking that way isn't completely terrible because it'll mostly be used for vjs-related things and not just people using `videojs.$` for other stuff that they would just use jquery for.

gkatsev:
LGTM. Going to make another PR after this to fix the tracks JSON that I commented about above.

PR #2775, opened 2015-11-04
Bower Path Fixes #2740

### LINKED ISSUE #2740: bower.json - The files listed under the main key are wrong
I tried installing the component using the Main Bower gulp plugin in my project.  This plugin is supposed to copy only the files listed in the main key that are the only files needed to run the module.  Well right now the listed files are the ones in src however it should have a list of the dist files.

DatTran:


bower.json
```diff
@@ -2,8 +2,8 @@
   "name": "video.js",
   "description": "An HTML5 and Flash video player with a common API and skin for both.",
   "main": [
-    "src/js/video.js",
-    "src/css/video-js.scss"
+    "dist/video-js.js",
```
gkatsev: this should still be `dist/video.js`.

DatTran:
Updated bower path fix

PR #2791, opened 2015-11-07
Update .travis.yml

paladox:
Test against node_js version 4 too.


gkatsev:
Thanks for the PR but I think I'd prefer to leave the build as is for just a bit longer until we get npm3 support and what not.

PR #2821, opened 2015-11-17
Do not display negative time

forbesjo:
Sometimes at the end of an HLS video the remaining time is calculated as -1 (current time is greater than duration by a hundredth of a second, rounded to 1 second) causing the remaining time display to show "--1:0-1"

This PR sets negative seconds to 0 so that only a valid time or dashed time is displayed.

### DIFF
--- a/src/js/utils/format-time.js
+++ b/src/js/utils/format-time.js
@@ -12,6 +12,7 @@
  * @function formatTime
  */
 function formatTime(seconds, guide=seconds) {
+  seconds = seconds < 0 ? 0 : seconds;
   let s = Math.floor(seconds % 60);
   let m = Math.floor(seconds / 60 % 60);
   let h = Math.floor(seconds / 3600);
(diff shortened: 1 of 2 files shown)


gkatsev:
Makes sense to me. LGTM

pam:

Commit: 1556bfb6662408ef4a321b759dc8736b045281fc
Build details: https://travis-ci.org/pam/video.js/builds/91626134

gkatsev:
I just realized that remaining time uses this as well: https://github.com/videojs/video.js/blob/8d9695b6adb352ace30fd95e25e61cbae6520e83/src/js/control-bar/time-controls/remaining-time-display.js#L56
We probably want a 3rd parameter to the function that determines whether it clips at 0 or not that we can use in duration and current time but it defaults to current functionality so it also isn't a breaking change.

forbesjo:
Remaining time is is a positive timestamp (duration - currentTime) with a '-' appended. I don't think formatTime is ever given a negative number because it would never return a correctly formatted timestamp.

gkatsev:
oh, I see. Missed that.

PR #2827, opened 2015-11-19
Replace multi-line comment with single line comments

nick11703:
When using Compass to compile the sass files, mulit-line comments are
not stripped in the output, causing malformed CSS output.

Example
```css
.video-js {
  /* display:inline-block would be closer to the video el's display:inline
   * but it results in flash reloading when going into fullscreen [#2205]
   */
  display: block;
  /* Make video.js videos align top when next to video elements */
  vertical-align: top;
  box-sizing: border-box;
  color: #fff;
  background-color: #000;
  position: relative;
  padding: 0;
  /* Start with 10px for base font size so other dimensions can be em based and
  easily calculable. */
  font-size: 10px;
  line-height: 1;
  /* Provide some basic defaults for fonts */
  font-weight: normal;
  font-style: normal;
  /* Avoiding helvetica: issue #376 */
  font-family: Arial, Helvetica, sans-serif;
  -webkit-user-select: none;
  -moz-user-select: none;
  -ms-user-select: none;
  user-select: none;
  /* Fix for Firefox 9 fullscreen (only if it is enabled). Not needed when
  checking fullScreenEnabled. */
}
```

Chrome's interpretation of the CSS
screen shot 2015-11-19 at 12 47 58 pm

### DIFF
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -1,16 +1,16 @@
-/**
- * Let's talk pixel math!
- * Start with a base font size of 10px (assuming that hasn't changed)
- * No Hover:
- * - Progress holder is 3px
- * - Progress handle is 9px
- * - Progress handle is pulled up 3px to center it.
- *
- * Hover:
- * - Progress holder becomes 5px
- * - Progress handle becomes 15px
- * - Progress handle is pulled up 5px to center it
- */
+//
+// Let's talk pixel math!
+// Start with a base font size of 10px (assuming that hasn't changed)
+// No Hover:
+// - Progress holder is 3px
+// - Progress handle is 9px
+// - Progress handle is pulled up 3px to center it.
+//
+// Hover:
+// - Progress holder becomes 5px
+// - Progress handle becomes 15px
+// - Progress handle is pulled up 5px to center it
+//
 
 .video-js .vjs-progress-control {
   @include flex(auto);
@@ -21,30 +21,30 @@
   display: none;
 }
 
-/* Box containing play and load progresses. Also acts as seek scrubber. */
+// Box containing play and load progresses. Also acts as seek scrubber.
 .video-js .vjs-progress-holder {
   @include flex(auto);
   @include transition(all 0.2s);
   height: 0.3em;
 }
 
-/* We need an increased hit area on hover */
+// We need an increased hit area on hover
 .video-js .vjs-progress-control:hover .vjs-progress-holder {
   font-size: 1.666666666666666666em
 }
 
-/* Also show the current time tooltip */
+// Also show the current time tooltip
 .video-js .vjs-progress-control:hover .vjs-mouse-display:after,
 .video-js .vjs-progress-control:hover .vjs-play-progress:after {
   display: block;
 
-  /* If we let the font size grow as much as everything else, the current time tooltip ends up
-     ginormous. If you'd like to enable the current time tooltip all the time, this should be disabled
-     to avoid a weird hitch when you roll off the hover. */
+  // If we let the font size grow as much as everything else, the current time tooltip ends up
+  // ginormous. If you'd like to enable the current time tooltip all the time, this should be disabled
+  // to avoid a weird hitch when you roll off the hover.
   font-size: 0.6em;
 }
 
-/* Progress Bars */
+// Progress Bars
 .video-js .vjs-progress-holder .vjs-play-progress,
 .video-js .vjs-progress-holder .vjs-load-progress,
 .video-js .vjs-progress-holder .vjs-load-progress div {
@@ -53,9 +53,9 @@
   height: 0.3em;
   margin: 0;
   padding: 0;
-  /* updated by javascript during playback */
+  // updated by javascript during playback
   width: 0;
-  /* Needed for IE6 *///
+  // Needed for IE6
   left: 0;
   top: 0;
 }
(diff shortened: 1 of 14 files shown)


gkatsev:
I'm okay with changing the single line comments to use `//` instead of `/* */`. I'm not a fan of changing multi-line comments to use single-line comments. Seems like this is a bug that compass should fix instead of us fixing our code.

nick11703:
Compass will ignore stripping multi-line comments and output them in the final CSS. The problem really come from [multi-line comments in CSS class definitions.](https://github.com/videojs/video.js/blob/master/src/css/components/_layout.scss#L2) Single line comments are stripped by compass during compilation. Can we agree to replace multi-line comments inside CSS definitions with single line comments? If so, I can update the pull request.

gkatsev:
Can you open an issue against compass regarding this? They should strip out comments that won't be valid in pure css. I'll pull it in then with the change for those.
Sounds good?

misteroneill:
It isn't Compass doing it, it's Sass. I don't think it's a bug in Sass, either. It may be a Chrome bug because it sees that commented out CSS property as an actual CSS property.

misteroneill:
I did a test. The following SCSS has its comment preserved, but shows no issues in the Chrome dev tools:

```scss
.video-js {
  /*
  baz baz baz baz
  bar bar bar bar
  */
  display: block;
}
```
But this shows the behavior in the attached screenshot:

```scss
.video-js {
  /*
  display: inline
  baz baz baz baz
  bar bar bar bar
  */
  display: block;
}
```

gkatsev:
Yeah, I just looked at videojs's css and the comments are preserved. Though, it's still working fine.

nick11703:
I don't believe it's a Compass issue either. Sass states it will ignore multi-line comments. I wonder if it's chrome picking up the ```display:inline-block``` in the comment.

nick11703:
Either way, if it's a Sass, Compass or browser issue, is it necessary to output the comments in the compiled CSS. They remain in the .scss files for the developer. To help make the web faster it would reduce final output by allowing single line comments to be stripped.

misteroneill:
FWIW, the comments do get stripped in `dist/video-js.min.css`, which is really what people should be using.

gkatsev:
I think I know what the exactly issue here.
Since comments aren't valid inside blocks, chrome looks at the block line by line to see whether it's valid or not and so if you have a multi-line comment, some of the things that are commended out could be interpreted incorrectly.

I looked through the changes and I think so of the things can just be moved outside of the blocks. The things that should stay inside the block can be changed to single line comments. I'll go through and make comments about it.

src/css/components/_progress.scss
```diff
@@ -21,30 +21,30 @@
...
-/* Also show the current time tooltip */
+// Also show the current time tooltip
 .video-js .vjs-progress-control:hover .vjs-mouse-display:after,
 .video-js .vjs-progress-control:hover .vjs-play-progress:after {
   display: block;
 
-  /* If we let the font size grow as much as everything else, the current time tooltip ends up
-     ginormous. If you'd like to enable the current time tooltip all the time, this should be disabled
-     to avoid a weird hitch when you roll off the hover. */
+  // If we let the font size grow as much as everything else, the current time tooltip ends up
```
gkatsev: move this out of the block

src/css/components/_progress.scss
```diff
@@ -83,7 +83,7 @@
 // Current Time "tooltip"
 .video-js .vjs-mouse-display:after,
 .video-js .vjs-play-progress:after {
-  /* By default this is hidden and only shown when hovering over the progress control */
+  // By default this is hidden and only shown when hovering over the progress control
```
gkatsev: move out of the block

src/css/components/_progress.scss
```diff
@@ -101,18 +101,18 @@
 }
 
 .video-js .vjs-load-progress {
```
gkatsev: move both the comments out of the block

src/css/components/_progress.scss
```diff
@@ -101,18 +101,18 @@
...
   background: rgba($secondary-background-color, $secondary-background-transparency);
 }
 
-/* there are child elements of the load progress bar that represent the
-specific time ranges that have been buffered */
+// there are child elements of the load progress bar that represent the
+// specific time ranges that have been buffered
 .video-js .vjs-load-progress div {
-  /* For IE8 we'll lighten the color */
+  // For IE8 we'll lighten the color
```
gkatsev: move both the comments out of the block

gkatsev:
@nick11703 if you make the changes discussed, I'll pull it in. Thanks!

nick11703:
@gkatsev No problem. I'll work on that soon. Thanks.

src/css/components/_progress.scss
nick11703: Both of these comments make sense where they are as single line ```//``` comments that will get stripped. Do you really want to move these?

gkatsev: I was thinking that this block is small and it's fairly obvious what is happening we can move it out. But yeah, I guess it's fine to leave it as is.

gkatsev:
@nick11703 hey, did you get a chance to go through and update the comments? (Looks like this also needs a rebase).

nick11703:
@gkatsev Sorry. Got caught up with the holidays. It's in my plans for today.

nick11703:
@gkatsev This should be good to go now. Let me know if you need anything else.

gkatsev:
LGTM

dmlap:
LGTM

PR #2835, opened 2015-11-21
Check for text track changes that occurred before tech was listening

gesinger:
If any code sets a track mode during player setup (e.g., `track.mode = 'showing'`) for non native text tracks, there is a race case where the player may not be listening for changes on the track. For instance:

```javascript
    html5: {
        nativeTextTracks: false
    }
});

var track = player.addTextTrack('captions');
track.mode = 'showing';
```

The caption settings will show the captions as on for the track, but the captions will not show up on the video unless a user switches to fullscreen or changes the caption settings (both will update display appropriately).

This change ensures that before videojs starts listening to change events on tracks, it gets any changes that may have happened between then and the start of setup.

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -335,6 +335,7 @@ class Tech extends Component {
       }
     });
 
+    textTracksChanges();
     tracks.addEventListener('change', textTracksChanges);
 
     this.on('dispose', function() {
--- a/test/unit/tracks/tracks.test.js
+++ b/test/unit/tracks/tracks.test.js
@@ -342,3 +342,13 @@ if (Html5.supportsNativeTextTracks()) {
     emulatedTt.on('addtrack', addtrack);
   });
 }
+
+test('should check for text track changes when emulating text tracks', function() {
+  let tech = new Tech();
+  let numTextTrackChanges = 0;
+  tech.on('texttrackchange', function() {
+    numTextTrackChanges++;
+  });
+  tech.emulateTextTracks();
+  equal(numTextTrackChanges, 1, 'we got a texttrackchange event');
+});


pam:

Commit: 8a3a97ecc936e34ca052570e4efb14ea1d8e1e45
Build details: https://travis-ci.org/pam/video.js/builds/92385457

test/unit/tracks/tracks.test.js
```diff
@@ -342,3 +342,27 @@ if (Html5.supportsNativeTextTracks()) {
...
+  player = TestHelpers.makePlayer({
+    techfaker: {
+      nativeTextTracks: false
+    }
+  });
+  equal(numTextTracksCalls, numTextTracksCallsNative + 1,
+    'we didn\'t have an extra text tracks update for non native text tracks');
+
+  Player.prototype.textTracks = origTextTracks;
+  Tech.prototype['featuresNativeTextTracks'] = false;
```
gkatsev: What if the original value was true?

gesinger: Good call. Will store the original value and reset it here.

pam:

Commit: dc174d890bef253d042e086aa8add0986afac9d9
Build details: https://travis-ci.org/pam/video.js/builds/92431612

gkatsev:
LGTM

src/js/tech/tech.js
```diff
@@ -322,6 +322,7 @@ class Tech extends Component {
     }
 
     let textTracksChanges = Fn.bind(this, function() {
+      let tracks = this.textTracks();
```
dmlap: Is this necessary? Seems like this function could not run unless `tracks` was already defined at the assignment in `emulateTextTracks` above.

gkatsev: It isn't strictly necessary but it makes testing this significantly easier.

src/js/tech/tech.js
```diff
@@ -322,6 +322,7 @@ class Tech extends Component {
...
       let updateDisplay = () => this.trigger('texttrackchange');
```
dmlap: I'm not totally sure this will be removed in the loop below since the function is created anew (I believe) every invocation of this method. That would be a pre-existing bug-- just thought it's worth mentioning.

test/unit/tracks/tracks.test.js
```diff
@@ -342,3 +342,28 @@ if (Html5.supportsNativeTextTracks()) {
...
   });
 }
+
+test('should check for text track changes when using non native text tracks', function() {
+  let origFeaturesNativeTextTracks = Tech.prototype['featuresNativeTextTracks'];
+  Tech.prototype['featuresNativeTextTracks'] = true;
+  let Player = Component.getComponent('Player');
+  let origTextTracks = Player.prototype.textTracks;
+  let numTextTracksCalls = 0;
+  Player.prototype.textTracks = function() {
```
dmlap: How about listening for `texttrackchange` events instead? It seems like then you wouldn't have to instrument the `textTracks` function and could remove the shadowed `tracks` variable.

gesinger: Yeah, the tricky part about testing it was that the testing needed to be done during player setup, making it hard to listen for specific events since they could’ve happened by the time we had a player reference. That said, since this is a unit test, it is better suited to simply check that emulateTextTracks on Tech fires `texttrackchange` before listening (as you mentioned), and we can avoid the entire player setup and nativeTextTracks setting.

src/js/tech/tech.js
gesinger: Tested, and you are right. The number of handlers grows each time the mode is set to showing, and doesn't shrink when set to modes other than showing. We should be able to simply move it outside of the textTracksChanges function and we can change the bind to that function.

Is it better to handle this in a separate PR, or the same (with another commit)?

gkatsev: Would better as another PR if you want. Otherwise, we'll fix it at some point :)

pam:

Commit: b02fdef05955a68184c458d5162a7d3174056f8f
Build details: https://travis-ci.org/pam/video.js/builds/92834268

gkatsev:
Sweet, that test looks a lot nicer now too.

src/js/tech/tech.js
gesinger: PR for change: https://github.com/videojs/video.js/pull/2849

gkatsev:
LGTM

mmcc:
Lgtm!

Issue #2834, opened 2015-11-20
Make videojs work with npm3


gkatsev:
https://github.com/videojs/video.js/blob/master/src/css/video-js.scss#L5

gkatsev:
Should be working.

PR #2838, opened 2015-11-21
Test against node_js 0.10, 4, 5

paladox:


gkatsev:
Thanks. Unfortunately, because of https://github.com/videojs/video.js/issues/2834 we can't really support node 5 right now. Hopefully, after thanksgiving I'll have time to look at what versions of node we'd want to be testing on. It's probably fine to just leave it on 0.10 for now.
Closing for now.

PR #2843, opened 2015-11-23
Test against node_js 0.10 and 4

paladox:


gkatsev:
Thanks for the PR but I think I'd prefer to leave the build as is for just a bit longer until we get npm3 support and what not.

PR #2844, opened 2015-11-23
Loop through sources then techs

forbesjo:
This PR is to make it so that selectSources iterates through the sources then techOrder. This allows browsers without MSE to play HLS videos through Flash instead of falling back to a mp4 source.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1700,26 +1700,26 @@ class Player extends Component {
    * @method selectSource
    */
   selectSource(sources) {
-    // Loop through each playback technology in the options order
-    for (var i=0,j=this.options_.techOrder;i<j.length;i++) {
-      let techName = toTitleCase(j[i]);
-      let tech = Tech.getTech(techName);
-      // Support old behavior of techs being registered as components.
-      // Remove once that deprecated behavior is removed.
-      if (!tech) {
-        tech = Component.getComponent(techName);
-      }
-      // Check if the current tech is defined before continuing
-      if (!tech) {
-        log.error(`The "${techName}" tech is undefined. Skipped browser support check for that tech.`);
-        continue;
-      }
+    // Loop through each source object
+    for (let a=0,b=sources;a<b.length;a++) {
+      // Loop through each playback technology in the options order
+      for (let i=0,j=this.options_.techOrder;i<j.length;i++) {
+        let techName = toTitleCase(j[i]);
+        let tech = Tech.getTech(techName);
+        // Support old behavior of techs being registered as components.
+        // Remove once that deprecated behavior is removed.
+        if (!tech) {
+          tech = Component.getComponent(techName);
+        }
+        // Check if the current tech is defined before continuing
+        if (!tech) {
+          log.error(`The "${techName}" tech is undefined. Skipped browser support check for that tech.`);
+          continue;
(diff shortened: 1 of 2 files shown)


pam:

Commit: 7010016e93a9e5400caaee48b44055b0f744dc78
Build details: https://travis-ci.org/pam/video.js/builds/92761449

dmlap:
We'd need a flag to enable this for backwards-compatibility.

forbesjo:
Took a test from @imbcmdth and part of his change.

pam:

Commit: c74e116233197da150011a55141bf2da5639150a
Build details: https://travis-ci.org/pam/video.js/builds/92797706

src/js/player.js
```diff
@@ -1700,36 +1700,26 @@ class Player extends Component {
...
+    // to select from them based on their priority.
+    let combinedSources = this.options_.sourceOrder ?
+      sources.reduce((acc, source) => acc.concat(techs.map(tech => ({source, tech}))), []) :
+      techs.reduce((acc, tech) => acc.concat(sources.map(source => ({source, tech}))), []);
+
+    let matchedSource = combinedSources
+      .find(({source, tech}) => tech.canPlaySource(source));
+
+    return matchedSource ?
+      { source: matchedSource.source, tech: matchedSource.tech.name } :
```
imbcmdth: I decided against using the `.name` property on the constructor because there is no guarantee that the constructor will be named or have the same name as the "tech" mapping.

For instance:
```javascript
Tech.registerTech('TotallyNotFoo', Foo);
```
Would have a name of `Foo` but a "techName" of `totallyNotFoo`.

src/js/player.js
```diff
@@ -1700,36 +1700,26 @@ class Player extends Component {
...
-        }
-      }
-    }
-
-    return false;
+    // Get only the techs that exist and are supported by the current platform
+    let techs =
+      this.options_.techOrder
+        .map(toTitleCase)
+        .map(techName => Tech.getTech(techName) || Component.getComponent(techName) ||
```
imbcmdth: While this was clever, I decided to make it more explicit since I wanted to add the comments from the old function back.

src/js/player.js
```diff
@@ -1700,36 +1700,26 @@ class Player extends Component {
...
+          log.error(`The "${techName}" tech is undefined. Skipped browser support check for that tech.`))
+        .filter(tech => tech && tech.isSupported());
+
+    // Depending on the truthiness of options.sourceOrder, we swap the order of techs and sources
+    // to select from them based on their priority.
+    let combinedSources = this.options_.sourceOrder ?
```
imbcmdth: This is very cool. I wonder if we can somehow combine the two very similar reduces here..

forbesjo:
Babel doesn't include .find :cry:

src/js/player.js
forbesjo: You can take my changes/some ideas from this PR since you've done most of the work.

gkatsev:
Babel doesn't shim anything. We're still dealing with an es5 baseline and not an es3 baseline anymore, at least.
Also, we have both this and #2847. We should consolidate on one of them.

forbesjo:
I'll close this one

PR #2849, opened 2015-11-24
Prevent non native text tracks from leaking cuechange event listeners

gesinger:
Due to a new updateDisplay function being generated on each invocation of the textTracksChanges function within emulateTextTracks of Tech, the cuechange event listener which called updateDisplay was not getting removed when looping through text tracks. This created a leak where updateDisplay could be called many times on a single track cuechange event.

An example of this can be seen by running the following:

```javascript
  html5: {
    nativeTextTracks: false
  }
});

player.on(‘ready’, function() {  
  var track = player.addTextTrack('captions');
  track.addCue(new VTTCue(1, 2, 'Test'));
  track.mode = 'showing';
  track.mode = 'showing';
  track.mode = 'showing';
  player.tech_.on('texttrackchange', function() { console.log('Text track change'); });
  player.play();
});
```

When a cuechange is reached (the first being at 1 second), texttrackchange will be triggered 3 times (one for each set of track.mode).

This change moves the updateDisplay function outside of the textTracksChanges function, so that it will not be regenerated on each call of textTracksChanges.

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -321,9 +321,8 @@ class Tech extends Component {
       window['WebVTT'] = true;
     }
 
-    let textTracksChanges = Fn.bind(this, function() {
-      let updateDisplay = () => this.trigger('texttrackchange');
-
+    let updateDisplay = () => this.trigger('texttrackchange');
+    let textTracksChanges = () => {
       updateDisplay();
 
       for (let i = 0; i < tracks.length; i++) {
@@ -333,7 +332,7 @@ class Tech extends Component {
           track.addEventListener('cuechange', updateDisplay);
         }
       }
-    });
+    };
 
     tracks.addEventListener('change', textTracksChanges);
 
--- a/test/unit/tracks/tracks.test.js
+++ b/test/unit/tracks/tracks.test.js
@@ -2,6 +2,7 @@ import ChaptersButton from '../../../src/js/control-bar/text-track-controls/chap
 import SubtitlesButton from '../../../src/js/control-bar/text-track-controls/subtitles-button.js';
 import CaptionsButton from '../../../src/js/control-bar/text-track-controls/captions-button.js';
 
+import TextTrack from '../../../src/js/tracks/text-track.js';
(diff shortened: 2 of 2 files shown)


pam:

Commit: 32a4b8dcd649be2b9e6448d835f76511a9d6cbf3
Build details: https://travis-ci.org/pam/video.js/builds/92882743

src/js/tech/tech.js
```diff
@@ -321,9 +321,8 @@ class Tech extends Component {
       window['WebVTT'] = true;
     }
 
-    let textTracksChanges = Fn.bind(this, function() {
-      let updateDisplay = () => this.trigger('texttrackchange');
-
+    let updateDisplay = Fn.bind(this, () => this.trigger('texttrackchange'));
```
gkatsev: no need to bind the error function (not that you technically can do that).

test/unit/tracks/tracks.test.js
```diff
@@ -342,3 +343,48 @@ if (Html5.supportsNativeTextTracks()) {
     emulatedTt.on('addtrack', addtrack);
   });
 }
+
+test('removes cuechange event when text track is hidden for emulated tracks', function() {
```
gkatsev: previously, we would end up with much larger `numTextTrackChanges`?

src/js/tech/tech.js
gesinger: Nice catch. Thanks!

test/unit/tracks/tracks.test.js
gesinger: Yeah. If you run this test without the change, `numTexttTrackChanges` reveals higher numbers (due to the old event listeners hanging around and triggering `texttrackchange` events.

pam:

Commit: 24d795335e0f741b1fa262382469d88bd74f201a
Build details: https://travis-ci.org/pam/video.js/builds/93072923

gkatsev:
LGTM

mmcc:
LGTM

PR #2863, opened 2015-11-26
Added `display: -moz-box` to display-flex mixin

oanhnn:
Added `display: -moz-box` to display-flex mixin to support Firefox 19- (buggy but mostly works)
Refs: https://css-tricks.com/using-flexbox/

### DIFF
--- a/src/css/_utilities.scss
+++ b/src/css/_utilities.scss
@@ -46,6 +46,7 @@
 
 @mixin display-flex($alignment: '', $justification: '') {
   display: -webkit-box;
+  display: -moz-box;
   display: -webkit-flex;
   display: -ms-flexbox;
   display: flex;


mmcc:
Firefox is an evergreen (auto-updating) browser, so at this point version 19 is ancient. *All* older (< 34) Firefox browsers account for less than 1.5% of Firefox usage, which means on the overall scale, this adds support for an infinitesimally small subset of users (keep in mind that CSS-Tricks blog post is > 2 years old).

If you have a specific reason that this should be pulled in, let's definitely revisit it, but we should be _removing_ vendor prefixes whenever we can, not adding them back.

mmcc:
Also, I was totally rude and remiss in not saying thanks for the pull request! Sorry for not pulling it in, but hopefully the reasoning makes sense.

gkatsev:
Is there a specific reason why you care about firefox 19? The last ESR build is firefox 38 and it was promoted to ESR just earlier this month.

oanhnn:
Thank you for reply.
I think add a line code, Not only it haven't error, but also it will cover more user's browsers.
:smile:

gkatsev:
Adding code isn't cheap. It can create bugs but also creates a maintenance burden, albeit, this one is relatively minor. Firefox 19 was released two years ago and is no longer maintained by mozilla, so, to pull this in we'd need a good reason to do so.

Issue #2852, opened 2015-11-25
support src being an empty string


gkatsev:
the spec basically says that when the src is set to an empty string, you want to reset the player and then stop and not try to loading anything. See step 9 of http://www.w3.org/TR/html5/embedded-content-0.html#concept-media-load-algorithm

Soviut:
This is currently how I test my error styles. Perhaps it would be a good idea to document how to put the player into an error state for error testing.

gkatsev:
You can trigger an error by calling `player.error()` with an object that has a `code` property and various others: `player.error({code: 4})`.

Soviut:
Ah nice, I assumed there was some way to do it; I should have guessed it was that simple.

gkatsev:
So, after trying playing around with setting the `src` to `''`, I realized that it does throw an error. Tested on this page: http://www.w3.org/2010/05/video/mediaevents.html
However, setting the `src` to `null` didn't throw and that might be an acceptable use.
Also, perhaps we want something like a `player.unload()` instead of setting the `src` to `null`?

Soviut:
Could we not just assume that an empty call to `player.src()` means remove call sources?

Also, I'd expect `player.unload()` to remove videoJS from the target and return the `<video>` tag to its original state, not change sources.

gkatsev:
@Soviut `player.src()` needs to act as a getter.
I guess `unload` isn't the right word for it. Just not sure that `player.src(null)` is the best way of implementing the user facing API for this.

Soviut:
Perhaps just document that `player.src({})` or `player.src([])` could clear the sources.

gkatsev:
Well, what I specifically want to happen with `src(null)` is to change the tech to Html5, and put in there a simple video element with no source. Basically, unset all state of the player as much as possible.

Soviut:
That sounds more like a reset. Perhaps a `player.reset()` or `player.init()` is a good idea and then that can be called if src is set to null as well.

dmlap:
If facilities exist in the HTML spec to "reset the video element" than I think we should stick to those alone. Having two ways of doing the same thing can be confusing.

gkatsev:
Trying to read the spec to see if there's a "reset" functionality. Seems like it might be just removing all source elements, deleting the src attribute, and then calling `load()`. Which is what I have hacked together locally currently.

misteroneill:
[...]

PR #2880, opened 2015-12-03
Add Player#reset function.

gkatsev:
Reset loads in the html5 tech and then resets the media element via a
tech call.
A techCall was used so that other techs could implement reset
functionality as well and we could theoretically not switch to the Html5
tech on reset.
Fixes https://github.com/videojs/video.js/issues/2852

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1924,13 +1924,14 @@ class Player extends Component {
   }
 
   /**
-   * Reset the player. Removes all sources, loads the Html tech, and calls `load`.
+   * Reset the player. Loads the first tech in the techOrder,
+   * and calls `reset` on the tech`.
    *
    * @return {Player} Returns the player
    * @method reset
    */
   reset() {
-    this.loadTech_('Html5', null);
+    this.loadTech_(toTitleCase(this.options_.techOrder[0]), null);
     this.techCall_('reset');
     return this;
   }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1924,6 +1924,18 @@ class Player extends Component {
   }
 
   /**
+   * Reset the player. Removes all sources, loads the Html tech, and calls `load`.
+   *
+   * @return {Player} Returns the player
+   * @method reset
+   */
+  reset() {
+    this.loadTech_('Html5', null);
+    this.techCall_('reset');
+    return this;
+  }
+
+  /**
    * Returns the fully qualified URL of the current source value e.g. http://mysite.com/video.mp4
    * Can be used in conjuction with `currentType` to assist in rebuilding the current source object.
    *
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -86,9 +86,6 @@ class Html5 extends Tech {
 
     this.triggerReady();
   }
-  clearMedia() {
-    Html5.clearMediaElement(this.el());
-  }
 
   /**
    * Dispose of html5 media element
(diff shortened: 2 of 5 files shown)


src/js/player.js
```diff
@@ -1900,6 +1900,18 @@ class Player extends Component {
   }
 
   /**
+   * Reset the player. Removes all sources, loads the Html tech, and calls `load`.
+   *
+   * @return {Player} Returns the player
+   * @method reset
+   */
+  reset() {
+    this.loadTech_('Html5', null);
```
gkatsev: Switch to the Html5 tech. Passing in `null` for `source` in `loadTech_` would load the tech in ["late init" mode](https://github.com/gkatsev/video.js/blob/reset/src/js/tech/html5.js#L33-L41) if it needed to switch from another tech.

dmlap: Should this switch to the first tech in the tech order?

src/js/tech/html5.js
```diff
@@ -1092,6 +1104,29 @@ Html5.disposeMediaElement = function(el){
...
+  while (i--) {
+    el.removeChild(sources[i]);
+  }
+
+  // remove any src reference.
+  // not setting `src=''` because that throws an error
+  el.removeAttribute('src');
+
+  if (typeof el.load === 'function') {
+    // wrapping in an iife so it's not deoptimized (#1060#discussion_r10324473)
```
dmlap: I would be surprised if this was called frequently enough where de-opt was a performance issue but it's fine to leave this way, too.

src/js/tech/tech.js
```diff
@@ -234,6 +234,13 @@ class Tech extends Component {
   }
 
   /**
+   * Reset the tech. Removes all sources and resets readyState.
+   *
+   * @method reset
+   */
+  reset() {}
```
dmlap: How will the Flash tech react to this? It seems like reset() will only work for HTML5.

Related: I forgot that src() accepts an array of source objects. What do you think about making `[]` equivalent to a reset?

gkatsev: Reset doesn't accept any input.

src/js/player.js
gkatsev: I was debating that but
1. we would need to implement something for flash in this case and probably would need to modify the swf
2. Setting a source should choose the appropriate tech after the reset
3. Html5 is basically the default tech.

Switching to Html5 and then resetting it is probably the easiest way to get to a nice base state where no sources are loaded and all readyStates are reset.

gkatsev: In an offline discussion we decided that using the first item in the techOrder is the best course of action as it matches what happens in videojs if you were to load it in a page without any sources.
See [flash](http://jsfiddle.net/gkatsev/njku477m/4/embedded/result/) and [html](http://jsfiddle.net/gkatsev/njku477m/5/embedded/result/) examples.

gkatsev:
Looks like I need to rebase this.

gkatsev:
Rebased.

src/js/tech/html5.js
```diff
@@ -86,6 +86,9 @@ class Html5 extends Tech {
 
     this.triggerReady();
   }
+  clearMedia() {
```
dmlap: Is this leftover? Seems unused.

gkatsev: Yep, leftover.

gkatsev: Removed.

dmlap:
LGTM

PR #2889, opened 2015-12-05
Fix keyboard (Space or Enter) operation of buttons. Fixes #2888.

### LINKED ISSUE #2888: Using Space key on the play/pause button in Firefox does not work
Since moving to V5, hitting Enter when the Play/pause button has focus in Firefox acts as expected, and using Enter or Space on Chrome works, but using the Space key on the Play/Pause button with Firefox does not work. It acts a little bit like the button is pressed twice, briefly playing or pausing the video and then switching back to playing or pausing, whichever was being done before the key press.

I was trying to track down if this is a duplicate of an issue which was already reported, and can't find one that's identical. Potentially related to #2669, though.

OwenEdwards:
When video.js moved to using button elements instead of DIVs for buttons in V5, it kept the keyboard handling to simulate a click on Space or Enter key press, even though the browser does this for buttons. This causes an issue with Firefox when performing keyboard operation of buttons like the Play/Pause button.

### DIFF
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -43,10 +43,24 @@ class Button extends Component {
       tabIndex: 0
     }, props);
 
-    // Add standard Aria info
+    if (tag === 'button') {
+      this.isButtonElement_ = true;
+
+      // Add attributes for button element
+      attributes = assign({
+        type: 'button' // Necessary since the default button type is "submit"
+      }, attributes);
+    } else {
+      this.isButtonElement_ = false;
+
+      // Add ARIA attributes for clickable non-button element
+      attributes = assign({
+        role: 'button'
+      }, attributes);
+    }
+
+    // Add common ARIA attributes
     attributes = assign({
-      role: 'button',
-      type: 'button', // Necessary since the default button type is "submit"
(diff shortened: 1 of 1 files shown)


src/js/button.js
```diff
@@ -43,10 +43,24 @@ class Button extends Component {
...
+
+      // Add ARIA attributes for clickable non-button element
+      attributes = assign({
+        role: 'button'
+      }, attributes);
+    }
+
+    // Add common ARIA attributes
     attributes = assign({
-      role: 'button',
```
gkatsev: is it bad to just have both assigned? Or just to make the final markup simpler?

gkatsev:
Have a question about role/type attributes.
LGTM.

gkatsev:
Going to close this because it's getting merged as part of #2897.

src/js/button.js
OwenEdwards: See https://www.w3.org/TR/html5/dom.html#aria-usage-note:

> NOTE: In the majority of cases setting an ARIA role and/or aria-* attribute that matches the default implicit ARIA semantics is unnecessary and not recommended as these properties are already set by the browser.

There has been further discussion about this within W3C, but that's the current recommendation.

PR #2893, opened 2015-12-07
Consistent width for fullscreen button and time-control

hartman:
This changes the width of the fullscreen button to 4em. For some
reason it was 3.8, but that doesn't make much sense and is not in line
with the other controls.

It also sets the textual time-control's width to auto with a 1em
padding to the left and right. Again this is more in line with the
behavior of the other controls.

I do find the entire interface slightly too wide to my liking, but that is exactly why it should be consistent, so that if required, it would be easier for me to override.


gkatsev:
LGTM. Thanks.

dmlap:
LGTM

PR #2896, opened 2015-12-08
Do a nullcheck for cues in updateForTrack. Fixes #2870

### LINKED ISSUE #2870: Cues are sometimes undefined in the updateForTrack function
I have seen several instances where `TextTrackDisplay.prototype.updateForTrack` tries to use cues that are `undefined`. In the `cues` array, I see that there are entries that are `undefined`.

This happens occasionally and the only time I have seen this behavior is when using 608 captions with the latest contrib-hls.

gkatsev:


src/js/tracks/text-track-display.js
```diff
@@ -141,6 +141,11 @@ class TextTrackDisplay extends Component {
 
     let i = cues.length;
     while (i--) {
+      let cue = cues[i];
+      if (!cue) {
+        continue;
+      }
+
       let cueDiv = cues[i].displayState;
```
misteroneill: Since you're creating the `cue` variable, maybe it should be used here?

gkatsev: haha, oops.

misteroneill:
LGTM

Issue #2769, opened 2015-11-03
Screen reader compatibility (accessibility) of control bar menus


OwenEdwards:
The fundamental structure of the 'menu buttons' using ARIA roles does not work because the div with `role="button"` which is the top-level element of the menu button contains actionable elements (menu items) as children. This is prohibited (although it's a little difficult to tease out of the ARIA spec exactly where that prohibition is spelled out). As a result, screen readers do not read out the menu structure correctly.

This is related to, but orthogonal to, #2155, and linked to #841, #1499, #1932 and #2136.

gkatsev:
Fixed by #3033.

PR #2897, opened 2015-12-08
Fix menu keyboard access and ARIA labeling for screen readers (Fixes #2155 and #2769)

### LINKED ISSUE #1446: Triggering the MenuButton using the keyboard cause the top menu item to be triggered.
1. Add a alert or console.log to vjs.MenuItem.prototype.onClick
2. Using the keyboard tab to a Menu Button such as Captions hit enter.
3. The top menu item is triggered.

### LINKED ISSUE #2155: Keyboard accessibility of control bar menus (e.g. captions, subtitles, etc)

OwenEdwards:
Implement ARIA-specified keyboard access to control bar menus (as in http://www.w3.org/TR/wai-aria-practices-1.1/#menu), and ARIA markup to indicate menu role and state to screen readers.

This also fixes #1446 / #1447, which seemed to have come back (or was never actually resolved) as an issue for keyboard access of menus.

This fix depends on #2889.

### DIFF
--- a/src/js/control-bar/control-bar.js
+++ b/src/js/control-bar/control-bar.js
@@ -38,6 +38,8 @@ class ControlBar extends Component {
   createEl() {
     return super.createEl('div', {
       className: 'vjs-control-bar'
+    }, {
+      'role': 'group' // The control bar is a group, so it can contain menuitems
     });
   }
 }
--- a/src/js/menu/menu-item.js
+++ b/src/js/menu/menu-item.js
@@ -42,7 +42,8 @@ class MenuItem extends Button {
   createEl(type, props, attrs) {
     return super.createEl('li', assign({
       className: 'vjs-menu-item',
-      innerHTML: this.localize(this.options_['label'])
+      innerHTML: this.localize(this.options_['label']),
+      tabIndex: -1
     }, props), attrs);
   }
 
--- a/src/js/menu/menu-item.js
+++ b/src/js/menu/menu-item.js
@@ -17,7 +17,18 @@ class MenuItem extends Button {
 
   constructor(player, options) {
     super(player, options);
+
+    this.selectable = options['selectable'];
+
     this.selected(options['selected']);
+
+    if (this.selectable) {
+      // TODO: May need to be either menuitemcheckbox or menuitemradio,
+      //       and may need logical grouping of menu items.
+      this.el_.setAttribute('role', 'menuitemcheckbox');
+    } else {
+      this.el_.setAttribute('role', 'menuitem');
+    }
   }
 
   /**
@@ -51,12 +62,14 @@ class MenuItem extends Button {
    * @method selected
    */
   selected(selected) {
-    if (selected) {
-      this.addClass('vjs-selected');
-      this.el_.setAttribute('aria-selected',true);
-    } else {
-      this.removeClass('vjs-selected');
-      this.el_.setAttribute('aria-selected',false);
+    if (this.selectable) {
+      if (selected) {
+        this.addClass('vjs-selected');
+        this.el_.setAttribute('aria-checked',true);
+      } else {
+        this.removeClass('vjs-selected');
+        this.el_.setAttribute('aria-checked',false);
+      }
     }
   }
 
--- a/src/js/menu/menu-item.js
+++ b/src/js/menu/menu-item.js
@@ -67,13 +67,18 @@ class MenuItem extends Button {
       if (selected) {
         this.addClass('vjs-selected');
         this.el_.setAttribute('aria-checked',true);
+        // aria-checked isn't fully supported by browsers/screen readers,
+        // so indicate selected state to screen reader in the control text.
+        this.controlText(', selected');
       } else {
         this.removeClass('vjs-selected');
         this.el_.setAttribute('aria-checked',false);
+        // Indicate un-selected state to screen reader
+        // Note that a space clears out the selected state text
+        this.controlText(' ');
       }
     }
   }
-
 }
 
 Component.registerComponent('MenuItem', MenuItem);
--- a/src/js/menu/menu.js
+++ b/src/js/menu/menu.js
@@ -48,7 +48,7 @@ class Menu extends Component {
     this.contentEl_ = Dom.createEl(contentElType, {
       className: 'vjs-menu-content'
     });
-    this.contentEl_.setAttribute('role', 'presentation');
+    this.contentEl_.setAttribute('role', 'menu');
     var el = super.createEl('div', {
       append: this.contentEl_,
       className: 'vjs-menu'
--- a/src/js/menu/menu.js
+++ b/src/js/menu/menu.js
@@ -15,6 +15,14 @@ import * as Events from '../utils/events.js';
  */
 class Menu extends Component {
 
+  constructor (player, options) {
+    super(player, options);
+
+    this.focusedChild_ = -1;
+
+    this.on('keydown', this.handleKeyPress);
+  }
+
   /**
    * Add a menu item to the menu
    *
@@ -25,6 +33,7 @@ class Menu extends Component {
     this.addChild(component);
     component.on('click', Fn.bind(this, function(){
       this.unlockShowing();
+      //TODO: Need to set keyboard focus back to the menuButton
     }));
   }
 
@@ -39,10 +48,12 @@ class Menu extends Component {
     this.contentEl_ = Dom.createEl(contentElType, {
       className: 'vjs-menu-content'
     });
+    this.contentEl_.setAttribute('role', 'presentation');
     var el = super.createEl('div', {
       append: this.contentEl_,
       className: 'vjs-menu'
     });
+    el.setAttribute('role', 'presentation');
     el.appendChild(this.contentEl_);
 
     // Prevent clicks from bubbling up. Needed for Menu Buttons,
@@ -53,6 +64,72 @@ class Menu extends Component {
     });
 
     return el;
+  }
+
+  /**
+   * Handle key press for menu
+   *
+   * @param {Object} event Event object
+   * @method handleKeyPress
+   */
+  handleKeyPress (event) {
+    if (event.which === 37 || event.which === 40) { // Left and Down Arrows
+      event.preventDefault();
+      this.stepForward();
+    } else if (event.which === 38 || event.which === 39) { // Up and Right Arrows
+      event.preventDefault();
+      this.stepBack();
+    }
+  }
+
+  /**
+   * Move to next (lower) menu item for keyboard users
+   *
+   * @method stepForward
+   */
+   stepForward () {
+     let stepChild = 0;
+
+     if (this.focusedChild_ !== undefined) {
+       stepChild = this.focusedChild_ + 1;
+     }
+     this.focus(stepChild);
+   }
+
+   /**
+    * Move to previous (higher) menu item for keyboard users
+    *
+    * @method stepBack
+    */
+  stepBack () {
+    let stepChild = 0;
+
+    if (this.focusedChild_ !== undefined) {
+      stepChild = this.focusedChild_ - 1;
+    }
+    this.focus(stepChild);
+  }
+
+  /**
+   * Set focus on a menu item in the menu
+   *
+   * @param {Object|String} item Index of child item set focus on
+   * @method focus
+   */
+  focus (item = 0) {
+    let children = this.children();
+
+    if (children.length > 0) {
+      if (item < 0) {
+        item = 0;
+      } else if (item >= children.length) {
+        item = children.length - 1;
+      }
+
+      this.focusedChild_ = item;
+
+      children[item].el_.focus();
+    }
   }
 }
 
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -39,8 +39,7 @@ class Button extends Component {
    */
   createEl(tag='button', props={}, attributes={}) {
     props = assign({
-      className: this.buildCSSClass(),
-      tabIndex: 0
+      className: this.buildCSSClass()
     }, props);
 
     if (tag === 'button') {
@@ -52,6 +51,11 @@ class Button extends Component {
       }, attributes);
     } else {
       this.isButtonElement_ = false;
+
+      // If the element isn't a button, it needs tabIndex=0 to allow keyboard focus
+      props = assign({
+        tabIndex: 0
+      }, props);
 
       // Add ARIA attributes for clickable non-button element
       attributes = assign({
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -43,10 +43,24 @@ class Button extends Component {
       tabIndex: 0
     }, props);
 
-    // Add standard Aria info
+    if (tag === 'button') {
+      this.isButtonElement_ = true;
+
+      // Add attributes for button element
+      attributes = assign({
+        type: 'button' // Necessary since the default button type is "submit"
+      }, attributes);
+    } else {
+      this.isButtonElement_ = false;
+
+      // Add ARIA attributes for clickable non-button element
+      attributes = assign({
+        role: 'button'
+      }, attributes);
+    }
+
+    // Add common ARIA attributes
     attributes = assign({
-      role: 'button',
-      type: 'button', // Necessary since the default button type is "submit"
       'aria-live': 'polite' // let the screen reader user know that the text of the button may change
     }, attributes);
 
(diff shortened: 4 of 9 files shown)


gkatsev:
Thanks, @OwenEdwards. I'll take a look at it ASAP.

OwenEdwards:
@gkatsev hold off on pulling this - I need to do more work on the "volume menu button" since keyboard handling may conflict (arrow keys apply to the volume slider, not to moving between menu items).

gkatsev:
Alright.
When we're focused inside of the menu, the arrow keys probably should just interact with the menu.
If we're just focused on the player in general, i.e., after hitting the big play button, the arrow keys should control volume/seeking.

OwenEdwards:
@gkatsev that would be nice to have, but I don't see that it's implemented in V5 right now? I thought this was something that was left to [videojs-hotkeys](https://github.com/ctd1500/videojs-hotkeys) "A plugin for Video.js that enables keyboard hotkeys when the player has focus"?

I was specifically referring to the fact that a [slider handles arrow keys when it has focus](https://github.com/videojs/video.js/blob/master/src/js/slider/slider.js#L178-L186), but the ARIA definition of a menu defines that [arrow keys move focus between menu items](https://github.com/OwenEdwards/video.js/blob/fix/menu-keyboard-access/src/js/menu/menu.js#L75-L83), so there's a conflict. By ARIA standards, the [Volume Menu Button](https://github.com/videojs/video.js/blob/master/src/js/control-bar/volume-menu-button.js) isn't really a 'menu', it's a custom widget with a button and a slider. It could even have a single keyboard focus, where arrow keys adjust volume and Space or Enter mute/unmute the player since right now its keyboard interaction is a little confused.

gkatsev:
Didn't know about that plugin and yeah, most of the hotkeys should still be left to the plugin. I think some stuff we probably should have in core. Like `<space>` should probably toggle the playback state when the player is in focus. But `f` to fullscreen isn't necessary to be in core. But it's debatable.

Ah, interesting. So, the issue is how apply the ARIA spec to the `volume menu button` which is a "menu" in vjs but not according to ARIA?

OwenEdwards:
@gkatsev right, that's the concise way of putting it! The volume menu button is a "menu" in VJS, but not a menu according to ARIA, where it's either a button and a slider, or a custom widget.

One thing which I'm still not clear about is that, for vjs menus, they open [on hover due to CSS](https://github.com/videojs/video.js/blob/master/src/css/components/menu/_menu-popup.scss#L22), but they also have this handling of [click](https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L203) and [lockShowing](https://github.com/videojs/video.js/blob/master/src/js/component.js#L955-L979) in JS. Do you know, is this to allow a CSS skin to change the onhover/onclick behavior? I haven't been able to track it down, and trying to get everything to work with keyboard-only operation without breaking existing behavior is tricky! ;-)

Let's split 'hotkeys' onto a separate GitHub issue - it ties into accessibility, but I don't want to muddy this issue too much!

gkatsev:
I think the `lock-showing` thing may be to help with exactly this situation. So you could select the menu and it will keep it open so you could keep interacting with the things inside the menu. Or something.
I know that it used to be badly broken in some versions of vjs 4. I'm not sure whether it was ever fixed or how it works in 5.x. Also, it doesn't seem to be doing anything with the volume menu button currently.

OwenEdwards:
@gkatsev @mmcc @heff can any of you explain what the `vjs-lock-showing` class and `lockShowing` method are trying to do? The more I dig, the more I think it's currently a bug.

For the Captions and Subtitles menus, they seem to do nothing because clicking on the menu button opens the menu (and calls `lockShowing`), but `unlockShowing` is called on [mouseout](https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L160-L163), so the menu closes if mouse hover moves off the menu. Similarly, clicking a second time on the menu button clears `vjs-lock-showing`, but the menu stays open because the [CSS treats `vjs-lock-showing` and `hover` identically](https://github.com/videojs/video.js/blob/master/src/css/components/menu/_menu-popup.scss#L22-L23). And `vjs-lock-showing` has no effect on the controls fading out during playback.

For the `volume-menu-button`, when it is `inline`, then it does seem like clicking on the mute button does sometimes lock the volume slider as visible, even when the mouse pointer moves off the mute button or slider. But this means _sometimes clicking on the mute button mutes the player, and sometimes it doesn't_!! I haven't been able to track down exactly when it does which one, but it means one control has two functions!!

As I try to fix keyboard accessibility of controls which are not just a button, issues with this lockShowing functionality are becoming a major roadblock, and I think they're actually usability issues becond keyboard accessibility. What do you guys think?

gkatsev:
I think `lock-showing` is just broken or not working correctly, though, with the best intentions.
As for the inline volume menu button, the issue is that the slider was overlayed on top of the mute toggle so you sometimes were clicking the slider instead of the mute toggle. I'm fixing it here: https://github.com/videojs/video.js/pull/2913
The "lock-showing" of the volume is, I believe, a side-effect of trying to make it keyboard accessible and keeping the slider/menu open when focused. Trying to address it here: https://github.com/videojs/video.js/pull/2638/files#diff-6b8a908fd9a1289ee34105f594366da4R73.

We probably should change lock-showing to apply when clicking on the menu button and get removed when clicking on it again and not be removed on `mouseout`. As for the volume menu button, is fiddling around with ARIA attributes help? Should we look into actually splitting it out into two components that are siblings instead of ancestor/descedant?

heff:
Re: lockShowing...if I remember right it was meant to allow for keyboard triggering of menus. So tab to the menu button, hit space bar to open the menu, tab to the first menu item and menu stays showing. I can't say if that's how it is now, but I think that would be the main use case to confirm is still working if you strip it out.

gkatsev:
Like the discussion [here](https://github.com/videojs/video.js/pull/2638/files#r47577698), I think the flaw is with keyboard+mouse interaction. I.e., if you tab into the menu and then mouse over it, it'll be hidden, but should it? We probably would want to know whether we tabbed into the control and only apply the mouseout event if opened the menu with the mouse.

gkatsev:
I opened a new issue that spun out of #2638 and is related to the lock-showing/keyboard/stuff discussion: https://github.com/videojs/video.js/issues/2918

OwenEdwards:
@gkatsev I think the issue with keyboard+mouse interaction is a bit of a distraction - I'm thinking of control interaction in this order:

* mouse only
* touchscreen only (which you've been working on)
* keyboard only (WCAG)
* keyboard-only with screen reader (WCAG)
* touchscreen with screen reader (WCAG)
* voice control (e.g. Dragon NaturallySpeaking) (WCAG)
* mouse+keyboard

I think mouse+keyboard is a concern, but not until we've got those other ones working!

So the questions I still feel hasn't been answered is:

1/. Do we want the Captions and Subtitles menu buttons to:

a/. Appear on hover (mouse) and on tap (touchscreen), or
b/. Appear on click (not on hover) and on tap

2/. Do we want a mechanism for mouse users to open the menu and leave it open after mouseout?

3/. Do we want the behavior in 1 and 2 to be changeable for site creators? Via CSS skin, or via Video.js parameters?

The preferred behavior of menus for keyboard-only users (including keyboard-only screen reader users) is defined by the ARIA specs, and should follow most or all of that pattern, but this does not affect the mouse and touchscreen operation of these controls.

Then we need to make a decision for the combined volume slider and mute button, in both inline and non-inline/vertical configuration. This does not have to match the menu operation, and the keyboard interaction with this combined control is definitely going to be different from keyboard interaction with the menu.

I do think we should split the `volume-menu-button` away from being a descendent of a menu, although it's possible that there could be some common ancestor for the two of them. In fact, I have been working on doing exactly this, and creating a component called a `popup`, which is part of what lead to this question about lock-showing; once we're clear on what control features we want to support, we'll can look at how much these two controls will have in common.

gkatsev:
I just realized that I'm doing some of the stuff you already had in here, for example in #2994. I'll go through and de-dupe and make suggestions and what not.

src/js/menu/menu-item.js
```diff
@@ -17,7 +17,18 @@ class MenuItem extends Button {
...
     super(player, options);
+
+    this.selectable = options['selectable'];
+
     this.selected(options['selected']);
+
+    if (this.selectable) {
+      // TODO: May need to be either menuitemcheckbox or menuitemradio,
+      //       and may need logical grouping of menu items.
+      this.el_.setAttribute('role', 'menuitemcheckbox');
```
gkatsev: locally, I've played around with the `menu` itself to a role of `listbox` and these to a role of `option`.
Using ChromeVox for testing it actually produced decent results since it told you how many options and whether they're selected or not.

src/js/control-bar/control-bar.js
```diff
@@ -38,6 +38,8 @@ class ControlBar extends Component {
   createEl() {
     return super.createEl('div', {
       className: 'vjs-control-bar'
+    }, {
+      'role': 'group' // The control bar is a group, so it can contain menuitems
```
gkatsev: wouldn't a `menubar` role be more "correct", assuming I've read the aria spec for that section correctly.

src/js/menu/menu-item.js
gkatsev: Not sure if this is a chromevox issue or not, but if I have captions in the list, when I expand the captions menu, it'll read the label of the last caption in the list.

gkatsev:
Also, to answer the questions in your last comment.
1. I think option a is what we want. Open on hover with mouse and tap on touchscreen
2. I think that clicking the button should lock the menu and leave it open until it is pressed again
3. I don't think we need to have an option for this but we should try to use CSS and html mostly with limited js functionality for this. That way, it could possibly be configured via CSS.

Moving the `volume-menu-button` to be a "popup" type role. Though, I wonder whether we can do it safely while maintaining backwards compatibility.

gkatsev:
Looks like ChromeVox, while nice for getting an initial idea for things, functions differently than VoiceOver and others.

src/js/menu/menu-item.js
gkatsev: Going to chalk it up to a bug in chromevox.

OwenEdwards: Don't base your understanding of whether accessibility is working on ChromeVox!!! You'll hurt your credibility with accessibility experts ;-) It's convenient because it's free, but Firefox+NVDA on a PC is a much better free, standards-compliant alternative. See http://webaim.org/projects/screenreadersurvey/.

That said, I've also been looking at some options around `listbox` and `option`. Semantically, the menus (e.g. the CC menu) are complicated because they have a current state (selected item), potentially a non-actionable title, a settings menu option (which is not "selectable", but is actionable), and a set of options of which one can be picked (we had that discussion about whether more than one could be picked, if we combine CC and Subtitles into a single menu, but I think we concluded they couldn't).

src/js/control-bar/control-bar.js
OwenEdwards: Unfortunately not, since we want to be able to add other control widgets into the Control Bar which can violate the restrictions of a `menubar`. Right now, the Control Bar is essentially a combination of a `menubar

src/js/menu/menu-item.js
gkatsev: I'll check out NVDA. I've been using OSX's VoiceOver the latter half of last week once I realized that ChromeVox isn't good enough.

gkatsev:
Also, I've been thinking that we should just merge this because it's progress towards better accessibility. I have some updates to my PR (#2994) that are based on these changes that hopefully also make it slightly better.
@OwenEdwards does that sound good? Also, is there a good way to contact you for some more real-time communication?

src/js/menu/menu-item.js
```diff
@@ -51,15 +63,22 @@ class MenuItem extends Button {
    * @method selected
    */
   selected(selected) {
-    if (selected) {
-      this.addClass('vjs-selected');
-      this.el_.setAttribute('aria-selected',true);
-    } else {
-      this.removeClass('vjs-selected');
-      this.el_.setAttribute('aria-selected',false);
+    if (this.selectable) {
```
dmlap: Just a general thought, not necessary for this PR: would we have an easier time with menu-button/menu-item if we started from a `<select>` element?

src/js/menu/menu-item.js
```diff
@@ -51,15 +63,22 @@ class MenuItem extends Button {
...
+        this.el_.setAttribute('aria-checked',true);
+        // aria-checked isn't fully supported by browsers/screen readers,
+        // so indicate selected state to screen reader in the control text.
+        this.controlText(', selected');
+      } else {
+        this.removeClass('vjs-selected');
+        this.el_.setAttribute('aria-checked',false);
+        // Indicate un-selected state to screen reader
+        // Note that a space clears out the selected state text
+        this.controlText(' ');
```
dmlap: It seems to me that this would end up clearing out the original control text after toggling between the selected and unselected states but I may be missing something. controlText() looks like a plain setter in Button. If the menu item text was original "English captions" it would change the text to ", selected" after selection and then clear it out entirely when it's unselected.

src/js/menu/menu.js
```diff
@@ -25,6 +33,7 @@ class Menu extends Component {
     this.addChild(component);
     component.on('click', Fn.bind(this, function(){
       this.unlockShowing();
+      //TODO: Need to set keyboard focus back to the menuButton
```
dmlap: @gkatsev I hope this is on your list to fix forward.

src/js/menu/menu.js
```diff
@@ -54,6 +65,72 @@ class Menu extends Component {
...
+   * Handle key press for menu
+   *
+   * @param {Object} event Event object
+   * @method handleKeyPress
+   */
+  handleKeyPress (event) {
+    if (event.which === 37 || event.which === 40) { // Left and Down Arrows
+      event.preventDefault();
+      this.stepForward();
+    } else if (event.which === 38 || event.which === 39) { // Up and Right Arrows
```
dmlap: If we're going to do keypress handling in a number of different places, we probably should consolidate these character codes.

dmlap:
@gkatsev I'm cool with moving forward with this, especially if you have additional fixes in the wings. I'd be curious to hear your thoughts on any of my comments, especially the thought about `<select>`.

src/js/menu/menu-item.js
gkatsev: Yes, just using a `<select>` should hopefully makes things easier, though, it'll make even less sense to have the caption settings menu button in it then, so, we'll have to put it elsewhere.
Also, `<select>` is *not* styleable at all.

gkatsev: Yes, just using a `<select>` should hopefully makes things easier, though, it'll make even less sense to have the caption settings menu button in it then, so, we'll have to put it elsewhere.
Also, `<select>` is *not* styleable at all.

src/js/menu/menu.js
gkatsev: Yes, yes we should.

src/js/menu/menu-item.js
OwenEdwards: We could rename `controlText` as `accessibilityText`, since it's the text equivalent for the visual of the button. For the play/pause button, the visual is from the videojs font set and looks like ▶(which is not announced by screen readers), while the `controlText` is `Play` or `Pause`.

For `menu-item`s, the visual is 'English captions' (which is handled correctly by screen readers), so `controlText` is not needed or used. I added the ', selected' or ' ' as essentially an alternate use of the `controlText` field.

src/js/menu/menu.js
OwenEdwards: Agreed! That was on my list to look at as park of this accessibility work; any thoughts on where it should go? utils/dom.js, perhaps? We need the ESC key from the modal dialog, too.

gkatsev: I've been thinking of grabbing this lib: https://github.com/timoxley/keycode
Seems to be small and tested and does exactly what we want.

OwenEdwards: The https://github.com/timoxley/keycode library seems like it would work. Potentially a little overkill (I was thinking of something as simple as the `this.keys` array from http://oaa-accessibility.org/example/15/), but why reinvent the wheel??

gkatsev: I opened an issue for this: https://github.com/videojs/video.js/issues/3022

gkatsev:
Superseded by #3033.

PR #2902, opened 2015-12-08
Set a min-width for the progress slider of 4em

hartman:
This is in line with the width of other controls. And at the default
size of the player, anything under 4em is rather unusable actually.


gkatsev:
LGTM

dmlap:
LGTM

Issue #2831, opened 2015-11-20
"videojs is not a function" error - in 5.2.1


MattS2K:
This error happens if you use auto-setup and it is initially called before the page has completed loading. Tracking through, it is caused by the following code:

```
var autoSetup = function autoSetup() {
  
     //cut code

        // If getAttribute isn't defined, we need to wait for the DOM.
      } else {
          autoSetupTimeout(1);
          break;
        }
    }

    // No videos were found, so keep looping unless page is finished loading.
  } else if (!_windowLoaded) {
      autoSetupTimeout(1);
    }
  };

// Pause to let the DOM keep processing
var autoSetupTimeout = function autoSetupTimeout(wait, vjs) {
    videojs = vjs;
    setTimeout(autoSetup, wait);
};

```

The calls to autoSetupTimeout are resetting videojs to undefined.

Soviut:
Your `autoSetupTimeout` function allows a vjs to be passed to it which then replaces videojs, however, your calls to it never pass vjs.

This passing of the vjs object seems unnecessary. Why not wait until videojs has loaded. You can use its callback to start your timeout since there's no point in starting it sooner.

Additionally, why not use a DOM ready event such as `DOMContentLoaded` (https://developer.mozilla.org/en-US/docs/Web/Events/DOMContentLoaded) to check if the page has finished loading before running the videojs setup?

MattS2K:
thanks for the response, but you're misunderstanding slightly - the code above is from videojs - its stomping all over itself! I have got around this myself, but posted it to make the videojs developers aware of the issue so they can fix it. I agree with your comments though.

svetlin-mladenov:
I am experiencing the same issue with 5.6.0 and judging from the code the problem is still present in 5.7.0.

Just to make it clear all the code below is from video.js source code itself.

Here is a possible ordering of events that triggers the bug:
 * video.js is loaded and `setup.autoSetupTimeout(1, videojs);` in [video.js](https://github.com/videojs/video.js/blob/master/src/js/video.js) is executed.
 * after a while the autoSetup function in [setup.js](https://github.com/videojs/video.js/blob/8d9695b6adb352ace30fd95e25e61cbae6520e83/src/js/setup.js) is run. The dom is not ready so `autoSetupTimeout(1);` is called at line 69. However `autoSetupTimeout` takes two arguments: a timeout and a videojs object. The second argument is not specified so it's assumed `undefined`
[...]

PR #2906, opened 2015-12-10
Fixing #2831

eckz:
Fixing #2831. Not setting videojs to undefined when the window is not loaded.


gkatsev:
#2831 has been fixed elsewhere in #3233. Thanks for the PR, though.

PR #2913, opened 2015-12-14
Width of buttons is 4em. So, left:4em;

gkatsev:
Also, we want to add a bit more margin on the ends for the volume-bar so
the bar head won't get cut off at the top or bottom.

### DIFF
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -25,7 +25,7 @@
 }
 
 .video-js .vjs-volume-bar {
-  margin: 1.35em;
+  margin: 1.35em 0.45em;
 }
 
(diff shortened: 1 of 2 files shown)


dmlap:
LGTM

richardbushell:
v5.4.5 with this pull BREAKS the vertical volume bar, see attached screengrab:-
vjs545 vertical volume error

src/css/components/_volume.scss
```diff
@@ -25,7 +25,7 @@
 }
 
 .video-js .vjs-volume-bar {
-  margin: 1.35em;
+  margin: 1.35em 0.45em;
```
gkatsev: This broke the vertical volume bar.

PR #2923, opened 2015-12-16
Correct adaptive layout selectors to match their intent

hartman:
There were a few problems
1: The fullscreen button was hidden in vjs-layout-x-small, unlike its
comments claimed and not matching with tiny, where it is visible.
2: The vjs-captions-control was hidden twice for tiny and x-small
3: The vjs-captions-control was hidden in small, even though the
intent was (per comments) for it to be visible
4: The vjs-volume-menu-button was present on x-small, because it was
incorrectly specified as .vjs-volume-button

All of this was partly because the lines for each layout were
different and thus 'hard' to compare between the layouts. I have now
split them in 4 groups/lines. Time controls, rate/progress, volume,
metadata. This will hopefully prevent similar problems in the future.

### DIFF
--- a/src/css/components/_adaptive.scss
+++ b/src/css/components/_adaptive.scss
@@ -5,10 +5,10 @@
   .vjs-custom-control-spacer { @include flex(auto); }
   &.vjs-no-flex .vjs-custom-control-spacer { width: auto; }
 
-  .vjs-current-time, .vjs-captions-button, .vjs-time-divider,
-  .vjs-progress-control, .vjs-duration, .vjs-remaining-time, .vjs-playback-rate,
-  .vjs-mute-control, .vjs-volume-control, .vjs-chapters-button, .vjs-captions-button,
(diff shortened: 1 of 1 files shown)


hartman:
ping.. anyone want to look at this ?

mmcc:
:heart::heart: the reorganization. Much cleaner.

I would remove fullscreen in tiny, that was almost certainly an oversight. Otherwise, looks good to me! :thumbsup:

hartman:
@mmcc actually, the comments on line 3 (above the definition of tiny), indicate the fullscreen button in tiny was intentional. Could be changed of course.

mmcc:
That it does. I could have sworn I meant for tiny to just be the play button, but two buttons is probably tiny enough :)

mmcc:
lgtm :shipit:

gkatsev:
LGTM

PR #2924, opened 2015-12-16
Change the innerText to make the Close button a better read

vdeshpande:
This change is to make screen readers read the close button as "Close Dialog" instead of just "Close" so that the label is clearly understood instead of a generic "Close" label as the innerText.


dmlap:
If you change one of the English phrases in the player, you need to update it in all the localization files as well. Also, we need to double check "Close" isn't used for anything that isn't related to closing a dialog.

gkatsev:
@vdeshpande hey, are you able to update this PR based on @dmlap's comments? Thanks.

vdeshpande:
@gkatsev Will do !

vdeshpande:
@dmlap "Close" isn't used in any of the language json files and also upon checking seems to be not used for anything that isn't related to closing a dialog.

OwenEdwards:
To be super captain-pedantic, shouldn't the change of the text for the `close-button` to "Close Dialog" be something which is passed as a parameter on instantiation from the `modal-dialog`? (In the same what that button's `controlText` is overridden for each instance). The `close-button` is modular, and could be used somewhere else, where "Close Dialog" wouldn't be appropriate. Like "Close mouth"?? ;-)

OwenEdwards:
I'm looking at this again, and I do think that the change you're proposing should be in `modal-dialog.js`, not in `close-button.js`. Something like:

https://github.com/videojs/video.js/blob/master/src/js/modal-dialog.js#L256-L262:

````
        // The close button should be a child of the modal - not its
        // content element, so temporarily change the content element.
        let temp = this.contentEl_;
        this.contentEl_ = this.el_;
        close = this.addChild('closeButton');
        close.controlText('Close Modal Dialog'); // Note that the term "Modal" is used a lot elsewhere in modal-dialog.js
        this.contentEl_ = temp;
        this.on(close, 'close', this.close);
````

If you look in `clickable-component.js` (and `button.js`, which extends it), the default text there is "*Need Text*"; I think the default for a `close-button` should be the same,or something like "Close Undefined Feature".

@dmlap, @gkatsev - am I being too pedantic here? I'll defer to your judgement on this.

gkatsev:
@OwenEdwards no, that sounds like the correct direction here. @vdeshpande would you be up for making that change?
Also, each of the language files should have their keys changed as part of this.

vdeshpande:
Sure. I'll give it a shot

vdeshpande:
@gkatsev , @OwenEdwards How does this look ?

vdeshpande:
Something went wrong here. 
I have opened up a fresh new PR to tackle this issue here : #3400 
I will close this PR.

PR #2946, opened 2015-12-24
Replace single quotes with double quotes, remove horizontal tables in Tracks page

CoWinkKeyDinkInc:
The horizontal tables breaks and turns into regular code as seen on the site http://docs.videojs.com/docs/guides/tracks.html  So removed the tables in the tables so it's just one big long vertical table so maybe it won't break like this.

Many parts of the docs still mentioned using the `<source>`'s  `type` attribute with single quotes.  This still works fine with double quotes, and to make sure some people don't get confused, I changed it to double quotes so it matches with every other HTML attribute.  Some parts of the docs already use it with double quotes anyway.


gkatsev:
LGTM.

PR #2947, opened 2015-12-27
add startpram 4 pseudo streaming

mysamimi:
when used mp4 video and set first techOrder flash and seek forwards or change the quality by videojs-resolution-switcher, video-js-swf get all before media. I think best solution is Pseudostreaming.


gkatsev:
Thanks for the PR; however, as mentioned in https://github.com/videojs/video-js-swf/pull/196, we're not looking to support this feature at this time.

PR #2953, opened 2015-12-28
New simple dev server

gkatsev:
I decided to add a new super simple dev server that also supports live reload. I also plan to try and figure out whether I can add in [`browserify-hmr`](https://www.npmjs.com/package/browserify-hmr) to this setup.
To get started with this is as easy as just running

``` sh
$ npm start
```

It will start up browserify using watchify and then open up the `index.html` file in your default browser. You can change this behavior via options.

``` sh
$ npm start -- --no-open
```

The initial `--` is a unix convention to know when to stop parsing options for the original command, in this case npm, and then pass it to the subscript. So, the option is `--no-open`. Or `--no-o` for short. This last command won't open a new browser window.
You can also add a `--live` option so that it'll do live-reload.

``` sh
$ npm start -- --live --no-open
```

This enables live reload which will reload the page if the javascript changes and watchify rebuilds the bundle.
When set, it'll also watch the sass files and recompile it and then also reload the page.

One of the main reasons for this is to use [budo](https://www.npmjs.com/package/budo) which is a browserify dev server that would actually delay the response of the things being browseried until the bundle finishes bundling.
The main downside currently is the doubling of the browserify options, though, they're unlikely to change soon.

### DIFF
--- a/build/bin/dev
+++ b/build/bin/dev
@@ -1,0 +1,69 @@
+#!/usr/bin/env node
+var path = require('path');
+var sh = require('shelljs');
+var nodeCli = require('shelljs-nodecli');
+var budo = require('budo');
+var minimist = require('minimist');
+var pkg = require('../../package.json');
+var version = pkg.version.split('.');
+var argv = minimist(process.argv, {
+  boolean: ['l', 'o'],
+  alias: {
+    l: 'live',
+    o: 'open'
+  },
+  default: {
+    o: false,
+    l: false
+  }
+});
+
+version = {
+  full: pkg.version,
+  major: version[0],
+  minor: version[1],
+  patch: version[2],
+  majorMinor: version[0] + version[1]
+};
+
+var b = budo(process.cwd() + '/src/js/video.js', {
+  // log to stdout
+  stream: process.stdout,
+  // open up the server the browser
+  open: argv.open,
+  port: 9999,
+  verbose: true,
+  live: argv.live,
+  dir: process.cwd(),
+  serve: 'video.js',
+  browserify: {
+    standalone: 'videojs',
+    transform: [
+      require('babelify').configure({
+        sourceMapRelative: process.cwd(),
+        loose: ['all']
+      }),
+      ['browserify-versionify', {
+        placeholder: '__VERSION__',
+        version: pkg.version
+      }],
+      ['browserify-versionify', {
+        placeholder: '__VERSION_NO_PATCH__',
+        version: version.majorMinor
+      }],
+      ['browserify-versionify', {
+        placeholder: '__SWF_VERSION__',
+        version: pkg.dependencies['videojs-swf']
+      }]
+    ]
+  }
+});
+
+b.watch('src/css/**/*.scss').on('watch', function(type, file) {
+  if (path.extname(file) === '.scss') {
+    nodeCli.exec('node-sass', 'src/css/video-js.scss', 'build/temp/video-js.css');
+    b.reload('build/temp/video-js.css');
+  } else {
+    b.reload(file);
+  }
+});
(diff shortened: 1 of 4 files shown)


build/grunt.js
```diff
@@ -14,6 +14,10 @@ module.exports = function(grunt) {
     patch: verParts[2]
   };
 
+  version.majorMinor = `${version.major}.${version.minor}`;
+  grunt.vjsVersion = version;
+
```
gkatsev: oops, extra whitespace.

build/grunt.js
```diff
@@ -14,6 +14,10 @@ module.exports = function(grunt) {
     patch: verParts[2]
   };
 
+  version.majorMinor = `${version.major}.${version.minor}`;
```
gkatsev: This part should get merged in regardless (I can open a new PR if we don't want the rest of this PR) because it could actually break things since `version.majorMinor` was being used before it was defined.

build/bin/dev
```diff
@@ -0,0 +1,71 @@
...
+
+version = {
+  full: pkg.version,
+  major: version[0],
+  minor: version[1],
+  patch: version[2],
+  majorMinor: version[0] + version[1]
+};
+
+var b = budo(process.cwd() + '/src/js/video.js', {
```
gkatsev: these paths and what not should be fiddled with before merging. Need to figure out exactly what I can put down. I used to have `'./'` here but it's broken when running through `npm start`.

gkatsev:
Looks like to support browserify-hmr, we'll probably need to write some browserify transform that adds the `module.hot` boilerplate appropriately to our modules. So, I think more than out-of-scope for this PR. However, I'll still be interested in trying it out at some point.

build/bin/dev
```diff
@@ -0,0 +1,71 @@
...
+      ['browserify-versionify', {
+        placeholder: '__SWF_VERSION__',
+        version: pkg.dependencies['videojs-swf']
+      }]
+    ]
+  }
+});
+
+if (argv.live) {
+  b.watch('src/css/**/*.scss').on('watch', function(type, file) {
```
misteroneill: Does this mean that it will only live reload on CSS changes or is the JS watching automatic?

package.json
```diff
@@ -14,6 +14,8 @@
   "homepage": "http://videojs.com",
   "author": "Steve Heffernan",
   "scripts": {
+    "budo": "budo",
```
misteroneill: Is the intent here to simply expose `budo` for developers wanting to call it manually (e.g. `npm run budo -- some-args`)?

misteroneill:
I dig it. :+1: I was thinking of writing a standalone package similar to this to use as a dependency for plugins. Do you think it would make sense to try to abstract this PR out into a separate project?

Would have to support passing it some options, so it might require an external script regardless. Perhaps for now it's better to have the plugin ecosystem use a different simple dev server?

build/bin/dev
gkatsev: Yep, js watching is automatic. This is specifically to rebuild the sass files when they change.
Also, I'm playing around with browserify-hmr instead of live reload for the js.

package.json
gkatsev: This is how I first started playing around with `budo`. We can leave it in or not.

gkatsev:
Yeah, abstracting it out totally makes sense to me. Something like `videojs-dev-server`. The main difference between this and plugins are the main browserify options. And we could keep the transforms enabled anyway so that people could use them if they want but we'd want to still allow the option to override or add extra browserify options.

build/bin/dev
misteroneill: Awesome. :+1:

package.json
misteroneill: Your call. I was just curious, really. :smile:

misteroneill:
Absolutely. That's the biggest unknown there - Browserify configuration.

gkatsev:
Given that we still haven't pulled it in yet, I'm just going to close this. I'll keep the branch around, though.

PR #2833, opened 2015-11-20
fix for returning current souce rather than blob

bcvio:

PR #2678, opened 2015-10-06
Feature/current source

chemoish:
Original PR: https://github.com/videojs/video.js/pull/2479

PR #2957, opened 2015-12-29
clear out currentSource on loadstart

gkatsev:
Starting with [v5.2.2](https://github.com/videojs/video.js/blob/master/CHANGELOG.md#522-2015-11-23) we've been returning the [cached src](https://github.com/videojs/video.js/pull/2833) for source handlers. This is important for techs that end up using MSE so that we return an the original source rather than the blob url that gets set.
However, there are cases when, in the html5 tech, the underlying video element could get a new video set, bypassing the videojs API. We want to actually pick up this change to return the new sources.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -477,6 +477,11 @@ class Html5 extends Tech {
    * @method setSrc
    */
   setSrc(src) {
+    let loadstartlistener = () => this.currentSource_ = null;
+
+    this.off(this.el_, 'loadstart', loadstartlistener);
+    this.one(this.el_, 'loadstart', () => this.one(this.el_, 'loadstart', loadstartlistener));
+
     this.el_.src = src;
   }
 


src/js/tech/html5.js
```diff
@@ -477,6 +477,11 @@ class Html5 extends Tech {
    * @method setSrc
    */
   setSrc(src) {
+    let loadstartlistener = () => this.currentSource_ = null;
```
gkatsev: The main question: should we dispose the current source handler?

gkatsev:
Opening it up now for discussion. This should get tests but I'll wait off till we decide the specific functionality.
@videojs/core-committers opinions? Thanks.

gkatsev:
(Also, the direct issue that sprouted this has been worked around for now, so, it's not a huge hurry)

chemoish:
I haven't been following, but is this `currentSource` similar to https://github.com/videojs/video.js/pull/2678?

gkatsev:
@chemoish sort of, actually. The Source Handler interface in videojs (what videojs-contrib-hls uses) gives the source (`{src, type}`) obj to the source handler to work on. It'll then cache it locally for the life of that source handler. The issue is that we keep returning that cached value even when the underlying video src has been changed because it didn't go through the videojs API (we didn't know it changed).

chemoish:
I understand the error, but didn't quite follow if the existing `currentSource_` implementation (https://github.com/videojs/video.js/pull/2833) makes #2678 obsolete.

#2678 doesn't handle anything other than attempting to store unmodified sources for MSE reasons.

gkatsev:
Ah, I see. No, it doesn't necessarily make it obsolete. #2833 made it so that `currentSrc` would now return the cached source in the case of MSE. It only returns the source url and not the type as well.
So, with MSE, and #2833 you would now get `http://example.com/video.m3u8` instead of `blob://example.com` when calling `player.currentSrc()`.
#2678 probably needs to get updated to use `currentSource_`. That feature could still be useful since it returns both the type and source of the video. `currentSource_` is the source object (what you pass to `player.src()` that is given to Source Handlers.

src/js/tech/html5.js
dmlap: Sounds good to me.

gkatsev: Also, `loadstartlintener` should probably be moved to a private prototype method so that removing it works correctly on subsequent calls to setSrc.

PR #2960, opened 2015-12-30
Update docstrings for autoplay and loop methods

huitsing:
Fix small copy/paste error in docstrings


gkatsev:
Heh, nice catch. Thanks. LGTM.

PR #2850, opened 2015-11-24
Errors fixes

gkatsev:
* hide the error display properly when clearing out an error
* allow clearing an error from the an error event handler

PR #2966, opened 2016-01-04
Fix for opening errorDisplay more than once

forbesjo:
Error display can be closed with error(null) therefore it should not be disposed on close.

This fix is related to #2850


gkatsev:
Makes sense. LGTM.

PR #2971, opened 2016-01-05
Do not autoplay if there is no source

forbesjo:
This PR prevents the player from calling play() with autoplay and no source. In Firefox this would fire a 'play' event (See http://jsbin.com/vubugev/edit?html,console,output). For some reason this doesn't happen in Chrome.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -720,7 +720,7 @@ class Player extends Component {
     // In Safari (5.1.1), when we move the video element into the container div, autoplay doesn't work.
     // In Chrome (15), if you have autoplay + a poster + no controls, the video gets hidden (but audio plays)
     // This fixes both issues. Need to wait for API, so it updates displays correctly
-    if (this.tag && this.options_.autoplay && this.paused()) {
+    if (this.src() && this.tag && this.options_.autoplay && this.paused()) {
       delete this.tag.poster; // Chrome Fix. Fixed in Chrome v16.
       this.play();
     }


gkatsev:
We probably should make sure that when a source is then added it'll get autoplayed. Does this happen in your testing? We may or may not want to have a similar check inside of `player#src`

forbesjo:
setSrc() in the Flash tech does check for autoplay and calls play(). The HTML tech just lets the video element deal with it. I've checked and after I call src() the video does automatically play.

forbesjo:
@gkatsev wondering if you have any more comments on this one.

gkatsev:
LGTM.
We probably should have a story at some point about fixing autoplay completely.

jamesgorrie:
The problem with this seems to be that if I have specified `<source src="../thing.m3u8" type="video/m3u8">` `player.src()` is then `""` (or `!`). Is this a bug when setting `src`? If so I can file it.

gkatsev:
`video/m3u8` isn't a correct mimetype for `m3u8`/HLS content.

jamesgorrie:
Apologies (And thanks for the heads up).

We are supplying multiple sources, and even when only serving the mp4 content we don't see the `player.src()` populated. Only when we add the `src` attribute to the `video` element itself.

e.g. of video element:
__Not working__
```
<video class="vjs-tech" preload="auto" aria-hidden="true">
  <source src="https://example.com/example.mp4" type="video/mp4">                
</video>
```

__Working__
```
<video class="vjs-tech" preload="auto" aria-hidden="true" src="https://example.com/example.mp4">
  <source src="https://example.com/example.mp4" type="video/mp4">                
</video>
```

gkatsev:
Can you open a separate issue for this? Thanks.

gkatsev:
Also, if you can include a working test case on a site like jsfiddle, that would be best.

jamesgorrie:
Okay - after quite a bit of digging, it appears it has to do with the fact that, if a page is loaded, we have a `currentSrc` set.

[This then doesn't run the `setSrc` function as the `current.src === this.el_.src`](https://github.com/videojs/video.js/blob/8f2bc92412c4b95641cf271faa5eadb3207c4147/src/js/tech/html5.js#L37)

So either this PR should include:
```JS
if ((this.src() || this.currentSrc()) && this.tag && this.options_.autoplay && this.paused()) {
//...
}
```

Or the `setSrc` method should be updated.

[Here is an example](https://jsbin.com/hohogu/2/edit?html,js,output) where autoplay won't work when we have `currentSrc` set from the browser.

jamesgorrie:
Created an issue here #3071

PR #2973, opened 2016-01-05
update videojs-font

gkatsev:


dmlap:
LGTM

misteroneill:
LGTM

gkatsev:
Actually, I need to add the wrapcodepoints task here as well :/

misteroneill:
LGTM, round 2

gkatsev:
Ok, so, I updated this PR based on https://github.com/videojs/font/pull/12

--------------
 Generate css using the codepoint-fixed sass files

The regular exported sass files will keep using the normal sass files
but users can use the new endpoint of `vjs.scss` which will use the
codepoint-fixed sass. They'll also need to manually wrap the codepoints
on their generated css file after the fact.

gkatsev:
The reason tests are failing is because it depends on videojs/font#12 to be merged and released.

misteroneill:
LGTM, third time's a charm!

PR #2984, opened 2016-01-08
Update CDN URL in setup guide

mister-ben:
The setup guide has broken links to non-existent 4.12 files. This updates the URL to 5.4.6.

Longer term, what's the best way to keep the markdown docs updated?


gkatsev:
LGTM.

PR #2985, opened 2016-01-08
Change manual timeUpdate event timing function to be fired more often

### LINKED ISSUE #2979: TextTrack Fallback is not called often enough to have reliable subtitles
Hi,

I did some testing of the tracks API and it seems, on edge (browser where HTML5 track api is not used), that checking for subtitle to display is done only 1 to 3 times per second (it seems to use the video progress event).
Subtitles in a professionnal environment is by frame and so the polling is not done often enough to have a consistent subtitling.

One way to get a better result would be using RequestAnimationFrame or setInterval to get a consistent polling and timing.

To test this issue, burn subtitles in the video and play the video with VTT track added. You will see that subtitles in burn video and in WebVTT are not appearing at the same time (at least in Edge).

I'm willing to provide a pull request if someone can tell me where the change would have to be done (I browsed a bit through the code, could not find the place where the text for the subtitles was updated, and the exact way the event is triggered).

Naouak:
Timeupdate event now use requestAnimationFrame with fallback to setTimeout to get call more often and to permit subtitles to have a better synchronization.

Fix #2979

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -193,10 +193,28 @@ class Tech extends Component {
    * @method trackCurrentTime
    */
   trackCurrentTime() {
-    if (this.currentTimeInterval) { this.stopTrackingCurrentTime(); }
-    this.currentTimeInterval = this.setInterval(function(){
-      this.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
-    }, 250); // 42 = 24 fps // 250 is what Webkit uses // FF uses 15
+    this.manualTimeUpdatesActive = true;
+    this.manualTimeUpdateTicker();
+  }
+
+  /**
+   * Trigger timeupdate event if not supported by browser
+   *
+   * @method manualTimeUpdateTicker
+   */
+  manualTimeUpdateTicker() {
+    let requestAnimationFrame =
+        window.requestAnimationFrame ||
+        window.webkitRequestAnimationFrame ||
+        window.mozRequestAnimationFrame ||
+        function(timingFn) {
+          setTimeout(timingFn, 16);
+        };
+
+    this.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
+    if(this.manualTimeUpdatesActive){
+      requestAnimationFrame(Fn.bind(this, this.manualTimeUpdateTicker));
+    }
   }
 
   /**
@@ -205,11 +223,7 @@ class Tech extends Component {
    * @method stopTrackingCurrentTime
    */
   stopTrackingCurrentTime() {
-    this.clearInterval(this.currentTimeInterval);
-
-    // #1002 - if the video ends right before the next timeupdate would happen,
-    // the progress bar won't make it all the way to the end
-    this.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
+    this.manualTimeUpdatesActive = false;
   }
 
   /**
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -193,16 +193,16 @@ class Tech extends Component {
    * @method trackCurrentTime
    */
   trackCurrentTime() {
-    this.manualTimeUpdatesActive = true;
-    this.manualTimeUpdateTicker();
+    this.manualTimeUpdatesActive_ = true;
+    this.manualTimeUpdateTicker_();
   }
 
   /**
    * Trigger timeupdate event if not supported by browser
    *
-   * @method manualTimeUpdateTicker
+   * @method manualTimeUpdateTicker_
    */
-  manualTimeUpdateTicker() {
+  manualTimeUpdateTicker_() {
     let requestAnimationFrame =
         window.requestAnimationFrame ||
         window.webkitRequestAnimationFrame ||
@@ -212,8 +212,8 @@ class Tech extends Component {
         };
 
     this.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
-    if(this.manualTimeUpdatesActive){
-      requestAnimationFrame(Fn.bind(this, this.manualTimeUpdateTicker));
+    if(this.manualTimeUpdatesActive_) {
+      requestAnimationFrame(Fn.bind(this, this.manualTimeUpdateTicker_));
     }
   }
 
@@ -223,7 +223,7 @@ class Tech extends Component {
    * @method stopTrackingCurrentTime
    */
   stopTrackingCurrentTime() {
-    this.manualTimeUpdatesActive = false;
+    this.manualTimeUpdatesActive_ = false;
   }
 
   /**


src/js/tech/tech.js
```diff
@@ -193,10 +193,28 @@ class Tech extends Component {
...
+    this.manualTimeUpdatesActive = true;
+    this.manualTimeUpdateTicker();
+  }
+
+  /**
+   * Trigger timeupdate event if not supported by browser
+   *
+   * @method manualTimeUpdateTicker
+   */
+  manualTimeUpdateTicker() {
```
gkatsev: this should probably be a private function, so, post-fix an `_`.

src/js/tech/tech.js
```diff
@@ -193,10 +193,28 @@ class Tech extends Component {
...
+    let requestAnimationFrame =
+        window.requestAnimationFrame ||
+        window.webkitRequestAnimationFrame ||
+        window.mozRequestAnimationFrame ||
+        function(timingFn) {
+          setTimeout(timingFn, 16);
+        };
+
+    this.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
+    if(this.manualTimeUpdatesActive){
```
gkatsev: we put spaces after `if` and before the `{`.

gkatsev:
LGTM. I'll make sure to run it locally before merging it in because it's a relatively major change.

Naouak:
I've been using it on my local build for several weeks and did not encounter any issue with it on Firefox, Chrome, Safari and Edge on Windows 10, Mac and Android(I deactivated the native version of tracks for every browser apart from iOS on my build).

dmlap:
requestAnimationFrame() is very frequent and `timeupdate` is a very popular event to add handlers to. Without good evidence that increasing this frequency will not bog down slow machines, I don't think we should make this change.  The HTML standard warns implementors to monitor the runtime of `timeupdate` event handlers (see [this note](https://html.spec.whatwg.org/multipage/embedded-content.html#playing-the-media-resource:current-playback-position-20)) and throttle them if necessary. Without those kind of safeguards, this seems dangerous.

Naouak:
I understand your point @dmlap but I don't think it is an issue. 

`requestAnimationFrame` is not time based but performance based. You have to actively ask for next animation frame to get you callback called. If you have a handler of timeupdate event that is slow, you will fire timeupdate less often because you called requestAnimationFrame too late.

Basically, if your timeupdate handler is taking 250ms, you will get 4 call of requestAnimationFrame per second at most. If your timeupdate handler is fast, you will get as much as your screen refresh rate per second (60hz in most case). It is like the note you linked asked.

OwenEdwards:
@Naouak regarding the frequency of `timeupdate` and `requestAnimationFrame`, I understand your point that the `timeupdate` will not get fired any more often than its performance will allow (as you say, performance based rather than time based), and I think that's an elegant solution, but is there a danger that it will still consume all the JS processing, and starve any other processing?

We're getting into the dynamics of real-time systems here, and priortization - I don't have much experience of that with JS, but I think that's what the point that @dmlap linked to was making - in the case of overload, does the system handle it gracefully or grind to a halt?

Naouak:
@OwenEdwards I'm doing some test with low end computers in order to test that right now and it doesn't seems to be problematic.
For reference, I'm using that requestAnimationFrame method to draw fairly complex subtitles on a canvas with a core2duo machine with old version of safari. There is no apparent issue at all and timing is doing good.
I can try to add some computing to the event to check.

I believe you are already something very wrong somewhere if you do something that last several dozens of milliseconds on timeupdate, .

gkatsev:
Also, we're focusing on requestAnimationFrame right now but we're falling back to a setTimeout of 16ms if rAF isn't available. I think that would be much more of an issue, particularly on low end harder and also in older browsers like IE8.

Naouak:
I think setTimeout won't respect the 16 ms timing and will do at minimum a 40ms delay per spec.

OwenEdwards:
@Naouak absolutely agree, if there's so much added to the event handler that the processing takes too long then it's a bigger problem. The point is, this problem typically won't be spotted by developers or site maintainers (who often have decent hardware), but by a user who will have no idea why the video player becomes unresponsive. Video.js doesn't necessarily need to fix that situation, but it would be preferable to be able to detect it in a way that means there won't be reports of Video.js being broken, especially if it happens when a site upgrades to a newer version of vjs.

Naouak:
@OwenEdwards Then we are to choose whether subtitling should have a good experience or  developpers should be good enough to not overload an event.

If you are really afraid that this event will block the browser, I can add a timer on the requestAnimationFrame execution and drop some callback if the timeupdate event takes too much time.

I personally need this PR for subtitles as current state has too much delay for a consistent and good user experience. If we are afraid that `timeupdate` is too often used by plugins for heavy computation, then let's trigger it less often and pass subtitles timing events to another event (e.g. : `subtitleTimeUpdate`) to get them to refresh more often.

I guess we should also add some mention in documentation that `timeupdate` should not be used with long synchronous tasks.

gkatsev:
@Naouak btw, how does native subtitles end up working with your usecase? Wouldn't they also not change fast enough?

Naouak:
@gkatsev Native is no good on several browser unfortunately. I have bug reports for Edge, Safari and [Chromium](https://code.google.com/p/chromium/issues/detail?id=576310) for timing issues. Ironically, the best support for subtitle as of today is done by Firefox (and it is perfect!) but videojs seems to have issue with it as the tech deactivate native support for firefox.

iOS is the worst offender with several seconds latency when there is a lot of cues. It makes subtitles unusable for movies in Safari iOS.

To be frank, I rewrite the values of featureNativeTextTracks on load to avoid using native track API today. I have this code before loading a player:

````javascript
var Html5 = videojs.getComponent('Html5');
Html5.prototype.featuresNativeTextTracks = (function(){
    var USER_AGENT = window.navigator.userAgent;
    var IS_IPHONE = (/iPhone/i).test(USER_AGENT);
    var IS_IPAD = (/iPad/i).test(USER_AGENT);
    var IS_IPOD = (/iPod/i).test(USER_AGENT);
    return IS_IPHONE || IS_IPAD || IS_IPOD;
}());
````

I have good results using the code in this PR on every platform I could try it on (Windows, Mac OS, Chrome and Firefox for Android). 95%+ of my users are using subtitles, I need them to be well synced (and burning them in the video is not a solution unfortunately).

**Comment by gkatsev** [comment]
We actually have an override for it:
```js
let player = videojs('foo', {
  html5: {
    nativeTextTracks: false
  }
});
```
Which would be the recommended approach. But how do you deal with the fact that on iphones you go to native fullscreen or on ipad that when a user goes fullscreen. Wouldn't they then lose subtitles?

Naouak:
Yes, my team plans to not make HTML5 player available on iOS and redirects users to our iOS application.

iOS has a lot of issues with subtitling that we can't support even if timing was good enough.

gkatsev:
So, I went back to read what the spec says about this matter. In the best practices section on metadata tracks, they [recommend using `cuechange` rather than `timeupdate`](https://html.spec.whatwg.org/multipage/embedded-content.html#best-practices-for-metadata-text-tracks:event-media-cuechange). Also, reading the [time marches on algorithn](https://html.spec.whatwg.org/multipage/embedded-content.html#time-marches-on) that runs when the media element's playback is changing makes it seem that `cuechange` can occur more frequently than `timeupdate` and that `timeupdate` should be throttled to 4-66Hz.
What this means for this PR? It means that we cannot accept it as is. However, if it were updated to have the text tracks use the requestAnimationFrame system for updating activeCues and triggering `cuechange` directly instead of the `timeupdate`, we could accept it.

gkatsev:
If someone can update this PR (or create a new one) that updates the emulated text track support to use requestAnimationFrame (with a fallback to setTimeout for older browsers) that would be awesome.

PR #2990, opened 2016-01-10
When options are passed via HTML booleans are strings

gaurav21r:
This is how I am invoking the Video Player

```HTML
<div class="video-background">
    <video id="main-video-player" data-setup='{ "controlBar":"false", "bigPlayButton": "false" }'>
      <source src="/media/{{video.videoLocation}}" type="video/mp4">
    </video>
</div>
```

```js
$(function () {
    videojs('main-video-player', {}, function () {
    ...
    })
})
```

But I get a couple of errors. A quick breakpoint in Chrome Dev tools reveal:

videojs

Based on http://stackoverflow.com/questions/263965/how-can-i-convert-a-string-to-boolean-in-javascript I thought this would be the best way to check for `"false"` as well. 

Let me know if this is good, or there's a better way to *parse* this `"false"`. I have a feeling there may be more such issues across the VideoJS code. Especially where we are setting options via HTML attributes and not JS.

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -517,8 +517,8 @@ class Component {
         }
 
         // Allow for disabling default components
-        // e.g. options['children']['posterImage'] = false
-        if (opts === false) {
+        // e.g. options['children']['posterImage'] = false or "false" (if set via HTML)
+        if ((opts === false) || (opts === "false")) {
           return;
         }
 


gkatsev:
Thanks for the PR. Unfortunately, we won't be accepting it. The `data-setup` attribute is meant to be given properly encoded JSON. That means that boolean attributes should not be quotes but kept unquoted. Updating this code to check for string values will open us up to a lot more issues in the long run.

PR #2991, opened 2016-01-11
Optimize active cues lookup time for HTML track fallback

Naouak:
Add an optimisation option to cue list. This optimisation permit a 90% performance increase when looking up for new cues to display on a large file.

Amended test for optimize version.

I did some test with a 2049 cues track file on my configuration.
With old method, each active cues lookup took a total more than 4 ms.
With this method, lookup take less than 0.5ms.

Here is a comparison of performances on Edge using both method. New method is obviously on the left, old method on the right.
image

Optimized version is a flag on object creation to avoid performances issue on startup and when creating activeCues list.

### DIFF
--- a/src/js/tracks/text-track-cue-list.js
+++ b/src/js/tracks/text-track-cue-list.js
@@ -14,8 +14,9 @@ import document from 'global/document';
  * };
  */
 
-let TextTrackCueList = function(cues) {
+let TextTrackCueList = function(cues, optimized) {
   let list = this;
+  this.optimized_ = optimized || false;
 
   if (browser.IS_IE8) {
     list = document.createElement('custom');
@@ -41,12 +42,45 @@ let TextTrackCueList = function(cues) {
 };
 
 TextTrackCueList.prototype.setCues_ = function(cues) {
-  let oldLength = this.length || 0;
   let i = 0;
   let l = cues.length;
 
-  this.cues_ = cues;
-  this.length_ = cues.length;
+  this.cues_ = [];
+  this.length_ = 0;
+  if (this.optimized_) {
+    this.quickCues_ = [];
+  }
+
+  for(i; i < l; i++){
+    this.addCue_(cues[i]);
+  }
+
+  this.length_ = this.cues_.length;
+};
+
+TextTrackCueList.prototype.optimizeAllCues_ = function() {
+  let cues = this.cues_;
+  this.quickCues_ = [];
+
+  for (let i = 0; i < cues.length; i++) {
+    this.optimizeCue_(cues[i]);
+  }
+};
+
+TextTrackCueList.prototype.optimizeCue_ = function(cue){
+  if(this.optimized_){
+    for (let j = cue['startTime']; j < cue['endTime'] + 100; j += 100) {
+      this.quickCues_[j / 100 | 0] = this.quickCues_[j / 100 | 0] || [];
+      this.quickCues_[j / 100 | 0].push(cue);
+    }
+  }
+};
+
+TextTrackCueList.prototype.addCue_ = function(cue) {
+  this.cues_.push(cue);
+  if(this.optimized_) {
+    this.optimizeCue_(cue);
+  }
 
   let defineProp = function(i) {
     if (!(''+i in this)) {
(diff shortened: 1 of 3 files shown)


src/js/tracks/text-track-cue-list.js
```diff
@@ -41,12 +42,45 @@ let TextTrackCueList = function(cues) {
...
+TextTrackCueList.prototype.optimizeAllCues_ = function() {
+  let cues = this.cues_;
+  this.quickCues_ = [];
+
+  for (let i = 0; i < cues.length; i++) {
+    this.optimizeCue_(cues[i]);
+  }
+};
+
+TextTrackCueList.prototype.optimizeCue_ = function(cue){
```
gkatsev: I don't understand what this function does. Can you write a doc comment for it?

src/js/tracks/text-track-cue-list.js
```diff
@@ -57,14 +91,47 @@ TextTrackCueList.prototype.setCues_ = function(cues) {
...
+    this.length_ = this.cues_.length;
+    if(this.optimized_){
+      this.optimizeAllCues_();
+    }
+  }
+};
 
-  if (oldLength < l) {
-    i = oldLength;
+TextTrackCueList.prototype.getActiveCuesByTime = function(time) {
```
gkatsev: we shouldn't change the public interface of this class.

gkatsev:
Made a few comments. There are some stylistic issues as well but we can get those updated after all the functionality is good and we're on the same page as to what this PR does.

Naouak:
I fixed some issues you reported.
For the change of public API, I can pass the function as private and use the private in `TextTrack` like in `TextTrack.prototype.addCue` and `TextTrack.prototype.removeCue` but it is a violation of visibility.

What is better, violating private or adding public APIs ?

gkatsev:
Inside the videojs codebase, we can use "private" methods, the convention is really more like "package protected". The point is that the public API to users of videojs should not change from the native text track api. Not all videojs users will be using the emulated text tracks which you're fixing.

src/js/tracks/text-track-cue-list.js
```diff
@@ -57,14 +106,47 @@ TextTrackCueList.prototype.setCues_ = function(cues) {
...
-    i = oldLength;
+  if (removed) {
+    this.length_ = this.cues_.length;
+    if(this.optimized_){
+      this.optimizeAllCues_();
+    }
+  }
+};
+
+TextTrackCueList.prototype.getActiveCuesByTime = function(time) {
```
gkatsev: Can you write a quick doc for this method as well? Also, postfix it with a `_`?

src/js/tracks/text-track-cue-list.js
```diff
@@ -57,14 +106,47 @@ TextTrackCueList.prototype.setCues_ = function(cues) {
...
+  let cues = this.quickCues_[time/100|0]||[];
+  let active = [];
 
-    for(; i < l; i++) {
-      defineProp.call(this, i);
+  for (let i = 0, l = cues.length; i < l; i++) {
+    let cue = cues[i];
```
gkatsev: is it possible that a cue was added to the track not ended up not being optimized so it doesn't show up? Is it even a case we should worry about?

Naouak: I mended `addCues_` and `removeCue_` functions checking for optimization to avoid that. 

`addCue_` was created to avoid optimizing too often when a WebVTT is parsed.
Before this PR, when a cue was added, every cues were pulled from the track and added again. This took a lot of time when there is a huge number of cues and optimization took also a lot longer.

Unless someone mess directly with the cue table it should not have any issue.

I will update my PR with documentation about `getActiveCuesByTime_` tomorrow.

src/js/tracks/text-track-cue-list.js
```diff
@@ -57,14 +106,47 @@ TextTrackCueList.prototype.setCues_ = function(cues) {
...
+};
+
+TextTrackCueList.prototype.removeCue_ = function(removeCue) {
+  let removed = false;
+
+  for (let i = 0, l = this.cues_.length; i < l; i++) {
+    let cue = this.cues_[i];
+    if (cue === removeCue) {
+      this.cues_.splice(i, 1);
+      removed = true;
```
gkatsev: should we break here? I don't think we're supposed to be able to have multiple of the same cue in the array.

src/js/tracks/text-track-cue-list.js
gkatsev: Sounds good.
Thanks.

gkatsev:
A few last comments:

1. I think a test for `getActiveCuesByTime_` is necessary, I do not think our tests for `activeCues` are sufficient to test this method. It can either be tested directly or via `activeCues`.
1. style
  1. our style guide asks for a space after block statement keyword, i.e., `for` and `if`.
  1. our style guide asks for whitespace around operators like `|` or `||` or `+` or `,`.

Unfortunately, we haven't had a chance to update our linter with these styles.

Thanks.

Naouak:
You are right for activeCues tests. I'm not well versed in TDD so I think I will need a bit of guidance for that part.

What test should there be apart from  "for several given time, are cues correct" ?

gkatsev:
That probably would cover most of the tests necessary. Some other stuff to try out is trying to get the activeCues/getCuesByTime when cues is empty.

gkatsev:
A lot has changed recently in tracks (because of the audio and video track work), it would be great if you could rebase (or merge) master and fix any issues. Sorry for this and thanks for the PR.

Naouak:
Unfortunately, I changed job and I can't legally work on this code anymore
without a written authorization from my company because it relates to stuff
I work on (French work's laws are a bit strange).

The committed code is legally cleared, I just can't work on it anymore.

PR #3002, opened 2016-01-14
Remove unused tracks from track list when changing source

### LINKED ISSUE #3000: Changing source does not clear tracks
It seems that if I dynamically change the source of a video.js instance, it does not clear the text tracks.

If on an HLS-supported browser, this can easily be seen by doing the following javascript on videojs.com:

```javascript
var hls = "http://media.streamonecloud.net/hls/account=6xINlI/item=HZ8kWupWhRqR/HZ8kWupWhRqR.m3u8";
var player = videojs.getPlayers()["preview-player"];
player.src(hls);
setTimeout(function() {
    player.src(hls);
    player.play();
}, 1000);
```

If I manually call `player.textTracks().removeTrack_` for all tracks it seems to work, but that seems like a private method.

This seems to be the case for both native and emulated text tracks as far as I could find.

nickygerritsen:
This should fix #3000 . When using native text tracks it removes unused ones on `loadstart`.

@gkatsev suggested using `inbandmetadatatrack` but that seems to be `""` for my HLS subtitle tracks so I guess we can not use it.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -94,6 +94,9 @@ class Html5 extends Tech {
         tl.addEventListener('change', Fn.bind(this, this[`handle${capitalType}TrackChange_`]));
         tl.addEventListener('addtrack', Fn.bind(this, this[`handle${capitalType}TrackAdd_`]));
         tl.addEventListener('removetrack', Fn.bind(this, this[`handle${capitalType}TrackRemove_`]));
+
+        // Remove (native) trackts that are not used anymore
+        this.on('loadstart', this[`removeOld${capitalType}Tracks_`]);
       }
     });
 
@@ -106,7 +109,6 @@ class Html5 extends Tech {
       this.handleTextTrackChange_ = Fn.bind(this, this.handleTextTrackChange);
       this.handleTextTrackAdd_ = Fn.bind(this, this.handleTextTrackAdd);
       this.handleTextTrackRemove_ = Fn.bind(this, this.handleTextTrackRemove);
-      this.removeOldTextTracks_ = Fn.bind(this, this.removeOldTextTracks);
       this.proxyNativeTextTracks_();
     }
 
@@ -141,8 +143,8 @@ class Html5 extends Tech {
       }
       
       // Stop removing old text tracks
-      if (type === 'text' && tl) {
-        this.off('loadstart', this.removeOldTextTracks_);
+      if (tl) {
+        this.off('loadstart', this[`removeOld${capitalType}Tracks_`]);
       }
     });
 
@@ -362,12 +364,17 @@ class Html5 extends Tech {
     this.audioTracks().removeTrack_(e.track);
   }
 
-  removeOldTextTracks() {
-    // This will loop over the texttracks and check if they are still used
+  /**
+   * This is a helper function that is used in removeOldTextTracks_, removeOldAudioTracks_ and
+   * removeOldVideoTracks_
+   * @param {Track[]} techTracks Tracks for this tech
+   * @param {Track[]} elTracks Tracks for the HTML5 video element
+   * @private
+   */
+  removeOldTracks_(techTracks, elTracks) {
+    // This will loop over the techTracks and check if they are still used by the HTML5 video element
     // If not, they will be removed from the emulated list
     let removeTracks = [];
-    const techTracks = this.textTracks();
-    const elTracks = this.el().textTracks;
     if (!elTracks) {
       return;
     }
@@ -390,8 +397,26 @@ class Html5 extends Tech {
 
     for (let i = 0; i < removeTracks.length; i++) {
       const track = removeTracks[i];
-      this.textTracks().removeTrack_(track);
+      techTracks.removeTrack_(track);
     }
+  }
+
+  removeOldTextTracks_() {
+    const techTracks = this.textTracks();
+    const elTracks = this.el().textTracks;
+    this.removeOldTracks_(techTracks, elTracks);
+  }
+
+  removeOldAudioTracks_() {
+    const techTracks = this.audioTracks();
+    const elTracks = this.el().audioTracks;
+    this.removeOldTracks_(techTracks, elTracks);
+  }
+
+  removeOldVideoTracks_() {
+    const techTracks = this.videoTracks();
+    const elTracks = this.el().videoTracks;
+    this.removeOldTracks_(techTracks, elTracks);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -282,6 +293,33 @@ class Html5 extends Tech {
     this.textTracks().removeTrack_(e.track);
   }
 
+  removeOldTextTracks() {
```
gkatsev: Make sure you consistently call this `removeOldTextTracks_`.

src/js/tech/html5.js
```diff
@@ -74,6 +74,7 @@ class Html5 extends Tech {
       this.handleTextTrackChange_ = Fn.bind(this, this.handleTextTrackChange);
       this.handleTextTrackAdd_ = Fn.bind(this, this.handleTextTrackAdd);
       this.handleTextTrackRemove_ = Fn.bind(this, this.handleTextTrackRemove);
+      this.removeOldTextTracks_ = Fn.bind(this, this.removeOldTextTracks);
```
gkatsev: missing the `_`.

src/js/tech/html5.js
```diff
@@ -282,6 +293,33 @@ class Html5 extends Tech {
...
+          break;
+        }
+      }
+
+      if (!found) {
+        removeTracks.push(techTrack);
+      }
+    }
+
+    for (let i = 0; i < removeTracks.length; i++) {
```
gkatsev:
```js
removeTracks.forEach((track) => this.textTracks().removeTrack_(track))
```
But really doesn't have to change.

src/js/tech/html5.js
```diff
@@ -282,6 +293,33 @@ class Html5 extends Tech {
     this.textTracks().removeTrack_(e.track);
   }
 
+  removeOldTextTracks() {
+    // This will loop over the texttracks and check if they are still used
+    // If not, they will be removed from the emulated list
+    let removeTracks = [];
+
+    for (let i = 0; i < this.textTracks().length; i++) {
+      let techTrack = this.textTracks()[i];
```
gkatsev: we probably should cache `this.textTracks()` in a variable outside the forloops so we don't have to make the look up each time.

src/js/tech/html5.js
```diff
@@ -282,6 +293,33 @@ class Html5 extends Tech {
...
+
+      let found = false;
+      for (let j = 0; j < this.el().textTracks.length; j++) {
+        if (this.el().textTracks[j] === techTrack) {
```
gkatsev: we should also probably cache a reference to the native text track and possibly also do a nullcheck around it as well. Since it is conceivably possible to have an html5 browser that doesn't have text track support.

src/js/tech/html5.js
```diff
@@ -282,6 +293,33 @@ class Html5 extends Tech {
     this.textTracks().removeTrack_(e.track);
   }
 
+  removeOldTextTracks() {
+    // This will loop over the texttracks and check if they are still used
+    // If not, they will be removed from the emulated list
+    let removeTracks = [];
+
+    for (let i = 0; i < this.textTracks().length; i++) {
```
gkatsev:
```js
let tt = this.textTracks();
let trackIndexOf = Array.prototype.indexOf.call.bind(this.el().textTracks);
Array.prototype.filter.call(tt,
                            (ourTrack) => trackIndexOf(ourTracks) === -1)
               .forEach((trackToRemove) => tt.removeTrack_(trackToRemove));
```
Why? Just felt like it. No change necessary.

gkatsev:
This needs to be rebased plus there's a few minor comments that need to be addressed.

nickygerritsen:
Totally, completely forgot about this PR. I'll add some replies above so we can tackle this one.

src/js/tech/html5.js
nickygerritsen: Hmmm I kinda copied this from the `handle...` functions. Why are they not postfixed with `_` but this one should be? As far as I know they actually should all be underscored?

src/js/tech/html5.js
nickygerritsen: Why is this `Fn.bind` dance here anyway btw? I'm wondering if it is needed for `removeOldTextTracks_`?

src/js/tech/html5.js
nickygerritsen: You mean like

```js
const textTracks = this.textTracks();
```
?

src/js/tech/html5.js
nickygerritsen: And with this you mean caching `this.el().textTracks` in some variable / constant and check if `this.el().textTracks` is `null`?

src/js/tech/html5.js
nickygerritsen: I kinda do like the other short version, but this one isn't really readable anymore :yum:

src/js/tech/html5.js
gkatsev: Oh I see, I must've missed that, I guess carry on...

src/js/tech/html5.js
gkatsev: yep

src/js/tech/html5.js
gkatsev: yep

nickygerritsen:
I assume we do not have to remove video/audio tracks, as these are never native?

gkatsev:
I think we might need to since we do listen to the addtrack on the native audio and video tracks as well. Thoughts, @brandonocasey?

BrandonOCasey:
I think Safari has support for native audio and video tracks so I think we will want to.

nickygerritsen:
OK then I'll try to mimic your way of doing this for all three track kinds.

nickygerritsen:
OK I think this does the trick. I removed all the `Fn.bind` goodness, as `Events.on` already does this. this also gets rid of a "strange" dance with the underscored and non-underscored version of the function.

gkatsev:
LGTM

src/js/tech/html5.js
```diff
@@ -94,6 +94,9 @@ class Html5 extends Tech {
         tl.addEventListener('change', Fn.bind(this, this[`handle${capitalType}TrackChange_`]));
         tl.addEventListener('addtrack', Fn.bind(this, this[`handle${capitalType}TrackAdd_`]));
         tl.addEventListener('removetrack', Fn.bind(this, this[`handle${capitalType}TrackRemove_`]));
+
+        // Remove (native) trackts that are not used anymore
```
gkatsev: oops, just realized this comment has `tracks` misspelt.

PR #3003, opened 2016-01-14
Make $primary-foreground-color a default var

mister-ben:
Add `!default` to make `$primary-foreground-color` overridable.


gkatsev:
LGTM.

PR #3006, opened 2016-01-15
return already created player when given an element

gkatsev:
As part of the tech 2.0 work, we decided to remote the player out of the tech. One of the changes in it [nulled out `tag.player`](https://github.com/videojs/video.js/blob/b48797bf1788f4207a2f03e8980c53a1b6a05a55/src/js/player.js#L590-L594). This is an issue because `videojs(el)` uses `tag.player` to see whether we have already created, so, if you try to get the player again with the same `el`, it'll actually create a brand new player for you.
An example:
```js
let el = document.querySelector('#my-video');
console.log(el.id); // "my-video"
let player = videojs(el);
console.log(el.id); // "my-video_html5_api"
let anotherPlayer = videojs(el);
console.log(el.id); // "my-video_html5_api_html5_api"
console.log(player == anotherPlayer); // false
```
The expected result should be something like this:
```js
let el = document.querySelector('#my-video');
console.log(el.id); // "my-video"
let player = videojs(el);
console.log(el.id); // "my-video_html5_api"
let anotherPlayer = videojs(el);
console.log(el.id); // "my-video_html5_api"
console.log(player == anotherPlayer); // true
```
There are two ways I can think off right now to fix this:
1. revert the changes that nulled out `tag.player`
2. when given an element and `tag.player` in null, we would want to parse out the id to remove the stuff like `html5_api` and then return `videojs.players[parsedId(tag.id)]` if it exists, otherwise actually go ahead and create a player

I feel like solution number 2 is more correct but it is also a lot more complicated and more prone to issues. Included below right now are two failing tests for the above usecase.
@videojs/core-committers @dconnolly thoughts?

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -269,6 +269,7 @@ class Player extends Component {
     // Update tag id/class for use as HTML5 playback tech
     // Might think we should do this after embedding in container so .vjs-tech class
     // doesn't flash 100% width/height, but class only applies with .video-js parent
+    tag.playerId = tag.id;
     tag.id += '_html5_api';
     tag.className = 'vjs-tech';
 
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -52,8 +52,8 @@ if (typeof HTMLVideoElement === 'undefined') {
  * @mixes videojs
  * @method videojs
  */
-var videojs = function(id, options, ready){
-  var tag; // Element of ID
+let videojs = function(id, options, ready){
+  let tag; // Element of ID
 
   // Allow for element or ID to be passed in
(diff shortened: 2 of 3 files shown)


heff:
Since the video tag is the only tech that doubles as an embed code, you could look at it from the embed code side and add tag.player in the player init, just for this purpose.

gkatsev:
In a conversation with @heff, we came to a simpler solution than 2 without resorting to 1. Instead, we're going to attach the `playerId` to the tag and use that to get back player references.

heff:
lgtm!

PR #3007, opened 2016-01-15
Update year range to 2016

pra85:


gkatsev:
Thanks. We've previously already updated our license to not include the year so, I'm going to close this and update the README to not include the year either.

pra85:
:+1:

PR #3017, opened 2016-01-22
Fix double-localization of the control text of the mute-toggle button.

OwenEdwards:
Since control text localization is entirely hidden inside the `controlText` method, the check for whether the mute button's control text has changed does not need to do localization.


gkatsev:
LGTM

Issue #2156, opened 2015-05-16
Keyboard accessibility of control bar after it is hidden during playback


OwenEdwards:
If a user uses tab and enter/space to play the video, focus shifts onto the control bar and the user can continue to use keyboard-only access.

However, once inactivity causes the controls to be hidden, the keyboard cannot be used to restore focus to the controls, so the video player is no longer accessible.

1Copenut:
This will be a big step forward. As soon as the video player loses focus, or the controls disappear from inactivity, it's no longer keyboard accessible, which is limiting. Users benefit from a way to scrub backwards via keyboard, toggle captions, play and pause video. For technical discussions, or even ones that are in-depth or dialog heavy, being able to scrub back, review, pause, playback are critical for learning and comprehension.

gkatsev:
We definitely should fix this. We probably need to listen to keyboard events and make the user active when that happens. I think we only make the user active on mouse movement. If someone cares to make a PR, it'll make it that much more likely to get completed earlier rather than later.
Thanks!

1Copenut:
@gkatsev Agreed. It might be useful to listen for the player object focus event as a trigger to show and (possibly) announce the control bar using aria-hidden="false" attribute on the control bar.

I'm just getting my JS leveled up, but I want to help. If that's testing, writing documentation, reading specs, so be it.

marguinbc:
Not sure  how keen anyone is on using another CSS class but I created a workaround that checks if any of the children of the control-bar have focus and applies a class called hasFocus to the control-bar. The hasFocus class simply prevents the transition and transform which occurs when the player looses focus.

    $('.vjs-control-bar > *').focus(function() {
      $('.vjs-control-bar').addClass('hasFocus');
    });
 
    $('.vjs-control-bar > *').focusout(function(e) {
      setTimeout(function () {
        if ($(':focus').closest('.vjs-control-bar').length <=0) {
          $('.vjs-control-bar').removeClass('hasFocus');
         }
      });
    });

and the CSS

.not-hover.vjs-has-started.vjs-paused.vjs-user-active .vjs-control-bar.hasFocus, 
 .not-hover.vjs-has-started.vjs-paused.vjs-user-inactive .vjs-control-bar.hasFocus, 
  .not-hover.vjs-has-started.vjs-playing.vjs-user-active .vjs-control-bar.hasFocus, 
[...]

PR #3023, opened 2016-01-22
Apply aria-hidden to control-bar when userinactive

gkatsev:
Defaults to aria-hidden:true because the controls are hidden by default.

The next step here is to try and prevent the control bar from hiding when it has keyboard focus.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2256,6 +2256,11 @@ class Player extends Component {
           this.addClass('vjs-user-active');
           this.trigger('useractive');
         } else {
+          // if we are focused anywhere in the player, don't go inactive
+          if (player.el().contains(document.activeElement)) {
+            return;
+          }
+
           // We're switching the state to inactive manually, so erase any other
           // activity
           this.userActivity_ = false;
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2257,7 +2257,7 @@ class Player extends Component {
           this.trigger('useractive');
         } else {
           // if we are focused anywhere in the player, don't go inactive
-          if (player.el().contains(document.activeElement)) {
+          if (this.el().contains(document.activeElement)) {
             return;
           }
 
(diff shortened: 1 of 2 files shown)


gkatsev:
And got it to not go inactive when focused inside the player. That was surprisingly easy.
Though, I do wonder whether we want to do that or not. I guess that we could make sure that when `<Esc>` is pressed, we would blur out of the player so it can go inactive or something.
@OwenEdwards thoughts?

src/js/player.js
```diff
@@ -2256,6 +2256,11 @@ class Player extends Component {
           this.addClass('vjs-user-active');
           this.trigger('useractive');
         } else {
+          // if we are focused anywhere in the player, don't go inactive
+          if (player.el().contains(document.activeElement)) {
```
gkatsev: maybe this should be `player.controlBar.el().contains(document.activeElement)`?
And then `<Esc>` would focus the player div?

src/js/player.js
```diff
@@ -2256,6 +2256,11 @@ class Player extends Component {
           this.addClass('vjs-user-active');
           this.trigger('useractive');
         } else {
+          // if we are focused anywhere in the player, don't go inactive
+          if (this.el().contains(document.activeElement)) {
```
gkatsev: maybe this should be `this.controlBar.el().contains(document.activeElement)`?
And then `<Esc>` would focus the player div?

gkatsev:
This should help with #2156 but it doesn't currently fix the trying to tab back into the player part of that issue.

gkatsev:
Talking this over with @OwenEdwards, we realized we want to do basically the opposite of what this PR does. When the controls go inactive, it should only be visual. The controls should still work for Screen Readers. I'll close this and approach this problem from that direction.

OwenEdwards:
@gkatsev "still work for Screen Readers" __and keyboard-only users__. I'm going to keep hammering on that point, that these are two different groups of users (although they experience related accessibility issues). Honestly, you'll miss me when I don't need to mention it again ;-)

gkatsev:
Yeah, it didn't feel right to not go inactive on focus but didn't understand exactly where I was going wrong. :D

PR #3032, opened 2016-01-25
Fix keyboard (Space or Enter) operation of buttons. Fixes #2888. Supercedes #2889

### LINKED ISSUE #2888: Using Space key on the play/pause button in Firefox does not work

OwenEdwards:
Better fix than #2889 for the problem with keyboard operation of buttons which was introduced by changing buttons to `<button>` elements, but leaving other clickable elements (menu buttons, poster image) as `<div>`s. By splitting them in different objects, this allows future changes to build on these two distinct bases.


gkatsev:
LGTM

src/js/clickable-component.js
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * @method handleKeyPress
+   */
+  handleKeyPress(event) {
+    // Support Space (32) or Enter (13) key operation to fire a click event
+    if (event.which === 32 || event.which === 13) {
+      event.preventDefault();
+      this.handleClick(event);
+    } else {
+      super.handleKeyPress(event); // Pass keypress handling up for unsupported keys
```
gkatsev: This ends up trying to call `Component`'s `handleKeyPress`, which doesn't exist.

src/js/button.js
```diff
@@ -13,122 +14,77 @@ import assign from 'object.assign';
...
   createEl(tag='button', props={}, attributes={}) {
     props = assign({
-      className: this.buildCSSClass(),
-      tabIndex: 0
+      className: this.buildCSSClass()
     }, props);
 
-    // Add standard Aria info
+    if (tag !== 'button') {
+      log.warn(`Creating a Button with an HTML element of ${tag} is deprecated; use ClickableComponent instead.`);
```
gkatsev: :+1:

src/js/clickable-component.js
gkatsev: We probably want the else to be:
```js
} else if (super.handleKeyPress) {
  super.handleKeyPress(event);
}
```

test/unit/clickable-component.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+  var player, testClickableComponent, el;
+
+  player = TestHelpers.makePlayer({
+  });
+
+  testClickableComponent = new ClickableComponent(player);
+  el = testClickableComponent.createEl();
+
+  ok(el.nodeName.toLowerCase().match('div'));
```
gkatsev: would be better to use `equal` here.
```js
equal(el.nodeName.toLowerCase(), 'div', 'the name of the element is "div"');
```

PR #3033, opened 2016-01-25
Fix menu keyboard access and ARIA labeling for screen readers (Fixes #2155 and #2769), supercedes #2897

### LINKED ISSUE #2155: Keyboard accessibility of control bar menus (e.g. captions, subtitles, etc)

OwenEdwards:
Building on #3032, this fixes keyboard operation of Menus without changing the DOM.

### DIFF
--- a/src/js/control-bar/text-track-controls/off-text-track-menu-item.js
+++ b/src/js/control-bar/text-track-controls/off-text-track-menu-item.js
@@ -25,6 +25,9 @@ class OffTextTrackMenuItem extends TextTrackMenuItem {
       'mode': 'disabled'
     };
 
+    // MenuItem is selectable
+    options['selectable'] = true;
+
     super(player, options);
     this.selected(true);
   }
--- a/src/js/control-bar/control-bar.js
+++ b/src/js/control-bar/control-bar.js
@@ -38,6 +38,8 @@ class ControlBar extends Component {
   createEl() {
     return super.createEl('div', {
       className: 'vjs-control-bar'
+    }, {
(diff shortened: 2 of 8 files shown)


src/js/control-bar/text-track-controls/off-text-track-menu-item.js
```diff
@@ -25,6 +25,9 @@ class OffTextTrackMenuItem extends TextTrackMenuItem {
       'mode': 'disabled'
     };
 
+    // MenuItem is selectable
+    options['selectable'] = true;
```
gkatsev: `options.selectable` should work fine here.

PR #3045, opened 2016-01-27
added "auto" attribute value support in width or height of the player

defli:
Currently player width or height getter (player.prototype.width()) is return 0 unless width or height attribute hasn't set as "auto" at video tag or hasn't at all. 

But some plugins ex. google ima uses width getter, so if user is using responsive design result will be false positive.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -341,10 +341,19 @@ class Player extends Component {
     let privDimension = dimension + '_';
 
     if (value === undefined) {
-      return this[privDimension] || 0;
+      // if attr of width or height is "auto",
+      // return computed width or height for responsiveness
+      let computedWidthOrHeight;
+      if (dimension === 'width') {
+        computedWidthOrHeight = this.el_.offsetWidth;
+      } else {
+        computedWidthOrHeight = this.el_.offsetHeight;
+      }
+
+      return this[privDimension] || computedWidthOrHeight || 0;
     }
 
-    if (value === '') {
+    if (value === '' || value === 'auto') {
       // If an empty string is given, reset the dimension to be automatic
       this[privDimension] = undefined;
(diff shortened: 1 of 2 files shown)


alidavut:
it's urgent for me as well :+1:

gkatsev:
Thanks for the PR, however, this is not a backwards compatible change so we can't pull it in as a minor version update. But there's more to consider.
We've had discussions about this previously and came to the conclusion that `width()` and `height()` should continue to mimic the `width` and `height` attributes of the native `video` element like we do with all our other methods like `currentTime()` and `duration()`. The native attributes on the `video` element return what size the video was configured with.
We've recently gotten a lot of issues related to this so I think adding other methods, perhaps `currentWidth()` and `currentHeight()` would be acceptable. These new methods should use `getComputedStyle` (and `el.currentStyle` for IE8).
As a workaround, it is recommended to use `getComputedStyle` on the player element to get the correct size of the player:
```js
var computedStyle = getComputedStyle(player.el());
var width = computedStyle.width;
var height = computedStyle.height;
```

defli:
Thanks for fast response. In my opinion, currentWidth() and currentHeight() would be nice. However there is bug in google ima plugin, may be I can update that plugin.

gkatsev:
Yeah, you should PR a fix for them using the `getComputedStyle`.

gkatsev:
Also, if you want to PR the new methods, that would be great too :D

defli:
Yes, I can implement new methods, is there a roadmap? Which steps should i follow? I have read the contribution manuel.

gkatsev:
@defli I'll write up an issue with the details about it and ping you when it's done.

defli:
@gkatsev thanks.

PR #3046, opened 2016-01-27
When removing a track also remove listeners

forbesjo:
This PR is to remove event listeners from removed text tracks. This would be useful for a plugin like videojs-playlist.

### DIFF
--- a/src/js/tracks/text-track-list.js
+++ b/src/js/tracks/text-track-list.js
@@ -111,7 +111,9 @@ TextTrackList.prototype.removeTrack_ = function(rtrack) {
   for (let i = 0, l = this.length; i < l; i++) {
     if (this[i] === rtrack) {
       track = this[i];
-      track.off();
+      if (track.off) {
+        track.off();
+      }
 
       this.tracks_.splice(i, 1);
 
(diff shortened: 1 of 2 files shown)


src/js/tracks/text-track-list.js
```diff
@@ -111,6 +111,7 @@ TextTrackList.prototype.removeTrack_ = function(rtrack) {
   for (let i = 0, l = this.length; i < l; i++) {
     if (this[i] === rtrack) {
       track = this[i];
+      track.off();
```
gkatsev: See the note in the doc for this method :)
The track in this list could be a native track element which won't have an `off` method. Can probably check `track instanceof TextTrack` or something.
However, because users are required to manage their listeners with native tracks, I wonder whether we should do this for the emulated tracks?

forbesjo: It looks like one of the tests caught that scenario. I'm checking for track.off existence since it seems like the input is still a TextTrack even though it's an element.

src/js/tracks/text-track-list.js
```diff
@@ -111,6 +111,9 @@ TextTrackList.prototype.removeTrack_ = function(rtrack) {
   for (let i = 0, l = this.length; i < l; i++) {
     if (this[i] === rtrack) {
       track = this[i];
+      if (track.off) {
```
gkatsev: This works. Sorry, what I meant was checking to see if it's an `instanceof` `videojs.TextTrack`.

forbesjo:
@gkatsev any other changes needed?

gkatsev:
LGTM. I'll work on getting these in this week.

PR #3053, opened 2016-01-29
Update Umuted to Unmute in lang files

mister-ben:
The lang files all have "Unmuted" as a key, but the [mute control uses "Unmute"](https://github.com/videojs/video.js/blob/master/src/js/control-bar/mute-toggle.js#L78).


gkatsev:
LGTM, though, is this the better approach or would changing the mute control's text be better?

mister-ben:
I think "unmute" makes more sense. Clicking unmutes.

"Unmuted" reads like a description of the current state, but in that case should be "muted".

gkatsev:
LGTM.

PR #3054, opened 2016-01-29
add fieldsets and labels to text-track-settings

marguinbc:
Multiple structural changes for accessibility
- Added fieldset and legends around colors and font settings
- Increased height to 20em to accommodate fieldset and legends
- Added css for fieldset and legends
- setDefault actually sets the default vs using {} null
- Removed --- as default option, replace with what we are actually setting as defaults, improved accessibility as screen readers now read, background color, black, etc.
- Added labels for primary comboboxes for colors, fonts
- Revised tests to check that new defaults are set, restored, etc.

Remaining tasks:
Properly label opacity fields for colors combo boxes
Fix one test I had problems with in 'should restore default settings'

### DIFF
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -6,7 +6,7 @@
   color: $primary-foreground-color;
   margin: 0 auto;
   padding: 0.5em;
-  height: 15em;
+  height: 20em;
   font-size: 12px;
   width: 40em;
 }
@@ -67,6 +67,24 @@
   margin-right: 10px;
 }
 
+.vjs-caption-settings .hiddenReadable {
+  position:absolute;
+  left: -10000px;
+  top: auto;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;  
+}
+
+.vjs-caption-settings fieldset {
+  margin-top: 1em;
+  margin-left: .5em;
+}
+
+.vjs-caption-settings h2 {
+  margin-bottom: 0;
+  margin-left: .5em;
+}
 .vjs-caption-settings input[type="button"] {
   width: 40px;
   height: 40px;
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -37,7 +37,7 @@ class TextTrackSettings extends Component {
       this.$('.vjs-bg-color > select').selectedIndex = 0;
       this.$('.window-color > select').selectedIndex = 0;
       this.$('.vjs-text-opacity > select').selectedIndex = 0;
-      this.$('.vjs-bg-opacity > select').selectedIndex = 0;
+      this.$('.vjs-bg-opacity > select').selectedIndex = 1;
       this.$('.vjs-window-opacity > select').selectedIndex = 0;
       this.$('.vjs-edge-style select').selectedIndex = 0;
       this.$('.vjs-font-family select').selectedIndex = 0;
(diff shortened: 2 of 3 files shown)


src/js/tracks/text-track-settings.js
```diff
@@ -235,113 +235,112 @@ function setSelectedOption(target, value) {
...
+                <option value="#00F">Blue</option>
+                <option value="#FF0">Yellow</option>
+                <option value="#F0F">Magenta</option>
+                <option value="#0FF">Cyan</option>
+              </select>
+              <span class="vjs-bg-opacity vjs-opacity">
                 <select>
-                  <option value="">---</option>
                   <option value="1">Opaque</option>
+                  <option value="0.8" selected>80% Opaque</option>
```
gkatsev: is adding another bg opacity necessary? Would it be better to change the value for `semi-transparent` instead?
The FCC rules for this call for `opaque`, `semi-transparent`, and `transparent` options.
@OwenEdwards does WCAG or a11y have anything to say about these values?

src/js/tracks/text-track-settings.js
```diff
@@ -235,113 +235,112 @@ function setSelectedOption(target, value) {
...
-              <select>
-                <option value="">---</option>
-                <option value="1">Opaque</option>
-                <option value="0.5">Semi-Opaque</option>
+         <div class="vjs-tracksettings-colors">
+          <fieldset>
+            <legend>Colors</legend>
+            <div class="vjs-fg-color vjs-tracksetting">
+              <label class="vjs-label" for="captions-foreground">Foreground</label>
+              <select id="captions-foreground">
```
gkatsev: all ids have to be unique which won't be the case if you load multiple players on a page.
To fix this, we probably want to do something like:
```js
// at the top of the page import guid
import guid from '../util/guid.js';

// then above the `let template`
let captionsForegroundId = `cpations-foreground${guid()}`;

// and there:
`<label for=${captionsForegroundId}>Foreground</label>
<select id=${captionsForegroundId}>...</select>
`
```

src/js/tracks/text-track-settings.js
OwenEdwards: @gkatsev this is an FCC requirement (FCC 12-9), not a WCAG requirement. Specifically, it calls for:

> four settings, opaque (100% opacity), semi-transparent (at 75% or 25% opacity), and transparent (0% opacity)

for `background` and for `window`, and three for `font` (hmm, actually transparent font is essentially what we need for description tracks! That gives me an idea!)

So yes, we need to add another `semi-transparent`. Whether we need to change the values (to 0.75 and 0.25) is something to discuss, too.

src/css/components/_captions-settings.scss
```diff
@@ -67,6 +67,24 @@
   margin-right: 10px;
 }
 
+.vjs-caption-settings .hiddenReadable {
```
OwenEdwards: `hiddenReadable` doesn't seem to be used in this PR. Can it be dropped?

src/js/tracks/text-track-settings.js
gkatsev: I'd vote to change the current semi-transparent from 50% to 75% and not add another option.

OwenEdwards:
Oh, @marguinbc, you actually dropped some functionality here I think. It wasn't clear, but the `---` values didn't denote `use the defaults`, they denoted `use the values in the WebVTT track, don't override them`. We need to work out a clearer way to put that in the settings dialog - maybe just `None`?

src/js/tracks/text-track-settings.js
OwenEdwards: @gkatsev looking at the `guid()` method, wouldn't a simpler approach be to use the `player.tag.id` (which is unique for each player instance) as part of the IDs in the Caption Player Dialog? It seems like the guid() is going to be unique for each element within the player, which is fine for how it's used within `dom.js`, but perhaps overkill here?

Otherwise, has there been any discussion about having a global `player.uniqueId`?

OwenEdwards: Oh, I answered my own question - each element has a `this.id_`, but `captionOptionsMenuTemplate()` is a static function, not an object method. So we would need to pass it a unique ID, at which point it makes little difference if it is `this.id_`, `this.player_.id()` or `Guid.newGUID()`.

OwenEdwards:
Hmm, @marguinbc @gkatsev I realize now that this logic around the `---` setting is essentially the wrong way round; WebVTT tracks usually have _no_ font settings, in which case they should use the video.js settings, and `---` is meaningless. If the text track does have font settings (color, background, etc.) then the option should be whether to override the track's setting with the user's preference or not. So it's really a user setting and an override checkbox. Does that make sense?

src/css/components/_captions-settings.scss
marguinbc: Sure thing.  Was planning on using them for labels that could be read by screen readers but not displayed but haven't added any to the form.

gkatsev:
@OwenEdwards well, there is the `reset/defaults` button which can unset user overrides but what if I just want to override the background color but not the other options? The `---` setting was previously used to denote the "don't do anything" options.

marguinbc:
What my intent was is since we have a 'selected' for each field, these sort of act as the default. So, I modified setDefault to use the same values.  

I also think from an accessibility standpoint, it makes sense for the selected value to be the currently set one and for screen readers to be able to read, foreground color, white...  

Doing so I did not account for the case where the user changed the default, saved them, went in to make another change, and decided NOT to apply that change after all. Reset/default would wipe out their previous changes and Done would apply the most recent changes.  Not sure how best to resolve that instance.

gkatsev:
Closing this one as #3281 supercedes it. Thanks for starting the ball rolling on this one @marguinbc!

PR #3060, opened 2016-02-01
Improve UX of current time tooltip

incompl:


src/css/components/_progress.scss
```diff
@@ -86,12 +86,12 @@
...
   display: none;
   position: absolute;
-  top: -2.4em;
+  top: -3.5em;
   right: -1.5em;
   font-size: 0.9em;
   color: #000;
   content: attr(data-current-time);
-  padding: 0.2em 0.5em;
+  padding: 6px 8px 8px 8px;
```
gkatsev: Shouldn't it have equal padding on top/bottom?

incompl: It is visually equal with this CSS. Probably due to the font.

gkatsev: I'm seeing it larger on the bottom.

incompl: What browser? This is what I'm seeing in Chrome:

screen shot 2016-02-01 at 2 53 58 pm

gkatsev: Chrome. I'm definitely seeing it as 6px and 8px.

gkatsev: So, after looking careful at it with @incompl, the computed style does actually lie here. What was causing as issue is the slight overlap of the control bar which caused it to look wrong to me. This is fixed by the latest commit to the PR.

dmlap:
LGTM

incompl:
Looks like this has two approvers and is ready to merge @dmlap @gkatsev

PR #3063, opened 2016-02-02
Enhance modularity of text track settings

OwenEdwards:
It seems much more logical to move all of the text track settings and styling into one component (`TextTrackSettings`), and leave `TextTrackDisplay` to do a fairly standard timing and display function. This way, future changes for text tracks (such as adding support for displaying several tracks, including a description track and a captions track) will be more modular.

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -9,21 +9,6 @@ import * as Fn from '../utils/fn.js';
 import document from 'global/document';
 import window from 'global/window';
 
-const darkGray = '#222';
-const lightGray = '#ccc';
-const fontMap = {
-  monospace:             'monospace',
-  sansSerif:             'sans-serif',
-  serif:                 'serif',
-  monospaceSansSerif:    '"Andale Mono", "Lucida Console", monospace',
-  monospaceSerif:        '"Courier New", monospace',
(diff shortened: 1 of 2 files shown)


gkatsev:
Maybe instead of of moving the styling into TextTrackSettings we put it in another file that we can import statically inside of TextTrackDisplay?

gkatsev:
my feeling is that `TextTrackSettings` is in charge of displaying the dialog to the user and also possibly caching this value between player loads. If we move the element styling portion from the `TextTrackDisplay` into `TextTrackSettings`, we're now having the `TextTrackSettings` be in charge of two separate tasks. The `TextTrackDisplay` is supposed to be in charge of displaying the cues, so, styling them is sort of in it's ball park as well. However, I do agree it will make the `TextTrackDisplay` much nicer if it delegated the actual styling to something else. The other thing shouldn't be the `TextTrackSettings`, I don't think. The styling function could still do a check to see whether `TextTrackSettings` are included in the player and be a no-op if it isn't.

OwenEdwards:
After discussion with @gkatsev, it seems like this isn't the way we're going to go right now with handling Captions Settings.

OwenEdwards:
Discussed this with @gkatsev. There's definitely room for improvement in the Captions Settings mechanism, but this isn't the right direction at this point.

Issue #3015, opened 2016-01-20
Caption track labels are always lowercase


thomaspeeters:
Because the caption menu is wrapped in a `div` with the `.vjs-menu` class, all caption labels have the `text-transform: lowercase;` rule applied to them.

I can fix this with some custom CSS like this:

```CSS
.vjs-captions-button .vjs-menu li
{
    text-transform: none;
}
```

However, I feel a cleaner solution should be possible, this shouldn't be the default behaviour for caption labels.

gkatsev:
I think we're going to call this a stylistic choice. Plus, it should be fairly easy to fix this.
```css
.vjs-menu li {
  text-transform: initial;
}
```

PR #3068, opened 2016-02-02
Remove text-transform: lowercase from .vjs-menu li

mcbanderson:
Removed text-transform: lowercase from .vjs-menu li in order to fix issue #3015


gkatsev:
@mmcc can you review this? Thanks.

gkatsev:
I think we're going to call this a stylistic choice. Plus, it should be fairly easy to fix this.
```css
.vjs-menu li {
  text-transform: initial;
}
```

PR #3077, opened 2016-02-04
Fix iPhone detection

misteroneill:
Because the Facebook iOS app's UIWebView identifies itself as both an iPad _and_ an iPhone, we need to exclude iPads from detection of iPhones.


gkatsev:
LGTM.

PR #3080, opened 2016-02-04
fixed reference to videojs-vtt dependency

alex-phillips:
The dependency for the VTT package has not been updated to reflect what is in `package.json`. Building the project then leaves out the VTT code.


gkatsev:
Good call. Thanks.

misteroneill:
LGTM

PR #3083, opened 2016-02-04
Android live display fix for HLS VOD

### LINKED ISSUE #3079: LiveDisplay initially shown on Android Chrome for VOD HLS
On Android Chrome (47, 48 and 49(dev) tested), `duration()` is initially `Infintity` for VOD HLS videos. As a result the live display is shown before playback begins. Depending on the network conditions/cache state it can be a second or longer from tapping play until a `durationchange` is fired with a finite time to show the non-live controls.

On iOS Safari and Chrome and the Android/Samsung stock browser, the duration is not `Infinity`.

Example: http://output.jsbin.com/kasawoh

mister-ben:
This fixes #3079 where Android Chrome shows the live display initially for VOD HLS videos. With this change the live display will not be added until after the video has progressed past 0 and the duration is still `Infinity`.

A couple of caveats:

- It uses user agent tests `browser.IS_ANDROID` and `browser.IS_CHROME` which probably isn't ideal
- For live HLS videos, the non-live controls will now be shown until playback starts. So the opposite issue, but I do think this is much less annoying than the flash of "LIVE" on a non-live video.

This also removes the LiveDisplay component's own duration checks to show/hide itself, and just uses CSS. This brings it into line with other components and avoids making the other changes in two places.

### DIFF
--- a/src/css/components/_live.scss
+++ b/src/css/components/_live.scss
@@ -1,10 +1,14 @@
 // We are assuming there is no progress bar and using the live display
 // to fill in the middle space. Live+DVR will need to adjust this.
 .video-js .vjs-live-control {
-  @include display-flex(flex-start);
-  @include flex(auto);
+  display: none;
   font-size: 1em;
   line-height: 3em;
+}
+
+.video-js.vjs-live .vjs-live-control {
+  @include display-flex(flex-start);
+  @include flex(auto);
 }
 
 .vjs-no-flex .vjs-live-control {
--- a/src/css/components/_live.scss
+++ b/src/css/components/_live.scss
@@ -1,14 +1,10 @@
 // We are assuming there is no progress bar and using the live display
 // to fill in the middle space. Live+DVR will need to adjust this.
 .video-js .vjs-live-control {
-  display: none;
-  font-size: 1em;
-  line-height: 3em;
-}
-
-.video-js.vjs-live .vjs-live-control {
   @include display-flex(flex-start);
   @include flex(auto);
+  font-size: 1em;
+  line-height: 3em;
 }
 
 .vjs-no-flex .vjs-live-control {
--- a/src/js/control-bar/live-display.js
+++ b/src/js/control-bar/live-display.js
@@ -3,6 +3,7 @@
  */
 import Component from '../component';
 import * as Dom from '../utils/dom.js';
+import * as browser from '../utils/browser.js';
 
 /**
  * Displays the live indicator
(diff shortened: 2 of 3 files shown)


gkatsev:
I'm not sure that the way this PR is currently is backwards compatible.

src/css/components/_live.scss
```diff
@@ -1,12 +1,16 @@
 // We are assuming there is no progress bar and using the live display
 // to fill in the middle space. Live+DVR will need to adjust this.
 .video-js .vjs-live-control {
-  @include display-flex(flex-start);
-  @include flex(auto);
+  display: none;
```
gkatsev: specifically, I'm not sure this change is backwards compatible.

mister-ben:
OK. I can revert the CSS and have the component controls its own visibility as before, it just means updating the logic in both player.js and live-display.js.

mister-ben:
Revisiting this as the flash of the live display is annoying. I've reverted the CSS so as not to introduce a breaking change. Now, the live checks in both player.js and live-display wait for a `timeupdate` where `currentTime() > 0` and `duration === Infinity` before showing the live display on Android Chrome.

One way to avoid the duplication in the player and the component would be for the player to trigger an 'live status change' event. Would that be a better appoach?

incompl:
How about a new `isLive` method? In [`videojs-contrib-ads` we have our own `isLive`](https://github.com/videojs/videojs-contrib-ads/blob/master/src/videojs.ads.js#L457-L464) and it's broken for the same reason you're making this fix, and I think we'll need to port the same currentTime() > 0 logic over there. Would be nice to get that in vjs proper and we could both use it. Presumably other people could use it too.

incompl:
Talking to @dmlap we think the best approach is to have the HTML5 tech duration implementation pave over this Android issue by not reporting Infinity when not live.

mister-ben:
I've given that a shot in #3476

tim1mw:
This issue appears to have regressed in the latest video.js with the HLS plugin, I've got two Android devices running Chrome which always play the video as live, so you can't seek. Firefox (on Linux and Android) shows the normal time bar, as does Chrome/Linux. I've put a copy of the a stream which demonstrates the problem here:

[http://www.my-place.org.uk/hls-test-stream/clip.m3u8](url)

The video clip comes from an webcam cam which is currently under test, video only no sound.

mister-ben:
That stream is showing as VOD for me on Android Chrome 61. You mention the HLS plugin, but I'm not able to test that stream using videojs-contrib-hls and overrideNative since the server does not add CORS headers. It would better to create a new issue with a complete reduced test case with information on exact versions of Video.js/browsers/plugins you're using.

Issue #2867, opened 2015-12-01
vjs-styles-defaults override defaults and cause page reflow


hartman:
I'm using videojs to progressively enhance my video elements. Unfortunately this is causing significant page reflows, which will impact the render speed of my pages. A requirement for these problems is to  asynchronously load your javascript.

* Problem one is that the video-js containers switch my players from inline-block to block. This causes page reflows for inlined elements. I can reset this in my own CSS, but I shouldn't have to.
* Second problem is vertical-alignment: top, which causes a redraw for inline elements. I can reset this in my own CSS, but I shouldn't have to.
* Third problem is the default size. This has higher specificity then my elements attributes. This is a problem, because the time by which my CSS is loaded and the defaults get applied to \<video class="video-js"> is significantly before the player setup 'reads' and sets my correct width and height on the individual players. I have made a [video](https://youtu.be/DlGLpaKmWng) of this problem.

This last one is really annoying to workaround. I cannot reset these defaults to 'initial' with pure CSS, because the per player classes with the correct w/h (which are added much later during player initialization), wouldn't have high enough specificity anymore. So really the only way to fix this is by using an inline style attribute and then after videojs setup remove that style attribute, which is... less then desirable.

There has to be a better way to deal with this... I'm reserving this space in my canvas for a reason (exactly to prevent reflows of the layout). To have the wrapping div ignore my original wish seems stupid. It this only to accommodate those who forget to reserve canvas area using the HTML attributes? In that case, why not set these fallback values as the last step of the players initialization, in case nothing else has set a width/height.

hartman:
Rolling a private patch for now to counter this:
https://phabricator.wikimedia.org/diffusion/ETMH/browse/master/patches/videojs.defaults.patch

gkatsev:
Going to close this a duplicate of #3051 just because it currently has the action item. It should be fixed as part of that.

PR #3093, opened 2016-02-09
Don't add style elements if VIDEOJS_NO_DYNAMIC_STYLE is set to true

### LINKED ISSUE #3051: How can i prevent video.js from putting styles in head

gkatsev:
This fixes #3051 (and #2867) by making sure that there's a way to make videojs doesn't add style elements to the DOM. This means that it will not have any default styles like default size dimensions. It also means that if the player is configured with dimensions, they will not be applied.
```html
<video width=600 height=300></video>
```
That will end up looking like (assuming the videojs css is still included):

And `player.width()` and `player.height()` will return `600` and `300` respectively.

TODO:
* [x] Update skins guide with this information as well as the fact that `base_styles.js` isn't currently wired up
* [x] Add tests to make sure this is working correctly
* [x] Add tests to make sure that videojs doesn't accidentally add *any* style elements to the DOM when `VIDEOJS_NO_BASE_THEME` is set to `true`

### DIFF
--- a/docs/guides/skins.md
+++ b/docs/guides/skins.md
@@ -1,6 +1,16 @@
 Skins
 =====
 
+## Base Skin
+The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)),
+and by default these styles are added to the dom for you!
+That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
+the styles you'd like to change.
+
+If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
+Keep in mind that without these base styles enabled, you'll need to manually include them.
+Video.js does not currently include the base skin automatically yet, so, this option isn't necessary.
+
 ## Default style elements
 Video.js uses a couple of dynamically, specifically, there's a default styles element as well as a player dimensions style element.
 They are used to provide extra default flexiblity with styling the player. However, in a lot of cases, users do not want this.
@@ -11,15 +21,9 @@ For example, the following player will end up having a width and height of 0 whe
 <video width=600 height=300></video>
 ```
 
-## Base Skin
-The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)),
-and by default these styles are added to the dom for you!
-That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
-the styles you'd like to change.
-
-If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
-Keep in mind that without these base styles enabled, you'll need to manually include them.
-Video.js does not currently include the base skin automatically yet, so, this option isn't necessary.
+### `Player#width` and `Player#height`
+When `VIDEOJS_NO_DYNAMIC_STYLE` is set, `Player#width` and `Player#height` will apply any width and height
+that is set directly to the video element (or whatever element the current tech uses).
 
 
 ## Icons
--- a/docs/guides/skins.md
+++ b/docs/guides/skins.md
@@ -14,7 +14,7 @@ Video.js does not currently include the base skin automatically yet, so, this op
 
 ## Default style elements
 Video.js uses a couple of style elements dynamically, specifically, there's a default styles element as well as a player dimensions style element.
-They are used to provide extra default flexiblity with styling the player. However, in a lot of cases, users do not want this.
+They are used to provide extra default flexiblity with styling the player. However, in some cases, like if a user has the HEAD tag managed by React, users do not want this.
 When `window.VIDEOJS_NO_DYNAMIC_STYLE` is set to `true`, video.js will *not* include these element in the page.
 This means that default dimensions and configured player dimensions will *not* be applied.
 For example, the following player will end up having a width and height of 0 when initialized if `window.VIDEOJS_NO_DYNAMIC_STYLE === true`:
--- a/docs/guides/skins.md
+++ b/docs/guides/skins.md
@@ -3,22 +3,23 @@ Skins
 
 ## Base Skin
 The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)),
-and by default these styles are added to the dom for you!
+and by default these styles are added to the DOM for you!
 That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
 the styles you'd like to change.
 
 If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
 Keep in mind that without these base styles enabled, you'll need to manually include them.
+
 Video.js does not currently include the base skin automatically yet, so, this option isn't necessary.
 
 ## Default style elements
-Video.js uses a couple of dynamically, specifically, there's a default styles element as well as a player dimensions style element.
+Video.js uses a couple of style elements dynamically, specifically, there's a default styles element as well as a player dimensions style element.
 They are used to provide extra default flexiblity with styling the player. However, in a lot of cases, users do not want this.
-When `window.VIDEOJS_NO_DYNAMIC_STYLE` is set to `true`, videojs will *not* include these element in the page.
+When `window.VIDEOJS_NO_DYNAMIC_STYLE` is set to `true`, video.js will *not* include these element in the page.
 This means that default dimensions and configured player dimensions will *not* be applied.
 For example, the following player will end up having a width and height of 0 when initialized if `window.VIDEOJS_NO_DYNAMIC_STYLE === true`:
 ```html
-<video width=600 height=300></video>
+<video width="600" height="300"></video>
 ```
 
 ### `Player#width` and `Player#height`
--- a/docs/guides/skins.md
+++ b/docs/guides/skins.md
@@ -4,8 +4,18 @@ Skins
 The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)), and by default these styles are added to the dom for you! That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
 the styles you'd like to change.
 
-If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = false` before Video.js is loaded. Keep in mind that without these base styles
+If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded. Keep in mind that without these base styles
 enabled, you'll need to manually include them.
+
+## Default style elements
+Video.js uses a couple of dynamically, specifically, there's a default styles element as well as a player dimensions style element.
+They are used to provide extra default flexiblity with styling the player. However, in a lot of cases, users do not want this.
+When `window.VIDEOJS_NO_BASE_THEME` is set to `true`, videojs will *not* include these element in the page.
+This means that default dimensions and configured player dimensions will *not* be applied.
+For example, the following player will end up having a width and height of 0 when initialized if `window.VIDEOJS_NO_BASE_THEME === true`:
+```html
+<video width=600 height=300></video>
+```
 
 ## Icons
 
--- a/docs/guides/skins.md
+++ b/docs/guides/skins.md
@@ -1,6 +1,17 @@
 Skins
 =====
 
+## Default style elements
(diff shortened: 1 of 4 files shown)


gkatsev:
I'm not certain that there's a good way of doing the 3rd TODO task above.

docs/guides/skins.md
```diff
@@ -1,20 +1,33 @@
...
+and by default these styles are added to the dom for you!
+That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
 the styles you'd like to change.
 
-If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = false` before Video.js is loaded. Keep in mind that without these base styles
-enabled, you'll need to manually include them.
+If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
+Keep in mind that without these base styles enabled, you'll need to manually include them.
+
+## Default style elements
```
gkatsev: this section is the main addition. The rest is mostly to keep lines from getting too long in the editor.

docs/guides/skins.md
```diff
@@ -1,20 +1,33 @@
...
 
-The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)), and by default these styles are added to the dom for you! That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
+The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)),
+and by default these styles are added to the dom for you!
+That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
 the styles you'd like to change.
 
-If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = false` before Video.js is loaded. Keep in mind that without these base styles
-enabled, you'll need to manually include them.
+If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
```
gkatsev: I changed `window.VIDEOJS_NO_BASE_THEME = false` to `window.VIDEOJS_NO_BASE_THEME = true` because that's how it's implemented and that's what the name implies.

gkatsev:
As part of this, it means that `player.width()` and `player.height()` will become no-ops because they edit the stylesheet that no longer gets included, however, that doesn't seem correct. I believe these methods need to *still* work even if `VIDEOJS_NO_BASE_THEME` is set.
What do @videojs/core-committers think?
I believe it's possible to have it work and be css overridable, even if the videojs css is included anyway.

heff:
It seems like if a user is setting `VIDEOJS_NO_BASE_THEME` it's a pretty clear indicator that they're using CSS to style the player and wouldn't be relying on player.width/height, or are technical *enough* not to need those. So to me it doesn't seem worth any extra complexity in the code to support that use case (assuming there's extra complexity). It could just be documented and/or throw a warning if the two are used together. I feel like we've done a lot of work to hold up the contract that the player will work like the video element out of the box, and if someone sets `VIDEOJS_NO_BASE_THEME` then they're intentionally stepping outside that contract (and they should understand that). My 2c.

gkatsev:
@heff does that mean that we should have a different mechanism for telling videojs to not use add style elements?
If `player.width()` and `player.height()` are analogous to `video.width` and `video.height` it means that they must always work.

heff:
I'm saying that I think using `VIDEOJS_NO_BASE_THEME` **plus** `player.width()` is an unlikely use case and could be considered breaking out of the contract where `player.width()` must always work like `video.width`. VIDEOJS_NO_BASE_THEME == "taking player sizing into your own hands". If you feel more strongly that player.width() should still uphold that contract in this use case, that's 100% cool with me.

gkatsev:
@heff so, I think we should take a step back a second to talk about what we're trying to accomplish and what is happening. Hopefully, the following writeup will clear it up and we can discuss what action we should take.

# What is the scenario?
videojs does *not* currently add the theme to the page by itself, a user is *required* to include either their own styles or the videojs styles (or both).
Videojs adds a style element for default player dimensions (`.video-js { width: 300px; height: 150px; }`). I've tried to look exactly why we do this but I could not find it. I do know there was a reason.
videojs adds a style element to the HEAD for the current player dimensions. When a user calls `player.width()` it then updates that stylesheet.
# The Problem
In a lot of cases, these dynamically inserted style elements are causing issues.
They can cause an issue when a user is using virtual DOM (like in #3051) or it can cause an issue in IE where IE uses insertion order for breaking ties in selector specificity.
# The need
There needs to be a way to make videojs *not* insert any style element into the DOM. This is possibly tangential to making videojs not insert a whole theme (which it doesn't do right now anyway).
# Current solution
The current solution piggy backs on the previously proposed `VIDEOJS_NO_BASE_THEME` property to not insert these elements.
# Possible requirements
* It is possible to have videojs not insert any style elements dynamically into the page
* `player#width` and `player#height` continue to function

I think it makes sense to have `player#width` and `player#height` to not work when `VIDEOJS_NO_BASE_THEME` is set, so, this leads me to think that we need a *different* signal to tell videojs to just not have dynamic stylesheets.

heff:
Ha, yeah, this is a lot to wrap your head around. I can't say I grasp on the entire issue yet, but a separate signal seems like it could make sense.

Here's the earlier research that led to the default 300x150.
http://blog.heff.me/post/75620807175/default-video-element-sizing

yareckon:
Hey, would love to have this as separate flag.  We use the base styling for things like the progress bar, but constrain the player ourselves with wrappers.  The styles in head are requiring us to increase specificity in our code a couple of places where we would rather not.

gkatsev:
@videojs/core-committers I think we want to do this under a separate flag like `VIDEOS_NO_DYNAMIC_STYLE` or something. Thoughts? I'd like to finish this up this week, if possible.

misteroneill:
I think it makes a ton of sense to not piggy-back on `VIDEOJS_NO_BASE_THEME` and go with a new flag like `VIDEOS_NO_DYNAMIC_STYLE`. Would this new flag trigger video.js to use a more old-school "update the `el().style`" approach or simply not do any sizing?

I agree that `width()` and `height()` ought to work in _some_ capacity even with `VIDEOS_NO_DYNAMIC_STYLE`.

gkatsev:
I think that the `width` and `height` function should continue to work. I would want to experiment with maybe just setting the `width` and `height` attributes on the video element directly for when the dynamic style els are disabled.

gkatsev:
I think this is good to go for review by @heff and other @videojs/core-committers.
When `VIDEOJS_NO_DYNAMIC_STYLE` is set it will *not* add any dynamic style elements to the DOM. In addition, `player.width` and `player.height` will apply the size to the `tech`'s element as `width` and `height` attributes, so, assuming no videojs sizes are included, it will be sized properly.

docs/guides/skins.md
```diff
@@ -1,20 +1,40 @@
 Skins
 =====
 
-The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)), and by default these styles are added to the dom for you! That means you can build a custom skin by simply taking advantage of the cascading aspect of CSS and overriding
+## Base Skin
+The base Video.js skin is made using HTML and CSS (although we use the [Sass preprocessor](http://sass-lang.com)),
+and by default these styles are added to the dom for you!
```
misteroneill: I feel like "dom" should be capitalized since it's an acronym.

docs/guides/skins.md
```diff
@@ -1,20 +1,40 @@
...
 the styles you'd like to change.
 
-If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = false` before Video.js is loaded. Keep in mind that without these base styles
-enabled, you'll need to manually include them.
+If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
+Keep in mind that without these base styles enabled, you'll need to manually include them.
+Video.js does not currently include the base skin automatically yet, so, this option isn't necessary.
+
+## Default style elements
+Video.js uses a couple of dynamically, specifically, there's a default styles element as well as a player dimensions style element.
```
misteroneill: Seems like there's something missing after "dynamically".

docs/guides/skins.md
gkatsev: It should.

docs/guides/skins.md
```diff
@@ -1,20 +1,40 @@
...
+They are used to provide extra default flexiblity with styling the player. However, in a lot of cases, users do not want this.
+When `window.VIDEOJS_NO_DYNAMIC_STYLE` is set to `true`, videojs will *not* include these element in the page.
```
misteroneill: Should "videojs" be "Video.js" or is there another preferred formatting?

docs/guides/skins.md
```diff
@@ -1,20 +1,40 @@
...
+This means that default dimensions and configured player dimensions will *not* be applied.
+For example, the following player will end up having a width and height of 0 when initialized if `window.VIDEOJS_NO_DYNAMIC_STYLE === true`:
+```html
+<video width=600 height=300></video>
```
misteroneill: No quotes on attributes intentionally?

docs/guides/skins.md
gkatsev: probably should video.js.

src/js/player.js
```diff
@@ -423,6 +425,24 @@ class Player extends Component {
    * @method updateStyleEl_
    */
   updateStyleEl_() {
+    if (window.VIDEOJS_NO_DYNAMIC_STYLE === true) {
+      /* jshint ignore:start */
+      let {width_: width = this.options_.width, height_: height = this.options_.width} = this;
```
misteroneill: Should this second `this.options_.width` be `this.options_.height`?

gkatsev: Yes...

src/js/video.js
```diff
@@ -99,21 +100,23 @@ let videojs = function(id, options, ready){
...
-  style = stylesheet.createStyleElement('vjs-styles-defaults');
-  let head = Dom.$('head');
-  head.insertBefore(style, head.firstChild);
-  stylesheet.setTextContent(style, `
-    .video-js {
-      width: 300px;
-      height: 150px;
-    }
+if (window.VIDEOJS_NO_DYNAMIC_STYLE !== true) {
+  let style = Dom.$('.vjs-styles-defaults');
```
misteroneill: I think the linter will complain that these `let` statements are not followed by an empty line.

src/js/player.js
gkatsev: This is why we have code reviews :)

src/js/video.js
gkatsev: Yes, though, videojs itself doesn't use vjsstandard yet
```
>.>
<.<
```

src/js/player.js
heff: This syntax excludes a lot of contributors from understanding what's happening here. Took me a minute. I think @mmcc is still working on it. :)

heff:
This is a clever way to handle it. lgtm

src/js/player.js
mmcc: loading

gkatsev: For those to read this in the future, what's happening is that it sets two variables called `width` and `height` to be the values of `this.width_` and `this.height_`, they also have default values of `this.options_.width` and `.this.options_.height` if `this.width_` and `this.height_` aren't set or are falsey.
That last bit is actually why I'm going to change it because we need to account for the possibility that the user wanted to set the width to be `0`. Why? Not sure, but it should still work.

docs/guides/skins.md
gkatsev: quotes are optional. I'll add them in anyway.

gkatsev:
Alright, this is ready for final review @heff @videojs/core-committers

heff:
lgtm!

docs/guides/skins.md
```diff
@@ -3,22 +3,23 @@ Skins
...
 the styles you'd like to change.
 
 If you don't want Video.js to inject the base styles for you, you can disable it by setting `window.VIDEOJS_NO_BASE_THEME = true` before Video.js is loaded.
 Keep in mind that without these base styles enabled, you'll need to manually include them.
+
 Video.js does not currently include the base skin automatically yet, so, this option isn't necessary.
 
 ## Default style elements
-Video.js uses a couple of dynamically, specifically, there's a default styles element as well as a player dimensions style element.
+Video.js uses a couple of style elements dynamically, specifically, there's a default styles element as well as a player dimensions style element.
```
chemoish: `styles element` => `style element`?

#donthitme

gkatsev: Heh, I think both are technically correct though a bit ambiguous. I was going for ((default styles) element) as opposed to a (default (style element)).
Maybe I should just change it to `default styles style element`?

Issue #3085, opened 2016-02-05
Keep localisations up to date


mister-ben:
There are now a few strings that are not in all or most of the lang files. Some used with `localize()`:

- 'Play video' - Big play button
- 'Close' (or 'Close dialog') - Close button
- 'Modal Window' - Modal dialog
- 'This is a modal window' - Modal dialog
- 'This modal can be closed by pressing the Escape key or activating the close button.' - Modal dialog

Plus a couple more set with `controlText`, which the component should localise:

- `', opens ' + options['kind'] + ' settings dialog'` - Captions settings menu
- ', selected' - Menu item

Would the easiest approach be to keep an en.json up to date? It should be more obvious that other lang files are missing entries.

Also, "dialog", "modal window" and "modal" being used interchangeably. Would that be confusing to screen reader users?

gkatsev:
Having an `en.json` sounds like a great idea.
We should probably update all usages of modal dialog, dialog, modal, window, etc, to be consistent.

gkatsev:
I'm going to close this as it's been "fixed" by #3096.

PR #3096, opened 2016-02-09
Added en.json as localisation template

mister-ben:
Adds en.json as a localisation template.

If dialog/modal/window is to be standardised (#3085) and Close will change to Close Dialog (#2924) imminently this will need to be updated.


gkatsev:
LGTM, thanks.

PR #3101, opened 2016-02-10
Buffer time-display component values to avoid refreshing HTML too often

Naouak:
HTML reparsing is a costly action for the browser and these components need to be updated only once per second. To avoid getting updating too often the HTML, the HTML will only update if the input data change.

### DIFF
--- a/src/js/control-bar/time-controls/current-time-display.js
+++ b/src/js/control-bar/time-controls/current-time-display.js
@@ -55,7 +55,7 @@ class CurrentTimeDisplay extends Component {
     let time = (this.player_.scrubbing()) ? this.player_.getCache().currentTime : this.player_.currentTime();
     let localizedText = this.localize('Current Time');
     let formattedTime = formatTime(time, this.player_.duration());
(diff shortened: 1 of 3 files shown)


src/js/control-bar/time-controls/current-time-display.js
```diff
@@ -55,7 +55,10 @@ class CurrentTimeDisplay extends Component {
     let time = (this.player_.scrubbing()) ? this.player_.getCache().currentTime : this.player_.currentTime();
     let localizedText = this.localize('Current Time');
     let formattedTime = formatTime(time, this.player_.duration());
-    this.contentEl_.innerHTML = `<span class="vjs-control-text">${localizedText}</span> ${formattedTime}`;
+    if(formattedTime !== this.formattedTime_){
```
gkatsev: needs whitespace after `if` and before `{`, in `remaining-time-display` as well.

gkatsev:
Thanks. This makes sense to me. Just needs a minor stylistic update.

dmlap:
LGTM

Issue #3070, opened 2016-02-03
TextTrackList doesn't work in IE8 with minified version.


steplerbox:
TextTrackList constructor: https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-list.js#L22
In bundled version constructor function is turning from function expression to named function expression:
`var TextTrackList = function TextTrackList(tracks)`
And In minified version it's getting different name:
`n = function p(a) {`
In IE8 p.prototype is not n.prototype, so IE8 hook does not work properly.

gkatsev:
[nooooooo](http://www.nooooooooooooooo.com/)

Also, could possibly be related to #3064.

gkatsev:
Closed by a2f5430

PR #3104, opened 2016-02-11
fix minified videojs in ie8

### LINKED ISSUE #3064: video.min.js errors in IE8, must use video.js
Something's busted about video.min.js since it breaks IE8 support. Could someone look into this or if you aren't going to fix it, at least update the docs with an html comment in the source code so people know

Use the below to see the error, change video.min.js to video.js and see how it works fine

&lt;!DOCTYPE html&gt;
&lt;html lang="en"&gt;
&lt;head&gt;
	&lt;link href="http://vjs.zencdn.net/5.5.3/video-js.css" rel="stylesheet"&gt;
	&lt;script src="http://vjs.zencdn.net/ie8/1.1.1/videojs-ie8.min.js" &gt;&lt;/script&gt;
	&lt;!-- must use video.js, IE8 errors if you use video.min.js --&gt;
	&lt;script src="http://vjs.zencdn.net/5.5.3/video.min.js" &gt;&lt;/script&gt;
&lt;/head&gt;

&lt;body&gt;
  &lt;video controls="controls" class="video-js vjs-fluid vjs-default-skin vjs-big-play-centered" preload="auto" data-setup='{}'&gt;
    &lt;source src="http://clips.vorwaerts-gmbh.de/VfE_html5.mp4" type="video/mp4"&gt;
    &lt;p class="vjs-no-js"&gt;
      To view this video please enable JavaScript, and consider upgrading to a web browser that
      &lt;a href="http://videojs.com/html5-video-support/" target="_blank"&gt;supports HTML5 video&lt;/a&gt;
    &lt;/p&gt;
  &lt;/video&gt;
&lt;/body&gt;
&lt;/html&gt;

gkatsev:
The issue here is that babel was turning function expressions like `let Foo = function() {}`
in names function expressions like `var Foo = function Foo() {}`. Afterwards, the minifier came through and cause these two variables to diverge causing issues in IE8. The fix is to switch to function statements.
Fixes #3064 and #3070.

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -149,6 +149,10 @@ module.exports = function(grunt) {
     },
     dist: {},
     watch: {
+      minify: {
+        files: ['build/temp/video.js'],
+        tasks: ['uglify']
+      },
       skin: {
         files: ['src/css/**/*'],
         tasks: ['sass', 'wrapcodepoints']
--- a/src/js/tracks/text-track-cue-list.js
+++ b/src/js/tracks/text-track-cue-list.js
@@ -14,7 +14,7 @@ import document from 'global/document';
  * };
  */
 
-let TextTrackCueList = function(cues) {
+function TextTrackCueList (cues) {
   let list = this;
 
   if (browser.IS_IE8) {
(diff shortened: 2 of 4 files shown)


gkatsev:
This is only a problem because of the special handling we do for tracks in IE8, specifically: https://github.com/videojs/video.js/blob/084bcf692ced2848973fe8b0c4fad6d745231b4c/src/js/tracks/text-track.js#L44-L52

misteroneill:
LGTM then

PR #3118, opened 2016-02-17
Add alternate CSS as video-js-cdn.css

misteroneill:
This stylesheet will automatically load the VideoJS EOT font from CDN for browsers that need it (old IE).

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -192,7 +192,7 @@ module.exports = function(grunt) {
       minify: {
         expand: true,
         cwd: 'build/temp/',
-        src: ['video-js.css'],
+        src: ['video-js.css', 'alt/video-js-cdn.css'],
         dest: 'build/temp/',
         ext: '.min.css'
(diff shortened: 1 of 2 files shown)


build/grunt.js
```diff
@@ -192,7 +192,7 @@ module.exports = function(grunt) {
       minify: {
         expand: true,
         cwd: 'build/temp/',
-        src: ['video-js.css'],
+        src: ['video-js.css', 'alt/video-js-cdn.css'],
```
gkatsev: would this minify task accidentally keep the minified version into `build/temp/`?

misteroneill: Everything ends up in the right place - `build/temp/alt/` and `dist/alt/`.

gkatsev: yeah, must be minify/grunt "magic" based on the src file.

dmlap:
Looks solid to me.

PR #3120, opened 2016-02-18
Don't break when local storage is disabled

forbesjo:
This PR is to fix an issue where attempting to restore caption settings when localstorage is disabled causes an error.

### DIFF
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -166,7 +166,9 @@ class TextTrackSettings extends Component {
       if (err) {
         log.error(err);
       }
-    } catch (e) { }
+    } catch (e) {
+      log.warn(e);
+    }
 
     if (values) {
       this.setValues(values);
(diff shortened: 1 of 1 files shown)


gkatsev:
LGTM.

src/js/tracks/text-track-settings.js
```diff
@@ -158,11 +158,15 @@ class TextTrackSettings extends Component {
...
-    if (err) {
-      log.error(err);
-    }
+    try {
+      [err, values] = safeParseTuple(window.localStorage.getItem('vjs-text-track-settings'));
+
+      if (err) {
+        log.error(err);
+      }
+    } catch (e) { }
```
dmlap: Shouldn't we log this exception?

forbesjo: I'll log the exception as a warning here as well as in `saveSettings()`.

BrandonOCasey:
LGTM

PR #3128, opened 2016-02-22
Flash tech duration getter returns Infinity instead of negative value

vit-koumar:
## Description
Flash tech `duration` getter should return Infinity instead of negative values (which some players do return) as stated in [HTML standard](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-duration).

This situation is solved in upper layer [Player object](https://github.com/vit-koumar/video.js/blob/master/src/js/player.js#L1353), but sometimes `duration` getter is called from inside the Flash tech for example [here](https://github.com/vit-koumar/video.js/blob/master/src/js/tech/flash.js#L274) which situation ends for example in inability of setting `currentTime` when duration is negative.

## Specific Changes proposed
Explicit `duration()` function defined in [js/tech/flash.js](https://github.com/vit-koumar/video.js/blob/master/src/js/tech/flash.js#L237) which returns Infinity instead of negative Number.

## Requirements Checklist
I will adapt the rest of project in case this change will be approved.

### DIFF
--- a/src/js/tech/flash.js
+++ b/src/js/tech/flash.js
@@ -230,6 +230,16 @@ class Flash extends Tech {
   }
 
   /**
+   * Get media duration
+   *
+   * @returns {Number} Media duration
+   */
+  duration() {
+    var duration = this.el_.vjs_getProperty('duration');
+    return duration >= 0 ? duration : Infinity;
+  }
+
+  /**
    * Load media into player
    *
    * @method load
@@ -312,7 +322,7 @@ class Flash extends Tech {
 // Create setters and getters for attributes
 const _api = Flash.prototype;
 const _readWrite = 'rtmpConnection,rtmpStream,preload,defaultPlaybackRate,playbackRate,autoplay,loop,mediaGroup,controller,controls,volume,muted,defaultMuted'.split(',');
-const _readOnly = 'networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoWidth,videoHeight'.split(',');
+const _readOnly = 'networkState,readyState,initialTime,startOffsetTime,paused,ended,videoWidth,videoHeight'.split(',');
 
 function _createSetter(attr){
   var attrUpper = attr.charAt(0).toUpperCase() + attr.slice(1);
--- a/src/js/tech/flash.js
+++ b/src/js/tech/flash.js
@@ -235,8 +235,12 @@ class Flash extends Tech {
    * @returns {Number} Media duration
    */
   duration() {
-    var duration = this.el_.vjs_getProperty('duration');
-    return duration >= 0 ? duration : Infinity;
+    if (this.readyState() === 0) {
+      return NaN;
+    } else {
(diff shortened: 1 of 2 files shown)


gkatsev:
Should this do a check to see if we have media enabled? Because the spec you linked says that a negative value can mean that there's no media. Otherwise, seems like a good change.

vit-koumar:
I think the spec states that if no media is loaded, then the return value must be NaN. Do you think there should be a check for media load and if no media is loaded, then the return value should be NaN regardless of the duration returned by the player?

vit-koumar:
So I changed the condition like if the readyState === 0 then it returns NaN (according to the standrd). Hope this is what you ment.

src/js/tech/flash.js
```diff
@@ -230,6 +230,20 @@ class Flash extends Tech {
...
   /**
+   * Get media duration
+   *
+   * @returns {Number} Media duration
+   */
+  duration() {
+    if (this.readyState() === 0) {
+      return NaN;
+    } else {
+      var duration = this.el_.vjs_getProperty('duration');
```
gkatsev: This should be a `let` instead.

gkatsev:
One final small change, LGTM otherwise. Thanks.

vit-koumar:
Thanks for the comment, I have amended the change. Is there anything else to do before merge, eg. some tests, docs, etc. ?

gkatsev:
Adding unit tests would be great!

vit-koumar:
Unit tests added.

gkatsev:
Hey, looks like this branch got into conflict with master, can you rebase or merge in master and fix the conflicts?

vit-koumar:
Rebased, conflict solved (+changed commit messages to conform your style)

gkatsev:
Awesome, thanks @vit-koumar.

BrandonOCasey:
LGTM

gkatsev:
Thanks @vit-koumar! Should have a release soon.

gkatsev:
@vit-koumar hey, looks like there are some errors on IE. Would you be able to take a look? See https://travis-ci.org/videojs/video.js/builds/145613240

vit-koumar:
Yes, I will have a look.

test/unit/tech/flash.test.js
```diff
@@ -197,6 +197,33 @@ test('play after ended seeks to the beginning', function() {
...
+      vjs_getProperty() {
+        return mockedDuration;
+      }
+    },
+    readyState: function() {
+      return mockedReadyState;
+    }
+  };
+  result = duration.call(mockFlash);
+  ok(Number.isNaN(result), 'duration returns NaN when readyState equals 0');
```
gkatsev: Looks like this and `Number.isFinite` below the two things causing issues. I'll fix it by including `es6-shim` in the test builds.

gkatsev:
@vit-koumar hey, I fixed it. See #3453.

vit-koumar:
Nice, thanks!

PR #3130, opened 2016-02-22
converted remaining text-track modules to full es6

BrandonOCasey:
## Description
Converted several text-track files to full es6 and added jsdoc comments to the files that I touched

## Specific Changes proposed
* added jsdoc comments to functions that did not previously have them
* linted changed files with vjsstandard were it would not conflict with current jshint
* converted text-track files to full es6 with classes

## Requirements Checklist
- [x] Manually Tested on Chrome, Firefox, Safari, and IE8

### DIFF
--- a/src/js/tracks/text-track-cue-list.js
+++ b/src/js/tracks/text-track-cue-list.js
@@ -14,7 +14,7 @@ import document from 'global/document';
  *   TextTrackCue? getCueById(DOMString id);
  * };
  *
- * @param {Array} cues cues to add to the list
+ * @param {Array} cues A list of cues to be initialized with
  * @class TextTrackCueList
  */
 
--- a/src/js/tracks/text-track-cue-list.js
+++ b/src/js/tracks/text-track-cue-list.js
@@ -4,7 +4,8 @@
 import * as browser from '../utils/browser.js';
 import document from 'global/document';
 
-/*
+/**
+ * A List of text track cues as defined in:
  * https://html.spec.whatwg.org/multipage/embedded-content.html#texttrackcuelist
  *
  * interface TextTrackCueList {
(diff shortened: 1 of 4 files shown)


src/js/tracks/text-track-enums.js
```diff
@@ -1,27 +1,39 @@
...
-  'metadata': 'metadata'
+const TextTrackKind = {
+  subtitles: 'subtitles',
+  captions: 'captions',
+  descriptions: 'descriptions',
+  chapters: 'chapters',
+  metadata: 'metadata'
 };
 
+/* jshint ignore:start */
```
gkatsev: it probably doesn't understand `const`.

src/js/tracks/text-track-list.js
```diff
@@ -18,131 +19,145 @@ import document from 'global/document';
  *   attribute EventHandler onaddtrack;
  *   attribute EventHandler onremovetrack;
  * };
+ *
+ * @param {Track[]} tracks A list of tracks to add to the list
```
gkatsev: "A list of tracks to initialize the list with" would be more accurate.

src/js/tracks/text-track-list.js
```diff
@@ -18,131 +19,145 @@ import document from 'global/document';
...
-TextTrackList.prototype.addTrack_ = function(track) {
-  let index = this.tracks_.length;
-  if (!(''+index in this)) {
-    Object.defineProperty(this, index, {
-      get: function() {
-        return this.tracks_[index];
-      }
+    this.trigger({
+      type: 'addtrack',
+      track
```
gkatsev: can we move the shorthand properties to the beginning of the object?

src/js/tracks/text-track.js
```diff
@@ -284,7 +68,7 @@ var loadTrack = function(src, track) {
   }));
 };
 
-var indexOf = function(searchElement, fromIndex) {
+const indexOf = function(searchElement, fromIndex) {
```
gkatsev: we can actually remove `indexOf`now because we're assuming an `es5` environment now. The changes will need to be made from usage similar to `indexOf.call(this.activeCues_, active[i]` to `this.activeCues_.indexOf(active[i])`.

src/js/tracks/text-track.js
```diff
@@ -318,4 +102,245 @@ var indexOf = function(searchElement, fromIndex) {
...
+
+    if (mode !== 'disabled') {
+      tt.tech_.on('timeupdate', timeupdateHandler);
+    }
+
+    Object.defineProperty(tt, 'kind', {
+      get() {
+        return kind;
+      },
+      set: Function.prototype
```
gkatsev: Since we're already in here, can we change `Function.prototype` to `() => {}` or `const noop = () => {}` and then use `noop`. It was fun while it lasted.

src/js/tracks/text-track.js
```diff
@@ -13,229 +13,13 @@ import window from 'global/window';
...
-  }
-}
-
-TextTrack.prototype = Object.create(EventTarget.prototype);
-TextTrack.prototype.constructor = TextTrack;
-
-/*
- * cuechange - One or more cues in the track have become active or stopped being active.
+/**
+ * Downloading stuff happens below this point
```
gkatsev: > parseCue takes a webvtt file contents and parses it into cues

gkatsev:
Made some comments, mostly minor changes.

src/js/tracks/text-track-list.js
BrandonOCasey: Would storing the array inside of the TextTrackList object rather than tracks_ work for this? or do we not want to remove that variable? This is the idea

``` JavaScript
add(track) {
    // auto increments length for us and stores it in our object under the correct index
    [].push.call(this, track);
  }
  remove(track) {
    for(let i = 0; i < this.length; i++) {
      if(this[i] === track) {
        // auto decrements length for us and splices right out of our object
        [].splice.call(this, i, 1);
        break;
      }
    }
  }

```

gkatsev: Unfortunately, we can't actually inherit from built-in currently.
Also, what I mean by my comment was to change this line from:
```js
this.trigger({
  type: 'addtrack',
  track
});
```
to
```js
this.trigger({
  track,
  type: 'addtrack'
});
```

BrandonOCasey:
@gkatsev OK I think I addressed your comments

src/js/tracks/text-track-cue-list.js
```diff
@@ -12,72 +13,93 @@ import document from 'global/document';
  *   getter TextTrackCue (unsigned long index);
  *   TextTrackCue? getCueById(DOMString id);
  * };
+ *
+ * @param {Array} cues cues to add to the list
```
gkatsev: This comment also needs to be updated. `cues to be initialized with` or something.

src/js/tracks/text-track.js
```diff
@@ -34,288 +100,223 @@ import XHR from 'xhr';
...
 
-      return cues;
-    },
-    set: Function.prototype
-  });
+    Object.defineProperty(tt, 'kind', {
+      get() {
+        return kind;
+      },
+      set: () => {}
```
dmlap: This is in the nitpicky category but I'm wondering if we should encode noops like this:
```js
{
  set(): {}
}
```
The fat arrow functions have extra functionality we don't want here and I think they'll end up getting transpiled into more code.

gkatsev: Looks like it's smart enough to not try anything: http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&playground=true&code=let%20o%20%3D%20{%0A%20%20set%28%29%20{}%2C%0A%20%20set2%3A%20%28%29%20%3D%3E%20{}%0A}%3B%0A%0Alet%20o2%20%3D%20{...o}%3B
However, I do like just using the shorthand, makes it cleaner.

dmlap: I'm in favor of sticking with the shorthand syntax for consistency (but good to know it gets optimized in the transpiling).

dmlap:
Other than the comment about fat-arrow functions, lgtm

dmlap:
LGTM

PR #3138, opened 2016-02-24
Listen to timeupdate on waiting

### LINKED ISSUE #3124: IE11 - currenTime()  and waiting wheel problem
In IE11 when call currenTime( seconds ) to go to a certain point, the image of the waiting wheel image remains and does not disappear. 

In Firefox, Safari, Chrome, Android, IE8 (with flash fallback) works well.

Versión of video.js:  5.8.0

gkatsev:
## Description
This is to workaround a bug currently present in IE11 and potentially in
other cases where we get the waiting event and set the vjs-waiting
class on the player which shows the spinner but then we don't get any
of the events that we use to remove the spinner.
This fixes #3124.

## Specific Changes proposed
Inside of `handleTechWaiting_`, listen to the first `timeupdate` event to clear the `vjs-waiting` class from the player.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -812,6 +812,7 @@ class Player extends Component {
   handleTechWaiting_() {
     this.addClass('vjs-waiting');
     this.trigger('waiting');
+    this.one('timeupdate', () => this.removeClass('vjs-waiting'));
   }
 
   /**
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -943,3 +943,11 @@ test('player#reset loads the first item in the techOrder and then techCalls rese
   equal(loadedSource, null, 'with a null source');
   equal(techCallMethod, 'reset', 'we then reset the tech');
 });
+
+test('Remove waiting class on timeupdate after tech waiting', function() {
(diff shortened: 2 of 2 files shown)


dmlap:
Should we have a test for this? (Actual question)

gkatsev:
I can try adding one.

gkatsev:
@dmlap I added a simple test.

dmlap:
LGTM

mctep:
Hello! Thank you for your job.

May be it is better to trigger `canplay` event in html5 tech only for IE11 after `waiting` and `timeupdate`?

In my case I use https://github.com/google/shaka-player and it triggers `timeupdate` right after `bufferingStart` event. And spinner works incorrect after this fix. So I need to trigger `waiting` event in next tick:

```js
class ShakaTech extends Tech {
 constructor() {
  // ...
  shakaPlayer.on('bufferingStart', () => {
    setTimeout(() => this.trigger('waiting'));
  })
  // ...
 }
}
```

It is not big problem. But I think it is not consistent and not obvious.

PR #3149, opened 2016-03-01
Keep tooltips inside

gkatsev:
## Description
Add a new option that makes the tooltips stay inside the bounds of the player.

## Specific Changes proposed
The new option is `keepWithin` and it's part of the `progressControl`. A new option is required because we can't change the defaults and the changes herein would be breaking otherwise.
Basically, we create a new progress control that we use for the tooltip which we bound with `min-width` and `max-width` to be half the width of the tooltip plus player-width minus half the width of the tooltip.
Also, this is an option because the default videojs skin doesn't have the progress bar go edge-to-edge of the player, so, there's no need to do the extra computations then. But if a skin uses an edge-to-edge progress bar and likes the time tooltips, then, this will be beneficial.

This contains the first attempt at doing this as well as part of the first commit.

## Requirements Checklist
- [x] Cleanup, make sure everything looks for the `keepWithin` option
- [x] Rename `keepWithin` to `keepTooltipsInside` or something
  - [x] Docs/guides updated
  - [x] Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))

### DIFF
--- a/src/js/control-bar/progress-control/mouse-time-display.js
+++ b/src/js/control-bar/progress-control/mouse-time-display.js
@@ -51,12 +51,27 @@ class MouseTimeDisplay extends Component {
   update(newTime, position) {
     let time = formatTime(newTime, this.player_.duration());
 
-    this.el().style.left = position + 'px';
+    this.el().style.left = this.clampPosition(position) + 'px';
     this.el().setAttribute('data-current-time', time);
   }
 
   calculateDistance(event) {
     return Dom.getPointerPosition(this.el().parentNode, event).x;
+  }
+
+  clampPosition(position) {
+    let playerWidth = parseFloat(getComputedStyle(this.player().el()).width);
+    let tooltipWidth = parseFloat(getComputedStyle(this.el(), ':after').width);
+    let tooltipWidthHalf = tooltipWidth / 2;
+    let actualPosition = position;
+
+    if (position < tooltipWidthHalf) {
+      actualPosition = Math.ceil(tooltipWidthHalf);
+    } else if (position > (playerWidth - tooltipWidthHalf)) {
+      actualPosition = Math.floor(playerWidth - tooltipWidthHalf);
+    }
+
+    return actualPosition;
   }
 }
 
--- a/src/js/control-bar/progress-control/mouse-time-display.js
+++ b/src/js/control-bar/progress-control/mouse-time-display.js
@@ -1,6 +1,7 @@
 /**
  * @file mouse-time-display.js
  */
+import window from 'global/window';
 import Component from '../../component.js';
 import * as Dom from '../../utils/dom.js';
 import * as Fn from '../../utils/fn.js';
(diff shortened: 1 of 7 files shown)


gkatsev:
This apparently doesn't work in firefox. Firefox always returns `auto` for the width of the tooltips whereas chrome is returning the pixel values for the width. Investigating.

gkatsev:
I found this firefox issue while trying to create an example on plnkr http://plnkr.co/edit/91wBL105FIOLuEf9JNxG?p=preview
I also created a reduced test case for just the issue https://jsfiddle.net/pkk0dp2k/2/
And unfortunately, the answer is that an element is required since pseudo elements don't give us nearly enough information via computed style.

gkatsev:
I believe this is ready to be reviewed now. I want to add a simple test that verifies that the correct elements are created when the option is set. Also, maybe some unit tests for some of the new methods like `clampPosition`.

src/js/control-bar/progress-control/mouse-time-display.js
```diff
@@ -21,6 +22,19 @@ class MouseTimeDisplay extends Component {
...
+    if (options.playerOptions &&
+        options.playerOptions.controlBar &&
+        options.playerOptions.controlBar.progressControl &&
+        options.playerOptions.controlBar.progressControl.keepTooltipsInside) {
+      this.keepTooltipsInside = options.playerOptions.controlBar.progressControl.keepTooltipsInside;
+    }
+
+    if (this.keepTooltipsInside) {
+      this.el().innerHTML = `<div class='vjs-time-tooltip'>`;
+      this.tooltip = this.el().querySelector('.vjs-time-tooltip');
```
misteroneill: I wonder if this would be preferable to adding to the DOM then querying it?
```js
this.tooltip = Dom.createEl('div', {className: 'vjs-time-tooltip'});
this.el().appendChild(this.tooltip);
```

src/js/control-bar/progress-control/mouse-time-display.js
```diff
@@ -53,11 +67,40 @@ class MouseTimeDisplay extends Component {
...
+      this.tooltip.innerHTML = time;
+      this.tooltip.style.right = `-${tooltipWidthHalf - difference}px`;
+    }
   }
 
   calculateDistance(event) {
     return Dom.getPointerPosition(this.el().parentNode, event).x;
   }
+
+  clampPosition(position) {
```
misteroneill: Seems like this might deserve a doc block to explain what "clamping" position is.

gkatsev: Yep, also, I need to add some general docs/guides for it.

src/js/control-bar/progress-control/mouse-time-display.js
gkatsev: that's probably better.

misteroneill:
LGTM

PR #3154, opened 2016-03-03
Add replay button

aydinkn:
Replay button has been added to the play-toggle


gkatsev:

mmcc:
My initial thought was that I think this probably falls inline more with a plugin or custom implementation, not core. This feels pretty similar to the argument for showing the big play button and poster at the end of a video...not a bad idea at all, just not "core".

That being said, this feels pretty standard these days for "custom" players. The only real argument against is that this isn't the video element's behavior, but it feels like we've strayed pretty far from that policy of late.

mmcc:
(also, thanks for the contribution, @aydinkn!)

gkatsev:
Yeah, I was leaning that way too, that maybe it should be a plugin.
Also, doesn't safari have a replay button at the end?

misteroneill:
If the intent of video.js is to track the `video` element and avoid deviating from the UI it presents - for reasons of maintaining focus and avoiding too much feature bloat - I think this ought to be a plugin. But definitely one that a _lot_ of people would use! :smile:

aydinkn:
Hi guys.
Ok we discuss replay property. 
> I develop a video.js PLUGIN my man. It is REPLAY the video. This is awesome

Strange the say it :)

I think this must be "core" video player property. Because "replay" is the one of the core action in all web and desktop players. video.js has already replay action over the video element. Then why we don't have a replay button?

gkatsev:
Thinking a bit more about this, I don't think we want to make this change in 5.x. We could probably make this change in the next major update.
Thanks for the PR and interest, @aydinkn. I hope you stick around and help out!

PR #3098, opened 2016-02-10
Basic descriptions track support

OwenEdwards:
Adds support for descriptions text tracks, with a separate menu button on the control bar, and display using the same mechanism as captions and subtitles. To handle the situation where a descriptions and a captions or subtitles track could be enabled at the same time, the TextTrackDisplay prefers non-descriptions tracks over descriptions tracks, and the descriptions menu button disables itself if another track is enabled.

This PR could probably use some more Unit tests to check the priority model is TextTrackDisplay, and the self-disabling in descriptions-button. Also, the CSS styling applied to the button when it is disabled (just making it 0.5 opacity) may need more work.

I'd appreciate any feedback, suggestions, etc.

PR #3163, opened 2016-03-08
Fix chapters menu

### LINKED ISSUE #3062: menus aren't getting sized correctly anymore
vjs-menu's width is set to 0 but vjs-menu-content is supposed to be set to 24em but ends up getting set to 'inherit', so, menus aren't showing up properly, at least for the chapters menu.
also, `vjs-selected` isn't getting applied properly to the selected element.

gkatsev:
## Description
Fixes #3062. The `left` and `width` properties for the chapters menu is unnecessary.
Also, we need to make sure that the chapters title gets added in first into the childrens array so that it's always at the top.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/text-track-controls/chapters-button.js
+++ b/src/js/control-bar/text-track-controls/chapters-button.js
@@ -90,11 +90,13 @@ class ChaptersButton extends TextTrackButton {
     let menu = this.menu;
     if (menu === undefined) {
       menu = new Menu(this.player_);
-      menu.contentEl().appendChild(Dom.createEl('li', {
+      let title = Dom.createEl('li', {
         className: 'vjs-menu-title',
         innerHTML: toTitleCase(this.kind_),
         tabIndex: -1
-      }));
+      });
+      menu.children_.unshift(title);
+      Dom.insertElFirst(title, menu.contentEl());
     }
 
     if (chaptersTrack && chaptersTrack.cues == null) {
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -70,11 +70,13 @@ class MenuButton extends ClickableComponent {
 
     // Add a title list item to the top
     if (this.options_.title) {
-      menu.contentEl().appendChild(Dom.createEl('li', {
+      let title = Dom.createEl('li', {
         className: 'vjs-menu-title',
         innerHTML: toTitleCase(this.options_.title),
         tabIndex: -1
       }));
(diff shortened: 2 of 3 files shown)


OwenEdwards:
FYI, there's no Chapters menu example in the existing repo, but there is one in the Descriptions example in #3098.

gkatsev:
Yep, for testing I just copied the standard oceans track and called it chapters.

gkatsev:
Also, looking into #3098 is what prompted me to fix this issue in the first place.

OwenEdwards:
Oh, great! I can take a look at this fix or yours this evening, and give you a LGTM.

gkatsev:
That would be great, thanks!

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -90,11 +90,13 @@ class ChaptersButton extends TextTrackButton {
     let menu = this.menu;
     if (menu === undefined) {
       menu = new Menu(this.player_);
-      menu.contentEl().appendChild(Dom.createEl('li', {
+      let title = Dom.createEl('li', {
```
OwenEdwards: 1/. Since this is a new Menu, what's the situation where it can have children already?

2/. meun-button.js already has this mechanism to add a title (using options.title) which doesn't seem to be used elsewhere - why isn't that mecanism used here?

OwenEdwards:
A couple of comments, otherwise LGTM. It exposes a limitation of the keyboard handling of menus, though, which I'll need to fix. :-(

src/js/control-bar/text-track-controls/chapters-button.js
gkatsev: 1. It's unlikely that it should be a problem, especially since it should be the first item.
2. I'll have to take a look at that and see whether it applies in this case or what its used for. Thanks.

src/js/menu/menu-button.js
```diff
@@ -70,11 +70,13 @@ class MenuButton extends ClickableComponent {
 
     // Add a title list item to the top
     if (this.options_.title) {
-      menu.contentEl().appendChild(Dom.createEl('li', {
+      let title = Dom.createEl('li', {
```
gkatsev: The menu title should probably be a separate dive inside menu with the menu items being in a separate UL, but we can't make the change in a minor update.

OwenEdwards:
LGTM.

PR #3165, opened 2016-03-09
Ignore titles in menu, if they exist.

### LINKED ISSUE #3164: Keyboard control of menus doesn't work for the Chapters menu
## Description

Keyboard acces to menus was fixed in #3033, but #3163 exposes that keyboard access doesn't work for a menu with a title (such as the Chapters menu).

### Expected behavior

Use Tab to move focus to the Captions menu button
Use Enter to open the menu
Observe that the arrow keys move focus up and down the menu

### Observed behavior

Then use Tab to move focus to the Chapters Menu button
Use Enter to open the menu
Arrow keys do not work

### Error output

`TypeError: children[item].el_ is undefined`

## Additional Information
#### videojs

Any versions after #3163 is merged.

#### browsers
what browser are affected?

Checked on Firefox, v44.

#### OSes

Checked on OS X.

### plugins

No.

gkatsev:
## Description
This is because of the change in #3163 where we added the title as an
item in the childrens array so that it is ordered correctly but this
breaks keyboard support because they expect the children to be lined up
correctly and also because they expect the children to be component
objects with an 'el_' property.
Checking to see if the first item is a 'vjs-menu-title' and adding one
to the 'item' allows us to ignore the title element and also select the
actual menu items.
Fixes #3164.

## Requirements Checklist

### DIFF
--- a/src/js/menu/menu.js
+++ b/src/js/menu/menu.js
@@ -117,7 +117,14 @@ class Menu extends Component {
    * @method focus
    */
   focus (item = 0) {
-    let children = this.children();
+    let children = this.children().slice();
+    let haveTitle = children[0].className &&
+      /vjs-menu-title/.test(children[0].className);
+
+    if (haveTitle) {
+      children.shift();
+    }
+
 
     if (children.length > 0) {
       if (item < 0) {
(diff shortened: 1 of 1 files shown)


gkatsev:
I think relying on the lack of `el_` to tell that it's a "title" isn't specific enough, since, maybe something else snuck in there somehow? We probably should also verify that `el_` exists before trying to call a method on it.

OwenEdwards:
LGTM. :+1:

PR #3166, opened 2016-03-10
[WIP] Feature audio/video tracks

benjipott:
## Description
https://github.com/videojs/video.js/issues/3137

capture d ecran 2016-03-10 a 17 24 29
capture d ecran 2016-03-11 a 16 07 21

ex with native hls sources : 
https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8
with mutli audio

after :  add switch bitrate button an switch audio button

##### Follow the embedded content specification to:
implement the sharing of common functionality between TextTracks, AudioTracks, and VideoTracks.
implement VideoTrack, VideoTrackList, and enumerations
implement AudioTrack, AudioTrackList, and enumerations

PR #3173, opened 2016-03-11
Integrate video & audio tracks with video.js

BrandonOCasey:
## Description
Implement `VideoTrack`,  `AudioTrack`, `VideoTrackList`, and `AudioTrackList` using common functionality from `TextTrack` and `TextTrackList`. Which is now shared in a base classes called `Track` and `TrackList`

## Specific Changes proposed
* Implement Audio/Video Track & Track List
* Move Common functionality into Track and TrackList classes so it can be shared
* Add unit tests for individual everything
* Add unit tests for full integration pieces

## Requirements Checklist


src/js/player.js
```diff
@@ -2473,6 +2477,32 @@ class Player extends Component {
...
+   * Get a video track list
+   * @link https://html.spec.whatwg.org/multipage/embedded-content.html#videotracklist
+   *
+   * @return {VideoTrackList} thes current video track list
+   * @method videoTracks
+   */
+  videoTracks() {
+    // cannot use techGet_ directly because it checks to see whether the tech is ready.
+    // Flash is unlikely to be ready in time but videoTracks should still work.
+    return this.tech_ && this.tech_['videoTracks']();
```
dmlap: This will return `undefined` when the tech is not ready. The spec says it should return an empty VideoTrackList in that case which seems nice if it's easy to pull off.

src/js/player.js
```diff
@@ -2473,6 +2477,32 @@ class Player extends Component {
...
+   * Get an audio track list
+   * @link https://html.spec.whatwg.org/multipage/embedded-content.html#audiotracklist
+   *
+   * @return {AudioTrackList} thes current audio track list
+   * @method audioTracks
+   */
+  audioTracks() {
+    // cannot use techGet_ directly because it checks to see whether the tech is ready.
+    // Flash is unlikely to be ready in time but audioTracks should still work.
+    return this.tech_ && this.tech_['audioTracks']();
```
dmlap: Same comment as for video tracks.

src/js/tech/flash.js
```diff
@@ -312,7 +312,7 @@ class Flash extends Tech {
 // Create setters and getters for attributes
 const _api = Flash.prototype;
 const _readWrite = 'rtmpConnection,rtmpStream,preload,defaultPlaybackRate,playbackRate,autoplay,loop,mediaGroup,controller,controls,volume,muted,defaultMuted'.split(',');
-const _readOnly = 'networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoTracks,audioTracks,videoWidth,videoHeight'.split(',');
+const _readOnly = 'networkState,readyState,initialTime,duration,startOffsetTime,paused,ended,videoWidth,videoHeight'.split(',');
```
dmlap: This will cause getters to be created for SWF callbacks that don't exist and will fail if called at runtime. Remember to test with Flash :) If there's no Flash interaction, you can write a regular ol' javascript method on the prototype to return an empty track list (or whatever you want).

gkatsev:
@benjipott would be great to get your input on this PR!

src/js/player.js
BrandonOCasey: should I make the same change for textTracks(). Or do we not want to change that API?

dmlap: I think that's a good idea-- I'd call the current behavior a bug. We can save that for another PR, though.

gkatsev: For background, we didn't do that for TextTracks because you probably shouldn't be calling these methods on a player that isn't ready (and thus with a tech that isn't ready) plus since the lists are stored on the tech, if we return an empty track list, we'll need to make sure that it is cached and that the tech that is loaded uses that instead of making a new one because otherwise the events like `addtrack` won't work for that one-off.

src/js/tech/html5.js
```diff
@@ -70,6 +72,13 @@ class Html5 extends Tech {
       }
     }
 
+    if (this.featuresNativeVideoTracks) {
+      this.handleVideoTrackChange_ = Fn.bind(this, this.handleVideoTrackChange);
```
dmlap: Why these aliases?

src/js/tech/html5.js
```diff
@@ -282,6 +342,63 @@ class Html5 extends Tech {
...
+    let vt = this.el().videoTracks;
+
+    if (vt && vt.addEventListener) {
+      vt.addEventListener('change', this.handleVideoTrackChange_);
+      vt.addEventListener('addtrack', this.handleVideoTrackAdd_);
+      vt.addEventListener('removetrack', this.handleVideoTrackRemove_);
+    }
+  }
+
+  handleVideoTrackChange(e) {
```
dmlap: Should these handlers be private?

src/js/tech/html5.js
```diff
@@ -282,6 +342,63 @@ class Html5 extends Tech {
...
+      type: 'change',
+      target: vt,
+      currentTarget: vt,
+      srcElement: vt
+    });
+  }
+
+  handleVideoTrackAdd(e) {
+    let track = e.track;
+    if(!(e.track instanceof VideoTrack)) {
```
dmlap: Is this for the native HTML case? If so, could you add a comment?

src/js/tech/tech.js
```diff
@@ -290,6 +298,30 @@ class Tech extends Component {
...
+   * @method initVideoTrackListeners
+   */
+  initVideoTrackListeners() {
+    let videoTrackListChanges = () => {
+      this.trigger('videotrackchange');
+    };
+
+    let tracks = this.videoTracks();
+
+    if (!tracks) return;
```
dmlap: If `videoTracks` always returns a track list, we can get rid of this short-circuiting.

src/js/tech/tech.js
```diff
@@ -313,6 +345,30 @@ class Tech extends Component {
     }));
   }
 
+
+  /**
+   * Initialize audiotrack listeners
+   *
+   * @method initAudioTrackListeners
+   */
+  initAudioTrackListeners() {
```
dmlap: These initializers are very similar and not 100% trivial. Is it possible for them to share more code?

src/js/tracks/audio-track-list.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * @file audio-track-list.js
+ */
+import TrackList from './track-list';
+
+/**
+ * A list of possiblee audio tracks. All functionality is in the
```
dmlap: `s/posiblee/possible/`

src/js/tracks/audio-track-list.js
```diff
@@ -0,0 +1,38 @@
...
+ * @class AudioTrackList
+ */
+class AudioTrackList extends TrackList {
+  addTrack_(track) {
+    super(track);
+    track.addEventListener('enabledchange', () => {
+      this.trigger('change');
+    });
+  }
+  addTrack(track) {
```
dmlap: What's the point of this method?

src/js/tech/html5.js
gkatsev: This is inline with how the native text tracks are done. The reason for them is that in one vendor, the functions weren't being properly bound if we were using `this.handle<Thing>Track<Action>` directly. Binding them here and then using the bound version in the listener fixes it.

src/js/tech/html5.js
gkatsev: Why is this even necessary?

src/js/tech/html5.js
```diff
@@ -282,6 +342,63 @@ class Html5 extends Tech {
...
+      type: 'change',
+      target: audioTrackList,
+      currentTarget: audioTrackList,
+      srcElement: audioTrackList
+    });
+  }
+
+  handleAudioTrackAdd(e) {
+    let track = e.track;
+    if(!(e.track instanceof AudioTrack)) {
```
gkatsev: Same here? Why is it necessary?

src/js/tech/html5.js
```diff
@@ -962,6 +1079,27 @@ Html5.supportsNativeTextTracks = function() {
   return supportsTextTracks;
 };
 
+/*
+ * Check to see if native video tracks are supported by this browser/device
+ *
+ * @return {Boolean}
+ */
+Html5.supportsNativeVideoTracks = function() {
```
gkatsev: #3166 has a nice method that took a string `text`, `audio`, or `video`, and returned whether that thing is supported. Would be good to cherry-pick that change, remembering that `supportsNativeTextTrack` needs to stay for backwards compat but it could be deprecated.

src/js/tech/tech.js
gkatsev: #3166 has them combined, we could cherry-pick that as well.

src/js/tracks/audio-track-list.js
gkatsev: We need a way for users of videojs to add tracks to the track lists, the lists are read-only and done by the "vendor". I don't think projects like contrib-hls should be using a "private" function like `addTrack_` so, I feel like renaming it to `addTrack` is best. Also, I think this is the time to probably bring in a stability index like https://nodejs.org/dist/latest-v5.x/docs/api/documentation.html because the public functions could go away one MSE full supports these natively in which case we could pick the additional tracks that way.

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,64 @@
...
+class AudioTrack extends Track {
+  constructor(options = {}) {
+    options.trackType = 'audio';
+    options.kind = AudioTrackEnums.AudioTrackKind[options.kind] || '';
+    // retval is only defined in ie8, the only time when we need it
+    let retval = super(options);
+    let audioTrack = this;
+
+    let enabled = false;
+    Object.defineProperty(audioTrack, 'enabled', {
```
gkatsev: we want to define these on `retval` because otherwise it won't work on IE8.

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,64 @@
...
+      get() { return enabled; },
+      set(newEnabled) {
+        // do nothing becasue:
+        // and invalid value passed in or the value passed in is the
```
gkatsev: > an invalid value

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,64 @@
...
+        // same as the current value
+        if (typeof newEnabled !== 'boolean' || newEnabled === enabled) {
+          return;
+        }
+
+        // We diverge from the spec here because we can only
+        // support one audio track at a time. So we
+        // make sure to disable any other audio track
+        // before we enable this one
+        if (newEnabled === true) {
```
gkatsev: no need to compare against `true`.

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,64 @@
...
+            // another audio track is enabled, disable it
+            at.enabled = false;
+          }
+        }
+        enabled = newEnabled;
+        this.trigger('enabledchange');
+      }
+    });
+
+    options.enabled = options.enabled || false;
```
gkatsev: why update options?

src/js/tracks/text-track.js
```diff
@@ -102,39 +99,22 @@ const loadTrack = function(src, track) {
...
-    let language = options.language || options.srclang || '';
-    let id = options.id || 'vjs_text_track_' + Guid.newGUID();
 
-    if (kind === 'metadata' || kind === 'chapters') {
+    if (options.kind === 'metadata' || options.kind === 'chapters') {
       mode = 'hidden';
     }
+    // retval will only be defined on IE8, which is when we need it
+    let retval = super(options);
+    let tt = this;
```
gkatsev: same issue as with `AudioTrack`.

src/js/tracks/text-track-list.js
```diff
@@ -20,144 +19,16 @@ import document from 'global/document';
...
-
-    if (!('' + index in this)) {
-      Object.defineProperty(this, index, {
-        get() {
-          return this.tracks_[index];
-        }
-      });
-    }
-
+    super(track);
```
gkatsev: does just calling super work here? Don't you need to do `super.addTrack_`?

src/js/tracks/text-track.js
```diff
@@ -102,39 +99,22 @@ const loadTrack = function(src, track) {
...
-      for (let prop in TextTrack.prototype) {
-        if (prop !== 'constructor') {
-          tt[prop] = TextTrack.prototype[prop];
-        }
-      }
-    }
-
-    tt.tech_ = options.tech;
-
+    options.kind = TextTrackEnum.TextTrackKind[options.kind] || 'subtitles';
```
gkatsev: we shouldn't be mutating the options object. We should be using new local variables.

src/js/tracks/text-track.js
```diff
@@ -262,9 +214,7 @@ class TextTrack extends EventTarget {
       tt.loaded_ = true;
     }
 
-    if (browser.IS_IE8) {
-      return tt;
-    }
+    return retval;
```
gkatsev: This should also still check for IE8 and return retval then. As do all the tracks.

src/js/tracks/track.js
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+    track.tech_ = options.tech;
+
+    let id = options.id || 'vjs_' + type + '_track_' + Guid.newGUID();
+    let kind = options.kind || '';
+    let label = options.label || '';
+    let language = options.language || '';
+
+    Object.defineProperties(track, {
```
gkatsev: `defineProperties` doesn't work in IE8, we can't use it.

src/js/tracks/video-track-list.js
```diff
@@ -0,0 +1,52 @@
...
+ * };
+ *
+ * @param {VideoTrack[]} tracks a list of video tracks to instantiate the list with
+ # @extends TrackList
+ * @class VideoTrackList
+ */
+class VideoTrackList extends TrackList {
+  constructor(tracks) {
+    super(tracks);
+    Object.defineProperty(this, 'selectedIndex', {
```
gkatsev: This still needs the IE8 fix with the custom element.

test/unit/tracks/text-track-list.test.js
```diff
@@ -2,155 +2,7 @@ import TextTrackList from '../../../src/js/tracks/text-track-list.js';
...
-  }, {
-    id: '3',
-    addEventListener() {},
-    off() {}
-  }
-];
-
 q.module('Text Track List');
-
-test('TextTrackList\'s length is set correctly', function() {
```
gkatsev: I assume these tests moved to something like the track-baseline?

src/js/tech/html5.js
BrandonOCasey: native Video and Audio Tracks won't be inherit from eventhandler so we have to use them to create a VideoTrack of our own that does inherit from eventhandler so that we can watch for events regardless of wether it is native or not.

test/unit/tracks/track.test.js
```diff
@@ -0,0 +1,69 @@
...
+q.module('Track');
+TrackBaseline(Track, {
+  id: '1',
+  kind: 'subtitles',
+  mode: 'disabled',
+  label: 'English',
+  language: 'en',
+  tech: new TechFaker()
+});
+test('defaults when items not provided', function() {
```
gkatsev: new line above this

test/unit/tracks/track.test.js
```diff
@@ -0,0 +1,69 @@
...
+const defaultTech = {
+  textTracks() {},
+  on() {},
+  off() {},
+  currentTime() {}
+};
+
+// do baseline track testing
+q.module('Track');
+TrackBaseline(Track, {
```
gkatsev: new line above this

test/unit/tracks/video-track-list.test.js
```diff
@@ -0,0 +1,47 @@
+import VideoTrackList from '../../../src/js/tracks/video-track-list.js';
+import VideoTrack from '../../../src/js/tracks/video-track.js';
+import EventTarget from '../../../src/js/event-target.js';
+
+q.module('Video Track List');
+test('trigger "change" event when "selectedchange" is fired on a track', function() {
```
gkatsev: new line above this

test/unit/tracks/video-track.test.js
```diff
@@ -0,0 +1,199 @@
...
+  videoTracks() {},
+  on() {},
+  off() {},
+  currentTime() {}
+};
+
+q.module('Video Track', {
+  beforeEach() {
+    this.videoTrackList = [];
+    Object.defineProperty(this.videoTrackList, 'selectedIndex', {
```
gkatsev: these tests need to run in IE8 as well, so, you cannot do this. Why not just create a VideoTrackList?

src/js/tech/html5.js
gkatsev: If you always use `addEventListener` with the tracks, they'll work regardless of whether they are native or not and we won't need to do any special handling of them.

src/js/player.js
BrandonOCasey: we should be able to do something like this here:
```JavaScript
audioTracks() {
  if(!this.tech_) {
    this.audioTracks_ = this.audioTracks_ || new AudioTrackList(); 
    return this.audioTracks
  }
  return this.tech_['audioTracks']();
}
```
`this.audioTracks_` is used when we loadTech_ so it should keep the same reference if thats the only place we load tech. Does that sound right?

src/js/tracks/audio-track.js
BrandonOCasey: we won't be covered by the parent returning retval only on IE8?

src/js/tracks/audio-track.js
BrandonOCasey: as in we don't need the if statement or just do the if statement as `if (newEnabled)`?

src/js/tracks/audio-track-list.js
BrandonOCasey: Does the implementation seen above fit that purpose? (`this.addTrack` calling `this.addTrack_`). The reason I went that route is to keeps the TrackList APIs all the same rather than having to do shenanigans with Video & Audio and leaving TextTrackList

src/js/tracks/text-track-list.js
BrandonOCasey: it does work, but I don't think that it should. I will make the change.

test/unit/tracks/text-track-list.test.js
BrandonOCasey: `TextTrackList` tests moved into track-list.test.js. I did not see the need to have a baseline as most of the tracklists are using the parent class mostly verbatim.

test/unit/tracks/video-track.test.js
BrandonOCasey: this unit test was created before I had VideoTrackList, so I will use that now instead

src/js/tracks/audio-track-list.js
dmlap: @gkatsev I don't think contrib-hls should be using private methods either.

@BrandonOCasey oh, text tracks don't have an addTrack()? This still feels a *little* weird but I'm ok with it if that's the case.

src/js/tracks/text-track.js
BrandonOCasey: we won't be covered by the parent returning retval only on IE8?

src/js/tech/html5.js
dmlap: I'm still confused :) Why weren't they bound properly? By "vendor", do you mean the HTML tech or Flash tech for example?

src/js/tracks/audio-track.js
dmlap: So you don't have use to this ugly comparison later `typeof options.enabled === 'undefined'`? :P

src/js/player.js
gkatsev: If that works, that's fine. Just wanted to give the background on it. Previously we didn't have that property that was used to initialize the tech with.

src/js/tech/html5.js
gkatsev: Vendor like safari, unfortunately, I don't remember the exact details. Was merged as part of https://github.com/videojs/video.js/pull/2425.

src/js/tracks/audio-track-list.js
gkatsev: Ah, I see your point. In that case, I guess it makes sense. We should still add a stability index as part of this because these methods could go away in the future.

gkatsev: Also, can you add a new line above this one?

src/js/tracks/audio-track.js
gkatsev: just `if (newEnabled)`

src/js/tracks/audio-track.js
gkatsev: Rather, I mean, we shouldn't be updating `options`, we should be using the `enabled` local var.

src/js/tracks/text-track.js
gkatsev: We should set `retval` to the response from super or `this` if it isn't in IE8 like it works in the main `track.js` file. Also, a more meaningful name than `retval` would be good.

test/unit/tracks/text-track-list.test.js
gkatsev: Right. That's fine.

src/js/tech/html5.js
BrandonOCasey: native Audio and Video Tracks won't have an addEventListener as they don't have a parent class.

src/js/tracks/text-track.js
BrandonOCasey: why not just always return this?

src/js/tech/html5.js
gkatsev: Huh, weird. Also, they don't actually emit any events so they don't technically need to inherit from EventTarget: https://html.spec.whatwg.org/multipage/embedded-content.html#audiotrack.
I wonder if we can come up with a separate mechanism letting the corresponding TrackList know that the selected/enabled options changed.

gkatsev: I really would prefer if we didn't have to wrap the native tracks if we can avoid it.

src/js/tracks/text-track.js
gkatsev: The way constructors in js work when called with `new` is that a new object is allocated and made available in via `this` but if you return a *different* object, the return value for `new Foo()` will be that different object. We need to do that for IE8 because IE8 doesn't let us use getters (`defineProperty`) on non-HTML elements. So, in IE8, we create a custom element, attach the getters and setters to it and then return that object. Only IE8 has this limitation and videojs 5.x must continue supporting IE8 at least for existing features and new features cannot break IE8, though, we could decide on a case-by-case basis whether new features should support IE8.

src/js/tracks/track.js
```diff
@@ -0,0 +1,55 @@
...
+      id: options.id || 'vjs_' + type + '_track_' + Guid.newGUID(),
+      kind: options.kind || '',
+      label: options.label || '',
+      language: options.language || ''
+    };
+
+    for(let key in trackProps) {
+      Object.defineProperty(track, key, {
+        get() { return trackProps[key]; },
+        set() {}
```
dmlap: @gkatsev why did we define these no-op setters?

src/js/tracks/video-track-enums.js
```diff
@@ -0,0 +1,31 @@
...
+ *   "alternative",
+ *   "captions",
+ *   "main",
+ *   "sign",
+ *   "subtitles",
+ *   "commentary",
+ *   "",
+ * };
+ */
+const VideoTrackKind = {
```
dmlap: It's a common idiom to define enums for these sorts of constants in many programming languages but I think it adds unnecessary abstraction here. The arguments in favor tend to come in two primary flavors:

1. It's easier to change with automated tools
2. It's good documentation

Point 1 doesn't hold for javascript because of its dynamic nature-- to be safe you have to do a find-replace anyways. Secondly, these constants are actually strings without a first class enum in the spec. We won't be changing them (beyond addition, maybe) anytime soon.

Point 2 makes a lot of sense but there are other ways of achieving it that don't impact runtime behavior. Our documentation allows you to add stub entries wherever you want that get placed into the generated API docs. That seems a better way of surfacing this info if it's important to users.

src/js/tracks/video-track-list.js
```diff
@@ -0,0 +1,53 @@
...
+    Object.defineProperty(videoTrackList, 'selectedIndex', {
+      get() {
+        for (let i = 0; i < this.length; i++) {
+          if (this[i].selected()) {
+            return i;
+          }
+        }
+        return -1;
+      },
+      set() {},
```
dmlap: @gkatsev same question about no-op setter.

@BrandonOCasey unnecessary trailing comma.

src/js/tracks/audio-track-list.js
```diff
@@ -0,0 +1,38 @@
...
+ * };
+ *
+ * @param {AudioTrack[]} tracks a list of audio tracks to instantiate the list with
+ # @extends TrackList
+ * @class AudioTrackList
+ */
+class AudioTrackList extends TrackList {
+  addTrack_(track) {
+    super.addTrack_(track);
+    track.addEventListener('enabledchange', () => {
```
benjipott: native audio track don't have addEventListener methods

src/js/tracks/video-track-list.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+    return videoTrackList;
+  }
+  addTrack_(track) {
+    super.addTrack_(track);
+    track.addEventListener('selectedchange', () => {
```
benjipott: native video track don't have addEventListener method

src/js/video.js
```diff
@@ -546,6 +548,20 @@ videojs.xhr = xhr;
 videojs.TextTrack = TextTrack;
 
 /**
+ * AudioTrack class
```
dmlap: Could you add comments about why these two classes are available publicly? (I assume it's so source handlers can create and add them at runtime)

src/js/tracks/video-track-list.js
BrandonOCasey: in the html5 tech we wrap native audio and video tracks with our own class (still up for debate wether we should be doing that or not) so they will have an even listener in that case.

test/unit/tracks/audio-track.test.js
```diff
@@ -0,0 +1,162 @@
...
+// do baseline track testing
+TrackBaseline(AudioTrack, {
+  id: '1',
+  language: 'en',
+  label: 'English',
+  kind: 'main',
+  tech: defaultTech
+});
+
+test('Can create an AudioTrack an enabled property', function() {
```
dmlap: The capital 'C' looks a little out-of-place here.

test/unit/tracks/audio-track.test.js
```diff
@@ -0,0 +1,162 @@
...
+});
+
+test('kind can only be one of several options, defaults to empty string', function() {
+  let track = new AudioTrack({
+    tech: this.tech,
+    kind: 'foo'
+  });
+
+  equal(track.kind, '', 'the kind is set to empty string, not foo');
+  notEqual(track.kind, 'foo', 'the kind is set to empty string, not foo');
```
dmlap: I would vote to skip this level of input validation. I think the number of people who will try to create tracks programmatically is very small and they're going to have to read the docs to understand how to do it anyways.

test/unit/tracks/audio-track.test.js
```diff
@@ -0,0 +1,162 @@
...
+  let eventsTriggered = 0;
+  track.addEventListener('enabledchange', () => {
+    eventsTriggered++;
+  });
+
+  // two events
+  track.enabled = true;
+  track.enabled = false;
+
+  // no event here
```
dmlap: Might be nice to turn these comments into assertions.

test/unit/tracks/audio-tracks.test.js
```diff
@@ -0,0 +1,114 @@
...
+  Html5.TEST_VID = {
+    audioTracks: []
+  };
+
+  ok(Html5.supportsNativeAudioTracks(), 'native audio tracks are supported');
+
+  Html5.TEST_VID = oldTestVid;
+});
+
+test('html5 tech does nt support native audio tracks if the video does not supports it', function() {
```
dmlap: > does *not* support native audio tracks if the video does not *support* it

test/unit/tracks/track-baseline.js
```diff
@@ -0,0 +1,42 @@
...
+    track.language = 'es';
+    track.id = '2';
+
+    equal(track.kind, options.kind, 'we have a kind');
+    equal(track.label, options.label, 'we have a label');
+    equal(track.language, options.language, 'we have a language');
+    equal(track.id, options.id, 'we have an id');
+  });
+
+  test('returns an instance of of itself on all browsers', function() {
```
dmlap: > of of

test/unit/tracks/video-track.test.js
```diff
@@ -0,0 +1,189 @@
...
+// do baseline track testing
+TrackBaseline(VideoTrack, {
+  id: '1',
+  language: 'en',
+  label: 'English',
+  kind: 'main',
+  tech: defaultTech
+});
+
+test('Can create an VideoTrack a selected property', function() {
```
dmlap: Kinda weird capital 'C' again.

dmlap:
Awesome tests!

src/js/player.js
dmlap: Sounds like an excellent solution to me.

src/js/tech/html5.js
dmlap: I don't like the fact we don't remember *why* we have to do this weird dance here. @BrandonOCasey would you mind giving this a shot without the binding thing in Safari?

src/js/tracks/audio-track.js
dmlap: Ah, ok.

src/js/tracks/video-track-list.js
dmlap: We could trigger all events on the owning TrackList object without much trouble and then stay closer to the spec.

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,62 @@
...
+        // We diverge from the spec here because we can only
+        // support one audio track at a time. So we
+        // make sure to disable any other audio track
+        // before we enable this one
+        if (newEnabled) {
+          let audioTrackList = audioTrack.tech_.audioTracks();
+          for(let i = 0; i < audioTrackList.length; i++) {
+            let at = audioTrackList[i];
+            // another audio track is enabled, disable it
+            at.enabled = false;
```
dmlap: I don't think a track should reach out to sibling tracks and read or modify them. Could this logic be moved up into the track list objects?

src/js/tracks/video-track-list.js
gkatsev: We're already proxying the native events from TrackList, so we don't need handle `selectedchange` for them, in fact, if shouldn't because we'll get double events.
I really would prefer not to wrap the native Tracks if we can avoid it.

src/js/tracks/video-track.js
```diff
@@ -0,0 +1,54 @@
...
+    let selected = false;
+
+    Object.defineProperty(videoTrack, 'selected', {
+      get() { return selected; },
+      set(newSelected) {
+        // an invalid value
+        if(typeof newSelected !== 'boolean' || newSelected === selected) {
+          return;
+        }
+        if(newSelected) {
```
dmlap: Same comment here as for audio track's `enabled` property. The track list seems like the track owner to me so I think it should be responsible for managing cross-track changes like this.

src/js/tracks/video-track-enums.js
BrandonOCasey: So is your proposal to add these lists to the Track files where they are used?

src/js/tech/tech.js
BrandonOCasey: should we keep initTextTrackListeners? since its public and we do not want to change the api.

src/js/tracks/track.js
gkatsev: `id`, `kind`, `label`, and `language`, are all read-only properties. Making the setter a no-op makes sure that they stay read-only.

src/js/tech/tech.js
gkatsev: Probably, as part of the release with this PR, I think we should introduce the stability index which would affect this method as well. We probably should have it log a deprecation warning.

src/js/tech/html5.js
gkatsev: I would be wary to change how it works for the text tracks themselves but OK with not doing it for the new video and audio track support.

src/js/tracks/video-track-enums.js
gkatsev: These "enums" beyond being a direct implementation from the spec do serve a useful purpose for us. It allows us to much more easily have a default `kind`.
I'm OK with moving all the "enums" into a single `tracks-enums.js` file and I could be convinced of moving them into their respective track files.

src/js/tracks/audio-track.js
gkatsev: This functionality is sort of weird regardless because the "vendor" (i.e., browser) that implements the audio tracks is sort of expected to make this change (assuming the vendor also only support one audio-track at a time). This functionality could be moved into a private function of the audio track list though, it isn't that much different.

src/js/tech/html5.js
```diff
@@ -282,6 +342,67 @@ class Html5 extends Tech {
...
+      currentTarget: vt,
+      srcElement: vt
+    });
+  }
+
+  handleVideoTrackAdd(e) {
+    let track = e.track;
+    // native video tracks will not inherit from EventTarget
+    // so we have to turn them into our own VideoTrack class
+    if(!(e.track instanceof VideoTrack)) {
```
gkatsev: space after `if`.

src/js/tech/tech.js
```diff
@@ -313,6 +345,30 @@ class Tech extends Component {
...
+    let audioTrackListChanges = () => {
+      this.trigger('audiotrackchange');
+    };
+
+    let tracks = this.audioTracks();
+
+    if (!tracks) return;
```
gkatsev: always use braces, even around single-line ifs.

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,62 @@
...
+          return;
+        }
+
+        // We diverge from the spec here because we can only
+        // support one audio track at a time. So we
+        // make sure to disable any other audio track
+        // before we enable this one
+        if (newEnabled) {
+          let audioTrackList = audioTrack.tech_.audioTracks();
+          for(let i = 0; i < audioTrackList.length; i++) {
```
gkatsev: space after `for`

src/js/tracks/video-track.js
```diff
@@ -0,0 +1,54 @@
...
+    // on IE8 this will be a document element
+    // for every other browser this will be a normal object
+    let videoTrack = super(options);
+    let selected = false;
+
+    Object.defineProperty(videoTrack, 'selected', {
+      get() { return selected; },
+      set(newSelected) {
+        // an invalid value
+        if(typeof newSelected !== 'boolean' || newSelected === selected) {
```
gkatsev: space after `if`

test/unit/tracks/audio-track-list.test.js
```diff
@@ -0,0 +1,43 @@
+import AudioTrackList from '../../../src/js/tracks/audio-track-list.js';
+import AudioTrack from '../../../src/js/tracks/audio-track.js';
+import EventTarget from '../../../src/js/event-target.js';
+
+q.module('Audio Track List');
```
gkatsev: new line after `q.module`

test/unit/tracks/audio-track.test.js
```diff
@@ -0,0 +1,162 @@
...
+  let track = new AudioTrack({
+    tech: this.tech,
+    kind: 'foo'
+  });
+
+  equal(track.kind, '', 'the kind is set to empty string, not foo');
+  notEqual(track.kind, 'foo', 'the kind is set to empty string, not foo');
+
+  // loop through all possible kinds to verify
+  for(let key in AudioTrackEnums.AudioTrackKind) {
```
gkatsev: space after `for`.

src/js/tracks/video-track-enums.js
gkatsev: Also, Video and Audio tracks are the same right? Moving it into one file for all 3 sounds like a good choice to reduce duplication.

BrandonOCasey: Audio and Video Tracks share a lots of enums and are listed in the same list. Many of them are audio only and many of them our video only.

src/js/tech/html5.js
BrandonOCasey: I like the `supportsNativeTracks` function from #3166 but I think that the supportsNativeTracks function actually needs to be specific to each type of track. Right now in `supportsNativeTextTrack`  we check a lot of very specific things that may or may not apply to Video and Audio tracks. I could see merging the audio and video track functions but it may just end up being confusing that way as we would always have a specific text function.

src/js/tracks/video-track-list.js
BrandonOCasey: there is no selectedchange event defined in the spec (just as there is no modechange, or enabledchange). So native TrackLists will not do this.

gkatsev: No, what I mean is that the audio-track list which we're proxying already will be handling the `change` event for when audio tracks change. We don't want to handle the `selectedchange` stuff ourselves which will happen if we wrap native tracks, no? Or does the wrapping basically just grab the properties and then ignore the original track?

src/js/tracks/track.js
dmlap: If you don't define a setter, you get a read-only property. Are you saying we might forget they should be read-only and introduce a setter in the future?

gkatsev: Being explicit is nice but I don't have a preference one way or the other. I'm not sure why I initially went with an explicit no-op. If you feel like we should get rid of it, assuming it doesn't break anything, I'm ok with that.

src/js/tracks/video-track-enums.js
dmlap: @gkatsev the spec does not define an actual enumeration type for `kind`, there's just [a table of strings](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-TrackList-getKind-categories). Contrast that with the [network or ready states](https://html.spec.whatwg.org/multipage/embedded-content.html#htmlmediaelement) on the media element, which do have runtime constants.

I don't understand your point about having a default `kind`. Would you mind elaborating?

gkatsev: Ah, interesting, only TextTrack has actual defined `enum`: https://html.spec.whatwg.org/multipage/embedded-content.html#texttrackkind...

Specifically, we can do something like:
```js
let kind = AudioTrackKind[options.kind] || "";
```
I was thinking we could also technically add a `_default` property to the `Kind` object (with the prefixed `_` to make sure it will never conflict with the spec) and then we could do:
```js
let kind = AudioTrackKind[options.kind] || AudioTrackKind._default;
```

So, yeah, it's even less of a real enum than for text tracks, but I think having the object in a single file exported under two names will reduce code duplication and allow other code to be a bit neater.

src/js/tracks/video-track-list.js
BrandonOCasey: would it be better if the Track Class went away since it seems like all tracks don't share a common parent (inherit from EventTarget). Then we could take the code from track and just make it a function that sets up the common properties?

gkatsev: I think it's fine for us to have our emulated Tracks inherit from our EventTarget. We need a way to singal to the list that the track was updated.

gkatsev: Also, our internal structure doesn't necessarily need to align perfectly with the IDL laid out in the spec. The specific objects just need to look and act as described in the spec :)

src/js/tracks/video-track-enums.js
BrandonOCasey: So the consensus here is to change the AudioTrackEnum and VideoTrackEnum to just VideoTrackKind and AudioTrackKind?

gkatsev: I think we should have a `track-kinds.js` file which exports three objects `AudioTrackKind`, `VideoTrackKind`, and `TextTrackKind`. Inside the file `AudioTracKind` and `VideoTrackKind` point at the same object.

src/js/tracks/track.js
BrandonOCasey: it seems that removing the noop causes trying to set these variables to throw an error. Is that what we want to happen?

src/js/tracks/video-track-enums.js
BrandonOCasey: Would the TextTrackMode enum still go in it's own file in that case?

src/js/tracks/track.js
gkatsev: I think the values should just not change rather than throw an error.

src/js/tracks/video-track-enums.js
gkatsev: Is it used anywhere other than `text-track.js`? If not, we could move it there or `track-kinds.js` could just be called `track-enums.js` and move it there. Even though it isn't really enums.

src/js/tracks/track.js
BrandonOCasey: ok then we will have to keep the noop's

src/js/tracks/video-track-enums.js
BrandonOCasey: yeah lets do move everything to track-enums

test/unit/tracks/audio-track-list.test.js
BrandonOCasey: done

src/js/tracks/audio-track-list.js
BrandonOCasey: see https://github.com/videojs/video.js/pull/3173#discussion_r56169414

src/js/tracks/track.js
dmlap: Where are you seeing an error? [Works for me](http://jsbin.com/vuzuxabafu/1/edit?html,console,output) in Chrome.

BrandonOCasey: hmmm maybe its just an  issue with the unit tests then. When I removed them anywhere that we were trying to set something without a setter ended up throwing an error. Let me try removing them again.

BrandonOCasey: TypeError: Cannot set property kind of #<VideoTrack> which has only a getter

BrandonOCasey: Upon further investigation this seems to happen when we are running our code in strict mode. I just added ["use strict;"][0] to the example you posted @dmlap and we get the errors. 
[0]: http://jsbin.com/pitidapulu/1/edit?html,console

src/js/player.js
```diff
@@ -2473,6 +2479,42 @@ class Player extends Component {
...
+   * Get a video track list
+   * @link https://html.spec.whatwg.org/multipage/embedded-content.html#videotracklist
+   *
+   * @return {VideoTrackList} thes current video track list
+   * @method videoTracks
+   */
+  videoTracks() {
+    // if we have not yet loadTech_, we create videoTracks_
+    // these will be passed to the tech during loading
+    if(!this.tech_) {
```
gkatsev: space after `if`.

src/js/player.js
```diff
@@ -2473,6 +2479,42 @@ class Player extends Component {
...
+      this.videoTracks_ = this.videoTracks_ || new VideoTrackList();
+      return this.videoTracks_;
+    }
+
+    return this.tech_['videoTracks']();
```
gkatsev: just dot notation is fine.
`this.tech_.videoTracks`

src/js/player.js
```diff
@@ -2473,6 +2479,42 @@ class Player extends Component {
...
+   * Get an audio track list
+   * @link https://html.spec.whatwg.org/multipage/embedded-content.html#audiotracklist
+   *
+   * @return {AudioTrackList} thes current audio track list
+   * @method audioTracks
+   */
+  audioTracks() {
+    // if we have not yet loadTech_, we create videoTracks_
+    // these will be passed to the tech during loading
+    if(!this.tech_) {
```
gkatsev: space after `if`

src/js/player.js
```diff
@@ -2473,6 +2479,42 @@ class Player extends Component {
...
+      this.videoTracks_ = this.videoTracks_ || new AudioTrackList();
+      return this.videoTracks_;
+    }
+
+    return this.tech_ && this.tech_['audioTracks']();
```
gkatsev: dot notation, also, no need to guard against `this.tech_`.

src/js/tech/html5.js
```diff
@@ -96,24 +115,28 @@ class Html5 extends Tech {
...
+
+    // ProxyNativeTextTracks
+    for (let i = 0; i < trackTypes.length; i++) {
+      let type = trackTypes[i];
+      let capitalType = type.charAt(0).toUpperCase() + type.slice(1);
+      let tl = this.el_[`${type}Tracks`];
+      let etl = this[`${type}Tracks`]();
+
+      if (tl && tl.removeEventListener) {
+        tl.removeEventListener('change', this[`handle${capitalType}TrackChange_`]);
```
gkatsev: does this actually unbind video and audio track change handlers?
I know that the videojs bind and event system does some fancy stuff to handle this case but it won't work if you're dealing with the native event system.

src/js/tech/html5.js
```diff
@@ -70,6 +70,25 @@ class Html5 extends Tech {
       }
     }
 
+    let trackTypes = ['audio', 'video'];
+
+    // ProxyNativeTextTracks
+    for (let i = 0; i < trackTypes.length; i++) {
```
gkatsev: can you change it to `trackTypes.forEach()` instead?

src/js/tech/html5.js
```diff
@@ -96,24 +115,28 @@ class Html5 extends Tech {
...
-      tt.removeEventListener('addtrack', this.handleTextTrackAdd_);
-      tt.removeEventListener('removetrack', this.handleTextTrackRemove_);
-    }
-
-    // clearout the emulated text track list.
-    let i = emulatedTt.length;
+    let trackTypes = ['audio', 'video', 'text'];
+
+    // ProxyNativeTextTracks
+    for (let i = 0; i < trackTypes.length; i++) {
```
gkatsev: forEach instead please?

src/js/tech/tech.js
```diff
@@ -313,6 +320,33 @@ class Tech extends Component {
...
+
+  /**
+   * Initialize audio and video track listeners
+   *
+   * @method initTrackListeners
+   */
+  initTrackListeners() {
+    const typesTracks = ['video', 'audio'];
+
+    for (let i = 0; i < typesTracks.length; i++) {
```
gkatsev: forEach

src/js/tracks/audio-track-list.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    return list;
+  }
+
+
+  addTrack_(track) {
+    super.addTrack_(track);
+    // native tracks don't have this
+    if(track.addEventListener) {
```
gkatsev: space after `if`

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,77 @@
...
+ *   readonly attribute DOMString language;
+ *   attribute boolean enabled;
+ * };
+ *
+ * @param {Object=} options Object of option names and values
+ * @class AudioTrack
+ */
+class AudioTrack extends Track {
+  constructor(options = {}) {
+      let settings = merge(options, {
```
gkatsev: indented too much

src/js/tracks/audio-track-list.js
```diff
@@ -0,0 +1,62 @@
...
+
+    if (browser.IS_IE8) {
+      for (let prop in AudioTrackList.prototype) {
+        if (prop !== 'constructor') {
+          list[prop] = AudioTrackList.prototype[prop];
+        }
+      }
+    }
+
+    return list;
```
gkatsev: I think it would still be good to only return `list` if it is IE8

src/js/tracks/audio-track.js
```diff
@@ -0,0 +1,77 @@
...
+
+    // if the user sets this track to selected then
+    // set selected to that true value otherwise
+    // we keep it false
+    if (settings.enabled) {
+      audioTrack.enabled = settings.enabled;
+    }
+    audioTrack.loaded_ = true;
+
+    return audioTrack;
```
gkatsev: only return if IE8

src/js/tracks/text-track-list.js
```diff
@@ -40,124 +37,14 @@ class TextTrackList extends EventTarget {
...
-    });
-
-    for (let i = 0; i < tracks.length; i++) {
-      list.addTrack_(tracks[i]);
-    }
-
-    if (browser.IS_IE8) {
-      return list;
-    }
+    return list;
```
gkatsev: only return if IE8

src/js/tracks/text-track.js
```diff
@@ -102,21 +101,25 @@ const loadTrack = function(src, track) {
...
+class TextTrack extends Track {
   constructor(options = {}) {
-    super();
-    if (!options.tech) {
-      throw new Error('A tech was not provided.');
-    }
-
-    let tt = this;
+    let settings = merge(options, {
+      trackType: 'text',
```
gkatsev: What's `trackType`?

src/js/tracks/text-track.js
```diff
@@ -255,16 +218,14 @@ class TextTrack extends EventTarget {
...
+      tt.src = settings.src;
+      loadTrack(settings.src, tt);
     } else {
       tt.loaded_ = true;
     }
 
-    if (browser.IS_IE8) {
-      return tt;
-    }
+    return tt;
```
gkatsev: only if in IE8

src/js/tracks/track.js
```diff
@@ -0,0 +1,55 @@
...
+      });
+    }
+    return track;
```
gkatsev: only in IE8 :)

src/js/tracks/video-track-list.js
```diff
@@ -0,0 +1,74 @@
...
+
+    if (browser.IS_IE8) {
+      for (let prop in VideoTrackList.prototype) {
+        if (prop !== 'constructor') {
+          list[prop] = VideoTrackList.prototype[prop];
+        }
+      }
+    }
+
+    return list;
```
gkatsev: only in IE8

src/js/tracks/text-track.js
BrandonOCasey: leftover from some previous change will get rid of that

src/js/tracks/audio-track-list.js
BrandonOCasey: is there any problem with always returning list though? the reason I ask is because if we only do it in the parent class on IE8 then we won't be able to do just this: 

``` JavaScript
  let track = super(settings);
```

We will have to do: 
``` JavaScript
  let track = super(settings);
  if(!track) {
    track = this;
  }
```

gkatsev: Ok, yeah, I don't think the benefits of wrapping each one in an IE8 block is worth it. So, ignore all those comments.

src/js/tech/html5.js
BrandonOCasey: I think these tests should cover that https://github.com/BrandonOCasey/video.js/blob/integrate-video-audio-track/test/unit/tech/html5.test.js#L253

gkatsev:
This is available on npm under version `5.9.0-0` or the `alpha` tag right now.

BrandonOCasey:
@gkatsev should I continue to make changes in this PR as they come up?

gkatsev:
@BrandonOCasey yes. Also, when you get a chance, rebasing against master would be good.

BrandonOCasey:
@gkatsev I think this feature is ready for a final review whenever you get the time

gkatsev:
@BrandonOCasey cool. Do you need another prerelease?

BrandonOCasey:
@gkatsev Yeah that may be a good first step

src/js/tracks/audio-track-list.js
```diff
@@ -0,0 +1,113 @@
...
+      // when we are disabling other tracks (since we don't support
+      // more than one track at a time) we will set changing_
+      // to true so that we don't trigger additional change events
+      if (this.changing_) {
+        return;
+      }
+      this.changing_ = true;
+      disableOthers(this, track);
+      this.trigger('change');
+      this.changing_ = false;
```
imbcmdth: I think lines 98 & 99 should be reversed. The way it is now, if an event handler for the 'change' event decided to alter the `enabled` audio track, we wouldn't emit any more 'change' events because `changing_` is still true. We should be done changing and back to normal before we trigger 'change'.

BrandonOCasey: that makes sense I will change it in video and audio track list

gkatsev:
@BrandonOCasey 5.9.0-1 is up, tagged alpha.

BrandonOCasey:
@gkatsev Rebased on master again and added some unit tests for track disposal in tech. I think we need another pre-release

BrandonOCasey:
@gkatsev okay we are really ready for another pre-release this time. Crossing my fingers that this is the last one.

test/unit/tech/tech.test.js
```diff
@@ -14,20 +21,10 @@ q.module('Media Tech', {
     this.clock = sinon.useFakeTimers();
     this.featuresProgessEvents = Tech.prototype['featuresProgessEvents'];
     Tech.prototype['featuresProgressEvents'] = false;
-    Tech.prototype['featuresNativeTextTracks'] = true;
```
gkatsev: why remove these?

src/js/tech/tech.js
```diff
@@ -237,26 +238,26 @@ class Tech extends Component {
...
-   * empty all of our track lists
+   * empty all of our video/audio track lists
    *
    * @private
    * @method emptyTrackLists_
+   * @param {Array|String} types type of tracks list to empty
    */
-  emptyTrackLists_() {
+  emptyTrackLists_(types) {
+    if (typeof types === 'string') {
```
gkatsev: you can do `types = [].concat(types)` and you won't need the if.

src/js/tech/tech.js
```diff
@@ -805,6 +806,12 @@ Tech.withSourceHandlers = function(_Tech){
     this.disposeSourceHandler();
     this.off('dispose', this.disposeSourceHandler);
 
+    // if we have a source and get another one
```
gkatsev: what if you have a tech that isn't a source handler get a new source? Will the these get cleared out then?

test/unit/tech/tech.test.js
BrandonOCasey: we may want to add tests for native text tracks (which these are mocking up) but they currently don't exist and I needed to test our implementation of text tracks.

src/js/tech/tech.js
BrandonOCasey: it seems like it should I will have to work on adding a test for that case, if you think that its a potential scenario

gkatsev: I think it would be good to handle.

gkatsev:
A new pre-release is now out as 5.9.0-2 (even though it'll go out as 5.10).

@benjipott hey, do you want to take another look at this PR? Would love to get some comments from you, especially if you're able to give the feature a shot.

src/js/tech/html5.js
```diff
@@ -96,25 +114,20 @@ class Html5 extends Tech {
...
-    // clearout the emulated text track list.
-    let i = emulatedTt.length;
-
-    while (i--) {
-      emulatedTt.removeTrack_(emulatedTt[i]);
-    }
-
+    // Un-ProxyNativeTracks
+    ['audio', 'video', 'text'].forEach((type) => {
+      let capitalType = type.charAt(0).toUpperCase() + type.slice(1);
```
benjipott: you can use videojs Utils.toTitleCase

src/js/tech/html5.js
```diff
@@ -14,6 +14,7 @@ import document from 'global/document';
 import window from 'global/window';
 import assign from 'object.assign';
 import mergeOptions from '../utils/merge-options.js';
+import TitleCase from '../utils/to-title-case.js';
```
benjipott: import toTitleCase from '../../utils/to-title-case.js'; (same as all imports in videojs)
https://github.com/videojs/video.js/search?utf8=%E2%9C%93&q=toTitleCase

benjipott:
I think tech should implement addVideoTrack addAudioTrack
if dashjs/videojs plugin implement audio switch button.
https://github.com/BrandonOCasey/video.js/blob/integrate-video-audio-track/src/js/tech/tech.js#L492

benjipott:
good job :+1: :)

gkatsev:
@benjipott we have `addTrack` as public on the AudioTrackList and VideoTrackList.
Also, cool! We have a PR (I think it's outdated now? @forbesjo) for adding the audio selection menu to videojs: https://github.com/videojs/video.js/pull/3223

src/js/tech/tech.js
```diff
@@ -33,10 +34,20 @@ import document from 'global/document';
 class Tech extends Component {
 
   constructor(options={}, ready=function(){}){
+    let defaults = {
```
gkatsev: I don't quite understand what this object is used for. How is this different from the `features` thing?

src/js/tech/tech.js
```diff
@@ -62,7 +69,17 @@ class Tech extends Component {
...
+      let capitalType = toTitleCase(type);
+
+      if (settings[`native${capitalType}Tracks`] === false) {
+        this[`featuresNative${capitalType}Tracks`] = false;
+      }
+
+      this[`${type}Tracks_`] = settings[`${type}Tracks`];
+    });
+
+    if (settings.nativeCaptions === false) {
```
gkatsev: this should be checking both `nativeCaptions` and `nativeTextTracks`

BrandonOCasey: I check nativeCaptions below the loop. Since its sort of a special case and it was in an or previously.

src/js/tech/tech.js
BrandonOCasey: by default all of the features are on if they are supported. If the user wants to then disable it, even if the feature was natively supported they could. I guess we don't really need defaults if we just want to check wether the option to keep nativeX on was passed in.

gkatsev: Oh, ok, I see now. It's just the flag to turn these things off.
Can HLS turn off native audio tracks early enough for that to be sufficient for it's usecase?

src/js/tech/tech.js
gkatsev: oh, I see. Sounds good then.

src/js/tech/tech.js
BrandonOCasey: we may have to set the option on videojs while we are loading HLS, but I think that we should be able to use this. Regardless the functionality was there for text tracks and this will mimic that. We may have another PR in the future to expose it further for HLS.

gkatsev: Sounds good to me.

gkatsev:
This PR possibly broke text tracks: https://github.com/videojs/video.js/issues/3263

src/js/tech/flash.js
dmlap: This issue needs to be resolved before this can be merged.

src/js/tech/html5.js
```diff
@@ -70,6 +71,24 @@ class Html5 extends Tech {
       }
     }
 
+    let trackTypes = ['audio', 'video'];
+
+    // ProxyNativeTextTracks
```
dmlap: It's a little weird this comment is in camel-case :)

src/js/tech/html5.js
```diff
@@ -70,6 +71,24 @@ class Html5 extends Tech {
...
+    trackTypes.forEach((type) => {
+      let capitalType = toTitleCase(type);
+
+      if (!this[`featuresNative${capitalType}Tracks`]) {
+        return;
+      }
+      let tl = this.el()[`${type}Tracks`];
+
+      if (tl && tl.addEventListener) {
+        tl.addEventListener('change', Fn.bind(this, this[`handle${capitalType}TrackChange_`]));
```
dmlap: I think you can use `this.on(target, event, listener)` to add these listeners with the added benefit they will be automatically disposed alongside the tech.

src/js/tech/tech.js
```diff
@@ -313,6 +341,32 @@ class Tech extends Component {
...
+    const trackTypes = ['video', 'audio'];
+
+    trackTypes.forEach((type) => {
+      let trackListChanges = () => {
+        this.trigger(`${type}trackchange`);
+      };
+
+      let tracks = this[`${type}Tracks`]();
+
+      tracks.addEventListener('removetrack', trackListChanges);
```
dmlap: Same comment from above about `this.on(elem, ...)`

src/js/tracks/track-enums.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+const TextTrackMode = {
+  disabled: 'disabled',
+  hidden: 'hidden',
+  showing: 'showing'
+};
+
+/* jshint ignore:start */
+// we ignore jshint here because it does not see
+// AudioTrackKind as defined here
```
dmlap: That's bizarre. Is there a bug in jshint for this issue?

src/js/tech/html5.js
gkatsev: tl could be a native object so, it's better/safer to use addEventListener.

src/js/tracks/track-enums.js
gkatsev: jshint (at least the version we use) doesn't understand ES6 very well. At some point, we'll have video.js's repo use vjsstandard.

src/js/tracks/video-track-list.js
```diff
@@ -0,0 +1,124 @@
+/**
+ * @file video-track-list.js
+ */
+import TrackList from './track-list';
+import * as browser from '../utils/browser.js';
+import document from 'global/document';
+
+/**
+ * anywhere we call this function we diverge from the spec
+ * as we only support one enabled audiotrack at a time
```
dmlap: copy-paste discrepancies below. Why is this duplicated?

dmlap:
@gkatsev @BrandonOCasey there are still a number of unaddressed comments in the PR. Removing the "needs: LGTM" for now.

src/js/tracks/video-track-list.js
BrandonOCasey: one needs to use selected (VideoTrack) and the other needs to use enabled (AudioTrack)

src/js/tech/tech.js
BrandonOCasey: I think Gary makes a good point above

src/js/tech/flash.js
BrandonOCasey: shouldn't Flash inheriting from Tech give us the methods that return an empty track list?

dmlap: Ah, this is a removal, not an addition. All good.

src/js/tech/html5.js
dmlap: Follow-up: it's better/safer to use `this.on(trackList, 'event', listener)` __but__  the current implementation of on() only works with Components or DOM nodes. I'm going to file an issue to fix that so we don't have to worry about manually cleaning up listeners for any EventTarget-like object.  We can stick with this for now.

dmlap:
LGTM

src/js/tracks/text-track.js
misteroneill: Was this still the case? Should this have a conditional around it or was there some other resolution here?

src/js/tech/tech.js
misteroneill: Was this question/concern resolved?

src/js/tracks/text-track.js
BrandonOCasey: I think we decided that it did not hurt anything to always return an instance of the object or on IE8 a dom object

src/js/tech/tech.js
BrandonOCasey: This code was pulled out, as we did not have a valid use case for it. I re-wrote history to do it so this comment still shows up.

misteroneill: 👍

src/js/tracks/text-track.js
misteroneill: 👍

misteroneill:
LGTM

gkatsev:
And tests passed in browserstack: https://travis-ci.org/videojs/video.js/builds/125066414

PR #3181, opened 2016-03-15
move timeout back into parseCues

gkatsev:
## Description
Make sure that vttjs has finished loading before proceeding with parsing cues.

## Specific Changes proposed
Specifically, move the code that does a setTimeout from `loadTrack` back into `parseCues`. Make `parseCues` into an asynchronous loop waiting till vttjs has loaded.

## Requirements Checklist
- [x] Reviewed by one Core Contributors

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -19,12 +19,22 @@ import XHR from 'xhr';
  * @param {String} srcContent webVTT file contents
  * @param {Track} track track to addcues to
  */
-const parseCues = function(srcContent, track) {
+const parseCues = function(srcContent, track, retryNum) {
   // Make sure that vttjs has loaded, otherwise, wait till it finished loading
   // NOTE: this is only used for the alt/video.novtt.js build
   if (typeof window.WebVTT !== 'function') {
+    let retry;
+
+    if (retryNum === 0) {
+      return log.error(`vttjs did not load, stopping trying to process ${track.src}`);
+    } else if (typeof retryNum !== 'number') {
+      retry = 3;
+    } else {
+      retry = retryNum - 1;
+    }
+
     return window.setTimeout(function() {
-      parseCues(srcContent, track);
+      parseCues(srcContent, track, retry);
     }, 100);
   }
 
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -71,7 +71,13 @@ const loadTrack = function(src, track) {
     // NOTE: this is only used for the alt/video.novtt.js build
     if (typeof window.WebVTT !== 'function') {
       if (track.tech_) {
-        track.tech_.on('vttjsloaded', () => parseCues(responseBody, track));
+        let loadHandler = () => parseCues(responseBody, track);
+        track.tech_.on('vttjsloaded', loadHandler);
+        track.tech_.on('vttjserror', () => {
+          log.error(`vttjs failed to load, stopping processing ${track.src}`);
+          track.tech_.off('vttjsloaded', loadHandler);
+        });
+
       }
     } else {
       parseCues(responseBody, track);
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -19,25 +19,7 @@ import XHR from 'xhr';
  * @param {String} srcContent webVTT file contents
  * @param {Track} track track to addcues to
  */
-const parseCues = function(srcContent, track, retryNum) {
-  // Make sure that vttjs has loaded, otherwise, wait till it finished loading
-  // NOTE: this is only used for the alt/video.novtt.js build
-  if (typeof window.WebVTT !== 'function') {
-    let retry;
-
-    if (retryNum === 0) {
-      return log.error(`vttjs did not load, stopping trying to process ${track.src}`);
-    } else if (typeof retryNum !== 'number') {
-      retry = 3;
-    } else {
-      retry = retryNum - 1;
-    }
-
-    return window.setTimeout(function() {
-      parseCues(srcContent, track, retry);
-    }, 100);
-  }
-
+const parseCues = function(srcContent, track) {
   let parser = new window.WebVTT.Parser(window,
                                         window.vttjs,
                                         window.WebVTT.StringDecoder());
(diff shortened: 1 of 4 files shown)


gkatsev:
Going to see if I can add a test for this so it won't happen again :)

gkatsev:
@chemoish can you review this one? You probably know the most about how tracks load next to me.

chemoish:
Does video.js have pub sub or anything we can leverage to listen to events?

I would prefer…

```
if (!window['WebVTT'] && this.el().parentNode != null) {
  let script = document.createElement('script');
  script.src = this.options_['vtt.js'] || '../node_modules/videojs-vtt.js/dist/vtt.js';
  script.onload = function () {
    console.log('load vtt.js');
  };
  this.el().parentNode.appendChild(script);
  window['WebVTT'] = true;
}
```

Instead of having a `setTimeout`.

src/js/tracks/text-track.js
```diff
@@ -20,6 +20,14 @@ import XHR from 'xhr';
  * @param {Track} track track to addcues to
  */
 const parseCues = function(srcContent, track) {
+  // Make sure that vttjs has loaded, otherwise, wait till it finished loading
+  // NOTE: this is only used for the alt/video.novtt.js build
+  if (typeof window.WebVTT !== 'function') {
+    return window.setTimeout(function() {
```
chemoish: If `window.WebVTT` never loads, this is infinite loop?

src/js/tracks/text-track.js
```diff
@@ -20,6 +20,14 @@ import XHR from 'xhr';
  * @param {Track} track track to addcues to
  */
 const parseCues = function(srcContent, track) {
+  // Make sure that vttjs has loaded, otherwise, wait till it finished loading
+  // NOTE: this is only used for the alt/video.novtt.js build
+  if (typeof window.WebVTT !== 'function') {
```
chemoish: Should this block `return;`?

Otherwise line 31 [https://github.com/videojs/video.js/pull/3181/files#diff-45ae6eaab623b37642b347fe7707682aR31] will fire an exception.

screen shot 2016-03-15 at 2 06 54 pm

src/js/tracks/text-track.js
gkatsev: Yes, it would. It should have a retry count, I guess.
Maybe changing it to send an event we can listen to is a better way of doing it, I'll see if I can do it quickly, I'd like to get this change in as quickly as possible.

src/js/tracks/text-track.js
gkatsev: Does it not return now? See `return window.setTimeout...`

chemoish: Must have not pulled that in… anyways should be good with return of some kind.

chemoish:
@gkatsev since this is a patch I am ok with this going in and fixing it later to support retry or pubsub.

I didn't step through the tests, but LGTM.

src/js/tracks/text-track.js
```diff
@@ -19,7 +19,25 @@ import XHR from 'xhr';
...
+    } else if (typeof retryNum !== 'number') {
+      retry = 3;
+    } else {
+      retry = retryNum - 1;
+    }
+
+    return window.setTimeout(function() {
+      parseCues(srcContent, track, retry);
+    }, 100);
+  }
```
imbcmdth: Is 400ms enough time to wait? Is there any guarantee that will be enough time on slow connections or heavily loaded pages?

gkatsev: I'm not sure, I guess on really slow pages or really slow connections, it will die.
Maybe we want up to 10 retries?

chemoish: Add `// TODO: listen to onload event or similar` (:

src/js/tracks/text-track.js
```diff
@@ -19,7 +19,25 @@ import XHR from 'xhr';
...
+
+    if (retryNum === 0) {
+      return log.error(`vttjs did not load, stopping trying to process ${track.src}`);
+    } else if (typeof retryNum !== 'number') {
+      retry = 3;
+    } else {
+      retry = retryNum - 1;
+    }
+
+    return window.setTimeout(function() {
```
dmlap: Why move the retry logic here from loadTrack()?

gkatsev: Because if it's in loadTrack it only ever gets called once and doesn't keep waiting for vttjs to load.

src/js/tracks/text-track.js
gkatsev: I could also have it do incremental back-off, 100 for first, 200 for second, 400 for 3rd, 800 for second. Total would be 1.5s.
Assuming we don't just do the onload.

src/js/tracks/text-track.js
```diff
@@ -19,7 +19,25 @@ import XHR from 'xhr';
  * @param {String} srcContent webVTT file contents
  * @param {Track} track track to addcues to
  */
-const parseCues = function(srcContent, track) {
+const parseCues = function(srcContent, track, retryNum) {
```
BrandonOCasey: should retryNum default to 0 here since we are not passing it in the first time?

gkatsev: actually, defaulting it to 3 (or 10) would make more sense plus get rid of some of the shenanigans I'm doing in lines 28-34.

src/js/tracks/text-track.js
gkatsev: @chemoish looks like it's not as urgent as other things, so, I'll spend time making it wait for vttjs to load instead of retrying :)

gkatsev:
Updated the code to wait for the load event of the script and have loadTrack delay parseCues until such an event. What's left is to have those event handlers be removed if the script fails to load and also test updates.

test/unit/tracks/text-track.test.js
```diff
@@ -254,3 +257,124 @@ test('fires cuechange when cues become active and inactive', function() {
...
+  xhr.respond(200, {}, 'WebVTT\n');
+
+  ok(!parserCreated, 'WebVTT is not loaded, do not try to parse yet');
+
+  testTech.trigger('vttjserror');
+  let errorSpyCall = log.error.getCall(0);
+  let offSpyCall = testTech.off.getCall(0);
+
+  ok(errorSpyCall.calledWithMatch('vttjs failed to load, stopping trying to process'),
+     'we logged the correct error after 3 timeouts');
```
dmlap: This message is outdated

dmlap:
Other than the update to the test message, LGTM

PR #3189, opened 2016-03-17
XHR and updatedisplay fixes for IE8

gkatsev:


src/js/tech/tech.js
```diff
@@ -333,6 +333,10 @@ class Tech extends Component {
       script.onerror = () => {
         this.trigger('vttjserror');
       };
+      this.on('dispose', () => {
```
dmlap: Dope.

dmlap:
Could use a little more exposition around the use of `proxyrequire` but otherwise LGTM

dmlap:
LGTM

Issue #3121, opened 2016-02-18
Announce Big Play Button on first tab into player on IE9 with JAWS


marguinbc:
The Big Play button is not being announced when using JAWS (version 15/16/17) on IE9. 

When you first tab into the player, the big play button should receive focus and be announced by JAWS. Somehow it is announced when you tab out and back into the BPB (shift+tab).  It also appears to work on other versions of IE and JAWS.

Confirmed this on VideoJS 5.7.1 on IE9 and JAWS 16

OwenEdwards:
@marguinbc is there a specific page this was happening on, or do you have a reduced test case?

I've been doing some investigation of this, and I *think* it's related to some weirdness around the fallback line inside the video element itself:

`<p class="vjs-no-js">To view this video please enable JavaScript, and consider upgrading to a web browser that <a href="http://videojs.com/html5-video-support/" target="_blank">supports HTML5 video</a></p>`

It seems like the link receives focus, but isn't announced correctly; then the BPB receives focus, and is announced, but then when focus moves back to the link (Shift-Tab), it is announced as the BPB.

OwenEdwards:
@marguinbc see #3187.

marguinbc:
So, it seems that the real issue is what you described in 3187 and the vjs-no-js div causing the issue on IE. If so, I'm thinking this can be closed.

PR #3194, opened 2016-03-22
Fix link adding an extra Tab stop (and confusing JAWS) with IE

### LINKED ISSUE #3187: The link in the <p class="vjs-no-js"> paragraph is not hidden from screen readers by IE9-11
## Description

The documentation recommends adding the line `<p class="vjs-no-js">To view this video please enable JavaScript, and consider upgrading to a web browser that <a href="http://videojs.com/html5-video-support/" target="_blank">supports HTML5 video</a></p>` inside the video tag - this CSS was added in 227700a2367657ecfe63b5af39df1bc7f9b2529b, with the following instructions:

````
/*  In IE8 w/ no JavaScript (no HTML5 shim), the video tag doesn't register.
    The .video-js classname on the video tag also isn't considered.
    This optional paragraph inside the video tag can provide a message to users
    about what's required to play video. */
````

Once video.js runs, that text is no longer visible, but the link is still in the keyboard focus (Tab) order for screen reader users. The text of the link seems to not be recognized correctly, so the link is announced as the previous focused element.

 ## Steps to reproduce
Using JAWS (16) on IE9, but also on IE11. On the http://www.videojs.com page, tab through the page to the big play button. Notice that there's a focus element before the BPB, and if you Shift-Tab back to it, it is announced as the BPB.

This link should be removed from the Tab order once video.js loads. (Note that `aria-hidden` does not remove content from the focus order - see https://www.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/ for preferred ways to hide this content from all users).

## Additional Information
#### videojs
v5.8.0
#### browsers
IE9, IE11 (Not tested on others)
#### OSes
Windows 7
### plugins
None

OwenEdwards:
## Description
Fix link adding an extra Tab stop (and confusing JAWS) with IE - fixes #3187 and #3121.

## Specific Changes proposed
Actually hide any links within the `<video>` tag, from Javascript, when video.js creates a new player.

## Requirements Checklist
- [x] Bug fixed

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -297,6 +297,14 @@ class Player extends Component {
     this.fluid(this.options_.fluid);
     this.aspectRatio(this.options_.aspectRatio);
 
+    // Hide any links within the video/audio tag, because IE doesn't hide them completely.
+    let links = tag.getElementsByTagName('a');
+    for (let i = 0; i < links.length; i++) {
+      let linkEl = links.item(i);
+      Dom.addElClass(linkEl, 'vjs-hidden');
+      linkEl.setAttribute('hidden', 'hidden');
+    }
+
     // insertElFirst seems to cause the networkState to flicker from 3 to 2, so
     // keep track of the original for later so we can know if the source originally failed
     tag.initNetworkState_ = tag.networkState;


gkatsev:
LGTM

PR #3213, opened 2016-03-28
Don't use icon font for time tooltips

### LINKED ISSUE #3210: Video.js uses default browser font on progress timers
As shown in the screenshot with the browser zoomed in, it uses the default font in the browser (which is obviously Times New Roman), which is inconsistent with the other font used in the far left, which is Arial.

wow

Obviously, this makes it much harder to read it since Times New Roman isn't designed to be read in such small print.  I also think you should make the font size slightly bigger overall.

mister-ben:
## Description
Prevents the browser showing time tooltips in a fallback font.
Fixes #3210

## Specific Changes proposed
Uses the regular text font rather than VideoJS for the time tooltips. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -42,7 +42,7 @@
 .video-js .vjs-progress-control:hover .vjs-time-tooltip,
 .video-js .vjs-progress-control:hover .vjs-mouse-display:after,
 .video-js .vjs-progress-control:hover .vjs-play-progress:after {
-  font-family: VideoJS;
+  font-family: $text-font-family;
   visibility: visible;
   font-size: 0.6em;
 }


gkatsev:
LGTM, though, I would like to see if there's some other style tweaks we would want to do to make sure it looks good.

gkatsev:
Also, could you rebase against stable? Thanks.

gkatsev:
ack, that's weird. I wonder why it didn't rebase cleanly...
Anyway, I guess I could just cherry-pick the commit manually.

mister-ben:
I reverted to from master while I was trying to figure that out. That file has quite a few changes between stable and master.

gkatsev:
Oh, right. Some of the changes are from the keep tooltips inside the player work. I guess we'll just merge it into master then.

PR #3215, opened 2016-03-28
use a brand new tag el for ie8

gkatsev:
Fix the tests on ie8. We need a new tag for IE8 and not reuse it when we create a new player.


misteroneill:
LGTM

Issue #3137, opened 2016-02-24
Add Audio/Video Track support to video.js


BrandonOCasey:
## Purpose
To extend the capabilities of video.js so that we can support multiple audio and video tracks much like we support multiple text tracks right now.

## Proposal
* Follow the embedded content specification to:
	* implement the sharing of common functionality between TextTracks, AudioTracks, and VideoTracks.
	* implement VideoTrack, VideoTrackList, and enumerations
	* implement AudioTrack, AudioTrackList, and enumerations
	* possibly implement HTMLSourceElement
* Add functionality to video.js to utilize the new track lists in api

## API Spec
The main portions of the API will be defined in accordance with [embedded content specification][a]
Integration with video.js with follow the current patterns that TextTrack has laid out where it makes sense.

## Acceptance Criteria Most Covered by #3173 
1. [x] Convert Current TextTrack functionality to use ES6 syntax - (#3130) 
1. [x] ~~move `tracks/text-track*` files into `tracks/text/text-track*` - (#3146)~~ Not Doing
1. [x] Convert text track related unit test to full es6 syntax - (#3148) 
1. [x] Move common functionality from TextTrackList so that it can be reused by other TracksLists (#3173)
1. [x] Move common functionality from TextTrack so that it can be reused by other Tracks (#3173)
1. [x] ~~Research and implement the [HTMLSourceElement][b] api if required~~ Not Doing
1. [x] Follow the [AudioTrack][1] specification and reuse existing functionality to create the following: (#3173) 
  - [x] [AudioTrack][1]
  - [x] [AudioTrackList][2]
  - [x] [AudioTrackEnum][3]
1. [x] Follow the [VideoTrack][4] specification and reuse existing functionality to create the following: (#3173)
  - [x] [VideoTrack][4]
  - [x] [VideoTrackList][5]
  - [x] [VideoTrackEnum][6]
1. [x] Add audioTracks getter to: (#3173)
  - [x] [player][7]
  - [x] [tech][8]
  - [x] [html5][9]
1. [x] Add videoTracks getter to player and tech (#3173)
  - [x] [player][7]
  - [x] [tech][8]
  - [x] [html5][9]

[a]: https://html.spec.whatwg.org/multipage/embedded-content.html
[b]: https://html.spec.whatwg.org/multipage/embedded-content.html#htmlsourceelement
[1]: https://html.spec.whatwg.org/multipage/embedded-content.html#audiotrack
[...]

PR #3223, opened 2016-03-29
Audio button

forbesjo:
This PR adds an audio button for audio track selection. It depends on ~~https://github.com/videojs/font/pull/13~~ and ~~#3137~~

### DIFF
--- a/src/js/control-bar/audio-track-controls/audio-track-button.js
+++ b/src/js/control-bar/audio-track-controls/audio-track-button.js
@@ -1,0 +1,64 @@
+/**
+ * @file audio-track-button.js
+ */
+import TrackButton from '../track-button.js';
+import Component from '../../component.js';
+import * as Fn from '../../utils/fn.js';
+import AudioTrackMenuItem from './audio-track-menu-item.js';
+
+/**
+ * The base class for buttons that toggle specific text track types (e.g. subtitles)
+ *
+ * @param {Player|Object} player
+ * @param {Object=} options
+ * @extends TrackButton
+ * @class AudioTrackButton
+ */
+class AudioTrackButton extends TrackButton {
+  constructor(player, options = {}) {
+    options.tracks = player.audioTracks && player.audioTracks();
+
+    super(player, options);
+
+    this.el_.setAttribute('aria-label', 'Audio Menu');
+  }
+
+  /**
+   * Allow sub components to stack CSS class names
+   *
+   * @return {String} The constructed class name
+   * @method buildCSSClass
+   */
+  buildCSSClass() {
+    return `vjs-audio-button ${super.buildCSSClass()}`;
+  }
+
+  /**
+   * Create a menu item for each audio track
+   *
+   * @return {Array} Array of menu items
+   * @method createItems
+   */
+  createItems(items = []) {
+    let tracks = this.player_.audioTracks && this.player_.audioTracks();
+
+    if (!tracks) {
+      return items;
+    }
+
+    for (let i = 0; i < tracks.length; i++) {
+      let track = tracks[i];
+
+      items.push(new AudioTrackMenuItem(this.player_, {
+        // MenuItem is selectable
+        'selectable': true,
+        'track': track
+      }));
+    }
+
+    return items;
+  }
+}
+
+Component.registerComponent('AudioTrackButton', AudioTrackButton);
+export default AudioTrackButton;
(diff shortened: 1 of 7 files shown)


src/js/control-bar/audio-track-controls/audio-track-button.js
```diff
@@ -0,0 +1,59 @@
+import MenuButton from '../../menu/menu-button.js';
+import Component from '../../component.js';
+import * as Fn from '../../utils/fn.js';
+import AudioTrackMenuItem from './audio-track-menu-item.js';
+
+class AudioTrackButton extends MenuButton {
```
gkatsev: what about having this button extend `TrackButton` and refactor `TextTrackButton` into a `TrackButton`? it already does the `removetrack` `addtracker` listening.

forbesjo: Sounds like a good idea.

src/js/control-bar/audio-track-controls/audio-track-button.js
```diff
@@ -0,0 +1,64 @@
...
+ * The base class for buttons that toggle specific text track types (e.g. subtitles)
+ *
+ * @param {Player|Object} player
+ * @param {Object=} options
+ * @extends TrackButton
+ * @class AudioTrackButton
+ */
+class AudioTrackButton extends TrackButton {
+  constructor(player, options = {}) {
+    options.tracks = player.audioTracks && player.audioTracks();
```
gkatsev: is this used anywhere?

src/js/control-bar/audio-track-controls/audio-track-button.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /**
+   * Create a menu item for each audio track
+   *
+   * @return {Array} Array of menu items
+   * @method createItems
+   */
+  createItems(items = []) {
+    let tracks = this.player_.audioTracks && this.player_.audioTracks();
```
gkatsev: why check that it exists? It *should* always be present.

src/js/control-bar/audio-track-controls/audio-track-button.js
gkatsev: yes, it is, in the TrackButton, missed it.

src/js/control-bar/audio-track-controls/audio-track-button.js
gkatsev: If you made a mock player to use with custom controls, this would fail, so, I guess we should keep it around for now, maybe remove it in a major bump.

src/js/control-bar/audio-track-controls/audio-track-button.js
```diff
@@ -0,0 +1,64 @@
...
+      return items;
+    }
+
+    for (let i = 0; i < tracks.length; i++) {
+      let track = tracks[i];
+
+      items.push(new AudioTrackMenuItem(this.player_, {
+        // MenuItem is selectable
+        'selectable': true,
+        'track': track
```
misteroneill: This object literal could be simplified to:

```js
{
  track,
  selectable: true
}
```

misteroneill: To elaborate, I feel like we should start writing code that will pass the video.js linter even if the project at large won't pass it.

src/js/control-bar/audio-track-controls/audio-track-menu-item.js
```diff
@@ -0,0 +1,71 @@
...
+ * @class AudioTrackMenuItem
+ */
+class AudioTrackMenuItem extends MenuItem {
+  constructor(player, options) {
+    let track = options.track;
+    let tracks = player.audioTracks();
+
+    // Modify options for parent MenuItem class's init.
+    options.label = track.label || track.language || 'Unknown';
+    options.selected = track.enabled;
```
misteroneill: In `AudioTrackButton` it's passing `selected` not `enabled` so I wonder if this line is correct...

misteroneill: Or if the `selected: true` above is pointless.

forbesjo: I'm not following, AudioTrackButton passes `selectable: true` to populate the menu with clickable menu items. It also passes track which has an `enabled` property which is really the currently active track. If it is active we want to indicate that it is selected by setting AudioTrackMenuItem's `selected` option.

misteroneill: It's because I am apparently illiterate. I mis-read `selectable` as `selected`. Ignore me. 😄

misteroneill:
LGTM

PR #3226, opened 2016-03-29
Fix 3140

seescode:
This fixes 3140 where control bar hides certain elements prematurely in IE.


gkatsev:
Does this make sass not turn the string into a utf8 character? If so, I wonder if we could do the same thing for our font and then we could get rid of the "wrapcodepoints" grunt script.

seescode:
image

It just outputs \0screen in the generated dist/video-js.css

gkatsev:
Cool, that's what we want to happen.

gkatsev:
Awesome, LGTM.

misteroneill:
LGTM

PR #3227, opened 2016-03-30
Fixes #3201 Add ARIA region role and ARIA label to player el

### LINKED ISSUE #3201: Add ARIA region role and ARIA label to player element
If a document contains multiple players, screen readers can inadvertently combine some of the controls. A specific example occurs when using JAWS in IE 9+. 

The solution is to add a "role" attribute with the value "region" to the player element, and also add an ARIA label identifying the player as an "audio player" or "video player", depending on the type. This has the benefit of adding the player to the screen reader navigation, and also fixing any bugs with the controls.

chrisauclair:
Fixes #3201. Fixes bug where JAWS in IE can't access all player controls if a document contains multiple players. Adds `role` of `region` and `aria-label` identifying media player type. Removes "video" from progress slider label.

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -35,7 +35,7 @@ class SeekBar extends Slider {
     return super.createEl('div', {
       className: 'vjs-progress-holder'
     }, {
-      'aria-label': 'video progress bar'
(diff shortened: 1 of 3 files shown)


OwenEdwards:
LGTM

gkatsev:
LGTM.

PR #3230, opened 2016-03-31
Attach handlers to ownerDocument, not document

andyearnshaw:
## Description
Currently, video.js has problems if the main script is loaded into a window but the player will be rendered into another window (e.g. an iframe).  Everywhere an event handler attaches to `document` assumes the document where the script was loaded, rather than where the player resides.  This is mostly apparent with the Slider component, that attaches _mousemove_ events on _mousedown_ and removes them on _mouseup_.

## Specific Changes proposed
This PR makes some headway towards context independence for the player.  It attaches events for the Slider and volume controls to the `ownerDocument` of the component's element instead of the global document object.

This allows for a use case we have with React components that may or may not render a video in an iframe, saving us from potentially reloading the script many times as the component updates.  A proper test suite for context independence might be required in the future if this is considered to be an important use case.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/volume-menu-button.js
+++ b/src/js/control-bar/volume-menu-button.js
@@ -7,7 +7,6 @@ import Popup from '../popup/popup.js';
 import PopupButton from '../popup/popup-button.js';
 import MuteToggle from './mute-toggle.js';
 import VolumeBar from './volume-control/volume-bar.js';
-import document from 'global/document';
 
 /**
  * Button for volume popup
@@ -131,7 +130,7 @@ class VolumeMenuButton extends PopupButton {
 
   handleMouseDown(event) {
     this.on(['mousemove', 'touchmove'], Fn.bind(this.volumeBar, this.volumeBar.handleMouseMove));
-    this.on(document, ['mouseup', 'touchend'], this.handleMouseUp);
+    this.on(this.el_.ownerDocument, ['mouseup', 'touchend'], this.handleMouseUp);
   }
 
   handleMouseUp(event) {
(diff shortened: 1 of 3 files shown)


gkatsev:
Seems like a good change. We'll need to make sure this is tested across browsers like IE8 and on mobile.

nickygerritsen:
It seems `ownerDocument` is supported "everywhere", even IE6: http://caniuse.com/#search=ownerdocument

So although we should probably test it doesn't break anything, it LGTM

gkatsev:
Verified in IE8, the only browser that could have potentially given us problems with this.

gkatsev:
thanks @andyearnshaw!

OwenEdwards:
Just noticed this PR - @gkatsev I'm a little concerned that the changes happen for the sliders but not for, e.g., `ClickableComponent`, which also attaches event handlers to the `document`. It seems like this could make some controls behave differently from others in the exact situation this is intended to address?

gkatsev:
@OwenEdwards yeah, those should probably change as well. Didn't realize that more things bind to the document. For normal users of videojs, they shouldn't be any difference since the element's owner document and `document` would be the same.

@andyearnshaw would you be up for going through and fixing the other references?

andyearnshaw:
@gkatsev sure, I'll try and set some time aside this week.

gkatsev:
Sounds good, thanks!

andyearnshaw:
Quick update, unfortunately we've had to switch back to JWPlayer because the MailOnline's VAST/VPAID plugin doesn't support non-linear content.  It's a real shame because JWPlayer's has similar issues to the ones I've contributed fixes for here.

The downside is that I haven't had much time to fix these other references, but it's on my list of things to do.

PR #3231, opened 2016-03-31
add missing tabindex to player controls for keyboard accessibility

marguinbc:
## Description
Tabbing to buttons implemented as divs requires tabindex for keyboard accessibility
Adds tabindex to three controls implemented as divs
Fixes tabindex and allows user to tab to controls on control bar

## Specific Changes proposed
One line change to add tabindex to attributes of controls

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -45,6 +45,7 @@ class Button extends ClickableComponent {
     attributes = assign({
       type: 'button', // Necessary since the default button type is "submit"
       'aria-live': 'polite' // let the screen reader user know that the text of the button may change
+      'tabindex' : 0 // add tabindex for keyboard accessibility to buttons
     }, attributes);
 
     let el = Component.prototype.createEl.call(this, tag, props, attributes);


src/js/button.js
```diff
@@ -45,6 +45,7 @@ class Button extends ClickableComponent {
     attributes = assign({
       type: 'button', // Necessary since the default button type is "submit"
       'aria-live': 'polite' // let the screen reader user know that the text of the button may change
+      'tabindex' : 0 // add tabindex for keyboard accessibility to buttons
```
gkatsev: If it's already a button element, we don't want to add a `tabindex`.
Though, I guess we do probably want to add a `tabindex` if it isn't a button.
Thoughts, @OwenEdwards?

marguinbc: For some reason, this code only adds tabindex for the three controls that the player complains are deprecated for using divs as buttons.  It doesn't add tabindex to actual button controls.

OwenEdwards: This specifically doesn't add `tabindex=0` to native `button` elements, where it isn't required. The `ClickableComponent` does add `tabindex=0` where non-button elements (typically `div`s) are used as buttons, otherwise they won't receive keyboard focus. I believe I've moved all the `extends Button`s over to `extends ClickableComponent`s where appropriate, so this all works out correctly.

If I've missed a control, or controls that aren't part of the video.js core extend a button and then override its element, then the test which produce the warning should also add a `tabindex=0`, for backward compatibility. Is this what you're referring to, @marguinbc, or are you expecting to see `tabindex` on native `button` elements still?

gkatsev:
We realized that the issue is actually in the [compatibility](https://github.com/videojs/videojs-4to5) script which turns button elements back into divs like in videojs 4. With the vjs5 change, this meant that those elements lost their tab index. Making the change there sounds like a better approach.

PR #3232, opened 2016-04-01
Is chrome on edge return true

benjipott:
## Description
Fix videojs.isChrome() on windows 10 edge return true

Mozilla/5.0 (Windows NT 10.0) AppleWebKit/537.36 (KHTML, like Gecko) **Chrome**/42.0.2311.135 Safari/537.36 Edge/12.10136

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -57,7 +57,8 @@ export const IS_OLD_ANDROID = IS_ANDROID && (/webkit/i).test(USER_AGENT) && ANDR
 export const IS_NATIVE_ANDROID = IS_ANDROID && ANDROID_VERSION < 5 && appleWebkitVersion < 537;
 
 export const IS_FIREFOX = (/Firefox/i).test(USER_AGENT);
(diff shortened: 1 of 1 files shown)


gkatsev:
LGTM.
Wondering if you were seeing any problems related to it or did you just notice that it was true?

benjipott:
Juste notice it, on windows 10 edge, if vjs / plugin test videojs.browser.IS_CHROME return true i think this const should return false

gkatsev:
yeah, we should definitely have it return the right value.

nickygerritsen:
@benjipott I was thinking: because Edge has kinda a strange UA string, isn't it better to put some `IS_EDGE` check at the top and for others add an `!IS_EDGE && ...`?

I can imagine we later add a Safari check and it will break again, because Edge also contains `Safari`.

So i.e.:

```javascript
export const IS_CHROME = !IS_EDGE && (/Chrome/i).test(USER_AGENT);
// etc.
```

nickygerritsen:
OK LGTM :)

misteroneill:
LGTM as well 👍

Issue #2386, opened 2015-07-20
Uncaught TypeError: videojs is not a function  error in 5.0, autoSetup


lomok:
I continue to get this message - I am using rc_2,  5.0.

It does not seem to affect anything -  people I have shown it to believe it is a timing issue.
```
video.js:9407 Uncaught TypeError: videojs is not a function
```
In chrome browser, clicking on the error in output window takes one to line 9407 --, which is :+1:       `var player = videojs(mediaEl);`

From the code:
```
// Check if any media elements exist
  if (mediaEls && mediaEls.length > 0) {

    for (var i = 0, e = mediaEls.length; i < e; i++) {
      var mediaEl = mediaEls[i];

      // Check if element exists, has getAttribute func.
      // IE seems to consider typeof el.getAttribute == 'object' instead of 'function' like expected, at least when loading the player immediately.
      if (mediaEl && mediaEl.getAttribute) {

        // Make sure this player hasn't already been set up.
        if (mediaEl.player === undefined) {
          var options = mediaEl.getAttribute('data-setup');

          // Check if data-setup attr exists.
          // We only auto-setup if they've added the data-setup attr.
          if (options !== null) {
            // Create new video.js instance.
line 9407 ----->>>>            var player = videojs(mediaEl);
          }
        }
```

heff:
This one should be fixed in the latest 5.0 RC. Let me know if you find differently.

clauso:
Hi,

I get this message with the current 5.0.0 in Firefox (https://github.com/videojs/video.js/releases/tag/v5.0.0).
Video plays and everything seems to work fine. However as we log client side js errors to our server through an API ( we want to notice if things are not working well ) this generates a whole lot of error messages.
Claus

grebett:
My personal workaround, in the video.js file, remove:

```javascript
// l. 11283
var videojs = undefined; // remove this line
```

and

```javascript
// l. 11344
var autoSetupTimeout = function autoSetupTimeout(wait, vjs) {
  videojs = vjs; // remove this line
  setTimeout(autoSetup, wait);
};
```
Since in videojs is already defined in the file, and not defined again after manually set it to undefined for ie8 fix. Not sure to understand all the issues of this bug, and don't have time too, but may be the beginning of a solution.

yairpe:
my workaround: change

          `if (options !== '{}' && options !== null) {					//  @@ if (options !== null) {
            // Create new video.js instance.
            var player = videojs(mediaEl);
           }`

[...]

PR #3233, opened 2016-04-01
do not reset videojs when autoSetupTimeout called on autoSetup retry

vtytar:
## Description
When autoSetupTimeout is called as a retry, it resets videojs variable to undefined, failing consequent auto-setup. See autoSetupTimeout in src/js/setup.js:

if (mediaEl && mediaEl.getAttribute) {
// ...
} else {
  autoSetupTimeout(1);
  break;
}

## Solution
Avoid videojs redefinition when autoSetupTimeout called without second argument (on autoSetup retry). It solves issue #2386.

### DIFF
--- a/src/js/setup.js
+++ b/src/js/setup.js
@@ -72,7 +72,8 @@ var autoSetup = function(){
 
 // Pause to let the DOM keep processing
 var autoSetupTimeout = function(wait, vjs){
-  videojs = vjs;
+  if (vjs)
+    videojs = vjs;
   setTimeout(autoSetup, wait);
 };
 


gkatsev:
Thanks!

src/js/setup.js
```diff
@@ -72,7 +72,8 @@ var autoSetup = function(){
 
 // Pause to let the DOM keep processing
 var autoSetupTimeout = function(wait, vjs){
-  videojs = vjs;
+  if (vjs)
```
gkatsev: our style guide requires to add curly braces around all blocks, but I can just add them during the merge.

misteroneill:
LGTM

PR #3237, opened 2016-04-04
Fixes #2156 make controls visible

### LINKED ISSUE #2156: Keyboard accessibility of control bar after it is hidden during playback

chrisauclair:
## Description
Fixes the a11y issue in #2156, where keyboard and screen reader users lose control focus when the control bar hides due to inactivity. I set the controls to `visibility: visible` and tested with NVDA and JAWS across all supported browsers.

For IE8, I opted to keep the controls hidden because `-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=0)";` won't make all of the children transparent. More work could be done on the controls for IE8, if someone can make a compelling case for it.

### DIFF
--- a/src/css/components/_control-bar.scss
+++ b/src/css/components/_control-bar.scss
@@ -10,15 +10,6 @@
   @include background-color-with-alpha($primary-background-color, $primary-background-transparency);
 }
 
-// IE8 is flakey with fonts, and you have to change the actual content to force
-// fonts to show/hide properly.
-// - "\9" IE8 hack didn't work for this
-// Found in XP IE8 from http://modern.ie. Does not show up in "IE8 mode" in IE9
-$ie8screen: "\\0screen";
-.vjs-user-inactive.vjs-playing .vjs-control-bar :before {
-  @media #{$ie8screen} { content: ""; }
-}
-
 // Video has started playing
 .vjs-has-started .vjs-control-bar {
   @include display-flex;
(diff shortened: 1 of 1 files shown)


src/css/components/_control-bar.scss
```diff
@@ -10,6 +10,15 @@
   @include background-color-with-alpha($primary-background-color, $primary-background-transparency);
 }
 
+// IE8 is flakey with fonts, and you have to change the actual content to force
```
gkatsev: why move this section?

src/css/components/_control-bar.scss
```diff
@@ -22,11 +31,17 @@
...
 .vjs-has-started.vjs-user-inactive.vjs-playing .vjs-control-bar {
-  visibility: hidden;
+  // Remain visible for screen reader and keyboard users
+  visibility: visible;
   opacity: 0;
 
   $trans: visibility 1.0s, opacity 1.0s;
   @include transition($trans);
+
+  // Make controls hidden in IE8 for now
```
gkatsev: sounds good to me

src/css/components/_control-bar.scss
gkatsev: oh, I see, it's for the variable. Maybe move just the variable but lease the section where it was?

gkatsev:
LGTM.
@OwenEdwards can you take a look? Thanks.

OwenEdwards:
LGTM :+1:

PR #3240, opened 2016-04-05
Add <style> to player tag instead of head

andyearnshaw:
## Description
Another portability fix (see #3230).  Currently, players moved to or created in a different window from the video.js script will not have the proper dimensions applied.

## Specific Changes proposed
The video dimensions stylesheet is added to the player element instead of `<head>`.  The `scoped` attribute is also set, though only supported by Firefox it also serves as a cosmetic validation for where the element is contained (scoped style elements are [permitted alongside flow content](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element:concept-element-contexts)).

By doing this, the element can be moved to another window or frame without losing these dimensions.  All browsers support `<style>` elements outside of `<head>`, though it is technically invalid without the `scoped` attribute.  Behaviour-wise, nothing has changed so current tests continue to pass and additional unit tests are unnecessary.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -288,9 +288,8 @@ class Player extends Component {
     // video element
     if (window.VIDEOJS_NO_DYNAMIC_STYLE !== true) {
       this.styleEl_ = stylesheet.createStyleElement('vjs-styles-dimensions');
-      let defaultsStyleEl = Dom.$('.vjs-styles-defaults');
-      let head = Dom.$('head');
-      head.insertBefore(this.styleEl_, defaultsStyleEl ? defaultsStyleEl.nextSibling : head.firstChild);
+      this.styleEl_.setAttribute('scoped', '');
+      this.el_.insertBefore(this.styleEl_, this.el_.firstChild);
     }
 
     // Pass in the width/height/aspectRatio options which will update the style el


gkatsev:
Thank you for the PR; unfortunately, we cannot accept it.

We used to have this in the player but we got rid of it. The reason why we cannot have it be inside the player element is that then it's really hard to override the player css.
So, if you wanted to size the player with just `.video-js` selector without the use of `!important`, you cannot do it because the player dimensions element has at least the same specificity but most definitely will come after your own css in the CSSOM.
Also, the scoped attribute apparently is probably not ever going to be standardized, unfortunately.

However, one thing you should take a look at -- which will be released as 5.9.0 later today -- is the [`VIDEOJS_NO_DYNAMIC_STYLE`](https://github.com/videojs/video.js/blob/master/docs/guides/skins.md#default-style-elements) property. This actually makes videojs not add any of these dynamic styles to the DOM at all and should help you make the player portable.

Thanks, hope to see more from you!

andyearnshaw:
Ah indeed, I can understand the specificity problem.  I think _scoped_ is probably going to be dropped from the spec too, as Firefox is the only browser to implement it and Blink have performance/optimisation concerns.  In the distant future this would probably be better handled using a shadow tree anyway.

Thanks for pointing me towards VIDEOJS_NO_DYNAMIC_STYLE :smile:.

gkatsev:
No problem, it just got added and released as part of 5.9.

PR #3247, opened 2016-04-11
added ie9-11 and edge

BrandonOCasey:
## Description
Added browser boolean constants for ie9, ie10, ie11, and EDGE.

## Specific Changes proposed
Add regex to detect browsers for ie9, ie10, ie11, and EDGE given the user agent.

## Requirements Checklist

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -58,7 +58,13 @@ export const IS_NATIVE_ANDROID = IS_ANDROID && ANDROID_VERSION < 5 && appleWebki
 
 export const IS_FIREFOX = (/Firefox/i).test(USER_AGENT);
 export const IS_CHROME = (/Chrome/i).test(USER_AGENT);
-export const IS_IE8 = (/MSIE\s8\.0/).test(USER_AGENT);
+export const IS_EDGE = (/Edge\/\d/i).test(USER_AGENT);
+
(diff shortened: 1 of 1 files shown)


src/js/utils/browser.js
```diff
@@ -58,7 +58,13 @@ export const IS_NATIVE_ANDROID = IS_ANDROID && ANDROID_VERSION < 5 && appleWebki
 
 export const IS_FIREFOX = (/Firefox/i).test(USER_AGENT);
 export const IS_CHROME = (/Chrome/i).test(USER_AGENT);
-export const IS_IE8 = (/MSIE\s8\.0/).test(USER_AGENT);
+export const IS_EDGE = (/Edge\/\d/i).test(USER_AGENT);
+
+export const IS_IE8 = (/MSIE\s8\.0/i).test(USER_AGENT);
+export const IS_IE9 = (/MSIE\s9\.0/i).test(USER_AGENT);
+export const IS_IE10 = (/MSIE\s10\.0/i).test(USER_AGENT);
+export const IS_IE11 = (/Trident\/7.0/i).test(USER_AGENT) && (/rv:11.0/).test(USER_AGENT);
```
dmlap: I'd like to only add these when we have a pressing need (so we don't get sloppy and use them more often). What is the use-case for these additional checks?

PR #3248, opened 2016-04-11
the first default text tracks will be show by default in emulated tracks

### LINKED ISSUE #1914: Subtitles default does not work

gkatsev:
## Description
As the title suggests, default emulated text tracks will now be shown by default. It will choose the first one from the list.

## Specific Changes proposed
* Native tracks will still be have the default attribute handled by the browser
* TextTrack objects now have a default property that represents whether they were created with it originally
* the first default emulated text track will be chosen to be shown by default
* Fixes #1914

## Requirements Checklist

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -76,6 +76,10 @@ class TextTrackDisplay extends Component {
           }
         }
 
+        // We want to show the first default track but captions and subtitles
+        // take precedence over descriptions.
+        // So, display the first default captions or subtitles track
+        // and otherwise the first default descriptions track.
         if (firstCaptions) {
           firstCaptions.mode = 'showing';
         } else if (firstDesc) {
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -62,11 +62,13 @@ class TextTrackDisplay extends Component {
       let modesToShow = {'captions': 1, 'subtitles': 1, 'descriptions': 1};
       let trackList = this.player_.textTracks();
 
-      for (let i = 0; i < trackList.length; i++) {
-        let track = trackList[i];
-        if (track.default && track.kind in modesToShow) {
-          track.mode = 'showing';
-          break;
+      if (trackList) {
+        for (let i = 0; i < trackList.length; i++) {
+          let track = trackList[i];
+          if (track.default && track.kind in modesToShow) {
+            track.mode = 'showing';
+            break;
+          }
         }
       }
     }));
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -58,6 +58,17 @@ class TextTrackDisplay extends Component {
         let track = tracks[i];
         this.player_.addRemoteTextTrack(track);
       }
+
+      let modesToShow = {'captions': 1, 'subtitles': 1, 'descriptions': 1};
+      let trackList = this.player_.textTracks();
+
+      for (let i = 0; i < trackList.length; i++) {
+        let track = trackList[i];
+        if (track.default && track.kind in modesToShow) {
+          track.mode = 'showing';
+          break;
+        }
+      }
     }));
   }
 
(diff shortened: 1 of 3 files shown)


dmlap:
This turns on captions/subtitles/descriptions if any of them are marked `default`? How does that line up with native browser behavior?

gkatsev:
@dmlap that's what chrome does. Safari does it's own thing where it remembers your last position. Really hard to test this. Looks like firefox only does it for english (so, probably if the srclang matches your browser's locale).

gkatsev:
Firefox also doesn't even have a captions button to toggle visibility of the current track on/off like chrome does.

gkatsev:
Also, I think most users *expect* it to just show the first default text track, regardless of the locale.

src/js/tracks/text-track-display.js
```diff
@@ -58,6 +58,19 @@ class TextTrackDisplay extends Component {
         let track = tracks[i];
         this.player_.addRemoteTextTrack(track);
       }
+
+      let modesToShow = {'captions': 1, 'subtitles': 1, 'descriptions': 1};
```
nickygerritsen: Is there a reason you use an object instead of an array?

src/js/tracks/text-track-display.js
```diff
@@ -58,6 +58,19 @@ class TextTrackDisplay extends Component {
...
+      let trackList = this.player_.textTracks();
+
+      if (trackList) {
+        for (let i = 0; i < trackList.length; i++) {
+          let track = trackList[i];
+          if (track.default && track.kind in modesToShow) {
+            track.mode = 'showing';
+            break;
```
nickygerritsen: Might be my misunderstanding of how it works natively, but is this also what happens normally? I.e. If I have both a caption and description having the default attribute, it now matters in which order I list them.

nickygerritsen:
Ah oops @dmlap asked a similar question already, my bad

src/js/tracks/text-track-display.js
gkatsev: Because `in` operator is `O(1)` vs `indexOf` is `O(n)`.

src/js/tracks/text-track-display.js
gkatsev: Actually, you're correct, in the descriptions PR, we decided with @OwenEdwards that descriptions should only default after captions/subtitles.

gkatsev:
@nickygerritsen heh, it's ok. It made me realize that I need to default descriptions *after* defaulting captions/subtitles.

gkatsev:
Here's an example: http://plnkr.co/edit/rncA4P?p=info

gkatsev:
I'll add a comment about defaulting to the first track unless there's both a caption/subtitles and a description, in which case captions/subtitles take priority.
I'll also see if I can move this block of code into the Tech constructor and add some tests.

gkatsev:
Ok, updated with tests and also a comment about what's happening.

nickygerritsen:
Nice, LGTM

surikat1978:
all made default and the first track does not pick up:
`<track kind="captions" src="http://website/uploads/subtitles/Focus.2015_asl/Focus.2015_Track01.vtt" srclang="en" label="Double" default >
  <track kind="captions" src="http://website/uploads/subtitles/Focus.2015_asl/Focus.2015_Track02.vtt" srclang="en" label="English" default >
  <track kind="captions" src="http://website/uploads/subtitles/Focus.2015_asl/Focus.2015_Track03.vtt" srclang="ru" label="Russian" default >`

surikat1978:
I remove this line : `"html5": { nativeTextTracks: false},`
and after the default 'captions in the Google Chrome works
But setting captions have disappeared

PR #3252, opened 2016-04-12
Use fonts 2.0 that don't require wrapping codepoints

gkatsev:
## Description
The title says it all

### DIFF
--- a/src/css/vjs.scss
+++ b/src/css/vjs.scss
@@ -1,3 +1,1 @@
-$icon-codepoints: true;
-
 @import "video-js";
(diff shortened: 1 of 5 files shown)


src/css/vjs.scss
```diff
@@ -1,3 +1 @@
-$icon-codepoints: true;
```
gkatsev: Removing this file entirely would technically be a breaking change, even removing `icon-codepoints` could be questionable but I don't think many people have been using it.

nickygerritsen: Agreed, don't really think this is a problem

nickygerritsen:
LGTM, makes the code a lot cleaner :)

misteroneill:
Very nice, LGTM. Agree that while it could be construed as a (very edge-casey) breaking change, it's not significant enough to merit a 6.0 👍

gkatsev:
I'll just revert the change if anyone complains.

PR #3253, opened 2016-04-12
Fix removeRemoteTextTracks not working with return value from addRemoteTextTracks

### LINKED ISSUE #2046: player.removeRemoteTextTrack(track) Failing
Attempting to remove a text track dynamically does not work in Firefox or Chrome.

reduced test case: http://partner.mediamobz.com/caption-test/

The buttons below the player trigger a removeRemoteTextTrack on the player for the current track and then dynamically add a new track.  The new track gets added, but the old track still exists in the menu and within the player.

This test case is running the latest CDN 4.12 version of videojs.

Should be working according to comments here: https://github.com/videojs/video.js/issues/1700

gkatsev:
## Description
Fix removeRemoteTextTracks not working with return value from addRemoteTextTracks

## Specific Changes proposed
Destructure and default the argument to `removeRemoteTextTrack` so that it accepts either the track itself or an object with a `track` property.
Also, add tests for this.
Fix #2046.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2579,7 +2579,9 @@ class Player extends Component {
    * @param {Object} track    Remote text track to remove
    * @method removeRemoteTextTrack
    */
-  removeRemoteTextTrack(track) {
+  // destructure the input into an object with a track argument, defaulting to arguments[0]
+  // default the whole argument to an empty object if nothing was passed in
+  removeRemoteTextTrack({track = arguments[0]} = {}) {
     this.tech_ && this.tech_['removeRemoteTextTrack'](track);
   }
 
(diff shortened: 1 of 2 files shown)


gkatsev:
We really need to switch off of jshint already :)

nickygerritsen:
Hmmm, are we doing the destructure thing more often? It feels ugly to me but that might just be me.

gkatsev:
It's in the language, and I made a comment about what's it's happening. With the destructuring, we can keep the body of the function nice and short and clear.

gkatsev:
If you think that even with the comment, it's too confusing, I can make the change.

nickygerritsen:
No it's fine, was just wondering.

PR #3254, opened 2016-04-13
revert 75116d4 adding chrome to travis

gkatsev:
This is because travis is failing to install the correct version of dependencies. We can re-add chrome back in at a later date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - 0.12
 before_install:
-  - export CHROME_BIN=/usr/bin/google-chrome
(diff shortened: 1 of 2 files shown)


nickygerritsen:
LGTM

kristerkari:
The install seems to fail because Chrome does not support the older Ubuntu version, it works if you tell Travis to use a newer version of Ubuntu:
- https://github.com/travis-ci/travis-ci/issues/5899#issuecomment-209677413
- https://github.com/kristerkari/travis-chrome/blob/master/.travis.yml

gkatsev:
@kristerkari sweet thanks. I'll make that change.

Issue #3202, opened 2016-03-23
How do we disable “captions settings” in  video.js player? Shows up in IE 11


lomok:
Hi-

I posted a variation of this earlier - same problem in firefox-
https://github.com/videojs/video.js/issues/2397
 and I find someone else had posted this on stack overflow-

Here it is:
"I am using video.js 5.7.1 on a single video player within a web page. The video element contains a "captions" track element. When the user clicks the CC button in IE 11, they see a menu containing the following:
	•	captions settings
	•	captions off
	•	english
I would like to completely remove the first menu item "captions settings" so that the menu only shows "captions off" and "english".
I'm sure it's something simple, but I can't seem to find how to disable/remove the "captions settings" menu item. What needs to happen to remove this option from the CC button on the player?"

http://stackoverflow.com/questions/35806897/how-do-we-disable-captions-settings-in-a-video-js-player

I am having these issues in ie 11- it was never answered by anyone.

Does anyone  how to stop this from occurring? I thought it may be an ie 11 html 5 issue, but other video players do not have this problem

Thanks!
![screenshot_2181](https://cloud.githubusercontent.com/assets/3836962/14003341/15000832-f10f-11e5-9730-45957818e13c.png)

gkatsev:
Unfortunately, we don't really have a good way of disabling it. Is there a reason why you want to hide it?

The "easiest" way to hide it is to set the `textTrackSettings` option to false and then add a bit of css to the page:
```js
var player = videojs(videoEl, {
  // make the text track settings dialog not initialize
  textTrackSettings: false
});
```
```css
/* hide the captions settings item from the captions menu */
.vjs-texttrack-settings {
  display: none;
}
```

gkatsev:
We probably should just not add the `captions settings` item if the component was removed from the player.

lomok:
I actually used this:
$(".vjs-texttrack-settings").css("display","none");
after making the player visible -it works great. Thanks so much.

gkatsev:
Great, thanks.

gkatsev:
Also, addressed by https://github.com/videojs/video.js/commit/18cdf08c0e784073d37d3dfc807dfb47eccdd344

nmaves:
Is this caption settings dialog only available in IE11?  If so why?

gkatsev:
[...]

Issue #2628, opened 2015-09-22
No method to dynamically change <tracks>


hparra:
This is probably related to #2369.

**I was looking for something similar to `player.src([{}])` but for tracks**. I have browsed issues at https://github.com/videojs/video.js/milestones/Text%20Tracks and read https://github.com/videojs/video.js/blob/stable/docs/guides/tracks.md but I am still confused by how related TextTrack API is supposed to work. Any clarification and known pitfalls would help me to take a stab at this.

I'm still on 4.x and would be implementing anything against that.

Related: Would a method to dynamically change all video child tags at once be useful? I could see why changing just the tracks mid-play would be a more difficult change.

hparra:
Anyone? @heff?

gkatsev:
The API for working with text tracks currently mimic the native text track APIs.
There's a `player.addTextTrack` which creates a programmatic text track and allows you to manually add cues to that track that you get.
Also, more useful to you, there's `player.addRemoteTextTrack`, this is a function that takes an object that looks pretty similar to the track element. It will then go and create the track for you. There's a corresponding `player.removeRemoteTextTrack` method to remove these. Note, you cannot remove tracks created by `player.addTextTrack`
You can also see what tracks are available by getting them in `player.textTracks()` or just the remote tracks with `player.remoteTextTracks()`.
A quick way to swap out all the tracks would be to grab each of them with `player.remoteTextTracks`, remove them with `player.removeRemoteTextTracks`, and then add new ones with `player.addRemoteTextTracks`. Maybe something like:
```js
var oldTracks = player.remoteTextTracks();
var i = oldTracks.length;
while (i--) {
  player.removeRemoteTextTrack(oldTracks[i]);
}
myNewTracks.forEach(function(track) {
  player.addRemoteTextTrack(track);
});
```

And yes, we really need to fix up and update our documentation relating to text tracks and how to use them.

gkatsev:
@hparra does my previous comment help? Sorry for the delayed responses.

hparra:
Thanks @gkatsev. Not sure yet, other fires... Were there (planned) changes in between 4.x and 5.x?

gkatsev:
There was a change between v4 and v5 around the main text track list object that videojs has but that shouldn't affect you.
Ok, let me know when you get a chance. I'll leave it open.

chemoish:
@gkatsev is there currently a way to detect the loading of remoteTextTracks?

[...]

Issue #1958, opened 2015-03-18
Out of Date Tracks Documentation


brimanning:
The [text tracks documentation](https://github.com/videojs/video.js/blob/stable/docs/api/vjs.TextTrack.md) is currently out-of-date.

It looks like the captions overhaul of [4.12](https://github.com/videojs/video.js/releases/tag/v4.12.0) is probably why.

Notable example issues:

- Methods missing (e.g. [`vjs.TextTracks.activate()` is no longer a method](https://github.com/videojs/video.js/blob/stable/docs/api/vjs.TextTrack.md#activate))
- Links to code are wrong for links to src/js/tracks.js (e.g. [`activeCues()`](https://github.com/videojs/video.js/blob/stable/docs/api/vjs.TextTrack.md#activecues) link in documentation returns a [404](https://github.com/videojs/video.js/blob/master/src/js/tracks.js#L270))

gkatsev:
Yes, you're correct. In 4.12, we reworked our text tracks completely. Unfortunately, documentation tends to fall behind :(

brimanning:
Of course. Makes sense to me. It just took some digging to realize that.

gkatsev:
Fixed by https://github.com/videojs/video.js/commit/18cdf08c0e784073d37d3dfc807dfb47eccdd344

PR #3256, opened 2016-04-13
Text tracks documentation updates

### LINKED ISSUE #1888: Add crossorigin documentation and explanation and update examples

gkatsev:
## Description
Update the documentation for Text Tracks.
Fixes #1888, #1958, #2628, and #3202.


gkatsev:
There is one functional change where it logs a warning if it thinks that the lack of `crossorigin` attribute could cause a problem with displaying text tracks, this is one of the requirements from #1888.

docs/guides/tracks.md
```diff
@@ -1,44 +1,50 @@
...
 but the text track served from another, you'll need to [enable CORS](http://enable-cors.org/) in order to do so.
+In addition to enabling CORS on the server serving the text tracks, you will need to add the [`crossorigin` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes) to the video element itself. This attribute has two values `anonymous` and `use-credentials`. Most users will want to use `anonymous` with cross-origin tracks.
+It can be added to the video element like so:
+```html
+<video class="video-js" crossorigin="anonymous">
+  <source src="http://vjs.zencdn.net/v/oceans.mp4" type="video/mp4">
+  <track src="http://example.com/oceans.vtt" kind="captions" srclang="en" label="English">
+</video>
+```
+One thing to be aware of is that in this case the video files themselves will *also* needs CORS headers applied to it. Since is because some browsers apply the crossorigin attribute to the video source itself and not just the tracks and is considered a [security concern by the spec](https://html.spec.whatwg.org/multipage/embedded-content.html#security-and-privacy-considerations).
```
OwenEdwards: "This" instead of "Since"

docs/guides/tracks.md
```diff
@@ -1,44 +1,50 @@
...
 - **Chapters**: Chapter titles that are used to create navigation within the video. Typically they're in the form of a list of chapters that the viewer can click on to go to a specific chapter.
-- **Descriptions** (not supported yet): Text descriptions of what's happening in the video for when the video portion isn't available, because the viewer is blind, not using a screen, or driving and about to crash because they're trying to enjoy a video while driving. Descriptions are read by a screen reader or turned into a separate audio track.
-- **Metadata** (not supported yet): Tracks that have data meant for javascript to parse and do something with. These aren't shown to the user.
+- **Descriptions**: Text descriptions of what's happening in the video for when the video portion isn't available, because the viewer is blind, not using a screen, or driving and about to crash because they're trying to enjoy a video while driving. Descriptions are read by a screen reader or turned into a separate audio track. Captions and Subtitles take precedence of Descriptions when they are selected.
```
OwenEdwards: "take precedence over" instead of "take precedence of". And maybe "Captions or Subtitles" rather than "Captions and Subtitles) - I don't think there's anything that mentions/specifies what happens if both a Caption *and* a Subtitle track are enabled?

docs/guides/tracks.md
```diff
@@ -52,157 +58,114 @@ The default attribute can be used to have a track default to showing. Otherwise
...
+
+### `player.textTracks() -> TextTrackList`
+This is the main interface into the text tracks of the player.
+It return a TextTrackList which lists all the tracks on the player.
+
+### `player.remoteTextTracks() -> TextTrackList`
+This is a helper method to get a list of all the tracks that were created from `track` elements or that were added to the player by the `addRemoteTextTrack` method. All these tracks are removeable from the player, where-as not all tracks from `player.textTracks()` are necessarily removeable.
+
+### `player.remoteTextTrackEls() -> HTMLTrackElementList`
+Another helper method, this is a list of all the `track` elements associates with the player. Both emulated or otherwise.
```
OwenEdwards: "associated" instead of "associates"

docs/guides/tracks.md
```diff
@@ -52,157 +58,114 @@ The default attribute can be used to have a track default to showing. Otherwise
...
+
+### `player.addRemoteTextTrack(Object options) -> HTMLTrackElement`
+This function takes an options object that looks pretty similar to the track element and returns a HTMLTrackElement.
+This object has a `track` property on it which is the actual TextTrack object.
+This `TextTrack` object is equivalent to the one that can be returned from `player.addTextTrack` with the added bonus that it can be removed from the player.
+Internally, video.js will either add a `<track>` element for you, or emulate that depending on whether native text tracks are supported or not.
+The options available are:
+* `kind`
+* `label`
+* `language` (also `srclang`)
```
OwenEdwards: Is it worth noting that `language` technically refers to the language used in the `label`, and `srclang` refers to the language of the track file itself? Or is that too much detail?

OwenEdwards:
A few minor editorial suggestions, otherwise LGTM.

docs/guides/tracks.md
gkatsev: `language` is the value that `srclang` gets translated to when the track element gets made into a track object, so, they're synonyms.

gkatsev:
Good changes, thanks.

docs/guides/tracks.md
OwenEdwards: Oh, I was confusing `language` with `lang` (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang); that's the thing which is distinct from `srclang`.

docs/guides/tracks.md
gkatsev: It'll pick the first text track that's a caption or a subtitle. See https://github.com/videojs/video.js/pull/3248

OwenEdwards: @gkatsev is that noted in this documentation somewhere? As we've discussed previously, in the case where both Captions and Subtitles are present for a video, the existing UI might give the impression that you can enable both independently (unlike the way I have the AD control disable itself if another track is enabled).

gkatsev: Ah, I see. No, that isn't mentioned anywhere. I'll add a section about default tracks. Native tracks could theoretically allow multiple captions to be selected together, though, I think our menu system disabled others.

OwenEdwards: There are three separate issues here:

1. Using the `default` attribute on a track, and how video.js handles the (error) case where there are multiple default tracks
2. What video.js does for [displaying tracks if more than one track is set to `showing`](https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-display.js#L117-140) (by any mechanism)
3. How the UI behaves (if you select one `kind` of track, it [disables any other tracks of the same kind](https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/text-track-menu-item.js#L81-93), but the display part doesn't support display of multiple `kind`s at once, as noted in 2). 

These three are obviously related, but in my mind are distinct. And now I've basically written the note myself! :-)

gkatsev: Haha, I'll add that in :)

OwenEdwards: 👍

OwenEdwards:
LGTM (in case my previous one didn't count!)

gkatsev:
Heh, thanks. I still have to add the other section about caption/subs+descriptions precedence with regards to defaults and menu selection and programmatic interfaces.

docs/guides/tracks.md
gkatsev: Added a section for this.

gkatsev:
@OwenEdwards the new section is in this commit: https://github.com/videojs/video.js/pull/3256/commits/33a7bca01a56b8c91277507165809671361354e4

OwenEdwards:
LGTM

PR #3260, opened 2016-04-15
Dispatch TextTrackCue enter and exit events

marc-ferrer:
## Description
To add missing `TextTrackCue` events (enter, exit) as described in the [standard](https://html.spec.whatwg.org/multipage/embedded-content.html#cue-events).
They are actually working in Chrome but not in Firefox, since video.js is not using native TextTracks in Firefox.

## Specific Changes proposed
Adds the the code to check for the cues that entered or exit since the last
check, to the `activeCues` getter which is run on every `timeUpdate` event.

## Requirements Checklist
- [x] Feature implemented
  - [ ] Docs/guides updated


gkatsev:
Thanks for the PR, I should be able to review it today. Also, the reason why it's working in chrome and not firefox is that in chrome we use the native text tracks but in firefox we force our emulated work. I have the beginning of a PR to make firefox use native text tracks as well (https://github.com/videojs/video.js/pull/3261) but it isn't quite ready yet.

gkatsev:
The emulated text track object currently implies that we're using vttjs's VTTCue object, though, it does support anything object, really.
I feel like that's necessary is to provide a mechanism to those VTTCues to be an EventTarget and thus be able to trigger events. Maybe even have a wrapper for vttjs's VTTCues that we use instead of using them directly.

Once we have those, we can always just do `cue.dispatchEvent({
  type: 'enter'
})` and not have to check to see that it's an `instanceof EventTarget`.

marc-ferrer:
That's what we have done in our project to support IE 11 (Chrome and Firefox's VTTCues already extend from `EventTarget`) since it does not have native VTTCue support either. We have a little shim that adds `dispatchEvent`, `addEventListener` and `removeEventListener` methods to VTTCue.

Unfortunately IE does not fulfill the `target` property in the `Event` object so as we need access to the cue that has triggered the `enter` event, we have ended adding `customEvent.cue = cue` (target is a ready only property).

Your solution of sending a simple object instead of an `Event` would solve our problems with the `target` property but we are not sure that it is the way to go when all browsers have support for `Events`.

Would it be fine to just add the EventTarget functionallity to VTTCue using a shim code? or do you prefer to have a customVTTCue that extends vttjs's VTTCue and implements `EventTarget`?

gkatsev:
Making a wrapper or another class that uses vttjs's VTTCue would mean that you could always call `dispatchEvent` without needing to check that is inherits from `window.EventTarget`. Making a wrapper for it that inherits from our own EventTarget is probably easiest.
As for using the `Event` object, it's probably fine except this still needs to work in IE8 and I'm not sure that works there.

gkatsev:
Hey, any updates on this? Thanks.

marc-ferrer:
Oops, I totally forgot about this... Past months have been really busy for us and apparently this issue ended up lost in a todo list. I might be able to retake this the next week.

marc-ferrer:
I'm reviewing this issue and I'm Having more troubles than I expected to find a proper solution.

How can I create a wrapper class that extends from `EventTarget` and `VTTCue` at the same time?

I could simply wrap the `VTTCue` inside a new CustomCue class that extends from EventTarget,

```
class CustomCue extends EventTarget {
  constructor(cue) {
    this.cue = cue
  }
}
```
Does not seems a correct solution as videojs would fail accessing `VTTCues` information.

Even thoug I could find a proper solution for extending both `VTTCue` and EventTarget at the same time,
cues are not created in this project, are they?

videojs-contrib-hls, and vttjs (through WebVTT parser) create cues and would need to use
the new CustomCues in order to be able to call cue.dispatchEvent.

gkatsev:
Going to close this PR because it isn't being actively worked on and it's out of date.
We'd still love to get this functionality in. Maybe we need think up of what needs to happen and create a plan for it. I'd be happy to talk about it some more either here or maybe in [slack](http://slack.videojs.com).

Issue #1862, opened 2015-02-13
Captions: Native Captions disabled in Firefox


gkatsev:
Investigate and fix why it's currently broken, so, we can use native captions in firefox.

chemoish:
Captions don't work for me entirely.

Firefox `40.0.3`

```bash
videojs@4.12.15
vtt.js@0.12.1
```

A request is made to obtain `node_modules/vtt.js/dist/vtt.js`, but vtt.js doesn't have a `dist` directory.

![screen shot 2015-10-14 at 5 43 07 pm](https://cloud.githubusercontent.com/assets/549319/10501903/0fbc397e-729b-11e5-8d17-8f3610a9d0e8.png)

`node_modules/vtt.js/lib/vtt.js`

However, fixing path errors don't have an effect on captions.

---

Maybe similar…

https://github.com/videojs/video.js/issues/2397

chemoish:
vtt.js issue seems to be addressed https://github.com/videojs/video.js/pull/2426

Dador:
Probably the reason why native captions not working with Firefox & video.js:
> Annoyingly, this doesn’t work for Firefox 35.0.1 which currently has all text tracks set to “disabled”, and the only way to override this value is to access the text track via the video itself.
More: http://www.iandevlin.com/blog/2015/02/javascript/dynamically-adding-text-tracks-to-html5-video

There is workaround on the link. Please take a look.

gkatsev:
Looks like we can re-enable native captions in current versions of firefox. They're working for me now.

gkatsev:
spoke too soon, I need to make sure that it's actually trying to use native captions and not just emulated captions.

gkatsev:
I figured out why tracks on firefox are broken. This also might have to do with why #2062 doesn't display native tracks.
When videojs moves the video element in the DOM, firefox had already started loaded the text tracks but then when the element is moved the already partially loaded tracks are killed by firefox and then it doesn't re-load it. So, we can try and do what we do on iOS in firefox or at least we'd want to remove the track elements and re-add them after moving the element in firefox.

gkatsev:
Setting `movingElementInDOM` to false on firefox fixes this issue!

gkatsev:
Well, maybe not fixes it but it seems to be helping out. At least I know what needs to happen here.

gkatsev:
[...]

Issue #1898, opened 2015-02-25
Tracks test fails in Firefox


heff:
[This test](https://github.com/videojs/video.js/blob/550d44332b6ec7df93d8d509c553cdcf348e9c3f/test/unit/tracks/tracks.js#L55) fails in Firefox because we don't use native captions there even though they're supported. The test should be combined with [this firefox test](https://github.com/videojs/video.js/blob/550d44332b6ec7df93d8d509c553cdcf348e9c3f/test/unit/tracks/tracks.js#L273).

gkatsev:
Closing because we're not going to be enabling native tracks in firefox because of issues described in #3261 and #1862.

PR #3261, opened 2016-04-15
Make firefox use native text tracks.

gkatsev:
## Description
Make firefox use native text tracks.
Fixes https://github.com/videojs/video.js/issues/1862.

## Specific Changes proposed
Pass player options to the tech.
Make firefox not allowed to move the element in the DOM and also clear
out the tracks from the element. In the end of the html5 constructor,
re-add the tracks from the player options.
This still causes firefox to print out a "Network Error" because it
aborts the original loading of the tracks.
screen shot 2016-04-15 at 14 05 54

It's possible that some work needs to happen so that we dont keep adding more tracks each time we switch from and to the html5 tech but this is the initial work.

Also, the control bar overlays the captions, though, the native control bar *also* overlays the captions.
screen shot 2016-04-15 at 14 12 00
screen shot 2016-04-15 at 14 14 03

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -551,6 +551,7 @@ class Player extends Component {
 
     // Grab tech-specific options from player options and add source and parent element to use.
     var techOptions = assign({
+      'playerOptions': this.options_,
       'nativeControlsForTouch': this.options_.nativeControlsForTouch,
       'source': source,
       'playerId': this.id(),
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -77,7 +77,7 @@ class Html5 extends Tech {
       this.proxyNativeTextTracks_();
 
       if (browser.IS_FIREFOX) {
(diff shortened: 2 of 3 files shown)


gkatsev:
And this could probably fix https://github.com/videojs/video.js/issues/1898 as well once it's mostly ready.

gkatsev:
Yep, #1898 is fixed by the latest commit (053e19c)

gkatsev:
Closing. I was trying to get tech switching working but we were losing the text tracks when coming back to the html5 tech because firefox wasn't firing the `addtrack` event on the tracks again. And so this would require another patch and who knows how much other stuff we could require to get this working correctly.
I'm going to open a bug with Mozilla about this but otherwise leave emulated tracks on with native captions disabled because of browser bug.

PR #3266, opened 2016-04-19
Fix to handleKeyPress in ClickableComponent

marguinbc:
## Description
Patch ClickableComponent but adding check to see if component is a DIV and not a button to determine if enter and space should be handled.

## Specific Changes proposed
Added condition to check if element is a DIV

## Requirements Checklist
- [x ] Bug fixed
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -87,9 +87,9 @@ class Button extends ClickableComponent {
    * @method handleKeyPress
    */
   handleKeyPress(event) {
+    // If the element is a div acting like a button, use super's handleKeyPress but 
     // Ignore Space (32) or Enter (13) key operation, which is handled by the browser for a button.
-    if (event.which === 32 || event.which === 13) {
-    } else {
+    if ((this.el_.tagName.toUpperCase() === 'DIV') && ((event.which !== 32) || (event.which !== 13))) {
       super.handleKeyPress(event); // Pass keypress handling up for unsupported keys
     }
   }


src/js/button.js
```diff
@@ -87,9 +87,9 @@ class Button extends ClickableComponent {
    * @method handleKeyPress
    */
   handleKeyPress(event) {
+    // If the element is a div acting like a button, use super's handleKeyPress but 
     // Ignore Space (32) or Enter (13) key operation, which is handled by the browser for a button.
-    if (event.which === 32 || event.which === 13) {
-    } else {
+    if ((this.el_.tagName.toUpperCase() === 'DIV') && ((event.which !== 32) || (event.which !== 13))) {
```
OwenEdwards: This needs to work if the element is a <span>, or anything else that isn't a <button>; why not use the same logic for the test as in line 40?

Also, `((event.which !== 32) || (event.which !== 13))` will always evaluate to true, even if which is 32 or 13!!

OwenEdwards:
The title and description of this PR are confusing because the "bug" is not in `ClickableComponent`, it's in `Button`. The whole point of splitting a Clickable Component from a Button was to fix a bug with Button handling, and the necessary changes were made elsewhere in video.js; this "bug" only shows up in the backward compatibility of the `Button`.

If we knew more about the other items which are extending `Button`, we could decide if we should actually honor their `<div>` passed as the `tag` parameter, or just ignore it and use a `<button>` element instead - that was a major goal in moving to V5. What do you think, @gkatsev & @heff?

gkatsev:
I think this may belong elsewhere, going to talk with @marguinbc offline.

src/js/button.js
marguinbc: Good point about div vs span and using the test on line 40.  The conditional logic was a brain fart, wanted to reverse the logic from 'do nothing if' and everything 'else' to 'do everything if' without an else.

PR #3267, opened 2016-04-19
Use background-color to cover video still

### LINKED ISSUE #1869: If cover art becomes scaled the borders around it are transparent meaning video can be seen below
Just found a minor issue that is when you have a cover art image that doesn't fit the player and it is scaled down, the player element is visible outside the scaled image.

This tweak sets the background-color property which means the player will always be completely covered if the poster does not fit exactly, or if the poster image cannot be loaded for some reason/whilst it's loading.

Here is what happens currently:
image

And here is what it looks like with the background-color property applied:
image

I have submitted a pull request at: https://github.com/videojs/video.js/pull/1868

forbesjo:
## Description
It looks like #2138 was lost in the 5.0 update. Fixes #1869 again.

## Specific Changes proposed
Use background-color to cover video still.

## Requirements Checklist

### DIFF
--- a/src/css/components/_poster.scss
+++ b/src/css/components/_poster.scss
@@ -4,6 +4,7 @@
   background-repeat: no-repeat;
   background-position: 50% 50%;
   background-size: contain;
+  background-color: #000000;
   cursor: pointer;
   margin: 0;
   padding: 0;


gkatsev:
LGTM, thanks.

PR #3269, opened 2016-04-19
Fix text track tests for older IEs

gkatsev:
## Description
Older IEs that don't recognize the track element don't read properties from it very well. So, in our tests we want to use `setAttribute` to set the track attributes.
The other test was changes because I originally thought it was necessary but we might as well have it because it's more readable.


misteroneill:
LGTM

Issue #2746, opened 2015-10-27
Accessibility of the Captions Settings dialog


OwenEdwards:
The Captions Settings dialog has some accessibility issues:
- [x] Field labels and fields are not explicitly associated
- [ ] Keyboard focus does not move into the dialog when it is opened
- [ ] Keyboard focus is not trapped inside the dialog while it is open
- [ ] Keyboard focus does not return to the control which opened the dialog when it is closed
- [ ] The extent (top and bottom) of the dialog is not indicated to screen readers
- [ ] The dialog cannot be closed with the Esc key
- [x] The meaning of '---' in the select fields is not clear
- [x] The control to close the dialog is labeled "Done" rather than "Close"
- [ ] The purpose of the "Defaults" button may not be obvious, and its effect may not be apparent to screen reader users

Some of these are very simple fixes, and some are lower priority than others, but all are fixable :-)

OwenEdwards:
Much of this can be adopted from @gdkraus' excellent example https://github.com/gdkraus/accessible-modal-dialog.

OwenEdwards:
@gkatsev, to clarify your points on the #2668 Modal Dialog thread,

1. I understand that, while the Captions Setting dialog is displayed then per FCC the user needs to see the effect of the settings I don't think having the Captions Setting dialog as a modal dialog excludes being able to see the current captions, unless it is implemented as a lightbox. The modal dialog doesn't have to be as big as the player window, does it?

2. Some players use an example of caption text in the settings dialog itself, so the existing captions *could* be obscured.

3. I guess I need to think about what I mean by modal dialog here; I get the point on #2668 that you want users to still be able to interact with the rest of the page, so the settings dialog shouldn't be modal in that sense, but I'm thinking of it more from a keyboard focus perspective, in terms of making sure that keyboard-only users can't accidentally move focus out of the dialog into player controls and then not be clear how to get focus back into the dialog.

Keyboard focus is something we need to think about broadly for accessibility, so let's give 3 some more thought. Can you clarify what you meant wrt 1 (and whether 2 might be an option? One of the issues I see with the Captions Settings dialog is sizing it for different video window sizes, which might make 2 more appealing?).

gkatsev:
1. Yep, the modal dialog could be sized down.
2. It's definitely doable to have a preview section just would require some more refactoring of text tracks and rendering portions.
[...]

PR #3281, opened 2016-04-26
Fix Captions Settings Dialog labeling for accessibility

OwenEdwards:
## Description

Based on #3054, this PR fixes screen reader labeling of the Captions Settings Dialog. This fixes point 1 of #2746.

## Specific Changes proposed

Add fieldsets & legends, and off-screen text for labels. Also, remove the '---' values.

## Requirements Checklist
- [x] Bug fixed

### DIFF
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -254,9 +254,8 @@ function captionOptionsMenuTemplate(uniqueId, dialogLabelId, dialogDescriptionId
 
   let template = `
     <div role="document">
-      <h1 id="${dialogLabelId}" class="vjs-control-text">Captions Settings Dialog</h1>
+      <div role="heading" aria-level="1" id="${dialogLabelId}" class="vjs-control-text">Captions Settings Dialog</div>
       <div id="${dialogDescriptionId}" class="vjs-control-text">Beginning of dialog window. Escape will cancel and close the window.</div>
-      <div class=""></div>
       <div class="vjs-tracksettings">
         <div class="vjs-tracksettings-colors">
           <fieldset class="vjs-fg-color vjs-tracksetting">
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -5,7 +5,6 @@ import Component from '../component';
 import * as Events from '../utils/events.js';
 import * as Fn from '../utils/fn.js';
 import log from '../utils/log.js';
-import * as Guid from '../utils/guid.js';
 import safeParseTuple from 'safe-json-parse/tuple';
 import window from 'global/window';
 
@@ -68,9 +67,18 @@ class TextTrackSettings extends Component {
    * @method createEl
    */
   createEl() {
+    let uniqueId = this.id_;
+    let dialogLabelId = 'TTsettingsDialogLabel-' + uniqueId;
+    let dialogDescriptionId = 'TTsettingsDialogDescription-' + uniqueId;
+
     return super.createEl('div', {
       className: 'vjs-caption-settings vjs-modal-overlay',
-      innerHTML: captionOptionsMenuTemplate()
+      innerHTML: captionOptionsMenuTemplate(uniqueId, dialogLabelId, dialogDescriptionId),
+      tabIndex: -1
+    }, {
+      role: 'dialog',
+      'aria-labelledby': dialogLabelId,
+      'aria-describedby': dialogDescriptionId
     });
   }
 
@@ -242,10 +250,13 @@ function setSelectedOption(target, value) {
   target.selectedIndex = i;
 }
 
-function captionOptionsMenuTemplate() {
-  let uniqueId = Guid.newGUID();
+function captionOptionsMenuTemplate(uniqueId, dialogLabelId, dialogDescriptionId) {
 
   let template = `
+    <div role="document">
+      <h1 id="${dialogLabelId}" class="vjs-control-text">Captions Settings Dialog</h1>
+      <div id="${dialogDescriptionId}" class="vjs-control-text">Beginning of dialog window. Escape will cancel and close the window.</div>
+      <div class=""></div>
       <div class="vjs-tracksettings">
         <div class="vjs-tracksettings-colors">
           <fieldset class="vjs-fg-color vjs-tracksetting">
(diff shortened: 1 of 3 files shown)


src/js/tracks/text-track-settings.js
```diff
@@ -241,14 +250,20 @@ function setSelectedOption(target, value) {
...
-        <div class="vjs-fg-color vjs-tracksetting">
-            <label class="vjs-label">Foreground</label>
-            <select>
-              <option value="">---</option>
-              <option value="#FFF">White</option>
+function captionOptionsMenuTemplate(uniqueId, dialogLabelId, dialogDescriptionId) {
+
+  let template = `
+    <div role="document">
+      <h1 id="${dialogLabelId}" class="vjs-control-text">Captions Settings Dialog</h1>
```
gkatsev: We probably should use a div here instead of h1, I had to change the elements used in videojs-dock because websites often style `h1` elements and it broke the title and descriptions of the plugin

src/js/tracks/text-track-settings.js
```diff
@@ -241,14 +250,20 @@ function setSelectedOption(target, value) {
...
+      <div id="${dialogDescriptionId}" class="vjs-control-text">Beginning of dialog window. Escape will cancel and close the window.</div>
+      <div class=""></div>
```
gkatsev: Looks superfluous.

test/unit/tracks/text-track-settings.test.js
```diff
@@ -9,6 +9,16 @@ const tracks = [{
   label: 'test'
 }];
 
+const defaultSettings = {
```
gkatsev: I'm not a fan of having these default values, I would prefer if vttjs just did whatever it wanted by default unless a user selects a value.
Maybe the `---` values should just change to be called `default`? or something?

src/js/tracks/text-track-settings.js
OwenEdwards: Even though `class="vjs-control-text"` means this one is visually hidden, and just there for Screen Readers?

gkatsev: Ah, right, I guess that's fine then.

test/unit/tracks/text-track-settings.test.js
OwenEdwards: I think you've got the logic of this backwards - basic Text Tracks do not specify a foreground or background color, and nor does vtt.js when it parses them, so there needs to be a "I want text displayed this way" setting, so "default" doesn't make sense.

Some cues in some tracks do specify a color or other styling, in which case the setting option ought to be "leave it as specified in the track" for those cues which have a specification in the track, or "force it to the styling which I've specified, overriding what's in the cue/track".

So "---" or "default" doesn't make sense, although there ought to be an extra checkbox (or slide switch) for each setting saying "override track's own setting".

gkatsev: Yeah, I meant with how the workflow from this dialog is. The selection boxes turn into an object which we then use down the line to see if we need to apply extra styles to the elements that vttjs produced.
By default, no such extra processing should be done; also, if the persistence option is set, we shouldn't record these default values.
Checkboxes that toggle each override could be ok, since that's sort of what the `---` option is supposed to represent.

gkatsev:
Hey, @OwenEdwards any updates on this one? Thoughts on my latest comment?

test/unit/tracks/text-track-settings.test.js
OwenEdwards: @gkatsev I needed to do some more research on this; I was wrong about "basic Text Tracks do not specify a foreground or background color, and nor does vtt.js when it parses them". 

The idea of `---` or `default` in the Captions Settings Dialog is somewhat bogus, since if a Text Track or cue doesn't specify its own styling, then the `default` is [buried in vtt.js](https://github.com/mozilla/vtt.js/blob/master/lib/vtt.js#L589-591), and is entirely arbitrary; neither the WebVTT standard nor anything from the FCC specifies a "default" setting for caption display if none is specified by the captions track. In my mind, the vtt.js default is equivalent to the [`Need Text` in video.js' components](https://github.com/videojs/video.js/blob/master/src/js/clickable-component.js#L95) - it's there in case nothing else is specified.

Which takes me back to my point that there shouldn't be a default in drop downs the video.js Captions Settings Dialog; there should be a default *set of settings* for color, opacity/transparency and font/outline/etc. which can match the existing behavior of vtt.js for consistency, and a new setting which should be an option of whether this *overrides* any color specified in the Text Track or not.

Whether the settings are saved by video.js and then restored (the "persistence option") is a separate feature from all of this.

gkatsev: I think we are in agreement.
The dropdowns shouldn't have any defaults or "empty values" and each setting should have an option of whether it will be used.

OwenEdwards: We're *almost* in agreement ;-) 

[x] The dropdowns shouldn't have any defaults or "empty values" 
[ ] See YouTube and iOS - the "checkbox" (whether for each value or for the settings as a whole) defines whether _styling in the WebVTT overrides the values set by the user_. That's not quite the same as whether or not to change what vtt.js does - the vtt.js defaults would disappear, and be replaced by the settings in video.js's Settings Dialog, although those setting values may or may not be changed by settings in the WebVTT file.

I can see the value in being able to completely turn off the Captions Settings functionality and not changing what vtt.js does (as was discussed elsewhere), and that could be implemented as a separate PR with a video.js parameter to control it.

Implementing this behavior of Captions Settings control won't be easy with the existing vtt.js though, and it's interface to `Services` and `WebVTTSet` object. We would need to fork vtt.js, and hope we can push a better user control API upstream to them.

gkatsev: We don't actually need to worry about the `Services` and `WebVTTSet` in vttjs because that's only used in native firefox. However, I wish that it was possible to get a build of vttjs without those. Also, we're still using a fork to enable our usage of vttjs the way we want.
I think I need to load up this PR and think about it some more.

OwenEdwards: We could suggest that vtt.js provides an API which allows user settings, including whether to override styling in the WebVTT or not, and that `Services` and `WebVTTSet` are one thing which drive that flexible API, but if `Services` does not exist, it's still possible to drive the user settings API. That's what I was thinking we should create in our fork, and then submit a PR to the Mozilla repo.

Because inside vtt.js is by far the easiest place to detect whether the WebVTT has styling or not ;-)

Issue #1951, opened 2015-03-16
EventEmitter.one needs to do the same dance as EventEmitter.on


gkatsev:
https://github.com/videojs/video.js/blob/master/src/js/event-emitter.js#L7-L14 and https://github.com/videojs/video.js/blob/master/src/js/event-emitter.js#L22-L24

gkatsev:
Fixed by #3285.

PR #3285, opened 2016-04-26
setSrc clears currentSource_ after loadstart

incompl:
## Description
currentSrc will now report correct source after the video source was changed by a third party.

This PR replaces #2957 

## Specific Changes proposed
* setSrc now clears currentSource_ after loadstart
*  EventTarget.prototype.one removes the addEventListener alias before calling Events.on so we don't get into an infinite type loop

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -11,7 +11,7 @@ EventTarget.prototype.on = function(type, fn) {
   // Remove the addEventListener alias before calling Events.on
   // so we don't get into an infinite type loop
   let ael = this.addEventListener;
-  this.addEventListener = Function.prototype;
+  this.addEventListener = () => {};
   Events.on(this, type, fn);
   this.addEventListener = ael;
 };
@@ -26,7 +26,7 @@ EventTarget.prototype.one = function(type, fn) {
   // Remove the addEventListener alias before calling Events.on
   // so we don't get into an infinite type loop
   let ael = this.addEventListener;
-  this.addEventListener = Function.prototype;
+  this.addEventListener = () => {};
   Events.one(this, type, fn);
   this.addEventListener = ael;
 };
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -23,7 +23,12 @@ EventTarget.prototype.off = function(type, fn) {
 EventTarget.prototype.removeEventListener = EventTarget.prototype.off;
 
 EventTarget.prototype.one = function(type, fn) {
+  // Remove the addEventListener alias before calling Events.on
+  // so we don't get into an infinite type loop
+  let ael = this.addEventListener;
+  this.addEventListener = Function.prototype;
   Events.one(this, type, fn);
+  this.addEventListener = ael;
 };
 
 EventTarget.prototype.trigger = function(event) {
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -99,7 +99,7 @@ class Html5 extends Tech {
 
     if (this.featuresNativeTextTracks) {
       if (crossoriginTracks) {
-        log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used. 
+        log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used.
             This may prevent text tracks from loading.`);
       }
 
@@ -549,7 +549,17 @@ class Html5 extends Tech {
    * @method setSrc
    */
   setSrc(src) {
+    let loadstartlistener = Html5.prototype.loadStartListener_;
+
+    this.off(this.el_, 'loadstart', loadstartlistener);
+    this.one(this.el_, 'loadstart', () => this.one(this.el_, 'loadstart', loadstartlistener));
+    this.disposeSourceHandler();
+
     this.el_.src = src;
+  }
+
+  loadStartListener_() {
+    this.currentSource_ = null;
   }
 
   /**
@@ -1147,7 +1157,6 @@ Html5.prototype['featuresNativeVideoTracks'] = Html5.supportsNativeVideoTracks()
  * @type {Boolean}
  */
 Html5.prototype['featuresNativeAudioTracks'] = Html5.supportsNativeAudioTracks();
-
 
 // HTML5 Feature detection and Device Fixes --------------------------------- //
 let canPlayType;
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -549,17 +549,7 @@ class Html5 extends Tech {
    * @method setSrc
    */
   setSrc(src) {
-    let loadstartlistener = Html5.prototype.loadStartListener_;
-
-    this.off(this.el_, 'loadstart', loadstartlistener);
-    this.one(this.el_, 'loadstart', () => this.one(this.el_, 'loadstart', loadstartlistener));
-
     this.el_.src = src;
-  }
-
-  loadStartListener_() {
-    this.currentSource_ = null;
-    this.disposeSourceHandler();
   }
 
   /**
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -99,7 +99,7 @@ class Html5 extends Tech {
 
     if (this.featuresNativeTextTracks) {
       if (crossoriginTracks) {
-        log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used.
+        log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used. 
             This may prevent text tracks from loading.`);
       }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -553,13 +553,13 @@ class Html5 extends Tech {
 
     this.off(this.el_, 'loadstart', loadstartlistener);
     this.one(this.el_, 'loadstart', () => this.one(this.el_, 'loadstart', loadstartlistener));
-    this.disposeSourceHandler();
 
     this.el_.src = src;
   }
 
   loadStartListener_() {
     this.currentSource_ = null;
+    this.disposeSourceHandler();
   }
 
   /**
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -842,6 +842,7 @@ Tech.withSourceHandlers = function(_Tech){
   _Tech.prototype.successiveLoadStartListener_ = function() {
     this.currentSource_ = null;
     this.disposeSourceHandler();
+    this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
   };
 
   /*
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -819,11 +819,11 @@ Tech.withSourceHandlers = function(_Tech){
 
       this.currentSource_ = source;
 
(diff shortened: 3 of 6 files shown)


gkatsev:
Fixes #1951. This still needs disposing the source handler, right?

incompl:
@gkatsev It has `this.disposeSourceHandler();` in `setSrc`

incompl:
Moved `disposeSourceHandler` to the right place

src/js/event-target.js
```diff
@@ -23,7 +23,12 @@ EventTarget.prototype.off = function(type, fn) {
 EventTarget.prototype.removeEventListener = EventTarget.prototype.off;
 
 EventTarget.prototype.one = function(type, fn) {
+  // Remove the addEventListener alias before calling Events.on
+  // so we don't get into an infinite type loop
+  let ael = this.addEventListener;
+  this.addEventListener = Function.prototype;
```
misteroneill: We have encountered issues with using `Function.prototype` as a no-op due to the way video.js tracks functions by a GUID property. I don't think it would be an issue here, but I think it's safer to never use it. Unfortunately. 😞

src/js/tech/html5.js
```diff
@@ -99,7 +99,7 @@ class Html5 extends Tech {
 
     if (this.featuresNativeTextTracks) {
       if (crossoriginTracks) {
-        log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used. 
+        log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used.
```
misteroneill: `tsml` will strip spaces between lines so this will end up with "isn't used.This may", which isn't the end of the world, but we could use my alternative package `tsmlj` which will join each line with a space.

test/unit/tech/html5.test.js
```diff
@@ -442,3 +445,28 @@ test('Html5#reset calls Html5.resetMediaElement when called', function() {
...
 });
+
+test('Html5#setSrc clears currentSource_ after loadstart', function() {
+
+  let thing = {
+    off: () => {},
+    one: (el, type, fun) => {
+      el.one(type, Fn.bind(thing, fun));
+    },
+    disposeSourceHandler: () => {},
```
gkatsev: we should verify that this has been called.

gkatsev:
@dmlap or @imbcmdth can you check this out as well and let us know if you think there could be any issues with it for contrib-hls.

src/js/event-target.js
incompl: What do you propose? `() => {}`?

src/js/tech/html5.js
incompl: Will tell me editor to cut it out with the whitespace trimming.

test/unit/tech/html5.test.js
incompl: Good idea

src/js/event-target.js
```diff
@@ -11,7 +11,7 @@ EventTarget.prototype.on = function(type, fn) {
   // Remove the addEventListener alias before calling Events.on
   // so we don't get into an infinite type loop
   let ael = this.addEventListener;
-  this.addEventListener = Function.prototype;
+  this.addEventListener = () => {};
```
dmlap: Point for discussion: should we be careful about using fat-arrows all over the place? I know calling addEventListener() with a non-`this` receiver is a weird thing to do but we could use a plain function here and avoid possibly subverting people's expectations.

src/js/event-target.js
```diff
@@ -23,7 +23,12 @@ EventTarget.prototype.off = function(type, fn) {
 EventTarget.prototype.removeEventListener = EventTarget.prototype.off;
 
 EventTarget.prototype.one = function(type, fn) {
+  // Remove the addEventListener alias before calling Events.on
+  // so we don't get into an infinite type loop
+  let ael = this.addEventListener;
+  this.addEventListener = () => {};
```
dmlap: What would happen if one of the listeners registers an event listener? Something like this:

```js
function maybeReRegister() {
  if (someCondition()) {
    player.one('play', maybeReRegister);
    return;
  }
  alert('done!');
};
player.one('play', maybeReRegister);
```

src/js/tech/html5.js
```diff
@@ -549,9 +549,19 @@ class Html5 extends Tech {
    * @method setSrc
    */
   setSrc(src) {
+    let loadstartlistener = Html5.prototype.loadStartListener_;
+
+    this.off(this.el_, 'loadstart', loadstartlistener);
+    this.one(this.el_, 'loadstart', () => this.one(this.el_, 'loadstart', loadstartlistener));
```
dmlap: Does this assume setSrc() is the only valid way to modify the video element's source? What about setSource()?

Is there a way this can be written that doesn't require adding and removing listeners after the tech's creation? You might be able to use the `readyState` of the HTML element, for instance.

test/unit/tech/html5.test.js
```diff
@@ -442,3 +444,30 @@ test('Html5#reset calls Html5.resetMediaElement when called', function() {
...
 });
+
+test('Html5#setSrc clears currentSource_ after loadstart', function() {
+
+  let disposed = false;
+  let thing = {
+    off: () => {},
+    one: (el, type, fun) => {
+      el.one(type, Fn.bind(thing, fun));
+    },
```
dmlap: Mocking one() and off() may make this test a little too artificial.

incompl: It does successfully test that setSrc clears contentSource_ after loadstart. Do you have any suggestions on how to make this more real?

src/js/event-target.js
dmlap: We're gonna change this to `function() {}`

src/js/event-target.js
dmlap: After reviewing, I don't think this is a problem right now but the event stuff sure is confusing.

src/js/event-target.js
gkatsev: I don't think it matters here at all since we just want an empty function while `Events.on` runs so we don't get into an infinite loop, so, there's no expectations being subverted here.
As for elsewhere, it really depends on what you want the context of your function to be since in arrow functions `this` and `arguments` are never bound.

src/js/tech/html5.js
dmlap: Also, this function literal wouldn't get cleared properly on subsequent calls to setSrc() which would cause the `loadstartlistener_` to get invoked on the wrong `loadstart`

src/js/event-target.js
gkatsev: I think it works fine because the event system creates a clone of the handlers before it starts triggering events: https://github.com/videojs/video.js/blob/2e2dbde4b4b70ecac22736477ca60d20a12cfadd/src/js/utils/events.js#L52-L63

src/js/tech/html5.js
gkatsev: Yeah, this doesn't account for `setSource` which is what ends up being used if a source handler is in effect.
https://github.com/videojs/video.js/blob/2e2dbde4b4b70ecac22736477ca60d20a12cfadd/src/js/player.js#L1916-L1919

gkatsev: why would `loadstartlistener` get called on the wrong `loadstart`?

src/js/tech/html5.js
```diff
@@ -549,9 +549,19 @@ class Html5 extends Tech {
    * @method setSrc
    */
   setSrc(src) {
+    let loadstartlistener = Html5.prototype.loadStartListener_;
```
misteroneill: Are we explicitly using the `Html5#loadStartListener_` function instead of `this.loadStartListener_` because we want to be sure that property hasn't been added to an instance of the tech as something else? Maybe this is a candidate for being module-level private instead of class-level pseudo-private?

src/js/tech/html5.js
dmlap: Here's the scenario:

1. Call setSrc()
1. Video element fires `loadstart`
1. Call setSrc()

loadStartListener_() would be removed on the second call to setSrc() but the inline listener that triggers the loadstartlistener on the next `loadstart` *would not*. So, the very next `loadstart` would run loadstartlistener.

test/unit/tech/html5.test.js
dmlap: I'd create an object that inherited from EventTarget right in the test here and use that.

src/js/tech/html5.js
gkatsev: Oh, right, the first `loadstart` listener would not be removed and would trigger again.

test/unit/tech/html5.test.js
gkatsev: We tried that. `thing` is a a `tech` mock who's `one` method signature is different. We could make it be a standalone `Component` instance, possibly.

src/js/tech/html5.js
incompl: No, it's just arbitrary.

gkatsev:
That looks pretty good to me. @dmlap can you re-review?

gkatsev:
Oh neat, github picked up that it's the same code (I pushed this branch to the repo so that it can run in browserstack).

src/js/tech/tech.js
```diff
@@ -814,18 +814,44 @@ Tech.withSourceHandlers = function(_Tech){
...
     return this;
   };
 
-   /*
-    * Clean up any existing source handler
-    */
-   _Tech.prototype.disposeSourceHandler = function(){
+  // On the first loadstart after setSource
+  _Tech.prototype.firstLoadStartListener_ = function() {
+    this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
```
dmlap: Why use one() if you want the successiveLoadStartListener_ to run on every subsequent loadstart?

src/js/event-target.js
dmlap: That's the crux of the discussion I think we should have. Using fat-arrows removes the ability of the caller to rebind `this`. I've felt it necessary to rebind `this` more than once in my life. Are we intentionally removing that ability or just doing it to save some keystrokes?

src/js/tech/tech.js
incompl: It's because `disposeSourceHandler` removes it. We work around this by re-adding it in `successiveLoadStartListener_` after `disposeSourceHandler` has been called.

src/js/tech/tech.js
```diff
@@ -814,18 +814,44 @@ Tech.withSourceHandlers = function(_Tech){
...
+
+    if (sh !== _Tech.nativeSourceHandler) {
+
+      this.currentSource_ = source;
+
+      // Catch if someone replaced the src without calling setSource.
+      // If they do, set currentSource_ to null and dispose our source handler.
+      this.off(this.el_, 'loadstart', _Tech.prototype.firstLoadStartListener_);
+      this.off(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
+      this.one(this.el_, 'loadstart', _Tech.prototype.firstLoadStartListener_);
```
dmlap: I'm not a fan of using event handler registration as instance-level state. The current set of registered event handlers are difficult to introspect and control flow is harder to trace. I don't see any logical issues with this code so we can move ahead with this but in general, I think adding an instance variable is preferable to this kind of solution.

dmlap:
I'm not in love with the solution but I can live with it. LGTM

Issue #3071, opened 2016-02-03
Autoplay doesn't work when `currentSrc` is set


jamesgorrie:
It appears, if a page is loaded, we then have a `currentSrc` set.

[This then doesn't run the `setSrc` function as the `current.src === this.el_.src`](https://github.com/videojs/video.js/blob/8f2bc92412c4b95641cf271faa5eadb3207c4147/src/js/tech/html5.js#L37)

So either this PR should include:
```JS
if ((this.src() || this.currentSrc()) && this.tag && this.options_.autoplay && this.paused()) {
//...
}
```

Or the `setSrc` method should be updated.

[Here is an example](https://jsbin.com/hohogu/2/edit?html,js,output) where autoplay won't work when we have `currentSrc` set from the browser.

rbran100:
@jamesgorrie Your fix works well in multiple browsers, I posted a PR above.

gkatsev:
That's really weird, I can reproduce it in the jsbin example but if I try to do it myself in the sandbox files, it works just fine in chrome 50.

gkatsev:
Ah, I figured out what the issue is. It's a Mixed Content issue. The jsbin page is running over https but it's trying to load an http video. For whatever reason it fails to load the first video but somehow succeeds for the second.

gkatsev:
Here's an updated example that seems to be working just fine: https://jsbin.com/vefuzohuma/edit?html,js,output

gkatsev:
@jamesgorrie @rbran100 hey, any of you have any thoughts about my comments and questions above?

rbran100:
@gkatsev When I read this mixed content seemed like a likely culprit, I put it on my list to test and get back to and just have not gotten there yet. sorry to leave you hanging.

gkatsev:
Thanks for getting back to me. Let me know if you get any further with that.

gkatsev:
@rbran100 cool, it might still be worse pulling that PR in. If you update it with the latest code we can pull it in.

gkatsev:
I merged the PR anyway, hopefully, it'll make life easier.

PR #3287, opened 2016-04-29
Chrome 50 does not set src attr on video element.

rbran100:
#3071 

## Description
Chrome does not set the src attr and this causes autoplay to never trigger.

## Specific Changes proposed
Check current src, it will hold the current source when multiple source video files are used.

## Requirements Checklist


gkatsev:
I made some comments in #3071, I'm not necessarily opposed to making this change but we should understand it fully before merging if we can.

gkatsev:
@rbran100 hey, would you want to update this PR against master to get it pulled in?

gkatsev:
Ended up rebasing this myself and pulilng it in. Thanks!

PR #3296, opened 2016-05-04
Add title to all clickable components

nickygerritsen:
## Description
As discussed on Slack yesterday, this will add a `title` to all clickable components.
I modified the play toggle test to also check for this, as this was the only test that seemed to test for this already.

Because `controlText` is already called in `createEl` I had to pass the element to set the title on to the function.

## Specific Changes proposed
This sets the `title` HTML attribute on all clickable components when setting `controlText_` initially or changing it using `controlText(...)`.

## Requirements Checklist

### DIFF
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -78,7 +78,7 @@ class ClickableComponent extends Component {
       el.appendChild(this.controlTextEl_);
     }
 
-    this.controlText(this.controlText_);
+    this.controlText(this.controlText_, el);
 
     return this.controlTextEl_;
   }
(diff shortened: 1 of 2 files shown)


src/js/clickable-component.js
```diff
@@ -78,23 +78,27 @@ class ClickableComponent extends Component {
...
    * Controls text - both request and localize
    *
-   * @param {String} text Text for element
+   * @param {String}  text Text for element
+   * @param {Element=} el Element to set the title on
    * @return {String}
    * @method controlText
    */
-  controlText(text) {
+  controlText(text, el=this.el()) {
```
gkatsev: is it possible that `this.el()` will end up returning nothing?

nickygerritsen: Only during the first call, because that is made from `createEl`, at which time `this.el()` does not eixst yet. Afterwards it should always exist, right?

gkatsev:
LGTM.

PR #3297, opened 2016-05-04
Update Dutch language file

nickygerritsen:
## Description
While working on #3296 I saw the Dutch language file was outdated and contained a wrong translation. I updated it to have the new values.

## Specific Changes proposed
This adds missing translations that were present in the English file and updates the translation of captions-related strings.

## Requirements Checklist


lang/nl.json
```diff
@@ -14,13 +14,24 @@
   "Unmute": "Geluid aan",
   "Playback Rate": "Weergavesnelheid",
   "Subtitles": "Ondertiteling",
-  "subtitles off": "Ondertiteling uit",
-  "Captions": "Ondertiteling",
-  "captions off": "Ondertiteling uit",
+  "subtitles off": "ondertiteling uit",
```
hartman: Capitalization ?

nickygerritsen: The source texts aren't capitalized either, so I didn't do that for the Dutch ones.

nickygerritsen: I _do_ wonder why the English versions are not capitalized, as IMHO it looks strange in the track menu's.
@videojs/core-committers any idea why this is the case?

hartman: And even if it is intentional, I think that in that particular spot, it would be a stylistic choice, for which there is the text-transform css property.

mister-ben: It's because the source text is the track's kind + ' off' in https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/off-text-track-menu-item.js#L23

I don't see a reason to not capitalise the translation if that's preferred. I'd rather capitalise the English too, but that would also require including the en translations which doesn't happen now (en.json is there as a template).

hartman: `label = options['kind'].charAt(0).toUpperCase() + options['kind'].slice(1);` ?

gkatsev: I'm chalking it up to a stylistic choice unless *everyone* except @mmcc prefers the capitalization :P

gkatsev: Also, the menu has a text-transform css applied to it in the default skin, so, it would all be lower case. Though, that's also easily solved via a text transform. Ultimately, it doesn't matter *that* much what the translation files say specifically, as long as the keys match what's in the code, a user can set whatever they want via css.

gkatsev:
LGTM

lang/nl.json
nickygerritsen: So what's the conclusion? With or without capitalization?

gkatsev: I would just leave it lowercase.

PR #3304, opened 2016-05-06
Add default styling for print

hartman:
## Description
Printing a page containing video.js currently is a rather messy state.
Controls are partially visible, as many of them are text (videojs-font), yet all images and backgrounds
are usually stripped by the user-agent from the print.

## Specific Changes proposed
Hide all elements other than the poster and/or the tech.

## Requirements Checklist
  - [ ] Docs/guides updated

By default in print, only show either the poster, or the current tech
element. This matches the behavior of the HTML5 video element for most
user agents.

The video-js poster currently uses a background image, which usually gets stripped by browsers,
when printing. I also propose changing the poster element to be an img element instead of a background image.

### DIFF
--- a/src/css/_print.scss
+++ b/src/css/_print.scss
@@ -1,5 +1,5 @@
 @media print {
-  .video-js *:not(.vjs-tech):not(.vjs-poster) {
+  .video-js > *:not(.vjs-tech):not(.vjs-poster) {
     visibility:hidden;
   }
 }
--- a/src/css/_print.scss
+++ b/src/css/_print.scss
@@ -1,0 +1,5 @@
+@media print {
+  .video-js *:not(.vjs-tech):not(.vjs-poster) {
+    visibility:hidden;
+  }
+}
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -1,7 +1,6 @@
 @import "variables";
 @import "private-variables";
 @import "utilities";
-@import "print";
 
 @import "node_modules/videojs-font/scss/icons";
 
@@ -40,3 +39,5 @@
 @import "components/adaptive";
 @import "components/captions-settings";
 @import "components/modal-dialog";
+
+@import "print";
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -1,6 +1,7 @@
 @import "variables";
 @import "private-variables";
 @import "utilities";
+@import "print";
 
 @import "node_modules/videojs-font/scss/icons";
 


src/css/_print.scss
```diff
@@ -0,0 +1,5 @@
+@media print {
+  .video-js *:not(.vjs-tech):not(.vjs-poster) {
+    visibility:hidden;
```
misteroneill: Are we sure that `visibility: hidden` does not create blank areas in the printed page? Is there some advantage to it over `display: none`?

hartman: Using visibility:hidden guarantees that we don't have a reflow of the page. 'you print what you see' in terms of dimensions. When using display:none; you are actively removing things from the canvas, which can influence the dimensions and behavior of other elements.

In reality, it probably won't matter much unless someone writes a 'bad' plugin.

hartman:
[This diff](https://github.com/hartman/video.js/commit/da1b634f26327989e4817bcc6690eb01f25dfb88) contains some further ideas on what I want to do for print. But it is probably better to deal with that in a later pull request.

src/css/_print.scss
```diff
@@ -0,0 +1,5 @@
+@media print {
+  .video-js > *:not(.vjs-tech):not(.vjs-poster) {
```
gkatsev: Is there an *easy* way of supporting IE8?

gkatsev:
One question that's mostly optional, LGTM otherwise.

src/css/_print.scss
mmcc: I vote we don't even ask this question anymore :)

gkatsev: Haha, that's why I stressed "easy". As part of 5.x, we should still try and support IE8, we don't have to try very hard though :)

gkatsev: And we definitely shouldn't actively break it.

mmcc: Yeah sorry, that was mostly a troll. You're right on all points, I was just adding the obligatory anti-IE8-support voice.

misteroneill:
LGTM if @gkatsev's question gets an answer 😄

src/css/_print.scss
hartman: Welk we could invert the selector to only hide things explicitly, but you would have to basically do that for every single plugin or element that might want to add something to the player DOM. I don't really think that that is maintainable in practice.

gkatsev: Yeah, not worth it.

PR #3314, opened 2016-05-11
null out currentSource_ on tech dispose and src change

gkatsev:
Follow up to https://github.com/videojs/video.js/pull/3285, we need `currentSource_` to be cleared out if we're switching sources on a tech and we're no longer using a source handler. Plus, when the tech is disposed, we should clear out the `currentSource_`.

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -813,6 +813,8 @@ Tech.withSourceHandlers = function(_Tech){
     // than clear all of our current tracks
     if (this.currentSource_) {
       this.clearTracks(['audio', 'video']);
+
+      this.currentSource_ = null;
     }
 
     if (sh !== _Tech.nativeSourceHandler) {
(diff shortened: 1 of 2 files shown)


src/js/tech/tech.js
```diff
@@ -813,6 +813,8 @@ Tech.withSourceHandlers = function(_Tech){
     // than clear all of our current tracks
     if (this.currentSource_) {
       this.clearTracks(['audio', 'video']);
```
gkatsev: @BrandonOCasey is there a reason why we don't want to just *always* call `clearTracks()` here?

BrandonOCasey: I'm not sure I entirely remember but I think it had to do with saving keeping Audio/Video Tracks that were added before we had a currentSource

gkatsev: Ok, I'll see what I can find out. Currently, the tests break because of this.

gkatsev: Ended up removing clearing out the `currentSource_` in `disposeSourceHandler` and only doing it inside this if statement.

incompl:
Looks good to me

misteroneill:
LGTM

Issue #3041, opened 2016-01-26
videojs 6.0 megathread


gkatsev:
Just opening this up with some thoughts of what could go into a videojs 6.0 and to get everyone's opinions on what stuff should go into it.
- make `sourceOrder` a default. This will make it so that we're playing back items based on source order regardless of the tech order. Currently, this option is set to `false` to maintain the older behavior of finding the best source based on the best tech.
- use the new fangled (yet to exist) fully _accessible_ volume control
- allow users to ask the player which plugins have been loaded on it
- better plugin support -- this has yet to be determined what it will be but popcornjs's plugins are pretty nice
- drop old browser support like IE8 and IE9.

More to come, stay tuned.

misteroneill:
Is there a policy (like a SLA) for how long the videojs org will "officially" support the previous version with patches?

gkatsev:
Nope but we can probably come up with something. It'll probably be useful because of the IE8 support since we could tell people to use the 5.x release for older browser support.

OwenEdwards:
At a minimum, I'd like to get accessibility fixes in there for:

#2729 Handling keyboard focus when the big play button is activated
#2156 Handling keyboard accessibility when the Control Bar is hidden

It'd be nice to add:
#2343 (need reworking) Support for description text tracks
#2746 Accessiblity of the Captions Settings dialog

And then the accessible volume control widget, and any accessibility improvements to the menus which might break the DOM.

I'm working on most of these, but I'll certainly need some help & advice.

gkatsev:
@OwenEdwards sounds good. Hopefully, most of these can even just go into vjs 5.x and don't need breaking a major version bump.

dmlap:
Does an API that lists plugins require a major version bump? Seems additive to me.

misteroneill:
@dmlap It does if you're trying to know which plugins have been invoked on a player (vs. what plugins are on the `Player.prototype`). I think the only way that could be achieved today is to wrap the function that gets added to the `Player.prototype` in something that tracks that, but that'd be a breaking change because the value of `Player.prototype.foo` would not be the same as what you passed to `.plugin()`.

I can see how a plugin system could be implemented piecemeal without requiring a major version bump. It's dead simple to continue supporting the current `.plugin()` form, but doing something like allowing a different set of parameters to be compatible with a more robust plugin system seems like it'd be do-able in a minor release.

That might be worth exploring.

hartman:
[...]

PR #3321, opened 2016-05-13
Add Tooltip component

norlin:
General Tooltip component for any ClickableComponent.

Allows to show properly positioned tooltips for buttons and so on.

By default, it’s disabled.

Could be enabled for all basic buttons with `{"tooltips": true}` option.

Could be enabled for specific control with per-component option 
`{"MyComponent": {"tooltip": true}}`

Allows to set up custom handler element for specific control:
`{"MyComponent": {"tooltip": "#myTooltipHandler"}}`

Not yet working for MenuButton and PopupButton since they’re using pseudo-element for icons.

Disabled for BigPlayButton even with general `{"tooltips": true}` option – but you can enable it with per-component option (will need to add additional styles as well).

Also it's possible to rewrite mouse-display-tooltip with this component.

### DIFF
--- a/src/css/components/_tooltip.scss
+++ b/src/css/components/_tooltip.scss
@@ -1,0 +1,23 @@
+.video-js .vjs-tooltip {
+  position: absolute;
+  top: -3em;
+  left: 0;
+  z-index: 10;
+
+  display: none;
+  padding: 6px 8px 8px 8px;
+
+  @include border-radius(0.3em);
+  @include background-color-with-alpha($primary-background-color, $primary-background-transparency);
+
+  color: $primary-foreground-color;
+  font-size: 0.9em;
+  font-family: $text-font-family;
+  white-space: nowrap;
+
+  pointer-events: none;
+}
+
+.video-js .vjs-tooltip-active {
+  display: block;
+}
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -8,6 +8,7 @@
 @import "components/big-play";
 @import "components/button";
 @import "components/close-button";
+@import "components/tooltip";
 
 @import "components/menu/menu";
 @import "components/menu/menu-popup";
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -38,6 +38,18 @@ class BigPlayButton extends Button {
     this.player_.play();
   }
 
+  /**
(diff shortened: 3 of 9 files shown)


nickygerritsen:
This is similar, but different, to #3296, where we enable it by default for all clickable components and just use the `title` attribute.

I'm not sure what's the opinion on that one vs. this one.

norlin:
@nickygerritsen: `title` attribute does not allow to control it's position and handler. In general, custom tooltips is much more customizable.

nickygerritsen:
I understand the differences :). I'm interested in knowing what the others think about this idea.

gkatsev:
I'll have to think about it but would be interesting to see what the other @videojs/core-committers think about this.
I wonder if it'll be better to just expose the correct information on the elements and the components so that this can be done outside of videojs, in a plugin, for example.

nickygerritsen:
If we do the latter I think adding a controltextchanged event would be enough already?

gkatsev:
@norlin does exposing the correct information and providing an event for when that changes sound good to you? If so, what information should be exposed besides the title text itself?

norlin:
@gkatsev are you sure a tooltips should be outside core vjs? Video.js itself uses a kind of tooltips for the progress-bar, for example. It could be unified with other tooltips.

gkatsev:
I totally didn't realize how long it's been since the last comment. Sorry for the late reply.

I'm not sure that tooltips should be outside of core. Though, I would much prefer if we develop and iterate on the tooltips outside of core. The tooltips on the progress bar (particularly with the keep tooltips inside option) are relatively custom and I don't think it would be easy to unify them with more generic tooltips. Plus, I don't think those can be unified in a backwards-compatible way.
We're already experimenting with making a repo for a feature we want but aren't sure how it'll look or work before bringing it into core. See [the videojs settings menu project](https://github.com/videojs/videojs-settings-menu).
I think at this point just providing the necessary information so that tooltips could be created in outside of core is where we're at. Maybe as part of a major version update, we could re-work and unify tooltips.

I'm going to close this PR because it isn't going to get merged as is. However, I do want to keep this discussion going either here or maybe in the [videojs 6.0 megathread](https://github.com/videojs/video.js/issues/3041) or in a new issue about tooltips. Thanks for getting the discussion started.

PR #3327, opened 2016-05-17
Fix test for tooltips in IE8

nickygerritsen:
This fixes the broken build from merging #3296.

Nothing too excited, but IE8 has no `"` in the class.


gkatsev:
LGTM

Issue #2382, opened 2015-07-20
5.0: Consider adding "sourcechange" event


nickygerritsen:
Our video.js based project allows to switch sources and to detect when a source has been switched we overwrote the `player.src` method to emit a `sourcechange` event.

However, we want this event to only trigger as soon as it is guaranteed that `player.src()` will return the new source.

Because setting the source using `player.src(...)` is done in a `player.ready()` block, this is cumbersome for us.

Currently we have code like the following:

``` js
// video.js sets source using ready function. Because that function is performed async, we also need to
// do our source change event triggering in a ready function, as we will otherwise trigger the change before
// video.js updates the source
player.ready(function()
{
    // Now we need to do a setTimeout, as otherwise it is not guaranteed that this code happens after the source change
    setTimeout(function()
    {
        player.trigger('sourcechange');
    }, 1);
});
```

It would be really useful if video.js itself could emit something like this `sourcechange` event as soon as `player.src()` would return the new source, i.e. as soon as the `techSet('src')` is performed.

Is this something that could be added?

gkatsev:
I think you'll find that the `loadstart` event already will function in just about the same way you want `sourcechange` to function.
Also, you should be able to set the handler even before the player is ready. As long as the player is available, you can set events on it.

nickygerritsen:
I will look into the `loadstart` event.

Setting the handler before the player is ready works, but in that case my `sourcechange` URL is emitted before the source is actually changed.

gkatsev:
Ah, I see. Well, hopefully, the `loadstart` event will be useful :)

nickygerritsen:
Loadstart does not work if preload = none on the video element i guess? We currently have at least some failing tests in that case if I switch to the loadstart event.

gkatsev:
Yes, you're right. In the `preload=none` case, it will only fire when the video gets played.
Can you elaborate on why you need to know exactly when a source has changed?

nickygerritsen:
So we are creating a quality switching plugin. When a user switches a quality while still playing or when the video is loaded, we want to seek to the time of the old source.

We use this source change event to start listening for duration changed events so we know when we can seek.

I'm also using the event to update my quality switching UI to determine if I should show the UI and what quality is currently active

heff:
[...]

Issue #1194, opened 2014-05-05
Handle the loadstart event at the player level


heff:
The loadstart event is the closest thing to something like a 'srcchange' event defined in the HTML5 spec. It's fired whenever the src is set or the source selection process begins, and is fired even if preload=none.

Currently the player relies on the tech to fire the loadstart event and bubble it up to the player, but this isn't exactly right because, in the scope of the player, that would only happen AFTER the source selection process, when the selected source is handed to the tech. That means if a source isn't selected, the loadstart event won't get fired and we won't know that src change was even attempted.

One place this could have an impact is the error state (the player has an existing error). We should be able to use the loadstart event to clear the error state, and possible other player states (networkState, readyState) if we decide to handle them at the player level.

Another option would be to create a `srcchange` or `contentupdate` event that the player fires whenever src() is used. The downside of this is we would be inventing a new event and detracting from how the html5 video element works, which we try to avoid, and could have unforeseen implications down the road.

The most difficult use case to work through so far is the ads use case, and understanding how we need the events to work in the process of switching between content and ad sources behind the scenes.

Also, wrap #1672 into this.
And #2382

IRC CHAT LOG: @dmlap, @heff

```
dmlap: heff: sorry, had to walk away. I like the "feed it directly to plugins" option but it gets kind of tricky if you consider playlists
dmlap: what is going to keep track of plugins that need metadata and re-initialize them when the src changes?
dmlap: maybe a playlist plugin could do that
dmlap: but the plugins would still have to register in some interesting way, or listen for an event
dmlap: either of which seems equivalent to just listening for loadstart and then querying the player
heff: yeah, that's true. So you like the 'store in options' method too then?
heff: I think the loadstart event is the right one to signal plugins to update their info
dmlap: yeah, that's my preference
dmlap: the only trick is when ads are involved
dmlap: we invented this new event, 'contentupdate' to indicate when src was changed for reasons *unrelated* to ads
dmlap: i wish we could just use loadstart but client side ads hijack the video element on mobile devices to do their business
dmlap: and they may emit multiple loadstarts
dmlap: i'm fine with code like this:
dmlap: player.on(player.ads ? 'contentupdate' : 'loadstart', function() { … })
dmlap: but it's not ideal
heff: yeah, I was thinking of adding a 'srcchange' event for similar reason, but then reading through the spec, the loadstart event happens at the beginning of the source selection process, and is triggered even if preload=none
[...]

PR #3329, opened 2016-05-18
Add beforesourcechange and sourcechange events

misteroneill:
## Description
This adds two new events which are triggered by the `Player#src()` method.

- `"beforesourcechange"` fires before a playable source is set on the player.
- `"sourcechange"` fires after the playable source has been set on the player (including after any tech-readiness and autoplay/preload behavior is triggered).

These are non-standard events, but I'd like to open discussion on adding them. They have a lot of practical application as we've seen at Brightcove and other companies where core contributors work. This may be ultimately more appropriate in a separate project - in order to keep the core more pure, but modifying the `Player` prototype seems out of the scope of plugins to me. An alternative might be coming up with some kind of flag that notes an event as non-standard.

Discuss!

## Requirements Checklist
- [x] Feature implemented
  - [x] Unit Tests updated
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1903,11 +1903,14 @@ class Player extends Component {
         // the tech loop to check for a compatible technology
         this.sourceList_([source]);
       } else {
-        let previous = this.currentSrc();
+        const previous = {
+          src: this.currentSrc(),
+          type: this.currentType()
+        };
 
         this.trigger('beforesourcechange', {
           current: previous,
-          next: source.src
+          next: source
         });
 
         this.cache_.src = source.src;
@@ -1936,7 +1939,7 @@ class Player extends Component {
 
           this.trigger('sourcechange', {
             previous,
-            current: source.src
+            current: source
           });
 
         // Set the source synchronously if possible (#2326)
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1903,6 +1903,13 @@ class Player extends Component {
         // the tech loop to check for a compatible technology
         this.sourceList_([source]);
       } else {
+        let previous = this.currentSrc();
+
+        this.trigger('beforesourcechange', {
+          current: previous,
+          upcoming: source.src
+        });
+
         this.cache_.src = source.src;
         this.currentType_ = source.type || '';
 
@@ -1926,6 +1933,11 @@ class Player extends Component {
           if (this.options_.autoplay) {
             this.play();
           }
+
+          this.trigger('sourcechange', {
+            previous,
+            current: source.src
+          });
 
         // Set the source synchronously if possible (#2326)
         }, true);
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1903,10 +1903,14 @@ class Player extends Component {
         // the tech loop to check for a compatible technology
         this.sourceList_([source]);
       } else {
-        const previous = {
-          src: this.currentSrc(),
-          type: this.currentType()
-        };
+        const previous = {src: this.currentSrc()};
+
+        // We only want to include the type if it exists. This keeps the
+        // objects sent from the sourcechange events similar: `type` will
+        // not be present if it's not known.
+        if (this.currentType_) {
+          previous.type = this.currentType_;
+        }
 
         this.trigger('beforesourcechange', {
           current: previous,
@@ -1937,10 +1941,12 @@ class Player extends Component {
             this.play();
           }
 
-          this.trigger('sourcechange', {
-            previous,
-            current: source
-          });
+          window.setTimeout(() => {
+            this.trigger('sourcechange', {
+              previous,
+              current: source
+            });
+          }, 0);
 
         // Set the source synchronously if possible (#2326)
         }, true);
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1907,7 +1907,7 @@ class Player extends Component {
 
         this.trigger('beforesourcechange', {
           current: previous,
-          upcoming: source.src
+          next: source.src
         });
 
         this.cache_.src = source.src;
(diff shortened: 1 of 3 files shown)


gkatsev:
You forgot to mention https://github.com/videojs/video.js/issues/2382

src/js/player.js
```diff
@@ -1903,6 +1903,13 @@ class Player extends Component {
         // the tech loop to check for a compatible technology
         this.sourceList_([source]);
       } else {
+        let previous = this.currentSrc();
+
+        this.trigger('beforesourcechange', {
```
gkatsev: I was going to ask if this should be triggered on line 1888 (and I do think that it probably should) but given how recursive `src` is, we probably don't want to move this up.

src/js/player.js
```diff
@@ -1903,6 +1903,13 @@ class Player extends Component {
...
+          current: previous,
```
gkatsev: should we send along the entire source object, including the type?

src/js/player.js
```diff
@@ -1927,6 +1934,11 @@ class Player extends Component {
             this.play();
           }
 
+          this.trigger('sourcechange', {
```
gkatsev: should this be wrapped in a setTimeout like mentioned in https://github.com/videojs/video.js/issues/2382? The source loading algorithm is asynchronous, so, if someone listens to `sourcechange` and calls `player.currentSrc()` they may not get the latest value.

src/js/player.js
misteroneill: My thought was that we'd only call it when we are changing to a playable source because that's what people will care about in most cases, I think.

gkatsev:
@nickygerritsen would love to get your input on this given you opened #2382.

src/js/player.js
misteroneill: Good point. 👍

src/js/player.js
misteroneill: Yeah, that probably makes sense.

misteroneill:
Also, I'd love to get some @videojs/core-committers thoughts on the notion of firing non-standard events like this. And whether we should namespace them to make it crystal clear they're not standard.

gkatsev:
Oh, cool, totally forgot that you can pass a dataobject as a second param to trigger.

dmlap:
Would you mind elaborating on some use-cases? I think it would be easier to evaluate if we were talking about specific examples where the native events aren't sufficient.

heff:
Don't forget about #1194! Some good discussion between @dmlap and I around this.

I'd be interested to know specifically how `sourcechange` would differ from `loadstart` in this case, and what would be the added benefits. If they are different enough, we should make sure there's a strict order of events between sourcechange and loadstart. It could get really confusing if they fire in different orders, and by the description it sounds like there's some edge cases where that could happen (without looking too deeply into it).

If they turn out to be not different enough, we could also consider just adding a `beforeloadstart`, which would be more of piggybacking on native events than creating new ones.

misteroneill:
The problem with `loadstart` is that it is not reliable.

Our existing analytics code attempted to shoehorn it and ended up with bugs and unintelligible event spaghetti. Different browsers behave differently - for example, in IE9-11, `play` fires before `loadstart` in some cases. And if you're on IE11 on Windows 8+ the order can be `play`, `playing`, and then `loadstart`. This is particularly noticeable when used in conjunction with something like `brightcove/videojs-playlist`, which, essentially, does this:

```js
player.src(...);
player.play();
```

These issues mean we cannot reliably detect when a new video has started without doing something like polling `currentSrc`, which is a much uglier solution in my mind than adding a non-native event.

I would be agreeable to the idea of making it `beforeloadstart` and removing the `sourcechange`, though. We don't have a need for `sourcechange` at the moment, though @nickygerritsen apparently did.

gkatsev:
Maybe for videojs-playlist, we should have `play` be called on a `setTimeout(0)`?

Also, related to #1194, @incompl was thinking that maybe it's time for videojs to make it's own set of events (based on the native events but maybe with different names) that have some guarantees. Like, a version of `loadstart` that will always be the first event you will see for a given source, or a `playing` event can never happen before a `firstcanplay`.

misteroneill:
I'm obviously a vote in favor of something like that. And when considering it, also consider the idea of namespacing non-native/standard events. Could be something like `'vjs:custom-event'`.

dmlap:
@misteroneill are the browser behaviors you're seeing not compliant with the standard? If that's the case, I think it's reasonable to normalize their behavior in video.js (and maybe that gets rid of the need for new events).

misteroneill:
As we discussed in person, the standard doesn't seem to give us any recommendation about ordering of events like `loadstart`, `play`, and `playing`.

I've been spending the morning observing behaviors in Chrome with different combinations of `preload`, `autoplay`, playlists, and video tech. There is a ton of variation.

HTML5, thankfully, is fairly consistent. It follows the `loadstart -> play -> playing` pattern for all combinations except `preload: "none"`, where it's `play -> loadstart -> playing` in most cases.

Flash is less consistent. In fact, with `autoplay: true` and playlists, we get patterns like:

```
play
playing
play
playing
play
playing
loadstart
play
playing
```

misteroneill:
Thought I'd note the logic we're now using internally to detect source changes: on `loadstart`, `canplay`, and `playing` we look for a new source and only once _all 3 events_ have triggered, we can consider the source changed. The list previously included `play`, but IE9 sometimes fires `play` before `currentSrc()` returns a new value.

Maybe this will help someone else struggling with similar issues.

gkatsev:
Closing again for now. We have https://github.com/brightcove/videojs-per-source-behaviors and we will potentially do some more work around this soon.

PR #3338, opened 2016-05-24
use direct versions for dependencies

gkatsev:


dmlap:
Interesting... I'm guessing we've hit too many accidental update issues recently?

gkatsev:
Well, also, the cdn deploy script doesn't like if it the swf or fonts have a range version.

dmlap:
Ah. LGTM.

PR #3342, opened 2016-05-27
only call source handler dispose once on the same source handler

BrandonOCasey:
## Description
Currently we will try to dispose of SourceHandlers twice when we call setSource multiple times, this causes videojs to throw an error on the second setSource

## Specific Changes proposed
* Null out sourceHandler_ on the tech when sourceHandler is disposed

## Requirements Checklist

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -855,6 +855,7 @@ Tech.withSourceHandlers = function(_Tech){
       this.off(this.el_, 'loadstart', _Tech.prototype.firstLoadStartListener_);
       this.off(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
       this.sourceHandler_.dispose();
+      this.sourceHandler_ = null;
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/js/tech/tech.js
```diff
@@ -794,6 +794,9 @@ Tech.withSourceHandlers = function(_Tech){
     * @return {Tech} self
     */
    _Tech.prototype.setSource = function(source){
+    if (!this.el_) {
```
gkatsev: should this ever happen?

test/unit/tech/tech.test.js
```diff
@@ -147,6 +147,97 @@ test('dispose() should clear all tracks that are added after creation', function
...
+    canHandleSource: function() {
+      return true;
+    },
+    handleSource: function(source, tech, options) {
+      return handler;
+    }
+  });
+
+  let tech = new MyTech();
+  tech.setSource({
```
gkatsev: we should set a non-known source here. Here's the test for the previous work: https://github.com/BrandonOCasey/video.js/blob/67e294f0faf8b52e5e88de4576500d1d465e4bee/test/unit/tech/tech.test.js#L453-L495

PR #3343, opened 2016-05-27
only call source handler dispose once on the same source handler

BrandonOCasey:
## Description
Currently we will try to dispose of SourceHandlers twice when we call setSource multiple times, this causes videojs to throw an error on the second setSource

## Specific Changes proposed
* Null out sourceHandler_ on the tech when sourceHandler is disposed

## Requirements Checklist


src/js/tech/tech.js
```diff
@@ -855,6 +855,7 @@ Tech.withSourceHandlers = function(_Tech){
       this.off(this.el_, 'loadstart', _Tech.prototype.firstLoadStartListener_);
       this.off(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
       this.sourceHandler_.dispose();
+      this.sourceHandler_ = null;
```
imbcmdth: Not much to argue about here.

LGTM!

gkatsev:
LGTM.

PR #3354, opened 2016-06-01
Fix to avoid duplicate chapters by building a new list each time

vdeshpande:
## Description
Duplicate chapters list gets added to the existing chapters item list whenever we try adding a chapter using `addRemoteTextTrack()`
This PR is to make sure that duplicate chapters do not get added to that list.

## Specific Changes proposed
Changes proposed within this PR are as follows:
1) We always keep our chapters item list clean and fresh each time a chapter is added, by always creating a **new chapters list** every single time
_This approach makes it easy to avoid duplicate chapters and only provide a new chapter item list whenever the user tries to create one._ 

## Requirements Checklist
  - [x] Docs/guides updated
  - [x] Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -105,8 +105,6 @@ class Component {
    */
   dispose() {
     this.trigger({ type: 'dispose', bubbles: false });
-
-    // Dispose all children.
     if (this.children_) {
       for (let i = this.children_.length - 1; i >= 0; i--) {
         if (this.children_[i].dispose) {
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -105,7 +105,7 @@ class Component {
    */
   dispose() {
     this.trigger({ type: 'dispose', bubbles: false });
-    
+
     // Dispose all children.
     if (this.children_) {
       for (let i = this.children_.length - 1; i >= 0; i--) {
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -105,6 +105,8 @@ class Component {
    */
   dispose() {
     this.trigger({ type: 'dispose', bubbles: false });
+    
+    // Dispose all children
     if (this.children_) {
       for (let i = this.children_.length - 1; i >= 0; i--) {
         if (this.children_[i].dispose) {
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -106,7 +106,7 @@ class Component {
   dispose() {
     this.trigger({ type: 'dispose', bubbles: false });
     
-    // Dispose all children
+    // Dispose all children.
     if (this.children_) {
       for (let i = this.children_.length - 1; i >= 0; i--) {
         if (this.children_[i].dispose) {
--- a/src/js/control-bar/text-track-controls/chapters-button.js
+++ b/src/js/control-bar/text-track-controls/chapters-button.js
@@ -99,7 +99,6 @@ class ChaptersButton extends TextTrackButton {
       });
       menu.children_.unshift(title);
       Dom.insertElFirst(title, menu.contentEl());
-      this.addChild(menu);
     } else {
         // We will empty out the menu children each time because we want a 
         // fresh new menu child list each time
@@ -146,13 +145,15 @@ class ChaptersButton extends TextTrackButton {
   // we try to put the chapter title into the DOM
 
   update() {
-      if(this.children_.length <= 1) {
-        this.firstCall = true;
-        super.update();
-      } else if(this.children_.length > 1) {
-          this.children_.pop();
-      }
+    
+    // We want to make sure that we are always using the code below only when
+    // we have update called on chapters and not anything else
+    if (this.kind_ === "chapters") {
+
+      //Call the parent update handler to add in the chapters menu for update
+      super.update();
     }
+  }
 }
 
 ChaptersButton.prototype.kind_ = 'chapters';
--- a/src/js/control-bar/text-track-controls/chapters-button.js
+++ b/src/js/control-bar/text-track-controls/chapters-button.js
@@ -100,15 +100,12 @@ class ChaptersButton extends TextTrackButton {
       menu.children_.unshift(title);
       Dom.insertElFirst(title, menu.contentEl());
       this.addChild(menu);
-    }
-    else {
-      // We will empty out the menu children each time because we want a 
-     // fresh new menu child list each time
-     items.forEach(function(item) {
-       menu.removeChild(item);
-     });
-     // Empty out the ChaptersButton menu items because we no longer need them
-     items = [];
+    } else {
+        // We will empty out the menu children each time because we want a 
+        // fresh new menu child list each time
+        items.forEach(item => menu.removeChild(item));
+        // Empty out the ChaptersButton menu items because we no longer need them
+        items = [];
     }
 
     if (chaptersTrack && chaptersTrack.cues == null) {
@@ -145,7 +142,17 @@ class ChaptersButton extends TextTrackButton {
     this.items = items;
     return menu;
   }
+  // Overriding the update button because duplicate chapters get created each time when
+  // we try to put the chapter title into the DOM
 
+  update() {
+      if(this.children_.length <= 1) {
+        this.firstCall = true;
+        super.update();
+      } else if(this.children_.length > 1) {
+          this.children_.pop();
+      }
+    }
 }
 
 ChaptersButton.prototype.kind_ = 'chapters';
--- a/src/js/control-bar/text-track-controls/chapters-button.js
+++ b/src/js/control-bar/text-track-controls/chapters-button.js
@@ -148,7 +148,7 @@ class ChaptersButton extends TextTrackButton {
     
     // We want to make sure that we are always using the code below only when
     // we have update called on chapters and not anything else
-    if (this.kind_ === "chapters") {
+    if (this.kind_ === 'chapters') {
 
       //Call the parent update handler to add in the chapters menu for update
       super.update();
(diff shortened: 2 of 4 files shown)


src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -78,7 +78,9 @@ class ChaptersButton extends TextTrackButton {
     let items = this.items = [];
 
     for (let i = 0, length = tracks.length; i < length; i++) {
-      let track = tracks[i];
+
+      // We will always choose the last track as our chaptersTrack
+      let track = tracks[length - 1];
```
gkatsev: wouldn't this always choose the last track? Don't we want the last chapters track?

gkatsev: Also, the track chosen here and in `createItems` is now different.

gkatsev: Actually, does this even need `createItems` since it's all happening in there? We should move the item creation login from this method into `createItems`.

vdeshpande: I think for the last chapters track, due to the break statement....it does not always choose the last track. 
Let's say we have multiple chapters list and if one the in-between chapters kind matches ; it will choose a track from in-between of the list and break out from the loop. 
I wanted to always make sure that we are taking the last track no matter how big the list is.

Also track chosen in `createItems` would be different, but we are popping out the items in `createMenu` while always maintaining a chapters list, so that would not be an issue I think.

Well we actually do not need `createItems` if we decide to take this approach.

gkatsev: but `length` never changes, so, it'll always choose `track[track.length - 1]`

gkatsev: I think moving things there is good because it'll be consistent with the other menu buttons.

vdeshpande: Ok, will try to see if I can move things around. :)

Regarding choosing the last track, earlier it was like this

```
for loop {
let track = tracks[i] 
(if track matches kind as chapters)  {
  chaptersTrack = track;
  break;
 }
}

Now let's say i = 1 and kind matches with chapters, 
but we want the 2nd track( i = 2) to be the correct and most recent 
chapters Track that has to be added to chapters list
```

This will cause the `chaptersTrack` to always take value of the first track, and break out from the loop.
If this happens, the last most recent track added never gets added to the list as `chaptersTrack` will always take value of the first previous track. Would this be not the case ?

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -98,7 +100,16 @@ class ChaptersButton extends TextTrackButton {
       menu.children_.unshift(title);
       Dom.insertElFirst(title, menu.contentEl());
     }
-
+    // We will empty out the children each time because we want a 
+    // fresh new list each time
+    
+    if (menu.contentEl_.childElementCount > 1) {
```
gkatsev: these should also be removed from `this.items`, so, probably better to loop through those and dispose each item.

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -112,6 +123,14 @@ class ChaptersButton extends TextTrackButton {
     if (chaptersTrack && chaptersTrack.cues && chaptersTrack.cues.length > 0) {
       let cues = chaptersTrack['cues'], cue;
 
+      // We want to empty out the item children but just keep the chapter title item
+      if (menu.children_.length > 2) {
```
gkatsev: we should try and avoid accessing "private" variables if there are other ways of removing things. Given that each item is added as a child, we should instead remove each child using the appropriate API method.

src/js/control-bar/text-track-controls/chapters-button.js
gkatsev: we should be able to loop through them and then removeChild them.

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -97,6 +99,16 @@ class ChaptersButton extends TextTrackButton {
       });
       menu.children_.unshift(title);
       Dom.insertElFirst(title, menu.contentEl());
+      this.addChild(menu);
+    }
+    else {
```
misteroneill: I think this should be on the previous line (i.e. `} else {`).

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -97,6 +99,16 @@ class ChaptersButton extends TextTrackButton {
...
+      // We will empty out the menu children each time because we want a
```
misteroneill: Indented one space too far.

src/js/component.js
```diff
@@ -105,8 +105,6 @@ class Component {
    */
   dispose() {
     this.trigger({ type: 'dispose', bubbles: false });
-
-    // Dispose all children.
```
gkatsev: why remove this?

src/js/menu/menu-button.js
```diff
@@ -41,6 +41,11 @@ class MenuButton extends ClickableComponent {
     if (this.menu) {
       this.removeChild(this.menu);
     }
+    // We need to remove the menu because we are already adding one before. 
+    // This can create duplicate menu items causing tests to fail
+    if(this.menu === undefined && this.firstCall) {
```
gkatsev: this duplicated the code from lines 41-34.

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -97,6 +99,13 @@ class ChaptersButton extends TextTrackButton {
       });
       menu.children_.unshift(title);
       Dom.insertElFirst(title, menu.contentEl());
+      this.addChild(menu);
```
gkatsev: should this line just not be here?

src/js/utils/dom.js
```diff
@@ -171,17 +171,20 @@ const elIdAttr = 'vdata' + (new Date()).getTime();
...
-  if (!id) {
-    id = el[elIdAttr] = Guid.newGUID();
-  }
-
-  if (!elData[id]) {
-    elData[id] = {};
-  }
-
-  return elData[id];
+  //if(el) {
```
gkatsev: work-in-progress checked in.

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -124,17 +133,26 @@ class ChaptersButton extends TextTrackButton {
...
     return menu;
   }
-
+  // Overriding the update button because duplicate chapters get created each time when
+  // we try to put the chapter title into the DOM
+
+  update() {
+      if(this.children_.length <= 1) {
+        this.firstCall = true;
+        super.update();
```
gkatsev: This can still be called if a non-chapters track is added. We should make sure that it only does anything when chapters are added.
The `update` method is an event handler.

src/js/component.js
vdeshpande: Added the comment back

gkatsev: This shouldn't be showing up as changed in the diff.

src/js/utils/dom.js
```diff
@@ -171,17 +171,19 @@ const elIdAttr = 'vdata' + (new Date()).getTime();
...
-
-  if (!id) {
-    id = el[elIdAttr] = Guid.newGUID();
-  }
-
-  if (!elData[id]) {
-    elData[id] = {};
-  }
+      
+    let id = el[elIdAttr];
```
gkatsev: The indent is off. It shouldn't be showing up as changed in the diff.

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -124,17 +132,28 @@ class ChaptersButton extends TextTrackButton {
...
     return menu;
   }
+  // Overriding the update button because duplicate chapters get created each time when
+  // we try to put the chapter title into the DOM
 
+  update() {
+    
+    // We want to make sure that we are always using the code below only when
+    // we have update called on chapters and not anything else
+    if (this.kind_ === 'chapters') {
```
gkatsev: this will always be true.

src/js/component.js
```diff
@@ -105,7 +105,7 @@ class Component {
    */
   dispose() {
     this.trigger({ type: 'dispose', bubbles: false });
-
+
```
gkatsev: whitespace

src/js/menu/menu-button.js
```diff
@@ -36,12 +36,13 @@ class MenuButton extends ClickableComponent {
    * @method update
    */
   update() {
+
```
gkatsev: extra whitespace

src/js/menu/menu-button.js
```diff
@@ -36,12 +36,13 @@ class MenuButton extends ClickableComponent {
...
    */
   update() {
+    
     let menu = this.createMenu();
 
     if (this.menu) {
       this.removeChild(this.menu);
     }
-
+
```
gkatsev: here as well

src/js/utils/dom.js
```diff
@@ -171,6 +171,7 @@ const elIdAttr = 'vdata' + (new Date()).getTime();
  * @function getElData
  */
 export function getElData(el) {
+
```
gkatsev: extra whitespace.

gkatsev:
LGTM

misteroneill:
LGTM

PR #3356, opened 2016-06-02
Improve Logging for IE < 11

misteroneill:
## Description
IE11 introduced expandable objects in the console. Previously, everything was `.join()`ed and, therefore, objects came through as `[object Object]`, which is not useful. This improves the ES6-ification of the log module and passes objects and arrays through `JSON.stringify()` before logging them in IE10 and lower.

Some opportunities for ES6 brevity were left out because JSHint does not like them; though, they would pass `videojs-standard`.

## Specific Changes proposed
- Modifications to the existing `utils/log` module.

## Requirements Checklist


gkatsev:
Can you add a test for the JSON.stringify functionality?

misteroneill:
You're right. 😄  I added a test for that behavior. Unfortunately, it meant exposing the `logByType` function and giving it an optional argument only used for testing, which broke `TextTrack` tests (which have been fixed).

Rebased on `master` too.

PR #3370, opened 2016-06-15
add `loadstart` event jsdoc

mboles:
Added loadstart event so it appears in documentation


gkatsev:
I was worried that having `handleTechLoadStart_` appear twice could cause issues but seems to be fine.
LGTM.

PR #3378, opened 2016-06-17
Fix NO_SRC when play is hit before the video elem src is set

ldayananda:
## Description
Previously, if play is triggered before the video element has a source, Video.js would throw PLAYER_ERR_NO_SRC and this message would flash in the player briefly.

## Specific Changes proposed
This checks whether the player has a source set on play, and if not, adds a loadstart handler to the tech to ensure that the source is loaded before playing.

## Requirements Checklist
- [x]  Bug fixed

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1293,7 +1293,7 @@ class Player extends Component {
    */
   play() {
     // Only calls the tech's play if we already have a src loaded
-    if (this.player_.currentSrc()) {
+    if (this.src() || this.currentSrc()) {
       this.techCall_('play');
     } else {
       this.tech_.one('loadstart', function() {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1292,7 +1292,13 @@ class Player extends Component {
    * @method play
    */
   play() {
-    this.techCall_('play');
+    /* Autoplay solution */
+    if (this.player_.currentSrc()) {
+      this.techCall_('play');
+    } else {
+      this.options_.autoplay = true;
+    }
+
     return this;
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1292,11 +1292,11 @@ class Player extends Component {
    * @method play
    */
   play() {
-    /* Autoplay solution */
+    // Only calls the tech's play if we already have a src loaded
     if (this.player_.currentSrc()) {
       this.techCall_('play');
     } else {
-      this.options_.autoplay = true;
+      this.autoplay(true);
     }
 
     return this;
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1296,10 +1296,14 @@ class Player extends Component {
     if (this.player_.currentSrc()) {
       this.techCall_('play');
     } else {
-      this.autoplay(true);
+      this.delayedPlay();
     }
 
     return this;
+  }
+
+  delayedPlay() {
+    this.techCall_('delayedPlay');
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/js/player.js
```diff
@@ -1292,7 +1292,13 @@ class Player extends Component {
    * @method play
    */
   play() {
-    this.techCall_('play');
+    // Only calls the tech's play if we already have a src loaded
+    if (this.player_.currentSrc()) {
+      this.techCall_('play');
+    } else {
+      this.autoplay(true);
```
incompl: What happens if the source is changed afterwards? Does that video autoplay too since we changed the setting?

test/unit/player.test.js
```diff
@@ -247,15 +247,15 @@ test('should hide the poster when play is called', function() {
   });
 
   equal(player.hasStarted(), false, 'the show poster flag is true before play');
-  player.play();
+  player.tech_.trigger('play');
```
incompl: Could you also fix this by mocking currentSrc()?

src/js/player.js
gkatsev: Another potential issue, on firefox when you hit play without a source, `autplay` is still set to `false`: https://jsfiddle.net/gkatsev/xnm01byt/

test/unit/player.test.js
ldayananda: I was originally getting an error about the method currentSrc() not existing(likely since the tests use a fake tech.) I looked at how some of the other tests were written and saw that they almost exclusively trigger play in this way so I went with this instead.

incompl: If it's consistent with the other tests that seems reasonable

src/js/player.js
gkatsev: And I just realized that if we want it to autoplay after a source is given, we need to make sure that the `play` event is fired when this is called but it isn't going to right now, I don't think...

ldayananda: With respect to the play event firing when autoplay is set, player.src(...) actually handles updating the video source and fires the play event if autoplay is true. In my general testing of this on different browsers, this seems to work.

gkatsev: Sorry, what I meant is that `play` should be fired when this function is called and not when the src is added and it starts playing.

src/js/tech/tech.js
```diff
@@ -308,6 +308,12 @@ class Tech extends Component {
     return createTimeRange();
   }
 
+  delayedPlay() {
+    this.el_.onloadstart = function() {
```
gkatsev: you should be able to do `this.tech_.one('loadstart'`

ldayananda:
Using .one at the player.js level as suggested by @gkatsev.

src/js/player.js
```diff
@@ -1292,7 +1292,15 @@ class Player extends Component {
    * @method play
    */
   play() {
-    this.techCall_('play');
+    // Only calls the tech's play if we already have a src loaded
+    if (this.player_.currentSrc()) {
```
gkatsev: You should be able to just do `this.currentSrc()` here, since the context is already the player itself. We also may want to do `this.src() || this.currentSrc()` to make sure that we catch issues where the `src` is set but `currentSrc` isn't (can happen when adblock is involved.

ldayananda:
Addressing potential holes when adblockers are used as per @gkatsev 's comment.

gkatsev:
LGTM

misteroneill:
LGTM

Issue #3013, opened 2016-01-20
video-js.css corrupt


arcontact:
I found several errors in http://vjs.zencdn.net/5.4.6/video-js.css:
`content: "";`
`@media �screen`
Maybe its related to character encoding?
Can you provide me minified CSS with UTF-8 encoding?

gkatsev:
None of those are errors and that file is fully utf-8 compliant. The issue is that sass converts the unicode codepoints into utf8 characters but when you view the css directly, it looks like a square because you don't have the font loaded in. We do have a fix in a prerelease version of videojs that will go around sass and have the final output of the css have the codepoints in them directly.
As for the `@media` clause, apparently, it's a trick to keep IE8 inline: https://github.com/videojs/video.js/blob/master/src/css/components/_control-bar.scss#L45-L52
Thanks, hope that helps clear things up.

arcontact:
Thank you very much i really appreciate your work.

thomaspeeters:
`@media �screen` is causing unexpected behaviour in IE11 on Windows 7 for me. It is supposed to be `@media \0screen`, [which should target only IE8](http://keithclark.co.uk/articles/moving-ie-specific-css-into-media-blocks/), but it appears to be broken.

Currently, the rule is applied in IE8 and higher, instead of just IE8. Changing the `�` to `\0` in the CSS fixed the issue for me.

The expected behaviour is for the @media rule to not be applied in IE11, but the actual behaviour is that it is applied.
Perhaps it would be better to separate IE8-specific CSS in the same way IE8-specific JS is in a separate file?

For clarity's sake I should add where I encountered this bug. I am overriding the default CSS so the controls aren't hidden during playback when the player is inactive.

These are my overriding rules:
```CSS
.video-js .vjs-control-bar {
    display: flex;
}

.vjs-has-started.vjs-user-inactive.vjs-playing .vjs-control-bar {
    visibility: visible;
    opacity: 1;
}
```

In Chrome and Firefox this results in a controlbar which is always visible, in IE11 this results in all the button icons disappearing when the player enters the `inactive` state.
Adding the following overriding rule **did not** work, for a reason unknown to me.

```CSS
@media screen {
    .vjs-user-inactive.vjs-playing .vjs-control-bar :before {
        content: inherit; }
}
```

thomaspeeters:
I just updated to the latest version (5.7.1) and noticed this issue is still present. Please re-open this issue, The IE8 media query is still broken.

[...]

PR #3380, opened 2016-06-18
rework build to seperate out ie8 specific sass

erikyuzwa:
## Description
* This is in reference to issue #3013 
* Slice out some identified lines of sass specific to `IE8` and create its own separate `.css` file if needed by developers

## Specific Changes proposed
* added `src/css/ie8.scss` to capture all ie8 specific sass.
* moved out some lines from `src/css/components/_control-bar.scss` which were ie8 specific.
* added some necessary modifications to the `grunt build` task to support this.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -199,13 +199,21 @@ module.exports = function(grunt) {
         src: ['video-js.css', 'alt/video-js-cdn.css'],
         dest: 'build/temp/',
         ext: '.min.css'
+      },
+      ie8: {
+        expand: true,
+        cwd: 'build/temp/ie8/',
+        src: ['videojs-ie8.css'],
+        dest: 'build/temp/ie8/',
+        ext: '.min.css'
       }
     },
     sass: {
       build: {
         files: {
           'build/temp/video-js.css': 'src/css/vjs.scss',
-          'build/temp/alt/video-js-cdn.css': 'src/css/vjs-cdn.scss'
+          'build/temp/alt/video-js-cdn.css': 'src/css/vjs-cdn.scss',
+          'build/temp/ie8/videojs-ie8.css': 'src/css/ie8.scss'
         }
       }
     },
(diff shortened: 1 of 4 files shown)


gkatsev:
Moving the IE8 stuff to another sass file, unfortunately, won't fix the issue. It needs to be turned into a plain old css file (so, sass won't turn the special characters into utf8/unicode characters).
Also, we don't want to create any breaking changes in patch or minor releases. Would be nice if this extra file can just be appended together with the file generated from sass.

gkatsev:
Also, wanted to thank you for getting started on this. Let me know if you can get to the changes above, otherwise, I can probably pull your changes in and make the other changes as well.

erikyuzwa:
gotcha. Thanks for the extra clarifications @gkatsev - I should be able to take care of this. No worries at all.

erikyuzwa:
the only thing outstanding is that the unminified `src/css/ie8.css` isn't being appended to `build/temp/video-js.css`

gkatsev:
Having the temp file not include the ie8 piece is fine, I think.

gkatsev:
While I think that `build/temp/video-js.css` is fine if it's missing, it does need to be in `dist/video-js.css`. Looking into what needs to change. Maybe we can still get it into `build/temp/video-js.css` anyway.

gkatsev:
Ah, we can use `concat` to do this. I'll make a PR for this against your PR.

gkatsev:
Made a PR for this, see https://github.com/erikyuzwa/video.js/pull/1
What it does is uses concat to add the ie8-only files into `build/temp/video-js.css` after the `sass` task compiles all the sass stuff into `build/temp/video-js.css`. Then we don't need the `cssmin` changes because the ie8 changes are already in the file.

misteroneill:
Love this. LGTM.

gkatsev:
landed in https://github.com/videojs/video.js/commit/d86d4b222203d460557697e6d7a411048d5e2d92 and https://github.com/videojs/video.js/commit/1ff9f381a23f571e427c713ea459f1436037d87a.
Thanks @erikyuzwa for starting this!

PR #3401, opened 2016-06-27
use pinned grunt-sass

gkatsev:
## Description
We are using functionality (bug) that is removed in a newer versions of node-sass, so, we want to pin to an earlier version until we figure things out.


dmlap:
LGTM

gkatsev:
Merged to stable. Will be closed when released.

Issue #3305, opened 2016-05-09
set 'nativeControlsForTouch'  and 'usingNativeControls' dosen't effect in mobile(ios,android)


yqb1989:
i saw this in the [source code](https://github.com/videojs/video.js/blob/master/src/js/tech/html5.js#L116)
```
    // Determine if native controls should be used
    // Our goal should be to get the custom controls on mobile solid everywhere
    // so we can remove this all together. Right now this will block custom
    // controls on touch enabled laptops like the Chrome Pixel
    if (browser.TOUCH_ENABLED && options.nativeControlsForTouch === true ||
        browser.IS_IPHONE ||
        browser.IS_NATIVE_ANDROID) {
      this.setControls(true);
    }
```
it seems that the **nativeControlsForTouch** will never effect if set to **false** in ios or android

Dracoborg:
Same issue here.

Phone: Nexus 5
OS: Android 6.0.1
Browsers:
    Chrome 50.0.2661.89
    Firefox 46.0.1

Also checked on Galaxy S5 but I don't know details.

gkatsev:
#3410 has been merged and fixed in master. Release soon.

PR #3410, opened 2016-07-01
usingNativeControls effect on mobile #3305

sashyro:
#3305


sashyro:
#3305

gkatsev:
Seems to work. Thanks.
One weird thing is that iphone will still keep it's big play button in the center, though, our controls work with it as well.

g-gabber:
@gkatsev 
I faced the same problem as you, that on iphone there is a big play button.
I managed to fix it using css from this post [hide-iphone-html5-video-play-button](http://stackoverflow.com/questions/8911247/hide-iphone-html5-video-play-button) on stackoverflow.

Issue #1316, opened 2014-06-26
Have videojs choose another source/tech combination when one fails


ange007:
Any news concerning https://github.com/videojs/video.js/issues/995?
As a whole some things changed, but not absolutely: http://jsfiddle.net/uLG8H/9/

Still strangeness in this example:
*if to click on the first link, and then on the second: the first video won't work, but the second will play.
*if to press at once the second video - it won't play.

_P.S. And why not to make an "techOrder" rechoice in case of failure previous?_

gkatsev:
I believe the output we want from this issue is the ability for videojs to choose another source/tech combination if the current one failed.
I'm going to update the title with this.

gkatsev:
I think tech selection is in a much better place in the newest versions of Video.js. Having Video.js choose another source when one failed is definitely a thing we want to work on. I'm working on writing up a roadmap for Video.js and retry on error is definitely going to be on there.
In the mean time, I'm closing this issue.

PR #3414, opened 2016-07-03
Allow the control bar to show on the error screen

### LINKED ISSUE #3323: Allow us to change sources even if we hit a media that is not supported in the currentbrowser
## Description
Briefly describe the issue.
Include a [reduced test case](https://css-tricks.com/reduced-test-cases/), we have a [starter template](http://jsbin.com/axedog/edit?html,output) on JSBin you can use.

When you select a media for example webm in Internet Explorer it causes it to show an error that says this media is not supported or something similar to that but you carn't do anything else, you carn't change source or do anything else. With ogv.js it supports the ogv format in Internet Explorer and Microsoft Edge with js.

The only way to get the player to do anything is to refresh the page and I think it selects the media it supports. But again if you select webm from source switcher it causes the error and you carn't do anything else until you refresh again. Please could you support changing sources even if the media your on is not supported. Would solve having to refresh everytime.

## Additional Information
#### videojs
5.10.1
#### browsers
what browser are affected?
All that doint support the format you click on
#### OSes
### plugins
https://github.com/kmoskwiak/videojs-resolution-switcher

paladox:
## Description

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

Allows us to change src to something that will work without needing to refresh the page.

Fixes #3323

### DIFF
--- a/src/css/components/_control-bar.scss
+++ b/src/css/components/_control-bar.scss
@@ -41,8 +41,9 @@ $ie8screen: "\\0screen";
 .vjs-controls-disabled .vjs-control-bar,
 .vjs-using-native-controls .vjs-control-bar,
 .vjs-error .vjs-control-bar {
-  // !important is ok in this context.
-  display: none !important;
+  // Allow you to change src even if
+  // your on the error screen.
+  z-index: 100;
 }
 
 // Don't hide the control bar if it's audio


paladox:
@gkatsev Hi, I managed to get control bar to show, but the error message and the big x doint disappear when I choose a different source that works.

gkatsev:
#3323 doesn't need a fix without the bigger issue at hand that needs fixing: having videojs try and choose a different source when an error occurs https://github.com/videojs/video.js/issues/1316.
You should be able to change the source just fine by calling `error(null)` like I mentioned in my comment on #3323: https://jsbin.com/cugeta/4/edit?html,js,console,output

paladox:
But if the control bar is hidden, we carn't go into it to change the source. Unless that's what  error(null)  does.

gkatsev:
Yes, `player.error(null)` tells videojs to ignore the error and hide the error display.

paladox:
Ok thanks, if we update the resolution plugin, would that work.

gkatsev:
Yeah, I think if the issue is that you can't switch resolutions, it should be fixed in that plugin. You can listen to the `error` event to make sure that you're only disabling the correct error.

paladox:
Ok thanks.

PR #3416, opened 2016-07-05
all components are case-sensitive

si458:
## Description
the components do not work correctly unless every first charactor is lowercase and the rest uppercase

## Specific Changes proposed
change first character to lowercase and rest uppercase

## Requirements Checklist
  - [ ] Docs/guides updated

all components are case-sensitive so first letter must be lowercase!

### DIFF
--- a/docs/guides/components.md
+++ b/docs/guides/components.md
@@ -12,13 +12,13 @@ The UI component architecture makes it easier to add child components to a paren
 
 ```js
 // Adding a new control to the player
-myPlayer.addChild('BigPlayButton');
+myPlayer.addChild('bigPlayButton');
 ```
 
 Every component has an associated DOM element, and when you add a child component, it inserts the element of that child into the element of the parent.
 
 ```js
-myPlayer.addChild('BigPlayButton');
+myPlayer.addChild('bigPlayButton');
 ```
 
 Results in:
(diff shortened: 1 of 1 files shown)


docs/guides/components.md
```diff
@@ -35,30 +35,30 @@ The actual default component structure of the Video.js player looks something li
 
 ```
 Player
```
gkatsev: The idea for this section to go with the "class names" rather than the instance names.

gkatsev:
I wonder if we should make a note that when you do `addChild` or `getChild` it should be camel-cased instead.
Also, whether those method should just be case-insensitive?
@videojs/core-committers should `addChild` and related methods be case-insensitive for component names?

misteroneill:
I think they should be case-insensitive in the sense that they accept lowerCamelCase and UpperCamelCase because the latter matches the class names, but not UPPERCASE or lowercase.

si458:
i think it be case-insensitive too :)

gkatsev:
I created a new issue describing this and what we want to happen: https://github.com/videojs/video.js/issues/3436
So, going to close this PR. Thanks for your time @si458, if you feel up to it, would love it if you make a PR to fix #3436.

PR #3426, opened 2016-07-11
Respect lang attribute on player and inherit correctly

mister-ben:
## Description
Currently, if language isn't set in the player options language will be determined by the `lang` attribute of the `html` element if set before falling back to the browser language preference.

`lang` should be respected if present on the video element. If not present, lang should be [inherited](https://www.w3.org/TR/html4/struct/dirlang.html#h-8.1.2) from the closest element with it set up to and including html.

## Specific Changes proposed
Checks for closest element with a `lang` attribute, not just `html`. The check is made in constructor rather than options prototype, as it needs to be relative to the specific player instance. A language in the player options still takes precedence.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -92,6 +92,25 @@ class Player extends Component {
     // see enableTouchActivity in Component
     options.reportTouchActivity = false;
 
+    // If language is not set, get the closest lang attribute
+    if (!options.language) {
+      if (typeof tag.closest === 'function') {
+        let closest = tag.closest('[lang]');
+        if (closest) {
+          options.language = closest.getAttribute('lang');
+        }
+      } else {
+        let element = tag;
+        while (element && element.nodeType === 1) {
+          if (Dom.getElAttributes(tag).hasOwnProperty('lang')) {
+            options.language = element.getAttribute('lang');
+            break;
+          }
+          element = element.parentNode;
+        }
+      }
+    }
+
     // Run base component initializing with new options
     super(null, options, ready);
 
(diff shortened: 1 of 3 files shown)


gkatsev:
LGTM.

gkatsev:
@mister-ben hey, looks like some of the tests are failing in IE, would you be able to take a look? See https://travis-ci.org/videojs/video.js/builds/145628625 (ignore the flash duration tests)

mister-ben:
Ack. Will check next week when I'm back from holiday.

gkatsev:
@mister-ben cool, I might end up fixing it myself. Would like to get a release out this week, if possible. I'll make sure to write here whether I managed to fix it or not.

gkatsev:
Figured it out. A pretty silly issue that's super hard to find: https://github.com/videojs/video.js/pull/3455

src/js/player.js
```diff
@@ -92,6 +92,25 @@ class Player extends Component {
...
+    // If language is not set, get the closest lang attribute
+    if (!options.language) {
+      if (typeof tag.closest === 'function') {
+        let closest = tag.closest('[lang]');
+        if (closest) {
+          options.language = closest.getAttribute('lang');
+        }
+      } else {
+        let element = tag;
+        while (element && element.nodeType === 1) {
```
OwenEdwards: Suggest you use `... && element.nodeType === Node.ELEMENT_NODE` here for clarity (see https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType).

mister-ben:
@gkatsev sorry about that. Thanks for sorting it out.

Issue #2931, opened 2015-12-18
Add an option to override format-time util


adamtal3:
There are many places in the code calling _utilsFormatTimeJs2['default']. Why is this util private? In v4 it was possible to override the time display format. In order to change it now. I have to override the prototype functions of 6 different components instead of just overriding format-time.

This is the override I've came up with:

``` javascript
    videojs.getComponent('PlayProgressBar').prototype.updateDataAttr = function (seconds) {
        var time = this.player_.scrubbing() ? this.player_.getCache().currentTime : this.player_.currentTime();
        this.el_.setAttribute('data-current-time', myCustomTimeFormat(time));
    }
    videojs.getComponent('CurrentTimeDisplay').prototype.updateContent = function updateContent() {
        // Allows for smooth scrubbing, when player can't keep up.
        var time = this.player_.scrubbing() ? this.player_.getCache().currentTime : this.player_.currentTime();
        var localizedText = this.localize('Current Time');
        var formattedTime = myCustomTimeFormat(time);
        this.contentEl_.innerHTML = '<span class="vjs-control-text">' + localizedText + '</span> ' + formattedTime;
    };
    videojs.getComponent('DurationDisplay').prototype.updateContent = function updateContent() {
        var duration = this.player_.duration();
        if (duration) {
            var localizedText = this.localize('Duration Time');
            var formattedTime = myCustomTimeFormat(duration);
            this.contentEl_.innerHTML = '<span class="vjs-control-text">' + localizedText + '</span> ' + formattedTime; // label the duration time for screen reader users
        }
    };
    videojs.getComponent('RemainingTimeDisplay').prototype.updateContent = function updateContent() {
        if (this.player_.duration()) {
            var localizedText = this.localize('Remaining Time');
            var formattedTime = myCustomTimeFormat(this.player_.remainingTime());
            this.contentEl_.innerHTML = '<span class="vjs-control-text">' + localizedText + '</span> -' + formattedTime;
        }
    };
    videojs.getComponent('MouseTimeDisplay').prototype.update = function update(newTime, position) {
        var time = myCustomTimeFormat(newTime);
        this.el().style.left = position + 'px';
        this.el().setAttribute('data-current-time', time);
    };
    videojs.getComponent('SeekBar').prototype.updateARIAAttributes = function updateARIAAttributes() {
        // Allows for smooth scrubbing, when player can't keep up.
        var time = this.player_.scrubbing() ? this.player_.getCache().currentTime : this.player_.currentTime();
[...]

PR #3433, opened 2016-07-16
Allow the user to override videojs.formatTime using a custom function #2931

adamtal3:
## Description

https://github.com/videojs/video.js/issues/2931
## Specific Changes proposed

Updated format-time.js to check if videojs.formatTime is a function and if it is - use it.

### DIFF
--- a/src/js/utils/format-time.js
+++ b/src/js/utils/format-time.js
@@ -4,7 +4,11 @@
  * Format seconds as a time string, H:MM:SS or M:SS
  * Supplying a guide (in seconds) will force a number of leading zeros
  * to cover the length of the guide
- *
+ */
+
+import window from 'global/window';
+
+/**
  * @param  {Number} seconds Number of seconds to be turned into a string
  * @param  {Number} guide   Number (in seconds) to model the string after
  * @return {String}         Time formatted as H:MM:SS or M:SS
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks for the PR. We appreciate the time and effort spent on it.

As it is right now, I don't think we can merge it because it assumes that videojs gets included as a global and doesn't allow a good mechanism for setting a custom format time method when videojs is required using browserify, webpack, or another bundler. Any solution to this problem must work regardless of how videojs is included and preferably not rely on separate global variables that videojs reads in.

Maybe what we want is to have a `setFormatTime` method that just takes a function and `formatTime` ends up using that method? As you mentioned in #2931, writing a custom string formatter isn't a great idea since users could end up with who knows what requirements.

I'm not sure it's necessary to keep this backwards compatible since it's a new feature; the fact that it worked previously is coincidental and not officially supported. Having a setter function would make it easier for us to change things internally without having to break the outward facing API. I'm not married to `setFormatTime`; let me know what you think about it.

adamtal3:
I see your point.

Adding a setFormatTime function that will receive a formatting function will be a better approach (This is just the quickest).

How should I address this? Should I add a videojs.setFormatTime = .... when I import the ... from the time-format.js which will save the custom formatting function in a local/global variable in the file's scope?

Any specific instructions?

gkatsev:
Probably the easiest is to have a `setFormatTime` method in both `video.js` and `format-time.js`  with the former calling the latter.
I'll think about it some more to see if there should be other constraints. The one thing I can think off is that the function being passed in only needs to get the time in seconds, since the `guide` would be a bit meaningless to it.

thijstriemstra:
I'd also like to use this eventually, my plugins use a custom formatTime to add milliseconds, currently unsupported by video.js.

gkatsev:
Closing this PR for now, the issue associated will remain open.

cedricve:
@adamtal3 do you have any progress on this?

Issue #2901, opened 2015-12-08
Capitals in tech name cause tech options to go missing


hartman:
Might be true for other component types as well, but at least for Tech.

If your Tech is named 'OgvJs' or 'HtmL5', than per the camelCase convention of options, you would expect that you can pass Tech options to your Tech like:

```
{
  ogvJs: {},
  htmL5: {}
}
```
or
```
{
  ogvjs: {},
  html5: {}
}
```
This however doesn't work. The options extra options are not present. If you inspect your Tech inside createEl, then you will see that the instance of your tech there shows this.name_ == null

Instead you have to rename your class to Html5 or Ogvjs, for your options to be copied into the options of the Tech constructor. This is confusing and unexpected.

PR #3439, opened 2016-07-19
Components are now accessible via `camelCase` and `UpperCamelCase` me…

### LINKED ISSUE #3436: `addChild` should not be sensitive to the first letter casing
When adding a component using `addChild` or getting a component using `getChild`, videojs expects the component to be in a `camelCased` style. However, the class names are `UpperCamelCased` which can be a bit confusing to users.
Instead, these methods (and any other related method) should accept the component name as both `camelCase` and `UpperCamelCase` but not `lowercase` or `UPPERCASE`.

This came up as a result of #3416.

chemoish:
Fixes #3436.
## Description

Components are not uniformly accessible through the same names. 

Example:
- `getChild` cannot retrieve a component by a `camelCase` name
- `addChild` can retrieve, then add a component by a `camelCase` name
## Specific Changes proposed

Enable components to be accessed via both `camelCase` and `UpperCamelCase` means.
## Requirements Checklist
  - [ ] Docs/guides updated

…ans.
- addChild
- getChild
- getComponent
- registerComponent

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -362,8 +362,6 @@ class Component {
         options = {};
       }
 
-      // If no componentClass in options, assume componentClass is the name lowercased
-      // (e.g. playButton)
       const componentClassName = options.componentClass || componentName;
 
       // Set name through options
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -347,7 +347,7 @@ class Component {
     let component;
     let componentName;
 
-    // If child is a string, create nt with options
+    // If child is a string, create component with options
     if (typeof child === 'string') {
       componentName = toTitleCase(child);
 
@@ -364,7 +364,7 @@ class Component {
 
       // If no componentClass in options, assume componentClass is the name lowercased
       // (e.g. playButton)
-      const componentClassName = options.componentClass || toTitleCase(componentName);
+      const componentClassName = options.componentClass || componentName;
 
       // Set name through options
       options.name = componentName;
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -306,6 +306,12 @@ class Component {
    * @method getChild
    */
   getChild(name) {
+    if (!name) {
+      return;
+    }
+
+    name = toTitleCase(name);
+
     return this.childNameIndex_[name];
   }
 
@@ -343,7 +349,7 @@ class Component {
 
     // If child is a string, create nt with options
     if (typeof child === 'string') {
-      componentName = child;
+      componentName = toTitleCase(child);
 
       // Options can also be specified as a boolean, so convert to an empty object if false.
       if (!options) {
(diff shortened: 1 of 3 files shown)


src/js/component.js
```diff
@@ -361,7 +361,7 @@ class Component {
 
       // If no componentClass in options, assume componentClass is the name lowercased
```
chemoish: comment might need to be updated.

src/js/component.js
```diff
@@ -1398,14 +1399,26 @@ class Component {
    * @method getComponent
    */
   static getComponent(name) {
-    if (Component.components_ && Component.components_[name]) {
-      return Component.components_[name];
+    if (!name) {
```
chemoish: i think something was trying to retrieve a component by name `''`, but i couldn't track it in the call stack.

might need refactoring, but i got out early and updated the implementation of `toTitleCase`.

src/js/utils/to-title-case.js
```diff
@@ -8,7 +8,11 @@
  * @private
  * @method toTitleCase
  */
-function toTitleCase(string){
+function toTitleCase(string) {
+  if (typeof string !== 'string') {
```
chemoish: might need to be changed, but will error out if a string isn't passed.

gkatsev: seems reasonable to throw here, maybe.

src/js/component.js
misteroneill: In the old code and the new code, a missing `Component.components_` or a missing `Component.components_[name]` will return `undefined` - the only difference is the old code was simpler. It seems like this block should go back to its old state, but with a `name = toTitleCase(name);` line before it:

```js
name = toTitleCase(name);
}
```

chemoish: will update, i don't know the state of the formatting rules, will ignore http://eslint.org/docs/rules/no-param-reassign.

gkatsev:
Looks like this needs to be rebased against master

chemoish:
@gkatsev I temporarily removed `vjsstandard` and rebased master.

chemoish:
@gkatsev updated again.

src/js/component.js
gkatsev: I think this comment could just be removed?

chemoish:
@gkatsev, Is that a known issue? I wasn't having that issue with FF 48, but I am having it with 49.0.1, does that have to do with me?

(removing the extraneous comment)

gkatsev:
@chemoish yup, I have a PR to fix the issue that should be merged soon: https://github.com/videojs/video.js/pull/3640

hartman:
@gkatsev does it per chance also effect #2901 ?

gkatsev:
@hartman unfortunately, it doesn't.

PR #3444, opened 2016-07-19
Separate fn.apply and stringify checks

### LINKED ISSUE #3442: `logByType` issue in IE10
## Description
I've only noticed this in IE10 (works fine in IE9 and IE11). In the function `logByType` (in `log.js` src file), the following error is thrown:

```
SCRIPT65535: Invalid calling object
```

If you simply surround the `fn(args.map....` call in a try/catch, it solves the issue. However, I don't know if this is the best solution.

## Steps to reproduce

1. Simply run the code in IE10. Although, I was also using the VAST/VPAID plugin.

### Expected
No exception to be thrown and VideoJS initialize normally.

### Actual
Exception thrown

### Error output
```
SCRIPT65535: Invalid calling object
```

#### videojs
Current `master` branch
#### browsers
IE10
#### OSes
Windows *
### plugins
VAST/VPAID plugin

misteroneill:
## Description
This fixes #3442 which was introduced by #3356.

IE10 correctly supports `console.*.apply`, but should also get the "stringification" treatment (because it uselessly logs `[object Object]` and similar). This avoids an issue where an "Invalid calling object" was triggered when trying to call console method(s) from a cached reference (without a proper context object).

## Specific Changes proposed
- Split out the logic around whether `console.*.apply` is supported and whether or not arguments should be stringified.

## Requirements Checklist
  - [x] ~~Unit Tests updated or fixed~~ N/A
  - [x] ~~Docs/guides updated~~ N/A
  - [x] ~~Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))~~

### DIFF
--- a/src/js/utils/log.js
+++ b/src/js/utils/log.js
@@ -38,11 +38,10 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   // add console prefix after adding to history
   args.unshift('VIDEOJS:');
 
-  // Old IE versions do not allow .apply() for some/all console method(s). And
   // IEs previous to 11 log objects uselessly as "[object Object]"; so, JSONify
   // objects and arrays for those less-capable browsers.
-  if (!fn.apply || stringify) {
-    fn(args.map(a => {
+  if (stringify) {
+    args = args.map(a => {
       if (a && typeof a === 'object' || Array.isArray(a)) {
         try {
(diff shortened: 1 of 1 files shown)


misteroneill:
Also: I tested (again) this change in all supported IE versions and it worked as expected.

gkatsev:
LGTM

PR #3449, opened 2016-07-21
Prevent coveralls from running locally

misteroneill:
## Description
The bash condition that previously existed in `package.json` to check for the `TRAVIS` env var was not working properly for everyone, causing `grunt coveralls` to run (and fail) locally.

Additionally, it would fail on Windows, since it was bash.

## Specific Changes proposed
This moves that condition checking for the `TRAVIS` env var to JavaScript, where it now works reliably.

## Requirements Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "homepage": "http://videojs.com",
   "author": "Steve Heffernan",
   "scripts": {
-    "test": "grunt test && if [ '$TRAVIS' ]; then grunt coveralls; fi;"
+    "test": "grunt test"
   },
   "repository": {
     "type": "git",
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -494,7 +494,8 @@ module.exports = function(grunt) {
   // Default task - build and test
   grunt.registerTask('default', ['test']);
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
   "homepage": "http://videojs.com",
   "author": "Steve Heffernan",
   "scripts": {
-    "test": "grunt test && if [ '$TRAVIS' ]; then grunt coveralls; fi;"
```
BrandonOCasey: you could also just double quote and escape travis here like `grunt test && if [ \"$TRAVIS\" ]; then grunt coveralls; fi;"` but if were going for windows build compatibility then grunt is he better way. LGTM other than that

gkatsev: Yeah, we may as well and try to make it work on windows.

gkatsev:
LGTM

package.json
misteroneill: The fact that using `"` instead of `'` would fix it suggests to me that sticking to JS instead of Bash is the way to go regardless of Windows support. 😄

BrandonOCasey: yeah bash is a fickle beast

PR #3450, opened 2016-07-21
Correct checks for IEs older than 11 in log tests

misteroneill:
## Description
This corrects test assertions for IE versions older than 11 in the tests for the `utils/log` module.

Because those browsers have their arguments `join`ed by default, they were causing some tests to fail when run in Browserstack against IE.

## Requirements Checklist


gkatsev:
LGTM

PR #3454, opened 2016-07-22
fix logging test errors

gkatsev:
This supercedes #3450.


gkatsev:
The tests that are failing now are because of Flash and lang. LGTM.

PR #3459, opened 2016-07-25
videojs-standard

misteroneill:
## Description
As discussed in-person on Thursday, we are in a slight lull at the moment and this is a great opportunity to clean up the video.js source code (and tests) to pass videojs-standard linting and abandon jshint.

## Specific Changes proposed
- Removes jshint
- Installs videojs-standard at 5.x
- Adds linting scripts/tasks
- Makes small code changes to address videojs-standard checks against `src/js/*.js`

## Requirements Checklist


build/grunt.js
```diff
@@ -114,14 +114,6 @@ module.exports = function(grunt) {
       build: ['build/temp/*'],
       dist: ['dist/*']
     },
-    jshint: {
```
misteroneill: This is a temporary step. The final PR in this series will restore automatic linting capabilities, but for now we want them off.

src/js/button.js
```diff
@@ -3,10 +3,7 @@
  */
 import ClickableComponent from './clickable-component.js';
 import Component from './component';
-import * as Events from './utils/events.js';
```
misteroneill: There are removals like this scattered around: they were unused modules.

_Note_: if we're only trying to import the module, we should be doing something like `import './utils/events'`.

src/js/clickable-component.js
```diff
@@ -121,14 +125,14 @@ class ClickableComponent extends Component {
    * @return {Component} The child component (created by this process if a string was used)
    * @method addChild
    */
-  addChild(child, options={}) {
-    // TODO: Fix adding an actionable child to a ClickableComponent; currently
+  addChild(child, options = {}) {
+    // TO_DO: Fix adding an actionable child to a ClickableComponent; currently
```
misteroneill: videojs-standard doesn't want you leaving `TODO` comments around (which makes sense because they almost never get done and only cause confusion down the line), but I didn't want to simply _delete_ them. So, I changed them to `TO_DO` which is kind of ugly, I know.

BrandonOCasey: I think the TODO rule is just a warning, should we just keep the warnings printing and work on fixing whatever it is we need to do?

misteroneill: You're right, I'll undo these changes!

src/js/base-styles.js
```diff
@@ -6,13 +6,17 @@
...
-if (window.VIDEOJS_NO_BASE_THEME) return;
+if (window.VIDEOJS_NO_BASE_THEME) {
+  return;
+}
 
 const styles = '{{GENERATED_STYLES}}';
 
-if (styles === '{{GENERATED'+'_STYLES}}');
+// Don't think we need this as it's a noop?
+// if (styles === '{{GENERATED'+'_STYLES}}');
```
gkatsev: I assume it was supposed to be something like
```js
if (styles === '{{GENERATED'+'_STYLES'}}) {
  return;
}
```
So that if somehow the styles weren't inline, it won't bother with the rest.

src/js/button.js
gkatsev: doesn't `import './utils/events.js'` put the exports into local scope? This is an issue for a "bag of functions" module.
`import * Events from './utils/events.js'` is more explicit in what we want.

src/js/player.js
```diff
@@ -26,19 +26,19 @@ import AudioTrackList from './tracks/audio-track-list.js';
...
 // Include required child components (importing also registers them)
-import MediaLoader from './tech/loader.js';
-import PosterImage from './poster-image.js';
-import TextTrackDisplay from './tracks/text-track-display.js';
-import LoadingSpinner from './loading-spinner.js';
-import BigPlayButton from './big-play-button.js';
-import ControlBar from './control-bar/control-bar.js';
-import ErrorDisplay from './error-display.js';
-import TextTrackSettings from './tracks/text-track-settings.js';
+import './tech/loader.js';
```
gkatsev: I would much prefer the earlier version. It's much more explicit and reduces cognitive load. I don't have to go to `./tracks/text-track-settings.js` to *know* that in here I can refer to it via `TextTrackSettings`.

src/js/player.js
```diff
@@ -516,14 +532,14 @@ class Player extends Component {
...
       // Otherwise calculate the height from the ratio and the width
-      height = width  * ratioMultiplier;
+      height = width * ratioMultiplier;
     }
 
     // Ensure the CSS class is valid by starting with an alpha character
-    if (/^[^a-zA-Z]/.test(this.id())) {
-      idClass = 'dimensions-'+this.id();
+    if ((/^[^a-zA-Z]/).test(this.id())) {
+      idClass = `dimensions-${this.id()}`;
```
gkatsev: did the linter complain about this? The reason we kept it the other way is because using the template strings here gives us no benefits over just string concatenation, and if anything, the old way is a bit cleaner.

src/js/player.js
```diff
@@ -2666,8 +2712,10 @@ class Player extends Component {
    */
   // destructure the input into an object with a track argument, defaulting to arguments[0]
   // default the whole argument to an empty object if nothing was passed in
-  removeRemoteTextTrack({track = arguments[0]} = {}) { // jshint ignore:line
-    this.tech_ && this.tech_['removeRemoteTextTrack'](track);
```
gkatsev: my one liner T__T

src/js/player.js
misteroneill: We'd need to turn off the linter for that line, then.

src/js/player.js
misteroneill: It complained about the infix operator not having spaces; so, I figured I'd just switch to the template string. Can switch back if desired.

src/js/base-styles.js
misteroneill: Hmm, thoughts on what to do, since it's clearly never done that?

src/js/player.js
```diff
@@ -2892,47 +2941,50 @@ Player.prototype.options_ = {
...
+// The following pointless assignment expressions are here only for
+// purposes of documentation tooling compatibility.
+
 /**
  * Fired when the user agent begins looking for media data
  *
  * @event loadstart
  */
-Player.prototype.handleTechLoadStart_;
+Player.prototype.handleTechLoadStart_ = Player.prototype.handleTechLoadStart_;
```
gkatsev: Does the linter complain about these? I think it would be nice to keep them as stubs, if possible.

src/js/video.js
```diff
@@ -721,12 +724,12 @@ videojs.insertContent = Dom.insertContent;
  * still support requirejs and browserify. This also needs to be closure
  * compiler compatible, so string keys are used.
  */
-if (typeof define === 'function' && define['amd']) {
-  define('videojs', [], function(){ return videojs; });
+if (typeof define === 'function' && define.amd) {
```
gkatsev: this probably should just be deleted altogether. But maybe in a separate PR.

src/js/button.js
misteroneill: I tend to agree, but the linter doesn't like unused variable names.

src/js/player.js
gkatsev: I think that in this particular case it would be a bit nicer.
```js
'dimensions-' + this.id();
`dimensions-${this.id()}`;
```

src/js/player.js
misteroneill: 🔥  burn it down 🔥

src/js/player.js
misteroneill: Yeah. It complained about not expecting an expression. We could alternatively do something like:

```js
Player.prototype.handleTechLoadStart_; // eslint-disable-line
```

src/js/player.js
gkatsev: What rule complains about this?

src/js/video.js
misteroneill: Yeah, I think so. Avoiding functional changes here is a good thing! 😄

src/js/button.js
gkatsev: if it isn't used then we shouldn't import it, sure.

src/js/base-styles.js
gkatsev: `base-styles` isn't even used anywhere anyway right now, so, feel free to just leave as is.

src/js/player.js
misteroneill: Fair enough! 👍

src/js/player.js
misteroneill: [`no-unused-vars`](http://eslint.org/docs/rules/no-unused-vars)

gkatsev: oooh, these are imported to make sure they're in the package. They're not used locally in player.js.
I wonder if there's a better way of handling this.
I guess this is fine then

misteroneill: Could also do `require('./tech/loader.js')`. I suppose it might be a little clearer that it's different from other imports?

src/js/base-styles.js
misteroneill: 👍

src/js/button.js
misteroneill: On the topic of using the `import 'foo';` style, I don't think it puts the exports into local scope. I think it simply runs the module code. MDN says:

> Import an entire module for side effects only, without importing any bindings.

src/js/player.js
gkatsev: We do have a comment above. Maybe just elaborate on it a bit?

src/js/button.js
misteroneill: Same as `require('foo.js');`

gkatsev: Ah, interesting. I guess that's exactly what we want in `player.js`.

gkatsev:
LGTM

src/js/media-error.js
```diff
@@ -8,13 +8,13 @@ import assign from 'object.assign';
  *
  * @param {Number} code The media error code
  */
-let MediaError = function(code){
+let MediaError = function(code) {
```
gkatsev: noting it here (but we shouldn't make this change as part of linting) but this potentially can cause a problem on IE8 since on line 22 we do `MediaError.defaultMessages` and IE8 is weird.

misteroneill: Good catch. Switching to `function MediaError(code) {` would solve that, right?

gkatsev: Yup.

misteroneill:
I merged all the disparate `vjsstandard-*` branches into this one.

misteroneill:
We are treating this as the "master" branch for standardization efforts.

gkatsev:
Landed!

PR #3460, opened 2016-07-25
src/js/utils passing videojs-standard

misteroneill:
## Description
The second linter PR. On top of #3459, this updates `src/js/utils`.

## Requirements Checklist

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -24,18 +24,21 @@ export const IS_IPHONE = (/iPhone/i).test(USER_AGENT) && !IS_IPAD;
 export const IS_IPOD = (/iPod/i).test(USER_AGENT);
 export const IS_IOS = IS_IPHONE || IS_IPAD || IS_IPOD;
 
-export const IOS_VERSION = (function(){
-  var match = USER_AGENT.match(/OS (\d+)_/i);
-  if (match && match[1]) { return match[1]; }
-})();
+export const IOS_VERSION = (function() {
+  let match = USER_AGENT.match(/OS (\d+)_/i);
+
+  if (match && match[1]) {
+    return match[1];
+  }
+}());
 
 export const IS_ANDROID = (/Android/i).test(USER_AGENT);
 export const ANDROID_VERSION = (function() {
   // This matches Android Major.Minor.Patch versions
   // ANDROID_VERSION is Major.Minor as a Number, if Minor isn't available, then only Major is returned
-  var match = USER_AGENT.match(/Android (\d+)(?:\.(\d+))?(?:\.(\d+))*/i),
-    major,
-    minor;
+  let match = USER_AGENT.match(/Android (\d+)(?:\.(\d+))?(?:\.(\d+))*/i);
+  let major;
+  let minor;
 
   if (!match) {
     return null;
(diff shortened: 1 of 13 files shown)


src/js/utils/browser.js
```diff
@@ -48,10 +51,10 @@ export const ANDROID_VERSION = (function() {
     return parseFloat(match[1] + '.' + match[2]);
   } else if (major) {
     return major;
-  } else {
-    return null;
   }
-})();
+  return null;
+}());
```
gkatsev: ugh, fine whatever.

misteroneill: Don't look at me!

gkatsev: I didn't add this rule!

src/js/utils/dom.js
```diff
@@ -493,17 +506,6 @@ export function getPointerPosition(el, event) {
 }
 
 /**
- * Determines, via duck typing, whether or not a value is a DOM element.
```
BrandonOCasey: what happened to isEl?

misteroneill: It moved up above because it was used before it was defined (which works because hoisting, but... it's against the rules).

src/js/utils/browser.js
misteroneill: It came all the way from the original fork of standard and _you_ specifically changed it to `"outside"`. 😉

src/js/utils/log.js
```diff
@@ -24,7 +26,7 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   // Was setting these once outside of this function, but containing them
   // in the function makes it easier to test cases where console doesn't exist
   // when the module is executed.
-  const fn = console && console[type] || function(){};
+  const fn = console && console[type] || function() {}; // eslint-disable-line
```
BrandonOCasey: cant we do `() =>` here? that may get rid of the need for eslint-disable-line too

src/js/utils/browser.js
misteroneill: Which is also the ESLint default.

src/js/utils/merge-options.js
```diff
@@ -41,7 +43,7 @@ const customizer = function(destination, source) {
  * provided objects
  * @function mergeOptions
  */
-export default function mergeOptions() {
+mergeOptions = function() {
```
BrandonOCasey: should probably make this `const mergeOptions` here and get rid of the let above

src/js/utils/browser.js
gkatsev: I plead the 5th.
(Actually, I don't remember setting that one... might be better to change it so we have less changes but I'm not attached to this.)

src/js/utils/log.js
misteroneill: The `eslint-disable-line` is there because of the reference to `console`. We _could_ use an arrow function, but I'm trying to avoid any changes that aren't necessary.

src/js/utils/merge-options.js
misteroneill: We can't because `customizer` uses `mergeOptions` and `mergeOptions` uses `customizer`. Perhaps a better solution is to remove the rule that says you can't use a function before it's defined (which is a silly rule unless you explicitly forbid function declarations, which we're no longer doing).

src/js/utils/log.js
gkatsev: What about `window.console` and `window.console[type]`?

misteroneill: It works, I'll push it up.

src/js/utils/browser.js
misteroneill: I'm fine with changing it.

src/js/utils/dom.js
BrandonOCasey: ahh I guess I missed that

PR #3461, opened 2016-07-26
videojs-standard Tracks

misteroneill:
## Description
Yet another PR in the Linter milestone! This time: `src/js/tracks`.

## Requirements Checklist

### DIFF
--- a/src/js/tracks/html-track-element-list.js
+++ b/src/js/tracks/html-track-element-list.js
@@ -7,7 +7,7 @@ import document from 'global/document';
 
 class HtmlTrackElementList {
   constructor(trackElements = []) {
-    let list = this;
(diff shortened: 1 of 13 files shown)


src/js/tracks/html-track-element-list.js
```diff
@@ -7,7 +7,7 @@ import document from 'global/document';
 
 class HtmlTrackElementList {
   constructor(trackElements = []) {
-    let list = this;
+    let list = this; // eslint-disable-line
```
misteroneill: There's a bunch of these. Normally, aliasing `this` is not allowed by videojs-standard. However, in these cases it's necessary due to the IE8 hack we see below. As such, we can just disable linting for this line since it's a legitimate use-case.

PR #3463, opened 2016-07-26
videojs-standard Utils

misteroneill:
## Description
Makes `src/js/utils/` pass videojs-standard.

Previous discussion in #3460 

## Requirements Checklist

### DIFF
--- a/src/js/utils/log.js
+++ b/src/js/utils/log.js
@@ -4,6 +4,8 @@
 import window from 'global/window';
 import {IE_VERSION} from './browser';
 
+let log;
+
 /**
  * Log messages to the console and history based on the type of message
  *
@@ -24,7 +26,7 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   // Was setting these once outside of this function, but containing them
   // in the function makes it easier to test cases where console doesn't exist
   // when the module is executed.
-  const fn = console && console[type] || function(){};
+  const fn = console && console[type] || function() {}; // eslint-disable-line
 
   if (type !== 'log') {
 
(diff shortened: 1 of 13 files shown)


src/js/utils/log.js
```diff
@@ -68,9 +71,9 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
  *
  * @function log
  */
-function log(...args) {
+log = function(...args) {
```
gkatsev: we may want to keep this as `function log` for the whole IE8 shenanigans.

src/js/utils/merge-options.js
```diff
@@ -41,7 +43,7 @@ const customizer = function(destination, source) {
  * provided objects
  * @function mergeOptions
  */
-export default function mergeOptions() {
+mergeOptions = function() {
```
gkatsev: keep this as it was for IE8...

gkatsev:
Two issues for IE8 compat.
LGTM otherwise.

src/js/utils/log.js
misteroneill: Yeah. I did it this way because it conflicts with `no-use-before-define`.

src/js/utils/merge-options.js
misteroneill: Yeah. I did it this way because it conflicts with no-use-before-define.

misteroneill: Why is this a problem in IE8?

gkatsev: Guess it doesn't need to change.
I was looking for changes from `function foo` to `foo = function` and got a bit overzealous.

PR #3464, opened 2016-07-26
videojs-standard Tracks

misteroneill:
## Description
Makes `src/js/tracks/` pass videojs-standard.

Previous discussion in #3461 

## Requirements Checklist

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -2,27 +2,59 @@
  * @file text-track-display.js
  */
 import Component from '../component';
-import Menu from '../menu/menu.js';
-import MenuItem from '../menu/menu-item.js';
-import MenuButton from '../menu/menu-button.js';
 import * as Fn from '../utils/fn.js';
-import document from 'global/document';
 import window from 'global/window';
 
 const darkGray = '#222';
 const lightGray = '#ccc';
 const fontMap = {
-  monospace:             'monospace',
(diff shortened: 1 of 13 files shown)


src/js/tracks/text-track-display.js
```diff
@@ -2,29 +2,61 @@
  * @file text-track-display.js
  */
 import Component from '../component';
-import Menu from '../menu/menu.js';
```
gkatsev: These are no longer used in this file I assume?

src/js/tracks/text-track.js
```diff
@@ -159,6 +160,8 @@ class TextTrack extends Track {
     let activeCues = new TextTrackCueList(tt.activeCues_);
     let changed = false;
     let timeupdateHandler = Fn.bind(tt, function() {
+      // This could be misleading. There is a getter on this property and if
```
gkatsev: Maybe better to say something like:
> Calling `this.activeCues` for the side-effects of updating itself due to it's nature as a getter function.
Do not remove or cues will stop updating.

Or something along the lines that explains better why it's being used here.

gkatsev:
A potential enhancement to a comment.
LGTM, otherwise.

src/js/tracks/text-track.js
misteroneill: That's better, I agree!

PR #3476, opened 2016-07-29
Suppress Infinity duration on Android Chrome before playback

mister-ben:
## Description

Re issue #3079 and the discussion in #3083. Android Chrome returns `Infinity` for VOD HLS until after playback has started. This causes issues such as an unwanted flash of the live display control.
## Specific Changes proposed

HTML5 tech will return `NaN` instead of `Infinity` if playback has not started. Fires a `durationupdate` event once the reported duration can be believed if the duration is still `Infinity`, so controls can update.
## Requirements Checklist
- [x] Unit test updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -435,7 +435,7 @@ class Html5 extends Tech {
     if (this.el_.duration === Infinity &&
       browser.IS_ANDROID && browser.IS_CHROME) {
       if (this.el_.currentTime === 0) {
-        let checkProgress = () => {
+        const checkProgress = () => {
           if (this.el_.currentTime > 0) {
             this.trigger('durationchange');
             this.off(this.player_, 'timeupdate', checkProgress);
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -428,6 +428,23 @@ class Html5 extends Tech {
    * @return {Number}
    */
   duration() {
+    // Android Chrome will report duration as Infinity for VOD HLS until after
+    // playback has started, which triggers the live display erroneously.
+    // Return 0 if playback has not started and trigger a durationupdate once
+    // the duration can be reliably known.
+    if (this.el_.duration === Infinity &&
+      browser.IS_ANDROID && browser.IS_CHROME) {
+      if (this.el_.currentTime === 0) {
+        let checkProgress = () => {
+          if (this.el_.currentTime > 0) {
+            this.trigger('durationchange');
+            this.off(this.player_, 'timeupdate', checkProgress);
+          }
+        };
+        this.on(this.player_, 'timeupdate', checkProgress);
+        return 0;
+      }
+    }
     return this.el_.duration || 0;
   }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -435,9 +435,14 @@ class Html5 extends Tech {
     if (this.el_.duration === Infinity &&
       browser.IS_ANDROID && browser.IS_CHROME) {
       if (this.el_.currentTime === 0) {
+        // Wait for the first `timeupdate` with currentTime > 0 - there may be
+        // several with 0
         const checkProgress = () => {
           if (this.el_.currentTime > 0) {
-            this.trigger('durationchange');
+            // Trigger durationchange for genuinely live video
+            if (this.el_.duration === Infinity) {
+              this.trigger('durationchange');
+            }
             this.off(this.player_, 'timeupdate', checkProgress);
           }
         };
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -430,7 +430,7 @@ class Html5 extends Tech {
   duration() {
     // Android Chrome will report duration as Infinity for VOD HLS until after
     // playback has started, which triggers the live display erroneously.
-    // Return 0 if playback has not started and trigger a durationupdate once
+    // Return NaN if playback has not started and trigger a durationupdate once
     // the duration can be reliably known.
     if (this.el_.duration === Infinity &&
       browser.IS_ANDROID && browser.IS_CHROME) {
@@ -442,10 +442,10 @@ class Html5 extends Tech {
           }
         };
         this.on(this.player_, 'timeupdate', checkProgress);
-        return 0;
+        return NaN;
       }
     }
-    return this.el_.duration || 0;
+    return this.el_.duration || NaN;
   }
 
   /**
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -446,6 +446,7 @@ class Html5 extends Tech {
             this.off(this.player_, 'timeupdate', checkProgress);
           }
         };
+
         this.on(this.player_, 'timeupdate', checkProgress);
         return NaN;
       }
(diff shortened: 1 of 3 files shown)


src/js/tech/html5.js
```diff
@@ -469,7 +469,26 @@ class Html5 extends Tech {
...
+      browser.IS_ANDROID && browser.IS_CHROME) {
+      if (this.el_.currentTime === 0) {
+        let checkProgress = () => {
+          if (this.el_.currentTime > 0) {
+            this.trigger('durationchange');
+            this.off(this.player_, 'timeupdate', checkProgress);
+          }
+        };
+        this.on(this.player_, 'timeupdate', checkProgress);
+        return 0;
```
dmlap: I believe [duration should be NaN](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-duration) when the true value is unknown.

src/js/tech/html5.js
```diff
@@ -469,7 +469,26 @@ class Html5 extends Tech {
...
+      }
+    }
+    return this.el_.duration || 0;
```
dmlap: Same note here about NaN.

dmlap:
Would you mind adding a unit test for this?

src/js/tech/html5.js
```diff
@@ -469,7 +469,26 @@ class Html5 extends Tech {
...
+    if (this.el_.duration === Infinity &&
+      browser.IS_ANDROID && browser.IS_CHROME) {
+      if (this.el_.currentTime === 0) {
+        let checkProgress = () => {
+          if (this.el_.currentTime > 0) {
+            this.trigger('durationchange');
+            this.off(this.player_, 'timeupdate', checkProgress);
+          }
+        };
+        this.on(this.player_, 'timeupdate', checkProgress);
```
misteroneill: Rather than an `on()` then an `off()` in the callback, what about using `one()`?

src/js/tech/html5.js
```diff
@@ -469,7 +469,26 @@ class Html5 extends Tech {
...
-  duration() { return this.el_.duration || 0; }
+  duration() {
+    // Android Chrome will report duration as Infinity for VOD HLS until after
+    // playback has started, which triggers the live display erroneously.
+    // Return NaN if playback has not started and trigger a durationupdate once
+    // the duration can be reliably known.
+    if (this.el_.duration === Infinity &&
+      browser.IS_ANDROID && browser.IS_CHROME) {
+      if (this.el_.currentTime === 0) {
+        let checkProgress = () => {
```
misteroneill: The newly released version of videojs-standard will complain about the use of `let` for names that are never reassigned (`const` is preferred).

src/js/tech/html5.js
misteroneill: Nevermind, I see why. 👍

mister-ben:
@misteroneill re the question I see on email notifications but not showing up here on why I've used `on()`/`off()` rather than `one()`, I get a few `timeupdate` events before `currentTime` and `duration` update.

src/js/tech/html5.js
```diff
@@ -469,7 +469,26 @@ class Html5 extends Tech {
...
+          if (this.el_.currentTime > 0) {
+            this.trigger('durationchange');
+            this.off(this.player_, 'timeupdate', checkProgress);
+          }
+        };
+        this.on(this.player_, 'timeupdate', checkProgress);
+        return NaN;
+      }
+    }
+    return this.el_.duration || NaN;
```
gkatsev: do the time components handle `NaN` correctly?

src/js/tech/html5.js
```diff
@@ -469,7 +469,26 @@ class Html5 extends Tech {
...
+    // Android Chrome will report duration as Infinity for VOD HLS until after
+    // playback has started, which triggers the live display erroneously.
+    // Return NaN if playback has not started and trigger a durationupdate once
+    // the duration can be reliably known.
+    if (this.el_.duration === Infinity &&
+      browser.IS_ANDROID && browser.IS_CHROME) {
+      if (this.el_.currentTime === 0) {
+        const checkProgress = () => {
+          if (this.el_.currentTime > 0) {
+            this.trigger('durationchange');
```
incompl: Is durationchange going to fire anyway?

src/js/tech/html5.js
mister-ben: They behave correctly and formatTime deals with it and Infinity.

src/js/tech/html5.js
gkatsev: this line is happening inside the `checkProgress` handler.

incompl: Right, which is set in `duration` and won't get set otherwise, that's my concern

gkatsev: Oh, I see what you mean.
I believe we should get at least one `durationchange` from the video loading. When we get a `durationchange` from the tech, [the player does a `techGet` on duration](https://github.com/videojs/video.js/blob/fc72fb90e7661594c0ea090e0e1e9838f8ec9486/src/js/player.js#L1013-L1015) so this gets called.
Though, it would be interesting to know whether it's possible for `durationchange` not to trigger and also when it triggers.

(not as useful but still leaving here: Our duration display updates itself and [calls `player.duration()`](https://github.com/videojs/video.js/blob/fc72fb90e7661594c0ea090e0e1e9838f8ec9486/src/js/control-bar/time-controls/duration-display.js#L54) on [`durationchange`](https://github.com/videojs/video.js/blob/fc72fb90e7661594c0ea090e0e1e9838f8ec9486/src/js/control-bar/time-controls/duration-display.js#L21).)

mister-ben: Without triggering the `durationchange` here the live display didn't appear for a live video.

test/unit/player.test.js
```diff
@@ -1044,3 +1044,21 @@ test('When VIDEOJS_NO_DYNAMIC_STYLE is set, apply sizing directly to the tech el
...
+  const player = TestHelpers.makePlayer();
+
+  browser.IS_ANDROID = true;
+  browser.IS_CHROME = true;
+
+  player.tech_.el_ = {
+    duration: Infinity,
+    currentTime: 0
+  };
+  ok(isNaN(player.tech_.duration()), 'returned NaN with currentTime 0');
```
gkatsev: We should use `Number.isNaN` instead. It's more correct and actually does what you expect it to do.

gkatsev: Also, es6-shim is included in the tests, so, it's available.

test/unit/player.test.js
```diff
@@ -1044,3 +1044,21 @@ test('When VIDEOJS_NO_DYNAMIC_STYLE is set, apply sizing directly to the tech el
   equal(player.tech_.el().width, 600, 'the width is equal to 600');
   equal(player.tech_.el().height, 300, 'the height is equal 300');
 });
+
+test('When Android Chrome reports Infinity duration with currentTime 0, return NaN', function() {
```
gkatsev: should this test be in `tech/html5.test.js` instead?

mister-ben: Yep, refactored and moved.

gkatsev:
LGTM.

Needs a rebase against master. Would you be able to do it? Otherwise, I can take care of it as part of the merge.

gkatsev:
This needs to be rebased against master.

incompl:
@gkatsev Looks like this branch has no conflicts with base branch, what needs to happen next?

vdeshpande:
@gkatsev , @mister-ben What needs to happen next here ?

mister-ben:
It's ready to go and will be pulled into the [5.13](https://github.com/videojs/video.js/milestone/13) release.

incompl:
Awesome

PR #3477, opened 2016-07-29
vjsstandard - Tests - Bottom half of test/unit

BrandonOCasey:
## Description
lint the bottom half of the files in the test/unit directory with `videojs-standard`

## Requirements Checklist

### DIFF
--- a/test/unit/plugins.test.js
+++ b/test/unit/plugins.test.js
@@ -1,27 +1,29 @@
+/* eslint-env qunit */
 import {IE_VERSION} from '../../src/js/utils/browser';
 import registerPlugin from '../../src/js/plugins.js';
 import Player from '../../src/js/player.js';
 import TestHelpers from './test-helpers.js';
 import window from 'global/window';
+import sinon from 'sinon';
 
-q.module('Plugins');
+QUnit.module('Plugins');
 
-test('Plugin should get initialized and receive options', function(){
-  expect(2);
+QUnit.test('Plugin should get initialized and receive options', function() {
+  QUnit.expect(2);
 
-  registerPlugin('myPlugin1', function(options){
-    ok(true, 'Plugin initialized');
-    ok(options['test'], 'Option passed through');
+  registerPlugin('myPlugin1', function(options) {
+    QUnit.ok(true, 'Plugin initialized');
+    QUnit.ok(options.test, 'Option passed through');
   });
 
-  registerPlugin('myPlugin2', function(options){
-    ok(false, 'Plugin initialized and should not have been');
+  registerPlugin('myPlugin2', function(options) {
+    QUnit.ok(false, 'Plugin initialized and should not have been');
   });
 
-  var player = TestHelpers.makePlayer({
-    'plugins': {
-      'myPlugin1': {
-        'test': true
+  const player = TestHelpers.makePlayer({
+    plugins: {
+      myPlugin1: {
+        test: true
       }
     }
   });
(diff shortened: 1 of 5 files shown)


test/unit/plugins.test.js
```diff
@@ -1,206 +1,214 @@
...
       }
     }
   });
 
   player.dispose();
 });
 
-test('Plugin should have the option of being initilized outside of player init', function(){
-  expect(3);
+QUnit.test('Plugin should have the option of being initilized ' +
```
gkatsev: I wonder if it's possible to tell eslint to ignore the long lines rule for tests.

test/unit/plugins.test.js
```diff
@@ -1,206 +1,214 @@
...
-  var player = TestHelpers.makePlayer({});
-  var plugin = function (name) {
-    registerPlugin(name, function (opts) {
-      this.on('test', function (event) {
+QUnit.test('Plugins should get events in registration order', function() {
+  const order = [];
+  const expectedOrder = [];
+  const pluginName = 'orderPlugin';
+  const player = TestHelpers.makePlayer({});
+  let i = 0;
```
gkatsev: we might as well move this into the forloop.

test/unit/plugins.test.js
```diff
@@ -1,206 +1,214 @@
...
-  var player = TestHelpers.makePlayer({});
-  var plugin = function (name) {
-    registerPlugin(name, function (opts) {
-      this.on('test', function (event) {
+QUnit.test('Plugins should not get events after ' +
+           'stopImmediatePropagation is called', function() {
+  const order = [];
+  const expectedOrder = [];
+  const pluginName = 'orderPlugin';
+  let i = 0;
```
gkatsev: same here.

test/unit/plugins.test.js
misteroneill: It's just a warning; so, I've been ignoring them.

BrandonOCasey: yeah its just a warning and I could put it back to the full line if we want

BrandonOCasey: we could use `"ignorePattern": /.*test(.*` on the max-len rule or some similar regex to at least stop us forcing it on test names

BrandonOCasey:
re-commenting on the `QUnit.test(` `max-len` issues, just so it does not get lost to the void: 

we could use "ignorePattern": /.*test(.* on the max-len rule or some similar regex. to turn it off for test names, or maybe test names should never be this long?

misteroneill:
In other projects, I'm using `tsmlj` for long strings like this:

```js
QUnit.test(tsmlj`
  this is a really long test name. it has a whole bunch of characters - even more
  than 80 or the warning limit of 90!
`, function(assert) {
  // ...
});
```

misteroneill:
LGTM

test/unit/plugins.test.js
```diff
@@ -1,206 +1,210 @@
+/* eslint-env qunit */
 import {IE_VERSION} from '../../src/js/utils/browser';
 import registerPlugin from '../../src/js/plugins.js';
 import Player from '../../src/js/player.js';
 import TestHelpers from './test-helpers.js';
 import window from 'global/window';
+import sinon from 'sinon';
```
gkatsev: if we're importing sinon directly, we should remove it from the `global-shim.js` file.

test/unit/video.test.js
```diff
@@ -1,133 +1,145 @@
...
+  const fixture = document.getElementById('qunit-fixture');
 
   fixture.innerHTML += '<video id="test_vid_id"><source type="video/mp4"></video>';
 
-  var vid = document.getElementById('test_vid_id');
+  const vid = document.getElementById('test_vid_id');
 
-  ok(vid.childNodes.length === 1);
-  ok(vid.childNodes[0].getAttribute('type') === 'video/mp4');
+  QUnit.ok(vid.childNodes.length === 1);
```
gkatsev: ugh, a lot of these need to change to use `equal` and have a message. Definitely not for this PR. Just didn't realize how many of these we still had.

test/unit/video.test.js
```diff
@@ -1,133 +1,145 @@
...
-  var fixture = document.getElementById('qunit-fixture');
-  fixture.innerHTML += '<video id="test_vid_id"></video><video id="test_vid_id2"></video>';
+QUnit.test('should return a video player instance', function() {
+  const fixture = document.getElementById('qunit-fixture');
 
-  var player = videojs('test_vid_id', { techOrder: ['techFaker'] });
-  ok(player, 'created player from tag');
-  ok(player.id() === 'test_vid_id');
-  ok(videojs.getPlayers()['test_vid_id'] === player, 'added player to global reference');
+  fixture.innerHTML += '<video id="test_vid_id">' +
```
gkatsev: what about splitting after `</video>`?

gkatsev:
A couple of minor comments.

test/unit/video.test.js
misteroneill: What about `strictEqual`? If `===` is better than `==` then it follows that `strictEqual` is better than `equal`. 😄

PR #3478, opened 2016-07-29
Linter v5 Compatibility

misteroneill:
## Description
This makes the source code pass v5 of videojs-standard.

## Requirements Checklist

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -64,7 +64,7 @@ class Component {
     // If there was no ID from the options, generate one
     if (!this.id_) {
       // Don't require the player ID function in the case of mock players
-      let id = player && player.id && player.id() || 'no_player';
+      const id = player && player.id && player.id() || 'no_player';
 
       this.id_ = `${id}_component_${Guid.newGUID()}`;
     }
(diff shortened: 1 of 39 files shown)


src/js/player.js
```diff
@@ -3015,10 +3015,10 @@ Player.prototype.handleVolumeChange_; // eslint-disable-line
  *
  * @event error
  */
-Player.prototype.handleError_ = Player.prototype.handleError_;
+Player.prototype.handleError_; // eslint-disable-line
```
BrandonOCasey: how does this work and is it needed? seems like it is never used from a quick code search

gkatsev: it's used for the jscond `@event` comment.

src/js/tracks/text-track.js
```diff
@@ -156,15 +156,15 @@ class TextTrack extends Track {
...
+    const activeCues = new TextTrackCueList(tt.activeCues_);
     let changed = false;
-    let timeupdateHandler = Fn.bind(tt, function() {
+    const timeupdateHandler = Fn.bind(tt, function() {
 
       // Accessing this.activeCues for the side-effects of updating itself
       // due to it's nature as a getter function. Do not remove or cues will
       // stop updating!
-      this.activeCues;
+      this.activeCues; // eslint-disable-line
```
BrandonOCasey: you may want to wrap this in the specific rule that you are turning off. This will protect use  in case we change linting in the future and also make it more apparent why this line needs to be disabled (although the comment above does help already)

BrandonOCasey:
LGTM other than my small comments

src/js/tracks/text-track.js
misteroneill: Good idea! 👍

src/js/component.js
```diff
@@ -1452,7 +1452,7 @@ class Component {
     subObj.extend = Component.extend;
 
     // Extend subObj's prototype with functions and other properties from props
-    for (let name in props) {
+    for (const name in props) {
```
gkatsev: are you sure this shouldn't be a `let`?

src/js/extend.js
```diff
@@ -67,7 +67,7 @@ const extendFn = function(superClass, subClassMethods = {}) {
   _inherits(subClass, superClass);
 
   // Extend subObj's prototype with functions and other properties from props
-  for (let name in methods) {
+  for (const name in methods) {
```
gkatsev: ditto about `let`

gkatsev:
I guess `const` in a `for..in` is fine? If so, 
LGTM.

misteroneill:
Yeah, ESLint complains about it. Which makes _some_ sense, I suppose.

PR #3480, opened 2016-08-01
add initial axe-core check

gkatsev:
## Description
Add a11y checks using axe-core. Currently, it's just being done with the default player created by `TestHelpers`.
We need to add a `descriptions` and `captions` track to not get any violations. 
We still should make sure that the control bar gets added and linted by axe or this won't be that helpful to us.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "xhr": "2.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


gkatsev:
Just going to close this. Some a11y testing will be done as part of https://github.com/videojs/video.js/pull/3487

PR #3483, opened 2016-08-01
vjsstandard - Tests - 1/2 of Tracks folder

BrandonOCasey:
## Description
linted half of the unit tests in the test/unit/tracks folder

## Requirements Checklist

### DIFF
--- a/test/unit/tracks/html-track-element-list.test.js
+++ b/test/unit/tracks/html-track-element-list.test.js
@@ -1,4 +1,4 @@
-import HTMLTrackElement from '../../../src/js/tracks/html-track-element.js';
+/* eslint-env qunit */
 import HTMLTrackElementList from '../../../src/js/tracks/html-track-element-list.js';
 import TextTrack from '../../../src/js/tracks/text-track.js';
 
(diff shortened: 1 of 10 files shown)


misteroneill:
LGTM

test/unit/tracks/html-track-element-list.test.js
```diff
@@ -28,31 +28,47 @@ const genericHtmlTrackElements = [{
   track: track2
 }];
 
-q.module('HTML Track Element List');
+QUnit.module('HTML Track Element List');
 
-test('HTMLTrackElementList\'s length is set correctly', function() {
-  let htmlTrackElementList = new HTMLTrackElementList(genericHtmlTrackElements);
+QUnit.test('HTMLTrackElementList\'s length is set correctly', function() {
```
gkatsev: if it isn't enabled yet, we should enable the eslint rule that allows us to wrap this in double quotes.

test/unit/tracks/text-track-controls.test.js
```diff
@@ -16,236 +18,266 @@ const track = {
...
+  QUnit.ok(!player.controlBar.descriptionsButton.hasClass('vjs-disabled'),
+           'descriptions control is NOT disabled');
 
   player.dispose();
 });
 
 if (!browser.IS_IE8) {
   // This test doesn't work on IE8.
-  // However, this test tests a specific with iOS7 where the TextTrackList doesn't report track mode changes.
+  // However, this test tests a specific with iOS7 where
```
gkatsev: I thought comments were exempt? Or was it only for when they had urls?

gkatsev:
LGTM

test/unit/tracks/text-track-controls.test.js
BrandonOCasey: only urls are exempt, so I think that means only urls at the end of a line that run way over

test/unit/tracks/html-track-element-list.test.js
BrandonOCasey: it seems to be allowed, would you like me to change it? I am not sure if eslint can pick lines like this up so that we can automatically know about them.

gkatsev: nah, it's fine for now.

PR #3484, opened 2016-08-01
vjsstandard - Tests - Other 1/2 of Tracks folder + Misc

BrandonOCasey:
## Description
linted the other half of the `test/unit/tracks` folder and some other misc files

## Requirements Checklist

### DIFF
--- a/test/api/api.js
+++ b/test/api/api.js
@@ -1,14 +1,13 @@
 /* eslint-env qunit */
-/* global videojs */
 /**
  * These tests run on the minified, window.videojs and ensure the needed
  * APIs still exist
  */
 import document from 'global/document';
 import window from 'global/window';
+const videojs = window.videojs;
 
 QUnit.module('Player API');
-
 QUnit.test('videojs should exist on the window', function() {
   QUnit.ok(window.videojs, 'videojs exists on the window');
 });
--- a/test/api/api.js
+++ b/test/api/api.js
@@ -1,212 +1,232 @@
+/* eslint-env qunit */
+/* global videojs */
 /**
  * These tests run on the minified, window.videojs and ensure the needed
  * APIs still exist
  */
+import document from 'global/document';
+import window from 'global/window';
 
-(function(){
+QUnit.module('Player API');
 
-q.module('Player API');
-
-test('videojs should exist on the window', function() {
-  ok(window.videojs, 'videojs exists on the window');
+QUnit.test('videojs should exist on the window', function() {
+  QUnit.ok(window.videojs, 'videojs exists on the window');
 });
 
-test('should be able to access expected player API methods', function() {
-  var player = videojs.getComponent('Player').prototype;
+QUnit.test('should be able to access expected player API methods', function() {
+  const player = videojs.getComponent('Player').prototype;
 
   // Native HTML5 Methods
-  ok(player.error, 'error exists');
-  ok(player.src, 'src exists');
-  ok(player.currentSrc, 'currentSrc exists');
-  ok(player.buffered, 'buffered exists');
-  ok(player.load, 'load exists');
-  ok(player.seeking, 'seeking exists');
-  ok(player.currentTime, 'currentTime exists');
-  ok(player.duration, 'duration exists');
-  ok(player.paused, 'paused exists');
-  ok(player.ended, 'ended exists');
-  ok(player.autoplay, 'autoplay exists');
-  ok(player.loop, 'loop exists');
-  ok(player.play , 'play exists');
-  ok(player.pause , 'pause exists');
-  ok(player.controls, 'controls exists');
-  ok(player.volume, 'volume exists');
-  ok(player.muted, 'muted exists');
-  ok(player.width, 'width exists');
-  ok(player.height, 'height exists');
-  ok(player.poster, 'poster exists');
-  ok(player.textTracks, 'textTracks exists');
-  ok(player.requestFullscreen, 'requestFullscreen exists');
-  ok(player.exitFullscreen, 'exitFullscreen exists');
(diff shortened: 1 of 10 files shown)


test/api/api.js
```diff
@@ -1,212 +1,232 @@
+/* eslint-env qunit */
+/* global videojs */
```
misteroneill: Perhaps this should go _after_ the explanatory comment below?

test/api/api.js
```diff
@@ -1,212 +1,232 @@
...
-  ok(player.width, 'width exists');
-  ok(player.height, 'height exists');
-  ok(player.poster, 'poster exists');
-  ok(player.textTracks, 'textTracks exists');
-  ok(player.requestFullscreen, 'requestFullscreen exists');
-  ok(player.exitFullscreen, 'exitFullscreen exists');
-  ok(player.playbackRate, 'playbackRate exists');
-  ok(player.networkState, 'networkState exists');
-  ok(player.readyState, 'readyState exists');
+  QUnit.ok(player.error, 'error exists');
```
misteroneill: I would've been _sooooo_ tempted to refactor this into a loop. I'm glad you resisted the urge! 😄

test/api/api.js
BrandonOCasey: I am actually going to define videojs at the top as `const videojs = window.videojs`

misteroneill:
LGTM

test/api/api.js
```diff
@@ -1,212 +1,231 @@
...
+  QUnit.ok(player.height, 'height exists');
+  QUnit.ok(player.poster, 'poster exists');
+  QUnit.ok(player.textTracks, 'textTracks exists');
+  QUnit.ok(player.requestFullscreen, 'requestFullscreen exists');
+  QUnit.ok(player.exitFullscreen, 'exitFullscreen exists');
+  QUnit.ok(player.playbackRate, 'playbackRate exists');
+  QUnit.ok(player.networkState, 'networkState exists');
+  QUnit.ok(player.readyState, 'readyState exists');
 
   // Unsupported Native HTML5 Methods
```
gkatsev: most of these are available now. Though, we shouldn't turn it on as part of this.

test/api/api.js
```diff
@@ -1,212 +1,231 @@
...
+  QUnit.ok(player.cancelFullScreen, 'cancelFullScreen exists');
 });
 
-test('should be able to access expected component API methods', function() {
-  var Component = videojs.getComponent('Component');
-  var comp = new Component({ id: function(){ return 1; }, reportUserActivity: function(){} });
+QUnit.test('should be able to access expected component API methods', function() {
+  const Component = videojs.getComponent('Component');
+  const comp = new Component({
+    id() {
```
gkatsev: making it an arrow function will make it even shorter :P

gkatsev: (no action required or necessary)

test/unit/tracks/text-track.test.js
```diff
@@ -265,56 +262,60 @@ test('tracks are parsed if vttjs is loaded', function() {
 
   // use proxyquire to stub xhr module because IE8s XDomainRequest usage
   let xhrHandler;
-  let TextTrack = proxyquire('../../../src/js/tracks/text-track.js', {
+  const TextTrack_ = proxyquire('../../../src/js/tracks/text-track.js', {
```
gkatsev: why rename?

test/unit/tracks/text-tracks.test.js
```diff
@@ -175,37 +166,38 @@ test('update texttrack buttons on removetrack or addtrack', function() {
...
-  track = document.createElement('track');
+  const tag = document.createElement('video');
+  const track = document.createElement('track');
+
   track.kind = 'captions';
   track.label = 'en';
   track.language = 'English';
   track.src = '#en.vtt';
   tag.appendChild(track);
-  track = document.createElement('track');
```
gkatsev: removing this line changes the test.

test/unit/tracks/text-tracks.test.js
```diff
@@ -216,23 +208,20 @@ test('update texttrack buttons on removetrack or addtrack', function() {
...
+  const track = document.createElement('track');
 
-  tag = document.createElement('video');
-  track = document.createElement('track');
   track.kind = 'captions';
   track.label = 'en';
   track.language = 'English';
   track.src = 'en.vtt';
   tag.appendChild(track);
-  track = document.createElement('track');
```
gkatsev: same here

test/unit/tracks/text-track.test.js
BrandonOCasey: since TextTrack is imported at the top eslint complains about redefining a variable

test/unit/tracks/text-tracks.test.js
BrandonOCasey: they are defined above with most of the other variable definitions

test/unit/tracks/text-tracks.test.js
BrandonOCasey: they are defined above with most of the other variable definitions

test/unit/tracks/text-tracks.test.js
gkatsev: the issue is that we want to create a new track here. With the current change, we end up with one track being created, have various attributes set and appended to tag in line 184/176 but then below this line it ends up re-using the same track element which we don't want.

BrandonOCasey:
@gkatsev changed the track creation so that we create two  tracks `const track1` and `const track2`. If it was I can revert it back to the way it was, but I think it will prevent confusion in the future.

gkatsev:
Making two tracks sounds good to me!

PR #3485, opened 2016-08-01
Control text fullscreen fix on ESC

vdeshpande:
## Description
This PR is to make sure the correct control text is displayed on the Fullscreen button after pressing the ESC key while in Fullscreen mode.

**Current Status:**
_(Steps to reproduce)_
1) Load the player on a page.
2) Press the fullscreen button, the player goes into fullscreen mode.
2) Hit the ESC key on your keyboard, and the player will go back to normal mode, however the control text on fullscreen button does not change or revert back to text as `Fullscreen`. It reads as `Non-Fullscreen` which is not correct.

**After this fix:**
Follow Step 1  and Step 2 as above. Press the ESC key and the control text should now revert back to the correct text as `Fullscreen`

**Before this fix (current scenario)**
before_fix_fullscreen
**After this fix**
after_fix_fullscreen

## Specific Changes proposed
The changes made in this fix makes sure that we have the correct control text when you exit fullscreen mode.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/player.js
```diff
@@ -1092,6 +1092,8 @@ class Player extends Component {
       this.addClass('vjs-fullscreen');
     } else {
       this.removeClass('vjs-fullscreen');
+      // Making sure the Control text reverts back after pressing ESC
+      this.controlBar.fullscreenToggle.controlText('Fullscreen');
```
gkatsev: the fullscreen toggle [handles](https://github.com/videojs/video.js/blob/fc72fb90e7661594c0ea090e0e1e9838f8ec9486/src/js/control-bar/fullscreen-toggle.js#L30-L38) the control text changes. We should make that component handle this use case correctly as well.

mister-ben:
Currently the control's text updates even when an attempt to switch to fullscreen is unsuccessful, such as when in an iframe without `allowfullscreen` - [example](http://jsbin.com/royepaz/edit?html,output). It would be good to fix that here too.

src/js/clickable-component.js
```diff
@@ -28,6 +28,7 @@ class ClickableComponent extends Component {
     this.on('click', this.handleClick);
     this.on('focus', this.handleFocus);
     this.on('blur', this.handleBlur);
+    player.on('fullscreenchange', this.handleFullscreenChange);
```
gkatsev: why not do this only in `fullscreen-toggle`? Generic clickable-components don't need to know about it.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -21,7 +21,21 @@ class FullscreenToggle extends Button {
...
   }
-
+  /**
+   * Handles Fullscreenchange on the component and change control text accordingly
+   *
+   * @method handleFullscreenChange
+   */
+  handleFullscreenChange() {
+    if (this.player_.isFullscreen()) {
+      this.addClass('vjs-fullscreen');
```
gkatsev: doesn't the player already handle this?

gkatsev:
There are test failures.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -21,7 +21,21 @@ class FullscreenToggle extends Button {
...
+      this.controlBar.fullscreenToggle.controlText('Non-Fullscreen');
+    }
+    else {
+      this.removeClass('vjs-fullscreen');
+      this.controlBar.fullscreenToggle.controlText('Fullscreen');
```
gkatsev: we should make sure the context for this method is the fullscreen toggle itself, otherwise it can be super confusing.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -12,6 +12,10 @@ import Component from '../component.js';
  */
 class FullscreenToggle extends Button {
 
+  constructor(player, options) {
+    super(player, options);
+    player.on('fullscreenchange', this.handleFullscreenChange);
```
gkatsev: if you change this to `this.on(player, 'fullscreenchange', this.handleFullscreenChange);` it'll make the context of the handler be `this` (the fullscreen toggle) plus it'll manage the event handler better since it'll be removed when this component is disposed and not necessarily just when the player is disposed.

gkatsev:
Also, would be nice to get a test for this, I think it'll just go into the `test/unit/controls.js`.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -21,7 +25,20 @@ class FullscreenToggle extends Button {
...
     return `vjs-fullscreen-control ${super.buildCSSClass()}`;
   }
-
+  /**
+   * Handles Fullscreenchange on the component and change control text accordingly
+   *
+   * @method handleFullscreenChange
+   */
+  handleFullscreenChange() {
+    let fst = this.player().controlBar.fullscreenToggle;
```
OwenEdwards: This seems unreliable - making the change that @gkatsev suggests would mean that `fst` would be entirely replaced in this method by `this`, which would make  FullscreenToggle much more robust and extensible.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -21,7 +25,19 @@ class FullscreenToggle extends Button {
...
+  /**
+   * Handles Fullscreenchange on the component and change control text accordingly
+   *
+   * @method handleFullscreenChange
+   */
+  handleFullscreenChange() {
+    if (this.player_.isFullscreen()) {
+      this.controlText('Non-Fullscreen');
+    }
+    else {
```
misteroneill: I don't think this style will pass videojs-standard.

test/unit/controls.test.js
```diff
@@ -112,3 +113,14 @@ test('should hide playback rate control if it\'s not supported', function(){
 
   ok(playbackRate.el().className.indexOf('vjs-hidden') >= 0, 'playbackRate is not hidden');
 });
+
+test('Fullscreen control text should be correct when fullscreenchange is triggered', function() {
+  var player = TestHelpers.makePlayer();
+  var fullscreentoggle = new FullscreenToggle(player);
```
misteroneill: I don't think this style will pass videojs-standard.

misteroneill:
I think this needs to be brought up to date with `master` because automated linting via videojs-standard has been added there.

vdeshpande:
@misteroneill How do the above changes look ?

gkatsev:
I think something went wrong in the rebase. You have extra commits in this branch now.

misteroneill:
LGTM

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -12,6 +12,10 @@ import Component from '../component.js';
  */
 class FullscreenToggle extends Button {
 
+  constructor(player, options) {
+    super(player, options);
+    this.on(player, 'fullscreenchange', this.handleFullscreenChange);
+  }
```
gkatsev: new line after this and before the next comment.

gkatsev:
One super small comment

vdeshpande:
@gkatsev done

misteroneill:
My LGTM still applies 😄

BrandonOCasey:
LGTM

DerekCalder:
This bug has returned. It is exactly as described in the original Description. I see it happening with version 7.6.6 in Chrome and Firefox. Here is a jsfiddle where it can be reproduced: https://jsfiddle.net/fzkLhs7e/1/. Should I open a new issue for this?

gkatsev:
Generally, best to open new issues for this as commenting on old things will likely get lost. Almost lost this but remembered seeing a comment about it as I'm currently refactoring our fullscreen support.

gkatsev:
I believe it will be fixed as part of https://github.com/videojs/video.js/pull/6422.

PR #3487, opened 2016-08-03
Basic accessibility testing with HTML_CodeSniffer in Grunt's test phase

OwenEdwards:
## Description
Add automated accessibility (`a11y`) testing to Grunt's testing. Use HTML_CodeSniffer, which does fairly reliable in-browser testing, to check for any automatically-detectable violations in the video.js player core, using `sandbox/descriptions.html.example` (since that file has a fairly comprehensive set of additional tracks for accessibility).

Note that automated testing can only detect a fraction of potential accessibility issues, since accessibility is inherently a functional issue rather than purely technical - passing these tests should not be considered a thorough test of accessibility, but just a baseline verification. (Note, for example, that it did flag a11y violations with the Captions Settings Dialog before #3281)

## Requirements Checklist
  - [x] Unit Tests updated or fixed - **N/A**
  - [ ] Docs/guides updated
  - [x] Example created - **N/A**

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -443,6 +443,19 @@ module.exports = function(grunt) {
           src: ['build/temp/video.js']
         }
       }
+    },
+    accessibility: {
+      options: {
+        accessibilityLevel: 'WCAG2AA',
+        reportLevels: {
+          notice: false,
+          warning: true,
+          error: true
+        }
+      },
+      test: {
+        src: ['sandbox/index.html']
+      }
     }
   });
 
@@ -451,6 +464,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('videojs-doc-generator');
   grunt.loadNpmTasks('chg');
   grunt.loadNpmTasks('gkatsev-grunt-sass');
+  grunt.loadNpmTasks('grunt-accessibility');
 
   const buildDependents = [
     'clean:build',
@@ -501,6 +515,8 @@ module.exports = function(grunt) {
   grunt.registerTask('dev', ['build', 'connect:dev', 'concurrent:watchSandbox']);
 
   grunt.registerTask('watchAll', ['build', 'connect:dev', 'concurrent:watchAll']);
+
+  grunt.registerTask('test-a11y', ['build', 'accessibility']);
 
   // Pick your testing, or run both in different terminals
   grunt.registerTask('test-ui', ['browserify:tests']);
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -515,14 +515,14 @@ module.exports = function(grunt) {
   grunt.registerTask('default', ['test']);
 
   // The test script includes coveralls only when the TRAVIS env var is set.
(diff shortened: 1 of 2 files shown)


build/grunt.js
```diff
@@ -190,6 +190,7 @@ module.exports = function(grunt) {
       swf:   { cwd: 'node_modules/videojs-swf/dist/', src: 'video-js.swf', dest: 'build/temp/', expand: true, filter: 'isFile' },
       ie8:   { cwd: 'node_modules/videojs-ie8/dist/', src: ['**/**'], dest: 'build/temp/ie8/', expand: true, filter: 'isFile' },
       dist:  { cwd: 'build/temp/', src: ['**/**', '!test*'], dest: 'dist/', expand: true, filter: 'isFile' },
+      a11y:  { src: 'sandbox/descriptions.html.example', dest: 'sandbox/descriptions.test-a11y.html' }, // Can only test a file with a .html or .htm extension
```
gkatsev: maybe `build/temp` is the directory we want to use?

OwenEdwards: The problem with that is that the `js` and `css` references from the example file(s) break, because they specifically point to `../build/temp/`. Any suggestions?

gkatsev: maybe it's just time to remove `.example` suffixes from the examples?

OwenEdwards: That would make thinks a *lot* simpler! I don't understand why they're `.example`, although I guess it's something to do with git/GitHub? But yes, it means all sorts of things may misidentify those files.

gkatsev: I think it's because if you force-add an ignored file, git will show it to you as changes when you edit that file.

OwenEdwards: Then why ignore `html` files in the sandbox? Why not let people try them out as they are, and edit them or add others, but not commit them? Or otherwise move the existing examples into  an example directory, with the option to copy them into the sandbox where they'll be ignored. I'm missing something about the intention of the sandbox and the examples, I guess!

gkatsev: Everything inside of `sandbox` is ignored. The idea is that you can do whatever you want in there and it won't be accidentally checked in. The issue arrises if we have a `index.html` file inside of `sandbox` (without the `.example`), when you make changes to that file, even though git is supposed to ignore `sandbox/` because `index.html` was previously checked in, it'll show up as modified. Then, if people were trying to make a PR, modified `sandbox/index.html` and committed by running `git add -A` the change to the `sandbox/index.html` would get added to the commit.

Maybe, we can move the `.example` files into something like `build/examples` and then have a grunt task that copies those files into the sandbox? We can even do it as a post-install (when you're doing `npm install` on a clone, for example).

gkatsev: I guess this is fine for now. Better to just have something than nothing.

gkatsev:
Closed via 7d85f27a8e48a46f821d3dafdae67d6850c1cb53

PR #3493, opened 2016-08-03
Fix the language table in the docs

BrandonOCasey:
## Description
Noticed that the language table on http://docs.videojs.com/docs/guides/languages.html is not correctly formatted. Looked at the markdown file for it and saw some malformed HTML. Fixed it

## Requirements Checklist


mister-ben:
Does this doc need to list language codes? Something like this from the [explanation of Navigator.language on MDN](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language) might suffice:

> A string representing the language version as defined in BCP 47. Examples of valid language codes include "en", "en-US", "fr", "es-ES", etc.

misteroneill:
I'm all for leaving the table out - especially when there's a link directly to an always-updated list of files/languages. This seems like wasteful documentation.

BrandonOCasey:
yeah that works too, I can just change it to a link

gkatsev:
I think the original idea was to have the list of names for the json files in an easily accessible place.

misteroneill:
The problem I see there is that you then have to remember to update this document anytime a new file is added/removed. Not that it's a common occurrence, but...

gkatsev:
this list has *all** the languages, not just the ones currently available in videojs.
Also, I wasn't necessarily saying we need to keep it, was just trying to give background on why it's there. If such a list is available in a more official place (specifically the two-letter language codes) we could just link it.

*not actually all, but a large selection of the most commonly used languages

BrandonOCasey:
http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

misteroneill:
Oh, I see, then ignore me. :)

BrandonOCasey:
@gkatsev @mister-ben @misteroneill So the consensus here is to change the table into a link right?  How about [this](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry)?

misteroneill:
LGTM

gkatsev:
LGTM

PR #3498, opened 2016-08-04
Disable HLS hack on Firefox for Android

kentuckyfriedtakahe:
## Description
Chrome reports HLS is not supported when it actually so a hack was added to work around it. Contrastingly Firefox for Android actually doesn't support HLS. This fixes the Firefox compat bug https://bugzilla.mozilla.org/show_bug.cgi?id=1194662 

## Specific Changes proposed
Firefox for Android will use alternative fallbacks if they are available rather than hitting a dead end.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1230,7 +1230,7 @@ const mp4RE = /^video\/mp4/i;
 
 Html5.patchCanPlayType = function() {
   // Android 4.0 and above can play HLS to some extent but it reports being unable to do so
-  if (browser.ANDROID_VERSION >= 4.0) {
+  if (browser.ANDROID_VERSION >= 4.0 && !browser.IS_FIREFOX) {
     if (!canPlayType) {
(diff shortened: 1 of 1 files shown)


mister-ben:
The tests need to be fixed, but this lets videojs-contrib-hls work on Android FF too.

gkatsev:
Ah, yeah, the [two tests](https://github.com/videojs/video.js/blob/master/test/unit/tech/html5.test.js#L104) just set `browser.ANDROID_VERSION` but since it's run in firefox `browser.IS_FIREFOX` is set.
We probably want to set that property to `false` in these two tests and add a test that verifies that when `browser.IS_FIREFOX` is true and `browser.ANDROID_VERSION` is 4.0 and higher, we don't patch `canPlayType`.

marco-c:
I've opened #3586, it contains the changes from this PR + fix for the tests + a new test for Firefox for Android.

gkatsev:
Closing in favor of #3586. That PR will get merged as two commits to give credit to both @kentuckyfriedtakahe and @marco-c.

PR #3504, opened 2016-08-05
fix errors for quotes rule

BrandonOCasey:
## Description
noticed that there were still two lint errors on my machine when running videojs-standard, Also we need to lint on PR's

## Requirements Checklist


BrandonOCasey:
handled by #3508

PR #3508, opened 2016-08-08
Use videojs-standard 5.2 and fix two errors

misteroneill:
## Description
This updates to videojs-standard 5.2 or higher and fixes two errors. The benefit of the upgrade is that we get correct exit codes from the videojs-standard CLI tool so the git pre-push hook actually works to prevent invalid pushes.

## Requirements Checklist

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -454,6 +454,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('gkatsev-grunt-sass');
 
   const buildDependents = [
+    'shell:lint',
     'clean:build',
 
     'browserify:build',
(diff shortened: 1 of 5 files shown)


build/grunt.js
```diff
@@ -482,6 +483,7 @@ module.exports = function(grunt) {
   );
 
   grunt.registerTask('dist', [
+    'shell:lint',
```
gkatsev: should we lint as part of running dist?

misteroneill: Hmm, probably not. Good call.

BrandonOCasey:
LGTM

gkatsev:
LGTM

Issue #3157, opened 2016-03-06
Swiping video causes video to play on Chrome Android


MaxLaumeister:
Related to the closed issue https://github.com/videojs/video.js/issues/1817, though this time it's Android rather than iOS.

For me, swipes are being mistaken as taps on Chrome for Android (Chrome 48.0.2564.95 on stock Android 6.0.1, Nexus 5) in Video.js 5.4.6. My site's mobile view is very dense with videos in some places, so the user is just about guaranteed to put their finger on one of the videos as they scroll past them. When this happens, the video registers the swipe as a deliberate tap and plays.

Here are steps to reproduce on the Video.js homepage using Chrome for Android: Wait until the page loads fully, then casually swipe up on the video to scroll down the page. Unless you make your swipe deliberately slow, the video will start to play as you scroll past it. A user who is just trying to scroll down the page will have videos start randomly playing, which is not ideal.

Here's a video (gif) of it happening on the Video.js homepage. Notice that when I scroll back up, the video has already started playing from when I swiped to scroll down:

![ezgif-2913097299](https://cloud.githubusercontent.com/assets/4030205/12501177/353805da-c06f-11e5-9098-9ac4d7627a92.gif)

amercier:
I'm having the same issue on Chrome 49.0.2623.91, Android 6.0.1, Nexus 5.
I was able to reproduce on both http://videojs.com/ (Video.js 5.8.0) and on my project (Video.js 4.12.13).

gabesoft:
This happens on IOS as well. But, only after pressing play on the video once. So to reproduce: 
- press play
- press pause
- scroll by touching the player

outlawtorn:
+1

mister-ben:
`assign()` returns `{}` [here](https://github.com/videojs/video.js/blob/master/src/js/component.js#L1183) on Android Chrome, so `touchDistance` is NaN. On iOS the object is copied as expected.

mister-ben:
Fixed in 5.11.3

PR #3514, opened 2016-08-09
Stop Android treating swipe as tap

mister-ben:
## Description
On Android, a short duration swipe is always considered a tap. Scrolling a page can easily start playback unintentionally - see #3157.

`assign()` returns `{}` [in the touchstart handler](https://github.com/videojs/video.js/blob/master/src/js/component.js#L1183), so `touchDistance` is always `NaN` [in the touchmove handler](https://github.com/videojs/video.js/blob/master/src/js/component.js#L1202).

It seems `pageX`, `pageY` etc are "own" properties of the Touch on iOS Safari (and of the object used in the tests) but not on Android Chrome, hence the difference in behaviour of `assign()`.

## Specific Changes proposed
Copy the specific properties needed instead of using `assign()`. Unless I'm missing something about the point of using `assign()` in the first place...

## Requirements Checklist

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -10,7 +10,6 @@ import * as Guid from './utils/guid.js';
 import * as Events from './utils/events.js';
 import log from './utils/log.js';
 import toTitleCase from './utils/to-title-case.js';
-import assign from 'object.assign';
 import mergeOptions from './utils/merge-options.js';
 
 /**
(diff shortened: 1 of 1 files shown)


gkatsev:
Looks good to me. Thanks for catching that @mister-ben!

outlawtorn:
I have this problem on firefox 50 and video.js 5.16.0 (laptop convertible to tablet)

Issue #3474, opened 2016-07-28
Issue with async play() in Chrome


diestrin:
## Description
Chrome recently changed the play() method to return a promise, and this is causing issue when you call load() or pause() before the promise is resolved:
_Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause()_
_Uncaught (in promise) DOMException: The play() request was interrupted by a call to load()_

## Steps to reproduce
1. Use Chrome
2. Go to http://videojs.com/advanced/ and open the web console
3. Play the video and click in the timeline in a place where the video isn't buffered yet (try using a network throttling if the video is loading too fast)
4. Move the timeline again to a place where the video isn't buffered yet (The issue will start after the second time you do this)

-- or --

1. Go to http://videojs.com/advanced/ and open the web console
2. Play any video
3. Quickly play another video before the 1st start to reproduce
4. You can do step 3 as many times as you want quickly and you'll see a lot of errors in the console

## Results
En error is logged, but the video is still working
### Expected
Wait for the play() promise to be resolved before calling another method
### Actual
The play is expected to be sync and thus pause() and load() are being callede inmediatelly
### Error output
Case 1: _Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause()_
Case 2: _Uncaught (in promise) DOMException: The play() request was interrupted by a call to load()_

## Additional Information
Bug report in chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=593273
#### videojs
5.10.7 and 5.11.0
#### browsers
Chrome 51.0.2704.103 (not only this version)
#### OSes
OSx 10.10.5 (not only this OS)
### plugins
No

wingchuihk:
Thank you Austin for the call this morning, and got a better understanding
on the product. I m looking forward to hear back from your team.

Have a great weekend!
Kenneth

gkatsev:
This will be fixed by https://github.com/videojs/video.js/pull/3518

misteroneill:
This should be fixed in 5.11.3

antonioribeiro:
I'm doing

    videojs.seek(seekTo).pause();

And receiving this

![image](https://cloud.githubusercontent.com/assets/3182864/20503984/54969d6c-b02c-11e6-949d-41fff926bc1d.png)

On 

    * Video.js 5.13.2 <http://videojs.com/>

Removing the pause fixes warning, but, of course, video start playing. Is this something I'm doing wrong?

[...]

PR #3518, opened 2016-08-10
Silence "The play() request was interrupted by a call to pause()"

mister-ben:
## Description
On browsers that return a promise for `play()` (so far Chrome), an `AbortError` DOMException is thrown if a `pause()` occurs before playback starts. This doesn't break anything, but logs an annoying error to the console. 

## Specific Changes proposed
Just catch the exception and do nothing.

## Requirements Checklist

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -441,8 +441,8 @@ class Html5 extends Tech {
 
     // Catch/silence error when a pause interrupts a play request
     // on browsers which return a promise
-    if (playPromise !== undefined) {
-      playPromise.catch((e) => {});
+    if (playPromise !== undefined && typeof playPromise.then === 'function') {
+      playPromise.then(null, (e) => {});
     }
   }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -437,7 +437,13 @@ class Html5 extends Tech {
    * @method play
    */
   play() {
-    this.el_.play();
+    const playPromise = this.el_.play();
+
+    // Catch/silence error when a pause interrupts a play request
+    // on browsers which return a promise
+    if (playPromise !== undefined) {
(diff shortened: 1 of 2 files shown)


src/js/tech/html5.js
```diff
@@ -437,7 +437,13 @@ class Html5 extends Tech {
    * @method play
    */
   play() {
-    this.el_.play();
+    const playPromise = this.el_.play();
+
+    // Catch/silence error when a pause interrupts a play request
+    // on browsers which return a promise
+    if (playPromise !== undefined) {
+      playPromise.catch((e) => {});
```
gkatsev: `catch` might be an issue on IE8.
Also, we should probably verify this is a promise before proceeding.

mister-ben: OK, reworked that.

test/unit/tech/html5.test.js
```diff
@@ -517,3 +517,20 @@ QUnit.test('Html5#reset calls Html5.resetMediaElement when called', function() {
 
   Html5.resetMediaElement = oldResetMedia;
 });
+
+QUnit.test('Exception in play promise should be caught', function() {
```
misteroneill: I think this test should be wrapped in a condition to check for `Promise` support. When it gets merged and runs in a browser that doesn't support `Promise` it will likely blow up.

gkatsev: we're running es6-shim on tests, so, Promise is always available.

misteroneill: Ah, yes. 👍

misteroneill:
In that case, LGTM

gkatsev:
LGTM

gkatsev:
This fixes https://github.com/videojs/video.js/issues/3474

PR #3525, opened 2016-08-12
Disable interaction with disabled clickable components

mister-ben:
## Description

`enable()` and `disable()` on clickable components is only cosmetic. "Disabled" implies the control should not be functional.
## Specific Changes proposed
- Remove event listeners on `disable()` and add back on `enable()`.
- Move adding listeners from constructor to `enable`
- Remove `tabindex` from disabled components and add `disabled` attribute to disabled buttons to prevent keyboard access.
## Requirements Checklist

### DIFF
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -146,6 +146,8 @@ class ClickableComponent extends Component {
   enable() {
     this.removeClass('vjs-disabled');
     this.el_.setAttribute('aria-disabled', 'false');
+    this.on('tap', this.handleClick);
+    this.on('click', this.handleClick);
     return this;
   }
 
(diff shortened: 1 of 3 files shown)


gkatsev:
There are two more listeners being set up in the constructor. Should we move all the listeners being added to `enable` and just call `enable` from the constructor?

gkatsev:
@OwenEdwards can you take a look at this?

LGTM, otherwise.

src/js/clickable-component.js
```diff
@@ -158,6 +164,13 @@ class ClickableComponent extends Component {
   disable() {
     this.addClass('vjs-disabled');
     this.el_.setAttribute('aria-disabled', 'true');
+    if (typeof this.tabIndex_ !== 'undefined') {
+      this.el_.removeAttribute('tabindex');
```
OwenEdwards: We had to enforce consistent use of `tabIndex` rather than `tabindex` throughout the code - it doesn't make any difference here, but it would be better for consistency.

OwenEdwards:
LGTM

Issue #3510, opened 2016-08-08
Audio track controls do work properly with video files including multiple audio tracks.


snyderizer:
## Description
In browsers that support the native AudioTracks API (recent versions of Safari and IE), using a video file that includes multiple audio tracks, the new video.js controls to select audio tracks do not allow you to switch back and forth between the audio tracks.

## Steps to reproduce

See example at http://jsbin.com/vuwoqujifo/edit?html,output

(This uses the video from Microsoft's sample http://samples.msdn.microsoft.com/Workshop/samples/media/audiotrackexample.htm
 linked from their documentation here https://msdn.microsoft.com/en-US/library/hh772483%28v=vs.85%29.aspx )

1. Play the video in browser that supports AudioTracks
2. Use the video.js audio track control to select a different audio track (e.g. "pl") 

### Expected
The audio track you select should be enabled, all others should be disabled.
View the example at http://samples.msdn.microsoft.com/Workshop/samples/media/audiotrackexample.htm with Safari or IE to see how the native player UI handles this.

### Actual
* Safari: The audio track you click does become enabled and the other tracks are disabled, but the controls make it appear that both tracks are selected, so it becomes impossible to switch from "en" to "pl" and then back to "pl"
* IE: The audio track you click becomes enabled, but the other tracks are not disabled, so multiple tracks are played simultaneously.  

### Error output
No errors

## Additional Information
Please include any additional information necessary here. Including the following:
#### browsers
Tested in browsers that support native AudioTracks API:
* Safari 9
* IE 11

gkatsev:
In the version of safari I'm running, I'm actually seeing the same behavior that you mention for IE, where as I enable other audio tracks, it keeps the previous track enabled.

snyderizer:
I had my Safaris confused, the behavior I described was for the developer preview "Release 10 (Safari 9.1.2, WebKit 11603.1.1)."
With the production version "Version 9.1.2 (11601.7.7)" is does keep the previous track enabled.

For my application it actually is preferable to be able to play multiple track simultaneously, but I need a way for the user to be able to deselect them.

PR #3538, opened 2016-08-16
disable the audio tracks that were not clicked on.  

snyderizer:
## Description
When an audio track menu item is clicked, the other audio tracks should be disabled. Potential fix for #3510. 

## Specific Changes proposed
In the AudioTrackMenuItem click handler, disable the tracks that were not clicked on, in addition to enabling the track that was clicked on.

## Requirements Checklist
  - [ ] Docs/guides updated

I did not see any existing unit tests for this class, did manual testing only.

(FWIW, `contrib feature submit` just printed "undefined" after I entered the description and didn't seem to create a pull request so I did this through Github.  Apologies if this is not formatted as you desire.)

### DIFF
--- a/src/js/control-bar/audio-track-controls/audio-track-menu-item.js
+++ b/src/js/control-bar/audio-track-controls/audio-track-menu-item.js
@@ -53,9 +53,7 @@ class AudioTrackMenuItem extends MenuItem {
     for (let i = 0; i < tracks.length; i++) {
       const track = tracks[i];
 
-      if (track === this.track) {
-        track.enabled = true;
-      }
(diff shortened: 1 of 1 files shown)


BrandonOCasey:
Thanks for the PR! The format looks fine from what I can tell. I think that `contrib` may be broken at the moment so that is why you are experiencing issues with it. We are working on a replacement. 

I think this looks good to me and through some of my own tests I can confirm that it works. LGTM

misteroneill:
Yeah, this LGTM. 

And the code is cleaner, IMHO.

gkatsev:
Merged to stable.

Issue #3536, opened 2016-08-16
Still broken in node


ljharb:
Even with https://github.com/videojs/video.js/pull/3502 applied locally, I get:
```sh
$PWD/node_modules/video.js/dist/video.js:19742
var BACKGROUND_SIZE_SUPPORTED = ('backgroundSize' in _globalDocument2['default'].createElement('video').style);
                                                                                 ^

TypeError: _globalDocument2.default.createElement is not a function
    at Object.140.global/document ($PWD/node_modules/video.js/dist/video.js:19742:82)
```

After trying to patch that locally, I get:
```sh
$PWD/node_modules/video.js/dist/video.js:15920
  supportsTextTracks = !!Html5.TEST_VID.textTracks;
                                       ^

TypeError: Cannot read property 'textTracks' of undefined
    at Function.Html5.supportsNativeTextTracks ($PWD/node_modules/video.js/dist/video.js:15920:40)
```

After patching that, I get:
```sh
$PWD/node_modules/video.js/dist/video.js:15940
  var supportsVideoTracks = !!Html5.TEST_VID.videoTracks;
                                            ^

TypeError: Cannot read property 'videoTracks' of undefined
    at Function.Html5.supportsNativeVideoTracks ($PWD/node_modules/video.js/dist/video.js:15940:45)
```

Could the project add a test that requires dist/video.js, in node, so as to ensure none of these continue?

gkatsev:
What specifically are you trying to with videojs in node?
I always find it weird that we get issues around `createElement` because we tend to use `global` which should return `min-document` for `document` in node and that is supposed to have a `createElement` method.
Also, the way that some of the internals of videojs work is that they expect to be run in a browser context and changing that is tricky and probably not worth the effort because they rely so much on things available in the browser.

gkatsev:
Oh, I just tried requiring global directly and it works fine but has issues if it's done through the current dist file in videojs. It should be fixed by #3445.

ljharb:
Our react components run in both the browser and server, but we use componentDidMount to only invoke videojs on the server. I need to be able to require videojs in both environments, and only have browser things invoke on the browser. In other words, I need to be able to require any module without side effects in any environment.

It's fine if videojs doesn't actually *work* in node - it just needs to be requireable without error.

gkatsev:
Ah, ok, making it requirable sounds fine to me.

gkatsev:
Fixed by #3540.

ljharb:
[...]

PR #3445, opened 2016-07-20
new build that outs source es6 files into es5 folder

gkatsev:
## Description
Use babel to compile the source into es5 so that packagers can use it properly.

## Specific Changes proposed
Update to latest version of babel, switch to babel-register, add grunt-babel, simply karma browserify config. use babel-preset-es2015-loose.
Make `main` point at the new `es5/video.js` file and include the `es5` build in the npm published folder. This probably should also be added to the bower build I guess?
There's still some issues around updating to babel 6 and using babel-preset-es2015-loose. Also, needs testing in IE8.
This will also address https://github.com/videojs/video.js/issues/2750 and related issues.

## Requirements Checklist
- [x] Fix babel 6 issues
- [x] Test in IE8
- [x] Is es2015-loose preset required? Is es2015 preset ok?
- [x] eliminate sourcemaps (fixes #3106)
- [x] verify that build process works on windows
- [ ] ~~Run babel through the test folder also for a simpler build~~
- [ ] ~~Make the build closer to the plugin generator build~~
- [ ] Reviewed by Core Contributors

The struck out ones can be done separately.

PR #3540, opened 2016-08-17
Make videojs requireable in node

gkatsev:
This is based on #3445 and makes videojs requirable in node, fixing #3536.

@ljharb the relevant commit is https://github.com/videojs/video.js/pull/3540/commits/ad27f77e8d763b124d7c6811d5e48069b90cbccd

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2937,7 +2937,7 @@ Player.prototype.options_ = {
     'textTrackSettings'
   ],
 
-  language: navigator.languages && navigator.languages[0] || navigator.userLanguage || navigator.language || 'en',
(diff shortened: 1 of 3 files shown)


ljharb:
Something else to consider, is making the "main" be an empty file (or a file that exports a function), and adding the "browser" field that points to the dist file - that way, a node require pulls the main, but a browserify/webpack/bundler will grab the "browser" version.

gkatsev:
@ljharb ah, interesting. Hadn't thought of that. Though, the changes here are fairly minimal, so, it's probably fine.

BrandonOCasey:
LGTM

PR #3559, opened 2016-08-22
fix some missing semi-colons, and other general scss issues

BrandonOCasey:
## Description
Found some rules that were using the same selector, that should be merged, found some properties that were missing semi-colons, and found one property that needed to be on two lines.

## Requirements Checklist

### DIFF
--- a/src/css/components/_fullscreen.scss
+++ b/src/css/components/_fullscreen.scss
@@ -1,8 +1,6 @@
 .video-js .vjs-fullscreen-control {
   cursor: pointer;
   @include flex(none);
-}
-.video-js .vjs-fullscreen-control {
   @extend .vjs-icon-fullscreen-enter;
 }
 // Switch to the exit icon when the player is in fullscreen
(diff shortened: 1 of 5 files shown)


src/css/components/_fullscreen.scss
```diff
@@ -1,8 +1,6 @@
 .video-js .vjs-fullscreen-control {
   cursor: pointer;
   @include flex(none);
-}
-.video-js .vjs-fullscreen-control {
```
gkatsev: does `sass` output change for this?
But yeah, makes sense to have this.

BrandonOCasey: let me do a quick diff, but the byte size was exactly the same

BrandonOCasey: after doing the diff there are no changes

gkatsev: Sounds good. I didn't imagine it would.

gkatsev:
Do we want to merge this in as a patch? If so, could you rebase this against stable? (you can change PR base branches in github now as well).

gkatsev:
LGTM

BrandonOCasey:
@gkatsev I don't think it matters where we do it, we can do it as a patch on stable if you want and I will make the changes

PR #3560, opened 2016-08-22
Fix MediaError

misteroneill:
We discovered that attempting to construct a video.js `MediaError` object would fail to properly copy a native `MediaError` object due to the `code` property being non-enumerable. Additionally, there were no tests for this module.

## Specific Changes proposed
- Fix construction of video.js `MediaError` objects from native `MediaError` objects by manually assigning the `code` property.
- Allow the video.js `MediaError` object to be redundantly constructed, such that if it is called with a video.js `MediaError` object, it simply returns the same object. This removes the need for two code forks found elsewhere in the code.

## Requirements Checklist

### DIFF
--- a/src/js/media-error.js
+++ b/src/js/media-error.js
@@ -6,16 +6,23 @@ import assign from 'object.assign';
 /*
  * Custom MediaError to mimic the HTML5 MediaError
  *
- * @param {Number} code The media error code
+ * @param {Number|String|Object|MediaError} value The media error code
  */
-let MediaError = function(code){
-  if (typeof code === 'number') {
-    this.code = code;
-  } else if (typeof code === 'string') {
+const MediaError = function(value) {
+  if (typeof value === 'number') {
+    this.code = value;
+  } else if (typeof value === 'string') {
     // default code is zero, so this is a custom error
-    this.message = code;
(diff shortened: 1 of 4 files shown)


src/js/media-error.js
```diff
@@ -6,22 +6,36 @@ import assign from 'object.assign';
...
     // default code is zero, so this is a custom error
-    this.message = code;
-  } else if (typeof code === 'object') {
-    assign(this, code);
+    this.message = value;
+  } else if (typeof value === 'object') {
+
+    // We assign the `code` property manually because native MediaError objects
+    // do not expose it as an own/enumerable property of the object.
+    if (typeof value.code === 'number') {
```
gkatsev: should we copy this before the `assign`?

gkatsev:
Can you rebase/change this PR against the stable branch?

gkatsev:
Two things, LGTM otherwise.

src/js/media-error.js
misteroneill: Hmm. Not sure there's a case where there would be a difference in behavior.

misteroneill:
Yeah, will rebase tomorrow morning.

PR #3562, opened 2016-08-23
Fix MediaError

misteroneill:
We discovered that attempting to construct a video.js `MediaError` object would fail to properly copy a native `MediaError` object due to the `code` property being non-enumerable. Additionally, there were no tests for this module.

_This is a re-opening of #3560 against `stable`._

## Specific Changes proposed
- Fix construction of video.js `MediaError` objects from native `MediaError` objects by manually assigning the `code` property.
- Allow the video.js `MediaError` object to be redundantly constructed, such that if it is called with a video.js `MediaError` object, it simply returns the same object. This removes the need for two code forks found elsewhere in the code.

## Requirements Checklist

### DIFF
--- a/src/js/media-error.js
+++ b/src/js/media-error.js
@@ -4,9 +4,16 @@
 import assign from 'object.assign';
 
 /*
- * Custom MediaError to mimic the HTML5 MediaError
+ * Custom MediaError class which mimics the standard HTML5 MediaError class.
  *
- * @param {Number|String|Object|MediaError} value The media error code
+ * @param {Number|String|Object|MediaError} value
+ *        This can be of multiple types:
+ *        - Number: should be a standard error code
+ *        - String: an error message (the code will be 0)
+ *        - Object: arbitrary properties
+ *        - MediaError (native): used to populate a video.js MediaError object
+ *        - MediaError (video.js): will return itself if it's already a
+ *          video.js MediaError object.
  */
 function MediaError(value) {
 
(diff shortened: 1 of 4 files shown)


gkatsev:
For future reference, you can now change the base of a PR without opening a new one: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

misteroneill:
Mind: blown! I've been wanting that feature for years.

src/js/media-error.js
```diff
@@ -4,24 +4,45 @@
...
+ *        - Object: arbitrary properties
+ *        - MediaError (native): used to populate a video.js MediaError object
+ *        - MediaError (video.js): will return itself if it's already a
+ *          video.js MediaError object.
  */
-let MediaError = function(code){
-  if (typeof code === 'number') {
-    this.code = code;
-  } else if (typeof code === 'string') {
+function MediaError(value) {
```
BrandonOCasey: should this use the es6 class syntax, and be a constructor?

BrandonOCasey:
LGTM other than my comment on es6 class syntax

src/js/media-error.js
misteroneill: I mean, it could, but I'd prefer not needing to go through another code review if it can be avoided.

test/unit/media-error.test.js
```diff
@@ -0,0 +1,65 @@
...
+ * Creates a real native MediaError object.
+ *
+ * @param  {Number} code
+ * @param  {String} [message]
+ * @return {MediaError}
+ */
+const createNativeMediaError = (code, message) => {
+  const err = Object.create(window.MediaError);
+
+  Object.defineProperty(err, 'code', {value: code});
```
gkatsev: this fails in IE8

gkatsev:
Merged to stable, closing.

PR #3570, opened 2016-08-24
Refactor TextTrackSettings

misteroneill:
## Description
The `TextTrackSettings` component has a lot of limitations: a large, hard-coded HTML string, repetitive code~~, and does not inherit from `ModalDialog` (which would provide accessibility benefits and more)~~.

This is the first in a series of 2 PRs. This one is focused on reducing repetitive code and making the component DRYer overall. The second PR will be focused on replacing the large HTML string with DOM generation methods. ~~And the third PR will be focused on inheriting from `ModalDialog` instead of `Component`.~~

**Update**: I think that inheriting from `ModalDialog` would be ideal, but it would also mean too much backward-incompatibility (and too much workaround code to achieve backward-compatibility). Let's reserve that for 6.0.

## Requirements Checklist


src/js/tracks/text-track-settings.js
```diff
@@ -321,17 +323,12 @@ class TextTrackSettings extends Component {
    * @method restoreSettings
    */
   restoreSettings() {
-    let err;
     let values;
 
     try {
-      [err, values] = safeParseTuple(window.localStorage.getItem('vjs-text-track-settings'));
```
gkatsev: I still think it's nice to use safeParseTuple. I think the reason why this is wrapped in a try/catch is to make sure that the call to localStorage doesn't break.

src/js/tracks/text-track-settings.js
```diff
@@ -1,12 +1,61 @@
...
+  fontPercent: {
+    selector: '.vjs-font-percent > select',
+    default: 2,
+    parser: (v) => v === '1.00' ? null : Number(v)
+  },
+  textOpacity: {
+    selector: '.vjs-text-opacity > select'
+  },
+  windowColor: {
+    selector: '.window-color > select'
```
gkatsev: `window-color` was always not prefixed by `vjs`? That should probably be fixed.

src/js/tracks/text-track-settings.js
```diff
@@ -236,83 +244,77 @@ class TextTrackSettings extends Component {
...
-      color: fgColor,
-      backgroundColor: bgColor
-    };
-
-    for (const name in result) {
-      if (result[name] === '' || result[name] === 'none' || (name === 'fontPercent' && result[name] === 1.00)) {
-        delete result[name];
+    let result = null;
+
+    iterateSelectConfigs((config, key) => {
```
gkatsev: maybe make a `reduceConfigs`?

src/js/tracks/text-track-settings.js
```diff
@@ -236,83 +244,77 @@ class TextTrackSettings extends Component {
...
-      fontPercent = fontPercent.toFixed(2);
-    }
+    iterateSelectConfigs((config, key) => {
+      const value = values[key];
+
+      if (!value) {
+        return;
+      }
 
-    setSelectedOption(this.$('.vjs-font-percent > select'), fontPercent);
```
gkatsev: why inline `setSelectedOption`? I think it makes this function a lot more readable.

gkatsev:
I guess the other parts of the proposal are still to come?

gkatsev:
Also, maybe make each part a separate PR?

src/js/tracks/text-track-settings.js
misteroneill: It seems like overkill to have two `try` blocks (even if one is hidden in a dependency). This way we only have one.

src/js/tracks/text-track-settings.js
misteroneill: Yeah. I was wondering about that. It probably should be. 👍

misteroneill:
Yeah, this part is going to be focused on DRYness.

src/js/tracks/text-track-settings.js
misteroneill: Fair enough.

src/js/tracks/text-track-settings.js
gkatsev: probably not necessary here but it allows us to more easily differentiate between the two errors.
Also, I wonder if safeJSONparse is used anywhere else.

misteroneill: I don't know that it does allow us to easily differentiate. Either way, you have an `Error` object that gets logged to the console. I suppose before it was sometimes logged as an error and sometimes as a warning, but... that doesn't seem like a significant benefit.

It's used in the `Player`. I dunno if it's worth including it at all since it's basically just a small wrapper around `try`/`catch`... but that's for another discussion/PR.

src/js/utils/obj.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Array-like iteration for objects.
+ *
+ * @param  {Object}   object
+ * @param  {Function} fn
+ *         A callback function which is called for each key in the object. It
+ *         receives the value and key as arguments.
+ */
+export function each(object, fn) {
```
BrandonOCasey: how much space would it take to bring in lodash here instead? (could provide some lodash code sharing between projects if it is not a lot more)

src/js/tracks/text-track-settings.js
```diff
@@ -236,83 +295,33 @@ class TextTrackSettings extends Component {
...
-   * .vjs-fg-color
-   * .vjs-text-opacity
-   * .vjs-bg-color
-   * .vjs-bg-opacity
-   * .window-color
-   * .vjs-window-opacity
+   * Sets text track settings from an object of values.
    *
-   * @param {Object} values Object with texttrack setting values
+   * @param {Object} values
```
BrandonOCasey: this should still have a description I think

src/js/utils/obj.js
misteroneill: I think it'd be a lot more. I'm willing to check to see how much more, but I suspect it's quite a bit (comparatively).

src/js/tracks/text-track-settings.js
misteroneill: Ah, yeah, good call. I think it was removed because at one point I was going to change it to an array with some backward compatibility for the object, but then went back to the object approach.

src/js/tracks/text-track-settings.js
```diff
@@ -58,7 +158,7 @@ function captionOptionsMenuTemplate(uniqueId, dialogLabelId, dialogDescriptionId
               </select>
             </span>
           </fieldset>
-          <fieldset class="window-color vjs-tracksetting">
+          <fieldset class="vjs-window-color vjs-tracksetting">
```
BrandonOCasey: will this still be backwards compatible (as it used to be called `window-color`)?

src/js/tracks/text-track-settings.js
```diff
@@ -175,38 +244,28 @@ class TextTrackSettings extends Component {
...
-      this.$('.window-color > select').selectedIndex = 0;
-      this.$('.vjs-text-opacity > select').selectedIndex = 0;
-      this.$('.vjs-bg-opacity > select').selectedIndex = 0;
-      this.$('.vjs-window-opacity > select').selectedIndex = 0;
-      this.$('.vjs-edge-style select').selectedIndex = 0;
-      this.$('.vjs-font-family select').selectedIndex = 0;
-      this.$('.vjs-font-percent select').selectedIndex = 2;
+    });
+
+    this.on(this.$('.vjs-default-button'), 'click', () => {
```
BrandonOCasey: can this be merged with the one above it since they both rely on `click` for `.vjs-default-button`?

src/js/tracks/text-track-settings.js
```diff
@@ -175,38 +244,28 @@ class TextTrackSettings extends Component {
...
-    Events.on(this.$('.window-color > select'), 'change', Fn.bind(this, this.updateDisplay));
-    Events.on(this.$('.vjs-text-opacity > select'), 'change', Fn.bind(this, this.updateDisplay));
-    Events.on(this.$('.vjs-bg-opacity > select'), 'change', Fn.bind(this, this.updateDisplay));
-    Events.on(this.$('.vjs-window-opacity > select'), 'change', Fn.bind(this, this.updateDisplay));
-    Events.on(this.$('.vjs-font-percent select'), 'change', Fn.bind(this, this.updateDisplay));
-    Events.on(this.$('.vjs-edge-style select'), 'change', Fn.bind(this, this.updateDisplay));
-    Events.on(this.$('.vjs-font-family select'), 'change', Fn.bind(this, this.updateDisplay));
+    });
+
+    Obj.each(selectConfigs, config => {
```
BrandonOCasey: 👍 so much better

src/js/tracks/text-track-settings.js
misteroneill: No, but we weren't using that selector and it could pretty easily be considered a bug because all our classes should be prefixed with `vjs-`.

src/js/tracks/text-track-settings.js
misteroneill: The previous one is for `.vjs-done-button`. 😄

BrandonOCasey: ah ok mis-read thanks!

BrandonOCasey:
* need to add a description to the mentioned doc block
* need to check how much size lodash will add gzipped
Other than that LGTM

src/js/tracks/text-track-settings.js
gkatsev: Also, the other benefit of safejsonparse is encapsulation. Yeah, not much help here but I think it does help in general. Supposedly, try/catch don't hurt much anymore in the latest and greatest js engines but older engines still have issues with deoping the entire function.

src/js/tracks/text-track-settings.js
gkatsev: If it breaks, we'll revert but it is a bug.

misteroneill:
Doc block is updated.

As for adding lodash, we already have lodash-compat in place. Using its methods - `each()` and `reduce()` - adds 2.5kb to the gzipped size. Removing lodash-compat and using lodash 4 instead adds 6.5kb. This is likely because lodash has multiple internal dependencies.

BrandonOCasey:
ok yeah maybe we can revisit it at a later time, when we have a better way to share dependencies between projects

src/js/tracks/text-track-settings.js
```diff
@@ -352,18 +358,18 @@ class TextTrackSettings extends Component {
     const values = this.getValues();
 
     try {
-      if (Object.getOwnPropertyNames(values).length > 0) {
-        window.localStorage.setItem('vjs-text-track-settings', JSON.stringify(values));
+      if (values) {
```
gkatsev: why remove the check for the size of values?

gkatsev: The default return value for `getValues` is an object which is truthy, so, we'll never enter the `else`

misteroneill: Good catch. That was from some earlier work where `getValues()` was not always returning an object.

BrandonOCasey:
LGTM

gkatsev:
@owenedwards can you take a look? Probably not much changes that matter towards accessibility in this PR.

OwenEdwards:
I don't see any changes that impact a11y. LGTM.

(Hoping we get rid of that chunk of HTML for the dialog at some point soon, too ;-) )

misteroneill:
@OwenEdwards Yep! That's the next PR!

misteroneill:
Going to close this and re-open against master when the other branch gets included.

Issue #3428, opened 2016-07-11
Dash sources not loading on 5.10.2+


RavWar:
I've stumbled upon an issue in videojs 5.10.2+ and dash. The dash sources are not loading most of the times.

It actually depends on how fast the manifest is loaded i think. Small and quick manifests load up normally much frequently than the big or slow ones. Once network throttling is set to slow (3G for example) then the small start to fail as well.

Here is the test case: http://jsbin.com/yesexuxeqe/edit?html,console,output
Note: if videojs version 5.10.2 is replaced with 5.9.2, it works okay every time.

After some investigation i've managed to pinpoint the error to this pr: https://github.com/videojs/video.js/pull/3285
Here is a normal startup flow (i've added additional log entries for loadstart events and firstLoadStartListener, successiveLoadStartListener functions):

![normal start](https://cloud.githubusercontent.com/assets/675182/16731313/1111e7be-4780-11e6-8a08-970cc01531de.png)

Here is a broken one:

![bad start](https://cloud.githubusercontent.com/assets/675182/16731314/11150dc2-4780-11e6-8a77-7c0c4ff1e6f8.png)

Mentioned pr introduced a feature to dispose source handlers on repetitive loadstart events. And in case of slow dash manifest - loadstart is actually fired twice which causes source handler to be disposed before it can even initiate the playback.

I have not had time to dig up the cause of repetitive loadstart event yet. Maybe someone can shed some light on why's that happening?

RavWar:
Also, i'm pretty sure this issue https://github.com/videojs/videojs-contrib-dash/issues/90 has the same cause

gkatsev:
Interesting. Thanks for digging in and finding the root cause. It does sound like the repetitive loadstarts are the issue here. From what I understand, the loadstart event should only ever be fired once. Might be worth also opening up an issue against dashjs itself with this.

richardbushell:
Ditto, we've been using videojs with videojs-contrib-dash for quite a while and also had to revert to vjs 5.9 after trying 5.10 at the weekend to avoid the Media Source error when trying to load the dash source.

richardbushell:
OK, here is the reply from dash.js:
https://github.com/Dash-Industry-Forum/dash.js/issues/1491#issuecomment-231959860
[...]

PR #3571, opened 2016-08-24
only reset currentSource_ on second loadstart if it has changed

BrandonOCasey:
## Description

Previously when a second `loadstart` we would assume that a new source is forthcoming and reset everything. This is not the case and there is currently an issue (#3428) where a second `loadstart` is being triggered on the same source and causing issues. Video.js should be able to handle the second `loadstart` for the same source even though it is not expected.
## Specific Changes proposed
- Keep track of the previousSource on each `loadstart`
- if  `loadStart` is triggered twice or more for the same source, ignore it
- update unit tests to reflect new functionality 
## Requirements Checklist

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -851,13 +851,23 @@ Tech.withSourceHandlers = function(_Tech) {
     return this;
   };
 
+  let previousSource;
+
   // On the first loadstart after setSource
   _Tech.prototype.firstLoadStartListener_ = function() {
+    previousSource = this.currentSource_;
     this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
   };
 
   // On successive loadstarts when setSource has not been called again
   _Tech.prototype.successiveLoadStartListener_ = function() {
+
+    // if the second loadstart is for the same source
+    // it is an invalid loadstart, set this event to trigger again
+    if (previousSource.src === this.currentSource_.src) {
+      return this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
+    }
+    previousSource = this.currentSource_;
     this.currentSource_ = null;
     this.disposeSourceHandler();
     this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -851,11 +851,9 @@ Tech.withSourceHandlers = function(_Tech) {
     return this;
   };
 
-  let previousSource;
-
   // On the first loadstart after setSource
   _Tech.prototype.firstLoadStartListener_ = function() {
-    previousSource = this.currentSource_;
+    this.previousSource_ = this.currentSource_;
     this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
   };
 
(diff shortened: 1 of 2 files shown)


src/js/tech/tech.js
```diff
@@ -851,13 +851,23 @@ Tech.withSourceHandlers = function(_Tech) {
...
+    previousSource = this.currentSource_;
     this.one(this.el_, 'loadstart', _Tech.prototype.successiveLoadStartListener_);
   };
 
   // On successive loadstarts when setSource has not been called again
   _Tech.prototype.successiveLoadStartListener_ = function() {
+
+    // if the second loadstart is for the same source
+    // it is an invalid loadstart, set this event to trigger again
+    if (previousSource.src === this.currentSource_.src) {
```
gkatsev: wouldn't this always be true, though? `this.currentSource_` only ever changes when `player.src()` is called and a source handler is set.

src/js/tech/tech.js
```diff
@@ -851,13 +851,23 @@ Tech.withSourceHandlers = function(_Tech) {
     return this;
   };
 
+  let previousSource;
```
gkatsev: this variable could end up being shared between Techs and players on a page.

BrandonOCasey: do we want it to be `this.previousSource_` then?

src/js/tech/tech.js
BrandonOCasey: I think that there are two scenarios, the first of which was being checked for already:

Scenario 1:
1. setSource is called
2. `loadstart` happens (`previousSource` will be set to `this.currentSource_` for thes first time)
3. currentSource_ is changed
4. `loadstart` happens again (`previousSource` will not equal `this.currentSource_` and `previousSource` will be set to `this.currentSource_` before it is set to null)

Scenario 2:
1. setSource is called
2. `loadstart` happens (`previousSource` will be set to `this.currentSource_` for thes first time)
3. `loadstart` happens again (`previousSource` is still set to `this.currentSource_` do nothing

src/js/tech/tech.js
misteroneill: Yeah, I think that's preferable.

src/js/tech/tech.js
gkatsev: Wouldn't we just want to compare `this.currentSource_` with `this.el_.currentSrc`?

misteroneill:
This seems like a reasonable change. I'd like to see if we could verify that it fixes the Dash.js issue.

src/js/tech/tech.js
BrandonOCasey: this.el.currentSrc will be a blob in some cases (dashjs being one of them)

BrandonOCasey:
Ok so there does not seem to be a good way to store the blob on `this.currentSource_` as suggested by @gkatsev. I think our best option for now is still to use `this.previousSource_`. I think that they will only match when this is a `loadstart` for the same source

gkatsev:
I don't think this covers the case where you a new source is put into the video element directly.
The `currentSource_` doesn't change then and the `previousSource` is set to the same `currentSource_`

BrandonOCasey:
could we use a https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver to watch for changes in the source and set currentSource appropriately?

misteroneill:
@gkatsev Wouldn't that be the case already? For example, if someone calls `setSource` which triggers the first loadstart then they manually change the source in the DOM, it'll trigger a new loadstart and fire the `_successiveLoadStartListener`, which will blow away the source handler?

This is kind of what the videojs-per-source-behaviors plugin tries to abstract away, but it'd be nice if video.js had some kind of comprehensive solution to this. I'm not sure it's possible, though, but `MutationObserver` is an interesting idea.

ruslandinov:
Hey guys, sorry for bothering you, but we also encounter this problem and would appreciate a fix. Is there any chance for the PR to be merged soon?

misteroneill:
@ruslandinov It's something that's been discussed, but we don't think we can get anything in until later this year for Video.js 6.0 because it would likely be a breaking change.

gkatsev:
I'm going to close this because I think we're going to try going a different route. We can always re-open later if necessary.

PR #3580, opened 2016-08-25
Exceptions in event handlers breaks further processing of event handlers

imbcmdth:
## Description
In the event dispatch function, we do not guard against event handlers raising exceptions. If that happens, any further processing of event handlers and *any code after the trigger* stops as the exception continues to unwind the stack.

This is particularly problematic if the event was triggered by a source handler during the initialization stages of the handler. The SourceHandler can end up in a state from which it can not recover through no fault of our own. 

See this example of a player broken by a simple exception thrown from an event handler:

http://jsbin.com/fipikerore/edit?html,output

## Specific Changes proposed
Simply add a try/catch around each invocation of an event handler function called from the the `dispatch` function. 

## Requirements Checklist
  - [ ] Docs/guides updated
  - [x] Example created

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -7,8 +7,9 @@
  * robust as jquery's, so there's probably some differences.
  */
 
-import  * as Dom from './dom.js';
-import  * as Guid from './guid.js';
+import * as Dom from './dom.js';
+import * as Guid from './guid.js';
+import log from './log.js';
 import window from 'global/window';
 import document from 'global/document';
 
@@ -57,7 +58,11 @@ export function on(elem, type, fn){
           if (event.isImmediatePropagationStopped()) {
             break;
           } else {
-            handlersCopy[m].call(elem, event, hash);
+            try {
+              handlersCopy[m].call(elem, event, hash);
+            } catch (e) {
+              log.error(e);
+            }
           }
         }
       }
--- a/test/unit/events.test.js
+++ b/test/unit/events.test.js
@@ -239,3 +239,21 @@ test('should have relatedTarget correctly set on the event', function() {
 
   Events.trigger(el2, { type:'click', relatedTarget:undefined });
 });
+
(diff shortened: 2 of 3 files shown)


src/js/utils/events.js
```diff
@@ -248,7 +249,11 @@ export function on(elem, type, fn) {
           if (event.isImmediatePropagationStopped()) {
             break;
           } else {
-            handlersCopy[m].call(elem, event, hash);
+            try {
```
gkatsev: this try/catch should probably be wrapped in an IIFE because try/catch

imbcmdth: That would make the code a lot uglier than it already is for little to no benefit. There should *never* be so many event handlers for any single event that the performance difference becomes an issue.

dmlap: I lean towards leaving this as-is.

test/unit/events.test.js
```diff
@@ -239,3 +239,21 @@ test('should have relatedTarget correctly set on the event', function() {
 
   Events.trigger(el2, { type:'click', relatedTarget:undefined });
 });
+
+QUnit.test('should execute remaining handlers after an exception in an event handler', function(assert) {
+  assert.expect(1);
+
+  const el = document.createElement('div');
+  const listener1 = function() {
+    throw new Error('GURU MEDITATION ERROR');
```
dmlap: Sweet.

test/unit/events.test.js
```diff
@@ -239,3 +239,21 @@ test('should have relatedTarget correctly set on the event', function() {
...
+  };
+  const listener2 = function() {
+    assert.ok(true, 'Click Triggered');
```
dmlap: I am not a fan of assertions in callbacks since it breaks reading order but that's totally a nitpick and not worth holding up the PR for.

dmlap:
LGTM

PR #3586, opened 2016-08-29
Disable HLS hack on Firefox for Android (with tests)

marco-c:
#3498 + tests.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1319,7 +1319,7 @@ const mp4RE = /^video\/mp4/i;
 
 Html5.patchCanPlayType = function() {
   // Android 4.0 and above can play HLS to some extent but it reports being unable to do so
-  if (browser.ANDROID_VERSION >= 4.0) {
+  if (browser.ANDROID_VERSION >= 4.0 && !browser.IS_FIREFOX) {
     if (!canPlayType) {
       canPlayType = Html5.TEST_VID.constructor.prototype.canPlayType;
     }
--- a/test/unit/tech/html5.test.js
+++ b/test/unit/tech/html5.test.js
@@ -137,6 +137,28 @@ QUnit.test('patchCanPlayType patches canplaytype with our function, conditionall
   Html5.unpatchCanPlayType();
 });
 
+QUnit.test('patchCanPlayType doesn\'t patch canplaytype with our function in Firefox for Android', function(assert) {
+  // the patch runs automatically so we need to first unpatch
+  Html5.unpatchCanPlayType();
+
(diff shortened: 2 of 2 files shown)


test/unit/tech/html5.test.js
```diff
@@ -131,14 +133,39 @@ QUnit.test('patchCanPlayType patches canplaytype with our function, conditionall
...
+  const video = document.createElement('video');
+  const canPlayType = Html5.TEST_VID.constructor.prototype.canPlayType;
+
+  browser.ANDROID_VERSION = 4.0;
+  browser.IS_FIREFOX = true;
+  Html5.patchCanPlayType();
+
+  assert.strictEqual(video.canPlayType,
+                 canPlayType,
+                 'original canPlayType and patched canPlayType should not be equal');
```
misteroneill: Looks like the assertion message needs to be updated.

marco-c: Leftover from copy & paste of the other test, fixed in b762008d22b1424e2b16140d6d98e7465b51d1fd.

misteroneill:
Thanks for this, @marco-c!

LGTM

gkatsev:
LGTM. We should merge this in as two separate commits when we get around to it.

src/js/tech/html5.js
```diff
@@ -1319,7 +1319,7 @@ const mp4RE = /^video\/mp4/i;
 
 Html5.patchCanPlayType = function() {
   // Android 4.0 and above can play HLS to some extent but it reports being unable to do so
-  if (browser.ANDROID_VERSION >= 4.0) {
+  if (browser.ANDROID_VERSION >= 4.0 && !browser.IS_FIREFOX) {
```
gkatsev: I just saw an update to the firefox beta which said they're getting HLS support. Is this something that we should possibly account for? Maybe check if firefox version is less than 50?

marco-c: The support in Firefox 50 is not polished, there are a lot of bugs (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1301050). It will probably be disabled in 50.
We're using a hacky solution, so it will never be like a properly supported codec.
I would leave it as is and expect Firefox not to return wrong values in `canPlayType`, if it will ever support HLS natively, it will be returned in `canPlayType`.

gkatsev: I guess we could also always update this at a later time once/if it officially ships.

curtisgibby:
Any idea when this will land in the core JS that Brightcove serves with its players? (e.g. http://players.brightcove.net/659677170001/default_default/index.js) Right now, that file references VideoJS version 5.11.7.

Issue #3606, opened 2016-09-08
Player not showing only with Android/Chrome and 3G/4G connections


Joan-Rojo1:
Hi there, 

I am facing an extremely rare issue. The player is not being displayed when using Android/Chrome and 3G/4G connections, with Wifi everything is ok. I have tried to debug it, but I am not getting any message on console. Here it is an example URL:

http://prem.publico.es/publicotv/577486/la-ola-de-incendios-en-galicia-hace-sospechar-que-son-intencionados

I dont know if any has faced this problem, but every help would be appreciated.

Thanks in advance,

Note: we use video.js 5.9.2 we use plugins for ads such as video.ima.js and video.ads.js, for example.

mister-ben:
Can you reduce that down to a minimal test case?

Joan-Rojo1:
Hi Ben,

Even you don't have to use your cell and mobile data to reproduce it. It's enough to use Chrome with developer tools and set throttling to any XG connection vs Wifi to see the difference. I can't find any difference in DOM. Here are 2 captures reproducing the problem:

http://oi66.tinypic.com/2wgf8n6.jpg

http://oi68.tinypic.com/apgzg6.jpg

Thanks a lot for answering,

misteroneill:
I tried all the way down to "Good 2G" and I saw a video player to appear and playback start. Granted, I had to wait almost 2 minutes for the ~700 requests and 5MB of data to finish. 😆 

The reason Ben asked about a reduced test case is that without one there is simply too much going on here to narrow a problem down to video.js specifically. There are too many factors at play that could be affecting your player to say that there's a bug in video.js.

BrandonOCasey:
here is the starter template for generating a reduced test case http://jsbin.com/axedog/edit?html,output

Joan-Rojo1:
Hi there, thanks a lot for your time and test. I will prepare a minimal test as you suggest and let you know next week. Have a nice weekend :)

mister-ben:
I believe the issue here is using fluid mode with preload disabled, when no other size information is available. The aspect ratio can't be detected and the intended default 16:9 isn't being used.

You're seeing this only on cellular because the browser forces preload to none on a cellular connection, but allows preload on wifi.

I think I have a fix, but you can work around it in the meantime by setting a width on the player, e.g. `width="100%"` or including an aspect ratio in the settings `{"aspectRatio": "16:9"}`.

Joan-Rojo1:
Hi Ben,

[...]

PR #3614, opened 2016-09-14
Use default aspect ratio for fluid players if width unknown

mister-ben:
## Description

If a player is fluid and does not have a width set, and preload is set to none, the height of the player is zero. This includes where preload is forced to none by mobile Chrome as in #3606.
## Specific Changes proposed
- If the player has the `.vjs-fluid` class when initialised, `fluid` is set to `true`, so adding the class behaves the same as `{fluid: true}` in the setup options.
- The `fluid(bool)` setter calls `player.updateStyleEl_()`. Otherwise it won't be triggered in `createEl()` if an aspect ratio is not also set.
- Corrects the test for a set `videoWidth()` in `updateStyleEl_()` - `videoWidth()` returns `0` if the width is unknown. This allows the default `16:9` to kick in rather than using `0:0`. 
## Requirements Checklist
- [x] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -410,8 +410,6 @@ class Player extends Component {
     this.height(this.options_.height);
     this.fluid(this.options_.fluid);
     this.aspectRatio(this.options_.aspectRatio);
-    // Style el has not updated yet if width, heigth and aspect ratio are undefined
-    // this.updateStyleEl_();
 
     // Hide any links within the video/audio tag, because IE doesn't hide them completely.
     const links = tag.getElementsByTagName('a');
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -2848,6 +2850,10 @@ class Player extends Component {
     const tagOptions = Dom.getElAttributes(tag);
     const dataSetup = tagOptions['data-setup'];
 
+    if (Dom.hasElClass(tag, 'vjs-fluid')) {
```
gkatsev: should we add a test that verifies that if `vjs-fluid` class is available then `player.fluid()` returns `true`?

gkatsev:
Sounds like a great idea.

src/js/player.js
mister-ben: Yes that makes sense, added.

gkatsev:
Trying out the new Projects and reviewing for this PR.

PR #3472, opened 2016-07-27
Refactoring chapters button handling and fixing several issues

cervengoc:
## Description
- Refactored `ChaptersButton`, broke logic into several methods.
- Fixed the issues in #3447 about in some browsers tracks have an empty `cues` array instead of null. Now we always subscribe to `load` event, and force an `update`. Also, track changes are handled, so chapters track can now be changed at runtime.
- Fixed the issue in #3447 about chapters menu items are not selectable. Now automatic update of the selected item based on player time works fine.
- Implemented the usage of the chapters track's `label` attribute as menu title, if it's present. If not, we fall back to the localized "Chapters" title.
- Refined the menu styling, so that `vjs-menu-title` telement won't get the `hover` effect, It would confuse users, because they might believe that the title item is a clickable item too.
## Testing

~~No unit tests for the functionality so far, because~~
- ~~there were no unit tests before either~~
- ~~I'm a new contributor and don't know the technical test environment~~

I've added some unit tests too.

Manually tested in IE 11, Chrome 51.

Issue #3447, opened 2016-07-20
Issues with chapters


cervengoc:
Hello guys, I had problems with chapters. I don't have time for the PR, but here are a list of issues which I've found and fixed locally.
- ChaptersButton listens to `load` event only when `cues == null`. In latest Chrome I have an empty `cues` property, so it seems like several browsers handle it differently. IMO the best would be to **always** subscribe to the `load` event and issue an `update`. Also, it would be needed to unsubscribe from it when player is disposed; when track is removed; etc. As a quick fix I refined the related `if` condition to `!chaptersTrack.cues || !chaptersTrack.cues.length`.
- ChaptersMenuItem should be set to `selectable`, otherwise it won't update it's selected style on `cuechange`.
- Chapters menu should take the chapters `track`'s `label` as menu title and fall back to "Chapters" if not provided.
- This is just a refactoring: it seems like ChaptersButton's `createItems` is incorretct, and not even called (`createMenu` is used instead), so it should be removed.
- There are several styling isssues with chapters: font is not `Arial` because it inherits `VideoJS` font, so menu items and title look ugly; menu items have strange padding, the text is not aligned to the middle vertically.

Please someone handle these in the next release.

gkatsev:
Thanks for the issue. We'll get around to it when we case. A PR is always greatly appreciated.
Some comments about each item below:

> ChaptersButton listens to load event only when cues == null
It probably makes sense to always listen to `load`. When this was implemented, `cues` was always `null` before it was loaded but it's possible that things have changed by now.

> ChaptersMenuItem should be set to selectable
Not sure that making the ChaptersMenuItem `selectable` is correct. Because we don't update update the "selected" item depending on the player's current time, I don't think it really makes sense for them to be `selectable`. Would the selected item but just the previously clicked item?
There definitely should be some visual cue that you've which item you've clicked on, though.

> Chapters menu should take the chapters track's label as menu title and fall back to "Chapters" 
That makes sense as an enhancement.

> it seems like ChaptersButton's createItems is incorrect
Yup, `createMenu` probably needs to be broken up and a large portion moved into `createItems`.

> font is not Arial because it inherits VideoJS font
The font-family was recently updated.

> menu items have strange padding
Not sure what these means.

> text is not aligned to the middle vertically
It looks aligned to me, but very possible it isn't perfect. Vertical alignment is hard in CSS.

cervengoc:
[...]

PR #3623, opened 2016-09-19
fix(chapters-button): Remove dead code and fix `selected` capability.

chemoish:
## Description

Fixes https://github.com/videojs/video.js/pull/3033, `selectable` with minimal code change. This doesn't account for any code style or implementation for leveraging `ChaptersTrackMenuItem` or extension of `TextTrackMenuItem`.

### DIFF
--- a/src/js/control-bar/text-track-controls/chapters-button.js
+++ b/src/js/control-bar/text-track-controls/chapters-button.js
@@ -3,7 +3,6 @@
  */
 import TextTrackButton from './text-track-button.js';
 import Component from '../../component.js';
-import TextTrackMenuItem from './text-track-menu-item.js';
 import ChaptersTrackMenuItem from './chapters-track-menu-item.js';
 import Menu from '../../menu/menu.js';
 import * as Dom from '../../utils/dom.js';
@@ -35,31 +34,6 @@ class ChaptersButton extends TextTrackButton {
    */
   buildCSSClass() {
     return `vjs-chapters-button ${super.buildCSSClass()}`;
-  }
-
-  /**
-   * Create a menu item for each text track
-   *
-   * @return {Array} Array of menu items
-   * @method createItems
-   */
-  createItems() {
-    const items = [];
-    const tracks = this.player_.textTracks();
-
-    if (!tracks) {
-      return items;
-    }
-
-    for (let i = 0; i < tracks.length; i++) {
-      const track = tracks[i];
-
-      if (track.kind === this.kind_) {
-        items.push(new TextTrackMenuItem(this.player_, {track}));
-      }
-    }
-
-    return items;
   }
 
   /**
@@ -124,7 +98,8 @@ class ChaptersButton extends TextTrackButton {
 
         const mi = new ChaptersTrackMenuItem(this.player_, {
           cue,
-          track: chaptersTrack
+          track: chaptersTrack,
+          selectable: true
         });
 
         items.push(mi);


src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -124,7 +98,8 @@ class ChaptersButton extends TextTrackButton {
 
         const mi = new ChaptersTrackMenuItem(this.player_, {
           cue,
-          track: chaptersTrack
+          track: chaptersTrack,
+          selectable: true
```
chemoish: Exact implementation of https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/text-track-button.js#L49.

src/js/control-bar/text-track-controls/chapters-button.js
```diff
@@ -38,31 +37,6 @@ class ChaptersButton extends TextTrackButton {
   }
 
   /**
-   * Create a menu item for each text track
-   *
-   * @return {Array} Array of menu items
-   * @method createItems
-   */
-  createItems() {
```
chemoish: Dead code never called. Item creation was all placed inside `createMenu` for some reason.

chemoish:
Partially, duplicative of https://github.com/videojs/video.js/pull/3472.

However, I think leveraging `TextTrackMenuItem` is probably better than forking implementation if possible.

src/js/control-bar/text-track-controls/chapters-button.js
OwenEdwards: Perhaps "selectable" wasn't a clear way to describe this, but I don't think that chapters can be "selected". Think of selectable as being whether the item behaves like a checkbox or radio button - would you want it "checked" in the menu? For captions and descriptions, yes, you would. For the captions settings menu, you wouldn't.

Having a chapter "selected" only makes sense if you have a mechanism which updates the current chapter as the video plays; otherwise it would be like leaving a checkmark on the progress bar of where you last clicked, even when playback has moved forward from that point.

gkatsev: They actually are supposed to get updated as mentioned in https://github.com/videojs/video.js/pull/3472 and https://github.com/videojs/video.js/issues/3447.

OwenEdwards: Okay, then we need to discuss the concept of "selected" some more. I see that chapter menu items will update as "selected" based on the timeline, but from an ARIA perspective, it's not strictly "selected". Again, think of it as a checkbox - if a checkbox is already checked, and you "click" on it, you don't expect the checkbox to stay checked. If it's a radio button you do, but then you don't expect it to have another effect. So this Chapters menu is a hybrid control, which is both a user control and a progress display. I understand what @chemoish  is trying to achieve visually, but it's not doing the correct thing logically, so just making it "selectable" isn't the correct approach. (Having said that, using the `controlText` to indicate `selected` to a screen reader was a hack too! Let's fix this "properly" before we go too far down the road to fix it at all).

gkatsev: Sounds like a plan. We should either see if we can get #3472 updated since it has a bunch of other good fixes or merge in #3472 and then fix it forward in a separate PR before releasing.

gkatsev:
Yeah, I think I'm going to close this now as this is available as part of #3472 as well, though, we still want more updates there. Please review that PR if you get the chance.

PR #3624, opened 2016-09-19
If we create a new element for movingMediaElInDOM, set playerId on the el again

### LINKED ISSUE #3283: issue to play video on ios with videojs API
I have some trouble to play a video with videojs API on ios 9.2. It work perfectly on desktop.

### Steps to reproduce :
1. run the source code on an ipad or an iphone
2. click on "action" button 

### Results :
loading animation appear and after black screen, the video don't playing.

### source code :
    <video id="my-video" class="video-js" controls="" data-setup="{}" width="640" height="264" >
        <source src="medias/videos/film_1080p.mp4" type='video/mp4'>
    </video>

    <button id="action">action</button>

    <script src="http://vjs.zencdn.net/5.9.2/video.js"></script>
    <script type="text/javascript">
        
        var video = videojs( 'my-video');

        document.getElementById('action').onclick = function(){
            video.play();
        }

    </script>

### versions

Video.js 5.9.2
ipad air on ios 9.2
iphone 6 on ios 9.3.1

gkatsev:
Fixes #3283. When `videojs` is called with an element, we try to see if we already have a player for that element. In some cases, like if `data-setup` is used with a manual call to `videojs`, we get called multiple times with the same element. On iOS, we cannot move the media element inside the DOM, so, we recreate it. In the process, we lose a marker we use for knowing whether we've created a player for that element.
This makes sure we keep that marker if we re-create the element.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -190,6 +190,8 @@ class Html5 extends Tech {
           })
         );
       }
+
+      el.playerId = this.options_.playerId;
     }
 
     // Update specific tag settings, in case they were overridden


gkatsev:
This should probably be rebased against stable and released as a patch along with the other items we have there.

misteroneill:
I noticed this was updating the Html5 tech. Is there any reason/need to update the Flash tech or is this isolated to Html5?

gkatsev:
It should only be an issue for when we delete and re-create the mediaEl. The original tag already gets the `playerId` set by default: https://github.com/videojs/video.js/blob/master/src/js/player.js#L321

misteroneill:
Sounds good. LGTM

gkatsev:
Closing in favor of https://github.com/videojs/video.js/pull/3648

PR #3629, opened 2016-09-20
remove es2015-loose since it's an option for es2015

gkatsev:
## Description
Followup from #3609. es2015-loose is deprecated because it's available as an option on the es2015 preset.


gkatsev:
Not sure. I'll take a look.

BrandonOCasey:
well the unit test failure is definitely odd...

gkatsev:
The test is broken because classList in firefox no longer throws an error.

PR #3634, opened 2016-09-22
grunt-contrib-uglify to 0.11

### LINKED ISSUE #3631: Contrib-uglify needs to be updated 0.11 to keep uglify in version 2.6
## Description

Uglify 2.7 introduced a [_breaking_ change](https://github.com/mishoo/UglifyJS2/pull/1179#issuecomment-248803837) with regards to IE8 support.
This means that our minified files that were released after 2.7 was released did not work in IE8.

contirb-uglify 0.11 switches the uglify dependency from being installed with a `^` to a `~` so it will never pick up 2.7.

misteroneill:
## Description
Fixes #3631 

Uglify 2.7 introduced a breaking change. This means our minified files after that point did not work in IE8. grunt-contrib-uglify released 0.11 which switches the uglify dependency from ^ to ~ and locks it down to the 2.6.x line, avoiding this implicit breakage.

## Requirements Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "grunt-contrib-connect": "~0.7.1",
     "grunt-contrib-copy": "^0.8.0",
     "grunt-contrib-cssmin": "~0.6.0",
-    "grunt-contrib-uglify": "^0.8.0",
+    "grunt-contrib-uglify": "~0.11.0",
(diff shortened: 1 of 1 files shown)


gkatsev:
Merged to stable.

PR #3637, opened 2016-09-23
update reportUserActivity

cucygh:
## Description

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

pass event to handleActivity and then we can use it in reportUserActivity method

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2432,12 +2432,12 @@ class Player extends Component {
       if (e.screenX !== lastMoveX || e.screenY !== lastMoveY) {
         lastMoveX = e.screenX;
         lastMoveY = e.screenY;
-        handleActivity();
+        handleActivity(e);
       }
     };
 
-    const handleMouseDown = function() {
-      handleActivity();
+    const handleMouseDown = function(e) {
+      handleActivity(e);
       // For as long as the they are touching the device or have their mouse down,
       // we consider them active even if they're not moving their finger or mouse.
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks for the PR; however, I'm not sure we actually need this change. `reportUserActivity` just sets a boolean and is already called in several places without any arguments. I think what we actually want is removing `event` from it's method signature as it is never used.

cucygh:
yeah,In fact, it is necessary for example different business use the same player and we need get all user‘s interaction such as ’click‘，if pass event arguments to reportUserActivity then we can make a plugin to override the API for different implementation of user interaction collecting without the player change

gkatsev:
I'm not really seeing the usecase and I cannot recommend [overriding that method](https://www.nczonline.net/blog/2010/03/02/maintainable-javascript-dont-modify-objects-you-down-own/).

Are the `useractive` and `userinactive` events not enough?

I'm going to close this PR for now. But please reply with your usecase and we can talk about whether it's something we want to support or not or maybe there's a better way of doing what you want to do. The PR can always be re-opened later if we feel like we want to bring it in.

cucygh:
the usecase is I'd like to know how much times the playButton is clicked,how much the VolumeSlide is cliked and how much times otherButton is clicked,at last send the info to server.

i see the code about useractive,it's simply true or false,i can't get the element directly and the handleUserActive_ method isn't a implementation

PR #3639, opened 2016-09-23
implement lifecycle hooks and trigger pre/post setup hooks

BrandonOCasey:
## Description

There is no current way to hook into the setup that video.js does, and this functionality would be useful for many plugins to use. (they can even initialize themselves with it)
## Specific Changes proposed
- Add several "hook" functions to the videojs object
- Add pre/post setup lifecycle event triggers
## Requirements Checklist

### DIFF
--- a/docs/guides/hooks.md
+++ b/docs/guides/hooks.md
@@ -1,66 +1,100 @@
 # Hooks
-Hooks exist so that users can latch on to certain Video.js lifecycle events
+Hooks exist so that users can "hook" on to certain video.js player lifecycle
 
 
 ## Current Hooks
-Currently the following lifecycle events are accept and use hooks
+Currently, the following hooks are avialable:
 
 ### beforesetup
-Called just before the player is created, so that the user can modify the options passed to the player or modify the video element that will be used for the player
-An example of this hook:
+`beforesetup` is called just before the player is created. This allows:
+* modification of the options passed to the video.js function (`videojs('some-id, options)`)
+* modification of the dom video element that will be used for the player
 
+`beforesetup` hook functions should:
+* take two arguments
+    1. videoEl: dom video element that video.js is going to use to create a player
+    2. options: options that video.js was intialized with and will later pass to the player during creation
+* return options that will merge and override options that video.js with intialized with
+
+Example: adding beforesetup hook
 ```js
     var beforeSetup = function(videoEl, options) {
+        // videoEl.id will be some-id here, since that is what video.js
+        // was created with
+
         videoEl.className += ' some-super-class';
 
-        // never autoplay because it makes me mad!
-        options.autoplay = false
+        // autoplay will be true here, since we passed in as such
+        (options.autoplay) {
+            options.autoplay = false
+        }
 
         // options that are returned here will be merged with old options
-        // automatically
+        // in this example options will now be
+        // {autoplay: false, controls: true}
         return options;
     };
+
+    videojs.hook('beforesetup', beforeSetup);
+    videojs('some-id', {autoplay: true, controls: true});
 ```
 
 ### setup
-Called just after the player is created, so that plugins or custom functionality can be initialized
+`setup` is called just after the player is created. This allows:
+* plugin or custom functionalify to intialize on the player
+* changes to the player object itself
 
+`setup` hook functions:
+* Take one argument
+    * player: the player that video.js created
+* Don't have to return anything
+
+Example: adding setup hook
 ```js
     var setup = function(player) {
         // initialize the foo plugin
         player.foo();
     };
+    var foo = function() {};
+
+    videojs.plugin('foo', foo);
+    videojs.hook('setup', setup);
+    var player = videojs('some-id', {autoplay: true, controls: true});
 ```
 
 ## Usage
 
 ### Adding
-In order to setup hooks you must first include Video.js in the page or script that you are using. Then you add hooks before running the videojs function like this:
+In order to use hooks you must first include video.js in the page or script that you are using. Then you add hooks using `videojs.hook(<name>, function)` before running the `videojs()` function.
 
+Example: adding hooks
 ```js
     videojs.hook('beforesetup', function(videoEl, options) {
         // videoEl will be the element with id=vid1
         // options will contain {autoplay: false}
     });
     videojs.hook('setup', function(player) {
-        // player will be the player that would get returned by the videojs() function
+        // player will be the same player that is defined below
+        // as `var player`
     });
-    videojs('vid1', {autoplay: false});
+    var player = videojs('vid1', {autoplay: false});
 ```
 
-After adding your hooks they will automatically be run at the correct time in the Video.js lifecycle
+After adding your hooks they will automatically be run at the correct time in the video.js lifecycle.
 
 ### Getting
-To access the array of hooks that currently exists and will be run on the Video.js object you can use the `hooks` function:
+To access the array of hooks that currently exists and will be run on the video.js object you can use the `videojs.hooks` function.
 
+Example: getting all hooks attached to video.js
 ```js
     var beforeSetupHooks = videojs.hooks('beforesetup');
     var setupHooks = videojs.hooks('setup');
 ```
 
 ### Removing
-To remove hooks from being executed during lifecycle events you will use `removeHook` like so:
+To stop hooks from being executed during the video.js lifecycle you will remove them using `videojs.removeHook`.
 
+Example: remove a hook that was defined by you
 ```js
     var beforeSetup = function(videoEl, options) {};
 
@@ -71,8 +105,9 @@ To remove hooks from being executed during lifecycle events you will use `remove
     videojs.removeHook('beforesetup', beforeSetup);
 ```
 
-You can also use `hooks` in conjunction with `removeHook` but it may have unexpected results if used during an asynchronous callbacks as other plugins/functionality may have added hooks.
+You can also use `videojs.hooks` in conjunction with `videojs.removeHook` but it may have unexpected results if used during an asynchronous callbacks as other plugins/functionality may have added hooks.
 
+Example: using `videojs.hooks` and `videojs.removeHook` to remove a hook
 ```js
     // add the hook
     videojs.hook('setup', function(videoEl, options) {});
--- a/docs/guides/hooks.md
+++ b/docs/guides/hooks.md
@@ -1,0 +1,86 @@
+# Hooks
+Hooks exist so that users can latch on to certain Video.js lifecycle events
+
+
+## Current Hooks
+Currently the following lifecycle events are accept and use hooks
+
+### presetup
+Called just before the player is created, so that the user can modify the options passed to the player or modify the video element that will be used for the player
+An example of this hook:
+
+```js
+    var preSetup = function(videoEl, options) {
+        videoEl.className += ' some-super-class';
+
+        // never autoplay because it makes me mad!
+        options.autoplay = false
+
+        // options that are returned here will be merged with old options
+        // automatically
+        return options;
+    };
+```
+
+### postsetup
+Called just after the player is created, so that plugins or custom functionality can be initialized
+
+```js
+    var postSetup = function(player) {
+        // initialize the foo plugin
+        player.foo();
+    };
+```
+
+
+## Usage
+
+### Adding
+In order to setup hooks you must first include Video.js in the page or script that you are using. Then you add hooks before running the videojs function like this:
+
+```js
+    videojs.hook('presetup', function(videoEl, options) {
+        // videoEl will be the element with id=vid1
+        // options will contain {autoplay: false}
+    });
+    videojs.hook('presetup', function(player) {
+        player will be the player that would get returned by the videojs function
+    });
+    videojs('vid1', {autoplay: false});
+```
+
+After adding your hooks they will automatically be run at the correct time in the Video.js lifecycle
+
+### Getting
+To access the array of hooks that currently exists and will be run on the Video.js object you can use the `hooks` function:
+
+```js
+    var preSetupHooks = videojs.hooks('presetup');
+    var postSetupHooks = videojs.hooks('postsetup');
+```
+
+### Removing
+To remove hooks from being executed during lifecycle events you will use `removeHook` like so:
+
+```js
+    var preSetup = function(videoEl, options) {};
+
+    // add the hook
+    videojs.hook('presetup', preSetup);
+
+    // remove that same hook
+    videojs.removeHook('presetup', preSetup);
+```
+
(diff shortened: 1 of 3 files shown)


src/js/video.js
```diff
@@ -96,10 +96,70 @@ function videojs(id, options, ready) {
...
   // Element may have a player attr referring to an already created player instance.
-  // If not, set up a new player and return the instance.
-  return tag.player || Player.players[tag.playerId] || new Player(tag, options, ready);
+  // If so return that otherwise set up a new player below
+  if (tag.player || Player.players[tag.playerId]) {
+    return tag.player || Player.players[tag.playerId];
+  }
+
+  videojs.hooks('presetup').forEach(function(hookFunction) {
+    options = videojs.mergeOptions(options, hookFunction(tag, options, ready));
```
gkatsev: Why pass `ready` to the hook?

src/js/video.js
```diff
@@ -96,10 +96,70 @@ function videojs(id, options, ready) {
...
+ */
+videojs.hooks_ = {};
+
+/**
+ * Get a list of hooks for a specific lifecycle event
+ *
+ * @param {String} type the lifecyle event to get hooks from
+ * @return {Array} an array of hooks, or an empty array if there are none
+ */
+videojs.hooks = function(type) {
```
gkatsev: Is hooks a too-generic term? would we want any more hooks on the main videojs object beyond `presetup` and `postsetup`?

src/js/video.js
BrandonOCasey: we probably don't need to

src/js/video.js
```diff
@@ -96,10 +96,70 @@ function videojs(id, options, ready) {
...
 
   // Element may have a player attr referring to an already created player instance.
-  // If not, set up a new player and return the instance.
-  return tag.player || Player.players[tag.playerId] || new Player(tag, options, ready);
+  // If so return that otherwise set up a new player below
+  if (tag.player || Player.players[tag.playerId]) {
+    return tag.player || Player.players[tag.playerId];
+  }
+
+  videojs.hooks('presetup').forEach(function(hookFunction) {
```
misteroneill: The `ModalDialog` component has similar lifecycle events, but the naming scheme is:

- `beforefoo`: Happens before the thing.
- `foo`: Happens after the thing.

Maybe we ought to follow that universally and use `beforesetup` and `setup`?

BrandonOCasey: we could, and previously I had it implemented like that, it just seemed like a lot of duplicate code

gkatsev: Wouldn't changing it just just involve changing `postsetup` on line 111 into `setup`?

misteroneill: And `presetup` to `beforesetup`? I don't see where the duplicate code concern comes from...

BrandonOCasey: These are the possible implementations with that naming scheme & the pro's/con's. do you see any other ways to implement it that sound better? (maybe I am just not seeing a good implementation?)   
> Also I wanted to note that using `setup` seem a bit weird to me since it will actually be called `postSetup` and not during setup

### Add lifecycle setters, getters, removers, and lists in a loop to the prototype 
* can make things confusing
* will not duplicate code
* will pollute the video.js object with any lifecycle event functions that we need to add
* may be more intuitive for a new user to call `beforeSetup` and `setup` to access things

### use the hooks terminology internally and add lifecycle  setters, getters, removers, and lists to the prototype in a loop
* Makes me think we should just use hooks directly
* will only pollute the video.js object with wrapper functions
* will not duplicate code
* may be more intuitive for the user

### implement lifecycle functions without any looping or hooks terminology
* Every time we need to expose a new event we have to duplicate functionality (adding a setter, getter, remover, and a list to video.js)
* will pollute video.js with wrapper functions
* lots of duplicate code
* may be intuitive for the user

BrandonOCasey: I think this is resolved

misteroneill: Yep, thanks!

docs/guides/hooks.md
```diff
@@ -0,0 +1,85 @@
+# Hooks
+Hooks exist so that users can latch on to certain Video.js lifecycle events
```
misteroneill: I think we ought to explain that this is a separate concept from per-player events. In fact, we should probably not use the term "events" at all, lest we cause confusion with events-events.

docs/guides/hooks.md
```diff
@@ -0,0 +1,85 @@
...
+        videoEl.className += ' some-super-class';
+
+        // never autoplay because it makes me mad!
+        options.autoplay = false
+
+        // options that are returned here will be merged with old options
+        // automatically
+        return options;
+    };
+```
```
misteroneill: I think this example (and the one below) is unclear without the `videojs.hook()` call.

docs/guides/hooks.md
```diff
@@ -0,0 +1,85 @@
+# Hooks
+Hooks exist so that users can latch on to certain Video.js lifecycle events
+
+
+## Current Hooks
+Currently the following lifecycle events are accept and use hooks
```
misteroneill: I think the heading can go away. And the phrasing/grammar here is weird. Maybe:

```
Currently, the following hooks are available:
```

docs/guides/hooks.md
```diff
@@ -0,0 +1,85 @@
...
+
+### beforesetup
+Called just before the player is created, so that the user can modify the options passed to the player or modify the video element that will be used for the player
```
misteroneill: Missing a period here.

docs/guides/hooks.md
gkatsev: maybe something like:
`Hooks exist so that users can hook into the video.js player lifecycle`

docs/guides/hooks.md
gkatsev: I think we should mention here (and not just in the example) what happens if multiple hooks are.
Also, state that the method takes in the video element and an options object and is expected to return a new options object.

src/js/video.js
```diff
@@ -96,10 +98,70 @@ function videojs(id, options, ready) {
...
   // Element may have a player attr referring to an already created player instance.
-  // If not, set up a new player and return the instance.
-  return tag.player || Player.players[tag.playerId] || new Player(tag, options, ready);
+  // If so return that otherwise set up a new player below
+  if (tag.player || Player.players[tag.playerId]) {
+    return tag.player || Player.players[tag.playerId];
+  }
+
+  videojs.hooks('beforesetup').forEach(function(hookFunction) {
+    options = videojs.mergeOptions(options, hookFunction(tag, videojs.mergeOptions({}, options)));
```
gkatsev: what happens if the `hookFunction` returns `null` or a string?
We probably should write some tests for this.

It may be easier to write tests if these are moved into a separate file and then copied over onto the videojs object but either way should work.

src/js/video.js
```diff
@@ -96,10 +98,70 @@ function videojs(id, options, ready) {
...
+
+/**
+ * Add a function hook to a specific videojs lifecycle event
+ *
+ * @param {String} type the lifecycle event to hook the function to
+ * @param {Function} fn the function to attach
+ */
+videojs.hook = function(type, fn) {
+  videojs.hooks_[type] = videojs.hooks_[type] || [];
+  videojs.hooks_[type].push(fn);
```
gkatsev: instead of using push, we should do `videojs.hooks_[type] = videojs.hooks_[type].concat(fn);`.
See https://github.com/videojs/mux.js/pull/121

src/js/video.js
```diff
@@ -96,10 +98,70 @@ function videojs(id, options, ready) {
...
+
+/**
+ * Remove a hook from a specific videojs lifecycle event
+ *
+ * @param {String} type the lifecycle event that the function hooked to
+ * @param {Function} fn the hooked function to remove
+ * @return {Function|undefined} the function that was removed or undef
+ */
+videojs.removeHook = function(type, fn) {
+  const hooks = videojs.hooks_[type] || [];
```
gkatsev: we should also slice here as well. See https://github.com/videojs/mux.js/pull/121 again.

BrandonOCasey:
I still have to write some unit tests for this functionality

docs/guides/hooks.md
misteroneill: > Conceptually similar to events, hooks provide users a way to intercept universal behaviors of video.js itself rather than individual players.

gkatsev: Are they similar to events though? You don't generally have the ability to do an action like what we're doing with these hooks.

src/js/video.js
```diff
@@ -96,10 +98,69 @@ function videojs(id, options, ready) {
...
+ */
+videojs.hooks_ = {};
+
+/**
+ * Get a list of hooks for a specific lifecycle event
+ *
+ * @param {String} type the lifecyle event to get hooks from
+ * @return {Array} an array of hooks, or an empty array if there are none
+ */
+videojs.hooks = function(type) {
```
gkatsev: Can we add a check to `hooks` that if it's called with two arguments either logs an error or calls `hook` for the user? Just ran into this problem and took a second to figure it out.

src/js/video.js
```diff
@@ -96,10 +98,69 @@ function videojs(id, options, ready) {
...
+};
+
+/**
+ * Add a function hook to a specific videojs lifecycle event
+ *
+ * @param {String} type the lifecycle event to hook the function to
+ * @param {Function|Array} fn the function to attach
+ */
+videojs.hook = function(type, fn) {
+  const hooks = videojs.hooks[type];
```
gkatsev: this should be `videojs.hooks_[type] || []` or `videojs.hooks(type);`

src/js/video.js
```diff
@@ -96,10 +98,69 @@ function videojs(id, options, ready) {
...
+
+  hooks[type].concat(fn);
```
gkatsev: should be `videojs.hooks_[type] = hooks.concat(fn);`

docs/guides/hooks.md
misteroneill: Fair enough, but they are conceptually similar in the sense that you're providing a function to be called when a certain condition is met. But, of course, I'm not married to any particular phrasing. 😄

src/js/video.js
```diff
@@ -96,10 +98,77 @@ function videojs(id, options, ready) {
...
   // Element may have a player attr referring to an already created player instance.
-  // If not, set up a new player and return the instance.
-  return tag.player || Player.players[tag.playerId] || new Player(tag, options, ready);
+  // If so return that otherwise set up a new player below
+  if (tag.player || Player.players[tag.playerId]) {
+    return tag.player || Player.players[tag.playerId];
+  }
+
+  videojs.hooks('beforesetup').forEach(function(hookFunction) {
+    const opts = hookFunction(tag, videojs.mergeOptions({}, options));
```
gkatsev: basically, each handler that doesn't return an object is ignored?

src/js/video.js
```diff
@@ -96,10 +98,77 @@ function videojs(id, options, ready) {
...
+ * @param {Function} fn the hooked function to remove
+ * @return {Boolean} the function that was removed or undef
+ */
+videojs.removeHook = function(type, fn) {
+  const index = videojs.hooks(type).indexOf(fn);
+
+  videojs.hooks_[type] = videojs.hooks_[type].slice();
+  videojs.hooks_[type].splice(index, 1);
+
+  return index > -1;
```
gkatsev: maybe also do this check above so we don't run the extra slice and splice unless necessary?

test/unit/video.test.js
```diff
@@ -166,3 +166,161 @@ QUnit.test('should expose DOM functions', function(assert) {
...
+
+  const errRetval = videojs.removeHook('bar', noop);
+
+  assert.equal(errRetval, false, 'should return false');
+  assert.equal(Object.keys(videojs.hooks_).length, 2, 'should have 2 hooks types');
+  assert.equal(videojs.hooks_.foo.length, 0, 'should have 0 foo hook');
+  assert.equal(videojs.hooks_.bar.length, 0, 'should have 0 bar hook');
+});
+
+QUnit.test('should be get all hooks for a type', function(assert) {
```
gkatsev: should be *able to* get all hooks for a type

test/unit/video.test.js
```diff
@@ -166,3 +166,161 @@ QUnit.test('should expose DOM functions', function(assert) {
...
+  assert.deepEqual(videojs.hooks_.foo, fooHooks, 'should return the exact foo list from videojs.hooks_');
+  assert.deepEqual(videojs.hooks_.bar, barHooks, 'should return the exact bar list from videojs.hooks_');
+});
+
+QUnit.test('should trigger beforesetup and setup during videojs setup', function(assert) {
+  const vjsOptions = {techOrder: ['techFaker']};
+  let setupCalled = false;
+  let beforeSetupCalled = false;
+  const beforeSetup = function(video, options) {
+    beforeSetupCalled = true;
```
gkatsev: what about asserting that `setupCalled` is still false in here?

test/unit/video.test.js
```diff
@@ -166,3 +166,161 @@ QUnit.test('should expose DOM functions', function(assert) {
...
+  videojs.hook('setup', setup);
+
+  const player = videojs(vid, vjsOptions);
+
+  assert.ok(player.options_, 'returning null in beforesetup does not lose options');
+  assert.equal(beforeSetupCalled, true, 'beforeSetup was called');
+  assert.equal(setupCalled, true, 'setup was called');
+});
+
+QUnit.test('beforesetup returns dont break videojs options', function(assert) {
```
gkatsev: Do we want a positive test case for this as well? 
Something like: `beforesetup` hooks modify options properly and merge each item in the list

src/js/video.js
BrandonOCasey: well they are still run, we just assume they don't want change options

gkatsev: Yup, makes sense. Meant the response is ignored.

src/js/video.js
BrandonOCasey: done

test/unit/video.test.js
BrandonOCasey: done, good catch

test/unit/video.test.js
BrandonOCasey: done, also added a check for `beforeSetupCalled` in the setup function

test/unit/video.test.js
```diff
@@ -327,3 +327,23 @@ QUnit.test('beforesetup returns dont break videojs options', function(assert) {
...
+
+QUnit.test('beforesetup options override videojs options', function(assert) {
+  const vjsOptions = {techOrder: ['techFaker'], autoplay: false};
+  const fixture = document.getElementById('qunit-fixture');
+
+  fixture.innerHTML += '<video id="test_vid_id"><source type="video/mp4"></video>';
+
+  const vid = document.getElementById('test_vid_id');
+
+  videojs.hook('beforesetup', function(options) {
```
gkatsev: can we add another hook that say adds `fluid: true` option?

PR #3640, opened 2016-09-23
check throwIfWhitespace first in hasElClass

gkatsev:
## Description
This is because the new spec for HTML has the classList (DOMTokenList)
.contains always return a value. Either true if it exists or false
otherwise.
## Requirements Checklist


gkatsev:
Also, this fixes the firefox tests that are breaking for us currently.

gkatsev:
Merged into stable.

PR #3644, opened 2016-09-26
Proxy ios webkit events into fullscreenchange

gkatsev:
## Description
Previously, the ios webkit events (webkitbeginfullscreen and
webkitendfullscreen) were only proxied into the fullscreenchange event
when requestFullscreen was called directly. So, on iPhones, we would
never be able to get this event. This makes it so these events are
always proxied.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -586,23 +586,25 @@ class Html5 extends Tech {
    * @method proxyWebkitFullscreen_
    */
   proxyWebkitFullscreen_() {
-    if ('webkitDisplayingFullscreen' in this.el_) {
-      const endFn = function() {
-        this.trigger('fullscreenchange', { isFullscreen: false });
-      };
-
-      const beginFn = function() {
-        this.one('webkitendfullscreen', endFn);
-
-        this.trigger('fullscreenchange', { isFullscreen: true });
-      };
-
-      this.on('webkitbeginfullscreen', beginFn);
-      this.on('dispose', () => {
-        this.off('webkitbeginfullscreen', beginFn);
-        this.off('webkitendfullscreen', endFn);
-      });
+    if (!('webkitDisplayingFullscreen' in this.el_)) {
+      return;
     }
+
+    const endFn = function() {
+      this.trigger('fullscreenchange', { isFullscreen: false });
+    };
+
+    const beginFn = function() {
+      this.one('webkitendfullscreen', endFn);
+
+      this.trigger('fullscreenchange', { isFullscreen: true });
+    };
+
+    this.on('webkitbeginfullscreen', beginFn);
+    this.on('dispose', () => {
+      this.off('webkitbeginfullscreen', beginFn);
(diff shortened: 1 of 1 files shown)


gkatsev:
This is failing the test due to the firefox issue which is fixed [here](https://github.com/videojs/video.js/pull/3640)

src/js/tech/html5.js
```diff
@@ -573,6 +577,28 @@ class Html5 extends Tech {
...
+      const webkitbeginFn = function() {
+        this.one('webkitendfullscreen', function() {
+          this.trigger('fullscreenchange', { isFullscreen: false });
+        });
+
+        this.trigger('fullscreenchange', { isFullscreen: true });
+      };
+
+      this.on('webkitbeginfullscreen', webkitbeginFn);
+      this.on('dispose', () => this.off('webkitbeginfullscreen', webkitbeginFn));
```
BrandonOCasey: I don't think `webkitendfullscreen` will be disposed correctly if we dispose the player while in full screen for some reason

gkatsev: Yup, you're correct. I'll update it.

gkatsev: Updated.

src/js/tech/html5.js
```diff
@@ -573,6 +577,33 @@ class Html5 extends Tech {
...
 
   /**
+   * Proxy iOS `webkitbeginfullscreen` and `webkitendfullscreen` into
+   * `fullscreenchange` event
+   *
+   * @private
+   * @method proxyWebkitFullscreen_
+   */
+  proxyWebkitFullscreen_() {
+    if ('webkitDisplayingFullscreen' in this.el_) {
```
BrandonOCasey: I would do `if (!'webkitDisplayingFullscreen' in this.el_)` if that works and return at the top

src/js/tech/html5.js
```diff
@@ -573,6 +577,33 @@ class Html5 extends Tech {
...
+      const beginFn = function() {
+        this.one('webkitendfullscreen', endFn);
+
+        this.trigger('fullscreenchange', { isFullscreen: true });
+      };
+
+      this.on('webkitbeginfullscreen', beginFn);
+      this.on('dispose', () => {
+        this.off('webkitbeginfullscreen', beginFn);
+        this.off('webkitendfullscreen', endFn);
```
BrandonOCasey: will this trigger an error if we are not in full screen? aka `this.one('webkitendfullscreen', endFn);` is already triggered and gone or was never added in the first place(in the case of never going to full screen)

gkatsev: Nope, neither of these should error out if it's already been removed.

src/js/tech/html5.js
gkatsev: Yeah, that would be a bit nicer.

gkatsev:
I've made the nesting change. I've also rebased against master, so, hopefully, the tests will pass without the firefox issue.

PR #3648, opened 2016-09-28
If we create a new element for movingMediaElInDOM, set playerId on the el again

gkatsev:
## Description
This is a rework of https://github.com/videojs/video.js/pull/3624 against stable. Fixes https://github.com/videojs/video.js/issues/3283.


gkatsev:
Merged into stable.

Issue #3645, opened 2016-09-26
Fix style `MouseTimeDisplay`


chemoish:
## Description

When hovering over the `ProgressControl` the `MouseTimeDisplay` activates outside of the expected boundary of the `SeekBar`.

I suspect it has to do with `margin: 0 0.45em 0 0.45em;` identified in https://github.com/videojs/video.js/blob/master/src/css/components/_slider.scss#L6 and possibly how parent element calculates the `MouseTimeDisplay`?
## Steps to reproduce
1. Go to http://videojs.com/
2. Play video
3. Attempt to hover to the edges of the `SeekBar`
### Expected

For the `MouseTimeDisplay` to be bound by the `SeekBar`.
### Actual

![image](https://cloud.githubusercontent.com/assets/549319/18855683/8c49429a-8409-11e6-9ae1-cebde7679f51.png)
## Additional Information

Please include any additional information necessary here. Including the following:
#### videojs

`5.11.6`
#### browsers
- `Chrome`
- `Firefox`
- `Safari`
#### OSes

OSX `10.11.5`

chemoish:
Possibly, easy, but naive fix would be to update the `ProgressControl` to have the `margin` instead.

gkatsev:
Huh, oops. Also, I'm thinking when we do a new major release for videojs we make the "keepToolstipsInside" option on the progress control be the default and only option, which, I think would also fix this issue.

PR #3658, opened 2016-09-29
style(_progress): Provide padding for the `vjs-progress-holder` while…

chemoish:
## Description

Fixes https://github.com/videojs/video.js/issues/3645. Looks like it doesn't effect `volume-bar`,  which is the only other dependent of `Slider`, because it manages its own margins. SEE: https://github.com/videojs/video.js/blob/master/src/css/components/_volume.scss#L28
## Specific Changes proposed

image
## Requirements Checklist
  - [ ] Docs/guides updated

… allowing it to resize on hover. Thereby containing the `MouseTimeDisplay` to the `vjs-progress-holder` instead of it bleeding outside.

NOTE: Guessing that the margin was put in place to provide some spacing between the slider and the surrounding controls. Because the slider itself, or rather, `vjs-progress-holder` becomes larger on hover, we cannot have the "padding" also be responsive on the same element.

### DIFF
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -15,6 +15,7 @@
 .video-js .vjs-progress-control {
   @include flex(auto);
   @include display-flex(center);
+  margin: 0 0.45em 0 0.45em;
   min-width: 4em;
 }
 
--- a/src/css/components/_slider.scss
+++ b/src/css/components/_slider.scss
@@ -3,7 +3,6 @@
   position: relative;
   cursor: pointer;
   padding: 0;
-  margin: 0 0.45em 0 0.45em;
 
   @include background-color-with-alpha($secondary-background-color, $secondary-background-transparency);
 }


src/css/components/_slider.scss
```diff
@@ -3,7 +3,6 @@
   position: relative;
   cursor: pointer;
   padding: 0;
-  margin: 0 0.45em 0 0.45em;
```
gkatsev: does the volume slider still look ok?

chemoish: image

image

src/css/components/_progress.scss
```diff
@@ -15,6 +15,7 @@
 .video-js .vjs-progress-control {
   @include flex(auto);
   @include display-flex(center);
+  margin: 0 0.45em 0 0.45em;
```
gkatsev: Looks like this would break the Brightcove skin by making the tooltips go out of bounds with the `keepTooltipsInside`. It probably would also break skins like https://github.com/mmcc/videojs-skin-twitchy.

src/css/components/_slider.scss
gkatsev: It seems like only this removal is necessary to fix the tooltip issue.

src/css/components/_progress.scss
gkatsev: I would just get rid of this line as it seems to be unnecessary.

chemoish: If there isn't any margin added it makes it very difficult to access the surrounding components because there is only 1px between the volume and the seekbar at position 0.

I didn't quite follow the comment.

gkatsev: Actually, you're right. There is a slight difference.
The issue is that with this line here, our tooltips look like this:
broken tooltip
notice how the right side is a bit cut off. Vs, this:
working tooltip

The position of the mouse marker in both those cases is correct but the tooltip isn't.

chemoish: It would seem that the Brightcove skin might need to include that in the source if it utilizes a fullscreen `SeekBar`?

Should video.js accommodate fullscreen `SeekBar` in its implementation?

Not sure best way to move forward on this.

gkatsev: it isnt just in fullscreen that's the issue. It gets cut off by the overflow. The videojs-skin-twichy would also have a similar issue.
But basically it means that we cannot make this change as is because it would break this usecase.

chemoish: Hmm, I am not sure how we would go about fixing this issue if the default styles directly conflict with other skins.

The default is embedded in the control bar, but the other is not...? I guess its kind of chicken/egg'ish.

If it makes sense for the default implementation to be the primary one, then the other skins would need to undo these changes?

The twitchy skin implements https://github.com/mmcc/videojs-skin-twitchy/blob/master/src/components/_slider.scss#L6 (which removes the inline `SeekBar` margin). It would also need to remove the `.vjs-progress-control` ones too (if they were added).

---

Not saying that this should be the case, just that the styles are conflicting—And its weird for the default implementation to have a `MouseTimeDisplay` that goes out of bounds, especially on the home page :p.

gkatsev:
I think I'm going to close this PR. Not sure exactly how to fix this without potentially breaking users. In videojs 6.0, we're going to transition to only having one kind of progress with mouse tooltips and we can make the fix there.

PR #3660, opened 2016-09-30
Refactor TextTrackSettings (Phase 2)

misteroneill:
We can review and merge this into the already-approved branch from #3570, then open a new PR for the combination of the two.

## Specific Changes proposed
This refactors the `TextTrackSettings` HTML blob into `createEl` calls. Also, we extend the `createEl` function to take a conventional `content` argument at the end, allowing nested calls to `createEl` to build up a DOM structure.

## Requirements Checklist


src/js/tracks/text-track-settings.js
```diff
@@ -22,36 +23,127 @@ const LOCAL_STORAGE_KEY = 'vjs-text-track-settings';
...
+    ]
   },
+
   color: {
-    selector: '.vjs-fg-color > select'
+    selector: '.vjs-fg-color > select',
+    id: 'captions-foreground-color-%s',
+    label: 'Color',
+    options: [
+      ['#FFF', 'White'],
```
BrandonOCasey: We should define the colors at the top once as a const, since we are using them in two places

src/js/tracks/text-track-settings.js
```diff
@@ -22,36 +23,127 @@ const LOCAL_STORAGE_KEY = 'vjs-text-track-settings';
...
   },
+
+  // Options for this object are defined below.
   windowOpacity: {
-    selector: '.vjs-window-opacity > select'
+    selector: '.vjs-window-opacity > select',
+    id: 'captions-window-opacity-%s',
+    label: 'Transparency',
+    options: [
+      ['0', 'Transparent'],
```
BrandonOCasey: maybe define this as a const as well since it is used in two places

src/js/tracks/text-track-settings.js
```diff
@@ -22,36 +23,127 @@ const LOCAL_STORAGE_KEY = 'vjs-text-track-settings';
...
+      ['script', 'Script'],
+      ['small-caps', 'Small Caps']
+    ]
   },
+
   fontPercent: {
     selector: '.vjs-font-percent > select',
+    id: 'captions-font-size-%s',
+    label: 'Font Size',
+    options: [
```
BrandonOCasey: why not use an object for options in all of these?

src/js/tracks/text-track-settings.js
```diff
@@ -273,25 +242,184 @@ class TextTrackSettings extends Component {
...
+    return [
+      createEl(
+        'label',
+        {className: 'vjs-label', textContent: config.label},
+        {for: id}),
+      createEl(
+        'select',
+        {id},
+        undefined,
+        config.options.map(o => {
```
BrandonOCasey: if options becomes an object this will have to change

src/js/tracks/text-track-settings.js
```diff
@@ -273,25 +242,184 @@ class TextTrackSettings extends Component {
...
+   * @return {Element}
+   * @method createElColors_
+   */
+  createElColors_() {
+    return createEl(
+      'div',
+      {className: 'vjs-tracksettings-colors'},
+      undefined,
+      [
+        createEl(
```
BrandonOCasey: maybe create all of these elements outside of the main `createEl` call so that this code is more understandable (as they will all have to have names, and you can add comments)

src/js/tracks/text-track-settings.js
```diff
@@ -273,25 +242,184 @@ class TextTrackSettings extends Component {
...
+   * @return {Element}
+   * @method createElFont_
+   */
+  createElFont_() {
+    return createEl(
+      'div',
+      {className: 'vjs-tracksettings-font'},
+      undefined,
+      [
+        createEl(
```
BrandonOCasey: same comment with this one as the `createElColors_` although it isn't nearly as bad.

src/js/tracks/text-track-settings.js
```diff
@@ -273,25 +242,184 @@ class TextTrackSettings extends Component {
...
-    }, {
-      'role': 'dialog',
-      'aria-labelledby': dialogLabelId,
-      'aria-describedby': dialogDescriptionId
-    });
+    const settings = createEl(
+      'div',
+      {className: 'vjs-tracksettings'},
+      undefined,
+      [this.createElColors_(), this.createElFont_(), this.createElControls_()]);
```
BrandonOCasey: should this `);` go on the next line?

src/js/tracks/text-track-settings.js
```diff
@@ -273,25 +242,184 @@ class TextTrackSettings extends Component {
...
+
+    const heading = createEl(
```
BrandonOCasey: might just be opinion here but I would like it if these blocks looked something like this: 
```JavaScript
const heading = createEl('div', {
  ...
  }, {
  ...
});
```

src/js/tracks/text-track-settings.js
```diff
@@ -327,7 +455,20 @@ class TextTrackSettings extends Component {
   }
 
   /**
-   * Restore text track settings
+   * Sets all <select> elements to their default values.
+   *
+   * @method setDefaults
+   */
+  setDefaults() {
+    Obj.each(selectConfigs, config => {
```
BrandonOCasey: I think we typically wrap this in another set of parens in video.js. `Obj.each((selectConfigs, config) =>`

src/js/tracks/text-track-settings.js
```diff
@@ -273,25 +242,184 @@ class TextTrackSettings extends Component {
...
+  /**
+   * Create controls for the component
+   *
+   * @private
+   * @return {Element}
+   * @method createElControls_
+   */
+  createElControls_() {
+    return createEl(
+      'div',
```
BrandonOCasey: same comment with this one as the createElColors_ although it isn't nearly as bad.

src/js/utils/dom.js
```diff
@@ -96,10 +96,11 @@ export function getEl(id) {
  * @param  {String} [tagName='div'] Name of tag to be created.
  * @param  {Object} [properties={}] Element properties to be applied.
  * @param  {Object} [attributes={}] Element attributes to be applied.
+ * @param  {String|Element|TextNode|Array|Function} [content] Contents for the element (see: `normalizeContent`)
  * @return {Element}
  * @function createEl
  */
-export function createEl(tagName = 'div', properties = {}, attributes = {}) {
+export function createEl(tagName = 'div', properties = {}, attributes = {}, content) {
```
gkatsev: This change would make it a minor, which is fine.

gkatsev: But it does seem like a nice thing to have.

src/js/tracks/text-track-settings.js
misteroneill: Also, ordering could be important.

src/js/tracks/text-track-settings.js
misteroneill: I feel like the function argument format we've decided on was:

```js
foo(
  'a',
  'b',
  'c');
```

src/js/tracks/text-track-settings.js
misteroneill: We're iterating over `selectConfigs`, though. If anything, it'd be `Obj.each(selectConfigs, (config) =>`.

src/js/tracks/text-track-settings.js
misteroneill: Yeah, this one could be simpler.

src/js/tracks/text-track-settings.js
misteroneill: I agree, personally, but this is another case of trying to be consistent is some way w/ function arguments.

src/js/tracks/text-track-settings.js
misteroneill: But it's apparent that this could be confusing, I'll add parens around `config`. 👍

misteroneill:
To be clear about the timeline/plan here, once this is approved, I'll close an re-open a new PR with both this one and #3570 together.

src/js/tracks/text-track-settings.js
```diff
@@ -27,14 +40,14 @@ const selectConfigs = {
...
     options: [
-      ['#000', 'Black'],
-      ['#FFF', 'White'],
-      ['#F00', 'Red'],
-      ['#0F0', 'Green'],
-      ['#00F', 'Blue'],
-      ['#FF0', 'Yellow'],
-      ['#F0F', 'Magenta'],
-      ['#0FF', 'Cyan']
+      COLOR_BLACK,
```
BrandonOCasey: 👍

gkatsev:
@OwenEdwards would you be able to take a look?

src/js/tracks/text-track-settings.js
```diff
@@ -301,7 +301,7 @@ class TextTrackSettings extends Component {
 
     return createEl('fieldset', {
       className: 'vjs-fg-color vjs-tracksetting'
-    }, undefined, [legend, select, opacity]);
+    }, undefined, [legend].concat(select, opacity));
```
gkatsev: why is the other way broken?

misteroneill: Because the `select` is an array.

gkatsev: Oh, I see now.

PR #3664, opened 2016-10-03
feat: add a safe getComputedStyle to videojs.

gkatsev:
## Description

This is used internally in the seek bar and mouse time display.
In firefox, in an iframe that is hidden with "display: none",
getComputedStyle() returns "null" which can break things.
See https://bugzilla.mozilla.org/show_bug.cgi?id=548397 for more
details.
## Specific Changes proposed
- add a `getComputedStyle` util
- expose `videojs.getComputedStyle`
- use the util in mouse time display and seek bar

### DIFF
--- a/src/js/utils/computed-style.js
+++ b/src/js/utils/computed-style.js
@@ -1,3 +1,5 @@
+import window from 'global/window';
+
 /**
  * A safe getComputedStyle with an IE8 fallback.
  *
@@ -6,16 +8,20 @@
  * that the player doesn't break in these cases.
  * See https://bugzilla.mozilla.org/show_bug.cgi?id=548397 for more details.
  *
- * @function getComputedStyle
+ * @function computedStyle
  * @param el the element you want the computed style of
  * @param prop the property name you want
  */
-export default function getComputedStyle(el, prop) {
-  if (typeof getComputedStyle === 'function') {
-    const cs = getComputedStyle(el);
+export default function computedStyle(el, prop) {
+  if (!el || !prop) {
+    return '';
+  }
+
+  if (typeof window.getComputedStyle === 'function') {
+    const cs = window.getComputedStyle(el);
 
     return cs ? cs[prop] : '';
   }
 
-  return el.currentStyle[prop];
+  return el.currentStyle[prop] || '';
 }
--- a/src/js/utils/get-computed-style.js
+++ b/src/js/utils/get-computed-style.js
@@ -1,0 +1,21 @@
+/**
+ * A safe getComputedStyle with an IE8 fallback.
+ *
+ * This is because in Firefox, if the player is loaded in an iframe with `display:none`,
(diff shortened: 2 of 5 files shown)


src/js/utils/get-computed-style.js
```diff
@@ -0,0 +1,21 @@
...
+ * then `getComputedStyle` returns `null`, so, we do a null-check to make sure
+ * that the player doesn't break in these cases.
+ * See https://bugzilla.mozilla.org/show_bug.cgi?id=548397 for more details.
+ *
+ * @function getComputedStyle
+ * @param el the element you want the computed style of
+ * @param prop the property name you want
+ */
+export default function getComputedStyle(el, prop) {
+  if (typeof getComputedStyle === 'function') {
```
BrandonOCasey: seems like we may want to check for el or prop being null here and return empty sting if that happens.

src/js/video.js
```diff
@@ -719,6 +720,20 @@ videojs.appendContent = Dom.appendContent;
...
+ * This is because in Firefox, if the player is loaded in an iframe with `display:none`,
+ * then `getComputedStyle` returns `null`, so, we do a null-check to make sure
+ * that the player doesn't break in these cases.
+ * See https://bugzilla.mozilla.org/show_bug.cgi?id=548397 for more details.
+ *
+ * @function getComputedStyle
+ * @param el the element you want the computed style of
+ * @param prop the property name you want
+ */
+videojs.getComputedStyle = computedStyle;
```
BrandonOCasey: do we want to call it `getComputedStyle`? Should it be `getComputedStyle` throughout the code rather than `computedStyle`?

src/js/utils/get-computed-style.js
misteroneill: Also, shouldn't this be `window.getComputedStyle`? 

I think this'll detect the module-level function instead of the global one.

src/js/video.js
misteroneill: I think it should be `computedStyle` so we don't confuse it with the global function.

src/js/utils/get-computed-style.js
gkatsev: Yeah, it'll be safer to do `window.`.

I guess checking that `el` and `prop` are available is good. I'll refrain from doing other type checking, though.

src/js/video.js
gkatsev: I'll do that.

src/js/utils/computed-style.js
```diff
@@ -6,16 +8,20 @@
...
+  }
+
+  if (typeof window.getComputedStyle === 'function') {
+    const cs = window.getComputedStyle(el);
 
     return cs ? cs[prop] : '';
   }
 
-  return el.currentStyle[prop];
+  return el.currentStyle[prop] || '';
```
BrandonOCasey: Nice catch here

PR #3676, opened 2016-10-08
Update almost all links to HTTPS

CoWinkKeyDinkInc:
## Description

After running into a simple error that was fixed by replacing it from HTTP to HTTPS, I made this PR to make it so websites that can serve HTTPS will serve it properly.
## Specific Changes proposed

This changes websites that work with HTTPS to use HTTPS links instead of the insecure HTTP links.  This also edits one link from whatwg.org to it links to the same article again.  This does not change links from w3c as I've found that it 404'd.
## Requirements Checklist
  - [] Docs/guides updated

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
 ## Description
 Briefly describe the issue.
-Include a [reduced test case](https://css-tricks.com/reduced-test-cases/), we have a [starter template](http://jsbin.com/axedog/edit?html,output) on JSBin you can use.
+Include a [reduced test case](https://css-tricks.com/reduced-test-cases/), we have a [starter template](https://jsbin.com/axedog/edit?html,output) on JSBin you can use.
 
 ## Steps to reproduce
 Explain in detail the exact steps necessary to reproduce the issue.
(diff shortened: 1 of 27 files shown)


gkatsev:
Thanks for the PR. Changing most of the links over to HTTPS makes sense to me.
Though, there's definitely links that either need to stay as http because they're used in tests or because the https urls don't currently work (for example, videojs.com itself doesn't work well over https right now).
In addition, URLs in documentation should probably just use the protocol-relative urls rather than http/https.

gkatsev:
Actually, most of the links can probably just be protocol relative as well.

CoWinkKeyDinkInc:
@gkatsev Don't worry, I didn't change any of the video.js links because I noticed they don't work with HTTPS.  I'll revert the HTTPS change on the test folders if you want.

CoWinkKeyDinkInc:
I don't know how to make protocol relative URLs

gkatsev:
Protocol relative urls just have `//` in front of them instead of `https://` or `http://`.

CoWinkKeyDinkInc:
Alright, I'll add these commits tomorrow

CoWinkKeyDinkInc:
So I reversed the changes to the text files, I think all the links that could use protocol relative URLS are already using them.

gkatsev:
Unfortunately, this got pretty out of date with master (partly my fault for not reviewing it in time). I'm still not sure that this is quite necessary but you can open a new PR if you update this.

CoWinkKeyDinkInc:
Okay, I'll take a look at it again and fix anything if it's needed.  Do you know why the builds failed btw?

PR #3678, opened 2016-10-11
Remove Flash and change DOM module

BrandonOCasey:
## Description

Remove the flash tech and tests from video.js. This requires a major version bump in video.js.
## Requirements Checklist


gkatsev:
Thoughts on re-adding Flash as a default plugin right now and then we can remove that plugin in december when we do the major version bump?

BrandonOCasey:
I think we may want to hold off on this until we can do a major version bump, seems like it will be a lot of work to resolve the circular dependencies here. (flash needs video.js to be loaded so that it can load itself, but video.js wants to include flash during load... etc)

src/js/video.js
```diff
@@ -719,6 +721,18 @@ videojs.appendContent = Dom.appendContent;
  */
 videojs.insertContent = Dom.insertContent;
 
+/**
+ * Export the Dom utilities for use in external plugins
+ * and Tech's
+ */
+videojs.Dom = Dom;
```
misteroneill: Looks like this is in here twice (also at L567).

BrandonOCasey:
update for the community. This is going to wait until `video.js` 6.0 due to some technical limitations, look for it in the next major release.

src/js/video.js
misteroneill: Also, can we expose it as `dom` instead? It's not a constructor; so, the casing doesn't make much sense.

src/js/video.js
```diff
@@ -719,6 +721,18 @@ videojs.appendContent = Dom.appendContent;
...
+
+/**
+ * Export the Url utilities for use in external plugins
+ * and Tech's
+ */
+videojs.Url = Url;
```
misteroneill: Can we expose this as `url` instead? It's not a constructor; so, the casing doesn't make much sense.

src/js/video.js
```diff
@@ -564,6 +564,8 @@ videojs.AudioTrack = AudioTrack;
  */
 videojs.VideoTrack = VideoTrack;
 
+videojs.Dom = Dom;
```
misteroneill: This can be removed.

BrandonOCasey:
pushed upstream to get tests running on browserstack see #3956

PR #3684, opened 2016-10-14
Add a note on Flash permissions in sandboxed environments

Haraldson:


misteroneill:
Thanks! Seems like a reasonable addition to me.

Haraldson:
Glad to hear it! It certainly would’ve saved me for days of debugging and googling. :)

PR #3685, opened 2016-10-14
Make the log work in IE9 when console is closed

misteroneill:
There is a bug in the existing logger when attempting to call console methods in IE9 while the console is closed. This addresses it.

## Requirements Checklist

### DIFF
--- a/src/js/utils/log.js
+++ b/src/js/utils/log.js
@@ -1,7 +1,7 @@
 /**
  * @file log.js
  */
-import window from 'global/window';
+import console from 'global/console';
 import {IE_VERSION} from './browser';
 
 /**
@@ -35,7 +35,7 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   //
   // This mainly prevents cases where old IE versions don't expose a `console`
   // object unless the console is actually open.
-  const fn = window.console && window.console[type];
+  const fn = console && console[type];
 
   if (!fn) {
     return;
@@ -62,7 +62,13 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   if (!fn.apply) {
     fn(args);
   } else {
-    fn[Array.isArray(args) ? 'apply' : 'call'](null, args);
+
+    // Make sure args is an array.
+    if (!Array.isArray(args)) {
+      args = [args];
+    }
+
+    fn.apply(null, args);
   }
 };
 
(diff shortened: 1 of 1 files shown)


src/js/utils/log.js
```diff
@@ -16,27 +16,30 @@ import {IE_VERSION} from './browser';
...
   log.history.push(args);
 
-  // add console prefix after adding to history
-  args.unshift('VIDEOJS:');
+  // If there's no console then don't try to output messages, but they will
+  // still be stored in `log.history`.
+  //
+  // This mainly prevents cases where old IE versions don't expose a `console`
+  // object unless the console is actually open.
+  const fn = window.console && window.console[type];
```
BrandonOCasey: i think that `global` actually exports console, so if you want to require console at the top rather than window you could use that.

src/js/utils/log.js
```diff
@@ -16,27 +16,30 @@ import {IE_VERSION} from './browser';
...
     args.unshift(type.toUpperCase() + ':');
   }
 
-  // add to history
+  args.unshift('VIDEOJS:');
+
+  // Add arguments to history
   log.history.push(args);
 
-  // add console prefix after adding to history
```
BrandonOCasey: I think that this comment is still relevant

src/js/utils/log.js
```diff
@@ -59,7 +62,7 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   if (!fn.apply) {
     fn(args);
   } else {
-    fn[Array.isArray(args) ? 'apply' : 'call'](console, args);
+    fn[Array.isArray(args) ? 'apply' : 'call'](null, args);
```
BrandonOCasey: why not force args into an array an always call apply?

src/js/utils/log.js
misteroneill: It's not. I moved that line about the `history.push` call. And, since `args` is added to `history` by reference, it never mattered if this was "after" the `history.push` call.

src/js/utils/log.js
misteroneill: I suppose we could do that, too. 👍

src/js/utils/log.js
```diff
@@ -16,27 +16,30 @@ import {IE_VERSION} from './browser';
...
   log.history.push(args);
 
-  // add console prefix after adding to history
-  args.unshift('VIDEOJS:');
+  // If there's no console then don't try to output messages, but they will
+  // still be stored in `log.history`.
+  //
+  // This mainly prevents cases where old IE versions don't expose a `console`
+  // object unless the console is actually open.
+  const fn = console && console[type];
```
BrandonOCasey: seems like we may still want the noop just in case here `|| function() {}`

misteroneill: The noop, unfortunately, was the cause of the problem because then it _had_ an `apply` method and reached the line where it tried to `apply(console, args)` and that's where it blows up.

BrandonOCasey:
#3686

PR #3686, opened 2016-10-14
Apply window.console instead of console, which is no longer defined

misteroneill:
## Description
This fixes the potential for a completely broken player under the following conditions:

- IE < 10
- `console` is closed, so undefined
- Call `videojs.log` or one of its methods

## Specific Changes proposed
No longer attempts to `apply(console, args)` because `console` is no longer defined as a local variable after e85c1c039153f4d031b602fa8dc6d4aa2130427f so the browser looks up the scope until it reaches the global scope and, in IE9, because there is no console unless it's open... throws a "console is undefined" error.

## Requirements Checklist

### DIFF
--- a/src/js/utils/log.js
+++ b/src/js/utils/log.js
@@ -62,7 +62,7 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   if (!fn.apply) {
     fn(args);
   } else {
-    fn[Array.isArray(args) ? 'apply' : 'call'](console, args);
+    fn[Array.isArray(args) ? 'apply' : 'call'](null, args);
   }
 };
 
--- a/src/js/utils/log.js
+++ b/src/js/utils/log.js
@@ -19,14 +19,6 @@ let log;
  */
 export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 11) => {
 
-  // If there's no console then don't try to output messages, but they will
-  // still be stored in `log.history`.
-  //
-  // Was setting these once outside of this function, but containing them
-  // in the function makes it easier to test cases where console doesn't exist
-  // when the module is executed.
-  const fn = window.console && window.console[type] || function() {};
-
   if (type !== 'log') {
 
     // add the type to the front of the message when it's not "log"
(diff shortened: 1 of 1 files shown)


forbesjo:
Looks good to me

gkatsev:
This doesn't seem right to me. The log methods expect to the called on the `console` object, so, if you apply `null` there, it's not getting called on `console` and may not work.

misteroneill:
You're right. Firefox doesn't like it. I'm gonna go back to what I was doing in my other PR.

src/js/utils/log.js
```diff
@@ -39,6 +31,19 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
...
+  //
+  // Was setting these once outside of this function, but containing them
+  // in the function makes it easier to test cases where console doesn't exist
+  // when the module is executed.
+  const fn = window.console && window.console[type];
+
+  // Bail out if there's no console.
+  if (!fn) {
+    return;
+  }
```
misteroneill: At this point, we've already stored the arguments in history; so, we can safely bail out if there's nothing to do. This is cheaper than potentially running `JSON.stringify` and calling a no-op.

BrandonOCasey: sounds like a better idea

src/js/utils/log.js
```diff
@@ -62,7 +67,7 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
   if (!fn.apply) {
     fn(args);
   } else {
-    fn[Array.isArray(args) ? 'apply' : 'call'](console, args);
+    fn[Array.isArray(args) ? 'apply' : 'call'](window.console, args);
```
BrandonOCasey: I assume using global console caused issues as well? and what about just calling apply and making args an array be default?

misteroneill: Yeah, the implied global `console` combined with the noop (which had `apply()`) was causing the problem in the first place.

As to the other change, I think I want to keep this fix pretty minimal. More changes are more opportunities for bugs!

BrandonOCasey: ok

PR #3687, opened 2016-10-14
Use npm-run instead of shelljs-nodecli

### LINKED ISSUE #3683: Loading "Gruntfile.js" tasks...ERROR
screen shot 2016-10-14 at 16 31 09

I tried to install the "conventional-changelog" but still not work.

misteroneill:
## Description
Fixes #3683 

## Specific Changes proposed
Stop using shelljs-nodecli and start using npm-run! The reason this fails is that shelljs-nodecli makes the assumption that the package and its executable are the same name (e.g. the `grunt` command is exposed by the `grunt` package).

## Requirements Checklist


gkatsev:
I guess this is fine? I've never had problems with shelljs-nodecli before. And it's been working fine for me, so, not sure exactly what the problem was.

misteroneill:
The problem is that the package name and the CLI name don't match. shelljs-nodecli assumes they do; so, when you look for `conventional-changelog` it looks for `node_modules/conventional-changelog/package.json` and fails to find it and gives up.

BrandonOCasey:
is there any reason not to switch to vanilla `shelljs` and use `shelljs.exec('node_modules/.bin/*')`?

misteroneill:
I don't think that would work on Windows and I believe there are cases where `node_modules/.bin` isn't the right path? Hence the existence of the `npm bin` command.

BrandonOCasey:
LGTM

gkatsev:
fwiw, shelljs-nodecli *is* picking up the local conventional-changelog bin for me...

gkatsev:
Or maybe not.

build/grunt.js
```diff
@@ -32,7 +32,7 @@ module.exports = function(grunt) {
       release: {
         tag_name: 'v'+ version.full,
         name: version.full,
-        body: nodeCli.exec('conventional-changelog', '-p videojs', {silent: true}).output
+        body: npmRun.execSync('conventional-changelog -p videojs', {silent: true})
```
gkatsev: This needs an `encoding: 'utf8'` options because a buffer is returned by default.

PR #3690, opened 2016-10-17
Advanced Plugins for 6.0

misteroneill:
# Advanced/Class-based Plugins for 6.0

As part of 6.0, we want to make some improvements to plugins while still supporting the current style of plugins for the most part. There are some minor backward-incompatibilities, but nothing that's likely to break existing plugins.

Though this is technically a breaking change, _the old stuff should still work!_
## Goals

These are the broad feature goals of this effort. The overarching approach will be to maintain simplicity throughout the implementation, avoiding assumptions and opinions whenever possible. For now, we are providing a foundation not a framework, but something both we and our users can build on easily.
- Maintain support for "basic" plugins (i.e. a function that is added to `Player.prototype`).
- Add a foundation for "enhanced" plugins with:
  - Support for the basic lifecycle of initialization and disposal.
  - Support for per-player state management.
- For both types of plugins, add the following:
  - Support for querying per-player plugin activation state.
  - Support for a conventional `VERSION` property.
## Specific Changes

Changes that are backward-incompatible are marked with **[BI]**.
- Add `videojs.registerPlugin` and `videojs.registerPlugins` and deprecate `videojs.plugin`.
  - **[BI]** Registering a plugin with a name that already exists or conflicts with a `Player.prototype` method will now throw an error.
- Add `videojs.getPlugin`, `videojs.getPlugins`, and `videojs.getPluginVersion`.
- Add `Player#usingPlugin` to detect if a plugin has been initialized on a player.
- Create `Plugin` base class for enhanced plugin functionality.
  - This includes static methods for de-registering plugins, but those are not globally exposed and mostly used for testing.
- Introduce semi-formal concept of mixins, which are functions that mutate an object in some way to add functionality:
  - Create `mixins/evented`, which provides objects with methods from `Events` (`on`, `one`, `off`, and `trigger`). This uses an `eventBusEl_` as an event bus because the `Events` module doesn't play very nicely with non-element objects.
  - Create `mixins/stateful`, which provides objects with a React-like `state` property and `setState` method - as well as triggering a `"statechanged"` event if the object has a `trigger` method.
## Requirements Checklist
  - [x] Docs/guides updated


src/js/plugins.js
```diff
@@ -1,17 +1,101 @@
...
+ *         be used as the init for your plugin.
+ */
+const registerPlugin = function(name, plugin) {
+  const normalized = normalizePlugin(plugin);
+
+  if (typeof name !== 'string' || !name.trim()) {
+    throw new Error('illegal plugin name; must be non-empty string');
+  }
+
+  if (plugins[name]) {
```
gkatsev: should also check `Player.prototype[name]`

dmlap:
I kind of like `player.plugin` and `player.foo` where `foo` is a plugin. Why should we deprecate these?

misteroneill:
@dmlap You mean `videojs.plugin`? Assuming that's the case, I felt like it was inconsistent with similar methods (`registerComponent` and `registerTech`).

As for `player.foo`, I'm not married to the new implementation, but the intent was to minimize (or eliminate) the surface area for conflicts between plugins and `Player.prototype` methods. With this new scheme, you could have a plugin called `play` or `pause`; the only limitation is that it can't share the same name as another plugin.

That said, what about an alternative where the plugin was still exposed as `player.foo` (for the initialization function) and then the other methods were attached to that on init, e.g. `player.foo.dispose()`. That's a pattern we've used in a few plugins.

gkatsev:
Fwiw, I don't think having a plugin called `play` is something we should support as long as it gets attached to the player prototype.

misteroneill:
For sure. That's more of a long-term design goal, since we're still mutating the `Player.prototype`.

gkatsev:
I think 6.0 is a good time to drop support for that.

heff:
Naming it registerPlugin makes sense to me. 👍 

The biggest question I see people run into with complex plugins is storing state/properties for just the plugin, and it'd be great if we could answer that question with this update. At the same time I would love to not lose the simplicity of the current API and the ability to create first-class functions on the player. e.g.

```js
// complex
videojs.registerPlugin('fooState', {
  init: function(options){
    this.foo = options.foo;

    // this.player_ automatically added to instance object 
    this.player_.play();
  },
  isFoo: function(){
    return this.foo;
  }
});

player.fooState({ foo: true });
player.fooState.isFoo();

// simple
videojs.registerPlugin('jumpTo10', function(){
    this.currentTime(10);
});

// Allowing plugin authors to do this is powerful, e.g. jQuery
player.jumpTo10();
```

That feels perty to me.

misteroneill:
@heff Thanks for the feedback. It seems we're on the same page here because what I've implemented is very similar (I think) what you described!

I can see keeping the jQuery-style plugins - I'm happy to un-deprecate that. And modifying the API to be more similar to what we have now as both you and @dmlap brought up.

This proposal also includes a concept of state for each plugin/player combination. Which can be understood as behaving as a very lightweight model/view implementation. It works kind of like this currently:

```js
videojs.registerPlugin('foo', {
  init: function(options) {
    this.foo_ = options.foo;
    this.player_; // this is the player.
    this.state_; // this is the state.
  },
  isFoo: function() {
    return this.foo_;
  }
});

player.plugin('foo').init({foo: true});
player.plugin('foo').isFoo(); // true

player.plugin('foo').on('statechange', function() {
  alert('foo changed to ' + this.plugin('foo').state().foo + '!');
  // Maybe do some DOM manipulation or something here?
});

player.plugin('foo').state({foo: false}); // "foo changed!"
```

That's without making any changes to what's in the PR at the moment. Like I said, the `player.plugin('foo')` calls will likely be converted to `player.foo`.

misteroneill:
I've been tinkering with replacing `player.plugin('foo')` with `player.foo`. And it's definitely do-able, but I worry about some of the unintuitive corner cases.

Particularly around the `this` context. If we try to stick closer to the current API for a plugin, I feel like this is the expectation:

```js
player.foo(); // this === player
player.foo.isFoo(); // this === player.foo
```

That's all well and good, but when I create a plugin like this:

```js
videojs.registerPlugin('foo', {
  init: function() {},
  isFoo: function() {}
});
```

My natural inclination would be that inside any of the functions I passed in for a "complex" plugin, `this` would be the plugin object.

So, there's a mismatch of expectations there. Either we make the `player.foo()` call bound to itself for consistency with the expectations from registering the plugin or we make `player.foo()` bound to the player and break that expectation.

misteroneill:
In general, my philosophy is to avoid tricky things like this. As it stands, my new implementation does a lot of confusing things, like creating a wrapper function on the `Player.prototype` then replacing that wrapper function with another wrapper function on a `player` itself. A wrapper which is, of course, bound to itself! 😆 

```js
  // Create a Player.prototype-level plugin wrapper that only gets called
  // once per player.
  plugins[name] = Player.prototype[name] = function(...firstArgs) {

    // Replace this function with a new player-specific plugin wrapper.
    const wrapper = function(...args) {
      if (this.active_) {
        this.teardown();
      }

      this.trigger('beforesetup');
      plugin.setup(...args);
      this.trigger('setup');
    };

    // Wrapper is bound to itself to preserve expectations.
    this[name] = Fn.bind(wrapper, wrapper);
```

misteroneill:
I'm tinkering with API ideas [in a JSBin](http://jsbin.com/mupani/edit?js).

heff:
I think it's fine to distinguish simple vs complex plugins, where the context of a simple function is the player and the context of complex plugin functions is the plugin. That's not a hard concept to learn, and the APIs are different enough.

```
// what happens here? should we magically dispose? or throw? or fail silently?
//    we could dispose OUR stuff but not call `dispose()`
//    we could provide an option to disposeOnSetup?
player.enhancedFoo();
player.enhancedFoo();
```
Ideally I think a console error logged. If people want to reset a plugin they need to call dispose first. If we try to magically support that I think it would actually lead to more confusion.

mmcc:
As we discussed quasi-IRL, I think it makes a lot of sense to follow the React API (particularly ES6) for plugins. I think we were circling around something like this:

```
class EnhancedPlugin extends Videojs.Plugin {
  constructor (opts) {
    super(opts);

    // default state
    this.state = {
      bar: 'baz'
    };
  }
  
  dispose () { }

  bar () {
    return this.state.bar;
  }

  doSomethingFun () {
    this.setState({bar: 'we are having fun now'});
  }
}
```

misteroneill:
Yep, that's very close to what I've got in mind. I'm going to be updating this PR this week with the changes we discussed.

misteroneill:
At this point, Component has been reimplemented as an evented object, but that change is going to be rolled back to keep the scope of this a bit more sane.

misteroneill:
And the `Component` changes are rolled back! I'll open a separate PR for those changes.

src/js/plugin.js
```diff
@@ -0,0 +1,281 @@
...
+  /**
+   * Register multiple plugins via an object where the keys are plugin names
+   * and the values are sub-classes of `Plugin` or anonymous functions for
+   * basic plugins.
+   *
+   * @param  {Object} plugins
+   * @return {Object}
+   *         An object containing plugins that were added.
+   */
+  static registerPlugins(plugins) {
```
dmlap: This doesn't seem to add much for end users that they couldn't achieve by repeating `registerPlugin()` a couple times in a row.

src/js/tracks/text-track-settings.js
```diff
@@ -1,168 +1,220 @@
 /**
  * @file text-track-settings.js
  */
-import Component from '../component';
```
dmlap: The changes to this file look nice but it seems unrelated to this PR.

test/unit/plugins.test.js
```diff
@@ -1,214 +0,0 @@
...
-  assert.deepEqual(order,
-            expectedOrder.slice(0, order.length),
-            'plugins should receive events in order of ' +
-            'initialization, until stopImmediatePropagation');
-
-  assert.equal(order.length, 1, 'only one event listener should have triggered');
-  player.dispose();
-});
-
-QUnit.test('Plugin that does not exist logs an error', function(assert) {
```
thijstriemstra: @misteroneill is this test replicated elsewhere since it's removed here? I have a hard time finding it or something similar.

src/js/plugin.js
misteroneill: Agreed. Will remove this and `deregisterPlugins`, which was really only added for my own convenience.

docs/guides/plugins.md
```diff
@@ -1,59 +1,344 @@
...
+
+### Register a Class-Based Plugin
+
+The registration process for class-based plugins is identical to [the process for basic plugins](#register-a-basic-plugin).
+
+```js
+videojs.registerPlugin('examplePlugin', ExamplePlugin);
+```
+
+### Setting up a Class-Based Plugin
```
misteroneill: I think that makes sense. I'll reshuffle this doc a bit to eliminate some of the duplication.

test/unit/plugins.test.js
misteroneill: This is now tested in player.test.js, under the "options: plugins" test. I've changed so it throws now instead of just logging. Failing loud in this case seems like a good idea.

misteroneill:
I made a few changes to the PR both in response to feedback and not. Those in the "not" category are:

- `Player#hasPlugin` is added. This is distinct from `Player#usingPlugin` slightly - it will report if the plugin exists rather than if it's been set up on the player.
  - Also, populating these methods is now done in `plugin.js`. Though, they are documented in `player.js` and exist as no-op there. The reason for this is that I wanted to avoid circular dependencies between these two modules.
- Passing a non-existent plugin in player options (e.g. `videojs('x', {plugins: {y: {}}})`) will now throw instead of simply logging an error.
- `Plugin.getPlugin()` will no longer return `Player.prototype` methods/properties. It didn't make sense there.

docs/guides/plugins.md
```diff
@@ -1,59 +1,324 @@
...
-        examplePlugin: {
-          exampleOption: true
-        }
-      }
+var Plugin = videojs.getPlugin('plugin');
+
+var ExamplePlugin = videojs.extend(Plugin, {
+
+  constructor: function(player, options) {
+    Plugin.prototype.call(this, player, options);
```
mister-ben: `Plugin.call(...)`

test/unit/plugins.test.js
thijstriemstra: Looks good, thanks!

misteroneill:
Yeah @BrandonOCasey

src/js/plugin.js
```diff
@@ -0,0 +1,342 @@
...
+    this.trigger('dispose');
+    this.off();
+
+    // Eliminate any possible sources of leaking memory by clearing up references
+    // between the player and the plugin instance and nulling out the plugin's
+    // state and replacing methods with a function that throws.
+    player[PLUGIN_CACHE_KEY][name] = false;
+    this.player = this.state = null;
+
+    this.dispose = () => {
```
BrandonOCasey: This seems like it is too much protection for the user, it could be overly restrictive to some plugins.

test/unit/mixins/evented.test.js
```diff
@@ -0,0 +1,455 @@
...
+});
+
+QUnit.test('on() errors', function(assert) {
+  const target = this.targets.a = evented({});
+
+  assert.throws(
+    function() {
+      target.on();
+    },
+    new Error('invalid event type; must be a non-empty string or array')
```
BrandonOCasey: maybe define this error up at the top for re-use

test/unit/mixins/evented.test.js
```diff
@@ -0,0 +1,455 @@
...
+    new Error('eventBusKey "foo" does not refer to an element'),
+    'throws if the target does not have an element at the supplied key'
+  );
+});
+
+QUnit.test('on() errors', function(assert) {
+  const target = this.targets.a = evented({});
+
+  assert.throws(
+    function() {
```
BrandonOCasey: maybe use a shorthand function so that this can fit on one line better?

test/unit/mixins/evented.test.js
```diff
@@ -0,0 +1,455 @@
...
+      target.on();
+    },
+    new Error('invalid event type; must be a non-empty string or array')
+  );
+
+  assert.throws(
```
BrandonOCasey: A lot of these assertions need descriptions

test/unit/mixins/evented.test.js
```diff
@@ -0,0 +1,455 @@
...
+
+  assert.ok(Obj.isObject(target), 'the target is still an object');
+  assert.ok(Dom.isEl(target.eventBusEl_), 'the target has an event bus element');
+  assert.strictEqual(typeof target.off, 'function', 'the target has an off method');
+  assert.strictEqual(typeof target.on, 'function', 'the target has an on method');
+  assert.strictEqual(typeof target.one, 'function', 'the target has a one method');
+  assert.strictEqual(typeof target.trigger, 'function', 'the target has a trigger method');
+});
+
+QUnit.test('evented() with exclusions', function(assert) {
```
BrandonOCasey: Maybe it would be worth it to make this function loop over all possible exclusions? If not then this should be changed to 'evented() with `one` function excluded'.

test/unit/mixins/evented.test.js
```diff
@@ -0,0 +1,455 @@
...
+  beforeEach() {
+    this.targets = {};
+  },
+
+  afterEach() {
+    Object.keys(this.targets).forEach(k => this.targets[k].trigger('dispose'));
+  }
+});
+
+QUnit.test('evented() mutations', function(assert) {
```
BrandonOCasey: I think that the name of this test should be something like "evented() An object is mutated as expected"

test/unit/mixins/evented.test.js
```diff
@@ -0,0 +1,455 @@
...
+
+  assert.throws(
+    function() {
+      target.off(evented({}), 'x', null);
+    },
+    new Error('invalid listener; must be a function')
+  );
+});
+
+QUnit.test('a.on("x", fn)', function(assert) {
```
BrandonOCasey: I think a lot of the test names after this need to be in more plain english

test/unit/mixins/stateful.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+  assert.ok(spy.called, 'the "statechanged" event occurred');
+
+  const event = spy.firstCall.args[0];
+
+  assert.strictEqual(event.type, 'statechanged');
+  assert.strictEqual(event.changes, changes, 'the changes object is sent along with the event');
+});
+
+QUnit.test('setState() without changes', function(assert) {
```
BrandonOCasey: += " does not trigger the `statechanged` event"

test/unit/mixins/stateful.test.js
```diff
@@ -0,0 +1,65 @@
...
+  assert.strictEqual(typeof target.setState, 'function', 'the target has a setState method');
+});
+
+QUnit.test('stateful() with defaults', function(assert) {
+  const target = stateful({}, {foo: 'bar'});
+
+  assert.strictEqual(target.state.foo, 'bar', 'the default properties are added to the state');
+});
+
+QUnit.test('setState()', function(assert) {
```
BrandonOCasey: setState() works as expected

test/unit/mixins/stateful.test.js
```diff
@@ -0,0 +1,65 @@
...
+  assert.strictEqual(typeof stateful, 'function', 'the mixin is a function');
+  assert.strictEqual(stateful(target), target, 'returns the target object');
+
+  assert.ok(Obj.isObject(target), 'the target is still an object');
+  assert.ok(Obj.isPlain(target.state), 'the target has a state');
+  assert.strictEqual(Object.keys(target.state).length, 0, 'the target state is empty by default');
+  assert.strictEqual(typeof target.setState, 'function', 'the target has a setState method');
+});
+
+QUnit.test('stateful() with defaults', function(assert) {
```
BrandonOCasey: stateful() with default state passed in

BrandonOCasey: It may be worth it to have stateful with no default state as well?

test/unit/mixins/stateful.test.js
```diff
@@ -0,0 +1,65 @@
+/* eslint-env qunit */
+import sinon from 'sinon';
+import EventTarget from '../../../src/js/event-target';
+import stateful from '../../../src/js/mixins/stateful';
+import * as Obj from '../../../src/js/utils/obj';
+
+QUnit.module('mixins: stateful');
+
+QUnit.test('stateful() mutations', function(assert) {
```
BrandonOCasey: stateful() mutates an object as expected.

test/unit/video.test.js
```diff
@@ -117,19 +116,13 @@ QUnit.test('should add the value to the languages object with lower case lang co
...
-
-  videojs.plugin(pluginName, pluginFunction);
-
-  const player = TestHelpers.makePlayer();
-
-  assert.ok(player.foo, 'should exist');
-  assert.equal(player.foo, pluginFunction, 'should be equal');
-  player.dispose();
+QUnit.test('should expose plugin functions', function(assert) {
+  assert.strictEqual(typeof videojs.registerPlugin, 'function');
```
BrandonOCasey: could use descriptions

src/js/mixins/evented.js
```diff
@@ -0,0 +1,377 @@
+/**
+ * @file mixins/evented.js
+ */
+import * as Dom from '../utils/dom';
+import * as Fn from '../utils/fn';
+import * as Events from '../utils/events';
+
+/**
```
misteroneill: Yes, @gkatsev asked me to move it to another PR.

src/js/video.js
```diff
@@ -433,51 +433,66 @@ videojs.mergeOptions = mergeOptions;
...
+ * Register a Video.js plugin
  *
- *       var player = this;
- *       var alertText = myPluginOptions.text || 'Player is playing!'
+ * @param  {String} name
+ * @param  {Function} plugin
+ *         A sub-class of `Plugin` or an anonymous function for basic plugins.
+ * @return {Function}
+ */
+videojs.registerPlugin = (name, plugin) => Plugin.registerPlugin(name, plugin);
```
misteroneill: I had a reason, but I can't remember what it was. So, it couldn't have been too compelling. 😆  I'll change these to be references.

src/js/video.js
```diff
@@ -433,51 +433,66 @@ videojs.mergeOptions = mergeOptions;
...
- *       text: 'Plugin added later!'
- *     });
- *     // Click play
- *     // --> Should alert 'Plugin added later!'
- * ```
+ * @param  {Object} plugins
+ * @return {Object}
+ *         An object containing plugins that were added.
+ */
+videojs.registerPlugins = (plugins) => Plugin.registerPlugins(plugins);
```
misteroneill: Addressed in another comment.

src/js/video.js
```diff
@@ -433,51 +433,66 @@ videojs.mergeOptions = mergeOptions;
...
+  log.warn('videojs.plugin() is deprecated; use videojs.registerPlugin() instead');
+  return Plugin.registerPlugin(name, plugin);
+};
+
+/**
+ * Get an object containing all available plugins.
+ *
+ * @return {Object}
+ */
+videojs.getPlugins = () => Plugin.getPlugins();
```
misteroneill: Addressed in another comment.

src/js/video.js
```diff
@@ -433,51 +433,66 @@ videojs.mergeOptions = mergeOptions;
...
+
+/**
+ * Get a single plugin by name.
+ *
+ * @param  {String} name
```
misteroneill: Thanks for the reminders. 👍

docs/guides/plugins.md
```diff
@@ -1,59 +1,324 @@
...
+
+#### The Player Plugin Name Property
+
+Both basic plugins and class-based plugins are set up by calling a method on a player with a name matching the plugin (e.g., `player.examplePlugin()`).
+
+However, with class-based plugins, this method acts like a factory function and it is _replaced_ for the current player by the plugin class instance:
+
+```js
+// `examplePlugin` has not been called, so it is a factory function.
+player.examplePlugin();
```
misteroneill: Hmmm, interesting idea.

src/js/player.js
```diff
@@ -256,11 +256,11 @@ class Player extends Component {
     if (options.plugins) {
       const plugins = options.plugins;
 
-      Object.getOwnPropertyNames(plugins).forEach(function(name) {
+      Object.keys(plugins).forEach(function(name) {
         if (typeof this[name] === 'function') {
           this[name](plugins[name]);
         } else {
-          log.error('Unable to find plugin:', name);
+          throw new Error(`plugin "${name}" does not exist`);
```
misteroneill: I feel like mis-configuration of the player is something we ought to be really loud about.

test/unit/mixins/evented.test.js
misteroneill: I don't think we ought to loop over all possible exclusions because that's... a lot. Especially when you consider that it can be an array of zero to four elements. I'll change it to be a reasonable subset.

misteroneill: On second thought, I'm removing the `exclude` option because the more I tinkered with tests, the more I realized it could be a source of bugs.

test/unit/mixins/evented.test.js
misteroneill: Do they? It's an optional argument...

test/unit/mixins/evented.test.js
misteroneill: I suppose it's the style used in most places. I've still never really drank the BDD Kool-aid entirely. 😄

src/js/mixins/evented.js
```diff
@@ -0,0 +1,377 @@
...
+ * Returns whether or not an object has had the evented mixin applied.
+ *
+ * @private
+ * @param  {Object} object
```
misteroneill: I think I've done a pass that adds missing descriptions for params and returns.

test/unit/video.test.js
misteroneill: I dunno, I feel like these are pretty self-explanatory. Is there some value I'm not understanding in providing descriptions for things like this?

misteroneill: Nevermind, I've updated all the tests.

docs/guides/plugins.md
```diff
@@ -1,59 +1,324 @@
...
+    super(player, options);
+
+    // Whenever the player emits a playing or paused event, we update the
+    // state if necessary.
+    this.on(player, ['playing', 'paused'], this.updateState);
+    this.on('statechanged', this.logState);
+  }
+
+  dispose() {
+    super();
```
misteroneill: Good catch, thanks.

docs/guides/plugins.md
```diff
@@ -1,59 +1,324 @@
...
+
+Now that we have a function that does something with a player, all that's left is to register the plugin with Video.js:
 
 ```js
-    videojs.plugin('examplePlugin', examplePlugin);
+videojs.registerPlugin('examplePlugin', examplePlugin);
 ```
 
-From this point on, your plugin will be added to the Video.js prototype and will show up as a property on every instance created. Make sure you choose a unique name that doesn't clash with any of the properties already in Video.js. Which leads us to...
+The only stipulation with the name of the plugin is that it cannot conflict with any existing player method. After that, any player will automatically have an `examplePlugin` method on its prototype!
```
misteroneill: Yeah, we do.

docs/guides/plugins.md
```diff
@@ -1,59 +1,324 @@
...
+```
+
+For now, this example class-based plugin does the exact same thing as the basic plugin described above - not to worry, we will make it more interesting as we continue!
+
+### Register a Class-Based Plugin
+
+The registration process for class-based plugins is identical to [the process for basic plugins](#register-a-basic-plugin).
+
+```js
+videojs.registerPlugin('examplePlugin', ExamplePlugin);
```
misteroneill: Sure, I'm adding a note after this code block that "classes" are still functions.

package.json
```diff
@@ -102,6 +102,7 @@
     "load-grunt-tasks": "^3.1.0",
     "markdown-table": "^1.0.0",
     "npm-run": "^4.1.0",
+    "permute": "^1.0.0",
```
BrandonOCasey: is this used?

docs/guides/plugins.md
BrandonOCasey: this has been resolved.

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+
+    // Bind all plugin prototype methods to this object.
+    Object.keys(Plugin.prototype).forEach(k => {
+      if (typeof this[k] === 'function') {
+        this[k] = Fn.bind(this, this[k]);
+      }
+    });
+
+    this.on('statechanged', this.handleStateChanged);
+    player.one('dispose', this.dispose);
```
BrandonOCasey: This comment was hidden by some code, changes. So here it is again: "why only one? Do plugins need to be initialized again every time the player is disposed?"

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+    this.trigger('dispose');
+    this.off();
+
+    // Eliminate any possible sources of leaking memory by clearing up
+    // references between the player and the plugin instance and nulling out
+    // the plugin's state and replacing methods with a function that throws.
+    player[PLUGIN_CACHE_KEY][name] = false;
+    this.player = this.state = null;
+
+    this.dispose = () => {
```
BrandonOCasey: Bringing back another comment "This seems like it is too much protection for the user, it could be overly restrictive to some plugins."

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    if (typeof plugin !== 'function') {
+      throw new Error(`illegal plugin for "${name}", must be a function, was ${typeof plugin}`);
+    }
+
+    pluginStorage[name] = plugin;
+
+    // Add a player prototype method for all sub-classed plugins (but not for
+    // the base Plugin class).
```
BrandonOCasey: I have two ideas here, maybe the API change would be too drastic, but here they are anyway:

would it be better to just make plugin registration work more like Components, ie: store Plugins that are currently in use on Plugin.plugins_[name]. Then force the user to initialize plugins through something like player.turnPluginOn('name');.
Just initialize right when the player starts (before ready). Then the plugin can attach to player events to do things that it needs to. Then the only way to get a plugin (which will always be an instance?) would be through player.getPlugin

src/js/plugin.js
misteroneill: Yeah. I mean, if a player is disposed, it's "gone" (for all intents and purposes). I don't think it makes a difference if it's `one` or `on`. Maybe I'll just use `on` for simplicity, since the listener will get removed anyway when the player is disposed.

src/js/plugin.js
BrandonOCasey: yeah it does make sense to store them on an actual private property.

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+/**
+ * Makes an object "evented" - granting it methods from the `Events` utility.
+ *
+ * By default, this adds the `off`, `on`, `one`, and `trigger` methods, but
+ * exclusions can optionally be made.
+ *
+ * @param  {Object} target
+ *         The object to which to add event methods.
+ *
+ * @param  {Object} [options]
```
BrandonOCasey: should be `[options={}]` since it has a default.

src/js/mixins/stateful.js
```diff
@@ -0,0 +1,91 @@
...
+ * @file mixins/stateful.js
+ */
+import * as Fn from '../utils/fn';
+import log from '../utils/log';
+import * as Obj from '../utils/obj';
+
+/**
+ * Set the state of an object by mutating its `state` object in place.
+ *
+ * @param  {Object|Function} next
```
BrandonOCasey: perhaps this should be names `newState` rather than `next`?

src/js/mixins/stateful.js
```diff
@@ -0,0 +1,91 @@
...
+      changes,
+      type: 'statechanged'
+    });
+  }
+
+  return changes;
+};
+
+/**
+ * Makes an object "stateful" - granting it a `state` property containing
```
BrandonOCasey: Should this state that the target must already be evented? Should we also make stateful objects evented in this function if they are not yet evented?

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+    this.player = player;
+
+    if (this.constructor === Plugin) {
+      throw new Error('Plugin must be sub-classed; not directly instantiated.');
+    }
+
+    // Make this object evented, but remove the added `trigger` method so we
+    // use the prototype version instead.
+    evented(this);
+    delete this.trigger;
```
BrandonOCasey: isn't there a way to exclude trigger from being added?

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+     *
+     * In all cases, an object containing the following properties is passed as a
+     * second argument to event listeners:
+     *
+     * - `name`: The name of the plugin that was set up.
+     * - `plugin`: The raw plugin function.
+     * - `instance`: For class-based plugins, the instance of the plugin sub-class,
+     *   but, for basic plugins, the return value of the plugin invocation.
+     *
+     * @event pluginsetup
```
BrandonOCasey: * should be `Plugin#pluginsetup` for consistency, and remove the `@memberof` below
* The objects properties can be documented using `@property` see this [example](https://github.com/videojs/video.js/blob/master/src/js/component.js#L138)
* Should have a custom type that is defined in this file, and returned by `getEventHash_`. This will prevent it from being typed twice. See an example [here](https://github.com/videojs/video.js/blob/master/src/js/utils/url.js#L31)

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+};
+
+class Plugin {
+
+  /**
+   * Plugin constructor.
+   *
+   * Subclasses should call `super` to ensure plugins are properly initialized.
+   *
+   * @param {Player} player
```
BrandonOCasey: should have `@fires Plugin#pluginsetup`

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+   * This returns that object or mutates an existing hash.
+   *
+   * @private
+   * @param  {Object} [hash={}]
+   *         An object to be used as event an event hash.
+   *
+   * @return {Object}
+   *         The event hash object with, at least, the following properties:
+   *
+   *         - `instance`: The plugin instance on which the event is fired.
```
BrandonOCasey: Use the custom typedef that should be defined for use in the `Player#pluginsetup` event.

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+  }
+
+  /**
+   * Triggers an event on the plugin object.
+   *
+   * @param  {Event|Object|string} event
+   *         A string (the type) or an event object with a type attribute.
+   *
+   * @param  {Object} [hash={}]
+   *         Additional data hash to pass along with the event. For plugins,
```
BrandonOCasey: Custom Typedef

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+   */
+  trigger(event, hash = {}) {
+    return Events.trigger(this.eventBusEl_, event, this.getEventHash_(hash));
+  }
+
+  /**
+   * Handles "statechanged" events on the plugin. No-op by default, override by
+   * subclassing.
+   *
+   * @param {Event} e
```
BrandonOCasey: add `@abstract` here to mention that it should be overridden.

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+   *        An object describing changes that occurred with the "statechanged"
+   *        event.
+   */
+  handleStateChanged(e) {}
+
+  /**
+   * Disposes a plugin.
+   *
+   * Subclasses can override this if they want, but for the sake of safety,
+   * it's probably best to subscribe the "dispose" event.
```
BrandonOCasey: add `@fires Plugin#dispose`

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+   */
+  dispose() {
+    const {name, player} = this;
+
+    /**
+     * Signals that a class-based plugin is about to be disposed.
+     *
+     * @event dispose
```
BrandonOCasey: * `@event Plugin#dispose` 
* remove `@memberof` & `@instance`
* add in `@type {EventTarget~Event}` as it is just a standard event

src/js/plugin.js
```diff
@@ -0,0 +1,436 @@
...
+   * @param  {string} name
+   *         The name of a plugin.
+   *
+   * @return {string}
+   *         The plugin's version or an empty string.
+   */
+  static getPluginVersion(name) {
+    const plugin = getPlugin(name);
+
+    return plugin && plugin.VERSION || '';
```
BrandonOCasey: do we want to enforce this? IE: throw if a plugin does not have a VERSION?

src/js/plugin.js
misteroneill: Given more thought, I think you're right on this. What made me convince myself was realizing that this could complicate plugins that want to override `dispose` and call `super.dispose()`.

docs/guides/plugins.md
```diff
@@ -1,59 +1,336 @@
...
+```
+
+> **Note:** Because ES6 classes are syntactic sugar on top of existing constructor function and prototype architecture in JavaScript, in all cases `registerPlugin`'s second argument is a function.
+
+### Key Differences from Basic Plugins
```
misteroneill: I think people are familiar with basic plugins and, if they are, will have certain expectations as to how they work. These could be subtle differences, so I wanted to call them out here.

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+ * @param  {string|Array} type
+ *         One or more event type(s).
+ *
+ * @param  {Function} listener
+ *         A listener function.
+ *
+ * @param  {string} [method="on"]
+ *         The event binding method to use.
+ */
+const listen = (target, type, listener, method = 'on') => {
```
misteroneill: Sounds good. Also moving `method` to be the second argument to better reflect the usage.

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+ * @type {Object}
+ */
+const mixin = {
+
+  /**
+   * Add a listener to an event (or events) on this object or another evented
+   * object.
+   *
+   * @param  {string|Array|Element|Object} targetOrType
+   *         If this is a string or array, it represents an event type(s) and
```
misteroneill: I'm rephrasing this a bit.

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+
+      // Use the same function ID as the listener so we can remove it later
+      // it using the ID of the original listener.
+      removeRemoverOnTargetDispose.guid = listener.guid;
+
+      listen(this, 'dispose', removeListenerOnDispose);
+      listen(target, 'dispose', removeRemoverOnTargetDispose);
+    }
+
+    return this;
```
misteroneill: Will remove it now that we seem to have decided to remove method chaining across the board.

src/js/mixins/stateful.js
misteroneill: Changing to `stateUpdates` (`newState` seems to imply that it replaces `state`, which it doesn't).

src/js/plugin.js
misteroneill: There was - `evented(this, {exclude: ['trigger']})` - but I felt like it was less clear. This feels more clear as to what's going on. I could be persuaded to go back, but maybe in another PR?

src/js/mixins/stateful.js
misteroneill: I don't think so. That creates a tight coupling where my idea of mixins are that they're composable and modular.

src/js/plugin.js
misteroneill: I don't think so. I think we want to stay relatively loose here.

src/js/mixins/stateful.js
misteroneill: No, it won't. It seems to me the `handle*` methods should be considered a convention for auto-binding to an event.

If someone makes an object stateful, but not evented, it doesn't make sense to treat it as if an event had fired on it.

misteroneill:
I believe all issues are addressed either through changes or responses. Ready for further review as needed.

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+  if (eventBusKey) {
+    if (!target[eventBusKey].nodeName) {
+      throw new Error(`The eventBusKey "${eventBusKey}" does not refer to an element.`);
+    }
+    target.eventBusEl_ = target[eventBusKey];
+  } else {
+    target.eventBusEl_ = Dom.createEl('span', {className: 'vjs-event-bus'});
+  }
+
+  ['off', 'on', 'one', 'trigger'].forEach(name => {
```
gkatsev: What about `Objecy.keys(mixin)` that way we're not repeating ourselves? And if we ever add another thing to mixin, it'll automatically get picked up.

docs/guides/plugins.md
```diff
@@ -1,58 +1,345 @@
...
-    videojs.plugin('examplePlugin', examplePlugin);
+videojs.registerPlugin('examplePlugin', examplePlugin);
 ```
 
-From this point on, your plugin will be added to the Video.js prototype and will show up as a property on every instance created. Make sure you choose a unique name that doesn't clash with any of the properties already in Video.js. Which leads us to...
+After that, any player will automatically have an `examplePlugin` method on its prototype!
+
+> **Note:** The only stipulation with the name of the plugin is that it cannot conflict with any existing plugin or player method.
+
+## Writing a Advanced Plugin
```
gkatsev: *an* advanced plugin.

docs/guides/plugins.md
```diff
@@ -1,58 +1,345 @@
...
-        examplePlugin: {
-          exampleOption: true
-        }
-      }
+const Plugin = videojs.getPlugin('plugin');
+
+class ExamplePlugin extends Plugin {
+
+  constructor(player, options) {
+    super(player);
```
gkatsev: options isn't passed here but it is in the other examples.

src/js/mixins/stateful.js
```diff
@@ -0,0 +1,92 @@
...
+ */
+const setState = function(stateUpdates) {
+
+  // Support providing the `stateUpdates` state as a function.
+  if (typeof stateUpdates === 'function') {
+    stateUpdates = stateUpdates();
+  }
+
+  if (!Obj.isPlain(stateUpdates)) {
+    log.warn('non-plain object passed to `setState`', stateUpdates);
```
gkatsev: if we're exiting this method early, should this be an error instead?

gkatsev: Also, what about just going with it regardless of what kind of object was passed in, assuming it isn't `null`. We could copy the properties from whatever was passed in.

src/js/mixins/stateful.js
```diff
@@ -0,0 +1,92 @@
...
+ * @param  {Object} [defaultState]
+ *         A default set of properties to populate the newly-stateful object's
+ *         `state` property.
+ *
+ * @return {Object}
+ *         Returns the `target`.
+ */
+function stateful(target, defaultState) {
+  target.state = Obj.assign({}, defaultState);
+  target.setState = Fn.bind(target, setState);
```
gkatsev: since `setState` already uses `this`, we probably could just do `target.setState = setState`.
But either way.

src/js/player.js
```diff
@@ -3341,6 +3341,40 @@ class Player extends Component {
...
+   * on this player. For that, [usingPlugin]{@link Player#usingPlugin}.
+   *
+   * @param  {string}  name
+   *         The name of a plugin.
+   *
+   * @return {boolean}
+   *         Whether or not this player has the requested plugin available.
+   */
+  hasPlugin(name) {
+    // While a no-op by default, this method is created in plugin.js to avoid
```
gkatsev: we can probably use the same trick that @BrandonOCasey has been using where we just have the doc comment but no actual function.
I wonder if we can specify partials for jsdoc.

src/js/plugin.js
```diff
@@ -0,0 +1,445 @@
...
+
+    // Make this object evented, but remove the added `trigger` method so we
+    // use the prototype version instead.
+    evented(this);
+    delete this.trigger;
+
+    stateful(this, this.constructor.defaultState);
+    markPluginAsActive(player, this.name);
+
+    // Bind all plugin prototype methods to this object.
```
gkatsev: why is this necessary?

src/js/plugin.js
```diff
@@ -0,0 +1,445 @@
...
+   * Gets a plugin by name if it exists.
+   *
+   * @param  {string} name
+   *         The name of a plugin.
+   *
+   * @return {Function|undefined}
+   *         The plugin (or `undefined`).
+   */
+  static getPlugin(name) {
+    return getPlugin(name);
```
gkatsev: the names being the same makes it a bit confusing. Took me a second to realize it wouldn't be an infinite loop.

gkatsev:
Would be so nice to have better managed plugins! Looking forward to pulling this in.

src/js/mixins/stateful.js
misteroneill: Sounds reasonable. I'll remove this block.

src/js/mixins/stateful.js
misteroneill: I think my reasoning was probably overly-paranoid defensiveness.

src/js/plugin.js
misteroneill: It's not. Overly-paranoid defensiveness again.

misteroneill:
I believe all feedback has been addressed.

docs/guides/plugins.md
```diff
@@ -1,58 +1,345 @@
...
-* [Grunt](http://gruntjs.com) for build management
-* [npm](https://www.npmjs.org) for dependency management
-* [QUnit](http://qunitjs.com) for testing
 
-## Step 1: Write Some Javascript
+- [Writing a Basic Plugin](#writing-a-basic-plugin)
+  - [Write a JavaScript Function](#write-a-javascript-function)
+  - [Register a Basic Plugin](#register-a-basic-plugin)
+- [Writing a Advanced Plugin](#writing-a-advanced-plugin)
+  - [Write a JavaScript Class/Constructor](#write-a-javascript-classconstructor)
```
BrandonOCasey: If the new doc tooling was merged into master, it might be better to switch over to that style of TOC if we can.

src/js/component.js
```diff
@@ -566,17 +566,46 @@ class Component {
   /**
    * Add an `event listener` to this `Component`s element.
    *
+   * ```js
```
BrandonOCasey: examples were moved out of doc comments in master

src/js/component.js
```diff
@@ -638,8 +667,47 @@ class Component {
...
   /**
-   * Remove an event listener from this `Component`s element. If the second argument is
-   * exluded all listeners for the type passed in as the first argument will be removed.
+   * Remove an event listener from this `Component`s element.
+   * ```js
+   *   var player = videojs('some-player-id');
+   *   var Component = videojs.getComponent('Component');
+   *   var myComponent = new Component(player);
+   *   var myFunc = function() {
+   *     var myComponent = this;
```
BrandonOCasey: examples were moved out of doc comments in master

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
+/**
+ * @file mixins/evented.js
```
BrandonOCasey: this should have a module definition as well so that it will show up correctly on the website. `@module evented`

src/js/mixins/stateful.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * @file mixins/stateful.js
```
BrandonOCasey: need `@module stateful` here

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+  /**
+   * Fire an event on this evented object, causing its listeners to be called.
+   *
+   * @param  {string|Object} event
+   *         An event type or an object with a type property.
+   *
+   * @param  {Object} [hash]
+   *         An additional object to pass along to listeners.
+   *
+   * @return {Object}
```
BrandonOCasey: This function does not seem to return anything

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+   *         If the first argument was a string or array, this may be the
+   *         listener function. Otherwise, this is a string or array of event
+   *         type(s).
+   *
+   * @param  {Function} [listener]
+   *         If the first argument was another evented object, this will be
+   *         the listener function; otherwise, _all_ listeners bound to the
+   *         event type(s) will be removed.
+   *
+   * @return {Object}
```
BrandonOCasey: This function does not seem to return anything.

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
+   * @param  {string|Array|Function} typeOrListener
+   *         If the first argument was a string or array, this should be the
+   *         listener function. Otherwise, this is a string or array of event
+   *         type(s).
+   *
+   * @param  {Function} [listener]
+   *         If the first argument was another evented object, this will be
+   *         the listener function.
+   *
+   * @return {Object}
```
BrandonOCasey: no return in the function

src/js/mixins/evented.js
```diff
@@ -0,0 +1,385 @@
...
```
BrandonOCasey: no return in the function.

src/js/mixins/stateful.js
```diff
@@ -0,0 +1,92 @@
...
+import * as Obj from '../utils/obj';
+
+/**
+ * Set the state of an object by mutating its `state` object in place.
+ *
+ * @param  {Object|Function} stateUpdates
+ *         A new set of properties to shallow-merge into the plugin state. Can
+ *         be a plain object or a function returning a plain object.
+ *
+ * @return {Object}
```
BrandonOCasey: should be `{Object|undefined}`

src/js/plugin.js
```diff
@@ -0,0 +1,445 @@
...
+    this[name] = () => instance;
+
+    return instance;
+  };
+};
+
+class Plugin {
+
+  /**
+   * Plugin constructor.
```
BrandonOCasey: I think these jsdocs need to be moved around a bit, see the jsdocs [from another class](https://github.com/videojs/video.js/blob/master/src/js/component.js#L23)

src/js/plugin.js
BrandonOCasey: yeah that works

src/js/plugin.js
BrandonOCasey: If we want to remain loose should we even bother having this function?

src/js/plugin.js
```diff
@@ -0,0 +1,445 @@
...
+ *         A factory function for the plugin sub-class.
+ */
+const createPluginFactory = (name, PluginSubClass) => {
+
+  // Add a `name` property to the plugin prototype so that each plugin can
+  // refer to itself by name.
+  PluginSubClass.prototype.name = name;
+
+  return function(...args) {
+    const instance = new PluginSubClass(...[this, ...args]);
```
BrandonOCasey: do we want to create a new plugin each time `player.pluginName()` is called? Don't we want to return the existing instance if one already exists that has not been disposed?

misteroneill: Note on line 153 that the plugin factory replaces itself with a function that simply returns the existing instance. If that instance is then disposed, it goes back to the factory.

src/js/component.js
misteroneill: Yeah, I've rolled back the changes to this file. Looks like it was a rebase mixup.

src/js/component.js
misteroneill: Yeah, I've rolled back the changes to this file. Looks like it was a rebase mixup.

theartoftraining99:
Virtualisation

PR #3694, opened 2016-10-19
Update the jsdoc comments to modern syntax - Part 1

BrandonOCasey:
## Description
- Add `@return` and description where missing from functions
- Add `@param` and description where missing
- Remove un-necessary doc comments such as `@class` and `@method`
- Update examples to match current code standards
- Update descriptions to make more sense across the board
- Keep all doc comments in a standard format
- Verify that documentation looked good after running jsdoc
- document `@events` and `@fires`
- Depends upon #3721 
## Requirements Checklist

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -23,19 +23,15 @@ class BigPlayButton extends Button {
   }
 
   /**
-   * This gets called when a `BigPlayButton` gets:
-   * - Clicked (via the `click` event, listening starts in the constructor)
-   * - Tapped (via the `tap` event, listening starts in the constructor)
-   * - Gains focus (via the `focus` event). Causes `ClickableComponent`
-   *   to listen for the `keydown` event. If the enter/space key gets
-   *   pressed before focus gets lost (via the `blur` event) this function gets called.
+   * This gets called when a `BigPlayButton` "clicked". See {@link ClickableComponent}
+   * for more detailed information on what a click can be.
    *
    * @param {EventTarget~Event} event
    *        The `keydown`, `tap`, or `click` event that caused this function to be
    *        called.
    *
-   * @listens {tap}
-   * @listens {click}
+   * @listens tap
+   * @listens click
    */
   handleClick(event) {
     this.player_.play();
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -5,37 +5,49 @@ import Button from './button.js';
 import Component from './component.js';
 
 /**
- * Initial play button. Shows before the video has played. The hiding of the
- * big play button is done via CSS and player states.
+ * The initial play button that shows before the video has played. The hiding of the
+ * `BigPlayButton` get done via CSS and `Player` states.
  *
- * @param {Object} player  Main Player
- * @param {Object=} options Object of option names and values
  * @extends Button
- * @class BigPlayButton
  */
 class BigPlayButton extends Button {
 
   /**
-   * Allow sub components to stack CSS class names
+   * Builds the default DOM `className`.
    *
-   * @return {String} The constructed class name
-   * @method buildCSSClass
+   * @return {string}
+   *         The DOM `className` for this object. Always returns 'vjs-big-play-button'.
    */
   buildCSSClass() {
     return 'vjs-big-play-button';
   }
 
   /**
-   * Handles click for play
+   * This gets called when a `BigPlayButton` gets:
+   * - Clicked (via the `click` event, listening starts in the constructor)
+   * - Tapped (via the `tap` event, listening starts in the constructor)
+   * - Gains focus (via the `focus` event). Causes `ClickableComponent`
+   *   to listen for the `keydown` event. If the enter/space key gets
+   *   pressed before focus gets lost (via the `blur` event) this function gets called.
    *
-   * @method handleClick
+   * @param {EventTarget~Event} event
+   *        The `keydown`, `tap`, or `click` event that caused this function to be
+   *        called.
+   *
+   * @listens {tap}
+   * @listens {click}
    */
-  handleClick() {
+  handleClick(event) {
     this.player_.play();
   }
-
 }
 
+/**
+ * The text that should display over the `BigPlayButton`s controls. Added to for localization.
+ *
+ * @type {string}
+ * @private
+ */
 BigPlayButton.prototype.controlText_ = 'Play Video';
 
 Component.registerComponent('BigPlayButton', BigPlayButton);
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -120,7 +120,7 @@ class Button extends ClickableComponent {
    * @param {EventTarget~Event} event
    *        The event that caused this function to get called.
    *
-   * @listens {keydown}
+   * @listens keydown
    */
   handleKeyPress(event) {
 
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -7,23 +7,26 @@ import log from './utils/log.js';
 import assign from 'object.assign';
 
 /**
- * Base class for all buttons
+ * Base class for all buttons.
  *
- * @param {Object} player  Main Player
- * @param {Object=} options Object of option names and values
  * @extends ClickableComponent
- * @class Button
  */
 class Button extends ClickableComponent {
 
   /**
-   * Create the component's DOM element
+   * Create the `Button`s DOM element.
    *
-   * @param {String=} type Element's node type. e.g. 'div'
-   * @param {Object=} props An object of properties that should be set on the element
-   * @param {Object=} attributes An object of attributes that should be set on the element
+   * @param {string} [tag=button]
+   *        Element's node type. e.g. 'button'
+   *
+   * @param {Object} [props={}]
+   *        An object of properties that should be set on the element.
+   *
+   * @param {Object} [attributes={}]
+   *        An object of attributes that should be set on the element.
+   *
    * @return {Element}
-   * @method createEl
+   *         The element that gets created.
    */
   createEl(tag = 'button', props = {}, attributes = {}) {
     props = assign({
@@ -62,13 +65,20 @@ class Button extends ClickableComponent {
   }
 
   /**
-   * Adds a child component inside this button
+   * Add a child `Component` inside of this `Button`.
    *
-   * @param {String|Component} child The class name or instance of a child to add
-   * @param {Object=} options Options, including options to be passed to children of the child.
-   * @return {Component} The child component (created by this process if a string was used)
-   * @deprecated
-   * @method addChild
+   * @param {string|Component} child
+   *        The name or instance of a child to add.
+   *
+   * @param {Object} [options={}]
+   *        The key/value store of options that will get passed to children of
+   *        the child.
+   *
+   * @return {Component}
+   *         The `Component` that gets added as a child. When using a string the
+   *         `Component` will get created by this process.
+   *
+   * @deprecated since version 5
    */
   addChild(child, options = {}) {
     const className = this.constructor.name;
@@ -80,10 +90,11 @@ class Button extends ClickableComponent {
   }
 
   /**
-   * Enable the button element
+   * Enable the `Button` element so that it can be activated or clicked. Use this with
+   * {@link Button#disable}.
    *
    * @return {Component}
-   * @method enable
+   *         Returns itself; method is chainable.
    */
   enable() {
     super.enable();
@@ -91,10 +102,11 @@ class Button extends ClickableComponent {
   }
 
   /**
-   * Disable the button element
+   * Enable the `Button` element so that it cannot be activated or clicked. Use this with
+   * {@link Button#enable}.
    *
    * @return {Component}
-   * @method disable
+   *         Returns itself; method is chainable.
    */
   disable() {
     super.disable();
@@ -102,9 +114,13 @@ class Button extends ClickableComponent {
   }
 
   /**
-   * Handle KeyPress (document level) - Extend with specific functionality for button
+   * This gets called when a `Button` has focus and `keydown` is triggered via a key
+   * press.
    *
-   * @method handleKeyPress
+   * @param {EventTarget~Event} event
+   *        The event that caused this function to get called.
+   *
+   * @listens {keydown}
    */
   handleKeyPress(event) {
 
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -10,15 +10,22 @@ import document from 'global/document';
 import assign from 'object.assign';
 
 /**
- * Clickable Component which is clickable or keyboard actionable, but is not a native HTML button
+ * Clickable Component which is clickable or keyboard actionable,
+ * but is not a native HTML button.
  *
- * @param {Object} player  Main Player
- * @param {Object=} options Object of option names and values
  * @extends Component
- * @class ClickableComponent
  */
 class ClickableComponent extends Component {
 
+  /**
+   * Creates an instance of this class.
+   *
+   * @param  {Player} player
+   *         The `Player` that this class should be attached to.
+   *
+   * @param  {Object} [options]
+   *         The key/value store of player options.
+   */
   constructor(player, options) {
     super(player, options);
 
@@ -28,13 +35,19 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Create the component's DOM element
+   * Create the `Component`s DOM element.
    *
-   * @param {String=} type Element's node type. e.g. 'div'
-   * @param {Object=} props An object of properties that should be set on the element
-   * @param {Object=} attributes An object of attributes that should be set on the element
+   * @param {string} [tag=div]
+   *        The element's node type.
+   *
+   * @param {Object} [props={}]
+   *        An object of properties that should be set on the element.
+   *
+   * @param {Object} [attributes={}]
+   *        An object of attributes that should be set on the element.
+   *
    * @return {Element}
-   * @method createEl
+   *         The element that gets created.
    */
   createEl(tag = 'div', props = {}, attributes = {}) {
     props = assign({
@@ -64,11 +77,13 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * create control text
+   * Create a control text element on this `Component`
    *
-   * @param {Element} el Parent element for the control text
+   * @param {Element} [el]
+   *        Parent element for the control text.
+   *
    * @return {Element}
-   * @method controlText
+   *         The control text element that gets created.
    */
   createControlTextEl(el) {
     this.controlTextEl_ = Dom.createEl('span', {
@@ -85,12 +100,17 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Controls text - both request and localize
+   * Get or set the localize text to use for the controls on the `Component`.
    *
-   * @param {String}  text Text for element
-   * @param {Element=} el Element to set the title on
-   * @return {String}
-   * @method controlText
+   * @param {string} [text]
+   *        Control text for element.
+   *
+   * @param {Element} [el=this.el()]
+   *        Element to set the title on.
+   *
+   * @return {string|ClickableComponent}
+   *         - The control text when getting
+   *         - Returns itself when setting; method can be chained.
    */
   controlText(text, el = this.el()) {
     if (!text) {
@@ -107,20 +127,20 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Allows sub components to stack CSS class names
+   * Builds the default DOM `className`.
    *
-   * @return {String}
-   * @method buildCSSClass
+   * @return {string}
+   *         The DOM `className` for this object.
    */
   buildCSSClass() {
     return `vjs-control vjs-button ${super.buildCSSClass()}`;
   }
 
   /**
-   * Enable the component element
+   * Enable this `Component`s element.
    *
-   * @return {Component}
-   * @method enable
+   * @return {ClickableComponent}
+   *         Returns itself; method can be chained.
    */
   enable() {
     this.removeClass('vjs-disabled');
@@ -136,10 +156,10 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Disable the component element
+   * Disable this `Component`s element.
    *
-   * @return {Component}
-   * @method disable
+   * @return {ClickableComponent}
+   *         Returns itself; method can be chained.
    */
   disable() {
     this.addClass('vjs-disabled');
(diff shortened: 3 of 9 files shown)


src/js/clickable-component.js
```diff
@@ -110,8 +127,8 @@ class ClickableComponent extends Component {
   /**
    * Allows sub components to stack CSS class names
    *
-   * @return {String}
-   * @method buildCSSClass
+   * @return {string}
+   *         The current css class for this ClickableComponent
```
misteroneill: This doesn't feel particularly accurate. It returns the default `className` for the component's element - not the current `className`. I think we should avoid using the term "css class" because it's got nothing to do with CSS.

src/js/clickable-component.js
```diff
@@ -120,10 +137,14 @@ class ClickableComponent extends Component {
...
    *
-   * @param {String|Component} child The class name or instance of a child to add
-   * @param {Object=} options Options, including options to be passed to children of the child.
-   * @return {Component} The child component (created by this process if a string was used)
-   * @method addChild
+   * @param {String|Component} child
+   *        The class name or instance of a child to add
+   *
+   * @param {Object} [options={}]
+   *        including options to be passed to children of the child.
```
misteroneill: Looks like we lost "Options, " here.

src/js/clickable-component.js
```diff
@@ -196,9 +228,11 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Handle Blur - Remove keyboard triggers
+   * Called when a ClickableComponent component loses focus. This function turns of
+   * the watching for `keydown` events (so they will no longer call `this.handleKeyPress`
```
misteroneill: Looks like "of" should be "off" but I feel like the terminology here should be "unbinds the listener for keydown events". Also missing a parentheses at the end.

src/js/close-button.js
```diff
@@ -1,24 +1,45 @@
...
- * @extends Button
- * @class CloseButton
- */
 class CloseButton extends Button {
 
+  /**
+   * The `CloseButton` component is a button which fires a "close" event
+   * when it is activated.
+   *
+   * @param {Object} player
```
misteroneill: I think these should say `@param {Player} player` if possible.

src/js/close-button.js
```diff
@@ -1,24 +1,45 @@
...
+   *
+   * @extends Button
+   */
   constructor(player, options) {
     super(player, options);
     this.controlText(options && options.controlText || this.localize('Close'));
   }
 
+  /**
+   * Returns the css attachments for this element
```
misteroneill: What are css attachments?

src/js/close-button.js
```diff
@@ -1,24 +1,45 @@
...
+   * @return {string}
+   *         The current css class for thes CloseButton
+   */
   buildCSSClass() {
     return `vjs-close-button ${super.buildCSSClass()}`;
   }
 
+  /**
+   * Triggers a close event when we are clicked,
+   * or focused and enter/space is pressed down
```
misteroneill: I think we shouldn't use personal language here (i.e. "we are" -> "the component is"). Also, comma is unnecessary here.

src/js/component.js
```diff
@@ -630,11 +708,17 @@ class Component {
...
+   *
+   * @param {Function|string} second
+   *        The event handler or event type
+   *
+   * @param {Function} third
+   *        The event handler
+   *
    * @return {Component}
-   * @method on
+   *         a reference to the component this function was called with
```
misteroneill: I think we should either come up with a convention for "Returns itself." because the same thing is phrased different ways in different files - or leave these out and allow the type (e.g. `{Component}`) imply it.

src/js/component.js
```diff
@@ -1135,17 +1330,19 @@ class Component {
...
   currentWidth() {
     return this.currentDimension('width');
   }
 
   /**
    * Get height of computed style
+   *
    * @return {Integer}
-   * @method currentHeight
+   *         the current heigth
```
misteroneill: Spelling

src/js/component.js
```diff
@@ -1293,15 +1494,20 @@ class Component {
...
+   *
+   * @return {number}
+   *         Returns the timeout ID
    */
   setTimeout(fn, timeout) {
     fn = Fn.bind(this, fn);
 
-    // window.setTimeout would be preferable here, but due to some bizarre issue with Sinon and/or Phantomjs, we can't.
+    // window.setTimeout would be preferable here,
+    // but due to some bizarre issue with Sinon and/or Phantomjs, we can't.
```
misteroneill: I don't think this comment is relevant anymore.

src/js/component.js
```diff
@@ -1397,16 +1616,18 @@ class Component {
...
-   * @static
-   * @method getComponent
+   *         the component that was registered under name
    */
   static getComponent(name) {
     if (Component.components_ && Component.components_[name]) {
       return Component.components_[name];
     }
 
+    // TODO: deprecated?
```
misteroneill: What's the question here?

src/js/error-display.js
```diff
@@ -49,6 +49,11 @@ class ErrorDisplay extends ModalDialog {
   }
 }
 
+/**
+ * Available options for an error display
```
misteroneill: I think this should be "Default options" instead of "Available options".

src/js/event-target.js
```diff
@@ -1,12 +1,50 @@
 /**
- * @file event-target.js
+ * @file src/js/event-target.js
  */
 import * as Events from './utils/events.js';
 
+/**
+ * An EventTarget class that seeks to emulate much of the builtin
+ * HTML EventTarget functionalify as well as add all kinds of ease
```
misteroneill: Should be "functionalify" and I think the sentence can end there. The rest is kind of vague.

src/js/event-target.js
```diff
@@ -1,12 +1,50 @@
...
+ * A list of events that the EventTarget is allowed to use
+ *
+ * @property EventTarget.prototype.allowedEvents_
+ * @private
+ */
 EventTarget.prototype.allowedEvents_ = {};
 
+/**
+ * This function is a cross browser wrapper around addEventListener
+ * with a much nicer syntax
```
misteroneill: I think it'd be better to describe what this does rather than its implementation, the same way the `off` method does.

src/js/clickable-component.js
```diff
@@ -33,11 +37,17 @@ class ClickableComponent extends Component {
...
+   *
+   * @param {Object} [props={}]
+   *        An object of properties that should be set on the element
+   *
+   * @param {Object} [attributes={}]
+   *        An object of attributes that should be set on the element
+   *
    * @return {Element}
-   * @method createEl
+   *         the element that was created
```
BrandonOCasey: it may be implied but I don't think its a bad thing to document

src/js/component.js
BrandonOCasey: well sometime `@return {Component}` will return a component that does not refer to the one that the function was called on. Although I do agree we need some sort of convention for it.

src/js/component.js
BrandonOCasey: do we want to warn in the log that this is deprecated?

src/js/component.js
BrandonOCasey: maybe it should just be "a reference to `this` object"?

src/js/close-button.js
```diff
@@ -6,20 +9,55 @@ import Component from './component';
...
+   * @param {Object} [options]
+   *        Object of option names and values
+   */
   constructor(player, options) {
     super(player, options);
     this.controlText(options && options.controlText || this.localize('Close'));
   }
 
+  /**
+   * Constructs a string to be used for the component elements
```
gkatsev: Does jsdoc not provide any functionality for inheritance? So we would only need to document these in Component and then everything would be linked?

src/js/component.js
```diff
@@ -770,17 +918,38 @@ class Component {
...
    * ```
    *
-   * @param  {Event|Object|String} event  A string (the type) or an event object with a type attribute
-   * @param  {Object} [hash] data hash to pass along with the event
-   * @return {Component}       self
-   * @method trigger
+   * @param {Event|Object|string} event
+   *        A string (the type) or an event object with a type attribute
+   *
+   * @param {Object} [hash]
```
gkatsev: is there an "optional" indicator?

src/js/component.js
```diff
@@ -1093,9 +1297,12 @@ class Component {
...
    * Get width or height of computed style
-   * @param  {String} widthOrHeight  'width' or 'height'
-   * @return {Number|Boolean} The bolean false if nothing was set
-   * @method currentDimension
+   *
+   * @param {string} widthOrHeight
+   *        'width' or 'height'
+   *
+   * @return {number|boolean}
+   8         the dimention or false if nothing was set
```
gkatsev: 8 instead of *.
dimension mispelled.

This doesn't return a boolean. It should default to `0`, I believe.

src/js/component.js
gkatsev: This is probably something that should just be disallowed/removed in 6.0

src/js/error-display.js
```diff
@@ -27,11 +30,10 @@ class ErrorDisplay extends ModalDialog {
   /**
    * Include the old class for backward-compatibility.
    *
-   * This can be removed in 6.0.
```
gkatsev: I wonder what this refers to specifically.

src/js/component.js
BrandonOCasey: ok I will add an @deprication tag to the function here

src/js/close-button.js
BrandonOCasey: anything that you override needs to be documented, we get anything that is not overridden, in the docs for the object. There are ways to say that this method should use the parents doc comments, but doing so causes jsdoc to link to the parent function rather than the current files function (so it seems like the current file does not implement it)

src/js/component.js
BrandonOCasey: brackets around the parameter indicate optional

BrandonOCasey:
@gkatsev @misteroneill ready for re-review

misteroneill:
As a general comment, I think it'd be good to have some kind of conventions regarding grammar. My _personal_ feeling is that things should generally follow English grammar. So, something like this:

```
* @return {Component}
*         Returns itself; method can be chained.
```

But that's me. If more people disagree, that's fine. Thoughts?

BrandonOCasey:
sounds good to me

src/js/big-play-button.js
```diff
@@ -5,41 +5,50 @@ import Button from './button.js';
...
   }
 
   /**
-   * Allow sub components to stack CSS class names
+   * Get the default css class for the BigPlayButton
    *
-   * @return {String} The constructed class name
-   * @method buildCSSClass
+   * @return {string}
+   *          returns default class name ie. 'vjs-big-play-button'
```
misteroneill: Couple grammatical updates here: `Returns default class name, i.e., 'vjs-big-play-button'.`

src/js/big-play-button.js
```diff
@@ -5,41 +5,50 @@ import Button from './button.js';
...
+   * @param {Object} [options]
+   *        Object of option names and values
+   */
   constructor(player, options) {
     super(player, options);
   }
 
   /**
-   * Allow sub components to stack CSS class names
+   * Get the default css class for the BigPlayButton
```
misteroneill: If we're leaving "css" here it should be capitalized.

src/js/big-play-button.js
```diff
@@ -5,41 +5,50 @@ import Button from './button.js';
...
 
+  /**
+   * constructor for BigPlayButton
+   *
+   * @param {Player} player
+   *        the player to attach this button to
+   *
+   * @param {Object} [options]
+   *        Object of option names and values
+   */
```
misteroneill: Both "constructor" and "the" should be capitalized in this block.

src/js/button.js
```diff
@@ -9,25 +9,37 @@ import assign from 'object.assign';
...
  * @extends ClickableComponent
- * @class Button
  */
 class Button extends ClickableComponent {
 
+  /**
+   * Constructor for Button
+   *
+   * @param {Player} player
+   *        the player to attach this button to
```
misteroneill: Capitalization.

src/js/button.js
```diff
@@ -84,9 +102,11 @@ class Button extends ClickableComponent {
   }
 
   /**
-   * Handle KeyPress (document level) - Extend with specific functionality for button
+   * Function that is called for this class when a key is pressed,
+   * Extend with specific functionality for any button you want to support
    *
-   * @method handleKeyPress
+   * @param {Event} event
+   *        the button event that caused this function to be called
```
misteroneill: Capitalization.

src/js/clickable-component.js
```diff
@@ -10,15 +10,24 @@ import document from 'global/document';
...
- * @param {Object} player  Main Player
- * @param {Object=} options Object of option names and values
  * @extends Component
- * @class ClickableComponent
  */
 class ClickableComponent extends Component {
 
+  /**
+   * constructor for a ClickableComponent
+   * but is not a native HTML button.
```
misteroneill: These look like they should be on the same line. And "constructor" should be capitalized. And there should be a comma after "ClickableComponent".

src/js/clickable-component.js
misteroneill: Capitalization, though. 😄

src/js/clickable-component.js
```diff
@@ -67,9 +82,11 @@ class ClickableComponent extends Component {
...
    * create control text
    *
-   * @param {Element} el Parent element for the control text
+   * @param {Element} [el]
+   *        Parent element for the control text
+   *
    * @return {Element}
-   * @method controlText
+   *         the control text element
    */
```
misteroneill: This could probably stand to have some capitalization added to both the `@return` and the description above.

src/js/clickable-component.js
```diff
@@ -88,10 +105,15 @@ class ClickableComponent extends Component {
...
-   * @method controlText
+   * @param {string} text
+   *        Text for element
+   *
+   * @param {Element} [el=this.el()]
+   *        Element to set the title on
+   *
+   * @return {string|ClickableComponent}
+   *         - The control text when getting
+   *         - Returns itself when setting; method can be chained.
```
misteroneill: Does this get Markdown-ified?

src/js/clickable-component.js
```diff
@@ -108,10 +130,14 @@ class ClickableComponent extends Component {
...
+   * Constructs a string to be used for the component elements
+   * default className
+   *
+   * This allows sub-components to stack class names with the
+   * parent
    *
-   * @return {String}
-   * @method buildCSSClass
+   * @return {string}
+   *         the default className for a component
```
misteroneill: Capitalization and sentences should end in periods.

src/js/clickable-component.js
```diff
@@ -162,25 +193,27 @@ class ClickableComponent extends Component {
...
 
   /**
-   * Handle KeyPress (document level) - Trigger click when Space or Enter key is pressed
+   * Called when this ClickableComponent has focus and when a key is pressed down.
+   * by default it will call `this.handleClick` when the key is space or enter
    *
-   * @method handleKeyPress
+   * @param {Event} event
+   *        the event that caused this to be called, in most cases
+   *        this will be the `keydown` event
```
misteroneill: I think we should leave out everything after the comma. And capitalize.

src/js/component.js
```diff
@@ -17,31 +17,50 @@ import mergeOptions from './utils/merge-options.js';
...
- *     button.trigger('customevent');
+ *
+ * Adding a button to the player
+ * ``` js
+ *   // adding a button to the player
+ *   var player = videojs('some-video-id');
+ *   var button = player.addChild('button');
+ *
+ *   console.log(button.el());
+ *   // will have the same html result as the previous example
```
misteroneill: I'm wondering if we should really have code examples in API docs or leave that to the manual documentation/guides.

BrandonOCasey: A part of me feels like we need to have doc comments that link to guides, and the guides should be the ones with the examples. Maybe another PR to rip examples out into manual documentation at a later date?

misteroneill: Sounds good.

src/js/component.js
```diff
@@ -130,53 +167,62 @@ class Component {
...
   player() {
     return this.player_;
   }
 
   /**
-   * Deep merge of options objects
-   * Whenever a property is an object on both options objects
-   * the two properties will be merged using mergeOptions.
+   * Deep merge of options objects with new options.
+   * > Note: When a property of is an object on both `obj` and `options`.
```
gkatsev: "When a property of is an object" is weird.
Maybe something like:
When both `obj` and `options` contain a property whose value is an object, the two properties get merged using mergeOptions

src/js/component.js
```diff
@@ -606,35 +687,63 @@ class Component {
...
-   * @param  {String|Component} first   The event type or other component
-   * @param  {Function|String}      second  The event handler or event type
-   * @param  {Function}             third   The event handler
+   *
+   * The benefit of using this over the following:
+   * - `VjsEvents.on(otherElement, 'eventName', myFunc)`
+   * - `otherComponent.on('eventName', myFunc)`
+   * Is that the listeners will get cleaned up when either component gets disposed.
+   * It will also bind `myComponent` as the context of `myFunc`.
+   * > **NOTE**: If you remove the element from the DOM that has used `on` you need to
```
gkatsev: I think usage of `NOTE` is consistent.

src/js/component.js
```diff
@@ -837,213 +1035,258 @@ class Component {
...
+   *
+   * @see [Information on CSS Selectors](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Getting_Started/Selectors)
    */
   $$(selector, context) {
     return Dom.$$(selector, context || this.contentEl());
   }
 
   /**
-   * Check if a component's element has a CSS class name
+   * Check if a component's element has a CSS `className`.
```
gkatsev: I think here it should stay as "class name". We specifically mean a generic CSS class name and not the `className` property on an html element.

gkatsev: This applied to `addClass` and `removeClass`

gkatsev: And others below.

src/js/component.js
```diff
@@ -837,213 +1035,258 @@ class Component {
...
+   * Things to know:
+   * - If the width or height in an number this will return the number
+   *   postfixed with 'px'.
+   * - If the width/height is a percent this will return the percent postfixed with
+   *   '%'
+   * - Hidden elements have a width of 0 with `window.getComputedStyle`. This function
+   *   defaults to the `Component`s `style.width` and falls back to
+   *   `window.getComputedStyle`. See [this]{@link http://www.foliotek.com/devblog/getting-the-width-of-a-hidden-element-with-jquery-using-width/}
+   *   for more information
+   * - If you want the `window.getComputedStyle` use {@link Component#currentWidth}
```
gkatsev: Maybe "if you want the computed style of the component, use `Component#current<<>>` instead"

src/js/component.js
```diff
@@ -1122,9 +1376,23 @@ class Component {
...
+   * computed style. Uses `window.getComputedStyle`.
+   *
+   * @typedef {Object} Component~DimensionObject
+   *
+   * @property {number} width
+   *           The width of the `Component`s computed style.
+   *
+   * @property {number} height
+   *           The height of the `Component`s computed style.
+   */
```
gkatsev: why is this comment broken up into two?

gkatsev: Oh, it's defining what a "DimensionObject" is. Maybe add a new line here?

src/js/component.js
```diff
@@ -1291,19 +1574,41 @@ class Component {
   }
 
   /**
-   * Creates timeout and sets up disposal automatically.
+   * A callback that has no parameters and is bound into `Component`s context.
+   *
+   * @callback Component~GenericCallback
+   * @this Component
+   */
```
gkatsev: Add a new line here?

src/js/component.js
```diff
@@ -1291,19 +1574,41 @@ class Component {
...
+  /**
+   * Creates a function that runs after an `x` millisecond timeout. This function is a
+   * wrapper around `window.setTimeout`. There are a few reasons to use this one
+   * instead though:
+   * 1. It gets cleared via  {@link Component#clearTimeout} when
+   *    {@link Component#dispose} gets called.
+   * 2. The function callback will gets turned into a {@link Component~GenericCallback}
+   *
+   * > Note: You can use `window.clearTimeout` on the id returned by this function. This
+   *         will cause its dispose listener will not get cleaned up! Please use
```
gkatsev: "This will cause its dispose listener *not* to get cleaned up"

src/js/component.js
```diff
@@ -1378,12 +1711,16 @@ class Component {
   }
 
   /**
-   * Registers a component
+   * Register a `Component` with `videojs` given the name and the component.
```
gkatsev: Should we add a note that Techs shouldn't be registered as Components?

src/js/event-target.js
```diff
@@ -1,12 +1,69 @@
...
+ * @class EventTarget
+ */
 const EventTarget = function() {};
 
+/**
+ * A Custom DOM event.
+ *
+ * @typedef {Object} EventTarget~Event
+ * @see [Properties]{@link https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent}
+ */
```
gkatsev: new line, please

src/js/event-target.js
```diff
@@ -1,12 +1,69 @@
...
+ *
+ * @callback EventTarget~EventListener
+ * @this {EventTarget}
+ *
+ * @param {EventTarget~Event} event
+ *        the event that triggered this function
+ *
+ * @param {Object} [hash]
+ *        hash of data sent during the event
+ */
```
gkatsev: new line, please

src/js/event-target.js
```diff
@@ -1,12 +1,69 @@
...
+/**
+ * An object containing event names as keys and true/false values as values. If an
+ * event name is set to a true value here {@link EventTarget#trigger} will have extra.
```
gkatsev: This sentence is incomplete. What will `trigger` have extra of?

src/js/event-target.js
```diff
@@ -35,6 +181,37 @@ EventTarget.prototype.one = function(type, fn) {
...
 };
 
+/**
+ * This function causes an event to happen. This will then cause any `event listeners`
+ * that are waiting for that event, to get called. If there are no `event listeners`
+ * for an event then nothing will happen.
+ *
+ * If the name of the Event that is being triggered is in `EventTarget.allowedEvents_`
+ * then trigger will also call the `on` + `uppercaseEventName` function. Example:
+ * the 'click' is in in `EventTarget.allowedEvents_` trigger will attempt to call
```
gkatsev: > `click` is in `allowedEvents_`, so, trigger will attempt to call `onclick` if it exists.

src/js/component.js
BrandonOCasey: Another benefit of having the `jsdoc` comments not contain examples. Is that we can lint markdown examples with eslint. Which cannot be done with jsdoc markdown examples.

PR #3703, opened 2016-10-23
Manual Documentation Improvements

misteroneill:
The first step in sweeping the `docs/` folder for completeness and accuracy.

This PR focuses on the entry-level guides and the README. It also features a few removals that have come from various conversations. And it removes the API guide because it was duplicating the setup guide.

## Requirements Checklist


README.md
```diff
@@ -1,65 +1,96 @@
...
-  <source src="really-cool-video.mp4" type="video/mp4">
-  <source src="really-cool-video.webm" type="video/webm">
+<video
+    id="my-player"
+    class="video-js"
+    controls
+    preload="auto"
+    poster="//vjs.zencdn.net/v/oceans.png"
+    data-setup='{}'>
+  <source src="//vjs.zencdn.net/v/oceans.mp4" type="video/mp4"></source>
```
gkatsev: I'd be inclined to leave the old fake urls.

README.md
```diff
@@ -1,65 +1,96 @@
...
-Video.js is a free and open source library, and we appreciate any help you're willing to give. Check out the [contributing guide](/CONTRIBUTING.md).
 
-_Video.js uses [BrowserStack](https://browserstack.com) for compatibility testing_
-## Building your own Video.js from source
-To build your own custom version read the section on [contributing code](/CONTRIBUTING.md#contributing-code) and ["Building your own copy"](/CONTRIBUTING.md#building-your-own-copy-of-videojs) in the contributing guide.
+Video.js is a free and open source library, and we appreciate any help you're willing to give - whether it's fixing bugs, improving documentation, or suggesting new features. Check out the [contributing guide][contributing] for more!
+
+_Video.js uses [BrowserStack][browserstack] for compatibility testing._
+
 ## License
```
gkatsev: Should we link the heading too? Makes it much more visible as well.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
 ```
 
-RIGHT
+> **Note:** The example with `controls="false"` can be a point of confusion for new developers - it will actually turn controls _on_!
+
+These are correct:
+
 ```html
 <video controls ...>
+<video ...>
```
gkatsev: should we also show `<video loop="loop">`? As that's a common thing to do as well.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
 ```
 
-> The biggest issue people run into is trying to set these values to false using false as the value (e.g. controls="false") which actually does the opposite and sets the value to true because the attribute is still included. If you need the attribute to include an equals sign for XHTML validation, you can set the attribute's value to the same as its name (e.g. controls="controls").
+### Precedence
```
gkatsev: While there sort of is a precedence here, I'm not sure we actually officially support it.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
-If autoplay is true, the video will start playing as soon as page is loaded (without any interaction from the user).
-NOT SUPPORTED BY APPLE iOS DEVICES. Apple blocks the autoplay functionality in an effort to protect its customers from unwillingly using a lot of their (often expensive) monthly data plans. A user touch/click is required to start the video in this case.
-```html
-<video autoplay ...>
-or
-{ "autoplay": true }
-```
+## Options Reference
 
+### Standard `<video>` Element Options
```
gkatsev: Can we link to MDN or something here as well, for more info?

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
 
-'none': Don't preload any of the video data. This will wait until the user clicks play to begin downloading.
+If `true`/present as an attribute, begins playback automatically.
 
-```html
-<video preload ...>
-or
-{ "preload": "auto" }
-```
+> **Note:** As of iOS 10, Apple offers `autoplay` support in Safari. However, in order for it to work, the `<video>` element must have `muted` and `playsinline` attributes!
```
gkatsev: If it doesn't have audio tracks it also works.
We should also double-check whether `playsinline` is required.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
+
+#### `children`
+
+> Type: `Array`
+
+#### `fluid`
+
+> Type: `Boolean`
+
+When `true`, the Video.js player will have a fluid size. In other words, it will scale to fit its container.
```
gkatsev: Can we add something like:
> If a class of `vjs-fluid` is added to the player before initialization, this option is automatically set.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
+
+#### `inactivityTimeout`
```
gkatsev: 0 means that the user will never become inactive.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
-The height attribute sets the display height of the video.
-```html
-<video height="480" ...>
-or
-{ "height": 480 }
+#### `sourceOrder`
+
+> Type: `Boolean`
+
+Tells Video.js to prefer the order of [`sources`](#sources) over [`techOrder`](#techOrder) in selecting a source and playback tech.
```
gkatsev: Should we add an example?

docs/guides/setup.md
```diff
@@ -1,124 +1,87 @@
...
 ```js
-videojs(document.getElementsByClassName('awesome_video_class')[0], {}, function() {
-  // You can grab an element by class if you'd like, just make sure
-  // if it's an array that you pick one (here we chose the first).
-});
+videojs(document.querySelector('.video-js'));
 ```
 
-\* If you have trouble playing back content you know is in the [correct format](http://blog.zencoder.com/2013/09/13/what-formats-do-i-need-for-html5-video/), your HTTP server might not be delivering the content with the correct [MIME type](http://en.wikipedia.org/wiki/Internet_media_type#Type_video). Please double check your content's headers before opening an [issue](/CONTRIBUTING.md).
+Regardless, in either case, the `data-setup` attribute is still supported for providing [options][options] to Video.js. Additionally, the `videojs` function can take options as its second argument. [Learn more about Video.js options][options].
```
gkatsev: I'm not sure this is strictly true.

docs/guides/setup.md
```diff
@@ -1,124 +1,87 @@
...
-```html
-<link href="//vjs.zencdn.net/5.4.6/video-js.min.css" rel="stylesheet">
-<script src="//vjs.zencdn.net/5.4.6/video.min.js"></script>
-```
+## Installing Video.js
 
-Alternatively you can always [go here](http://videojs.com/getting-started/) to get the latest URL for videojs CDN.
+Video.js is officially available via CDN, npm, and Bower.
 
-We include a stripped down Google Analytics pixel that tracks a random percentage (currently 1%) of players loaded from the CDN. This allows us to see (roughly) what browsers are in use in the wild, along with other useful metrics such as OS and device. If you'd like to disable analytics, you can simply include the following global **before** including Video.js:
```
gkatsev: this section should be kept.

docs/guides/setup.md
```diff
@@ -1,124 +1,87 @@
...
 
-### Bower
-```
-$ bower install --save video.js
-```
+> Note: Video.js works with `<video>` _and_ `<audio>` elements, but for simplicity we'll refer only to `<video>` elements going forward!
 
+The core strength of Video.js is that it decorates a [standard `<video>` element][w3c-video] and emulates its associated [events and APIs][w3c-media-events], while providing a customizable DOM-based UI.
 
-### Self Hosted. ###
```
gkatsev: We should keep this section as well.

docs/index.md
```diff
@@ -1,37 +1,69 @@
-[Video.js homepage](http://videojs.com)
+# [Video.js][vjs] Documentation
 
-<h1>Video.js Documentation</h1>
+There are two categories of docs: [Guides][guides] and [API docs][api].
```
gkatsev: what about having `API Docs` link to the second below instead?

docs/guides/setup.md
gkatsev: There's a "build-your own" section in the CONTRIBUTING.md, we could potentially just link to that.

docs/guides/options.md
```diff
@@ -1,138 +1,352 @@
...
+
+Video.js indicates that the user is interacting with the player by way of the `"vjs-user-active"` and `"vjs-user-inactive"` classes and the `"useractive"` event.
+
+The `inactivityTimeout` determines how many milliseconds of inactivity is required before declaring the user inactive.
+
+#### `language`
+
+> Type: `String`
+
+A [two-letter code][lang-codes] matching one of the available languages in the player. This sets the initial language for a player, but it can always be changed.
```
mister-ben: Not necessarily two-letter. And while it's true the player's language can be changed later on, that can lead to confusion as localisable text does not get updated.

docs/guides/options.md
mister-ben: +1 This can be confusing

README.md
misteroneill: Why is that? I feel like this gives people something they could copy/paste and see it working.

docs/guides/options.md
misteroneill: This seems relevant, though. The sort of thing I'd want to know without having to experiment for myself. I feel like it's the implicit reality.

docs/guides/options.md
misteroneill: Good call. Rather than outlining how it works here, I'm going to link to the WebKit blog post.

docs/guides/options.md
misteroneill: Good point 👍

docs/guides/setup.md
misteroneill: Ah, yes, I thought the GA thing was in the "Getting Started" document, but it seems not.

docs/guides/setup.md
misteroneill: Another thing I thought was covered by "Getting Started"... will bring it back with accurate information!

docs/guides/setup.md
misteroneill: Actually, I think we should move this to the "Getting Started" guide. It's silly to have two documents covering the same thing.

docs/index.md
misteroneill: I think that's fine for now, but I wonder if we need the API guide at all.

docs/guides/options.md
gkatsev: I'm not comfortable with documenting a behavior we haven't been testing with, whether it's there implicitly or not.

misteroneill: Fair enough.

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
 
-```html
-<video controls autoplay preload="auto" ...>
-```
 
-Alternatively, you can use the data-setup attribute to provide options in the [JSON](http://json.org/example.html) format. This is also how you would set options that aren't standard to the video tag.
+- [Standard `<video>` Element Options](#standard-video-element-options)
+  - [`autoplay`](#autoplay)
+  - [`controls`](#controls)
+  - [`height`](#height)
```
BrandonOCasey: Looking at the spec [here](https://www.w3.org/TR/html5/embedded-content-0.html#htmlmediaelement) we may be missing (or maybe we have not implemented them):
* `playbackRate` 
* `defaultPlaybackRate`
* `volume`
* `defaultMuted`

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+A URL to an image that displays before the video begins playing. This is often a frame of the video or a custom title screen. As soon as the user hits "play" the image will go away.
+
+### `preload`
+
+> Type: `String`
+
+Suggests to the browser whether or not the video data should begin downloading as soon as the `<video>` element is loaded. Supported values are:
+
+- `'auto'`
```
BrandonOCasey: perhaps these should be sub sections `####`?

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+  - [`aspectRatio`](#aspectratio)
+  - [`children`](#children)
+  - [`fluid`](#fluid)
+  - [`inactivityTimeout`](#inactivitytimeout)
+  - [`language`](#language)
+  - [`languages`](#languages)
+  - [`notSupportedMessage`](#notsupportedmessage)
+  - [`plugins`](#plugins)
+  - [`sourceOrder`](#sourceorder)
+  - [`sources`](#sources)
```
BrandonOCasey: Do we want to specify anything from the [default options](https://github.com/videojs/video.js/blob/master/src/js/player.js#L2752)? (Seems like some of these may not even exist anymore...)

README.md
```diff
@@ -1,65 +1,96 @@
...
-Simply add these includes to your document's
-`<head>`:
+## Quick Start
+
+Thanks to the awesome folks over at [Fastly][fastly], there's a free, CDN hosted version of Video.js that anyone can use. Add these tags to your document's `<head>`:
 
 ```html
-<link href="//vjs.zencdn.net/5.8/video-js.min.css" rel="stylesheet">
-<script src="//vjs.zencdn.net/5.8/video.min.js"></script>
+<link href="//vjs.zencdn.net/5.11/video-js.min.css" rel="stylesheet">
```
BrandonOCasey: is there any way that we can prevent the docs from needing to be updated every time we update the video.js version?

README.md
```diff
@@ -1,65 +1,96 @@
...
 
+[![Build Status][travis-icon]][travis-link]
+[![Coverage Status][coveralls-icon]][coveralls-link]
+[![Slack Status][slack-icon]][slack-link]
 
-[![NPM](https://nodei.co/npm/video.js.png?downloads=true&downloadRank=true)](https://nodei.co/npm/video.js/)
+[![NPM][npm-icon]][npm-link]
 
-> Video.js is a web video player built from the ground up for an HTML5 world. It supports HTML5 and Flash video, as well as YouTube and Vimeo (through [plugins](https://github.com/videojs/video.js/wiki/Plugins)). It supports video playback on desktops and mobile devices. This project was started mid 2010, and the player is now used on over ~~50,000~~ ~~100,000~~ 200,000 websites.
+> Video.js is a web video player built from the ground up for an HTML5 world. It supports HTML5 and Flash video, as well as YouTube and Vimeo (through [plugins][plugins]). It supports video playback on desktops and mobile devices. This project was started mid 2010, and the player is now used on over ~~50,000~~ ~~100,000~~ 200,000 websites.
```
BrandonOCasey: is the 200,000 number still accurate? or is it more?

README.md
BrandonOCasey: I think using real urls is good, we just need to make sure we think about the docs if any of these urls change in the future.

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+## Standard `<video>` Element Options
+
+Each of these options is also available as a [standard `<video>` element attribute][video-attrs]; so, they can be defined in all three manners [outlined above](#setting-options). Typically, defaults are not listed as this is left to browser vendors.
+
+### `autoplay`
+
+> Type: `Boolean`
+
+If `true`/present as an attribute, begins playback automatically.
```
BrandonOCasey: "when the player is ready"

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+  Load only the meta data of the video, which includes information like the duration and dimensions of the video.
+
+- `'none'`
+
+  Don't preload any data. The browser will wait until the user hits "play" to begin downloading.
+
+### `src`
+
+> Type: `String`
```
BrandonOCasey: Do we want to specify that `source` elements within the video element can also be used for this same purpose? Should we define those in the file?

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+### `aspectRatio`
+
+> Type: `String`
+
+Puts the player in [fluid](#fluid) mode and the value is used when calculating the dynamic size of the player. The value should represent a ratio - two numbers separated by a colon (e.g. `"16:9"` or `"4:3"`).
+
+### `children`
+
+> Type: `Array`
```
BrandonOCasey: needs a description

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+Video.js indicates that the user is interacting with the player by way of the `"vjs-user-active"` and `"vjs-user-inactive"` classes and the `"useractive"` event.
+
+The `inactivityTimeout` determines how many milliseconds of inactivity is required before declaring the user inactive. A value of `0` indicates that there is no `inactivityTimeout` and the user will never be considered inactive.
+
+### `language`
+
+> Type: `String`
+
+A [language code][lang-codes] matching one of the available languages in the player. This sets the initial language for a player, but it can always be changed.
```
BrandonOCasey: I think that this defaults to english, especially if there is no translation available in the language that you want to use.

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+### `languages`
+
+> Type: `Object`
+
+Customize which languages are available in a player. The keys of this object will be [language codes][lang-codes] and the values will be objects with English keys and translated values.
+
+Learn more about [languages in Video.js](languages.md).
+
+> **Note**: Generally, this option is not needed and it would be better to pass your custom languages to `videojs.addLanguage()`, so they are available in all players!
```
BrandonOCasey: should we deprecate this option then? is there a use case for it?

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+> Type: `String`
+
+Allows overriding the default message that is displayed when Video.js cannot play back a media source.
+
+### `plugins`
+
+> Type: `Object`
+
+This supports having plugins be initialized automatically with custom options when the player is initialized - rather than requiring you to initialize them manually.
```
BrandonOCasey: It would be good to specify something like: `For more information on plugins see here`

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
 
+Although, since the `plugins` option is an object, the order of initialization is not guaranteed!
 
-Individual Options
-------------------
+### `sourceOrder`
 
-> ### Note on Video Tag Attributes ###
-> With HTML5 video tag attributes that can only be true or false (boolean), you simply include the attribute (no equals sign) to turn it on, or exclude it to turn it off. For example, to turn controls on:
+> Type: `Boolean`
```
BrandonOCasey: This should be an `Array`

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+
+- `'auto'`
+
+  Start loading the video immediately (if the browser supports it). Some mobile devices will not preload the video in order to protect their users' bandwidth/data usage. This is why the value is called 'auto' and not something more conclusive like `'true'`.
+
+  _This tends to be the most common and recommended value as it allows the browser to choose the best behavior._
+
+- `'metadata'`
+
+  Load only the meta data of the video, which includes information like the duration and dimensions of the video.
```
BrandonOCasey: "Sometimes the metadata will be loaded by downloading a few frames of the video."

BrandonOCasey:
General comments for this file. 
* We should use casing as defined in [eslint](https://github.com/videojs/eslint-config-videojs/blob/master/eslintrc.json#L179-L187) to be consistent with jsdoc comments and JavaScript itself

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
-or
-{ "controls": true }
-```
+- for each tech:
+  - for each source:
+    - if tech can play source, use this tech/source combo
+
+With `sourceOrder: true`, the algorithm becomes:
 
+- for each source:
```
BrandonOCasey: 👍

docs/guides/setup.md
```diff
@@ -1,124 +1,186 @@
...
 ```
 
-Alternative Setup for Dynamically Loaded HTML
----------------------------------------------
-If your web page or application loads the video tag dynamically (ajax, appendChild, etc.), so that it may not exist when the page loads, you'll want to manually set up the player instead of relying on the data-setup attribute. To do this, first remove the data-setup attribute from the tag so there's no confusion around when the player is initialized. Next, run the following javascript some time after the Video.js javascript library has loaded, and after the video tag has been loaded into the DOM.
-```js
-videojs("example_video_1", {}, function(){
-  // Player (this) is initialized and ready.
-});
+> **Note:** The example with `controls="false"` can be a point of confusion for new developers - it will actually turn controls _on_!
```
BrandonOCasey: 👍

docs/guides/setup.md
```diff
@@ -1,124 +1,186 @@
...
+
+These are correct:
+
+```html
+<video controls ...>
+<video loop="loop" ...>
```
BrandonOCasey: is this correct? I think the way advised above says not to use boolean attributes like this (even though it will work in this case)

docs/guides/setup.md
```diff
@@ -1,124 +1,186 @@
...
+
+```js
+var player = videojs('my-player');
+
+player.on('ready', function() {
+  this.addClass('my-example');
+});
+```
+
+In each case, the callback is called asynchronously - _even if the player is already ready!_
```
BrandonOCasey: Just a general question here, not really about the docs, should we support all three use cases?

docs/index.md
```diff
@@ -1,37 +1,59 @@
...
-* [Tracks](./guides/tracks.md) - Tracks are used for displaying text information over a video, selecting different audio tracks for a video, or selected different video tracks for a video.
+#### [Options][guides-options]
 
-### Customizing
+There are a number of options that can be used to change how the player behaves, starting with the HTML5 media options like autoplay and preload, and expanding to Video.js specific options.
+
+#### [Tracks][guides-tracks]
 
-* [API](./guides/api.md) - The Video.js API allows you to control the video through javascript or trigger event listeners, whether the video is playing through HTML5, flash, or another playback technology.
+Tracks are used for displaying text information over a video, selecting different audio tracks for a video, or selecting different video tracks.
```
BrandonOCasey: if we have a video track guide, we may want to specify that video tracks are untested and may not work.

docs/index.md
```diff
@@ -1,37 +1,59 @@
...
 
-* [Skins](./guides/skins.md) - You can change the look of the player across playback technologies just by editing a CSS file. The skins documentation gives you a intro to how the HTML and CSS of the default skin is put together.
+### Customizing
```
BrandonOCasey: Should we add a guide to `Component`s?

docs/index.md
```diff
@@ -1,37 +1,59 @@
...
+#### [Skins][guides-skins]
 
-* [Tech](./guides/tech.md) - A 'playback technology' is the term we're using to represent HTML5 video, Flash, and other video plugins, as well as other players like the YouTube player. Basically anything that has a unique API to audio or video. Additional playback technologies can be added relatively easily.
+You can change the look of the player across playback technologies just by editing a CSS file. The skins documentation gives you a intro to how the HTML and CSS of the default skin is put together.
 
-* [Plugins](./guides/plugins.md) - You can package up interesting Video.js customizations and reuse them elsewhere. Find out how to build your own plugin or use one created by someone else.
+#### [Plugins][guides-plugins]
 
-### Resources
+You can package up interesting Video.js customizations and reuse them elsewhere. Find out how to build your own plugin or use one created by someone else.
```
BrandonOCasey: I think that it would be good to add a link to the wiki were all the plugins are listed.

README.md
misteroneill: According to builtwith.com it's over 400k now. Will update.

README.md
misteroneill: Sure, we could do some kind of token replacement here. I'm hesitant to do that in this PR. I want to set up some tooling related to docs in a separate PR.

docs/guides/options.md
misteroneill: I didn't list things that don't appear to be supported in the code as `options`.

docs/guides/options.md
misteroneill: Yeah, like I said, I left out things that don't appear to be used.

- `vtt.js` - Adding.
- `starttime` - I think `starttime` was assuming it was coming from an attribute like `starttime="1"`. Probably old code and probably not worth documenting.
- `createEl`, `initChildren`, `reportTouchActivity` - Wouldn't make sense to document since it's always set to `false` explicitly. And I think its existence is a bit of a hack to begin with.
- `nativeControlsForTouch` - Adding here and also under tech options.
- `starttime` - Addressed above.
- `nativeCaptions` and `nativeTextTracks` - Adding under tech options.
- `*Tracks` - These aren't available as options are they?

README.md
BrandonOCasey: maybe add an issue or story and link it here, sounds good though.

docs/guides/options.md
BrandonOCasey: I will add an issue to see if we want to support them

docs/guides/options.md
misteroneill: Well, the `src` is synonymous with the `src` attribute. The variety of ways to set it is outlined at the top of this section.

misteroneill:
The concern I have with adding an example for each option is that it becomes _extremely_ redundant and verbose. The documentation for how to set options is in `setup.md` and there is a blurb at the top of each section explaining commonalities of options in that section (for example, the ability for autoplay to be set via attribute or option).

docs/guides/options.md
```diff
@@ -1,138 +1,312 @@
...
+  - [`languages`](#languages)
+  - [`notSupportedMessage`](#notsupportedmessage)
+  - [`plugins`](#plugins)
+  - [`sourceOrder`](#sourceorder)
+  - [`sources`](#sources)
+  - [`techOrder`](#techorder)
+- [Component Options](#component-options)
+  - [`children`](#children-1)
+  - [`${componentName}`](#componentname)
+- [Tech Options](#tech-options)
```
BrandonOCasey: This is what I mean: 
```js
var sourceHandleOptions = {something:true};
videojs({
  html5: {hls: sourceHandlerOptions},
  flash: {hls: sourceHandlerOptions},
});
```

BrandonOCasey:
My thinking was not an example for each option, but more like an example value for anything that is not a boolean. Such as aspectRatio or width.

README.md
BrandonOCasey: I would even link to `builtwith.com` here so that it i apparent that we are not just making a number up. 👍

docs/guides/options.md
misteroneill: This is the same as the component option and I wanted to keep all the `children` stuff together. I'll link to it, though!

docs/guides/options.md
misteroneill: I'm not sure, honestly. The only thing I can think of is if you want different players on a page to support different language sets. @gkatsev do you know of a use-case/reason for it?

docs/guides/setup.md
misteroneill: Good question, but one that I think is outside the scope of this PR. :)

docs/guides/options.md
misteroneill: Yeah, that'd probably be a 6.0 thing and should probably not block this PR.

docs/guides/options.md
misteroneill: Let's leave the tracks options off if we don't want to publicize it. Also, we should figure out the `starttime` thing. It's not a standard attribute, so I suspect it's legacy and could probably be removed.

docs/guides/options.md
misteroneill: Hmm, well, that's what I'm saying, right? `{hls: sourceHandlerOptions}` gets passed through to source handlers. Or does HLS do some magic that's not in Video.js proper?

misteroneill:
Gotcha. I'll see if there's anything that warrants example(s).

docs/guides/options.md
gkatsev: I think it's mostly for convenience, if you have just one player you could just configure it in the player options.

docs/guides/options.md
BrandonOCasey: hmmm seems like your right sourceHandlers do get all of techs options and not just specific source handler options. I guess maybe just specify that in the Tech options section?

README.md
BrandonOCasey: Added in #3799

docs/guides/options.md
BrandonOCasey: #3780

docs/guides/options.md
BrandonOCasey: #3781

docs/guides/options.md
BrandonOCasey: I think that the convenience may be warranted in this case, but maybe it would be better to make the `languages` option do the same thing as `videojs.addLanguage`?

README.md
```diff
@@ -1,65 +1,105 @@
...
 
+[![Build Status][travis-icon]][travis-link]
+[![Coverage Status][coveralls-icon]][coveralls-link]
+[![Slack Status][slack-icon]][slack-link]
 
-[![NPM](https://nodei.co/npm/video.js.png?downloads=true&downloadRank=true)](https://nodei.co/npm/video.js/)
+[![NPM][npm-icon]][npm-link]
 
-> Video.js is a web video player built from the ground up for an HTML5 world. It supports HTML5 and Flash video, as well as YouTube and Vimeo (through [plugins](https://github.com/videojs/video.js/wiki/Plugins)). It supports video playback on desktops and mobile devices. This project was started mid 2010, and the player is now used on over ~~50,000~~ ~~100,000~~ 200,000 websites.
+> Video.js is a web video player built from the ground up for an HTML5 world. It supports HTML5 and Flash video, as well as YouTube and Vimeo (through [plugins][plugins]). It supports video playback on desktops and mobile devices. This project was started mid 2010, and the player is now used on over ~~50,000~~ ~~100,000~~ ~~200,000~~ 400,000 websites.
```
BrandonOCasey: I think this comment was lost before, I am adding it back in here:

I think we should link to https://trends.builtwith.com/media/VideoJS here so that it i apparent that we are not just making a number up.

docs/guides/setup.md
BrandonOCasey: Also isn't there an option to make this synchronous somehow?

BrandonOCasey: #3782

gkatsev: We don't have any way to make it synchronous. It was removed in 5.0, since previously, it depended on what the tech did, so, html5 was sync but flash was async. Now it's always async. However, #3782 makes sense. I think we may not want all 3, but maybe two of them.

misteroneill: Don't forget, the `Component#ready` function can take the second `sync` argument, which I recall was a point of contention early in the 5.x process.

BrandonOCasey: yeah thats the one that I am talking about, since `player.ready` can be synchronous should we mention that here?

gkatsev: I did forget. And everyone should also forget :P

misteroneill: I think it should remain undocumented at best.

README.md
misteroneill: They shouldn't, they're on the CDN.

docs/guides/options.md
```diff
@@ -1,138 +1,351 @@
...
+Although, since the `plugins` option is an object, the order of initialization is not guaranteed!
 
-Individual Options
-------------------
+See [the plugins guide](plugins.md) for more information on Video.js plugins.
 
-> ### Note on Video Tag Attributes ###
-> With HTML5 video tag attributes that can only be true or false (boolean), you simply include the attribute (no equals sign) to turn it on, or exclude it to turn it off. For example, to turn controls on:
+### `sourceOrder`
+> Type: `Array`
```
gkatsev: `sourceOrder` is a boolean.

gkatsev: Also, should we mention now that in 6.0 it will "default" to true, whatever that means in the world of middleware.

gkatsev: and it defaults to `false` right now.

docs/guides/options.md
```diff
@@ -1,138 +1,351 @@
...
-### preload ###
-The preload attribute informs the browser whether or not the video data should begin downloading as soon as the video tag is loaded. The options are auto, metadata, and none.
+Defines the order in which Video.js techs are preferred. By default, this means that the `Html5` tech is preferred, but Video.js will fall back to `Flash` if no `Html5`-compatible source can be found.
 
-'auto': Start loading the video immediately (if the browser agrees). Some mobile devices like iPhones and iPads will not preload the video in order to protect their users' bandwidth. This is why the value is called 'auto' and not something more final like 'true'.
+### `vtt.js`
+> Type: `string`
 
-'metadata': Load only the meta data of the video, which includes information like the duration and dimensions of the video.
+Allows overriding the default URL to vtt.js, which may be loaded asynchronously to polyfill support for `WebVTT`.
```
gkatsev: We should add a note that it'll be used in the novtt build of videojs.

docs/guides/options.md
```diff
@@ -1,138 +1,351 @@
...
 
-```html
-<video ... data-setup='{ "controlBar": { "muteToggle": false } }'></video>
+```js
+videojs.options.flash.swf = '//path/to/videojs.swf'
 ```
 
-The [components guide](./components.md) has an excellent breakdown of the structure of a player, you
-just need to remember to nest child components in a `children` array for each level.
+#### `nativeCaptions`
```
gkatsev: This is an alias to `nativeTextTracks`. I think it should be mentioned below instead of having its own section.

misteroneill:
Since all my sub-PRs (in my fork) were approved, I have merged all 5 branches into one and rebased on `master`. This should be good to go. 👍

PR #3706, opened 2016-10-24
fix: make sure that document.createElement exists before using

### LINKED ISSUE #3665: es5 build calls document.createElement unchecked
The es5 build contains this JS:
```js
// HTML5 Element Shim for IE8
if (typeof HTMLVideoElement === 'undefined') {
   _document2['default'].createElement('video');
   _document2['default'].createElement('audio');
   _document2['default'].createElement('track');
}
```

In node, `HTMLVideoElement` does not exist, but neither does `document.createElement`.

Can this also check `_document2['default'].createElement` before trying to call it?

gkatsev:
If you try and require videojs in an environment that doesn't implement `document.createElement` properly -- like in Node.js -- you could potentially get an error. This checks that `window.document && window.document.createElement` is available before calling `createElement`.
We specifically check `window.document` so that `global` module won't cause issues with it's shimming of `document.createElement` in `global/document.

Fixes #3665

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -33,7 +33,9 @@ import xhr from 'xhr';
 import Tech from './tech/tech.js';
 
 // HTML5 Element Shim for IE8
-if (typeof HTMLVideoElement === 'undefined') {
+if (typeof HTMLVideoElement === 'undefined' &&
+    window.document &&
+    window.document.createElement) {
   document.createElement('video');
   document.createElement('audio');
   document.createElement('track');


src/js/video.js
```diff
@@ -33,7 +33,9 @@ import xhr from 'xhr';
 import Tech from './tech/tech.js';
 
 // HTML5 Element Shim for IE8
-if (typeof HTMLVideoElement === 'undefined') {
+if (typeof HTMLVideoElement === 'undefined' &&
+    window.document &&
+    window.document.createElement) {
```
misteroneill: What about using `document` here instead of `window.document`? We're already importing it on line 8.

gkatsev: As I mentioned in the description, we can't use `document` because `global/document` tries to shim the DOM API in node.

misteroneill: Ignore me, then. Chalk it up to lack of sleep. 😆

PR #3708, opened 2016-10-24
Update the jsdoc comments to modern syntax - Part 3

BrandonOCasey:
## Description
- player.js
- setup.js
- slider.js

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -44,106 +44,219 @@ import './tracks/text-track-settings.js';
 // Import Html5 tech, at least for disposing the original video tag.
 import './tech/html5.js';
 
+// The following tech events are simply re-triggered
+// on the player when they happen
 const TECH_EVENTS_RETRIGGER = [
   /**
-   * Fired while the user agent is downloading media data
+   * Fired while the user agent is downloading media data.
+   *
+   * @event Player#progress
+   * @type {EventTarget~Event}
+   */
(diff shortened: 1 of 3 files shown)


src/js/player.js
```diff
@@ -889,21 +1074,38 @@ class Player extends Component {
...
+       * @type {EventTarget~Event}
+       */
       this.trigger('firstplay');
     } else {
       // reset the hasStarted state
       this.hasStarted(false);
+      /**
+       * @event Player#loadstart
+       * @type {EventTarget~Event}
+       */
```
misteroneill: This is documented above - does it need to be documented again?

misteroneill: Same is true of `firstplay` in a few places.

src/js/player.js
```diff
@@ -44,106 +44,219 @@ import './tracks/text-track-settings.js';
...
+ *   </video>
  * ```
- * After an instance has been created it can be accessed globally using `Video('example_video_1')`.
  *
- * @param {Element} tag        The original video tag used for configuring options
- * @param {Object=} options    Object of option names and values
- * @param {Function=} ready    Ready callback function
- * @class Player
+ * After an instance has been created it can be accessed globally again
+ * by calling `videojs('example_video_1');`
```
BrandonOCasey: sure I will add that in

PR #3709, opened 2016-10-24
fix: remove unnecessary comments from video.min.js

### LINKED ISSUE #3707: Uglify does not work
CDN is fine, but the "video.min.js" from "video-js-5.11.7.zip" and newer is not properly minified:
filesize `5.11.6`: 281 kb
filesize `5.11.7+`: 494 kb

gkatsev:
This involves updating to latest version of uglify and setting
"screwIE8" to false since that is a default in 2.7. However, this makes
"preserveComments" "some" work as it is supposed to work.

Fixes #3707


gkatsev:
I still need to verify that this works in IE8.

gkatsev:
Verified that it works in IE8.

Issue #3106, opened 2016-02-12
access to video.min.js.map is forbidden


frontendschlampe:
[Error] Failed to load resource: the server responded with a status of 403 (Forbidden) https://vjs.zencdn.net/5.4.6/video.min.js.map

see ticket https://github.com/videojs/video.js/issues/700

OsX El Captain
Safari 9.0.3

http://stackoverflow.com/questions/18407543/video-js-map-throwing-a-404-not-found

gkatsev:
hm... I thought we got that taken care of.

tjbee:
Has this been resolved?

gkatsev:
Nope, unfortunately, still an issue.

gkatsev:
Will be fixed once videojs 3.12 is released as part of #3445. Closing.

jemoreno:
This is still an issue

gkatsev:
What version of videojs is it still an issue for?

jemoreno:
http://vjs.zencdn.net/5.8/video.js.map
http://vjs.zencdn.net/5.9/video.js.map
http://vjs.zencdn.net/5.10/video.js.map
http://vjs.zencdn.net/5.11/video.js.map

They all return a 403 error

jemoreno:
any update on this?

gkatsev:
This will be removed from 5.11 soon (https://github.com/videojs/video.js/pull/3710) and already isn't available in 5.12. We'll consider adding *proper* sourcemaps sometime later.

PR #3710, opened 2016-10-24
fix: remove sourcemaps from the output

### LINKED ISSUE #3701: Sourcemaps don't work right
## Description
Sourcemaps provided with videojs do show up in the source tree of a browser, but do not permit setting of a breakpoint in the right spot. When minified file is included, then sourcemaps don't work at all - showing a source that went into a minification, without listing files individually. Perhaps its time to switch to gulp with better sourcemap tools.

## Steps to reproduce
Open Chrome Developer Tools

1. Go to sources tab
2. Expand source tree provided by video.js script
3. Set a breakpoint in src/js/video.js somewhere inside videojs function
4. Refresh page

### Expected
Breakpoint to stop exactly where you place it and permit inspection of the code.

### Actual
Breakpoints jump all over the code (maps are not in sync with code).

#### videojs
5.11.8
#### browsers
Chrome, and any other that support sourcemaps
#### OSes
MacOS El Capitan
### plugins
None

gkatsev:
Source Maps, while nice, aren't perfect and the ones we were outputting
weren't right.

Fixes #3701, #3106.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 ## HEAD (Unreleased)
 * greenkeeper updated karma dependencies ([view](https://github.com/videojs/video.js/pull/3523))
 * update to latest uglify to fix preserve comments issue. Disable screw ie8 option. ([view](https://github.com/videojs/video.js/pull/3709))
+* remove sourcemap generation ([view](https://github.com/videojs/video.js/pull/3710))
 
 --------------------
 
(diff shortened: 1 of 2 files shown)


gkatsev:
Yeah, we can try and figure out sourcemaps in the future. But frankly, I've yet to have found real benefit from having sourcemaps.

stsvilik:
Benefit is quite great - being able to step through the code looking for potential problems. You should seriously consider using gulp - will build your code faster and would not require you to transpile/build to fs. It is not as bad as you think.

Issue #3272, opened 2016-04-20
Component.on() is too picky


dmlap:
## Description

Component.prototype.on() is really handy because it ties the lifetime of listeners to the owning Component. It's smart enough to work with basically anything that implements an EventTarget-like interface but we [have runtime checks](https://github.com/videojs/video.js/blob/ac3771a329f587b713d33b5085e401c153f894ea/src/js/component.js#L665) that restrict it to only DOM node-ish and Component-ish objects. We should check for the presence of add/removeEventListener instead to make on() useful for objects like native TextTrackLists.
#### videojs

5.x
#### browsers

All.
#### OSes

All.

PR #3727, opened 2016-10-28
fix(events): support adding events to the window object

boushley:
## Fix problem with not being able to add events to the window

When using the `Component#on` function, events added to the `window` object (like a `resize` event) the event handler is silently ignored.

This is caused because the current code uses the presence of a `nodeName` property to determine if the target is a DOM node. For the most part this is a good test, but it doesn't work for the `window` object.
## Specific Changes proposed

This updates the logic for adding event listeners to check if the target is the window object, in addition to the current `nodeName` property check.
This change also adds a final `else` to this if statement that logs a warning so that developers are aware that their event is not being added.
## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -666,18 +666,17 @@ class Component {
       // Add the same function ID so we can easily remove it later
       cleanRemover.guid = fn.guid;
 
-      // Check if this is a DOM node
-      if (first.nodeName) {
-        // Add the listener to the other element
-        Events.on(target, type, fn);
-        Events.on(target, 'dispose', cleanRemover);
-
-      // Should be a component
-      // Not using `instanceof Component` because it makes mock players difficult
-      } else if (typeof first.on === 'function') {
+      // If we are attaching to a component like object use the preferred `on` method
+      if (typeof target.on === 'function') {
         // Add the listener to the other component
         target.on(type, fn);
         target.on('dispose', cleanRemover);
+      } else if (Events.canAttachEvent(target)) {
+        // Add the listener to the other element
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks for the PR. Would you be interested in updating this to make it a bit more generic as mentioned in https://github.com/videojs/video.js/issues/3272?

boushley:
@gkatsev, I missed your comment, sorry about that.

I can definitely update the method according to #3272, was actually working through the issues and came across that one tonight and so was coming here to comment on that.

I'll hopefully get this updated tomorrow.

boushley:
@gkatsev updated to address #3272

boushley:
Also, my brain keeps getting messed up by the fact that in pull 3727 I'm fixing issue 3272 :)

gkatsev:
Heh, that's pretty funny re: issue numbers.
Cool, I'll check it out.

misteroneill:
This is good. I am doing some event-related stuff for 6.0 and this will make a great addition. I can't wait to rebase! 👍

boushley:
Are there any changes we need on this before landing it?

gkatsev:
Probably no more code changes. Needs some testing and verification.

gkatsev:
Unfortunately, I had to revert this because IE8 fails and there's too many logs in the tests :(

gkatsev:
Looks like I can't re-open this. Maybe because I didn't use the "revert" button to revert this.

gkatsev:
See https://travis-ci.org/videojs/video.js/builds/180822016

PR #3736, opened 2016-11-03
Add option to have remoteTextTracks automatically 'garbage-collected' when sources change

imbcmdth:
## Description
Right now "remoteTextTracks" persist between source changes and must be removed manually by the user. This is problematic in many cases. It means plugins that change sources, such as the playlist API or videojs-contrib-ads, have to be extremely careful about changes the source and take care to cleanup remoteTextTracks. 

We do this because the Video.js API aligns very closely with the media element's native API. I propose that we could be doing more to iron-out the rough patches in the native API. In my opinion, the fact that previous changes to the media-element (adding text track elements) may change the behavior of a source is one of those rough patches.

## Specific Changes proposed
`Tech#addRemoteTextTrack` now accepts a second parameter - a boolean named `manualCleanup` which defaults to true preserving backwards compatibility. When that value is set to `false` the text track will be removed from the video element whenever a source change occurs.

A new instance property named `autoRemoteTextTracks_` is added to `Tech` that keeps track of which TextTracks have automatic life-cycle management.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2525,9 +2525,9 @@ class Player extends Component {
    *
    * @param {Object} options    Options for remote text track
    */
-  addRemoteTextTrack(options) {
+  addRemoteTextTrack(...theArgs) {
     if (this.tech_) {
-      return this.tech_.addRemoteTextTrack(options);
+      return this.tech_.addRemoteTextTrack(...theArgs);
     }
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2532,9 +2532,9 @@ class Player extends Component {
    * @deprecated The default value of the "manualCleanup" parameter will default
    * to "false" in upcoming versions of Video.js
    */
-  addRemoteTextTrack(...theArgs) {
+  addRemoteTextTrack(options, manualCleanup) {
     if (this.tech_) {
-      return this.tech_.addRemoteTextTrack(...theArgs);
+      return this.tech_.addRemoteTextTrack(options, manualCleanup);
     }
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2521,9 +2521,16 @@ class Player extends Component {
   }
 
   /**
-   * Add a remote text track
+   * Creates a remote text track object and returns an html track element.
    *
-   * @param {Object} options    Options for remote text track
+   * @param {Object} options The object should contain values for
+   * kind, language, label, and src (location of the WebVTT file)
+   * @param {Boolean} [manualCleanup=true] if set to false, the TextTrack will be
+   * automatically removed from the video element whenever the source changes
+   * @return {HTMLTrackElement} An Html Track Element.
+   * This can be an emulated {@link HTMLTrackElement} or a native one.
+   * @deprecated The default value of the "manualCleanup" parameter will default
+   * to "false" in upcoming versions of Video.js
    */
   addRemoteTextTrack(...theArgs) {
     if (this.tech_) {
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -613,7 +613,6 @@ class Html5 extends Tech {
    *
    * @param {Object} options The object should contain values for
    * kind, language, label and src (location of the WebVTT file)
-   * @method createRemoteTextTrack
    */
   createRemoteTextTrack(options) {
     if (!this.featuresNativeTextTracks) {
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -607,6 +607,14 @@ class Html5 extends Tech {
     return this.el_.addTextTrack(kind, label, language);
   }
 
+  /**
+   * Creates either native TextTrack or an emulated TextTrack depending
+   * on the value of `featuresNativeTextTracks`
+   *
+   * @param {Object} options The object should contain values for
+   * kind, language, label and src (location of the WebVTT file)
+   * @method createRemoteTextTrack
+   */
   createRemoteTextTrack(options) {
     if (!this.featuresNativeTextTracks) {
       return super.createRemoteTextTrack(options);
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -607,18 +607,10 @@ class Html5 extends Tech {
     return this.el_.addTextTrack(kind, label, language);
   }
 
-  /**
-   * Creates a remote text track object and returns a html track element
-   *
-   * @param {Object} options The object should contain values for
-   * kind, language, label and src (location of the WebVTT file)
-   * @return {HTMLTrackElement}
-   */
-  addRemoteTextTrack(options = {}) {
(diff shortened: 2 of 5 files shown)


src/js/tech/tech.js
```diff
@@ -806,39 +839,29 @@ Tech.withSourceHandlers = function(_Tech) {
...
+    this.off('dispose', this.disposeSourceHandler);
+
     if (!sh) {
       // Fall back to a native source hander when unsupported sources are
       // deliberately set
       if (_Tech.nativeSourceHandler) {
         sh = _Tech.nativeSourceHandler;
       } else {
         log.error('No source hander found for the current source.');
+        return;
```
imbcmdth: The addition of this `return` and moving of lines 843-844 to above this conditional block are to fix a tiny issue that may have never been manifest in the wild but which I discovered during a test run.

If a Tech has no `nativeSourceHandler` property and `selectSourceHandler` returned `null` because no matching source handlers were found, then the line later on: `this.sourceHandler_ = sh.handleSource(source, this, this.options_);` would throw an exception because `sh` would stay `null`.

src/js/tech/tech.js
```diff
@@ -531,20 +548,33 @@ class Tech extends Component {
...
+  addRemoteTextTrack(options, manualCleanup) {
     const track = mergeOptions(options, {
       tech: this
     });
-
     const htmlTrackElement = new HTMLTrackElement(track);
 
+    if (manualCleanup !== true && manualCleanup !== false) {
+      // deprecation warning
+      log.warn('Calling addRemoteTextTrack without explicitly setting the "manualCleanup" parameter to `true` is deprecated and default to `false` in future version of video.js');
```
imbcmdth: I am open to some nice word-smithing here. I feel like coming up with error/warnings is not my strong suit.

src/js/tech/tech.js
```diff
@@ -531,20 +548,33 @@ class Tech extends Component {
    *
    * @param {Object} options The object should contain values for
    * kind, language, label and src (location of the WebVTT file)
+   * @param {Boolean} manualCleanup if set to false, the TextTrack will be
+   * automatically removed from the video element whenever the source changes
    * @return {HTMLTrackElement}
    * @method addRemoteTextTrack
    */
-  addRemoteTextTrack(options) {
+  addRemoteTextTrack(options, manualCleanup) {
```
BrandonOCasey: I would make this  `, manualCleanup = true)` here.

src/js/tech/tech.js
```diff
@@ -531,20 +548,33 @@ class Tech extends Component {
    *
    * @param {Object} options The object should contain values for
    * kind, language, label and src (location of the WebVTT file)
+   * @param {Boolean} manualCleanup if set to false, the TextTrack will be
```
BrandonOCasey: this should be `* @param {boolean} [manualCleanup=true]`, since its optional and defaults to true

BrandonOCasey: also you should add an `@depricated` tag to the bottom stating that "The `manualCleanup` parameter will be removed in favor of always doing automatic cleanup in the future."

src/js/tech/tech.js
BrandonOCasey: is there any reason why we need to deprecate this and not just make the change? This almost seems like a bug. If we decide not to deprecate and to just make the change we can probably make this change much smaller.

gkatsev: This is because currently, you have to manually cleanup, we're wanting to change it so that by default tracks are cleaned up automatically.

src/js/tech/tech.js
gkatsev: Having a default option here would make it difficult to do the deprecation warning.

src/js/tech/tech.js
imbcmdth: Done

src/js/tech/html5.js
```diff
@@ -607,18 +607,10 @@ class Html5 extends Tech {
...
 
-  /**
-   * Creates a remote text track object and returns a html track element
-   *
-   * @param {Object} options The object should contain values for
-   * kind, language, label and src (location of the WebVTT file)
-   * @return {HTMLTrackElement}
-   */
-  addRemoteTextTrack(options = {}) {
+  createRemoteTextTrack(options) {
```
gkatsev: this should probably be private.

src/js/tech/tech.js
```diff
@@ -410,18 +412,7 @@ class Tech extends Component {
...
-   * @method emulateTextTracks
-   */
-  emulateTextTracks() {
-    const tracks = this.textTracks();
-
-    if (!tracks) {
-      return;
-    }
-
+  addWebVttScript() {
```
gkatsev: private

src/js/tech/tech.js
```diff
@@ -543,6 +560,14 @@ class Tech extends Component {
     return createTrackHelper(this, kind, label, language);
   }
 
+  createRemoteTextTrack(options) {
```
gkatsev: private

src/js/player.js
```diff
@@ -2525,9 +2525,9 @@ class Player extends Component {
    *
    * @param {Object} options    Options for remote text track
```
BrandonOCasey: this should be `{TextTrackList[]} theArgs track to add as remote text tracks`, also theArgs is a weird name, shouldn't it be tracks?

src/js/tech/html5.js
```diff
@@ -640,11 +639,20 @@ class Html5 extends Tech {
...
+  }
 
-    // store HTMLTrackElement and TextTrack to remote list
-    this.remoteTextTrackEls().addTrackElement_(htmlTrackElement);
-    this.remoteTextTracks().addTrack_(htmlTrackElement.track);
+  /**
+   * Creates a remote text track object and returns a html track element
+   *
+   * @param {Object} options The object should contain values for
+   * kind, language, label and src (location of the WebVTT file)
```
BrandonOCasey: theRest needs to be added to the doc comment, and the name is a bit weird (maybe it will make more sense to me with a doc comment.)

src/js/player.js
```diff
@@ -2521,13 +2521,20 @@ class Player extends Component {
...
+   * @param {Object} options The object should contain values for
+   * kind, language, label, and src (location of the WebVTT file)
+   * @param {Boolean} [manualCleanup=true] if set to false, the TextTrack will be
+   * automatically removed from the video element whenever the source changes
+   * @return {HTMLTrackElement} An Html Track Element.
+   * This can be an emulated {@link HTMLTrackElement} or a native one.
+   * @deprecated The default value of the "manualCleanup" parameter will default
+   * to "false" in upcoming versions of Video.js
+   */
+  addRemoteTextTrack(...theArgs) {
```
BrandonOCasey: this needs to be changed from `...theArgs`

PR #3743, opened 2016-11-04
Use ES6 and allow V8 to optimize mergeOptions

misteroneill:
## Description
The current implementation causes the `mergeOptions` function to be de-optimized. This change improves readability by switching to ES6 syntax _and_ it results in a roughly 75% improvement in the performance of this function by transpiling to a `for` loop instead of `slice.call`.

## Requirements Checklist


BrandonOCasey:
LGTM

PR #3744, opened 2016-11-04
Unify all Track and TrackList APIs

BrandonOCasey:
## Description
* Unify the Video/Audio/Text track APIs

## Requirements Checklist
  - [ ] Tested using BrowserStack/Internet Explorer

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -24,8 +24,7 @@ import mergeOptions from './utils/merge-options.js';
 import textTrackConverter from './tracks/text-track-list-converter.js';
 import ModalDialog from './modal-dialog';
 import Tech from './tech/tech.js';
-import AudioTrackList from './tracks/audio-track-list.js';
-import VideoTrackList from './tracks/video-track-list.js';
+import {ALL as TRACK_TYPES} from './tracks/track-types';
 
 // The following imports are used only to ensure that the corresponding modules
 // are always included in the video.js package. Importing the modules will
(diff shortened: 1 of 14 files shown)


src/js/player.js
```diff
@@ -650,22 +649,25 @@ class Player extends Component {
...
       'preload': this.options_.preload,
       'loop': this.options_.loop,
       'muted': this.options_.muted,
       'poster': this.poster(),
       'language': this.language(),
       'vtt.js': this.options_['vtt.js']
-    }, this.options_[techName.toLowerCase()]);
+    };
+
+    Object.keys(TRACK_TYPES).forEach((type) => {
```
misteroneill: I wonder if this array ought to be stored somewhere instead of calling `Object.keys(TRACK_TYPES)` in multiple places. Either exported from `tracks/track-types` or as a private variable in this module.

src/js/player.js
```diff
@@ -2676,6 +2600,20 @@ class Player extends Component {
   }
 }
 
+Object.keys(TRACK_TYPES).forEach(function(type) {
```
misteroneill: I could be wrong, but it looks like we lost documentation for these methods?

src/js/tracks/track-types.js
```diff
@@ -0,0 +1,44 @@
...
+import VideoTrack from './video-track';
+import HTMLTrackElement from './html-track-element';
+
+import mergeOptions from '../utils/merge-options';
+
+export const NORMAL = {
+  audio: {
+    List: AudioTrackList,
+    Track: AudioTrack,
+    Capital: 'Audio'
```
misteroneill: I think these should maybe be `capital:` instead of `Capital:`. I feel like, as-is, it could be confused for a constructor because of the naming convention. And maybe `capitalName` or something?

gkatsev:
I don't think we should change the API method names.

src/js/tracks/track-types.js
BrandonOCasey: yeah I think that capitalName makes sense.

BrandonOCasey:
I think that it would be possible to put them back to `remoteTextTrackEl` if we want. Any other concerns?

src/js/player.js
gkatsev: track_types could probably expose a `names` field.

BrandonOCasey:
closed so that we can test with browserstack. See #3783

PR #3745, opened 2016-11-04
fix hooks unit test in ie8

BrandonOCasey:
## Description
Fix a unit test in ie8 that is currently broken

## Requirements Checklist

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -107,14 +107,14 @@ function videojs(id, options, ready) {
   }
 
   videojs.hooks('beforesetup').forEach(function(hookFunction) {
-    const opts = hookFunction(tag, videojs.mergeOptions({}, options));
+    const opts = hookFunction(tag, mergeOptions(options));
 
     if (!opts || typeof opts !== 'object' || Array.isArray(opts)) {
       videojs.log.error('please return an object in beforesetup hooks');
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/js/video.js
```diff
@@ -107,14 +107,14 @@ function videojs(id, options, ready) {
   }
 
   videojs.hooks('beforesetup').forEach(function(hookFunction) {
-    const opts = hookFunction(tag, videojs.mergeOptions({}, options));
+    const opts = hookFunction(tag, mergeOptions(options));
```
BrandonOCasey: it was silly to use `videojs.` here but not really a necessary change.

test/unit/video.test.js
```diff
@@ -336,7 +336,7 @@ QUnit.test('beforesetup options override videojs options', function(assert) {
 
   const vid = document.getElementById('test_vid_id');
 
-  videojs.hook('beforesetup', function(options) {
+  videojs.hook('beforesetup', function(tag, options) {
```
BrandonOCasey: I guess video tags in ie8, don't have an `autoplay` property... whoops

gkatsev: Ah, yeah, you have to do `getAttribute` on IE8. Heh, that's pretty funny.

PR #3757, opened 2016-11-07
Object Type-Detection and Replacing object.assign

misteroneill:
## Description
This pulls in a few updates I made to `utils/obj` from #3690 so they can be more isolated and make it into 5.x. 

Perhaps more impactful, this removes `object.assign` as a dependency and re-implements the functionality we were depending on. _This change reduces the minified size of Video.js by over 2KB off the gzipped size._

## Specific Changes proposed
- Introduce the `isObject` method: essentially, `typeof x === 'object'` with a null check.
- Introduce the `isPlain` method: taken from `mergeOptions`, this does a reasonable job at identifying plain objects.
- Remove `object.assign`.

## Requirements Checklist

### DIFF
--- a/src/js/utils/obj.js
+++ b/src/js/utils/obj.js
@@ -28,6 +28,7 @@
  * @return {Mixed}
  *         The new accumulated value.
  */
+import objectAssign from 'object.assign';
 
 /**
  * Array-like iteration for objects.
@@ -62,4 +63,18 @@ export function each(object, fn) {
 export function reduce(object, fn, initial = 0) {
   return Object.keys(object).reduce(
     (accum, key) => fn(accum, object[key], key), initial);
+}
+
+/**
+ * Object.assign-style object merge/extend.
+ *
+ * @param  {Object} target
+ * @param  {Object} ...sources
+ * @return {Object}
+ */
+export function assign(...args) {
+
+  // This exists primarily to isolate our dependence on this third-party
+  // polyfill. If we decide to move away from it, we can do so in one place.
+  return objectAssign(...args);
 }
--- a/src/js/utils/obj.js
+++ b/src/js/utils/obj.js
@@ -78,3 +78,27 @@ export function assign(...args) {
   // polyfill. If we decide to move away from it, we can do so in one place.
   return objectAssign(...args);
 }
+
+/**
+ * Returns whether a value is an object of any kind - including DOM nodes,
+ * arrays, regular expressions, etc. Not functions, though.
+ *
+ * This avoids gotchas like `typeof null === 'object'`.
+ *
+ * @param  {Object} value
+ * @return {Boolean}
+ */
+export function isObject(value) {
+  return !!value && typeof value === 'object';
+}
+
+/**
+ * Returns whether an object appears to be a "plain" object - that is, a
+ * direct instance of `Object`.
+ *
+ * @param  {Object} value
+ * @return {Boolean}
+ */
+export function isPlain(value) {
+  return isObject(value) && value.constructor === Object;
+}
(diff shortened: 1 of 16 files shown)


src/js/utils/obj.js
```diff
@@ -30,3 +30,52 @@ export function reduce(object, fn, initial = 0) {
...
+ * Returns whether a value is an object of any kind - including DOM nodes,
+ * arrays, regular expressions, etc. Not functions, though.
+ *
+ * This avoids gotchas like `typeof null === 'object'`.
+ *
+ * @param  {Object} value
+ * @return {Boolean}
+ */
+export function isObject(value) {
+  return !!value && typeof value === 'object';
```
gkatsev: this directly contradicts the jsdoc comment.

src/js/utils/obj.js
```diff
@@ -30,3 +30,52 @@ export function reduce(object, fn, initial = 0) {
...
+
+/**
+ * Returns whether an object appears to be a "plain" object - that is, a
+ * direct instance of `Object`.
+ *
+ * @param  {Object} value
+ * @return {Boolean}
+ */
+export function isPlain(value) {
+  return isObject(value) && value.constructor === Object;
```
gkatsev: why remove the `toString` check.

BrandonOCasey:
is there any reason not to use the `mergeOptions` in place of `assign` rather than having a new function? Is there any reason to even have `mergeOptions`? should `mergeOptions` just use this new function?

misteroneill:
@BrandonOCasey Yeah, I do think we ought to not have the partial lodash dependency there, but that function is a bit more complex in that it's recursive and has the "customizer" thing; so, it's a bit more prone to unforeseen breaking changes. That said, I do plan to take a stab at it.

src/js/utils/obj.js
misteroneill: I think a copy-paste mistake.

src/js/utils/obj.js
misteroneill: Will clarify! 👍

BrandonOCasey:
SPACE SAVING AHOY! 🚀

src/js/utils/fn.js
```diff
@@ -31,7 +31,34 @@ export const bind = function(context, fn, uid) {
...
+ *
+ * @param  {Function} fn
+ *         The function to be throttled.
+ *
+ * @param  {Number}   wait
+ *         The number of milliseconds by which to throttle.
+ *
+ * @return {Function}
+ */
+export const throttle = function(fn, wait) {
```
gkatsev: can this be done in a separate PR?

src/js/utils/obj.js
```diff
@@ -30,3 +31,55 @@ export function reduce(object, fn, initial = 0) {
...
+}
+
+/**
+ * Returns whether an object appears to be a "plain" object - that is, a
+ * direct instance of `Object`.
+ *
+ * @param  {Object} value
+ * @return {Boolean}
+ */
+export function isPlain(value) {
```
BrandonOCasey: would it be better to use https://lodash.com/docs/4.16.6#isPlainObject?

src/js/utils/obj.js
```diff
@@ -30,3 +31,55 @@ export function reduce(object, fn, initial = 0) {
...
+ * Returns whether a value is an object of any kind - including DOM nodes,
+ * arrays, regular expressions, etc. Not functions, though.
+ *
+ * This avoids the gotcha where using `typeof` on a `null` value
+ * results in `'object'`.
+ *
+ * @param  {Object} value
+ * @return {Boolean}
+ */
+export function isObject(value) {
```
BrandonOCasey: would it be better to just use https://lodash.com/docs/4.16.6#isObject?

misteroneill: My next PR is eliminating lodash as a dependency. It's not worth the extra weight of including lodash just for this.

PR #3760, opened 2016-11-08
Remove lodash

misteroneill:
## Description
This removes lodash-compat as a dependency (though, lodash is preserved as a dev dep). _This reduces the minified+gzipped size of Video.js by 4KB._

## Specific Changes proposed
- Remove lodash-compat as a dependency.
- Add lodash as a dev dependency.
- Reimplement a pared down throttle function in our `fn` utility.
- Reimplement `_.merge` behavior in `mergeOptions`.

## Requirements Checklist

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -5,7 +5,7 @@ import npmRun from 'npm-run';
 module.exports = function(grunt) {
   require('time-grunt')(grunt);
 
-  let _ = require('lodash-compat');
+  let _ = require('lodash');
   let pkg = grunt.file.readJSON('package.json');
   let license = grunt.file.read('build/license-header.txt');
   let bannerCommonData = _.pick(pkg, ['version', 'copyright']);
--- a/build/options-customizer.js
+++ b/build/options-customizer.js
@@ -1,4 +1,4 @@
-import _ from 'lodash-compat';
+import _ from 'lodash';
 /**
  * Customizes _.merge behavior in `gruntOptions` to concatenate
  * arrays. This can be overridden on a per-call basis to
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
   "dependencies": {
     "babel-runtime": "^6.9.2",
     "global": "4.3.0",
-    "lodash-compat": "3.10.2",
     "safe-json-parse": "4.0.0",
     "tsml": "1.0.1",
     "videojs-font": "2.0.0",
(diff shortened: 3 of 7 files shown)


misteroneill:
Note: this includes commits from #3757

src/js/utils/obj.js
```diff
@@ -30,3 +31,55 @@ export function reduce(object, fn, initial = 0) {
...
 }
+
+/**
+ * Object.assign-style object shallow merge/extend.
+ *
+ * @param  {Object} target
+ * @param  {Object} ...sources
+ * @return {Object}
+ */
+export function assign(target, ...sources) {
```
mmcc: FWIW, I use a similar helper extensively at work (Redux app) and we went ahead and defaulted to setting the target to an empty object so the result is always a shallow copy. Not sure if that makes sense here, though.

boushley:
Have we checked the size difference of using our own vs using the individual lodash packages like: https://www.npmjs.com/package/lodash.throttle and

boushley:
Dang, fat fingered the submit button instead of paste. That's what I get for looking at issues on my phone...

https://www.npmjs.com/package/lodash.assign

misteroneill:
@boushley It's a good thought. I did an experiment a few weeks ago with those packages but they have internal dependencies (internal lodash functions, etc) that add a lot more weight than these implementations. lodash.assign for example weighs in at 1.6KB minified and gzipped, which would negate the file size savings.

src/js/utils/obj.js
misteroneill: How does that work if the target is the first argument and the sources are the rest?

boushley:
@misteroneill glad to hear you already investigated it. I suspected there was something like that :)

Given that's been investigated I think these are great alternatives that save quite a bit of bulk :shipit:

src/js/utils/obj.js
gkatsev: I think it basically ends up being `assign({}, target, source1, source2)`.

mmcc: ^^ this.

gkatsev: However, there's one nice usecase of not forcing target to be an empty object:
```js
Object.assign(el.style, {
  'color': 'blue',
  'background': 'red'
});
```
Or similar. But yeah, *most* times, it probably does want to just be an empty object.

misteroneill: Could have another method - `assignNew` or something. That's a bad name, but you get my drift. Sometimes it's convenient to mutate an object.

PR #3768, opened 2016-11-09
test(dom): fix removeElClass test in Safari 10.

gkatsev:
Safari 10 automatically dedupes duplicate class names in an element. So,
our test was failing because we had an extra "foo" in the check. This is
an unlikely scenario that has browser variations, so, better to just
remove it.


hartman:
Alternatively, you could have directly read back the .class value after setting it, and have had all subsequent asserts, depend on the browser having deduped itself.

PR #3772, opened 2016-11-10
fix(html5): exit early on emulated tracks in html5

gkatsev:
addRemoteTextTrack and removeRemoteTextTrack in Html5 tech assumed that
they are always called for native text tracks. However, Html5 tech can
have emulated text tracks. For those, we just exit early after the super
methods are called.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -657,6 +657,10 @@ class Html5 extends Tech {
   addRemoteTextTrack(options, manualCleanup) {
     const htmlTrackElement = super.addRemoteTextTrack(options, manualCleanup);
 
+    if (!this.featuresNativeTextTracks) {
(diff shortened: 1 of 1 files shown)


boushley:
This works and so is great!

I think I would actually prefer using the non-negated if here for both cases. It cleans up the add track case significantly, and the remove track case isn't so big that the extra indentation causes lots of problems. The reason for this is that it makes the return path consistent. For remove this isn't a big deal, but for add it's nice to have a single return since it makes it easier to verify that both cases return the same data.

That said, this is a pretty minor thing for such small methods, so I'm fine as is.

boushley:
I like it!

And if you turn on ignore whitespace it looks nice and clean: https://github.com/videojs/video.js/pull/3772/files?w=1

PR #3774, opened 2016-11-11
fix(flash): lower currentTime cache to 25ms

gkatsev:
Caching the value of currentTime in flash for more than `35ms` seems to cause video playback in HLS via contrib-hls to stop. Set to `25ms` as a nice round number.


gkatsev:
Closing. Talking about this offline, we decided that we should just revert the caching change and try to address it with a better and more complete solution.

PR #3786, opened 2016-11-16
Added option to disable native tracks

forbesjo:
## Description
This PR provides a tech level option to disable native audio and video tracks.

## Specific Changes proposed
This adds `nativeAudioTracks` and `nativeVideoTracks` tech options, this will disable hooking into the native track APIs. This would be useful when using videojs-contrib-hls on Edge.

Usage:
```js
videojs(video, {
  html5: {
    nativeAudioTracks: false
  }
});
```

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -80,7 +80,13 @@ class Tech extends Component {
       this.manualTimeUpdatesOn();
     }
 
-    if (options.nativeCaptions === false || options.nativeTextTracks === false) {
+    ['Text', 'Audio', 'Video'].forEach((track) => {
+      if (options[`native${track}Tracks`] === false) {
+        this[`featuresNative${track}Tracks`] = false;
+      }
+    });
+
+    if (options.nativeCaptions === false) {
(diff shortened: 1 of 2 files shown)


gkatsev:
Also, just realized that https://github.com/videojs/video.js/blob/master/src/js/tech/html5.js#L83-L125 would need to be tweaked a bit to use this as well.
As is, this PR just adds a way to disable the feature switch.

forbesjo:
Why would that block need to be tweaked?

gkatsev:
For the native audio tracks, we want to not add the native tracks to our audio tracks list, right? That block still ads native tracks to their respective lists, even if native versions are disabled.

forbesjo:
My change exposes an option to set this https://github.com/videojs/video.js/blob/master/src/js/tech/html5.js#L89
which will cause it to return early and not hook into the native tracks API.

gkatsev:
D'oh. Totally missed that when scanning that section. Carry on.

Issue #3734, opened 2016-11-02
Better Techs Proposal


gkatsev:
Having worked with Techs and Source Handlers for a while now, we recently found that there was a missing use case to our current support matrix.
The two main issues we were facing
1. we wanted to have a source handler or a tech that could use already preset source handlers or techs. 
For example, if we had a CMS that can put a video id in as a source for the video and we wanted a way to have a source handler/tech translate this video id into an HLS url and have it play back through the contrib-hls source handler.
2. we want to allow techs and source handlers to have a separate timeline that is displayed to the user.
This can be useful if you're working with a live playlist and want to allow the user to seek within it or you have server-side ads.

An interesting solution we were investigating was making source handlers behave like express middleware. During source selection, videojs would cycle through each of the source handlers giving each the source object and allow the source handler to modify it before it is handed to the next source handler. While this sounds like a good idea it only solves issue 1. To solve issue 2, we would need to modify the code a lot so that each method dealing with time was passed through each source handler in each direction. Plus, we already had a way to sit between what the underlying playback engine does and what the control bar and player produce: the Tech.

My proposal is that we update source handlers so that when you inherit from techs, you get to use its source handlers. If a source handler is registered on your tech and it applied to the same source as your parent tech, it would get chosen first. In addition, `"current"` timing methods will be added to the player and tech that the control bar will use -- I have them called `currentCurrentTime` and `currentDuration` in a local proof of concept, but we probably should come up with better names. These will default to use the normal `currentTime` and `duration` methods but if a tech wanted to provide an alternate timeline, they could override these methods. This is not to say that source handlers as middleware shouldn't happen but this proposal is to get the most benefit with the least amount of work.

In addition, work should be done to make techs that load asynchronously not break the player completely. Potentially, re-trying auto-setup on players with errors each time that `registerTech` is called.

In short:
* Allow source handlers to be inherited by Techs
* New Source Handlers can be registered to child techs and they will be chosen first over parent tech's source handlers
* Add new "current" timing methods like `currentCurrentTime` and `currentDuration` that are used by the control bar.
  * Default to the standard `currentTime` and `duration` methods.
  * Can be overridden by child techs to provide an alternate timeline for display
* Make asynchronously loaded techs work with auto-setup.

gkatsev:
Would love to have some input from @bcvio @imbcmdth @heff @incompl @dmlap @misteroneill and whoever else wants to chime in.

gkatsev:
[...]

PR #3788, opened 2016-11-16
feat: middleware

gkatsev:
## Description
This is the initial PR for middleware which hopes to solve the problems from https://github.com/videojs/video.js/issues/3734.
Middleware are objects with two special functions `setSource` and `setTech`. On top of these special methods, it can have getters and setters. Currently, the only allowed ones are `currentTime`, `duration`, and `setCurrentTime`. In the future it can be extended to allow more. Also, methods like `play` could probably be treated as setters and events originating from the video element can be treated as getters and events triggered on the player could be treated as setters. But that's more in the future.

The core of middleware is that video types are basically a routing tree. So, you can "use" a middleware on a particular type but you can also "use" a tech on a particular type using special syntax since techs need a bit extra handling.
```js
// example HLS usage
videojs.use('application/x-mpegurl', require('contrib-hls'));
videojs.use('application/vnd.apple.mpegurl', require('contrib-hls'));

// example tech usage
videojs.use('video/mp4', 'videojs/html5');
videojs.use('application/x-mpegurl', 'videojs/html5');
```

The contrib-hls object could look something like this:
```js
module.exports = {
  setSource(srcObj, next) {
    // next expects an error as a first argument to let the middleware know whether you can handle the current srcObj
    // you can also modify this srcObj and return something else instead which will continue down the middleware tree
    next(false, srcObj);
  },

  setTech(tech) {
    // this gets the tech that ends up getting chosen to back everything. Store it if you need it.
    // the tech still acts as the final source of truth for everything
  },

  currentTime(previousCt) {
    // getters get the value from further up the chain and should return a transformed value, if necessary.
    // In a simple case, the middleware would get the value from the tech and would expect to return a value for the player to use
    return previousCt / 2;
  },

  setCurrentTime(newCt) {
    // setters get the value from further up the chain and should return a transformed value, if necessary.
   // In a simple case, the middleware would get the value from the player and would expect to return a value for the tech to use
  return newCt;
  }
};
```

I have a simple live example here: http://plnkr.co/edit/Ap9Fvg21lpYW3Rnlk9ty?p=preview


package.json
```diff
@@ -20,9 +20,15 @@
     "build": "grunt dist",
     "change": "grunt chg-add",
     "clean": "grunt clean",
+    "grunt": "grunt",
```
gkatsev: Most of this will get removed before it's ready. Makes working with this locally a bit easier.
It'll probably get pulled into a separate PR.

src/js/player.js
```diff
@@ -1255,6 +1266,24 @@ class Player extends Component {
...
+      return value;
+    };
+  }
+
+  middlewareSet(method, arg) {
+    return this.middleware_.reduce(this.middlewareIterator(method), arg);
+  }
+
+  middlewareGet(method) {
+    return this.middleware_.reduceRight(this.middlewareIterator(method), undefined);
```
gkatsev: Finally, I can use `reduce` and `reduceRight`!

src/js/video.js
```diff
@@ -736,19 +736,9 @@ videojs.insertContent = Dom.insertContent;
...
-/*
- * Custom Universal Module Definition (UMD)
- *
- * Video.js will never be a non-browser lib so we can simplify UMD a bunch and
- * still support requirejs and browserify. This also needs to be closure
- * compiler compatible, so string keys are used.
- */
-if (typeof define === 'function' && define.amd) {
-  define('videojs', [], () => videojs);
+module.exports = videojs;
```
gkatsev: This will get pulled into a separate PR.

boushley:
I haven't dug into the code yet, but from the description it seems odd that for `setSource` and `setCurrentTime` follow different patterns.

What is the purpose of the `error` argument passed to `next` in `setSource`?

src/js/video.js
```diff
@@ -736,19 +736,9 @@ videojs.insertContent = Dom.insertContent;
...
 
-// checking that module is an object too because of umdjs/umd#35
-} else if (typeof exports === 'object' && typeof module === 'object') {
-  module.exports = videojs;
-}
+videojs.use = middlewareUse;
+videojs.use('video/mp4', 'videojs/html5');
```
gkatsev: How to register the techs for native functionality is still something I'm a bit unsure about.

src/js/tech/middleware.js
```diff
@@ -0,0 +1,63 @@
...
+    mw.setSource(src, function(err, _src) {
+
+      // something happened, try the next middleware on the current level
+      if (err) {
+        return ssh(src, middleware.slice(1), next, acc);
+      }
+
+      // we've succeeded, now we need to go deeper
+      acc.push(mw);
+      ssh(_src, middlewares[_src.type] || [], next, acc);
```
boushley: We need some way in this case to prevent us from calling the same middleware over and over. If the middleware leaves the `src.type` unchanged and doesn't send along an error, we will keep calling the same middleware over and over.

src/js/tech/middleware.js
```diff
@@ -0,0 +1,63 @@
...
+      ssh(src, middlewares[mw] || [], next, acc);
+    }
+
+  // try the current middleware
+  } else if (mw) {
+    mw.setSource(src, function(err, _src) {
+
+      // something happened, try the next middleware on the current level
+      if (err) {
+        return ssh(src, middleware.slice(1), next, acc);
```
boushley: Instead of doing a `return` here which implies that we care about the return value, can we just put an `else` on the if statement and add the following two lines to it.

This will also make the code more consistent with other recursive calls in this function which don't return.

src/js/tech/middleware.js
```diff
@@ -0,0 +1,63 @@
...
+
+    // we've reached the end, so, we need to get a tech and return.
+    if (split[0] === 'videojs') {
+      const tech = Tech.getTech(toTitleCase(split[1]));
+
+      next(tech, src, acc);
+
+    // we've reached a fork, so, we need go deeper
+    } else {
+      ssh(src, middlewares[mw] || [], next, acc);
```
boushley: So this makes it possible to do something like `videojs.use('video/foo', 'my-standard-middleware-stack')` and when that "middleware" is hit it would then start processing the `my-standard-middleware-stack` set of middlewares, right? Seems interesting.

Although assuming that set of middleware eventually changes the type back to something standard we need to determine if we will ever call the same middleware twice.

src/js/tech/middleware.js
```diff
@@ -0,0 +1,63 @@
+import Tech from './tech.js';
+import toTitleCase from '../utils/to-title-case.js';
+
+const middlewares = {};
+
+export function use(type, middleware) {
+  (middlewares[type] = middlewares[type] || []).push(middleware);
```
boushley: I'm too new to the project to know the general code style guidelines, but I find that lines like this while clever add an unnecessary obstacle to reading the code.

I feel that dividing this up into two lines makes it more idiomatic, and thus easier to read:

```
middlewares[type] = middlewares[type] || []
middlewares[type].push(middleware)
```

src/js/player.js
```diff
@@ -1255,6 +1266,24 @@ class Player extends Component {
...
+      if (mw[method]) {
+        return mw[method](value);
+      }
+
+      return value;
+    };
+  }
+
+  middlewareSet(method, arg) {
+    return this.middleware_.reduce(this.middlewareIterator(method), arg);
```
boushley: Hooray for the right tool for the job! (`reduce` here and `reduceRight` below!)

src/js/player.js
```diff
@@ -1255,6 +1266,24 @@ class Player extends Component {
...
+  }
+
+  middlewareGet(method) {
+    return this.middleware_.reduceRight(this.middlewareIterator(method), undefined);
+  }
```
boushley: It feels like these three middleware functions might belong in the `middleware.js` file instead. Just be that they accept a list of middlewares to process that would be passed in.

src/js/video.js
boushley: Definitely something we need to think about. Because as it stands you can never add a middleware that handles `setSource` for 'video/mp4' since there is a tech registered as the first item always.

src/js/player.js
gkatsev: The methods should be private. The reason why they aren't in `middleware.js` and why `setSource` is special is because `setSource` traverses through our middleware tree and sets up the specific middleware to be used for the player at this time. Though, I guess it can be moved into `middleware.js` and then `this.middleware_` could be passed in.

src/js/tech/middleware.js
boushley: :D

src/js/player.js
boushley: Right, I was assuming that `this.middleware_` would be passed in.

The `player.js` file is big enough as it is. Seems like iterating over a list of middlewares should be in the middlewares file instead.

gkatsev:
Should middleware be given a reference to the `player`?

boushley:
It seems like it could add some new capabilities. In fact to implement the use case I talked about before (unmute sets volume to above 0) you would need to handle the mute setter and be able to set volume on the player.

This could also cause problems if someone does something silly like tries to set `currentTime` on the player while in a middleware that is handling `currentTime`.

src/js/player.js
```diff
@@ -1843,6 +1872,31 @@ class Player extends Component {
...
+    }
+
+    this.cache_.source = src;
+    this.cache_.src = src.src;
+
+    middleware.setSource(src, (tech, src_, mws) => {
+      this.middleware_ = mws;
+      this.loadTech_(tech.name, src_);
+      middleware.setTech(mws, this.tech_);
+    });
```
mmcc: If `tech.name` is the only thing used here, we could destructure in the args. I don't feel strongly about this one, just something I've started doing a little more of lately.

```
middleware.setSource(src, ({ name }, src_, mws) => {
  this.middleware_ = mws;
  this.loadTech_(name, src_);
  middleware.setTech(mws, this.tech_);
});
```

src/js/tech/middleware.js
mmcc: If we're picking this nit, I generally prefer `concat` to `|| []`.

```
middlewares[type] = [].concat(middlewares[type], middlware)
```

This is a little bit of nonsense here since we're ok with mutating the original array, so ¯\\\_(ツ)_/¯

src/js/tech/middleware.js
```diff
@@ -0,0 +1,70 @@
...
+import assign from 'object.assign';
+
+const middlewares = {};
+
+export function use(type, middleware) {
+  (middlewares[type] = middlewares[type] || []).push(middleware);
+}
+
+export function setSource(src, next) {
+  setTimeout(()=>ssh(src, middlewares[src.type] || [], next, []), 1);
```
mmcc: I thought Standard required spaces here (i.e. `() => foo()`)?

I don't feel crazy strong either way, but on [line 16](https://github.com/videojs/video.js/pull/3788/files#diff-b1bec4e9f074fe980f0471fcaaf4df75R16) there are spaces so we're not even being consistent within the same file.

gkatsev: Apparently not but yeah, I agree that there should be more whitespace. Originally added it in to verify that async works.

src/js/player.js
```diff
@@ -2379,6 +2393,91 @@ class Player extends Component {
...
+      } else {
+        // load this technology with the chosen source
+        this.loadTech_(sourceTech.tech, sourceTech.source);
+      }
+
+    } else {
+      return true;
+    }
+
+    return;
```
BrandonOCasey: should we return false here to keep the return value consistent?

src/js/player.js
```diff
@@ -2379,6 +2393,91 @@ class Player extends Component {
...
+      middleware.setTech(mws, this.tech_);
+    });
+
+    return this;
+  }
+
+  src_(source) {
+    const sourceTech = this.selectSource([source]);
+
+    if (sourceTech) {
```
BrandonOCasey: seems like this should be 
```js
if (!sourceTech) {
  return true;
}
```

src/js/player.js
```diff
@@ -2379,6 +2393,91 @@ class Player extends Component {
...
+      if (sourceTech.tech === this.techName_) {
```
BrandonOCasey: It would probably be good to have a similar shortout here. 

```js 
if (sourceTech.tech !== this.techName_) {
  // load this technology with the chosen source
  his.loadTech_(sourceTech.tech, sourceTech.source);
  return false;
}
```

src/js/player.js
```diff
@@ -2379,6 +2393,91 @@ class Player extends Component {
...
+        this.ready(function() {
+
+          // The setSource tech method was added with source handlers
+          // so older techs won't support it
+          // We need to check the direct prototype for the case where subclasses
+          // of the tech do not support source handlers
+          if (this.tech_.constructor.prototype.hasOwnProperty('setSource')) {
+            this.techCall_('setSource', source);
+          } else {
+            this.techCall_('src', source.src);
```
BrandonOCasey: Should we deprecate the old way?

src/js/player.js
```diff
@@ -1530,7 +1535,12 @@ class Player extends Component {
    */
   techCall_(method, arg) {
     // If it's not ready yet, call method when it is
+    if (method in middleware.allowedSetters) {
+      return middleware.set(this.middleware_, this.tech_, method, arg);
+    }
+
```
BrandonOCasey: seems like we should check do this here, to prevent checking `this.tech_` in two places below.

```js
if (!this.tech_) {
  return
}  
```

Also it seems like `this.tech_.ready()` already checks if the tech is ready? Seems like we could merge a lot of this code together to make it much simpler.

src/js/player.js
```diff
@@ -1562,6 +1572,10 @@ class Player extends Component {
   techGet_(method) {
     if (this.tech_ && this.tech_.isReady_) {
 
+      if (method in middleware.allowedGetters) {
```
BrandonOCasey: This function could also use some similar cleanup, maybe we can do both in another PR, but its always bothered me that it seems so messy an nested.

src/js/tech/middleware.js
BrandonOCasey: I also think that "Type" should be renamed to `mimeType`. I am also wondering if it would make more sense for middleware to come first in the arguments order.  I want to `use()` `hls` for `x` type(s). I also thing that type should be able to be an array.

src/js/tech/middleware.js
```diff
@@ -0,0 +1,79 @@
+import { assign } from '../utils/obj.js';
+
+const middlewares = {};
+
+export function use(type, middleware) {
+  (middlewares[type] = middlewares[type] || []).push(middleware);
+}
+
+export function setSource(setTimeout, src, next) {
+  setTimeout(() => ssh(src, middlewares[src.type], next), 1);
```
BrandonOCasey: will we get this far if a `src` does not have a `type`? Should we be checking for this somewhere?

BrandonOCasey: should this be async?

src/js/tech/middleware.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export function setTech(middleware, tech) {
```
BrandonOCasey: middlewareList since it seems like we want it to be an array here?

test/index.html
```diff
@@ -9,6 +9,9 @@
 <body>
   <div id="qunit"></div>
   <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
+  <script>
+    QUnit.config.reorder = false;
```
BrandonOCasey: Should this be in the generator/spellbook?

src/js/tech/middleware.js
misteroneill: @mmcc The only problem with `concat` in that case is that if `middlewares[type]` is `undefined`, you'll get an array like `[undefined, middleware]` which I think is probably undesirable.

src/js/tech/middleware.js
```diff
@@ -0,0 +1,79 @@
...
+  return (value, mw) => {
+    if (mw[method]) {
+      return mw[method](value);
+    }
+
+    return value;
+  };
+}
+
+function ssh(src = {}, middleware = [], next, acc = []) {
```
misteroneill: Agreed. I keep thinking of SSH.

src/js/video.js
```diff
@@ -767,19 +768,6 @@ videojs.insertContent = Dom.insertContent;
...
-  define('videojs', [], () => videojs);
-
-// checking that module is an object too because of umdjs/umd#35
-} else if (typeof exports === 'object' && typeof module === 'object') {
-  module.exports = videojs;
-}
+module.exports = videojs;
 
-export default videojs;
+videojs.use = middlewareUse;
```
misteroneill: There a reason this comes after `module.exports`?

test/index.html
misteroneill: I don't think so. I think setting `reorder` to `false` is really used to mask bad cleanup in tests. I think @gkatsev may have cleaned that up in another PR as well.

test/unit/player.test.js
```diff
@@ -283,6 +281,7 @@ QUnit.test('should asynchronously fire error events during source selection', fu
   });
 
   this.clock.tick(1);
+  this.clock.tick(1);
```
misteroneill: Could we add an explanation as to why we need to tick 2ms instead of 1ms?

package.json
```diff
@@ -25,7 +25,13 @@
     "start": "grunt watchAll",
     "test": "grunt test",
     "toc": "doctoc",
-    "docs": "jsdoc -r src/js -d docs/api -c .jsdoc.json"
+    "docs": "jsdoc -r src/js -d docs/api -c .jsdoc.json",
+    "predev": "babel src/js -d es5",
+    "dev": "run-p babel browserify",
+    "dev-test": "run-p dev test-watch",
+    "babel": "babel -w src/js -d es5",
```
gkatsev: I'm going to pull these out of this PR when getting ready to merge and make a separate PR out of it.

src/js/tech/middleware.js
gkatsev: I went for the two line method.

src/js/player.js
```diff
@@ -829,6 +834,8 @@ class Player extends Component {
       'techId': `${this.id()}_${techName}_api`,
       'videoTracks': this.videoTracks_,
       'textTracks': this.textTracks_,
+      'remoteTextTracks': this.remoteTextTracks_,
+      'remoteTextTrackEls_': this.remoteTextTrackEls_,
```
gkatsev: Unfortunately, this change does need to be in this PR, some of the tests rely on tracks and will fail without it because techs are now async.

src/js/player.js
gkatsev: You're right, we probably want to put the middleware section into one of the ifs.
It's possible that just doing .ready() works, in which case, I can see if it's worth trying to simplify it for this PR.

src/js/player.js
gkatsev: Same here. I'd prefer to do the cleanup in a separate PR but might do it if it helps make the middleware part more clear as well.

src/js/player.js
gkatsev: I don't think we're quite ready to do that yet.

src/js/player.js
gkatsev: yeah, makes it less nested.

src/js/player.js
gkatsev: makes sense.

src/js/video.js
gkatsev: Nope.

src/js/video.js
```diff
@@ -767,19 +768,6 @@ videojs.insertContent = Dom.insertContent;
...
 videojs.computedStyle = computedStyle;
 
-/*
- * Custom Universal Module Definition (UMD)
- *
- * Video.js will never be a non-browser lib so we can simplify UMD a bunch and
- * still support requirejs and browserify. This also needs to be closure
- * compiler compatible, so string keys are used.
- */
-if (typeof define === 'function' && define.amd) {
```
gkatsev: Yeah, I'll pull it out in favor of that. I think I did this because I thought something was broken with something I tried.

test/index.html
gkatsev: I rebased against master, so, I'll try again without it.

test/unit/player.test.js
```diff
@@ -596,6 +595,9 @@ QUnit.test('make sure that controls listeners do not get added too many times',
   player.dispose();
 });
 
+/*
+ * sourceOrder is no longer used with middleware
+ *
 QUnit.test('should select the proper tech based on the the sourceOrder option', function(assert) {
```
gkatsev: I'll remove it right before merging.

test/unit/player.test.js
gkatsev: It's because the error is triggered async, so, we need one for it, but we need an additional one for the player creation itself.

test/index.html
gkatsev: Yep, it can be removed now.

test/unit/player.test.js
gkatsev: I added a comment

src/js/tech/middleware.js
gkatsev: Yep, the name should definitely change.

gkatsev:
I have a question about what `.src` and friends should return. Right now it looks like this:
```html
  <video width=600 height=300 lang="es" id="vid1" class="video-js vjs-default-skin" controls >

      <source src="foo" type="video/foo">
    <source 
      src="//d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8"
      type="application/x-mpegurl"
      >
    <source src="http://vjs.zencdn.net/v/oceans.mp4" type='video/mp4'>
    <source src="http://vjs.zencdn.net/v/oceans.webm" type='video/webm'>
    <source src="http://vjs.zencdn.net/v/oceans.ogv" type='video/ogg'>

    <track kind="captions" src="text.vtt" srclang="en" label="English" default></track>

  </video>
```
And the output of the methods is this:
```js
player.src()
// "blob:http://localhost:8000/a4360e78-1eec-4f58-9be5-3e0f12cab696"
player.currentSrc()
// "//d2zihajmogu5jn.cloudfront.net/sintel/master.m3u8"
player.currentSource()
// {type: "video/foo", src: "foo"}
player.currentSources()
/* [
  {
    "type": "video/foo",
    "src": "foo"
  },
  {
    "type": "application/x-mpegurl",
    "src": "https://s3.amazonaws.com/_bc_dml/example-content/bipbop-advanced/bipbop_16x9_variant.m3u8"
  },
  {
    "type": "video/mp4",
    "src": "http://vjs.zencdn.net/v/oceans.mp4"
  },
  {
    "type": "video/webm",
    "src": "http://vjs.zencdn.net/v/oceans.webm"
  },
  {
    "type": "video/ogg",
    "src": "http://vjs.zencdn.net/v/oceans.ogv"
  }
] */
```

dmlap:
@gkatsev I don't think `player.src()` should return a blob URL unless the app developer (i.e. not the middleware developer) set a blob URL as the source. That was one of the big design goals for source handlers and I think it's still a good idea.

gkatsev:
@dmlap That is the current behavior with source handlers. Also, historically, we've made the methods like `.src()` just match what the video element returns. We've definitely already diverged a bit with regards to `currentSrc()` in this case. Now might be the time to diverge a bit more if we feel that's the right direction.
I kind of like that we have a way to get exactly the src from the video element.

Also, I should add that in the above scenario the `foo` source is the one that was selected.

misteroneill:
I think it makes sense to keep things as similar to the current state as possible. Or adhering to the standard (in the case of `src()` and `currentSrc()`).

As an aside, the `blob:` URL is a consequence of HLS, right?

src/js/player.js
gkatsev: we can't actually do this cleanup here because we need to return a value from this function and `this.ready()` doesn't guarantee a value, even if you're running it synchronously. However, I think I was able to simplify `techCall`.

gkatsev:
After some discussion with @dmlap, we think that `currentSrc()` should regular the actual source that is being set, like `foo`. `src()` should probably return `foo` as well in this case but maybe it should instead return the `m3u8` source. Also, `currentType()` should match the type of the source in `currentSrc()`.

misteroneill:
So, if it's returning the source that was originally set (pre-middleware activity), is there going to be a way to get the post-middleware source?

gkatsev:
I think it would then be up to the middleware and potential corresponding plugin to surface. Plus, there's always asking the tech directly.

gkatsev:
I've moved the grunt and package.json changes into a separate PR: https://github.com/videojs/video.js/pull/3896 and thus I reverted the changes in [this PR](https://github.com/videojs/video.js/pull/3788/commits/7429e65ab712ecbdccebb4acc9b2145bf470a006).

gkatsev:
@mmcc @boushley @misteroneill I've changed the PR from WIP to RFC. I think it's in decent shape now just needs closer inspection.

BrandonOCasey:
@gkatsev It would be a good idea to run the jsdoc linter over this code

src/js/tech/middleware.js
```diff
@@ -0,0 +1,88 @@
...
+      setSourceHelper(_src,
+          src.type === _src.type ? mwrest : middlewares[_src.type],
+          next,
+          acc);
+    });
+  } else if (mwrest.length) {
+    setSourceHelper(src, mwrest, next, acc);
+  } else {
+    // const techs = [].concat(middlewares['*']);
+    // const tech = techs.filter((tech) => !!tech.tech.canPlayType(src.type))[0].name;
```
misteroneill: Should these commented lines just be removed?

PR #3794, opened 2016-11-18
Include vtt.js using require/browserify

BrandonOCasey:
## Description
Right now we concat vtt.js right into the video.js bundle.

## Specific Changes proposed
* add code that wraps around the require for vtt.js so that it can be disabled
* use `envify` to enable or disable a build with vtt.js
* use `unreachable-branch-transform` to remove the require call
* use `browserify-shim` to wrap vtt.js if it is being included

## Requirements Checklist
  - [ ] Docs/guides updated


build/grunt.js
```diff
@@ -24,7 +25,17 @@ module.exports = function(grunt) {
     plugin: [
       ['bundle-collapser/plugin'],
       ['browserify-derequire']
-    ]
+    ],
```
misteroneill: Does videojs-standard not complain about this trailing comma?

src/js/tech/tech.js
```diff
@@ -421,7 +421,12 @@ class Tech extends Component {
     if (!window.WebVTT && this.el().parentNode !== null && this.el().parentNode !== undefined) {
       const script = document.createElement('script');
 
-      script.src = this.options_['vtt.js'] || '../node_modules/videojs-vtt.js/dist/vtt.js';
+      if (process.env.VTT === false) {
+        script.src = this.options_['vtt.js'] || 'https://cdn.rawgit.com/gkatsev/vtt.js/vjs-v0.12.1/dist/vtt.js';
+      }
```
misteroneill: I dig it. Proper pre-processing! 👍

build/grunt.js
BrandonOCasey: hmmm I think that it should? maybe we ignore the whole build directory.

misteroneill: Yeah, it does. Oh well. I don't have a problem with hanging commas. 😆

gkatsev:
I just realized that unfortunately, this doesn't solve all our needs :(
We need to be able to properly support webpack bundling as well as browserify.

gkatsev:
Looks like both envify and unreachable-branch-transform can be run separately from browserify. We could run those on tech.js after running babel before publishing and that should make it so webpack can require it manually. I guess it'll have to be run manually after the alt and dist files are generated so that they don't step on each others toes.

gkatsev:
Another option is to use something like https://github.com/erikdesjardins/babel-plugin-transform-dead-code-elimination, seems like it supports env vars by default and you can configure babel to run or not run plugins depending on env as well.

gkatsev:
Looks like that package is going to be deprecated in favor of babili. It's possible we can use just parts of babili to do what we want. For example `transform-inline-environment-variables` or `transform-node-env-inline` with `minify-dead-code-elimination` or `minify-guarded-expressions`

BrandonOCasey:
I think that we currently have no way to include vtt.js with webpack like your describing (as we concat it onto the browser dist). We default to not including `videojs-vtt.js` in the es5 dist. Do we want to require `videojs-vtt.js` it in the es5 dist by default?

BrandonOCasey:
See #3919

Issue #3755, opened 2016-11-07
Cannot read property 'vdata' after calling dispose


pedromcunha:
## Description
I know this issue has been posted before but I don't think any of the issues posted have been the same as my scenario (maybe I'm wrong, not sure). I've been struggling with an error, namely this one:

`bugsnag-3.min.js:1 VIDEOJS: ERROR: TypeError: Cannot read property 'vdata1478551913479' of null`

I'm working on Angular application that removes the video element from the dom. During the directive's clean up phase I make a call to the dispose method of the videojs element. Upon some further investigation it seems like the dispose method actually removes the element from the dom. So when we get into this helper method here:

```
function hasElData(el) {
  var id = el[elIdAttr];

  if (!id) {
    return false;
  }

  return !!Object.getOwnPropertyNames(elData[id]).length;
}
```

It throws the error mentioned above. Now, I know one quick way of fixing this. I squelched the error by adding this code to the top of the function:

```
 if(!el)
     return false;
```

This works but I don't necessarily want to fork videojs. Is this a fixable issue or am I just doing something completely kooky? Also, I should mention I'm using the videojs hls contrib plugin.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.

1. initialize a videojs player
2. call dispose on said player
3. Error

### Error output
`bugsnag-3.min.js:1 VIDEOJS: ERROR: TypeError: Cannot read property 'vdata1478551913479' of null`

### versions
5.11.x
5.12.x
#### videojs
video.js
#### browsers
Chrome, firefox
#### OSes
Mac OSX
### plugins
video-contrib-hls.js

gkatsev:
That error generally means that something -- a plugin, component, or event handler -- is still trying to use the player in some way after being disposed.
Do you have a reduced test case we can take a look at? Would make it a lot easier to get the root of the problem.

pedromcunha:
Here's an isolated test case, it's using videojs hls plugin as well as angularjs(1.5.x). As you can see the issue is reproducible and it's basically identical to my application code.

https://jsbin.com/yoziqikaha/3/edit?html,js,output

gkatsev:
Looks like an issue with contrib-hls not disposing properly. It seems like we get a `sourceclose` event that we try to do stuff with but it happens after videojs has already disposed to the player for the most part.
[...]

PR #3796, opened 2016-11-21
Fix error when disposing player in combination with HLS plugin

arendjr:
## Description

When I'm using the HLS plugin, I get a "Cannot read property 'vdata1234567890' of null" error when I try to dispose the player. I've tracked the issue down to the following stack trace (sorry if the line numbers don't make sense, both projects have been built into my own project when I'm debugging them):

    hasElData                     video.js:18351
    off                           video.js:19132
    off                           video.js:1365
    (anonymous function)          videojs-contrib-hls.js:14460
    data.dispatcher               video.js:19103
    trigger                       video.js:19213
    EventTarget.trigger           video.js:6242
    ***sourceclose (async)***
    (anonymous function)          video.js:13565
    Html5.disposeMediaElement     video.js:13569
    dispose                       video.js:12718
    dispose                       video.js:8084

Now if we look inside the *disposeMediaElement* method, we see it calls `el.load()` at the end inside an anonymous function. But apparently this call is asynchronous, which I suspect is the cause that when the HLS plugin calls `this.player_.off('mediachange', this.onPlayerMediachange_);` inside its `sourceclose` handler (the top-most anonymous function in the trace above), `player.el_` has become `null` by then.

So I think it might be unreasonable to expect the plugin to anticipate its listeners will be called when the player is already disposed (especially as it appears the cause for this asynchronous invocation is not with the plugin itself), so instead I propose to just attempt to handle this case more gracefully. In this instance, by adding a check inside the `off()` function so that if any plugins are too late unsubscribing their listeners, at least it won't result in errors.

## Specific Changes proposed
Add a `null`-check inside `off()` so unsubscribing listeners from non-existing element never triggers errors anymore. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -279,7 +279,7 @@ export function on(elem, type, fn) {
  */
 export function off(elem, type, fn) {
   // Don't want to add a cache object through getElData if not needed
-  if (!Dom.hasElData(elem)) {
+  if (!elem || !Dom.hasElData(elem)) {
     return;
   }
 


misteroneill:
Thanks for the PR, @arendjr!

My concern with it is that it is merely hiding an error in videojs-contrib-hls rather that fixing the issue, which is that videojs-contrib-hls is not properly cleaning up its listeners/dealing with asynchrony.

Would you mind reporting a bug to videojs-contrib-hls?

gkatsev:
After a lot of back and forth, I do think that we need to prevent the error from being thrown in this case. However, we need to make sure that the error is logged properly as an error.

Also, the HLS team is aware of the issue but it's a bit tricky. This would also fix https://github.com/videojs/video.js/issues/3755

gkatsev:
Unfortunately, we cannot accept this PR as is. We should update all methods that use `hasElData` to log an error (but not throw) in this particular case.

PR #3798, opened 2016-11-21
feat(texttracks): always use emulated text tracks

gkatsev:
Chrome has a bug where if you add a remote text track and try to use it programmatically, you won't get any cues displayed. So, just switch to always emulated mode.

Also, add `IS_SAFARI` and `IS_ANY_SAFARI` to the browsers.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -767,25 +767,7 @@ Html5.canControlPlaybackRate = function() {
  * @return {Boolean}
  */
 Html5.supportsNativeTextTracks = function() {
-  let supportsTextTracks;
-
-  // Figure out native text track support
-  // If mode is a number, we cannot change it because it'll disappear from view.
-  // Browsers with numeric modes include IE10 and older (<=2013) samsung android models.
-  // Firefox isn't playing nice either with modifying the mode
-  // TODO: Investigate firefox: https://github.com/videojs/video.js/issues/1862
-  supportsTextTracks = !!Html5.TEST_VID.textTracks;
-  if (supportsTextTracks && Html5.TEST_VID.textTracks.length > 0) {
-    supportsTextTracks = typeof Html5.TEST_VID.textTracks[0].mode !== 'number';
-  }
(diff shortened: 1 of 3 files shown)


misteroneill:
😖 

LGTM

src/js/tech/html5.js
```diff
@@ -767,25 +767,7 @@ Html5.canControlPlaybackRate = function() {
...
-  }
-  if (supportsTextTracks && browser.IS_FIREFOX) {
-    supportsTextTracks = false;
-  }
-  if (supportsTextTracks && !('onremovetrack' in Html5.TEST_VID.textTracks)) {
-    supportsTextTracks = false;
-  }
-
-  return supportsTextTracks;
+  return browser.IS_ANY_SAFARI;
```
BrandonOCasey: wasn't the issue only with chrome? should we still support native on firefox?

gkatsev: We don't currently support native tracks on Firefox anyway. Firefox 50 may be better but older versions definitely had issues.
Also, I think it would be better to be consistent as much as we can.

BrandonOCasey: ok maybe we should just have an issue/TODO comment to revisit this at another time?

gkatsev: Well, now is the time. Either we should strive to always try and use native tracks or we should try and always use emulated tracks. Safari is a bit of a special case, particularly because of iOS and HLS that we allow native tracks for it.

BrandonOCasey: I am all for always using emulated tracks, because we can control all of the behavior

OwenEdwards: I agree with a move to doing emulated tracks wherever possible; I've been meaning to file an issue because, in order to support Descriptions tracks, you *have* to use emulated tracks even if the browser claims it supports text tracks, and having to switch to emulated tracks just *because* you want to add support for Descriptions tracks isn't going to be popular, if sites have become used to native track styles and controls. Emulated tracks allow the kind of cross-browser consistency, similar to control bar consistency, that is one of video.js' greatest strengths.

PR #3800, opened 2016-11-23
test: fix tests on older IE

gkatsev:


test/unit/test-helpers.js
```diff
@@ -139,22 +139,22 @@ const TestHelpers = {
    * @param  {string} eventType
    */
   triggerDomEvent(element, eventType) {
-    let event;
```
gkatsev: this change is for removing windows line-endings.

test/unit/player.test.js
```diff
@@ -1269,19 +1280,23 @@ QUnit.test('should allow to register custom player when any player has not been
...
 });
 
 QUnit.test('should not allow to register custom player when any player has been created', function(assert) {
   const tag = TestHelpers.makeTag();
   const player = videojs(tag);
 
   class CustomPlayer extends Player {}
-  try {
+
+  assert.throws(function() {
```
gkatsev: `assert.throws` is better.

test/unit/player.test.js
```diff
@@ -1269,19 +1280,23 @@ QUnit.test('should allow to register custom player when any player has not been
 
   assert.equal(player instanceof CustomPlayer, true, 'player is custom');
   player.dispose();
+
+  // reset the Player to the original value;
+  videojs.registerComponent('Player', Player);
```
gkatsev: This fixes the actual issue with custom player on IE8.

src/js/component.js
```diff
@@ -1443,7 +1443,9 @@ class Component {
     if (name === 'Player' && Component.components_[name]) {
       const Player = Component.components_[name];
 
-      if (Player.players && Object.keys(Player.players).length > 0) {
+      if (Player.players &&
+          Object.keys(Player.players).length > 0 &&
+          Object.keys(Player.players).map((playerName) => Player.players[playerName]).every(Boolean)) {
```
gkatsev: Not sure whether we want a comment here but basically, if all the items in Player.players are null, you're allowed to register a custom player. This is partly so we can reset the Player component properly for the test.

misteroneill: I think it deserves some kind of comment. Basically, what you just wrote! 😆

gkatsev: Done

test/unit/test-helpers.js
BrandonOCasey: good catch!

test/unit/tracks/text-track-controls.test.js
```diff
@@ -435,7 +436,11 @@ test('chapters should be displayed when remote track added and load event fired'
 
   equal(chaptersEl.track.cues.length, 2);
 
-  TestHelpers.triggerDomEvent(chaptersEl, 'load');
+  if (player.tech_.featuresNativeTextTracks) {
+    TestHelpers.triggerDomEvent(chaptersEl, 'load');
```
BrandonOCasey: why/how are these two statements different?

gkatsev: On IE8 we use a custom DOM element for text tracks, however, IE8 doesn't allow us to trigger custom events on a custom element.
So, because everywhere we support native text tracks, we can trigger a real DOM event, we use that and where we use emulated tracks, we just trigger `load` directly on the chapters element object.

gkatsev: I can add a comment in the code for this, if you think it's necessary.

BrandonOCasey: So should this check for IE8 instead? or maybe just add a comment to specify that this is a protection for IE8?

gkatsev: I decided that a general solution would be better than specifically checking for IE8. However, I'll add a comment about this.

PR #3802, opened 2016-11-26
Make registerComponent only work with Components

misteroneill:
## Description
This addresses one of the pre-existing cards from the 6.0 project. It prevents techs (and others) from being registered via `registerComponent`.

There is an open question as to whether or not we want to implement similar protections to `registerTech` and adopt `deregisterComponent` and `deregisterTech` methods similar to the way the plugins PR does for plugins.

## Specific Changes proposed
- `registerComponent` now throws if given an object that is not a subclass of `Component` (or `Component` itself).
- `registerComponent` now throws if given a non-empty string as its `name` argument.

## Requirements Checklist

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1541,12 +1541,16 @@ class Component {
    * @return {Component}
    *         The `Component` that was registered.
    */
-  static registerComponent(name, comp) {
-    if (!name) {
-      return;
+  static registerComponent(name, Comp) {
+    if (typeof name !== 'string' || !(/\S/).test(name)) {
+      throw new Error('illegal "name"; must be non-empty string');
     }
 
-    name = toTitleCase(name);
+    if (!Component.prototype.isPrototypeOf(Comp.prototype) && Component !== Comp) {
+      throw new Error('illegal "Comp"; must be a subclass of Component');
+    }
+
+    name = toTitleCase(name.trim());
 
     if (!Component.components_) {
       Component.components_ = {};
@@ -1566,9 +1570,9 @@ class Component {
       }
     }
 
-    Component.components_[name] = comp;
+    Component.components_[name] = Comp;
 
-    return comp;
+    return Comp;
   }
 
   /**
@@ -1661,4 +1665,5 @@ class Component {
 }
 
 Component.registerComponent('Component', Component);
+
 export default Component;
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1571,16 +1571,19 @@ class Component {
       Component.components_ = {};
     }
 
-    if (name === 'Player' && Component.components_[name]) {
-      const players = Component.components_[name].players;
+    const Player = Component.getComponent('Player');
+
+    if (name === 'Player' && Player) {
+      const players = Player.players;
+      const playerNames = Object.keys(players);
 
       // If we have players that were disposed, then their name will still be
       // in Players.players. So, we must loop through and verify that the value
       // for each item is not null. This allows registration of the Player component
       // after all players have been disposed or before any were created.
       if (players &&
-          Object.keys(players).length > 0 &&
-          Object.keys(players).map((playerName) => players[playerName]).every(Boolean)) {
+          playerNames.length > 0 &&
+          playerNames.map((pname) => players[pname]).every(Boolean)) {
         throw new Error('Can not register Player component after player has been created.');
       }
     }
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1535,44 +1535,45 @@ class Component {
    * @param {string} name
    *        The name of the `Component` to register.
    *
-   * @param {Component} comp
+   * @param {Component} ComponentToRegister
    *        The `Component` class to register.
    *
    * @return {Component}
    *         The `Component` that was registered.
    */
-  static registerComponent(name, Comp) {
-    if (typeof name !== 'string' || !(/\S/).test(name)) {
-      throw new Error('illegal "name"; must be non-empty string');
+  static registerComponent(name, ComponentToRegister) {
+    if (typeof name !== 'string' || !name.length) {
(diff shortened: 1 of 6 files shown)


src/js/component.js
```diff
@@ -1425,16 +1425,20 @@ class Component {
...
-    if (!name) {
-      return;
+  static registerComponent(name, Comp) {
+    if (typeof name !== 'string' || !(/\S/).test(name)) {
+      throw new Error('illegal "name"; must be non-empty string');
     }
 
-    name = toTitleCase(name);
+    if (!Component.prototype.isPrototypeOf(Comp.prototype) && Component !== Comp) {
+      throw new Error('illegal "Comp"; must be a subclass of Component');
```
BrandonOCasey: I think this error is a little vague. Could be more like "The component ${name} that was ordered for registration, was not a subclass of `Component`. This is required.

src/js/component.js
```diff
@@ -1425,16 +1425,20 @@ class Component {
...
-   * @param {Object} comp The component to register
+   * @param {Object} Comp The component to register
    * @static
    * @method registerComponent
    */
-  static registerComponent(name, comp) {
-    if (!name) {
-      return;
+  static registerComponent(name, Comp) {
+    if (typeof name !== 'string' || !(/\S/).test(name)) {
```
BrandonOCasey: will this match empty string? EDIT: it may be a good idea just to comment what that regex matches

src/js/component.js
```diff
@@ -1425,16 +1425,20 @@ class Component {
...
+      throw new Error('illegal "name"; must be non-empty string');
     }
 
-    name = toTitleCase(name);
+    if (!Component.prototype.isPrototypeOf(Comp.prototype) && Component !== Comp) {
+      throw new Error('illegal "Comp"; must be a subclass of Component');
+    }
+
+    name = toTitleCase(name.trim());
```
BrandonOCasey: Should this go at the top? Then we would just have to check for empty string?

BrandonOCasey: it might be worth warning that the component name contains trailing/leading spaces and that we are trimming it here?

src/js/component.js
misteroneill: Sure. It may be overkill, though. I dunno. I am sometimes accused of being over-defensive. 😄 

It matches a string with any non-space character. The opposite of `\s`.

src/js/component.js
misteroneill: I tend to prefer shorter errors, but perhaps the part before the semicolon could be clearer.

src/js/component.js
misteroneill: I probably shouldn't trim it honestly. But I'll see about moving it up.

misteroneill: I am removing the `trim()`.

src/js/component.js
```diff
@@ -1471,12 +1476,6 @@ class Component {
     if (Component.components_ && Component.components_[name]) {
       return Component.components_[name];
     }
-
-    if (window && window.videojs && window.videojs[name]) {
-      log.warn(`The ${name} component was added to the videojs object when it should be registered using videojs.registerComponent(name, component)`);
-
-      return window.videojs[name];
-    }
```
misteroneill: Removed this section since it was deprecated in 5.x.

src/js/component.js
```diff
@@ -1425,13 +1425,18 @@ class Component {
...
+   * @param {Object} ComponentToRegister The component to register
    * @static
    * @method registerComponent
    */
-  static registerComponent(name, comp) {
-    if (!name) {
-      return;
+  static registerComponent(name, ComponentToRegister) {
+    if (typeof name !== 'string' || !name.length) {
+      throw new Error(`illegal component name, "${name}"; must be a non-empty string`);
```
BrandonOCasey: won't this throw an "uncaught reference error" or something like that if name is undefined?

misteroneill: That's a good point.

misteroneill: Actually, no, it shouldn't. It will only get to the second part of the expression if it's a string.

src/js/component.js
```diff
@@ -1535,15 +1535,20 @@ class Component {
...
+   * @param {Component} ComponentToRegister
    *        The `Component` class to register.
    *
    * @return {Component}
    *         The `Component` that was registered.
    */
-  static registerComponent(name, comp) {
-    if (!name) {
-      return;
+  static registerComponent(name, ComponentToRegister) {
```
gkatsev: I think we also want to check if comp is a Tech and throw an error in that case.

src/js/component.js
```diff
@@ -1553,22 +1572,22 @@ class Component {
     }
 
     if (name === 'Player' && Component.components_[name]) {
-      const Player = Component.components_[name];
+      const players = Component.components_[name].players;
```
gkatsev: is it guaranteed that `Component.components_[name]` exists? I know when I added the check below it wasn't.

misteroneill: It's not, but it's checked above in line 1574. I guess that could be clearer. Will update.

src/js/component.js
```diff
@@ -1535,15 +1535,34 @@ class Component {
...
    */
-  static registerComponent(name, comp) {
-    if (!name) {
-      return;
+  static registerComponent(name, ComponentToRegister) {
+    if (typeof name !== 'string' || !name) {
+      throw new Error(`Illegal component name, "${name}"; must be a non-empty string.`);
+    }
+
+    const Tech = Component.getComponent('Tech');
```
BrandonOCasey: should we use this or import?

misteroneill: `Tech` depends on `Component`, so `import` is out of the question unfortunately.

BrandonOCasey: ok makes sense

PR #3805, opened 2016-11-28
added an faq

BrandonOCasey:
## Description
Added an FAQ under guides

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/docs/guides/faq.md
+++ b/docs/guides/faq.md
@@ -1,50 +1,62 @@
 # FAQ
 
 ## Q: What is video.js?
-video.js is an extendable wrapper around the native video element. It does the following:
+video.js is an extendable framework/library around the native video element. It does the following:
   * Offers a plugin API so that different types of video can be handed to the native
     video element. IE:  HLS, Flash, or HTML5 video.
   * Unifies the native video api across browsers (polyfilling support for features
-  * if necessary)
+    if necessary)
   * Offers an extendable and themable UI
   * Takes care of accessibility for the user (in-progress)
+  * Has a set of core plugins that offer support for tons of additional video formats.
+    * [videojs-contrib-hls](http://github.com/videojs/videojs-contrib-hls)
+    * [videojs-contrib-dash](http://github.com/videojs/videojs-contrib-hls)
+  * Support for DRM video via a core plugin
+    * [videojs-contrib-eme](https://github.com/videojs/videojs-contrib-eme)
+  * Lots of plugins which offer support for all kinds of features. See the [plugin list on videojs.com](http://videojs.com/plugins)
 
 ## Q: How do I install video.js?
-Currently video.js can be installed using bower, npm, or by serving a release file from
-a github tag. For information on doing any of those see the [install guide](http://videojs.com/getting-started/).
+Currently video.js can be installed using bower, npm, serving a release file from
+a github tag, or even using a CDN hosted version. For information on doing any of those
+see the [install guide](http://videojs.com/getting-started/).
 
 ## Q: What do video.js version numbers mean?
 video.js follows [semver](http://semver.org/) which means that the API should not change
-out from under a userunless there is a major version increase.
+out from under a user unless there is a major version increase.
 
 ## Q: How can I troubleshoot playback issues?
 See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on troubleshooting playback issues
 
 ## Q: A video does not play in a specific browser. Why?
-Chrome and Safari both rely on byte-range requests for playing back video. So It is important to
-Make sure that the location which is serving your video files supports byte-range requests.
-Often PHP serving video is an issue as it does not support byte-range requests by default.
+See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on
+why this might be happening. Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues).
+Make sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
+with your issue.
 
 ## Q: Why does the entire video download before playback? Why does the video load for a long time?
-This could mean that your video is not encoded with the metadata -- things like duration,
-video dimensions, etc -- at the beginning of the video. This means that the browser will
-need to load the entire video before it know how long the video is and what to do with
-the video.
+See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on
+why this might be happening. Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues).
+Make sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
 
 ## Q: I see an error thrown that mentions `vdata12345`. What is that?
-This error is thrown when an element that is associated with a component is removed
-from the DOM but the event handlers associated with the element are not removed. This
-is almost always due to event listeners not being disposed when dispose is called on
-a component.
+See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on
+why this might be happening. Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues).
+Make sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
 
 ## Q: What media formats does video.js support?
-This depends on the browsers support, video.js will work off of that. If Flash is
-avialable Flash videos can also be played but only if the Flash tech is included with
-video.js.
+This depends on the formats supported by the browser's HTML5 video element, and the playback
+techs made available to video.js. For example, video.js 5 includes the Flash tech by default which
+enables the playback of FLV video where the Flash plugin is available.
 
 ## Q: How can I hide the links to my video/subtitles/audio/tracks?
-This is outside of the scope of video.js and there are things that can be done to make
-it harder, but it will never be completely possible.
+It's impossible to hide the network requests a browser makes and difficult to
+sufficiently obfuscate URLs in the source. Techniques such as token authentication may
+help but are outside of the scope of video.js.
+
+For content that must be highly secure [videojs-contrib-eme](https://github.com/videojs/videojs-contrib-eme) adds DRM support.
 
 ## Q: I think I found a bug with video.js or I want to add a feature. What should I do?
 ### if you think that you can fix the issue or add the feature
@@ -64,7 +76,12 @@ problem that you are facing in isolation. Think of it as example page that repro
 the issue in the least amount of possible code. We have a [starter example](http://jsbin.com/axedog/edit?html,output)
 for reduced test cases.
 
-## Q: How do a make a plugin for video.js?
+## Q: What is a plugin?
+A plugin is a group of reusable functionality that can be re-used by others. For instance a plugin could add
+a button to video.js that makes the video replay 10 times in a row before it stops playback for good. If such
+a plugin existed and was published users could include it on their page to share that functionality.
+
+## Q: How do I make a plugin for video.js?
 See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js.
 
 ## Q: How do I add a button to video.js?
@@ -83,7 +100,7 @@ See the [video.js github wiki](https://github.com/videojs/video.js/wiki/Skins).
 ## Q: Does video.js work as an audio only player?
 Yes! It can be used to play audio only files in a `<video>` or `<audio>` tag. The
 difference being that the `<audio>` tag will not have a blank display area and the `<video>`
-tag will.
+tag will. Note that audio only will not work with the Flash playback tech.
 
 ## Q: Does video.js support audio tracks?
 Yes! See the [audio tracks guide](/docs/guides/audio-tracks.md) for information on using audio tracks.
@@ -105,7 +122,9 @@ video.js itself does not support MPEG DASH, but a project called [videojs-contri
 which is maintained by the video.js organization.
 
 ## Q: Does video.js support live video?
-Yes! Video.js adds support for live videos.
+Yes! Video.js adds support for live videos via the Flash tech which supports RTMP streams.
+[videojs-contrib-hls](http://github.com/videojs/videojs-contrib-hls) will add support for live HLS video
+if you add it to your page with video.js.
 
 ## Q: Can video.js be required in node.js?
 Yes! Please submit an issue if this does not work. See the question on submiting issues
--- a/docs/guides/faq.md
+++ b/docs/guides/faq.md
@@ -1,91 +1,103 @@
-## FAQ
+# FAQ
 
-**Q: What is video.js?**
-> video.js is an extendable wrapper around the native video element. It does the following:
->  * Offers a plugin API so that different types of video can be handed to the native video element. IE:  HLS, Flash, or HTML5 video.
->  * Polyfills and unifies the native video api across browsers
->  * Offers an extendable and themable UI
->  * Takes care of accessibility for the user (in-progress)
+## **Q: What is video.js?**
+video.js is an extendable wrapper around the native video element. It does the following:
+  * Offers a plugin API so that different types of video can be handed to the native
+    video element. IE:  HLS, Flash, or HTML5 video.
+  * Unifies the native video api across browsers (polyfilling support for features
+  * if necessary)
+  * Offers an extendable and themable UI
+  * Takes care of accessibility for the user (in-progress)
 
-**Q: How do I install video.js?**
-> Currently video.js can be installed using bower, npm, or by serving a release file
-> from a github tag. For information on doing any of those see the [install guide](http://videojs.com/getting-started/).
+## **Q: How do I install video.js?**
+Currently video.js can be installed using bower, npm, or by serving a release file from
+a github tag. For information on doing any of those see the [install guide](http://videojs.com/getting-started/).
 
-**Q: What do video.js version numbers mean?**
-> video.js follows [semver](http://semver.org/)
+## **Q: What do video.js version numbers mean?**
+video.js follows [semver](http://semver.org/) which means that the API should not change
+out from under a userunless there is a major version increase.
 
-**Q: How can I troubleshoot playback issues?**
-> See the [troubleshooting guide](/docs/guides/troubleshooting.md)
+## **Q: How can I troubleshoot playback issues?**
+See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on troubleshooting playback issues
 
-**Q: What media formats does video.js support?**
-> This depends on the browsers support, video.js will work off of that. If Flash is avialable Flash videos
-> can also be played but only if the Flash tech is included with video.js.
+## **Q: What media formats does video.js support?**
+This depends on the browsers support, video.js will work off of that. If Flash is
+avialable Flash videos can also be played but only if the Flash tech is included with
+video.js.
 
-**Q: How can I hide the links to my video?**
-> This is outside of the scope of video.js and there are things that can be done to make it harder, but it will
-> never be completely possible.
+## **Q: How can I hide the links to my video/subtitles/audio/tracks?**
+This is outside of the scope of video.js and there are things that can be done to make
+it harder, but it will never be completely possible.
 
-**Q: I think I found a bug with video.js. What should I do?**
-> If you know how to code and feel comfortable looking into the issue, we will
-> gladly review a pull request if you open one on the [video.js repo](https://github.com/videojs/video.js/pulls).
-> If you don’t have any idea what that means, please open an [issue on the video.js repo](https://github.com/videojs/video.js/issues)
-> and make sure that you follow the issue template so that we can better assist you.
-> If the bug relates to a plugin, please file the issue against the plugin’s repository.
+## **Q: I think I found a bug with video.js or I want to add a feature. What should I do?**
+### if you think that you can fix the issue or add the feature
+Submit a pull request to the [video.js repo](https://github.com/videojs/video.js/pulls).
+Make sure to follow the [contributing guide](/CONTRIBUTING.md#contributing-code) and
+the [pull request template](/.github/PULL_REQUEST_TEMPLATE.md).
 
-**Q: What is a [reduced test case](https://css-tricks.com/reduced-test-cases/)?**
-> A [reduced test case](https://css-tricks.com/reduced-test-cases/) is an example of the problem that you
-> are facing in isolation. Basically it is an example page that reproduces the issue in the least amount
-> of possible code. We have a [starter example](http://jsbin.com/axedog/edit?html,output)
+### If you don't think you can fix the issue or add the feature
+Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues). Make
+sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
+with your issue.
 
-**Q: How do a make a plugin for video.js?**
-> See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js
+## **Q: What is a [reduced test case](https://css-tricks.com/reduced-test-cases/)?**
+A [reduced test case](https://css-tricks.com/reduced-test-cases/) is an example of the
+problem that you are facing in isolation. Think of it as example page that reproduces
+the issue in the least amount of possible code. We have a [starter example](http://jsbin.com/axedog/edit?html,output)
+for reduced test cases.
 
-**Q: How do I add a button to video.js?**
-> See the [button guide](/docs/guides/button.md) for information on adding a button to video.js
+## **Q: How do a make a plugin for video.js?**
+See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js.
 
-**Q: Where can I find a list of video.js plugins?**
-> A list of plugins is avialable on [videojs.com](http://videojs.com/plugins).
+## **Q: How do I add a button to video.js?**
+See the [button guide](/docs/guides/button.md) for information on adding a button to video.js.
 
-**Q: How can I get my plugin listend on the website?**
-> Add the 'videojs-plugin' keyword to your package.json and publish your package to npm.
+## **Q: Where can I find a list of video.js plugins?**
+A list of plugins is avialable on [videojs.com](http://videojs.com/plugins).
 
-**Q: Where can I find a list of video.js skins?**
-> See the [video.js github wiki](https://github.com/videojs/video.js/wiki/Skins).
+## **Q: How can I get my plugin listend on the website?**
+Add the 'videojs-plugin' [keyword to your array in package.json](https://docs.npmjs.com/files/package.json#keywords)
+and publish your package to npm.
 
-**Q: Does video.js work as an audio only player?**
-> Yes! Video.js can be used with an `<audio>` tag the same way that it is used with a `<video>` tag.
+## **Q: Where can I find a list of video.js skins?**
+See the [video.js github wiki](https://github.com/videojs/video.js/wiki/Skins).
 
-**Q: Does video.js support multiple audio tracks?**
-> Yes! See the guide for [audio tracks](/docs/guides/audio-track.md)
+## **Q: Does video.js work as an audio only player?**
+Yes! It can be used to play audio only files in a `<video>` or `<audio>` tag. The
+difference being that the `<audio>` tag will not have a blank display area and the `<video>`
+tag will.
 
-**Q: Does video.js support multiple video tracks?**
-> Theoretically it does but the code has yet to be tested, if it works or you find an
-> issues feel free to open a pull request or submit an issue. The guide for video tracks is
-> [here](/docs/guides/video-track.md)
+## **Q: Does video.js support audio tracks?**
+Yes! See the [audio tracks guide](/docs/guides/audio-tracks.md) for information on using audio tracks.
 
-**Q: Does video.js support multiple text tracks?**
-> Yes! See the guide for [text tracks](/docs/guides/text-track.md)
+## **Q: Does video.js support video tracks?**
+The code for video tracks exists but it has not been tested. See the [video tracks guide](/docs/guides/video-tracks.md)
+for more information on using video tracks.
 
-**Q: Does video.js support HLS (HTTP Live streaming) video?**
-> video.js supports HLS video if the native HTML5 element supports HLS
-> (e.g. Safari, Edge, Android Chrome and iOS). For browsers without native
-> support see the [videojs-contrib-hls](https://github.com/videojs/videojs-contrib-hls) project
-> which adds support.
+### **Q: Does video.js support text tracks (captions, subtitles, etc)?**
+Yes! See the [text tracks guide](/docs/guides/text-track.md) for information on using text tracks.
 
-**Q: Does video.js support MPEG Dash video?**
-> video.js itself does not support MPEG DASH, but a project called
-> [videojs-contrib-dash](https://github.com/videojs/videojs-contrib-dash) which is maintained by
-> the video.js organization does.
+## **Q: Does video.js support HLS (HTTP Live streaming) video?**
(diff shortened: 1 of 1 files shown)


docs/guides/faq.md
```diff
@@ -0,0 +1,82 @@
+## FAQ
+
+**Q: How do I install video.js?**
+> Currently video.js can be installed using bower, npm, or by serving a release file
+> from a github tag. For information on doing any of those see the [install guide](http://videojs.com/getting-started/).
+
+**Q: What is video.js?**
```
misteroneill: I think this ought to be first question.

docs/guides/faq.md
```diff
@@ -0,0 +1,82 @@
...
+> video.js is an extendable wrapper around the native video element. It does the following:
+>  * Offers a plugin API so that different types of video can be handed to the native video element. IE:  HLS, flash, or html5 video.
+>  * Polyfills and unifies the native video api across browsers
+>  * Offers an extendable and themable UI
+>  * Takes care of accessibility for the user (in-progress)
+
+**Q: How is video.js versioned?**
```
misteroneill: I think I would rephrase this question as "What do video.js version numbers mean?"

docs/guides/faq.md
```diff
@@ -0,0 +1,82 @@
...
+> See the [troubleshooting guide](/docs/guides/troubleshooting.md)
+
+**Q: What is a [reduced test case](https://css-tricks.com/reduced-test-cases/)?**
+> A [reduced test case](https://css-tricks.com/reduced-test-cases/) is an example of the problem that you
+> are facing in isolation. Basically it is an example page that reproduces the issue in the least amount
+> of possible code. We have a starter example [here](http://jsbin.com/axedog/edit?html,output)
+
+**Q: What media formats does video.js support?**
+> This depends on the browsers support, video.js will work off of that. If flash is avialable flash videos
+> can also be played but only if the flash tech is included with video.js.
```
misteroneill: Flash should be capitalized, since it's a proper noun.

docs/guides/faq.md
```diff
@@ -0,0 +1,82 @@
...
+>  * Offers an extendable and themable UI
+>  * Takes care of accessibility for the user (in-progress)
+
+**Q: How is video.js versioned?**
+> video.js follows [semver](http://semver.org/)
+
+**Q: How can I troubleshoot playback issues?**
+> See the [troubleshooting guide](/docs/guides/troubleshooting.md)
+
+**Q: What is a [reduced test case](https://css-tricks.com/reduced-test-cases/)?**
```
misteroneill: I think this question ought to go below the "I think I found a bug..." question.

docs/guides/faq.md
```diff
@@ -0,0 +1,82 @@
...
+> If the bug relates to a plugin, please file the issue against the plugin’s repository.
+
+**Q: How do a make a plugin for video.js?**
+> See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js
+
+**Q: How do I add a button to video.js?**
+> See the [button guide](/docs/guides/button.md) for information on adding a button to video.js
+
+**Q: Where can I find a list of video.js plugins?**
+> See the video.js wiki for a running list [here](http://videojs.com/plugins).
```
misteroneill: We should probably "deprecate" the wiki page and point it to this page. Also, "See the video.js wiki" is inaccurate here.

misteroneill: Also, we should probably mention how a plugin gets added to that list (i.e. it's published to npm with the "videojs-plugin" keyword, IIRC).

docs/guides/faq.md
```diff
@@ -0,0 +1,82 @@
...
+> can also be played but only if the flash tech is included with video.js.
+
+**Q: How can I hide the links to my video?**
+> This is outside of the scope of video.js and there are things that can be done to make it harder, but it will
+> never be completely possible.
+
+**Q: I think I found a bug with video.js. What should I do?**
+> If you know how to code and feel comfortable looking into the issue, we will
+> gladly review a pull request if you open one on the [video.js repo](https://github.com/videojs/video.js/pulls).
+> If you don’t have any idea what that means, please open an [issue here](https://github.com/videojs/video.js/issues)
```
misteroneill: I think it's generally considered bad form from a usability/accessibility perspective to use a bunch of "here" links. An alternative would be:

```
If you don't have any idea what that means, please [open an issue on GitHub](...)
```

Linking relevant text makes the link more semantically meaningful for uses browsing with a keyboard or other assistive technology.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Q: How do I install video.js?**
+> Currently video.js can be installed using bower, npm, or by serving a release file
+> from a github tag. For information on doing any of those see the [install guide](http://videojs.com/getting-started/).
+
+**Q: What do video.js version numbers mean?**
+> video.js follows [semver](http://semver.org/)
+
+**Q: How can I troubleshoot playback issues?**
+> See the [troubleshooting guide](/docs/guides/troubleshooting.md)
```
BrandonOCasey: This guide does not currently exist, we should probably work on adding it though.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+**Q: What is a [reduced test case](https://css-tricks.com/reduced-test-cases/)?**
+> A [reduced test case](https://css-tricks.com/reduced-test-cases/) is an example of the problem that you
+> are facing in isolation. Basically it is an example page that reproduces the issue in the least amount
+> of possible code. We have a [starter example](http://jsbin.com/axedog/edit?html,output)
+
+**Q: How do a make a plugin for video.js?**
+> See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js
+
+**Q: How do I add a button to video.js?**
+> See the [button guide](/docs/guides/button.md) for information on adding a button to video.js
```
BrandonOCasey: This guide does not exits, is something like this going to be added by the manual documentation PR's?

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+**Q: Does video.js work as an audio only player?**
+> Yes! Video.js can be used with an `<audio>` tag the same way that it is used with a `<video>` tag.
+
+**Q: Does video.js support multiple audio tracks?**
+> Yes! See the guide for [audio tracks](/docs/guides/audio-track.md)
+
+**Q: Does video.js support multiple video tracks?**
+> Theoretically it does but the code has yet to be tested, if it works or you find an
+> issues feel free to open a pull request or submit an issue. The guide for video tracks is
+> [here](/docs/guides/video-track.md)
```
BrandonOCasey: I think this is added by the manual documentation PR's

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+> Yes! Video.js adds support for live videos.
+
+**Q: Can video.js be required in node.js?**
+> Yes! Please submit an issue if this does not work.
+
+**Q: Does video.js work with webpack?**
+> Yes! Please submit an issue if this does not work.
+
+**Q: is video.js on bower?**
+> Yes! See the [install guide](http://videojs.com/getting-started/) for more information.
```
BrandonOCasey: is there a better link for this that is not going to change when the website changes?

docs/guides/faq.md
misteroneill: Good point. I don't think that's in scope at the moment, but we can answer the question at a very basic level here and update it to point to a guide when we have one?

We could add a section to the component guide rather than a whole new guide. Just another option.

docs/guides/faq.md
misteroneill: Yeah, but it's `video-tracks.md` instead.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
+## FAQ
+
+**Q: What is video.js?**
+> video.js is an extendable wrapper around the native video element. It does the following:
+>  * Offers a plugin API so that different types of video can be handed to the native video element. IE:  HLS, Flash, or HTML5 video.
+>  * Polyfills and unifies the native video api across browsers
```
gkatsev: It doesn't usually polyfill anything unless completely broken. It generally sits on top to make a unified version.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
+## FAQ
+
+**Q: What is video.js?**
```
gkatsev: I think each question should be a heading.
So, at the top, FAQ should be `#` and each question should be `##`.
```markdown
# FAQ

## What is video.js?
Video.js is a video player.
```

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Q: How can I hide the links to my video?**
+> This is outside of the scope of video.js and there are things that can be done to make it harder, but it will
+> never be completely possible.
+
+**Q: I think I found a bug with video.js. What should I do?**
+> If you know how to code and feel comfortable looking into the issue, we will
+> gladly review a pull request if you open one on the [video.js repo](https://github.com/videojs/video.js/pulls).
+> If you don’t have any idea what that means, please open an [issue on the video.js repo](https://github.com/videojs/video.js/issues)
+> and make sure that you follow the issue template so that we can better assist you.
```
gkatsev: link to the issue template here.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Q: What media formats does video.js support?**
+> This depends on the browsers support, video.js will work off of that. If Flash is avialable Flash videos
+> can also be played but only if the Flash tech is included with video.js.
+
+**Q: How can I hide the links to my video?**
+> This is outside of the scope of video.js and there are things that can be done to make it harder, but it will
+> never be completely possible.
+
+**Q: I think I found a bug with video.js. What should I do?**
```
gkatsev: We should link to the appropriate section of the CONTRIBUTING.md.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+> A list of plugins is avialable on [videojs.com](http://videojs.com/plugins).
+
+**Q: How can I get my plugin listend on the website?**
+> Add the 'videojs-plugin' keyword to your package.json and publish your package to npm.
+
+**Q: Where can I find a list of video.js skins?**
+> See the [video.js github wiki](https://github.com/videojs/video.js/wiki/Skins).
+
+**Q: Does video.js work as an audio only player?**
+> Yes! Video.js can be used with an `<audio>` tag the same way that it is used with a `<video>` tag.
```
gkatsev: Not sure if we need to add it here but the spec (and videojs's different between the two going forward) is that the only difference between the two is that with `video` you have a display area and with `audio`, you do not.
You can play audio files in a video element and you can play a video file in an audio element.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Q: Does video.js support multiple audio tracks?**
+> Yes! See the guide for [audio tracks](/docs/guides/audio-track.md)
+
+**Q: Does video.js support multiple video tracks?**
+> Theoretically it does but the code has yet to be tested, if it works or you find an
```
gkatsev: I think it would be better to say something like:
> The code for supporting video tracks exists but it has not been tested.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+**Q: How can I get my plugin listend on the website?**
+> Add the 'videojs-plugin' keyword to your package.json and publish your package to npm.
+
+**Q: Where can I find a list of video.js skins?**
+> See the [video.js github wiki](https://github.com/videojs/video.js/wiki/Skins).
+
+**Q: Does video.js work as an audio only player?**
+> Yes! Video.js can be used with an `<audio>` tag the same way that it is used with a `<video>` tag.
+
+**Q: Does video.js support multiple audio tracks?**
```
gkatsev: We should call it just `audio tracks` and `video tracks` below.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+> Yes! See the guide for [audio tracks](/docs/guides/audio-track.md)
+
+**Q: Does video.js support multiple video tracks?**
+> Theoretically it does but the code has yet to be tested, if it works or you find an
+> issues feel free to open a pull request or submit an issue. The guide for video tracks is
+> [here](/docs/guides/video-track.md)
+
+**Q: Does video.js support multiple text tracks?**
```
gkatsev: just `text tracks`. Also, potentially add captions as well.

docs/guides/faq.md
```diff
@@ -0,0 +1,91 @@
...
+> Yes! See the guide for [text tracks](/docs/guides/text-track.md)
+
+**Q: Does video.js support HLS (HTTP Live streaming) video?**
+> video.js supports HLS video if the native HTML5 element supports HLS
+> (e.g. Safari, Edge, Android Chrome and iOS). For browsers without native
```
gkatsev: Chrome on Android is called `Chrome for Android`.

docs/guides/faq.md
gkatsev: I think the byte range/php hosting question need to be in this FAQ, unless the troubleshooting guide is going to be an FAQ of more technical topics.

BrandonOCasey: sure just tell me how to word it and I will add it in here, I think that it can be in both places.

docs/guides/faq.md
BrandonOCasey: should the `>` stay or should the text just be under the heading?

gkatsev: Just text under the heading.

docs/guides/faq.md
BrandonOCasey: is there a guide planned on this? As it seems like it has come up a fair amount

gkatsev: Not right now. There's actually some work that needs to be done for that too, like actually disabling the display area for an audio element instead of having ~40px of display area.

BrandonOCasey: ok I will open an issue to add a troubleshooting and an audio only guide

BrandonOCasey:
Added issues to make guides for [troubleshooting](#3810) and [audio only](#3811)

docs/guides/faq.md
```diff
@@ -0,0 +1,103 @@
+# FAQ
+
+## **Q: What is video.js?**
```
gkatsev: I don't think we need the bolding or `Q:` anymore.

gkatsev:
A couple more questions:

## The video does not play in my browser of choice
Make sure that however you are serving your video files supports byte-range requests. Chrome and Safari both rely on byte range requests for playing back video, so, the video will not be able to play.
This is likely the case if you are serving the file using PHP as it does not support byte-range requests by default

## The browser loads the entire video before playing or takes a long time to load
This could mean that your video is not encoded with the metadata -- things like duration, video dimensions, etc -- at the beginning of the video. The browser then needs to load the entire video before it know how long the video is and what to do with the video.

## I see an error thrown like `vdata12345`
This probably means that a component used was not being disposed properly. The element associated with the component was removed from the DOM but event handlers associated with the component are still live.

@mister-ben you replied to a lot of questions like the first two, can you proof read them? Also, can you take a look at this PR?

mister-ben:
## The video does not play in my browser of choice
Make sure that however you are serving your video files supports byte-range requests. Chrome and Safari both rely on byte range requests for playing back video, so, the video will not be able to play. On other browsers you may experience different behaviour, for example the video may play but return to the beginning whenever you try to seek.

This is commonly a problem if your source URL is pointing to a PHP or other script that is fetching and returning the file. Your script must respond correctly to range headers.

## The browser loads the entire video before playing or takes a long time to load
This could mean that your video is not encoded with the metadata -- things like duration, video dimensions, etc -- at the beginning of the video. This metadata is known as the "moov atom" in MP4. Depending on the browser the whole video may need to be downloaded before it's playable, or playback may be able to start but seeking beyond the downloaded range won't work.

Many encoders put the metadata in the correct place by default, others may require you to choose a "fast start" or "optimise for streaming" option.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
+# FAQ
+
+## Q: What is video.js?
+video.js is an extendable wrapper around the native video element. It does the following:
```
mister-ben: framework rather than wrapper?

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+  * Offers a plugin API so that different types of video can be handed to the native
+    video element. IE:  HLS, Flash, or HTML5 video.
+  * Unifies the native video api across browsers (polyfilling support for features
+  * if necessary)
```
mister-ben: no bullet on this line

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+  * Offers an extendable and themable UI
+  * Takes care of accessibility for the user (in-progress)
+
+## Q: How do I install video.js?
+Currently video.js can be installed using bower, npm, or by serving a release file from
+a github tag. For information on doing any of those see the [install guide](http://videojs.com/getting-started/).
```
mister-ben: Alternatively, you can use the CDN hosted version.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Q: What do video.js version numbers mean?
+video.js follows [semver](http://semver.org/) which means that the API should not change
+out from under a userunless there is a major version increase.
```
mister-ben: user unless

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+This could mean that your video is not encoded with the metadata -- things like duration,
+video dimensions, etc -- at the beginning of the video. This means that the browser will
+need to load the entire video before it know how long the video is and what to do with
+the video.
+
+## Q: I see an error thrown that mentions `vdata12345`. What is that?
+This error is thrown when an element that is associated with a component is removed
+from the DOM but the event handlers associated with the element are not removed. This
+is almost always due to event listeners not being disposed when dispose is called on
+a component.
```
mister-ben: If troubleshooting is going to be a separate doc, this would be better there

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Q: What media formats does video.js support?
+This depends on the browsers support, video.js will work off of that. If Flash is
+avialable Flash videos can also be played but only if the Flash tech is included with
+video.js.
```
mister-ben: This depends on the formats supported by the browser's HTML5 video element, and the playback techs made available to video.js. For example,  video.js 5 includes the Flash tech by default which enables the playback of FLV video where the Flash plugin is available.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues). Make
+sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
+with your issue.
+
+## Q: What is a [reduced test case](https://css-tricks.com/reduced-test-cases/)?
+A [reduced test case](https://css-tricks.com/reduced-test-cases/) is an example of the
+problem that you are facing in isolation. Think of it as example page that reproduces
+the issue in the least amount of possible code. We have a [starter example](http://jsbin.com/axedog/edit?html,output)
+for reduced test cases.
```
mister-ben: http://jsbin.com/axedog has an old version of video.js

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Q: How do a make a plugin for video.js?
+See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js.
```
mister-ben: A "what is a plugin?" question would be useful first.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+Add the 'videojs-plugin' [keyword to your array in package.json](https://docs.npmjs.com/files/package.json#keywords)
+and publish your package to npm.
+
+## Q: Where can I find a list of video.js skins?
+See the [video.js github wiki](https://github.com/videojs/video.js/wiki/Skins).
+
+## Q: Does video.js work as an audio only player?
+Yes! It can be used to play audio only files in a `<video>` or `<audio>` tag. The
+difference being that the `<audio>` tag will not have a blank display area and the `<video>`
+tag will.
```
mister-ben: While Flash is still around, worth noting that audio only doesn't work in Flash.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Q: Does video.js support audio tracks?
+Yes! See the [audio tracks guide](/docs/guides/audio-tracks.md) for information on using audio tracks.
+
+## Q: Does video.js support video tracks?
+The code for video tracks exists but it has not been tested. See the [video tracks guide](/docs/guides/video-tracks.md)
+for more information on using video tracks.
```
mister-ben: "Video track support is in development" ?

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+video.js supports HLS video if the native HTML5 element supports HLS (e.g. Safari, Edge,
+Chrome for Android, and iOS). For browsers without native support see the [videojs-contrib-hls](https://github.com/videojs/videojs-contrib-hls)
+project which adds support.
+
+## Q: Does video.js support MPEG Dash video?
+video.js itself does not support MPEG DASH, but a project called [videojs-contrib-dash](https://github.com/videojs/videojs-contrib-dash)
+which is maintained by the video.js organization.
+
+## Q: Does video.js support live video?
+Yes! Video.js adds support for live videos.
```
mister-ben: The Flash tech supports simple RTMP streams, and [videojs-contrib-hls](https://github.com/videojs/videojs-contrib-hls) add support for live HLS.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Q: Can video.js be required in node.js?
+Yes! Please submit an issue if this does not work. See the question on submiting issues
+to find out how.
```
mister-ben: Duplicates the "how do I install" question earlier.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
+# FAQ
+
+## Q: What is video.js?
+video.js is an extendable wrapper around the native video element. It does the following:
+  * Offers a plugin API so that different types of video can be handed to the native
+    video element. IE:  HLS, Flash, or HTML5 video.
+  * Unifies the native video api across browsers (polyfilling support for features
+  * if necessary)
+  * Offers an extendable and themable UI
+  * Takes care of accessibility for the user (in-progress)
```
mister-ben: I think it's worth introducing core plugins like videojs-contrib-* as part of the ecosystem early as they make a big difference to the feature set.

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+## Q: A video does not play in a specific browser. Why?
+Chrome and Safari both rely on byte-range requests for playing back video. So It is important to
+Make sure that the location which is serving your video files supports byte-range requests.
+Often PHP serving video is an issue as it does not support byte-range requests by default.
+
+## Q: Why does the entire video download before playback? Why does the video load for a long time?
+This could mean that your video is not encoded with the metadata -- things like duration,
+video dimensions, etc -- at the beginning of the video. This means that the browser will
+need to load the entire video before it know how long the video is and what to do with
+the video.
```
mister-ben: I drafted some expanded answers for these at https://github.com/videojs/video.js/pull/3805#issuecomment-263834925

docs/guides/faq.md
```diff
@@ -0,0 +1,120 @@
...
+a component.
+
+## Q: What media formats does video.js support?
+This depends on the browsers support, video.js will work off of that. If Flash is
+avialable Flash videos can also be played but only if the Flash tech is included with
+video.js.
+
+## Q: How can I hide the links to my video/subtitles/audio/tracks?
+This is outside of the scope of video.js and there are things that can be done to make
+it harder, but it will never be completely possible.
```
mister-ben: It's impossible to hide the network requests a browser makes and difficult to sufficiently obfuscate URLs in the source. Techniques such as token authentication may help but are outside of the scope of video.js.

For content that must be highly secure [videojs-contrib-eme](https://github.com/videojs/videojs-contrib-eme) adds DRM support.

mister-ben:
These are some drafts I'd made but they're not in a Q&A format. Some duplicate and some belong in troubleshooting:

## Formats

- Make sure you are using formats that video.js can play. Which formats can be played depends on your browser, OS and which tech plugins you are using.

- Most browsers can play MP4 with h264 video. If you want to have a single source, and neither live streaming nor adaptive streaming is not a consideration, MP4 is a good choice.

- For adaptive streaming or live streaming, use HLS or DASH. You'll need to use (videojs-contrib-hls) or (videojs-contrib-dash). HLS is convenient as mobile browsers have native support.

- It's not only the file container format (e.g. MP4) that is important but also the codecs used. MP4 support in HTML5 browsers is specifically for MP4 with h264 video and MP3 or AAC audio.

- Some low-end phones save video in the 3GP format but give the file an MP4 extension. These files will not play.

- The Flash tech supports FLV video and partially RTMP. These will not play without Flash. If you are still using Flash videos, you need to migrate away as browsers end Flash support. SWF is not a video format and will not play event in the Flash tech.

## Embedding

- Always specify the correct mime type for your videos, whether using `<source>` elements or the `src()` API. This is used to determine what video.js will attempt to play.

## Hosting video

- Your server must support [byte range requests](LINK). Most servers support this by default. If you are proxying the media files via a server side script, this script must implement ranges. The impact of not doing this ranges from seeking being broken to no playback at all (on iOS).

- Your server must return the correct mime type for the format.

- If you are using HLS or DASH your server must [implement CORS headers](enablecors.org)

## Fullscreen

- If your player is in an iframe, make sure it and any parent iframes have the fullscreen attributes `allowfullscreen web…`. Without these, the browser will not allow fullscreen mode.

docs/guides/faq.md
misteroneill: Hmm, I don't really feel like either is quite appropriate. Library is probably the most accurate term. I think I've used "decorates a native video element" as a description before.

gkatsev: videojs.com calls it "The Player Framework"

misteroneill:
As discussed, we'll try to stick to content and organization in this PR and I can do my grammar nitpicking in a separate venue. :)

docs/guides/faq.md
BrandonOCasey: this is still up in the air, but I think we may make a troubleshooting guide and link this question there when we have it.

docs/guides/faq.md
BrandonOCasey: @gkatsev can this be updated or do we have to change the url?

docs/guides/faq.md
BrandonOCasey: I think all the code is done, but we never got around to testing it. It could very well work for some use cases at this point.

docs/guides/faq.md
BrandonOCasey: I think this question was to address the fact that video.js can be used server side in node.js. Some packages on npm can only be used in the browser, as weird as that sounds.

BrandonOCasey:
@mister-ben Thanks for all the troubleshooting info. I added another [PR for troubleshooting](#3814)

docs/guides/faq.md
misteroneill: The guide will now mention that it's "experimental". It should work _in theory_ but since nobody supports it yet...

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+## Q: What is a plugin?
+A plugin is a group of reusable functionality that can be re-used by others. For instance a plugin could add
+a button to video.js that makes the video replay 10 times in a row before it stops playback for good. If such
+a plugin existed and was published users could include it on their page to share that functionality.
+
+## Q: How do I make a plugin for video.js?
+See the [plugin guide](/docs/guides/plugins.md) for information on making a plugin for video.js.
+
+## Q: How do I add a button to video.js?
+See the [button guide](/docs/guides/button.md) for information on adding a button to video.js.
```
misteroneill: Can we comment this out until we get a button guide?

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Q: Where can I find a list of video.js plugins?
+A list of plugins is avialable on [videojs.com](http://videojs.com/plugins).
+
+## Q: How can I get my plugin listend on the website?
```
misteroneill: "listend" -> "listed"

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+Add the 'videojs-plugin' [keyword to your array in package.json](https://docs.npmjs.com/files/package.json#keywords)
+and publish your package to npm.
```
misteroneill: Maybe mention that the generator provides this by default? 😄

docs/guides/faq.md
misteroneill: The guide will mention that it's an experimental technology at the top.

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+### Q: Does video.js support text tracks (captions, subtitles, etc)?
+Yes! See the [text tracks guide](/docs/guides/text-track.md) for information on using text tracks.
+
+## Q: Does video.js support HLS (HTTP Live streaming) video?
+video.js supports HLS video if the native HTML5 element supports HLS (e.g. Safari, Edge,
+Chrome for Android, and iOS). For browsers without native support see the [videojs-contrib-hls](https://github.com/videojs/videojs-contrib-hls)
+project which adds support.
+
+## Q: Does video.js support MPEG Dash video?
+video.js itself does not support MPEG DASH, but a project called [videojs-contrib-dash](https://github.com/videojs/videojs-contrib-dash)
```
misteroneill: "but a project" -> "but there is a project"

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+which is maintained by the video.js organization.
+
+## Q: Does video.js support live video?
+Yes! Video.js adds support for live videos via the Flash tech which supports RTMP streams.
+[videojs-contrib-hls](http://github.com/videojs/videojs-contrib-hls) will add support for live HLS video
+if you add it to your page with video.js.
+
+## Q: Can video.js be required in node.js?
+Yes! Please submit an issue if this does not work. See the question on submiting issues
```
misteroneill: "submiting" -> "submitting"

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+to find out how.
+
+## Q: Does video.js work with webpack?
+Yes! Please submit an issue if this does not work. See the question on submiting issues
```
misteroneill: "submiting" -> "submitting"

docs/guides/faq.md
```diff
@@ -0,0 +1,139 @@
...
+See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on
+why this might be happening. Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues).
+Make sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
+
+## Q: I see an error thrown that mentions `vdata12345`. What is that?
+See the [troubleshooting guide](/docs/guides/troubleshooting.md) for information on
+why this might be happening. Open an [issue on the video.js repo](https://github.com/videojs/video.js/issues).
+Make sure that you follow the [issue template](/.github/ISSUE_TEMPLATE.md) and the
+[contributing guide](/CONTRIBUTING.md#filing-issues) so that we can better assist you
```
BrandonOCasey: do we want to have two places discussing the same thing in detail? or should player-workflow link to contributing as well?

docs/guides/faq.md
BrandonOCasey: So should I just bring this answer in line with the others? Or should I keep the mention of it not having been tested.

docs/guides/faq.md
```diff
@@ -0,0 +1,179 @@
...
+Yes! Please [submit an issue or open a pull request][pr-issue-question] if this does not work.
+
+## Q: Does video.js work with webpack?
+Yes! Please [submit an issue or open a pull request][pr-issue-question] if this does not work.
+
+## Q: Is video.js on bower?
+Yes! See the [install guide][install-guide] for more information.
+
+<!-- guides -->
+[plugin-guide]: /docs/guides/plugins
```
gkatsev: all these links need the .md extension on them.

PR #3827, opened 2016-12-02
Remove TimeRanges without an index deprecation warning

misteroneill:
Remove TimeRanges without an index deprecation warning.

## Requirements Checklist

### DIFF
--- a/src/js/utils/time-ranges.js
+++ b/src/js/utils/time-ranges.js
@@ -2,7 +2,6 @@
  * @file time-ranges.js
  * @module time-ranges
  */
-import log from './log.js';
 
 /**
  * Returns the time for the specified index at the start or end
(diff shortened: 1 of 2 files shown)


src/js/utils/time-ranges.js
```diff
@@ -1,16 +1,10 @@
-import log from './log.js';
-
 function rangeCheck(fnName, index, maxIndex) {
   if (index < 0 || index > maxIndex) {
     throw new Error(`Failed to execute '${fnName}' on 'TimeRanges': The index provided (${index}) is greater than or equal to the maximum bound (${maxIndex}).`);
   }
 }
 
 function getRange(fnName, valueIndex, ranges, rangeIndex) {
-  if (rangeIndex === undefined) {
```
misteroneill: Yeah, at first I thought the `rangeCheck` would throw, but it won't. I'll add that.

PR #3828, opened 2016-12-02
Remove deprecation for a Button created with a non-button element

misteroneill:
Remove deprecation for a Button created with a non-button element - throw instead.

## Requirements Checklist

### DIFF
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -29,23 +29,13 @@ class Button extends ClickableComponent {
    *         The element that gets created.
    */
   createEl(tag = 'button', props = {}, attributes = {}) {
+    if (tag !== 'button') {
+      throw new Error('Buttons must be created with a "button" tag.');
+    }
+
     props = assign({
       className: this.buildCSSClass()
     }, props);
-
-    if (tag !== 'button') {
-      log.warn(`Creating a Button with an HTML element of ${tag} is deprecated; use ClickableComponent instead.`);
-
-      // Add properties for clickable element which is not a native HTML button
-      props = assign({
-        tabIndex: 0
-      }, props);
-
-      // Add ARIA attributes for clickable element which is not a native HTML button
-      attributes = assign({
-        role: 'button'
-      }, attributes);
-    }
 
     // Add attributes for button element
     attributes = assign({
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -16,8 +16,9 @@ class Button extends ClickableComponent {
   /**
    * Create the `Button`s DOM element.
    *
-   * @param {string} [tag=button]
-   *        Element's node type. e.g. 'button'
+   * @param {string} [tag="button"]
+   *        The element's node type. This argument is IGNORED: no matter what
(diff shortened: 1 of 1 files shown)


src/js/button.js
```diff
@@ -26,24 +26,14 @@ class Button extends ClickableComponent {
    * @method createEl
    */
   createEl(tag = 'button', props = {}, attributes = {}) {
+    if (tag !== 'button') {
+      throw new Error('Buttons must be created with a "button" tag.');
```
BrandonOCasey: why do we even allow a tag to be passed in?

BrandonOCasey: https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-display.js#L165
https://github.com/videojs/video.js/blob/master/src/js/control-bar/live-display.js#L38 
https://github.com/videojs/video.js/blob/master/src/js/control-bar/control-bar.js#L39
https://github.com/videojs/video.js/blob/master/src/js/loading-spinner.js#L19 
https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L114
https://github.com/videojs/video.js/blob/master/src/js/popup/popup-button.js#L62
https://github.com/videojs/video.js/blob/master/src/js/player.js#L492

Most createEl calls that we have don't seem to follow this format (other than base classes) like clickable-component and component basically.

gkatsev: I think the reason for this was originally backwards compatibility. I wonder if we can now just ignore the passed in tag and and always create a `button`. We'd want to still log the warning, though. So, maybe just throwing is fine :)

@OwenEdwards thoughts?

misteroneill: Well, we definitely want to accept `props` and `attributes`, I think, for custom buttons. But I suppose changing the signature isn't the worst thing in the world.

BrandonOCasey: making it would be good to not throw? Doesn't seem like a fatal error to me, but I could be persuaded either way.

misteroneill: Yeah. I could see just calling `log.error` in this case.

OwenEdwards: There were still a couple of components which are clickable <div>s instead of <button>s - one of them was the poster image (I don't remember the other). I'd been discouraged from making DOM changes, so I made this split for backward-compatibility. Ultimately (in v6.0?) the `clickable-component` and the `button` components should be merged again, and the option to pass a tag removed.

BrandonOCasey: #3837

src/js/button.js
```diff
@@ -28,25 +28,11 @@ class Button extends ClickableComponent {
    * @return {Element}
    *         The element that gets created.
    */
-  createEl(tag = 'button', props = {}, attributes = {}) {
+  createEl(props = {}, attributes = {}) {
```
gkatsev: After some thought, I'm unsure whether we want to change the method signature altogether. It totally makes sense to me to ignore it and always create a button element but it feels weird for this method to have a different signature from all other `createEl`s.

PR #3830, opened 2016-12-02
Add support for `addtrack` on edge even though it doesn't support `removetrack`

BrandonOCasey:
## Description
* Don't change `featuersNativeTextTracks` as it technically does not have the full feature
* Add a check for `browser.IS_EDGE` and proxyNativeTracks if we are on edge.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -119,7 +119,10 @@ class Html5 extends Tech {
       this.on('loadstart', this[`removeOld${capitalType}Tracks_`]);
     });
 
-    if (this.featuresNativeTextTracks) {
+    //  edge only supports the `addtrack` event, so it
+    //  does not technically have native support, but we want to use the support
+    // that it does have.
+    if (this.featuresNativeTextTracks || browser.IS_EDGE) {
       if (crossoriginTracks) {
         log.warn(tsml`Text Tracks are being loaded from another origin but the crossorigin attribute isn't used.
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -119,7 +119,10 @@ class Html5 extends Tech {
       this.on('loadstart', this[`removeOld${capitalType}Tracks_`]);
     });
 
-    if (this.featuresNativeTextTracks) {
+    //  edge only supports the `addtrack` event, so it
+    //  does not technically have native support, but we want to use the support
+    // that it does have.
+    if (this.featuresNativeTextTracks || browser.IS_EDGE) {
```
forbesjo: Wont this conflict with the option `html5.nativeCaptions = false` on Edge?

gkatsev: The idea is that we proxy the native tracks always .I think what's missing is that in the `addtrack` handler we want to make sure that in the case of `this.featuresNativeTextTracks === false`, we only add the track for in-band tracks.

misteroneill:
Is this something we still want to do in light of our discoveries that Edge has terrible native caption support?

gkatsev:
I don't think we want this unless we run into issues with emulated captions or contrib-hls on edge.

misteroneill:
Ok, let's close this for now, in that case.

PR #3833, opened 2016-12-05
Fix error in Fn.throttle that broke MouseTimeDisplay

misteroneill:
## Description
This fixes a breakage in 5.14.0 of the MouseTimeDisplay UI that was caused by my work in 761b877626e9f056464f3ef12923d0bb768c276b

## Specific Changes proposed
- Make the `Fn.throttle` function actually work.
- Add a test to verify and verify manually (which I did).

## Requirements Checklist

### DIFF
--- a/src/js/utils/fn.js
+++ b/src/js/utils/fn.js
@@ -55,7 +55,7 @@ export const bind = function(context, fn, uid) {
  * @return {Function}
  */
 export const throttle = function(fn, wait) {
-  let last;
+  let last = Date.now();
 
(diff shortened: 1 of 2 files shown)


src/js/utils/fn.js
```diff
@@ -55,7 +55,7 @@ export const bind = function(context, fn, uid) {
  * @return {Function}
  */
 export const throttle = function(fn, wait) {
-  let last;
+  let last = Date.now();
```
gkatsev: welp

PR #3834, opened 2016-12-05
tests: add Edge to browserstack tests

gkatsev:
## Description
Add Edge as a browser to run tests on in browserstack.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -144,7 +144,6 @@ function getCustomLaunchers(){
     edge_bs: {
       base: 'BrowserStack',
       browser: 'edge',
-      browser_version: '14',
       os: 'Windows',
       os_version: '10'
     },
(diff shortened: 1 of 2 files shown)


test/karma.conf.js
```diff
@@ -140,6 +141,14 @@ function getCustomLaunchers(){
       os_version: 'Yosemite'
     },
 
+    edge_bs: {
+      base: 'BrowserStack',
+      browser: 'edge',
+      browser_version: '14',
```
forbesjo: You might want to leave browser version off so Browserstack can automatically use the latest Edge.

forbesjo:
One comment otherwise LGTM

test/karma.conf.js
gkatsev: Makes sense to me. Didn't realize that's how the other evergreens were set up.

PR #3829, opened 2016-12-02
Restore all outlines

misteroneill:
This restores outlines for everything for accessibility purposes.

The progress bar looks awkward, but I will be working on that in another PR.

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] Reviewed by Two Core Contributors

PR #3836, opened 2016-12-05
Time Tooltips

misteroneill:
This eliminates the `keepTooltipsInside` option and makes that the default behavior.

It will also involve some refactoring to optimize performance since this is somewhere that we spend a fair amount of time.

## Requirements Checklist


gkatsev:
We should also fix https://github.com/videojs/video.js/pull/3658 as part of this PR.

gkatsev:
Or rather, fix https://github.com/videojs/video.js/issues/3645

src/css/_utilities.scss
```diff
@@ -24,7 +24,6 @@
   border: 0;
   clip: rect(0 0 0 0);
   height: 1px;
-  margin: -1px;
```
misteroneill: I'll copy/paste the commit description here:

> This looks like it was copy/pasted from HTML5 Boilerplate, but there
does not seem to be a purpose to the `margin: -1px;` line and it
causes unsightly "bumps" in the outline of certain elements (sliders).
>
> I browsed through 5 years of history of HTML5 Boilerplate and gave up
looking for a justification for this.
>
> The WebAIM link they even used to cite (and the snook.ca link they
currently cite) does not mention having this margin:
>
> http://webaim.org/techniques/css/invisiblecontent/

misteroneill:
I rebased out the changes from #3829 - I don't think they were truly necessary here.

misteroneill:
This isn't perfect, but I think it's getting close. There are a few very minor edge cases to iron out and the tests need fixing.

gkatsev:
This looks pretty nice.
I noticed that there's still weird behavior related to #3645, particularly if you mouse over from the right side into the progress control. I can show you some time. It's rather minor and an improvement from what we currently have.

gkatsev:
Ok, I think we want to merge this for the RC without addressing the issue but we should address the issue shortly after.

PR #3838, opened 2016-12-06
Deprecate the use of `starttime` in player.js

### LINKED ISSUE #3781: does the starttime/startTime option work? Do we want to keep it?
## Description
Right now I see an option for start time referenced in a few places under two different names. I am thinking that it is a legacy option that probably does not work anymore. Do we want to make it work again, or fix the option?

Code Links: 
* [starttime](https://github.com/videojs/video.js/blob/master/src/js/player.js#L1023)
* [startTime](https://github.com/videojs/video.js/blob/master/src/js/player.js#L657)?

BrandonOCasey:
## Description
Deprecate the use of `starttime` in player.js. It seems like the `startTime` option is actually used in flash so for the time being that will have to stay. Fixes #3781 

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1271,12 +1271,14 @@ class Player extends Component {
    *
    * @fires Player#firstplay
    * @listens Tech#firstplay
+   * @deprecated In 6.0 passing the `starttime` option to the player will be deprecated
    * @private
    */
   handleTechFirstPlay_() {
     // If the first starttime attribute is specified
     // then we will start at the given offset in seconds
     if (this.options_.starttime) {
+      log.warn('Passing the `starttime` option to the player will be deprecated in 6.0');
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -1271,12 +1271,14 @@ class Player extends Component {
    *
    * @fires Player#firstplay
    * @listens Tech#firstplay
+   * @deprecated In 6.0 passing the `starttime` option to the player will be deprecated
```
misteroneill: Rather than "In 6.0" I wonder if we can phrase this as "As of 6.0"?

BrandonOCasey: seems like a good idea

gkatsev:
I think this one can be merged in as a minor.

PR #3847, opened 2016-12-09
Text track button does not hide after remove tracks

mctep:
Hello! I write own tech that supports text tracks. How to clear text tracks, after load next source in player?

I try to remove **text** (upd) tracks like below:

```js
const tracks = player.textTracks();
tracks.tracks_.forEach(track => tracks.removeTrack_(track));
```

But `TextTrackButton` push OFF `item` for disabling any text tracks and return items with single element.

For the first displaying component it is work because in constructor of `TrackButton` it hides when length of items less than 2:

```js
if (this.items.length <= 1) {
  this.hide();
}
```

But when `TextTracks` updates in `MenuButton` it hides only when length of items is 0:

```js
if (this.items && this.items.length === 0) {
  this.hide();
} else if (this.items && this.items.length > 1) {
  this.show();
}
```

So I think it would better that `TextTrackButton` should return empty array of items if there are no tracks in player. This PR is just example solution.

### DIFF
--- a/src/js/control-bar/text-track-controls/text-track-button.js
+++ b/src/js/control-bar/text-track-controls/text-track-button.js
@@ -38,14 +38,14 @@ class TextTrackButton extends TrackButton {
    *         Array of menu items that were created
    */
   createItems(items = []) {
-    // Add an OFF menu item to turn all tracks off
-    items.push(new OffTextTrackMenuItem(this.player_, {kind: this.kind_}));
-
     const tracks = this.player_.textTracks();
 
-    if (!tracks) {
+    if (!tracks || !tracks.length) {
       return items;
     }
+
+    // Add an OFF menu item to turn all tracks off
+    items.push(new OffTextTrackMenuItem(this.player_, {kind: this.kind_}));
 
     for (let i = 0; i < tracks.length; i++) {
       const track = tracks[i];


gkatsev:
Your way of removing tracks isn't our recommended method for this. What about doing something like the following instead?
```js
// only remote tracks are actually removable
const tracks = player.remoteTextTracks();

let i = 0;
while (i--) {
  player.removeRemoteTextTrack(tracks[i]);
}
```

mctep:
Thank you. I will use only remote text tracks.
However it does not fix problem with hiding `TextTrackButton` after removing all text tracks.

gkatsev:
Trying it out locally, the button is removed for me. Do you have a [reduced test case](https://css-tricks.com/reduced-test-cases/)?

misteroneill:
I have confirmed that removing a remote text track removes the CC button when using video.js 5.12.

Here is a [reduced test case](http://output.jsbin.com/tisube).

gkatsev:
Thanks for confirming @misteroneill. Based on that, I don't think this PR is necessary. If you're seeing an issue with the button being removed, could you open an issue with a reduced test case?

Thanks.

mctep:
Hm. Thank you. I will look into it more.

mctep:
Oh. This issue actual for subtitles tracks and `SubtitlesButton`. Not for captions. There is threshold In `CaptionsButton`.

Here is modified your [reduced test case](http://jsbin.com/cepiyisezo)

gkatsev:
Hm... Sounds like we want to move the threshold piece of code from captionsButton to the TextTrackButton since I think that descriptions tracks would have the same issue now as well.

gkatsev:
Can you open an issue with this? If you want to attemp that solution, that would be awesome as well. Thanks for bringing this to our attention @mctep.

PR #3853, opened 2016-12-12
Log Levels

misteroneill:
## Description
This adds support for log levels to video.js' logger. Log levels act as a whitelist of logger method(s) that are allowed. The following levels are defined here:

* `off`: Matches none. Any value that can be cast to `false` will have this effect. The most restrictive.
* `all` (default): Matches all Video.js-provided functions (`log`, `log.warn`, and `log.error`).
* `warn`: Matches `log.warn` and `log.error` calls.
* `error`: Matches only `log.error` calls.

Also, this improves the `log.history` handling with the ability to `clear()` it as well as toggle it off and on using `disable()` and `enable()`.

## Specific Changes proposed
- Adds the `log.level()` method which is a getter/setter for log levels.
- Adds the `log.levels` object as a container for the various logging levels.
- **BREAKING CHANGE:** `log.history` is now a function that returns a shallow clone of the internal history array rather than the array itself.
- Adds the `log.history.clear`, `log.history.disable`, `log.history.enable` methods.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -170,10 +170,10 @@ help but are outside of the scope of video.js.
 For content that must be highly secure [videojs-contrib-eme][eme] adds DRM support.
 
 ## Q: Can I turn off video.js logging?
-Yes! This can be achieved by adding the following code _before_ including video.js:
+Yes! This can be achieved by adding the following code _after_ including Video.js, but _before_ creating any player(s):
 
 ```js
-window.VIDEOJS_DEFAULT_LOG_LEVEL = 'off';
+videojs.log.level('off');
 ```
 
 ## Q: What is a plugin?
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -169,6 +169,13 @@ help but are outside of the scope of video.js.
 
 For content that must be highly secure [videojs-contrib-eme][eme] adds DRM support.
 
+## Q: Can I turn off video.js logging?
+Yes! This can be achieved by adding the following code _before_ including video.js:
+
+```js
+window.VIDEOJS_DEFAULT_LOG_LEVEL = 'off';
+```
+
 ## Q: What is a plugin?
 
 A plugin is a group of reusable functionality that can be re-used by others. For instance a plugin could add
@@ -197,6 +204,15 @@ the [plugins guide][plugin-guide] for more information.
 ## Q: Where can I find a list of video.js skins?
 
 See the [video.js github wiki][skins-list].
+
+## Q: Can I disable the `<style>` elements video.js adds by default?
+Yes! You can add the following code _before_ including video.js:
+
+```js
+window.VIDEOJS_NO_DYNAMIC_STYLE = true;
+```
+
+For more information, see the [skins guide][skin-style-els].
 
 ## Q: Does video.js work as an audio only player?
 
@@ -307,3 +323,5 @@ Yes! Please [submit an issue or open a pull request][pr-issue-question] if this 
 [semver]: http://semver.org/
 
 [starter-example]: http://jsbin.com/axedog/edit?html,output
+
+[skin-style-els]: /docs/guides/skins.md#additional-style-elements
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -205,15 +205,6 @@ the [plugins guide][plugin-guide] for more information.
 
 See the [video.js github wiki][skins-list].
 
-## Q: Can I disable the `<style>` elements video.js adds by default?
-Yes! You can add the following code _before_ including video.js:
-
-```js
-window.VIDEOJS_NO_DYNAMIC_STYLE = true;
-```
-
-For more information, see the [skins guide][skin-style-els].
-
 ## Q: Does video.js work as an audio only player?
 
 Yes! It can be used to play audio only files in a `<video>` or `<audio>` tag. The
@@ -323,5 +314,3 @@ Yes! Please [submit an issue or open a pull request][pr-issue-question] if this 
 [semver]: http://semver.org/
 
 [starter-example]: http://jsbin.com/axedog/edit?html,output
-
-[skin-style-els]: /docs/guides/skins.md#additional-style-elements
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -19,6 +19,7 @@
   * [Q: How can I autoplay a video on a mobile device?](#q-how-can-i-autoplay-a-video-on-a-mobile-device)
 * [Q: How can I play RTMP video in video.js?](#q-how-can-i-play-rtmp-video-in-videojs)
 * [Q: How can I hide the links to my video/subtitles/audio/tracks?](#q-how-can-i-hide-the-links-to-my-videosubtitlesaudiotracks)
+* [Q: Can I turn off video.js logging?](#q-can-i-turn-off-videojs-logging)
 * [Q: What is a plugin?](#q-what-is-a-plugin)
 * [Q: How do I make a plugin for video.js?](#q-how-do-i-make-a-plugin-for-videojs)
 * [Q: Where can I find a list of video.js plugins?](#q-where-can-i-find-a-list-of-videojs-plugins)
@@ -170,11 +171,14 @@ help but are outside of the scope of video.js.
 For content that must be highly secure [videojs-contrib-eme][eme] adds DRM support.
 
 ## Q: Can I turn off video.js logging?
+
 Yes! This can be achieved by adding the following code _after_ including Video.js, but _before_ creating any player(s):
 
 ```js
 videojs.log.level('off');
 ```
+
+For more information, including which logging levels are available, check out the [debugging guide][debug-guide].
 
 ## Q: What is a plugin?
 
@@ -314,3 +318,5 @@ Yes! Please [submit an issue or open a pull request][pr-issue-question] if this 
 [semver]: http://semver.org/
 
 [starter-example]: http://jsbin.com/axedog/edit?html,output
+
+[debug-guide]: ./guides/debug.md
--- a/docs/guides/debugging.md
+++ b/docs/guides/debugging.md
@@ -1,0 +1,111 @@
+# Debugging
+
+## Table of Contents
+
+* [Logging](#logging)
+  * [API Overview](#api-overview)
+  * [Log Safely](#log-safely)
+  * [Log Objects Usefully](#log-objects-usefully)
+  * [Log Levels](#log-levels)
+  * [History](#history)
+
+## Logging
+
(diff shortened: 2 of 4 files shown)


src/js/utils/log.js
```diff
@@ -22,17 +25,19 @@ let log;
...
 
-    // add the type to the front of the message when it's not "log"
+    // Add the type to the front of the message when it's not "log".
     args.unshift(type.toUpperCase() + ':');
   }
 
-  // add to history
-  log.history.push(args);
+  // Add a clone of the args at this point to history.
+  log.history.push([].concat(args));
```
misteroneill: Changed this to a clone because I think the intent was for `VIDEOJS:` to _not_ get added to history, but it didn't work that way.

src/js/utils/log.js
```diff
@@ -8,6 +8,9 @@ import {isObject} from './obj';
 
 let log;
 
+// This is the private tracking variable for logging level.
+let level = 'debug';
```
misteroneill: I guess I'm not 100% sure on this one either. I wonder if it makes more sense to turn logging off and make it opt-in for 6.0?

gkatsev: Maybe add an `info` and default it to that? `info` would be just `videojs.log`.

src/js/utils/log.js
```diff
@@ -93,6 +99,76 @@ log = function(...args) {
...
+ * - `debug` (default): Matches only Video.js-provided functions (`log`,
+ *   `log.warn`, and `log.error`).
+ * - `warn`: Matches `log.warn` and `log.error` calls.
+ * - `error`: Matches only `log.error` calls.
+ *
+ * @type {Object}
+ */
+log.levels = {};
+
+Object.defineProperties(log.levels, {
```
gkatsev: this will break in IE8.

misteroneill: D'oh. I guess they can be normal properties.

src/js/utils/log.js
misteroneill: So it would show `log` calls but not `warn` or `error`?

gkatsev: Yup.

src/js/utils/log.js
misteroneill: Fair point. Well, we can also simply have it use normal properties for browsers that don't support `Object.defineProperties`.

src/js/utils/log.js
misteroneill: That's not something I would've ever considered to be a use-case, honestly.

misteroneill:
I updated JSDoc, but it's not perfectly correct because this structure (function with functions as properties) doesn't really fit within their OO worldview.

src/js/utils/events.js
```diff
@@ -1,13 +1,15 @@
...
  * @module events
- *
- * Event System (John Resig - Secrets of a JS Ninja http://jsninja.com/)
- * (Original book version wasn't completely usable, so fixed some things and made Closure Compiler compatible)
- * This should work very similarly to jQuery's events, however it's based off the book version which isn't as
- * robust as jquery's, so there's probably some differences.
  */
 
+/*
+Event System (John Resig - Secrets of a JS Ninja http://jsninja.com/)
```
BrandonOCasey: this should probably be reverted, since it is fixed in #3854

src/js/utils/log.js
```diff
@@ -76,32 +83,124 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
...
+} else {
+  log.levels.DEFAULT = level;
+}
+
+/**
+ * Get or set the current logging level. If a string matching a key from
+ * {@link log.levels} is provided, acts as a setter. Regardless of argument,
+ * returns the current logging level.
+ *
+ * @param  {String} [lvl]
```
BrandonOCasey: should be lowercase `{string}` rather than uppercase

src/js/utils/log.js
```diff
@@ -8,9 +8,13 @@ import {isObject} from './obj';
 
 let log;
 
+// This is the private tracking variable for logging level.
+let level = 'all';
```
BrandonOCasey: It may be better to set the default log level in the same location  as the check for `window.VIDEOJS_DEFAULT_LOG_LEVEL`

src/js/utils/log.js
```diff
@@ -76,32 +83,124 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
...
+ *
+ * @type {Object}
+ */
+log.levels = {};
+
+if (Object.defineProperties) {
+  Object.defineProperties(log.levels, {
+    all: {
+      enumerable: true,
+      value: 'log|warn|error'
```
BrandonOCasey: typical log level implementations that I have seen use number rather rather than regex matches to see if they should log. I think that this works fine here and it does allow us to only log from one level if we want to, but I figured that I would throw in another idea.

Here are the log levels I see:
* 0 - Off
* 1 - Error
* 2 - Warn
* 3 - log

Typically you set your log level to the number or string representation of a level, so 3 or `log` by default. Then everything below that level is logged. So if the log level is 3 we log everything, if the log level is 2 we log only warnings and errors, if the log level is 1 we only log errors, and finally if the log level is 0 we log nothing at all.

misteroneill: That was my original implementation, but the problem with numerically-indexed levels like that is that they are really inflexible and difficult to expand. For example, @gkatsev had the suggestion of a log level that was `log` _only_ (not `warn` or `error`). That doesn't fit neatly in this sort of sequential system.

BrandonOCasey: ahh ok, just wanted to make sure the more traditional system was considered, but I do see the benefits here.

docs/guides/faq.md
```diff
@@ -74,6 +74,13 @@ help but are outside of the scope of video.js.
 
 For content that must be highly secure [videojs-contrib-eme][eme] adds DRM support.
 
+## Q: Can I turn off video.js logging?
+Yes! This can be achieved by adding the following code _before_ including video.js:
+
+```js
+window.VIDEOJS_DEFAULT_LOG_LEVEL = 'off';
```
gkatsev: can we make this be settable per-player also, not just globally?

misteroneill:
Yeah, it could be a minor. I'll make the other change.

docs/guides/faq.md
misteroneill: I think the only way to do this would be to really refactor the logging to create new console wrappers for each player (alongside the one for video.js itself). Probably not worth doing for a minor.

misteroneill:
As discussed offline, we are going to refactor this somewhat to allow for per-player logging and log-levels, which means it will become a major update and roll into 6.0.

docs/faq.md
```diff
@@ -153,6 +160,15 @@ the [plugins guide][plugin-guide] for more information.
 
 See the [video.js github wiki][skins-list].
 
+## Q: Can I disable the `<style>` elements video.js adds by default?
```
BrandonOCasey: Should this be in another PR?

BrandonOCasey: Should this be in another PR?

src/js/utils/log.js
```diff
@@ -76,32 +83,127 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
...
+      enumerable: true,
+      value: ''
+    },
+    warn: {
+      enumerable: true,
+      value: 'warn|error'
+    }
+  });
+} else {
+  log.levels = {
```
BrandonOCasey: what about making this the default and overriding it if `Object.defineProperties` exists? We could also loop through the original log levels to make this code smaller.

src/js/utils/log.js
```diff
@@ -76,32 +83,127 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
...
+    off: '',
+    warn: 'warn|error'
+  };
+}
+
+// Determine the default logging level.
+level = 'all';
+
+if (window.VIDEOJS_DEFAULT_LOG_LEVEL &&
+    log.levels.hasOwnProperty(window.VIDEOJS_DEFAULT_LOG_LEVEL)) {
```
BrandonOCasey: do we want to log a warning if the default log level is invalid?

src/js/utils/log.js
```diff
@@ -76,32 +83,127 @@ export const logByType = (type, args, stringify = !!IE_VERSION && IE_VERSION < 1
...
+  level = window.VIDEOJS_DEFAULT_LOG_LEVEL;
+}
+
+if (Object.defineProperty) {
+  Object.defineProperty(log.levels, 'DEFAULT', {value: level});
```
BrandonOCasey: why not make `log.levels.DEFAULT = level;` the default and override it if `Object.defineProperty` exists?

docs/faq.md
misteroneill: Absolutely. No idea how it got in here... weird.

src/js/utils/log.js
misteroneill: Eliminating the `defineProperties` usage altogether.

src/js/utils/log.js
misteroneill: Killing off this global entirely.

src/js/utils/log.js
misteroneill: Doing something simple like you're suggesting, but getting rid of the `defineProperty` crap altogether.

src/js/utils/log.js
misteroneill: No longer relevant as `window.VIDEOJS_DEFAULT_LOG_LEVEL` was removed.

docs/faq.md
```diff
@@ -169,6 +169,13 @@ help but are outside of the scope of video.js.
 
 For content that must be highly secure [videojs-contrib-eme][eme] adds DRM support.
 
+## Q: Can I turn off video.js logging?
```
gkatsev: Should we list valid log levels here?

misteroneill: Added debugging guide, which includes that information. And added a link to that guide here.

docs/guides/debugging.md
```diff
@@ -0,0 +1,111 @@
...
+By passing a string, the logging level can be changed to one of the available logging levels:
+
+```js
+videojs.log.level('error'); // show only error messages and suppress others
+videojs.log('foo'); // does nothing
+videojs.log.warn('foo'); // does nothing
+videojs.log.error('foo'); // logs "foo" as an error
+```
+
+Available logging levels are:
```
gkatsev: Can we make this a `####` heading? That way it shows up in the TOC as well.

misteroneill: Done.

PR #3862, opened 2016-12-15
test: move vjs hooks test module to separate file

gkatsev:


gkatsev:
Yes. Yesterday when I was trying to add a test to videojs, I added it to the bottom of the file and then went to filter to the "video.js" qunit module and couldn't find my test. It's best to leave each qunit module in a separate file.

gkatsev:
According to `ack`, it's the only one.

PR #3888, opened 2016-12-23
Fix extra warn logs on already initialized player references

gesinger:
## Description
Extra warnings were being logged when getting a reference to a player via videojs('player_id').

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/test/unit/video.test.js
+++ b/test/unit/video.test.js
@@ -61,25 +61,27 @@ function(assert) {
   const player = videojs('test_vid_id', { techOrder: ['techFaker'] });
 
   assert.ok(player, 'created player from tag');
-  assert.ok(player.id() === 'test_vid_id');
-  assert.ok(!warnLogs.length, 'no warn logs');
+  assert.equal(player.id(), 'test_vid_id', 'player has the right ID');
+  assert.equal(warnLogs.length, 0, 'no warn logs');
 
   const playerAgain = videojs('test_vid_id');
 
-  assert.ok(player === playerAgain, 'did not create a second player from same tag');
-  assert.ok(!warnLogs.length, 'no warn logs');
+  assert.equal(player, playerAgain, 'did not create a second player from same tag');
+  assert.equal(warnLogs.length, 0, 'no warn logs');
 
   const playerAgainWithOptions = videojs('test_vid_id', { techOrder: ['techFaker'] });
 
-  assert.ok(player === playerAgainWithOptions,
-            'did not create a second player from same tag');
+  assert.equal(player,
+               playerAgainWithOptions,
+               'did not create a second player from same tag');
   assert.equal(warnLogs.length, 1, 'logged a warning');
   assert.equal(warnLogs[0],
-               'Player "test_vid_id" is already initialised. ' +
-                 'Options will not be applied.',
+               'Player "test_vid_id" is already initialised. Options will not be applied.',
                'logged the right message');
 
   log.warn = origWarnLog;
+
+  player.dispose();
 });
 
 QUnit.test('should return a video player instance from el html5 tech', function(assert) {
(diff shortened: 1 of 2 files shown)


test/unit/video.test.js
```diff
@@ -45,6 +46,42 @@ QUnit.test('should return a video player instance', function(assert) {
...
+  assert.ok(player === playerAgainWithOptions,
+            'did not create a second player from same tag');
+  assert.equal(warnLogs.length, 1, 'logged a warning');
+  assert.equal(warnLogs[0],
+               'Player "test_vid_id" is already initialised. ' +
+                 'Options will not be applied.',
+               'logged the right message');
+
+  log.warn = origWarnLog;
+});
```
gkatsev: `player` needs to be disposed.

test/unit/video.test.js
```diff
@@ -45,6 +46,42 @@ QUnit.test('should return a video player instance', function(assert) {
...
+
+  const player = videojs('test_vid_id', { techOrder: ['techFaker'] });
+
+  assert.ok(player, 'created player from tag');
+  assert.ok(player.id() === 'test_vid_id');
+  assert.ok(!warnLogs.length, 'no warn logs');
+
+  const playerAgain = videojs('test_vid_id');
+
+  assert.ok(player === playerAgain, 'did not create a second player from same tag');
```
gkatsev: `assert.equal` would be preferred over `ok` and `===`

test/unit/video.test.js
```diff
@@ -45,6 +46,42 @@ QUnit.test('should return a video player instance', function(assert) {
...
+  assert.ok(!warnLogs.length, 'no warn logs');
+
+  const playerAgainWithOptions = videojs('test_vid_id', { techOrder: ['techFaker'] });
+
+  assert.ok(player === playerAgainWithOptions,
```
gkatsev: again `assert.equal`.

test/unit/video.test.js
```diff
@@ -45,6 +46,42 @@ QUnit.test('should return a video player instance', function(assert) {
...
+            'did not create a second player from same tag');
+  assert.equal(warnLogs.length, 1, 'logged a warning');
+  assert.equal(warnLogs[0],
+               'Player "test_vid_id" is already initialised. ' +
```
gkatsev: you can keep this string on one line (for now)

test/unit/video.test.js
```diff
@@ -45,6 +46,42 @@ QUnit.test('should return a video player instance', function(assert) {
...
+    warnLogs.push(args);
+  };
+
+  fixture.innerHTML += '<video id="test_vid_id"></video>';
+
+  const player = videojs('test_vid_id', { techOrder: ['techFaker'] });
+
+  assert.ok(player, 'created player from tag');
+  assert.ok(player.id() === 'test_vid_id');
+  assert.ok(!warnLogs.length, 'no warn logs');
```
gkatsev: maybe better to do an `equals` check with with `length` and `0`?

test/unit/video.test.js
```diff
@@ -45,6 +46,42 @@ QUnit.test('should return a video player instance', function(assert) {
...
+  log.warn = (args) => {
+    warnLogs.push(args);
+  };
+
+  fixture.innerHTML += '<video id="test_vid_id"></video>';
+
+  const player = videojs('test_vid_id', { techOrder: ['techFaker'] });
+
+  assert.ok(player, 'created player from tag');
+  assert.ok(player.id() === 'test_vid_id');
```
gkatsev: `assert.equal` instead and don't forget a message.

gesinger:
Updated. Thanks for the review @gkatsev !

test/unit/video.test.js
```diff
@@ -45,6 +46,44 @@ QUnit.test('should return a video player instance', function(assert) {
   player2.dispose();
 });
 
+QUnit.test('should log about already initalized players if options already passed',
```
misteroneill: I think we'd ultimately want this test to use `sinon.spy(log, 'warn')` instead and to use the `TestHelpers.makePlayer()` function to create a mock player, but I don't think it should stop the PR from going in since there is time-sensitivity here.

PR #3889, opened 2016-12-23
fix: support require()-ing video.js

### LINKED ISSUE #3869: Allow video.js to be run in node
## Description
I am server rendering my frontend using node.  We are unable to server render any modules that import video.js.  Due to running `document.createElement`

https://github.com/videojs/video.js/blob/881cfcb346b588013768f538faf3d5f20ce42109/src/js/utils/browser.js#L74

## Steps to reproduce

1. run node dist/video.js
2. See stacktrace `TypeError: _document2.default.createElement is not a function`

### Expected
Video.js to not crash

### Actual
It crashes running  `document.createElement`

### Error output
`TypeError: _document2.default.createElement is not a function`

## Additional Information
#### videojs
5.12.6
#### browsers
what browser are affected? None
#### OSes
what platforms (operating systems and devices) are affected? node any

cc @ljharb

misteroneill:
## Description
This makes a number of small fixes to support pulling in video.js via `require()` in Node.

I have verified it in the following ways (using Node 4.6):

```sh
node -e "require('./dist/video.js')"
```

```sh
node -e "require('./es5/video.js')"
```

```
node
> require('./dist/video.js')
```

```
node
> require('./es5/video.js')
```

All these were tested because we had noticed some odd behavior when using the inline `-e` command vs. the REPL. Success was considered "didn't throw an error".

> **Note**: It is not expected that _calling_ `videojs()` the function in Node will work. That will throw an error, but loading it as a module should not throw.

Fixes #3869.

## Specific Changes proposed
- Introduce the `Dom.isReal()` function, which makes an educated assumption about the "realness" of the `document` object.
- Wrap code here and there with checks against `Dom.isReal()` as well as other defensive code.

## Requirements Checklist

### DIFF
--- a/src/js/setup.js
+++ b/src/js/setup.js
@@ -4,6 +4,7 @@
  *
  * @module setup
  */
+import * as Dom from './utils/dom';
 import * as Events from './utils/events.js';
 import document from 'global/document';
 import window from 'global/window';
@@ -15,6 +16,12 @@ let videojs;
  * Set up any tags that have a data-setup `attribute` when the player is started.
  */
 const autoSetup = function() {
+
+  // Protect against breakage in non-browser environments.
+  if (!Dom.isReal()) {
+    return;
+  }
+
   // One day, when we stop supporting IE8, go back to this, but in the meantime...*hack hack hack*
   // var vids = Array.prototype.slice.call(document.getElementsByTagName('video'));
   // var audios = Array.prototype.slice.call(document.getElementsByTagName('audio'));
@@ -89,10 +96,10 @@ function autoSetupTimeout(wait, vjs) {
     videojs = vjs;
   }
 
-  setTimeout(autoSetup, wait);
+  window.setTimeout(autoSetup, wait);
 }
 
-if (document.readyState === 'complete') {
+if (Dom.isReal() && document.readyState === 'complete') {
   _windowLoaded = true;
 } else {
   /**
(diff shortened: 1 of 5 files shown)


src/js/setup.js
```diff
@@ -89,10 +96,10 @@ function autoSetupTimeout(wait, vjs) {
     videojs = vjs;
   }
 
-  setTimeout(autoSetup, wait);
+  window.setTimeout(autoSetup, wait);
```
ljharb: `setTimeout` is a global in node too, so this change isn't really needed

src/js/utils/dom.js
```diff
@@ -56,6 +56,15 @@ function classRegExp(className) {
 }
 
 /**
+ * Whether the current DOM interface appears to be real.
+ *
+ * @return {Boolean}
+ */
+export function isReal() {
+  return document === window.document && typeof document.createElement === 'function';
```
ljharb: this would need a `typeof window !== 'undefined'` and `typeof document !== 'undefined'`, since otherwise they're a reference error in node.

gkatsev: since we're using `global`, it returns either window or global depending on the environment, so, `window` should never end up being undefined.

src/js/setup.js
misteroneill: True, but it's consistent with the rest of the codebase where we are using `window.setTimeout`, which will work in Node because we're using the `global` package.

PR #3896, opened 2016-12-28
chore: better dev experience

gkatsev:
npm start now runs `grunt dev` which in turn runs the connect server,
sets up babel watching, browserify watching (for both src and test
files), and css and alternate builds watching.

Additionally, it will copy over the dev build files into the dist folder
so that if you're linking videojs and you have a project that gets
video.js from the dist, you can link and develop with more easy without
rerunning all of `grunt dist` each time.

Additionally, this specifically does not run more time consumming
operations like minification.

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -127,6 +127,15 @@ module.exports = function(grunt) {
     },
     dist: {},
     watch: {
+      dist: {
+        files: [
+          'build/temp/video.js',
+          'build/temp/alt/video.novtt.js',
+          'build/temp/video-js.css',
+          'build/temp/alt/video-js-cdn.css'
+        ],
+        tasks: ['copy:dist']
+      },
       novtt: {
         files: ['build/temp/video.js'],
         tasks: ['concat:novtt']
@@ -142,10 +151,6 @@ module.exports = function(grunt) {
       babel: {
         files: ['src/js/**/*.js'],
         tasks: ['babel:es5']
-      },
-      jshint: {
-        files: ['src/**/*', 'test/unit/**/*.js', 'Gruntfile.js'],
-        tasks: 'jshint'
       }
     },
     connect: {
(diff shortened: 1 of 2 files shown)


build/grunt.js
```diff
@@ -142,10 +151,6 @@ module.exports = function(grunt) {
       babel: {
         files: ['src/js/**/*.js'],
         tasks: ['babel:es5']
-      },
-      jshint: {
```
gkatsev: we no longer use jshint.

build/grunt.js
```diff
@@ -127,6 +127,15 @@ module.exports = function(grunt) {
     },
     dist: {},
     watch: {
+      dist: {
+        files: [
```
gkatsev: these are the files that users most likely try to get from the dist folder, so, we copy on changes to them.

build/grunt.js
```diff
@@ -405,6 +410,14 @@ module.exports = function(grunt) {
         'watch:babel',
         'browserify:tests'
       ],
+      dev: [
```
gkatsev: we specifically don't run `watch:minify` because it is a time consuming process.

build/grunt.js
```diff
@@ -405,6 +410,14 @@ module.exports = function(grunt) {
...
+        'browserify:watch',
+        'browserify:tests',
+        'watch:novtt',
+        'watch:skin',
+        'watch:dist',
+        'shell:babel'
```
gkatsev: we run `shell:babel` to use babel's built-in file watching. `grunt-babel` provides watching using grunt's `watch` module that we use above, however, it recompiles *all* the files instead of the files that were changed.

package.json
```diff
@@ -22,12 +22,13 @@
...
     "lint": "vjsstandard",
-    "start": "grunt watchAll",
+    "start": "grunt dev",
     "test": "grunt test",
     "docs": "npm run docs:lint && npm run docs:api",
     "docs:api": "jsdoc -r src/js -d docs/api -c .jsdoc.json",
     "docs:lint": "remark -- './**/*.md'",
-    "docs:fix": "remark --output -- './**/*.md'"
+    "docs:fix": "remark --output -- './**/*.md'",
+    "babel": "babel src/js -d es5"
```
gkatsev: a simple babel command to compile our source files to the destination folder.

build/grunt.js
```diff
@@ -437,6 +450,12 @@ module.exports = function(grunt) {
       }
     },
     shell: {
+      babel: {
+        command: 'npm run babel -- --w',
```
gkatsev: this runs our npm babel script (a simple babel command) with the `--w` flag for watch. When a file changes, just that one will be compiled.

PR #3860, opened 2016-12-14
Remove method Chaining from videojs

BrandonOCasey:
## Description
* Remove method chaining from video.js
* Returning a Promise from play should happen in a separate PR

## Specific Changes proposed
List of methods that were changed:
* src/js/component.js
  * ready
  * addClass
  * removeClass
  * toggleClass
  * show
  * hide
  * lockShowing
  * unlockShowing
  * setAttribute
  * removeAttribute
  * dimension - (Only when setting)
  * dimensions
  * on
  * off
  * one
  * trigger
  * height - (only when setting)
  * width - (only when setting)
* src/js/modal-dialog.js
  * close
  * open
  * empty
  * fill
  * fillWith
* src/js/slider/slider.js
  * vertical - (only when setting)
* src/js/tech/tech.js
  * withSourceHandlers/setSource
* src/js/clickable-component.js
  * controlText - getter
  * enable
  * disable
* src/js/menu/menu-button.js
  * enable
  * disable
* src/js/player.js
  * pause
  * play
  * scrubbing - (only when setting)
  * currentTime - (only when setting)
  * duration - (only when setting)
  * volume - (only when setting)
  * hasStarted - (only when setting)
  * muted - (only when setting)
  * isFullscreen - (only when setting)	
  * requestFullscreen
  * exitFullscreen
  * src - (only when setting)
  * load
  * reset
  * preload - (only when setting)
  * autoplay - (only when setting)
  * loop - (only when setting)
  * poster - (only when setting)
  * controls - (only when setting)
  * usingNativeControls - (only when setting)
  * playbackRate - (only when setting)
  * isAudio - (only when setting)
  * language - (only when setting)
  * dimension - (only when setting)
  * error - (only when setting)
  * userActive - (only when setting)

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] Change has been verified in an actual browser (Chome, Firefox, IE)
- [x] Unit Tests updated or fixed
- [ ] Reviewed by Two Core Contributors

PR #3907, opened 2017-01-03
Return the native Promise from play()

BrandonOCasey:
## Description
Return a Promise from `play()` originally this was a PR off of #3860 (https://github.com/BrandonOCasey/video.js/pull/4). This has been changed to be standalone.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1570,20 +1570,36 @@ class Player extends Component {
   /**
    * start media playback
    *
-   * @return {Player}
-   *         A reference to the player object this function was called on
+   * @return {Promise|undefined}
+   *         Returns a `Promise` if the browser supports it or a Promise library was
+   *         passed in as an option. This will return the native play `Promise` if play()
+   *         returns a `Promise` for this browser.
    */
   play() {
-    // Only calls the tech's play if we already have a src loaded
-    if (this.src() || this.currentSrc()) {
-      this.techCall_('play');
-    } else {
+    const Promise_ = this.options_.Promise || window.Promise;
+
+    const play_ = (resolve) => {
+      // Only calls the tech's play if we already have a src loaded
+      if (this.src() || this.currentSrc()) {
+        return resolve(this.techGet_('play'));
+      }
+
       this.tech_.one('loadstart', function() {
-        this.play();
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
Still have to work on this comment from gkatsev https://github.com/BrandonOCasey/video.js/pull/4#pullrequestreview-14032894

BrandonOCasey:
#3927

src/js/player.js
```diff
@@ -1570,20 +1570,24 @@ class Player extends Component {
...
    */
   play() {
     // Only calls the tech's play if we already have a src loaded
     if (this.src() || this.currentSrc()) {
-      this.techCall_('play');
-    } else {
-      this.tech_.one('loadstart', function() {
-        this.play();
-      });
+      return this.techGet_('play');
```
gkatsev: oh, this is an interesting case for the techget/techcall dichotomy. `play` is still technically a setter, which is why `techcall` is used, but we want its return value. I wonder if we need to get rid the the difference. This is particularly important because of how middleware work. We're bubbling `techCall`s from the player to the tech and `techGet`s from the tech to the player.

BrandonOCasey: I also think that we should get rid of the difference

src/js/player.js
```diff
@@ -1570,20 +1570,24 @@ class Player extends Component {
...
     }
 
-    return this;
+    this.tech_.one('loadstart', function() {
```
mister-ben: If there's a source but it's of an unknown type, this throws as `this.tech_` is undefined.

BrandonOCasey: should I just wrap the loadstart listener in an `if (this.tech_)` check then?

gkatsev: `this.ready()`

misteroneill:
@mister-ben Could you take another look at this? I think the issue you brought up was addressed.

mister-ben:
Sure, will get a chance to look properly next week.

thecotne:
so on chrome 62 where native videoTag.play() returns promise 

this should return promise right?

i am asking becouse it does not return anything

gkatsev:
Video.js 6 will, yes. Video.js 5 will not.

PR #3908, opened 2017-01-03
chore: don't run tests on travis if only docs were changed

gkatsev:
Check to see if we're on travis and if the changes that we're running against only include files in the `docs/` folder or are `.md` files and if so, don't run any tests. Locally, or, if non-doc changes, tests will still run.

### DIFF
--- a/build/docs-only.js
+++ b/build/docs-only.js
@@ -1,0 +1,19 @@
+import sh from 'shelljs';
+import path from 'path';
+
+
+export default function(commit, commitRange) {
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
would it be better to do this if only src/*, test/*, or package.json changed?

gkatsev:
I'd prefer to black-list stuff rather than whitelist, particularly since the blacklist is smaller.

gkatsev:
Um... it should for as much as `git` commands work on windows.

gkatsev:
Actually, the check is only made on travis, so, should be fine. Should always run tests locally.

PR #3910, opened 2017-01-04
jsdoc output

gkatsev:
Uses the TUI JSDoc template. All configuring of jsdoc was moved to the jsdoc.json file.
There's currently a bug with TUI regarding the tutorials (https://github.com/nhnent/tui.jsdoc-template/issues/12) but it should via gh-pages currently because it'll now autorender markdown into html.

Left to do:
* [x] run jsdoc on publish
* [x] tweak configuration of jsdoc
* [x] make sure that tutorials are rendering correctly
* [x] figure out what to do about TOCs in jsdoc
* [ ] fix jsdoc's README


BrandonOCasey:
Do we want to do custom styling for video.js in another PR?

BrandonOCasey:
Can we change the logo in the top left? (from toastui to videojs)

gkatsev:
Yes, I believe it's all customizable in some way or another. I'll do it in this PR if it isn't too complicated.

BrandonOCasey:
I also noticed that some of the links are broken, I think it would be good to have some sort of automated tests to check for broken links (maybe in another PR as well)

BrandonOCasey:
I also noticed that examples seem to be cut off at the bottom? (might be a bug in tui)
http://gkatsev.github.io/video.js/docs/api/tutorial-components.html

gkatsev:
the tutorials are loaded in in-an iframe, so, you should be able to just scroll :), but yeah, the iframe is potentially not sized correctly.

Also, which links do you see broken?

BrandonOCasey:
the Guides link at http://gkatsev.github.io/video.js/docs/api/index.html leading to http://gkatsev.github.io/video.js/docs/api/guides. Should probably be `/guides`? I guess its not up yet though.

BrandonOCasey:
I also noticed that the API link at the top doesn't just link to the current page.

gkatsev:
Ah, I think that's because it's just inlining the `docs/api.md` file as a "main" and the links are still relative to the markdown structure. That's something that we'd want to fix. We potentially could just write a different page to use as the front of the docs site.

gkatsev:
The latest updates depend on these changes: https://github.com/nhnent/tui.jsdoc-template/pulls/gkatsev

gkatsev:
Looks like to get tocs to work, we'd need to update the template to add [a custom renderer to marked to add embedded anchor tags for headings](https://github.com/chjj/marked#overriding-renderer-methods), I'm going to call that not a blocker.

The only real update that's still necessary and I think we should do it separately from this, probably, is to update the index.md so we have a better output on the main docs page.

package.json
```diff
@@ -25,10 +25,12 @@
     "start": "grunt dev",
     "test": "grunt test",
     "docs": "npm run docs:lint && npm run docs:api",
-    "docs:api": "jsdoc -r src/js -d docs/api -c .jsdoc.json",
+    "jsdoc": "jsdoc",
```
BrandonOCasey: this will only cause "There are no input files to process.", maybe we should skip this script?

package.json
```diff
@@ -25,10 +25,12 @@
...
+    "docs:api": "jsdoc -c .jsdoc.json",
     "docs:lint": "remark -- './**/*.md'",
     "docs:fix": "remark --output -- './**/*.md'",
-    "babel": "babel src/js -d es5"
+    "babel": "babel src/js -d es5",
+    "prepublish": "not-in-install && npm run docs:api || in-install"
```
BrandonOCasey: If I understand this right this script will only build the docs `prepublish` but not when `npm install` is run on the project. Is that right? (if so this is awesome, and I had no idea that we could do this)

.jsdoc.json
```diff
@@ -1,4 +1,38 @@
...
+  "templates": {
+    "default": {
+      "staticFiles": {
+        "include": ["build/docs/"]
+      }
+    },
+    "logo": {
+      "url": "//videojs.com/img/logo.png",
+      "height": "30px",
+      "width": "214px"
```
BrandonOCasey: Is there anyway to include videojs.com as a link somewhere on the navigation?

.jsdoc.json
```diff
@@ -1,4 +1,38 @@
...
+    "encoding": "utf8"
+  },
+  "templates": {
+    "default": {
+      "staticFiles": {
+        "include": ["build/docs/"]
+      }
+    },
+    "logo": {
+      "url": "//videojs.com/img/logo.png",
```
BrandonOCasey: should this logo url have http:// on the front? doesn't seem to load for me without it.

package.json
gkatsev: Yup, that's what this does.

.jsdoc.json
gkatsev: are you loading it via file:///? That won't work then. But we could potentially just do http://

.jsdoc.json
gkatsev: Hm... I can definitely add a link on the footer. Making the logo a link would be awesome but might need an update to the template.

package.json
gkatsev: This is like the `grunt` task above, it's so that you can more easily run `jsdoc` commands on the cli without installing it globally if `jsdoc -c .jsdoc.json` doesn't suit what you're trying to do.

.jsdoc.json
BrandonOCasey: without http:// on the front chrome seems to default to file://. Even with that when I open the generated index.html file the logo is not loaded.

.jsdoc.json
BrandonOCasey: ahh didn't notice the footer link maybe add an issue to add a link in for the logo to the website.

package.json
BrandonOCasey: seems outside of the scope of what scripts should do. I think its easy enough to do `./node_modules/.bin/jsdoc`

.jsdoc.json
gkatsev: I'll add it as a note in 6.0 project.

package.json
gkatsev: It's a pretty common pattern nowadays.

PR #3916, opened 2017-01-04
fix: expose tech but warn without safety var

gkatsev:


gkatsev:
The tests need to be updated.

src/js/player.js
```diff
@@ -920,32 +920,27 @@ class Player extends Component {
...
-      To disable this alert and error, please pass in an object with
-      \`IWillNotUseThisInPlugins\` to the \`tech\` method. See
-      https://github.com/videojs/video.js/issues/2617 for more info.
-    `;
-
-    window.alert(errorText);
-    throw new Error(errorText);
+    if (!safety) {
+      const warnText = `
+        Using the tech directly can be dangerous. I hope you know what you're doing.
```
dmlap: ha!

dmlap:
LGTM

src/js/player.js
```diff
@@ -920,32 +920,27 @@ class Player extends Component {
...
+        See https://github.com/videojs/video.js/issues/2617 for more info.
+      `;
+
+      log.warn(warnText);
```
misteroneill: The variable seems superfluous. Is there a reason not to pass a string directly to `warn()`?

gkatsev: Nope.

src/js/player.js
```diff
@@ -920,32 +920,25 @@ class Player extends Component {
...
-      To disable this alert and error, please pass in an object with
-      \`IWillNotUseThisInPlugins\` to the \`tech\` method. See
-      https://github.com/videojs/video.js/issues/2617 for more info.
-    `;
-
-    window.alert(errorText);
-    throw new Error(errorText);
+    if (!safety) {
+      log.warn(`
+        Using the tech directly can be dangerous. I hope you know what you're doing.
```
BrandonOCasey: won't this add lots of whitespace to the start of the warning?

src/js/player.js
```diff
@@ -920,32 +920,25 @@ class Player extends Component {
...
+   * It will print a warning by default about the danger of using the tech directly
+   * but any argument that is passed in will silence the warning.
    *
-   * @param {Object} safety
-   *        An object that must contain `{IWillNotUseThisInPlugins: true}`
-   *
-   * @param {boolean} safety.IWillNotUseThisInPlugins
-   *        Must be set to true or else this function will throw an error.
+   * @param {Any} safety
+   *        Anything passed in to silence the warning
```
BrandonOCasey: technically it will have to be a true value, due to the if check (we may want to check if it is `=== undefined` below)

src/js/player.js
```diff
@@ -920,32 +920,25 @@ class Player extends Component {
...
+   * Return a reference to the current {@link Tech}.
+   * It will print a warning by default about the danger of using the tech directly
+   * but any argument that is passed in will silence the warning.
    *
-   * @param {Object} safety
-   *        An object that must contain `{IWillNotUseThisInPlugins: true}`
-   *
-   * @param {boolean} safety.IWillNotUseThisInPlugins
-   *        Must be set to true or else this function will throw an error.
+   * @param {Any} safety
```
BrandonOCasey: I think that the type here should actually be `*`. https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#the-all-type

gkatsev: changed it to `*` and also made it optional via `[safety]`.

src/js/player.js
gkatsev: Yes, updated.

PR #3919, opened 2017-01-05
Require `videojs-vtt.js` via require rather than concat

BrandonOCasey:
## Description
Use require on `videojs-vtt.js` rather than adding on to the end of the bundle after building.
This PR replaces #3794 

## Specific Changes
* cleanup `build/grunt.js` for browserify
* use `aliasify` to exclude `videojs-vtt.js` for novtt builds
* allow fallback to old script vtt include
* fix a bug where `shell:babel` was passing `--w` to babel rather than `-w` or `--watch`

## Requirements Checklist

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -326,7 +326,7 @@ module.exports = function(grunt) {
       },
       buildnovtt: {
         options: browserifyGruntOptions({transform: [
-          ['aliasify', {aliases: {'videojs-vtt.js': false}}],
+          ['aliasify', {aliases: {'videojs-vtt.js': false}}]
         ]}),
         files: {
           'build/temp/alt/video.novtt.js': ['es5/video.js']
@@ -344,7 +344,7 @@ module.exports = function(grunt) {
       watchnovtt: {
         options: browserifyGruntOptions({
           transform: [
-            ['aliasify', {aliases: {'videojs-vtt.js': false}}],
+            ['aliasify', {aliases: {'videojs-vtt.js': false}}]
           ],
           watch: true,
           keepAlive: true,
@@ -500,7 +500,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('chg');
   grunt.loadNpmTasks('grunt-accessibility');
 
-  const buildDependents = [
+  grunt.registerTask('build', [
     'shell:lint',
     'clean:build',
 
@@ -519,13 +519,11 @@ module.exports = function(grunt) {
     'copy:swf',
     'copy:ie8',
     'vjslanguages'
-  ];
+  ]);
 
-  grunt.registerTask('build', buildDependents);
-  grunt.registerTask('build:dist', buildDependents.map(task => task));
   grunt.registerTask('dist', [
     'clean:dist',
-    'build:dist',
+    'build',
     'copy:dist',
     'copy:examples',
     'zip:dist'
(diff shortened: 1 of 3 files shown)


gkatsev:
As you mentioned @BrandonOCasey, we should look at how safe it is to bring this in as a minor since it's theoretically possible. Would make it a bit easier for current suers.

build/grunt.js
```diff
@@ -326,32 +318,39 @@ module.exports = function(grunt) {
       }
     },
     browserify: {
-      options: browserifyGruntOptions(),
       build: {
```
gkatsev: so, `build` is now the main task and takes the job of both `build` and `dist`?

build/grunt.js
```diff
@@ -531,12 +522,7 @@ module.exports = function(grunt) {
   ];
 
   grunt.registerTask('build', buildDependents);
-
-  grunt.registerTask(
-    'build:dist',
-    buildDependents.map(task => task === 'browserify:build' ? 'browserify:dist' : task)
-  );
-
+  grunt.registerTask('build:dist', buildDependents.map(task => task));
```
gkatsev: We don't need to map here anymore. Or we could just get rid of it.

build/grunt.js
BrandonOCasey: Yeah, the difference between the two was that `dist` would  replace the vtt.js string path with a url to the cdn. Since we will require vtt.js by default when we build it isn't really needed to do them separately

src/js/tech/tech.js
```diff
@@ -522,13 +523,25 @@ class Tech extends Component {
    *
    * @fires Tech#vttjsloaded
    * @fires Tech#vttjserror
-   * @fires Tech#texttrackchange
    */
   addWebVttScript_() {
     if (!window.WebVTT && this.el().parentNode !== null && this.el().parentNode !== undefined) {
+      const vtt = require('videojs-vtt.js');
```
gkatsev: so, basically, this requires vttjs and then below, we make sure that it becomes available globally unless the vttjs option was provided. In the novtt build, a transform tells browserify not to include this file in the output?

src/js/tech/tech.js
```diff
@@ -522,13 +523,25 @@ class Tech extends Component {
...
+
+      // load via require if avialable and vtt.js script location
```
gkatsev: `av*i*alable` to `ava*i*lable` (i after the a).

build/grunt.js
BrandonOCasey: we may just want to get rid of the task

src/js/tech/tech.js
BrandonOCasey: Yeah, we use the vtt.js option if we have it, if not we use the require if it is available (which it won't be for novtt builds), otherwise we use the cdn url.

The transform that you mention works like this:
[`aliasify`](https://github.com/benbria/aliasify) when passed [`false`](https://github.com/benbria/aliasify#stubbing-out-packages) for a package alias will turn the require call into an empty object `{}` before browserify includes in in the bundle.

gkatsev: Cool, just wanted to confirm I'm understanding it correctly.

BrandonOCasey:
@gkatsev fixed the typo and removed `build:dist` ready for re-review when you get the time.

Issue #3839, opened 2016-12-06
Backward seeking issue in Chrome 55


bitactive:
## Description
Chrome 55 introduced bug in VideoJS that results in wrong progressbar position after seeking backward.
 It can be reproduced on latest stable 5.12.6 and master 5.14.1

## Steps to reproduce
Open latest Chrome 55 on Windows
1. Go to http://videojs.com/
2. Click play on the video player
3. Quickly seek to position near end (for example 0:40)
4. After video start playing do another seek to beginning (for example 0:10)

## Results
In most cases progressbar does not move to seeked position.
Sometimes (depends on connection speed) progressbar move to seeked position, start buffering and go back just after start playing.
### Expected
Progressbar move to seeked position.

### Actual
Video does seek to selected position, but progres bar does not reflect this status.

## Additional Information
Chrome 54 works fine. Bug appear in Chrome 55 on Windows (tested on Win7 & 10).
Other html5 players are working fine.

Bug exists on latest stable 5.12.6 and master 5.14.1

RodrigoMaster7:
I got a temporary solution in version 5.12.6, replacing:
```
  SeekBar.prototype.handleMouseDown = function handleMouseDown(event) {
    _Slider.prototype.handleMouseDown.call(this, event);

    this.player_.scrubbing(true);

    this.videoWasPlaying = !this.player_.paused();
    this.player_.pause();
  };
```
For This:

```
SeekBar.prototype.handleMouseDown = function handleMouseDown(event) {
    _Slider.prototype.handleMouseDown.call(this, event);

    this.player_.scrubbing(true);

  };
```

Please tell me if this works for you.

(Translated by Google)

bitactive:
Yes, i can confirm that this patch fixes the issue.

gkatsev:
Interesting, I can confirm this. Unfortunately, just removing the pause isn't a good solution. The reason is that if you're seeking via scrubbing (if you press and hold the mouse on the progress bar and then move it around), removing the pause causes the video to continue to play if you wait for a second.

I think a potential fix would be to just wrap the two lines that were removed in a setTimeout:
```js
SeekBar.prototype.handleMouseDown = function handleMouseDown(event) {
    _Slider.prototype.handleMouseDown.call(this, event);

    this.player_.scrubbing(true);

    this.setTimeout(function() {
        this.videoWasPlaying = !this.player_.paused();
         this.player_.pause();
    });
};
```
This seems to work but not sure whether there are any implications for this.

[...]

PR #3921, opened 2017-01-06
Pause player before seeking in seek bar mousedown

MattiasBuelens:
## Description
I ran into the backwards seeking problem with Chrome 55, as previously reported in #3839. After some investigation, I found it was also reported in the Chromium bug tracker as [bug 675556](https://bugs.chromium.org/p/chromium/issues/detail?id=675556), and it should be fixed in Chrome 57.

I noticed that video.js 5.15.0 fixes this in #3886 by delaying the pause until the first `mousemove` or after 100ms (whichever happens first). However, the comments from the Chromium bug indicate that there is a simpler solution. The problem occurs when you call when you call `video.pause()` immediately after setting `video.currentTime`, however it does not occur when you pause **before** seeking.

## Specific Changes proposed
This change reverts #3886. Instead, it ensures that `SeekBar.handleMouseDown` first pauses the player before calling its super method (which in turn calls `handleMouseMove` which seeks the player).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -122,12 +122,12 @@ class SeekBar extends Slider {
    * @listens mousedown
    */
   handleMouseDown(event) {
-    super.handleMouseDown(event);
-
     this.player_.scrubbing(true);
 
     this.videoWasPlaying = !this.player_.paused();
     this.player_.pause();
+
+    super.handleMouseDown(event);
   }
 
   /**
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -127,10 +127,7 @@ class SeekBar extends Slider {
     this.player_.scrubbing(true);
 
     this.videoWasPlaying = !this.player_.paused();
-
-    this.pauseTimer_ = this.setTimeout(function() {
-      this.player_.pause();
-    }, 100);
+    this.player_.pause();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


gkatsev:
Interesting find. I wonder whether we should keep #3886 in case anything like this ever happens again in the future. Plus, I kind of like that in #3886 we only pause if necessary. What do you think?

MattiasBuelens:
@gkatsev I agree that it's nicer not having to pause and play when the user clicks the seek bar instead of dragging it. I haven't tested whether it's faster to just seek versus pause/seek/play, but I don't think there would be a noticeable performance difference.

Personally, I prefer synchronous solutions over random delays in my own code, but I understand why you'd want to keep #3886. If the pause timer ever turns out to be a problem, at least you now have an alternative solution. Feel free to close this PR, no hard feelings. 😛

gkatsev:
Synchronous solutions are definitely nice.

@videojs/core-committers what are your thoughts?

heff:
I would vote to remove the setTimeout. It feels like addressing the symptom instead of the cause, and setTimeout patches like to find ways to bite you back.

I don't know where this fits in, but for me (Chrome 57 (canary) Mac) on videojs.com (5.14.1), when I seek forward then back like the bug reported, the progress bar stays at the forward position and stops progressing with the current time. I guess that means it's not fixed in 57?

MattiasBuelens:
@heff What version of Chrome Canary are you running? The latest version (57.0.2976.0) no longer has the problem on videojs.com for me.

heff:
I was at 57.0.2971.0. Just updated to 57.0.2976.5 and the issue went away.

misteroneill:
A 👍 for preferring synchronous solutions. And another 👍 for change sets that are more red than green.

MattiasBuelens:
@misteroneill Without the commit to revert #3886, the change set would be +2/-2. So... yellow? 😛

PR #3929, opened 2017-01-10
Add check to ensure `hasAttribute` is a function

stramel:
## Description
For elements which parent doesn't have `hasAttribute` it needs do another check to prevent erroring out.

## Specific Changes proposed
Added a check to ensure the tag's parentNode has the `hasAttribute` function

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -473,7 +473,7 @@ class Player extends Component {
   createEl() {
     const tag = this.tag;
     let el;
-    const playerElIngest = this.playerElIngest_ = tag.parentNode && tag.parentNode.hasAttribute('data-vjs-player');
+    const playerElIngest = this.playerElIngest_ = tag.parentNode && tag.parentNode.hasAttribute && tag.parentNode.hasAttribute('data-vjs-player');
 
     if (playerElIngest) {
       el = this.el_ = tag.parentNode;


gkatsev:
Can you elaborate on when the tag's `parentNode` is not going to have a `hasAttribute` method? Want to understand the usecase better before pulling it in.

stramel:
@gkatsev I am using the `<template>` element which creates a `DocumentFragment` to encompass a video player.

I am specifically creating a Polymer component that extends the `video` element. When I use the [ready lifecycle callback](https://www.polymer-project.org/1.0/docs/devguide/registering-elements#lifecycle-callbacks) the local DOM has been initialized but not attached to the document. This means that the `DocumentFragment` is the parent. Using the attached lifecycle callback, does not hit this case since it has been attached and the parent is then a DOM element. However, the attached callback runs each time the element is attached. I was concerned there may be issues with videojs attempting to initialize the element multiple times. Thus I chose to use the ready callback which will only be called once.

Hope this clarifies things! 


REFRENCES:

- [template](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template)
- [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment)

gkatsev:
Ah, yup, makes sense! Thanks.

Issue #3918, opened 2017-01-05
`mouseup` over `MuteToggle` after dragging `VolumeBar` to zero incorrectly fires `handleClick` on `MuteToggle`


kevinlitchfield:
After clicking and dragging `VolumeBar` to zero, if the user releases the mouse over `MuteToggle`, `handleClick` fires on `MuteToggle`. Watch the checkbox tracking the `muted` property in these screencaptures:

### `mouseup` over `MuteToggle`, `handleClick` fired on `MuteToggle` and `muted` is toggled to `true`

![click](https://cloud.githubusercontent.com/assets/11491159/21685972/9973213a-d331-11e6-91b1-0038039d8d70.gif)

### `mouseup` over `VolumeBar`, `handleClick` not fired on `MuteToggle` and `muted` remains `false`

![noclick](https://cloud.githubusercontent.com/assets/11491159/21685943/7707220e-d331-11e6-80b5-14085b6c3e15.gif)

Currently, this bug has no implications for the user, because as discussed in #3909, clicking `MuteToggle` after setting the volume slider to 0% has no user-visible effect. But fixing it is important to addressing #3909, because if clicking `MuteToggle` when `volume` is zero sets `volume` to something greater than zero, then users will be able to accidentally trigger that behavior by `mouseup`ing over `MuteToggle` after dragging `VolumeBar` to zero.

misteroneill:
Thanks for the nice report. 👍

gkatsev:
This is probably something we should fix for 5.x. As part of 6.0, we're moving to a better, and more accessible volume control which would actually be two separate controls, so, it potentially won't be an issue. But we should keep an eye out for this when working on it. /cc @BrandonOCasey.

BrandonOCasey:
This should be fixed in master now that #3957 is merged in

gkatsev:
I think this is fixed with the new volume behavior in 6.0.

PR #3935, opened 2017-01-11
Accessible Volume Control

BrandonOCasey:
## Description
Separate VolumeBar and MuteToggle from VolumeMenuButton and make it accessible.

## Specific Changes proposed
* Remove VolumeMenuButton
* Add MuteToggle to ControlBar
* Add VolumeControl to ControlBar
* Make VolumeControl Work with VolumeBar
* Make VolumeControl Work with MuteToggle
* Verify Accessibility

## Requirements Checklist
  - [x] Change has been verified in an actual browser 
    - [x] Chome /w Voiceover on OSX
    - [x] Firefox /w Voiceover on OSX
    - [x] Safari  /w Voiceover on OSX
    - [x] Edge /w JAWS
    - [x] IE 11 /w JAWS
    - [x] IE 10 /w JAWS
    - [x] IE 9 /w JAWS
    - [x] IE 8 /w JAWS 


src/js/control-bar/mute-toggle.js
```diff
@@ -92,7 +95,7 @@ class MuteToggle extends Button {
     // Don't rewrite the button text if the actual text doesn't change.
     // This causes unnecessary and confusing information for screen reader users.
     // This check is needed because this function gets called every time the volume level is changed.
-    const toMute = this.player_.muted() ? 'Unmute' : 'Mute';
+    const toMute = this.player_.muted() ? 'Mute Toggle Unmute' : 'Mute Toggle Mute';
```
gkatsev: this would change our localizations, I think it would be better not to change it, if we don't need to.

src/js/control-bar/play-toggle.js
```diff
@@ -68,7 +68,7 @@ class PlayToggle extends Button {
     this.removeClass('vjs-paused');
     this.addClass('vjs-playing');
     // change the button text to "Pause"
-    this.controlText('Pause');
+    this.controlText('Play Toggle Pause');
```
gkatsev: localization change

src/js/control-bar/volume-control/volume-control.js
```diff
@@ -36,6 +36,23 @@ class VolumeControl extends Component {
         this.removeClass('vjs-hidden');
       }
     });
+
+    this.on(this.volumeBar, ['focus', 'slideractive'], () => {
```
gkatsev: this is the change from the volume-menu-button to make changing volume more easily?

src/js/control-bar/control-bar.js
```diff
@@ -56,7 +55,8 @@ class ControlBar extends Component {
 ControlBar.prototype.options_ = {
   children: [
     'playToggle',
-    'volumeMenuButton',
+    'muteToggle',
```
gkatsev: what about instead of making them direct children of the control bar have a second component that has them as children? That way, we can style them better as part of the accessible control that works like the volume-menu-button but also separately, for those that want them separate.

BrandonOCasey: I am thinking that we can make `MuteToggle` and `VolumeBar` both a child of `VolumeControl`, what do you think?

src/js/control-bar/volume-control/volume-control.js
BrandonOCasey: yeah this makes it so that you can click and drag anywhere rather than having to click and drag directly on the volume slider.

src/js/control-bar/control-bar.js
gkatsev: I think we should keep mutetoggle and volumecontrol working as is if they're used by themselves.

BrandonOCasey:
@gkatsev ready for another review, just need to get the vertical volume bar working again.

src/js/control-bar/volume-control/check-volume-support.js
```diff
@@ -0,0 +1,27 @@
...
+  // hide volume controls when they're not supported by the current tech
+  if (player.tech_ && player.tech_.featuresVolumeControl === false) {
+    self.addClass('vjs-hidden');
+  }
+  self.on(player, 'loadstart', function() {
+    if (player.tech_.featuresVolumeControl === false) {
+      self.addClass('vjs-hidden');
+    } else {
+      self.removeClass('vjs-hidden');
+    }
```
misteroneill: What about replacing this conditional block with:

```js
self.toggleClass('vjs-hidden', player.tech_.featuresVolumeControl === false);
```

src/js/control-bar/volume-control/mute-toggle.js
```diff
@@ -24,23 +25,14 @@ class MuteToggle extends Button {
...
+    this.on(player, 'loadstart', this.update);
 
-      if (player.tech_.featuresVolumeControl === false) {
-        this.addClass('vjs-hidden');
-      } else {
-        this.removeClass('vjs-hidden');
-      }
-    });
+    // update muted icon on volumechange
+    this.on(player, 'volumechange', this.update);
```
misteroneill: What about combining this with the above for:

```js
this.on(player, ['loadstart', 'volumechange'], this.update);
```

BrandonOCasey:
Fixes #3918

src/css/components/_control.scss
```diff
@@ -36,3 +36,28 @@
   display: table-cell;
   vertical-align: middle;
 }
+
+// visually hide a control-bar component without changing the position
+// and without hiding it from screen readers. This also preserves
+// a transition animation.
+.video-js .vjs-control .vjs-visual-hide-vertical {
```
gkatsev: is it necessary to have two of these? Why not just set both the height and width to be `1px` in all cases?
What about other visual hiding techniques like `left: -9999px;` or the recommended approach from webaim: http://webaim.org/techniques/css/invisiblecontent/#absolutepositioning?

src/js/control-bar/volume-panel.js
```diff
@@ -0,0 +1,89 @@
...
+   * @param {Object} [options={}]
+   *        The key/value store of player options.
+   */
+  constructor(player, options) {
+
+    // pass the inline option down to the VolumeControl if
+    // the VolumeControl is on.
+    if (options.inline === false && (!options.volumeControl || isPlain(options.volumeControl))) {
+      options.volumeControl = options.volumeControl || {};
+      options.volumeControl.inline = options.inline;
```
gkatsev: maybe we should translate the `inline` option to vertical here instead of even at the volume control level?

src/js/control-bar/volume-panel.js
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Create the `Component`'s DOM element
+   *
+   * @return {Element}
+   *         The element that was created.
+   */
+  createEl() {
+    let orientationClass = 'vjs-volume-panel-horizontal';
+
+    if (this.options_.inline === false) {
```
gkatsev: if you set the default to be vertical, this if can be simplified to be `if (this.options_.inline) { horizontal }`

src/css/components/_control.scss
BrandonOCasey: We need to have two in order to correctly `transition` components like the volume bar that are hidden and show on the `contral-bar`. If we were to do the `width` and `heigth` at the same time the horizontal volume-bar would appear to move up and left as it transitions out which looked a little weird and was not the current behavior. A vertical volume bar would have a similar transition problem.

I looked at the other hiding techniques but the fact that they position the element in a different place means that we would lose the transition.

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -25,9 +25,17 @@ class VolumeBar extends Slider {
...
    */
   constructor(player, options) {
+    if (options.inline === false) {
+      options.vertical = true;
+    }
+
     super(player, options);
+
+    // hide this control if volume support is missing
+    checkVolumeSupport(this, player);
```
BrandonOCasey: agreed

BrandonOCasey:
seems like I need to revisit the inline vs vertical option overall, will do

src/js/control-bar/volume-panel.js
BrandonOCasey: Well when I think of the default volumeBar/volumeControl I think of it as horizontal, so I think that horizontal should stay the default.

src/css/components/_volume.scss
```diff
@@ -1,27 +1,32 @@
...
   @extend .vjs-icon-volume-mid;
 }
 
 .video-js .vjs-volume-control {
-  width: 5em;
+  width: auto;
+  margin-right: 1em;
+  @include flex(none);
+  @include display-flex(center);
+  @include transition(all 0.4s);
```
gkatsev: we should avoid using transition all but instead only transition the specific properties we want. Using `all` here potentially can cause performance issues since the browsers then try to transition all the properties even if we only care about width or height, for example.
We don't have to solve it immediately, but we should do it before a 6.0 release.

src/css/components/_control.scss
gkatsev: Ok, I see what there are two. I still wonder if there's some way to simplify it.

Also, I just noticed that when the volume control is hidden as part of the volume panel, it adds a pixel to the width of the panel, we probably don't want that. So, we could potentially use the webaim hiding technique on a delay after the height/width transition?

Also, I noticed a small hiccup with the animation:
volume control animating as part of volume panel
Notice the small white bar that shows up at of the control closing. This potentially stems from the same issue as the last paragraph.

gkatsev:
Also, comments aside, this looks great. Thanks.

src/js/control-bar/volume-panel.js
gkatsev: Yup, just thinking of how to simplify the block.
Maybe just switch to `!this.options.inline`, checking a boolean against `false` is a bit redundant, I think.

docs/guides/components.md
```diff
@@ -316,6 +317,22 @@ Player
 
 ## Specific Component Details
 
+### Volume Panel
+
+The `VolumePanel` includes the `MuteToggle` and the `VolumeBar` Components, which will be hidden if volume changes are not supported. There is one important option for the `VolumePanel` which can make your `VolumeControl` appear vertically over the `MuteToggle`. This can be set by passing `VolumePanel` `{inline: false}` as the default behavior is a horizontal `VolumeControl` with `{inline: true}`.
```
gkatsev: `VolumeControl` not `VolumeBar` in the first sentence.

src/css/components/_control.scss
BrandonOCasey: yeah I originally had a setTImeout call on `hide()` that would run after `0.4s` that made the 1px go away but it made things a lot more complex. Hopefully we can find another fix for this. Maybe this should go with the IE8/IE9 style fix PR?

gkatsev: Fixing this as another task sounds good to me. But it has to be fixed before 6.0.

src/js/control-bar/volume-control/check-volume-support.js
gkatsev: `self.toggleClass('vjs-hidden', !player.tech_.featuresVolumeControl)`

src/js/control-bar/volume-panel.js
BrandonOCasey: we have to check against false here since most of the time inline we not be defined.

src/js/control-bar/volume-control/check-volume-support.js
```diff
@@ -0,0 +1,27 @@
...
+ *        The component that should be hidden if volume is unsupported
+ *
+ * @param {Player} player
+ *        A reference to the player
+ *
+ * @private
+ */
+const checkVolumeSupport = function(self, player) {
+  // hide volume controls when they're not supported by the current tech
+  if (player.tech_ && player.tech_.featuresVolumeControl === false) {
```
gkatsev: `featuresVolumeControl` is a boolean, so, I guess just:
```js
if (player.tech_ && !player.tech_.featuresVolumeControl) {
```

src/js/control-bar/volume-control/volume-control.js
```diff
@@ -22,20 +24,80 @@ class VolumeControl extends Component {
...
+    this.on(this.volumeBar, ['blur'], () => this.hide());
+  }
+
+  /**
+   * Remove the visual hidden state from the `VolumeControl`.
+   */
+  show() {
+    this.shouldHide_ = false;
+
+    if (this.options_.vertical === true) {
```
gkatsev: `vertical` is already a boolean, no need to compare against `true`

src/js/control-bar/volume-control/volume-control.js
```diff
@@ -22,20 +24,80 @@ class VolumeControl extends Component {
...
+    // if we are currently locked to the showing state
+    // don't hide, but store that we should hide when
+    // lockShowing_ turns to a false value.
+    if (this.lockShowing_) {
+      this.shouldHide_ = true;
+      return;
+    }
+
+    // animate hiding the bar via transitions
+    if (this.options_.vertical === true) {
```
gkatsev: `vertical` is already a boolean, no need to compare against `true`

src/js/control-bar/volume-control/volume-control.js
```diff
@@ -45,8 +107,14 @@ class VolumeControl extends Component {
    *         The element that was created.
    */
   createEl() {
+    let orientationClass = 'vjs-volume-horizonal vjs-visual-hide-horizontal';
+
+    if (this.options_.vertical === true) {
```
gkatsev: `vertical` is already a boolean, no need to compare against `true`

src/js/control-bar/volume-control/volume-control.js
```diff
@@ -22,20 +24,80 @@ class VolumeControl extends Component {
...
-  constructor(player, options) {
-    super(player, options);
+  constructor(player, options = {}) {
 
-    // hide volume controls when they're not supported by the current tech
-    if (player.tech_ && player.tech_.featuresVolumeControl === false) {
-      this.addClass('vjs-hidden');
+    // Pass the vertical option down to the VolumeBar if
+    // the VolumeBar is turned on.
+    if (typeof options.vertical !== 'undefined' && (!options.volumeBar || isPlain(options.volumeBar))) {
```
gkatsev: would be good to default `vertical` to false so it's always a boolean, though, `undefined` will get the results we want as well but being explicit is nice.

src/js/control-bar/volume-panel.js
gkatsev: Ah, I see. What about making sure that `inline` gets set with a value in all cases?

src/css/components/_control.scss
BrandonOCasey: ok I will add it to the list of css fixes in a note on the Project

BrandonOCasey:
Pushing this branch upstream so that tests will run on browserstack, See #3957

Issue #3909, opened 2017-01-03
Clicking muted icon after setting volume slider to 0% has no user-visible effect


kevinlitchfield:
Clicking the muted icon (<img width="24" alt="screen shot 2017-01-03 at 6 25 01 pm" src="https://cloud.githubusercontent.com/assets/11491159/21626879/f6de45bc-d1e1-11e6-8e87-b6f0a9271d9a.png">) after setting the volume slider to 0% has no user-visible effect.

![videojs](https://cloud.githubusercontent.com/assets/11491159/21626859/bef7164c-d1e1-11e6-90e3-70941a619e49.gif)

This is because:

- The muted icon is displayed whenever `volume` is `0`, regardless of whether `muted` is `true` or `false`.
- Clicking the muted icon toggles the `muted` property and has no effect on `volume`.

So in this case, since `volume` is `0`, the muted icon doesn't change on repeated clicks, despite the changing `muted` property and despite the changing tooltop text ("Mute"/"Unmute"). This has confused users of the app I use Video.js in.

Two suggestions for a solution:

### 1. YouTube solution

![youtube](https://cloud.githubusercontent.com/assets/11491159/21626815/7f0c1d0c-d1e1-11e6-94f4-f0712dc11a37.gif)

When clicking the muted icon after moving the volume slider to zero, **unmute and set `volume` to the last value before it was manually set to `0`**. This is what YouTube does, and I think this is the best UX.

### 2. macOS solution

When clicking the muted icon after moving the volume slider to zero, **unmute and show the normal volume icon, while keeping `volume` at `0`**. This is what macOS does.

![macos](https://cloud.githubusercontent.com/assets/11491159/21626808/71ffa598-d1e1-11e6-9248-2981c396613f.gif)

Would the maintainers consider a PR addressing this issue?

mister-ben:
There was a discussion about this a little while back, maybe on Slack. I don't recall whether a consensus was reached but I'd agree that it would be better UX if clicking unmute were to set volume to something greater than 0 if volume is 0.

gkatsev:
We're in the process of making a major update to videojs. Making the change sounds ok to me.
@videojs/core-committers any thoughts?

heff:
[...]

PR #3942, opened 2017-01-13
Improve mute button usability

kevinlitchfield:
Clicking `MuteToggle` when `volume` is 0 now has more user-friendly behavior. Previously, clicking MuteToggle when `volume` was 0 had no user-visible effect (as discussed in https://github.com/videojs/video.js/issues/3909, which this PR addresses).

There are three possible scenarios I'm aware of for setting the volume to 0 and then dealing with a click on `MuteToggle` that need to be taken into account:

1. **User drags `VolumeBar` to 0 with mouse or finger.** In this case, `volume` should be restored to the level it was at when the user started dragging `VolumeBar`.
1. **User steps `volume` to 0 with keyboard.** In this case, clicking `MuteToggle` should restore the volume to its last value before it reached 0.
1. **`volume` is set directly to 0 with external JS, by a plugin, etc.** In this case, it should be up to the programmer to control what level the volume should be set to when the user clicks `MuteToggle`, but it should never remain at 0.

How this solution works:

- Clicking `MuteToggle` when volume is zero sets `volume` to the value of a new property, `lastVolume`.
- `lastVolume` is updated to match `volume` when `Player#volume` sets a new value for `volume`, unless the new value of `volume` is 0.
- When a user starts to drag `VolumeBar`, `volume` is saved into another new property, `volumeBeforeDrag`. When the user finishes dragging `VolumeBar`, if `volume` is 0, `lastVolume` is updated to match `volumeBeforeDrag`.

**Note:** This should not be merged until https://github.com/videojs/video.js/issues/3918 is fixed. As explained in that bug report: 

> If clicking `MuteToggle` when `volume` is zero sets `volume` to something greater than zero, then users will be able to accidentally trigger that behavior by `mouseup`ing over `MuteToggle` after dragging `VolumeBar` to zero.

### DIFF
--- a/src/js/control-bar/mute-toggle.js
+++ b/src/js/control-bar/mute-toggle.js
@@ -53,7 +53,15 @@ class MuteToggle extends Button {
    * @listens click
    */
   handleClick(event) {
-    this.player_.muted(this.player_.muted() ? false : true);
+    const vol = this.player_.volume();
+    const lastVolume = this.player_.lastVolume();
+
+    if (vol === 0) {
+      this.player_.volume(lastVolume);
+      this.player_.muted(false);
+    } else {
+      this.player_.muted(this.player_.muted() ? false : true);
+    }
   }
 
   /**
--- a/src/js/control-bar/mute-toggle.js
+++ b/src/js/control-bar/mute-toggle.js
@@ -54,7 +54,7 @@ class MuteToggle extends Button {
    */
   handleClick(event) {
     const vol = this.player_.volume();
-    const lastVolume = this.player_.lastVolume();
+    const lastVolume = this.player_.lastVolume_();
 
     if (vol === 0) {
       this.player_.volume(lastVolume);
--- a/src/js/control-bar/volume-control/volume-bar.js
+++ b/src/js/control-bar/volume-control/volume-bar.js
@@ -25,7 +25,8 @@ class VolumeBar extends Slider {
    */
   constructor(player, options) {
     super(player, options);
-
+    this.on('slideractive', this.updateVolumeBeforeDrag);
+    this.on('sliderinactive', this.updateLastVolume);
     this.on(player, 'volumechange', this.updateARIAAttributes);
     player.ready(() => this.updateARIAAttributes);
   }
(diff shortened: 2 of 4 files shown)


BrandonOCasey:
Overall this is great! Thanks for the PR!

I think that there may be a way to make this a bit simpler: 
* Remove `volumeBeforeDrag` and listening for `sliderinactive` in the `VolumeBar` constructor
* `Player#lastVolume` is updated by `VolumeBar`:
  1. `VolumeBar` listens to `slideractive` on itself (as it does in the current implementation)
  2. The `slideractive` listener records the starting volume and starts listening for one `sliderinactive` event on itself
  3. The `sliderinactive` listener:
     * Sets `lastVolume` to the current volume unless it is zero or the volume from the `slideractive` listener if its not zero
     * Does nothing if the volume was changed from 0 to 0

I also think that we should probably make  `lastVolume` on the player a private function prefixed with an `_` at the end of it. What does everyone else think? @videojs/contributors

src/js/player.js
```diff
@@ -1831,6 +1839,48 @@ class Player extends Component {
...
+   *         - 0.5 is half volume or 50%
+   *
+   * @return {Player|number}
+   *         a reference to the calling player when setting and the
+   *         current volume as a percent when getting
+   */
+  lastVolume(percentAsDecimal) {
+    if (percentAsDecimal !== undefined) {
+      this.lastVolume_ = percentAsDecimal;
+      return this;
```
gkatsev: This should just `return` when used as setter.

src/js/player.js
```diff
@@ -1831,6 +1839,48 @@ class Player extends Component {
...
+    *         The new last volume as a decimal percent:
+    *         - 0 is muted/0%/off
+    *         - 1.0 is 100%/full
+    *         - 0.5 is half volume or 50%
+    *
+    * @return {Player|number}
+    *         a reference to the calling player when setting and the
+    *         current volume as a percent when getting
+    */
+  volumeBeforeDrag(percentAsDecimal) {
```
gkatsev: this should also be marked as private

src/js/control-bar/mute-toggle.js
```diff
@@ -65,7 +65,15 @@ class MuteToggle extends Button {
    * @listens click
    */
   handleClick(event) {
-    this.player_.muted(this.player_.muted() ? false : true);
+    const vol = this.player_.volume();
+    const lastVolume = this.player_.lastVolume();
```
gkatsev: is it possible for the `lastVolume` to end up being 0? Is it worth handling? At that point, I guess would be to set the volume to 0.5?

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -112,6 +112,24 @@ class VolumeBar extends Slider {
...
   }
 
+  /**
+   * When user starts dragging the VolumeBar, store the volume and listen for
+   * the end of the drag. When the drag ends, if the volume was set to zero,
+   * set lastVolume to the stored volume.
+   *
+   * @listens slideractive
+   */
+  updateLastVolume() {
```
BrandonOCasey: we probably want this function to be private (only used internally) just like `Player#lastVolume_()`

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -112,6 +112,24 @@ class VolumeBar extends Slider {
...
+    const volumeBeforeDrag = this.player_.volume();
+
+    this.on('sliderinactive', function setLastVolume() {
```
BrandonOCasey: You can actually use `this.one('sliderinactive'` here that way you don't have to use `this.off` below. I would also say that you can just use an arrow function here rather than a named one. ex: 
```js
this.one('sliderinactive', () => {
// current function goes here
});
```

src/js/player.js
```diff
@@ -1873,6 +1880,26 @@ class Player extends Component {
...
+   * @param  {number} [percentAsDecimal]
+   *         The new last volume as a decimal percent:
+   *         - 0 is muted/0%/off
+   *         - 1.0 is 100%/full
+   *         - 0.5 is half volume or 50%
+   *
+   * @return {number}
+   *         the current value of lastVolume as a percent when getting
+   */
+  lastVolume_(percentAsDecimal) {
```
gkatsev: this should have an `@private` jsdoc directive.

src/js/control-bar/volume-control/volume-bar.js
gkatsev: also, `@private` jsdoc directive

kevinlitchfield:
@BrandonOCasey @gkatsev Thanks so much for the feedback! Have addressed.

gkatsev:
@kevinlitchfield thanks so much for your work!

PR #3953, opened 2017-01-18
test: fix tests

gkatsev:
## Description

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated


.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": [ ["es2015", {"loose": true}] ],
-  "plugins": ["transform-es3-property-literals", "transform-es3-member-expression-literals", "inline-json"]
+  "presets": [ "es3", ["es2015", {"loose": true}] ],
```
BrandonOCasey: How much size does this add to the video.js?

gkatsev: It doesn't. It's equivalent to using the es3 plugins but because of when and how plugins and presets are run, having the preset is more reliable for making sure that things are quoted correctly.

BrandonOCasey: we should probably switch this in the generator/spellbook

gkatsev: Yeah, might be good, though, I wouldn't worry too much as long as it continues to work.

PR #3959, opened 2017-01-19
refactor: Evented Components

misteroneill:
## Description
The PR for advanced plugins introduced the concept of mixins and added two: evented and stateful.

This PR refactors `Component` to use the evented mixin, granting it event broadcast/handling capabilities.

I will open a second PR for implementing stateful components.

## Requirements Checklist
  - ~~Docs/guides updated~~
  - ~~Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))~~

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -4,11 +4,11 @@
  * @file component.js
  */
 import window from 'global/window';
+import evented from './mixins/evented';
 import * as Dom from './utils/dom.js';
 import * as DomData from './utils/dom-data';
 import * as Fn from './utils/fn.js';
 import * as Guid from './utils/guid.js';
-import * as Events from './utils/events.js';
 import log from './utils/log.js';
 import toTitleCase from './utils/to-title-case.js';
 import mergeOptions from './utils/merge-options.js';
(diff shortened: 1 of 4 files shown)


src/js/component.js
```diff
@@ -134,9 +137,6 @@ class Component {
     this.childIndex_ = null;
     this.childNameIndex_ = null;
 
-    // Remove all event listeners.
-    this.off();
```
gkatsev: does this happen automatically now?

src/js/player.js
```diff
@@ -348,6 +349,9 @@ class Player extends Component {
 
     this.el_ = this.createEl();
 
+    // Make this an evented object and use `el_` as its event bus.
+    evented(this, {eventBusKey: 'el_'});
```
gkatsev: Is this necessary if it happens in `Component`?

src/js/component.js
misteroneill: Yes, it does. It's part of why the mixin is applied to instances instead of prototypes (to answer your other question).

It's a toss-up. You can have this `off()` on `dispose` behavior built-in or you can apply the mixin to a prototype and manually call `this.off()` in the `dispose` method of that class.

src/js/player.js
misteroneill: See other comment.

src/js/component.js
gkatsev: Sounds good. Just wanted an answer the question before we were to merge it in :)

PR #3964, opened 2017-01-21
Update Vietnamese string

ngoisaosang:
Fixed and updated


gkatsev:
Thanks @ngoisaosang! By the way, in the future, you don't need to close the PR, if you just push another commit to the same branch the PR was made from, the PR will be updated.

PR #3969, opened 2017-01-23
Update MW to require a factory, add *-mw

gkatsev:
## Description
As I was playing around with middleware, in the first RC, I realized that the current way of `videojs.use` with an object was lacking. Specifically, if you have multiple players on the page, the middleware would need to be able to handle tech management itself, which doesn't seem right. Also, the mw didn't know about what player it was attached to.
Now, `videojs.use` takes in a function that given a player, returns a middleware object. The middleware object looks the same as the current object we use for registration. This allows us to have per-player instances of middleware and because it's a factory, a user can use a class or Component or Plugin object for managing their middleware if they want, as long as the methods are available.
For example:
```js
videojs.use('video/foo', (player) => ({
  setSource(src, next) {
    if (player.playbackRate() > 1) {
      next(null, {
       src:  'foo',
       type: 'video/mp4'
      });
    } else {
      next(new Error('we don't support this));
    }
  }
}));
```

Also, this PR adds the ability to register "star middleware" (* mw),  which will always be run through right before selecting the tech.

### DIFF
--- a/src/js/tech/middleware.js
+++ b/src/js/tech/middleware.js
@@ -32,8 +32,11 @@ export function set(middleware, tech, method, arg) {
 }
 
 export const allowedGetters = {
+  buffered: 1,
   currentTime: 1,
-  duration: 1
+  duration: 1,
+  seekable: 1,
+  played: 1
 };
 
 export const allowedSetters = {
--- a/src/js/tech/middleware.js
+++ b/src/js/tech/middleware.js
@@ -61,6 +61,7 @@ function setSourceHelper(src = {}, middleware = [], next, player, acc = [], last
   // then call the mw's setSource method
   } else if (mwFactory) {
     const mw = mwFactory(player);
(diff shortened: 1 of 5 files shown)


src/js/tech/middleware.js
```diff
@@ -15,8 +15,8 @@ export function getMiddleware(type) {
   return middlewares;
 }
 
-export function setSource(setTimeout, src, next) {
-  setTimeout(() => setSourceHelper(src, middlewares[src.type], next), 1);
+export function setSource(player, setTimeout, src, next) {
+  setTimeout(() => setSourceHelper(src, middlewares[src.type], next, player), 1);
```
gkatsev: actually, I just realized I don't need to pass in `setTimeout` here. But it's the `setTimeout` from `player`, so, that if `player` is disposed`, so will this timeout.

PR #3972, opened 2017-01-23
docs(guides): add basic ReactJS integration example

revolunet:
## Description

Just some doc for a clean ReactJS integration

## Specific Changes proposed

add `guides/react.md `

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/docs/guides/react.md
+++ b/docs/guides/react.md
@@ -1,0 +1,61 @@
+# video.js and ReactJS integration
+
+Here's a basic ReactJS player implementation.
+
+It just instantiate the video.js player on `componentDidMount` and destroy it on `componentWillUnmount`.
+
+```jsx
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import videojs from 'video.js'
+
+// this loads video.js CSS using webpack loaders
+require('!style-loader!css-loader!video.js/dist/video-js.css')
+
+export default class VideoPlayer extends React.Component {
+  componentDidMount() {
+    // get the <video> DOM node
+    const videoNode = ReactDOM.findDOMNode(this).querySelector('video');
+
+    // instantiate video.js
+    this.player = videojs(videoNode, this.props, function onPlayerReady() {
+      console.log('onPlayerReady', this)
+    });
+  }
+
+  // destroy player on unmount
+  componentWillUnmount() {
+    if (this.player) {
+      this.player.dispose()
+    }
+  }
+
+  // wrap the player in a div with a `data-vjs-player` attribute
+  // so videojs won't create additional wrapper in the DOM
+  // see https://github.com/videojs/video.js/pull/3856
+  render() {
+    return (
+      <div data-vjs-player>
+        <video className="video-js"></video>
+      </div>
+    )
+  }
+}
+```
+
+You can then use it like this :
+
+```jsx
+// see https://github.com/videojs/video.js/blob/master/docs/guides/options.md
+const videoJsOptions = {
+  autoPlay: true,
+  controls: true,
+  sources: [{
+    src: '/path/to/video.mp4',
+    type: 'video/mp4'
+  }]
+}
+
+return <VideoPlayer { ...videoJsOptions } />
+```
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -269,6 +270,12 @@ Yes! Please [submit an issue or open a pull request][pr-issue-question] if this
 
 Yes! Please [submit an issue or open a pull request][pr-issue-question] if this does not work.
 
+Use `require('!style-loader!css-loader!video.js/dist/video-js.css')` to inject video.js CSS.
```
BrandonOCasey: I would add "Be sure to" on the front of this sentence to make it a bit more readable.

docs/guides/react.md
```diff
@@ -0,0 +1,61 @@
...
+It just instantiate the video.js player on `componentDidMount` and destroy it on `componentWillUnmount`.
+
+```jsx
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import videojs from 'video.js'
+
+// this loads video.js CSS using webpack loaders
+require('!style-loader!css-loader!video.js/dist/video-js.css')
```
BrandonOCasey: I think we may want to exclude this line and the comment above as it is more webpack specific

docs/guides/react.md
```diff
@@ -0,0 +1,61 @@
+# video.js and ReactJS integration
+
+Here's a basic ReactJS player implementation.
+
+It just instantiate the video.js player on `componentDidMount` and destroy it on `componentWillUnmount`.
```
BrandonOCasey: * ~~instantiate~~ instantiates
* ~~destroy~~ destroys

docs/guides/react.md
```diff
@@ -0,0 +1,61 @@
...
+    return (
+      <div data-vjs-player>
+        <video className="video-js"></video>
+      </div>
+    )
+  }
+}
+```
+
+You can then use it like this :
```
BrandonOCasey: remove the extra space before `:`

revolunet:
Thanks. i added a footnote in react.md about webpack+css.

docs/guides/react.md
```diff
@@ -0,0 +1,62 @@
...
+```jsx
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import videojs from 'video.js'
+
+export default class VideoPlayer extends React.Component {
+  componentDidMount() {
+    // get the <video> DOM node
+    const videoNode = ReactDOM.findDOMNode(this).querySelector('video');
```
chemoish: Thoughts on `<video ref={(c) => { this.videoNode = c; }} />` instead of `ReactDOM`?

docs/guides/react.md
```diff
@@ -0,0 +1,62 @@
...
+    type: 'video/mp4'
+  }]
+}
+
+return <VideoPlayer { ...videoJsOptions } />
+```
+
+If you use webpack, you can embed video.js CSS like this:
+
+`require('!style-loader!css-loader!video.js/dist/video-js.css')`
```
chemoish: I find this line more confusing than anything.

If they are using `Webpack`, then they already know about `loaders`, otherwise they won't get very far. 

If they already know about `loaders`, then they must also know the power they have to transform files (https://webpack.js.org/guides/code-splitting-css/).

You don't need to have `Webpack` to include styles and recommending the inline approach might be less than ideal?

:two: cents.

docs/guides/react.md
revolunet: Not sure what is "best" way, but you're right, using `ref` is more straightforward

docs/guides/react.md
revolunet: well, now that the tooling is stable and mainstream (create-react-app, nwb...) i'm not sure that everyone master webpack loaders config. also the path to the CSS is not obvious so i think its worth mentionning ?

revolunet: for people using browserify or something else, we could also mention alternatives ?

gkatsev: What about just mentioning that you can find the css file at that path and not mention bundler specific stuff?
We also link to it in the [`package.json`](https://github.com/videojs/video.js/blob/master/package.json#L6)

fgarcia:
Now I think the most practical solution is avoiding tooling references and
just mention the problem: "dont forget your css"

docs/guides/react.md
revolunet: ok, removed the mention to webpack 😢

PR #3979, opened 2017-01-25
Expose Tech#resize as Player#resize

mister-ben:
## Description
The HTML5 tech has a [resize](https://html.spec.whatwg.org/multipage/embedded-content.html#event-media-resize) which fires when the video's intrinsic dimensions change. ~`Player` has a `resize` event when the player dimensions change.~ This can be confusing, and the tech's `resize` is useful.

`Player` had a `resize` event, but currently does not. 

## Specific Changes proposed
~Trigger a `videoresize` event on `Player` when `Tech` emits `resize`. The existing `Player#resize` does not change.~

Retrigger tech `resize` on `Player`.

## Requirements Checklist
  - [x] Docs/guides updated (jsdoc)

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -192,13 +192,13 @@ const TECH_EVENTS_RETRIGGER = [
   'ratechange',
 
   /**
-   * Fires when the playing speed of the audio/video is changed
+   * Fires when the video's intrinsic dimensions change
    *
    * @event Player#resize
    * @type {event}
    */
   /**
-   * Retrigger the `ratechange` event that was triggered by the {@link Tech}.
+   * Retrigger the `resize` event that was triggered by the {@link Tech}.
    *
    * @private
    * @method Player#handleTechResize_
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -192,6 +192,22 @@ const TECH_EVENTS_RETRIGGER = [
   'ratechange',
 
   /**
+   * Fires when the playing speed of the audio/video is changed
+   *
+   * @event Player#resize
+   * @type {event}
+   */
+  /**
+   * Retrigger the `ratechange` event that was triggered by the {@link Tech}.
+   *
+   * @private
+   * @method Player#handleTechResize_
+   * @fires Player#resize
+   * @listens Tech#resize
+   */
+  'resize',
+
+  /**
    * Fires when the volume has been changed
    *
    * @event player#volumechange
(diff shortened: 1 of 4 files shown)


src/js/player.js
```diff
@@ -880,6 +880,29 @@ class Player extends Component {
     this.on(this.tech_, 'posterchange', this.handleTechPosterChange_);
     this.on(this.tech_, 'textdata', this.handleTechTextData_);
 
+    /**
+     * Fires when the video's intrinsic width and/or height has changed.
+     * This differs from Player#resize, which is triggered when the player
+     * is resized.
+     *
+     * @event player#videoresize
```
BrandonOCasey: I think `player#videoresize` will have to be `Player#videoresize` with a capital P.

src/js/tech/html5.js
```diff
@@ -848,6 +848,7 @@ Html5.Events = [
   'play',
   'pause',
   'ratechange',
+  'resize',
```
gkatsev: this actually made me think maybe we should rename the dimension resize event on the player?
Some people (I know I have) have done `player.on(Html5.Events, () =>)` but the `resize` won't be the correct one.

I propose we change the player-level resize to `dimensionchange` and bring `resize` through unchanged. @videojs/core-committers thoughts?

heff:
Makes sense to me, though `dimensionchange` is interchangeable with `resize` so kind of ambiguous. Maybe `playerresize`, in the same vein as `playerready`?

src/js/tech/html5.js
mister-ben: Makes sense. I agree with @heff on the ambiguity, `playerresize` sounds good to me.

gkatsev: Sounds good to me.

mister-ben:
Updated to just pass through `resize` modified.

Although `Component.dimension()` triggers `resize`, `Player.dimension()` does not, so there's no `resize` event on the player to rename. Ought there to be?

gkatsev:
Seems we want to change what component triggers as well. Maybe generic components will trigger `componentresize` and then player would do `playerresize`?

mister-ben:
What should trigger playerresize on player? The way components work, it would only be triggered within `player.dimension(w, h)` which isn't that useful. Should we try to track the player size and trigger if it changes in `addClass` and `updateStyleEl_()`? That would still miss resizes from external causes like a parent element being resized.

gkatsev:
I saw a simple module recently that implemented proper resize handling by creating  hidden iframe the size of the element and listening and redispatching the resize event from it.

gkatsev:
@mister-ben also, I think it's fine for this PR to just make `playerresize` triggered where it currently does. We can make it trigger on other resizes in another PR.

mister-ben:
I'm pretty sure player hasn't triggered `resize` for a while - I can't find anywhere that triggers it in master, so this should probably go in as is. Also rebased.

gkatsev:
Wouldn't player trigger `resize` because component triggers resize?

mister-ben:
Component only triggers `resize` in `Component.dimension()` https://github.com/videojs/video.js/blob/master/src/js/component.js#L888

`Player.dimension()` does not
https://github.com/videojs/video.js/blob/master/src/js/player.js#L614

gkatsev:
Ah, cool. Yeah, I guess this is good to go then.

PR #3958, opened 2017-01-19
Fix CSS for Accessible Volume Control & IE <= 9

BrandonOCasey:
## Description
* Fix IE8/IE9 styles for volume control
* Fix IE8/IE9 styles for remaining time display (Fixes #3952)
* ~~This PR is based upon #3935~~ (Merged)
* Fixes IE10 transitions which were not implemented previously

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [x] If necessary, more likely in a feature request than a bug fix
  - Change has been verified in:
    - [x] Safari
    - [x] Chrome
    - [x] Firefox
    - [x] IE8
    - [x] IE9
    - [x] IE10
    - [x] IE11
    - [x] Edge
- [ ] Reviewed by Two Core Contributors

PR #3981, opened 2017-01-25
move most volume panel functionality into css state

gkatsev:
## Description
Built from https://github.com/videojs/video.js/pull/3958
Moves as much transition and "logic" into CSS as possible.
Use some javascript to make sure that things are sized properly when tabbing with the keyboard.

### DIFF
--- a/src/js/control-bar/volume-panel.js
+++ b/src/js/control-bar/volume-panel.js
@@ -46,16 +46,20 @@ class VolumePanel extends Component {
     // hide this control if volume support is missing
     checkVolumeSupport(this, player);
 
-    // when the mouse leaves the VolumePanel area hide the VolumeControl (slider/bar)
-    this.on(['mouseenter', 'touchstart'], () => this.volumeControl.show());
-    this.on(['mouseleave', 'touchend'], () => this.volumeControl.hide());
+    // while the slider is active (the mouse has been pressed down and
+    // is dragging) or in focus we do not want to hide the VolumeBar
+    this.on(this.volumeControl, ['slideractive'], this.sliderActive_);
+    this.on(this.muteToggle, 'focus', this.sliderActive_);
 
-    // when any child of the VolumePanel gets or loses focus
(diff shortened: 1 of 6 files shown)


BrandonOCasey:
I am not seeing the `VolumeControl`/`VolumeBar` being hidden on Firefox/Chorme

gkatsev:
As in if you're just adding the volume control directly? Those are always shown unless when part of the volume panel.

BrandonOCasey:
seems like it was a cache issue, I see it working now

src/js/control-bar/volume-panel.js
```diff
@@ -45,16 +45,20 @@ class VolumePanel extends Component {
...
-    // show/hide the VolumeControl (slider/bar)
-    this.children().forEach((child) => {
-      this.on(child, ['focus'], () => this.volumeControl.show());
-      this.on(child, ['blur'], () => this.volumeControl.hide());
-    });
+    this.on(this.volumeControl, ['sliderinactive'], this.sliderInactive_);
+    this.on(this.muteToggle, 'blur', this.sliderInactive_);
+  }
+
+  sliderActive_() {
```
BrandonOCasey: These will need jsdoc comments or maybe these should just be local functions?

BrandonOCasey:
I am not seeing the horizontal `VolumeControl`/`VolumeBar` hide on `IE9 Windows 7`(HTML) or `IE8 Windows 7` (Flash)

gkatsev:
Horizontal volume control isn't hidden on IE8 and 9. Always expanded.

BrandonOCasey:
ah ok well then other than my comment it LGTM

BrandonOCasey:
Actually one last thing, I am not seeing the vertical volume bar hide on IE8. Works on IE9

BrandonOCasey:
@gkatsev tested, the only issue as you said is the missing ball on IE8. I think thats acceptable.

PR #3983, opened 2017-01-26
fix `width:auto` on ie8 & ie9 for remaining-time-display

BrandonOCasey:
## Description
fix `width:auto` on ie8 & ie9 for remaining-time-display

## Requirements Checklist

### DIFF
--- a/src/css/_utilities.scss
+++ b/src/css/_utilities.scss
@@ -115,3 +115,10 @@
   @extend %fill-parent;
   text-align: center;
 }
+
+// don't wrap whitespace for the time control and make the text container 0 width
+// fixes a bug with 'width: auto' in IE 8/9
+.vjs-ie-auto-width-fix {
+  width: 0px !important;
+  white-space: nowrap;
+}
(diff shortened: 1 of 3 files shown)


src/css/_utilities.scss
```diff
@@ -115,3 +115,10 @@
   @extend %fill-parent;
   text-align: center;
 }
+
+// don't wrap whitespace for the time control and make the text container 0 width
+// fixes a bug with 'width: auto' in IE 8/9
+.vjs-ie-auto-width-fix {
```
gkatsev: Thinking about this a bit more. Can we re-use the vjs-no-flex class that gets added to the player div? It'll definitely get added in IE8 and 9.

BrandonOCasey: I will test that

BrandonOCasey: seems like `vjs-no-flex` does not fix the width problem

gkatsev: Sorry, I meant, can we rely on the vjs-no-flex class rather than introducing another one? i.e., move the properties from this selector into that?

PR #3985, opened 2017-01-26
Make `registerTech` add that tech to the default `techOrder`

BrandonOCasey:
## Description
Make `registerTech` add that tech to the default `techOrder`

## Requirements Checklist

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -802,6 +802,9 @@ class Tech extends Component {
     middleware.use('*', {name, tech});
 
     Tech.techs_[name] = tech;
+    if (name !== 'Tech') {
+      Tech.defaultTechs_.push(name.toLowerCase());
+    }
     return tech;
   }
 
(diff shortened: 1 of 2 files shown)


src/js/tech/tech.js
```diff
@@ -802,6 +802,9 @@ class Tech extends Component {
     middleware.use('*', {name, tech});
 
     Tech.techs_[name] = tech;
+    if (name !== 'Tech') {
+      Tech.defaultTechs_.push(name.toLowerCase());
```
gkatsev: techOrder expects them to be in camel case.

BrandonOCasey:
@gkatsev I changed it to convert to `camelCase` but I am wondering if we can't just add it to TechOrder as is. Seems like we call `toTitleCase` on `techOrder` when we use it anyway. Maybe we should even make the `techOrder` search case insensitive? That last idea seems like it might be a bit out of scope here though.

gkatsev:
We should just do what we did for components where it doesn't care whether it's camel or title cased.

BrandonOCasey:
Do we want to do that in this PR? if not I will make another to do it so we can just get this one merged to get `videojs-flash` RC.0 out

gkatsev:
I guess doing it in another PR would be cleaner

BrandonOCasey:
#3986

PR #3987, opened 2017-01-26
fix: allow changing volume in full height of volume control

gkatsev:
This makes it so you don't have to be precise in changing the volume.

One thing I'm not sure is whether I should change the cursor for the entirety of the volume control. I guess I probably should.

I'll be making another PR soon to allow similar functionality on the progress control.

### DIFF
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -20,6 +20,7 @@
 }
 
 .video-js .vjs-volume-control {
+  cursor: pointer;
   margin-right: 1em;
   @include display-flex;
 }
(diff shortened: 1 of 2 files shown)


gkatsev:
This has a small issue where if you drag out of the volume control, it'll hide but I think that should be fixed after #3981 has been merged. I could either rebase this and update here or make a separate PR for the lock showing aspect then.

misteroneill:
I think the cursor ought to reflect the interactive state. So, if you're hovering over something you can interact with, it should change.

gkatsev:
The issue that's failing the build is the sass issue that was just fixed by #4001, so, I'm just going to merge this in.

PR #3991, opened 2017-01-27
fix: updating time tooltips when player not in DOM

gkatsev:
If the time tooltips try to update when the player isn't in the DOM,
they error out because the bounding rects for the player will not be available.
This does a null check and exists early.


gkatsev:
Looks like I should investigate IE8 more, since 3 restarts didn't work.

gkatsev:
waiting the weekend and restarting tests unstuck it and it finally passed, hurray.

Issue #3890, opened 2016-12-23
Subtitles Button does not hide after remove subtitles tracks


mctep:
According to https://github.com/videojs/video.js/pull/3847

mctep:
@gkatsev What to do with [ChaptersButton@update](https://github.com/videojs/video.js/blob/4fd902229fc1bd39f5abd8b033cfaf5a96c4de2b/src/js/control-bar/text-track-controls/chapters-button.js#L55)? As I understand there is should not be the logic with threshold in `ChaptersButton`. So we cannot simple move update method from `CaptionsButton` to `TextTrackButton` I think.

May be It better to make `createItems` for each button to return empty array in a specific cases? 
And I do not understand why `threshold` sets from 2 to 1 when [featuresNativeTextTracks]( https://github.com/videojs/video.js/blob/4fd902229fc1bd39f5abd8b033cfaf5a96c4de2b/src/js/control-bar/text-track-controls/captions-button.js#L58) enabled.

Other way I see it make yet another `TextTrackButtonWithThreshold` 😄  class.
Or set some `option` in each button constructors like `option.threshold` and use it in common `TextTrackButton.update` method. But it is not clear what to do with `featuresNativeTextTracks`.

gkatsev:
Maybe a threshold of -1 disables this for the chapters button. I would prefer not to add another class :P

The reason why captions has a threshold of 1 for native and 2 for emulated is that we add an extra item to the menu for emulated captions for caption settings.

mctep:
Good catch about `threshold = -1` 👍 

My third way with constructors is wrong because we have not info about tech in it and tech could be changed in player lifetime.

What do you think about this way:

```js
class SomeButton extends TextTrackButton {
  update() {
    this.threshold = someSpecificThresholdForThisButton();
    super.update();
  }
}
```

Or make `getMenuItemsTreshold` for `TextTrackButton` and redefine it in any `SomeButton`.

Or rewrite `createItems` 😄  that could return empty array without any additional buttons and any threshold logic when there are no tracks in track list. Why not?

justinanastos:
This might be a totally naive suggestion, but it worked when modifying the source code locally. I did not think of edge cases, but this seems to work by checking to see if there one item left in the list and if it is an `instanceof OffTextTrackMenuItem`

#3993

justinanastos:
I believe this issue has been resolved by #3993

gkatsev:
[...]

PR #3993, opened 2017-01-27
fix(MenuButton) Implement unified method of showing/hiding MenuButton

justinanastos:
## Description
When remote subtitle text tracks are removed and there are none left, the subtitle button button persists in the menu bar. This also applies to the descriptions button as seen in #3994.

Reduced test case created by @mctep: http://jsbin.com/navoruk/1/edit?html,js,output

Addresses #3890, #3994

## Specific Changes proposed

- Implement a `getHideThreshold` function in `menu-button` and override it in `captions-button`.
- Remove `menu-button` descendants' update methods

## Requirements Checklist

### DIFF
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -58,9 +58,9 @@ class MenuButton extends ClickableComponent {
     this.buttonPressed_ = false;
     this.el_.setAttribute('aria-expanded', 'false');
 
-    if (this.items && this.items.length === 0) {
+    if (this.items && this.items.length <= this.hideThreshold_) {
       this.hide();
-    } else if (this.items && this.items.length > 1) {
+    } else {
       this.show();
     }
   }
@@ -74,6 +74,16 @@ class MenuButton extends ClickableComponent {
   createMenu() {
     const menu = new Menu(this.player_);
 
+    /**
+     * Hide the menu if the number of items is less than or equal to this threshold. This defaults
+     * to 0 and whenever we add items which can be hidden to the menu we'll increment it. We list
+     * it here because every time we run `createMenu` we need to reset the value.
+     *
+     * @protected
+     * @type {Number}
+     */
+    this.hideThreshold_ = 0;
+
     // Add a title list item to the top
     if (this.options_.title) {
       const title = Dom.createEl('li', {
@@ -81,6 +91,8 @@ class MenuButton extends ClickableComponent {
         innerHTML: toTitleCase(this.options_.title),
         tabIndex: -1
       });
+
+      this.hideThreshold_ += 1;
 
       menu.children_.unshift(title);
       Dom.insertElFirst(title, menu.contentEl());
--- a/src/js/control-bar/text-track-controls/captions-button.js
+++ b/src/js/control-bar/text-track-controls/captions-button.js
@@ -40,33 +40,6 @@ class CaptionsButton extends TextTrackButton {
   }
 
   /**
-   * Update caption menu items
-   *
-   * @param {EventTarget~Event} [event]
-   *        The `addtrack` or `removetrack` event that caused this function to be
-   *        called.
-   *
-   * @listens TextTrackList#addtrack
-   * @listens TextTrackList#removetrack
-   */
-  update(event) {
-    let threshold = 2;
-
-    super.update();
-
-    // if native, then threshold is 1 because no settings button
-    if (this.player().tech_ && this.player().tech_.featuresNativeTextTracks) {
-      threshold = 1;
-    }
-
-    if (this.items && this.items.length > threshold) {
-      this.show();
-    } else {
-      this.hide();
-    }
-  }
-
-  /**
    * Create caption menu items
    *
    * @return {CaptionSettingsMenuItem[]}
@@ -77,6 +50,8 @@ class CaptionsButton extends TextTrackButton {
 
     if (!(this.player().tech_ && this.player().tech_.featuresNativeTextTracks)) {
       items.push(new CaptionSettingsMenuItem(this.player_, {kind: this.kind_}));
+
+      this.hideThreshold_ += 1;
     }
 
     return super.createItems(items);
--- a/src/js/control-bar/text-track-controls/text-track-button.js
+++ b/src/js/control-bar/text-track-controls/text-track-button.js
@@ -40,6 +40,7 @@ class TextTrackButton extends TrackButton {
   createItems(items = []) {
     // Add an OFF menu item to turn all tracks off
     items.push(new OffTextTrackMenuItem(this.player_, {kind: this.kind_}));
+    this.hideThreshold_ += 1;
 
     const tracks = this.player_.textTracks();
 


src/js/menu/menu-button.js
```diff
@@ -58,7 +59,7 @@ class MenuButton extends ClickableComponent {
     this.buttonPressed_ = false;
     this.el_.setAttribute('aria-expanded', 'false');
 
-    if (this.items && this.items.length === 0) {
+    if (this.items && (this.items.length === 0 || (this.items.length === 1 && this.items[0] instanceof OffTextTrackMenuItem))) {
```
OwenEdwards: I don't this should be handled in the `MenuButton` component; it should be in the sub-class component `CaptionsButton`, using the [`update`](https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/captions-button.js#L52-L67) method which explicitly understands the different `threshold` if the OffTextTrackMenuItem is included.

OwenEdwards: Oh, my mistake - I was thinking of the `CaptionsSettings` menu item, not the `OffTextTrackMenuItem`.

justinanastos:
@OwenEdwards Any more feedback?

OwenEdwards:
@justinanastos I'm still confused why this happens on the `subtitles` menu button, and not on the `captions` menu button, but you're fixing it on the generic `menuButton`. Can you give me a little more explanation of why that is?

justinanastos:
@OwenEdwards In trying to explain why I implemented this the way I did, I realized there was a smarter way of handing it.

First, when using emulated text tracks, video.js will add a "captions settings" menu item, which is an instance of `OffTextTrackMenuItem`:

captions menu button

subtitles menu button

When removing the caption button, video.js will use a threshold for how many menu items exist for showing or hiding the button in [`control-bar/text-track-controls/subtitles-button`](https://github.com/videojs/video.js/blob/5.x/src/js/control-bar/text-track-controls/captions-button.js#L42-L67):

```javascript
/**
 * Update caption menu items
 *
 * @param {EventTarget~Event} [event]
 *        The `addtrack` or `removetrack` event that caused this function to be
 *        called.
 *
 * @listens TextTrackList#addtrack
 * @listens TextTrackList#removetrack
 */
update(event) {
  let threshold = 2;

  super.update();

  // if native, then threshold is 1 because no settings button
  if (this.player().tech_ && this.player().tech_.featuresNativeTextTracks) {
    threshold = 1;
  }
  if (this.items && this.items.length > threshold) {
    this.show();
  } else {
    this.hide();
  }
}
```

---

A more appropriate solution would be to add a similar function in [`control-bar/text-track-controls/subtitles-button`](https://github.com/videojs/video.js/blob/5.x/src/js/control-bar/text-track-controls/subtitles-button.js#L40)

---

This PR has been updated to add an `update` function to [`control-bar/text-track-controls/subtitles-button`](https://github.com/videojs/video.js/blob/5.x/src/js/control-bar/text-track-controls/subtitles-button.js#L40). A "subtitles settings" menu item is never added for native or emulated subtitles, so the threshold will always be `1`.

OwenEdwards:
@justinanastos okay, that seems like a cleaner fix. The problem is, this also affect the Descriptions button (see #3994). If you look at [the update method in MenuButton](https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js#L78-L82), you'll see that it hides the menu if there are zero menu items, shows the menu if there are **more than one items**, but **doesn't explicitly do anything if there is just one item**!

So you need to look at which other menus extend the MenuButton (TrackButton, TextTrackButton, etc.), and see how far up the hierarchy you can push this change so that it affects all the menus which need fixing (e.g. does the AudioTrackButton need fixing?) without you having to put this fix in multiple places. @gkatsev may have some thoughts on that, too.

gkatsev:
In #3890, I was thinking that we should make the threshold a generic thing for text track buttons and maybe track buttons in general. That way, captions button can set its own threshold and subtitles could use the default. And then a value of -1 or something could disable this hiding.

This PR solves the specific issue with subtitles, so, maybe it might be worth pulling it in anyway, but I think we should make it generic that way we don't have the same code in the various track buttons.

gkatsev:
Given that @mister-ben did a similar thing in #4028, I'd be inclined to pull this in, and then making the more general fix later.

OwenEdwards:
Hmm, I'm just concerned that the `Descriptions` button, which is essentially *identical* to the `subtitles` button except with a different track `kind` and button `controlText` is going to be the orphaned control if we don't fix the identical issue in both of them by moving this fix up the hierarchy.

gkatsev:
I don't disagree. However, we have a fix for subtitles, so, why not get it fixed?

OwenEdwards:
I guess I'm a fan if "do it right the first time" :-) If we pull this patch, we'll either want to undo it later and apply it higher up the hierarchy to fix #3994, or it'll get copied into `DescriptionsButton`, and that way lies madness ;-)

OwenEdwards:
@gkatsev I guess another way to look at this is - was it intentional, or is it a bug, that in the `update` method for the core component `MenuButton`, it hides if there are no items, shows if there are more than one items, but *does nothing if there is exactly one item*? If it's a bug, it should be fixed (I'm having a hard time imagining a use case for a menu with just one item, but I guess it's possible), no?

gkatsev:
The default menu button should probably should for one and more and then TextTracks buttons need the threshold fix applied to it as a general solution, which each type of text track specifying if it needs a higher threshold (like captions button sometimes needing a threshold of 2). AudioTrackButton could be shown or not with only one item.

Looks like MenuButton has been around like that for a long time.

justinanastos:
@gkatsev I didn't fully understand what you meant in #3890 until now. This makes total sense and would be pretty simple to implement. Let me give it a shot.

gkatsev:
Cool, that would be awesome, @justinanastos. Let me know if you have any questions, feel free to drop by our slack (http://slack.videojs.com), too.

justinanastos:
@gkatsev I did not implement the `-1` behavior because, as far as I can tell, it won't be used anywhere. All menus appear to have a title, so we'll never actually want to show the menu with no content, right?

src/js/menu/menu-button.js
```diff
@@ -37,6 +37,15 @@ class MenuButton extends ClickableComponent {
   }
 
   /**
+   * Hide the menu if the number of items is less than or equal to this threshold. This defaults
+   * to 1 beacuse all menus have a title.
+   * @type {Number}
+   */
+  get hideThreshold() {
```
gkatsev: these should be regular methods rather than getters.

justinanastos: Updated @gkatsev .

src/js/menu/menu-button.js
```diff
@@ -37,6 +37,16 @@ class MenuButton extends ClickableComponent {
   }
 
   /**
+   * Hide the menu if the number of items is less than or equal to this threshold. This defaults
+   * to 1 beacuse all menus have a title.
```
OwenEdwards: Not all menus have a title - the Subtitles menu and the Descriptions menu have an "off" button, which should be hidden if it is the only item. <Captain Pedantic ;-) >

justinanastos: Updated @OwenEdwards

src/js/menu/menu-button.js
```diff
@@ -37,6 +37,16 @@ class MenuButton extends ClickableComponent {
   }
 
   /**
+   * Hide the menu if the number of items is less than or equal to this threshold. This defaults
```
OwenEdwards: Ultimately I'd love to see each menu component actually set their threshold as they add items to the menu which can be hidden (title, off button, caption settings button, etc), but this looks like it will do for now, and covers both #3890 and #3994.

gkatsev: You're right. Maybe just making it be a private instance variable would be easier in the long run.
At least should be a private method.

@justinanastos thoughts?

OwenEdwards: That would avoid the multiple uses of ` if (this.player().tech_ && this.player().tech_.featuresNativeTextTracks) {` in this component - just increment the threshold count at the same time it adds the Captions Settings Menu Item.

justinanastos: I'll give this idea a shot.

justinanastos:
@OwenEdwards and @gkatsev : I updated the logic to use a private instance variable that is incremented when menus are added.

src/js/menu/menu-button.js
```diff
@@ -29,6 +29,14 @@ class MenuButton extends ClickableComponent {
 
     this.update();
 
+    /**
+     * Hide the menu if the number of items is less than or equal to this threshold. This defaults
+     * to 0 and whenever we add items to the menu we'll increment it.
+     *
+     * @type {Number}
+     */
+    this.hideThreshold_ = 0;
```
gkatsev: this should probably move above `this.update()`. `createItems` gets called as part of `createMenu` which is called in `update`, so, we need to make sure that this threshold is initialized properly.

src/js/menu/menu-button.js
```diff
@@ -29,6 +29,14 @@ class MenuButton extends ClickableComponent {
 
     this.update();
 
+    /**
+     * Hide the menu if the number of items is less than or equal to this threshold. This defaults
+     * to 0 and whenever we add items to the menu we'll increment it.
+     *
+     * @type {Number}
```
gkatsev: does this need to be marked `@private`?

OwenEdwards:
This looks *much* better! Once you've made the changes @gkatsev requested, and fixed the tests, it'll be a really clean solution to the various problems.

src/js/menu/menu-button.js
justinanastos: Yes it does @gkatsev

justinanastos: Correction @gkatsev : It should be `@protected` because descendants should be allowed access.

gkatsev: Ah, didn't realize that existed.

justinanastos: Me neither, I googled where to put the `@private` tag and saw it.

src/js/menu/menu-button.js
```diff
@@ -29,6 +29,14 @@ class MenuButton extends ClickableComponent {
 
     this.update();
 
+    /**
+     * Hide the menu if the number of items is less than or equal to this threshold. This defaults
+     * to 0 and whenever we add items to the menu we'll increment it.
```
OwenEdwards: If I'm going to be pedantic, this ought to be "... and whenever we add items *which can be hidden* to the menu ..." I don't want future additions to the code to increment this when they add non-hideable items!

gkatsev:
@justinanastos can we get a version of this against master as well? If you don't have a chance to do it, let us know and we can port it over for you.

justinanastos:
Sure thing @gkatsev

PR #3996, opened 2017-01-30
Let font import be relative if possible

aequasi:
## Description
This would allow the font import to pull from the node_modules directory relatively, instead of absolutely, for projects that use a bundler, and bundle this on their own 

## Specific Changes proposed
Changed import to use `~`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -2,7 +2,7 @@
 @import "private-variables";
 @import "utilities";
 
-@import "node_modules/videojs-font/scss/icons";
+@import "~videojs-font/scss/icons";
 
 @import "components/layout";
 @import "components/big-play";


gkatsev:
Is the `~` a known sass thing?

aequasi:
thats a good question actually. It might only be a webpack thing. Can we make this import configurable? like fontawesome does

gkatsev:
Yup, looks like it's a webpack thing. Also, apparently, it's currently broken (see #3998).
I'm going to close this out for now.

As for making that configurable, that's probably fine as long as the default build works correctly :). Are you importing the sass directly over the pre-built files?

gkatsev:
Might be better to move the discussion into a "make font @import configurable" issue.

PR #4004, opened 2017-01-31
feat: allow seeking in full height of progress control

gkatsev:


src/js/control-bar/progress-control/progress-control.js
```diff
@@ -28,6 +29,9 @@ class ProgressControl extends Component {
     super(player, options);
     this.handleMouseMove = Fn.throttle(Fn.bind(this, this.handleMouseMove), 25);
     this.on(this.el_, 'mousemove', this.handleMouseMove);
+
+    this.throttledHandleMouseMove = throttle(bind(this, this.handleMouseSeek), 25);
```
BrandonOCasey: maybe it would be a good idea to add a doc comment for this method, or even add it to the class and have it do the same thing it does now.

gkatsev:
@BrandonOCasey I renamed the method to `throttledHandleMouseSeek` to match with `handleMouseSeek` and I added a jsdoc comment to it.

PR #4005, opened 2017-01-31
feat: toggle playback with space when focused on seekbar

gkatsev:


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -166,6 +166,44 @@ class SeekBar extends Slider {
...
+  handleKeyPress(event) {
+
+    // Support Space (32) or Enter (13) key operation to fire a click event
+    if (event.which === 32 || event.which === 13) {
+      event.preventDefault();
+      this.handleAction(event);
+    } else if (super.handleKeyPress) {
+
+      // Pass keypress handling up for unsupported keys
+      super.handleKeyPress(event);
```
OwenEdwards: My thought on this was that unhandled keypresses by each control bar element would bubble up here to higher components; for example, if Space isn't supported by a control bar component, it would bubble all the way up to the control bar itself, which would interpret it as 'toggle playback state'. That way, you wouldn't need to put this kind of handling into each sub-component, and the control bar itself could implement the full set of keyboard shortcuts that are becoming the de-facto standard for YouTube (Space, Up Arrow, Down Arrow, Home, End, 0..9, F, M, etc.) But I realize this isn't what's happening here - it's bubbling up to the component this extends, not its structural parent. So maybe the root `Component` needs to pass it up to its parent component, if it has one?

gkatsev: Hm... yeah, that sounds like a good idea in the long term.
We should see if we can make that pluggable or if it's worth bringing it into core. There's already the hotkeys plugin: https://github.com/ctd1500/videojs-hotkeys

OwenEdwards: I remember that hotkeys plugin, but I don't think it handles pressing Space properly, given that Space should activate a button if a button has focus.

The structure could be to pass all unrecognized keypresses up to the control bar, or even all the way up to the player, and allow a plugin to decide what to do with them there?

gkatsev: Any objections to this PR with figuring out the rest at a later point?

OwenEdwards:
No objections. LGTM

Issue #3699, opened 2016-10-21
Title attribute value in closed-caption menu is broken


stsvilik:
## Description

When CC menu items are hovered by mouse, they display empty or invalid text.
## Steps to reproduce

Hover mouse over CC menu item to see title text appear.

In Chrome
1. Put player on pause
2. Inspect CC button and set :hover state to checked
3. Inspect menu item which appears on-hover (caption settings is good example)
4. Pay attention to "title" attribute of li tag
### Expected

Title text to match values in the items.
### Actual

Empty or invalid text appears.
### Error output

No.

This issue can be observed on Video.js homepage in "Funny or Die" demo.
#### videojs

5.11.8?
#### browsers

All
#### OSes

All

gkatsev:
Can you please elaborate on what you mean by invalid text?
I believe right now the title attribute for the captions menu items is set to only tell whether the item is selected or not.
I'm not sure it would make sense to have the title attribute be the text itself, seems a bit redundant. What do you think, @OwenEdwards?

stsvilik:
@gkatsev I would expect a title for say "english" track to say at least "english" instead of " " which looks like a small square blocking the content. As for "caption settings" item, title is ", opens captions settings dialog" do you find it correct? Where is comma coming from? To be honest I don't think title is necessary at all - who would ever wait long enough to see it show up?

gkatsev:
Yeah, we should potentially just remove it altogether. I think the comma is there for screenreaders, but not positive.

OwenEdwards:
This is the same issue that @chemoish mentioned in his [comment](https://github.com/videojs/video.js/issues/3447#issuecomment-244456398) on #3447.

When `title` attributes were added, they just pulled in the `controlText` value. This makes sense for controls which have an icon and no visible text, since the controlText is both what is presented by a screen reader and what is shown on mouse-hover as a tooltip.

For controls which have visible text (e.g. `MenuItem`s), to support screen reader users the controlText was set to additional information about the button which is required for accessibility. Unfortunately, this doesn't work with the way that the title text was implemented. It doesn't make sense to have any title attribute on a control with a text label - if the control isn't clear from the text itself, then it shouldn't rely on the title tooltip because that isn't shown for sighted keyboard-only users.

So either the setting of the title text needs to be changed, or controlText needs to be more clearly defined and potentially the components need a separate feature for screen reader specific text.

What do you think, @gkatsev and @chemoish?

hartman:
[...]

PR #4006, opened 2017-01-31
fix: hide font-icons from assitive technology

### LINKED ISSUE #3982: Font icons should be hidden from assistive technology (screen readers)
## Description

The font icons are inserted into the page using special characters in the CSS, e.g.:

````
.vjs-icon-play:before, .video-js .vjs-big-play-button:before, .video-js .vjs-play-control:before {
    content: "\f101";
}
````

giving:

````
<button class="vjs-play-control vjs-control vjs-button vjs-paused" type="button" aria-live="polite" title="Play" aria-disabled="false">
    <!-- Icon is inserted here in the DOM -->
    <span class="vjs-control-text">Play</span>
</button>
````

Screen readers include the `\f101` character in the name of the button, together with the `vjs-control-text`. No problems have been reported with this yet, but the extra character can be seen, for example, with VoiceOver on macOS, which displays the spoken text on the screen. This extra character will also be displayed on Refreshable Braille Displays, and could cause confusion.

## Recommendation

The font icon should be inserted in a `<span aria-hidden="true"></span>` inside the control, so that the special character is hidden from screen readers as purely decorative.

## Additional Information
#### videojs
v5.15
#### browsers
Browser and screen readers including IE11 & JAWS18, Firefox & NVDA, Safari & VoiceOver
#### OSes
Windows 7, macOS
### plugins
no.

gkatsev:
Wrap font-icons in a span with aria-hidden=true on it.

Fixes #3982.

### DIFF
--- a/src/css/components/_audio.scss
+++ b/src/css/components/_audio.scss
@@ -1,3 +1,3 @@
-.video-js .vjs-audio-button {
+.video-js .vjs-audio-button .vjs-icon-placeholder {
   @extend .vjs-icon-audio;
 }
--- a/src/css/components/_big-play.scss
+++ b/src/css/components/_big-play.scss
@@ -21,7 +21,7 @@
 
   // Since the big play button doesn't inherit from vjs-control, we need to specify a bit more than
   // other buttons for the icon.
-  &:before {
+  & .vjs-icon-placeholder:before {
     @extend %icon-default;
   }
 }
--- a/src/css/components/_big-play.scss
+++ b/src/css/components/_big-play.scss
@@ -17,11 +17,11 @@
   @include border-radius(0.3em);
   @include transition(all 0.4s);
 
-  @extend .vjs-icon-play;
-
   // Since the big play button doesn't inherit from vjs-control, we need to specify a bit more than
   // other buttons for the icon.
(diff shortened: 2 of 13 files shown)


OwenEdwards:
Doesn't work for the Big Play Button. Otherwise LGTM.

OwenEdwards:
@gkatsev, as a side note - given how simple these changes were (nice work!), I wonder if we should also put the `MenuItem` text in its own `<span class="???">`? It could have its own class name, and it **wouldn't** have `aria-hidden`, but may give us consistent DIV/SPAN structuring so that we can make changes in v6 without changing the DOM?

Come to think of it, we could also insert a `<span aria-hidden="true" class="vjs-icon-placeholder"></span>` at the start of the `MenuItem` too, which could be a placeholder in case we want to implement an icon and visible label text, as I mentioned in https://github.com/videojs/video.js/issues/3699#issuecomment-275264808, item 2?

gkatsev:
Adding the span is additive, so, we can always add it at a later date. However, putting the menuitem text into its own span sounds like a good idea.

gkatsev:
I opened up https://github.com/videojs/video.js/issues/4017 for the menuitem DOM structure changes

PR #4008, opened 2017-02-01
Change accessibility test in grunt.js to remove unnecessary warning message

OwenEdwards:
## Description
Minor change in the grunt file to remove unnecessary warning messages from the accessibility test

## Specific Changes proposed
(See above)

## Requirements Checklist

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -483,8 +483,8 @@ module.exports = function(grunt) {
           error: true
         },
         ignore: [
-          // Ignore the warning about needing <optgroup> elements
-          'WCAG2AA.Principle1.Guideline1_3.1_3_1.H85.2'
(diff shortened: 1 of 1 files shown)


OwenEdwards:
The example it tests does follow accessibility guidelines - it's just that automatic testing can't distinguish when one thing is turned off by Javascript, and another thing is turned on, so it thinks the two elements overlay each other. That's why it's flagged as a warning, not an error - it's recommending that this issue is checked manually. There isn't a standard way to flag to the checker that this is okay, and shouldn't cause a warning, so the easiest thing is to turn off that warning. Alternatively we could turn off all warnings, and just look for errors, but this gives better coverage than that approach.

gkatsev:
We'd want a similar fix against master.

PR #4010, opened 2017-02-01
feat: getTech now accepts `TitleCase` or `camelCase`

### LINKED ISSUE #3986: Make TechOrder accept `camelCase` and `TitleCase` tech names
## Description
Currently `techOrder` only accepts tech names that are `camelCase` others will be ignored. It would be much more user friendly and closer to other conventions if we accept `TitleCase` as well.

## Proposed Changes
* After #3985 is merged remove the `camelCase` conversion.
* Anywhere that we use techOrder should search for tech using `camelCase` or `TitleCase`. Most of there are in `player.js`

BrandonOCasey:
## Description
`getTech` now works with `TitleCase` and `camelCase` tech names

Fixes #3986
## Specific Changes proposed
* update techOptions to look for `TitleCase`/`camelCase` options
* remove all possible `toTitleCase` calls from player.js
* remove deprecated usage of Tech as Component
* add a unit test to verify that registerTech works
* change defaultTech_ to defaultTechOrder_

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -858,6 +858,10 @@ class Player extends Component {
     // Initialize tech instance
     const TechClass = Tech.getTech(techName);
 
+    if (!TechClass) {
+      throw new Error(`No Tech named '${titleTechName}' exists! '${titleTechName}' should be registered using videojs.registerTech()'`);
+    }
+
     this.tech_ = new TechClass(techOptions);
 
     // player.triggerReady is always async, so don't need this to be async
@@ -2147,9 +2151,6 @@ class Player extends Component {
     const techs =
       this.options_.techOrder
         .map((techName) => {
-          // `Component.getComponent(...)` is for support of old behavior of techs
-          // being registered as components.
-          // Remove once that deprecated behavior is removed.
           return [techName, Tech.getTech(techName)];
         })
         .filter(([techName, tech]) => {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -806,14 +806,17 @@ class Player extends Component {
       this.unloadTech_();
     }
 
+    const titleTechName = toTitleCase(techName);
+    const camelTechName = techName.charAt(0).toLowerCase() + techName.slice(1);
+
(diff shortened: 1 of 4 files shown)


src/js/player.js
```diff
@@ -840,6 +843,8 @@ class Player extends Component {
       techOptions[props.getterName] = this[props.privateName];
     });
 
+    assign(techOptions, this.options_[titleTechName]);
```
gkatsev: does `assign` handle undefineds?

src/js/player.js
```diff
@@ -851,14 +856,9 @@ class Player extends Component {
     }
 
     // Initialize tech instance
-    let TechComponent = Tech.getTech(techName);
+    const TechClass = Tech.getTech(techName);
 
-    // Support old behavior of techs being registered as components.
-    // Remove once that deprecated behavior is removed.
-    if (!TechComponent) {
-      TechComponent = Component.getComponent(techName);
```
gkatsev: I guess we should throw an error in this case?

BrandonOCasey: I will improve the error handing here

BrandonOCasey: I don't know if we show check `getComponent` at all. I think we should just throw an error if `TechComponent`/`TechClass` does not exist. We could check `getComponent` here but I think that we should be able to have a Tech and a Component with the same name (as weird as that would be)

src/js/player.js
BrandonOCasey: Going to add some unit tests to assign to find out, will open another PR with those

src/js/player.js
gkatsev: yeah, I meant to just throw if `TechClass` isn't available. We shouldn't try to call `getComponent` anymore.

src/js/player.js
BrandonOCasey: See #4014 for tests, seems that undefined/null/empty will do nothing

src/js/player.js
BrandonOCasey: Done

PR #4015, opened 2017-02-01
docs: fixup global members

gkatsev:
There are still two globals that I can't figure out how to fix properly:
* TextTrackCue typedef in the TextTrackCueList
* mergeOptions

Though, I don't think those two should hold this PR up.

### DIFF
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -91,7 +91,6 @@ class PlayToggle extends Button {
   /**
    * Add the vjs-ended class to the element so it can change appearance
    *
-   * @method handleEnded
    */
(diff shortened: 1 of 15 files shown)


BrandonOCasey:
`TextTrackCue` should probably be `TextTrackCueList~TextTrackCue` to keep it in that classes scope? I could be wrong on that though. LGTM otherwise

gkatsev:
I tried using that but let me try again.

BrandonOCasey:
If that change worked I approve this PR

gkatsev:
Yeah, I think so. It seems like it sometimes just gets cached or something so sometimes just refreshing the page doesn't seem to help. The change definitely doesn't hurt, though.

PR #4016, opened 2017-02-01
Add flash as a dev dependency

BrandonOCasey:
## Description
Also add an example of using flash in the sandbox

## Requirements Checklist
  - [x] Example created

### DIFF
--- a/sandbox/index.html.example
+++ b/sandbox/index.html.example
@@ -10,13 +10,19 @@
   <!-- Load the source files -->
   <link href="../build/temp/video-js.css" rel="stylesheet" type="text/css">
   <script src="../build/temp/video.js"></script>
+  <script src="../node_modules/videojs-flash/dist/videojs-flash.js"></script>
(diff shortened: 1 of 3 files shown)


sandbox/index.html.example
```diff
@@ -10,12 +10,6 @@
   <!-- Load the source files -->
   <link href="../build/temp/video-js.css" rel="stylesheet" type="text/css">
   <script src="../build/temp/video.js"></script>
-
-  <!-- Set the location of the flash SWF -->
```
gkatsev: what about just adding the flash tech to this sandbox? That way if you're fiddling with this file and then run it on IE8, it won't just stop working.

Having a flash example which prefers flash as the tech order sounds good either way.

BrandonOCasey: sure I will do that

BrandonOCasey: done

PR #4018, opened 2017-02-01
fix: focus play toggle from Big Play Btn on play

### LINKED ISSUE #2729: Movement of keyboard focus from the 'big play' button to the full control bar to support keyboard accessibility
A couple of changes would fix accessibility issues related to the 'big play' button:
1. When the big play button is activated using the keyboard, **focus should be moved** to the regular play button in the control bar
2. There should be some indication to screen reader users that only the big play button is available before playback is started, and that the full control bar will become available when play starts (or that it has become available once the big play button has been activated, but this is not an ideal solution because that is the point when audio from the video playback will start).

This is related to, but distinct from. #2156.

gkatsev:
If the control bar and playtoggle exist, focus the playtoggle after
triggering play via the keyboard from the Big Play Button.
If `play()` returns a promise, wait until it resolves to focus,
otherwise, use a setTimeout.

Fixes #2729

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -40,6 +40,7 @@ class BigPlayButton extends Button {
     const playToggle = cb && cb.getChild('playToggle');
 
     if (!playToggle) {
+      this.player_.focus();
       return;
     }
 
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -34,7 +34,22 @@ class BigPlayButton extends Button {
    * @listens click
    */
   handleClick(event) {
-    this.player_.play();
+    const playPromise = this.player_.play();
+
+    const cb = this.player_.getChild('controlBar');
+    const playToggle = cb && cb.getChild('playToggle');
+
+    if (!playToggle) {
+      return;
+    }
+
+    if (playPromise) {
+      playPromise.then(() => playToggle.focus());
+    } else {
+      this.setTimeout(function() {
+        playToggle.focus();
+      }, 1);
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/js/big-play-button.js
```diff
@@ -34,7 +34,22 @@ class BigPlayButton extends Button {
...
    */
   handleClick(event) {
-    this.player_.play();
+    const playPromise = this.player_.play();
+
+    const cb = this.player_.getChild('controlBar');
+    const playToggle = cb && cb.getChild('playToggle');
+
+    if (!playToggle) {
+      return;
```
OwenEdwards: If there isn't a playToggle control, focus ought to move to either the player's `<div>`, or the `<video>` element, rather than just falling into a black hole! I guess the `<video>` element won't be there if the player uses a Flash tech, so the `<div>` makes sense.

gkatsev: We can focus on the tech.

OwenEdwards: Strictly, focus ought to move to something which is actionable, but it can move to something which isn't if pressing Tab from there will get the user to an actionable control (rather than, for example, the next focusable element _after_ the video.js player)

gkatsev: Makes sense, I'll just focus the player element and we can always tweak it in the future if we think there's something better.

gkatsev: Updated to focus on the player element in this case.

richardbushell:
@gkatsev
As per #4039, could this be included in latest 5.x release too please...

gkatsev:
@OwenEdwards @richardbushell ported over in https://github.com/videojs/video.js/pull/4132/

richardbushell:
@gkatsev Terrific, thanks...

PR #4023, opened 2017-02-02
fix: accessibility bugs with the VolumeBar

### LINKED ISSUE #4021: aria-valuenow and aria-valuetext are not initially set correctly
## Description
The initial state of the volume slider has `aria-valuenow` set to 0, and no `aria-valuetext` attribute. As a result, a screen reader (e.g. VoiceOver on macOS) announces the slider as "0, volume level), even though the volume is set to (and displayed as) 100%. The `aria-valuetext` is not set until the volume is changed.

## Steps to reproduce
1. On a Mac, turn on VoiceOver (command-F5)
2. Open a page with the video.js player, in Safari
3. Use the Tab key to move keyboard focus to the Big Play Button, and press Enter
4. Use the Tab key to move focus into the Control Bar, and then onto the volume control slider

### Expected
Expect VoiceOver to announce (and display) that the volume is 100%

### Actual
VoiceOver announces "0, volume level"

### Error output
_None_

## Additional Information
#### videojs
v6.0.0-RC.1
#### browsers
Safari with VoiceOver (and probably others)
#### OSes
macOS (Sierra, 10.12.2)
### plugins
_No_

### LINKED ISSUE #4022: The aria-valuetext for the volume control slider contains two decimal places on a 0 to 100% scale!
## Description
Screen reader announces volume as, e.g. "eight-point-zero-zero-percent". Since a keyboard-only user can only move the volume control in steps of 10, the two decimal places of precision in `aria-valuetext` are completely unnecessary!

## Steps to reproduce
1. Use a screen reader, such as VoiceOver on macOS, to read out the value of the volume control slider!

## Additional Information
#### videojs
v6.0.0-RC.1
#### browsers
All
#### OSes
All
### plugins
_none_

BrandonOCasey:
## Description
* Fixes #4021: `aria-valuenow` and `aria-valuetext` not set initially
* Fixes #4022: drop the decimal places on `aria-valuenow` and `aria-valuetext`

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/volume-control/volume-bar.js
+++ b/src/js/control-bar/volume-control/volume-bar.js
@@ -27,7 +27,7 @@ class VolumeBar extends Slider {
     super(player, options);
     this.on('slideractive', this.updateLastVolume_);
     this.on(player, 'volumechange', this.updateARIAAttributes);
-    player.ready(() => this.updateARIAAttributes);
+    player.ready(() => this.updateARIAAttributes());
   }
 
   /**
@@ -106,7 +106,7 @@ class VolumeBar extends Slider {
    */
   updateARIAAttributes(event) {
     // Current value of volume bar as a percentage
-    const volume = (this.player_.volume() * 100).toFixed(2);
+    const volume = Math.round((this.player_.volume() * 100)).toString();
 
     this.el_.setAttribute('aria-valuenow', volume);
     this.el_.setAttribute('aria-valuetext', volume + '%');


src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -27,7 +27,7 @@ class VolumeBar extends Slider {
     super(player, options);
     this.on('slideractive', this.updateLastVolume_);
     this.on(player, 'volumechange', this.updateARIAAttributes);
-    player.ready(() => this.updateARIAAttributes);
+    player.ready(() => this.updateARIAAttributes());
```
gkatsev: heh

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -106,7 +106,7 @@ class VolumeBar extends Slider {
    */
   updateARIAAttributes(event) {
     // Current value of volume bar as a percentage
-    const volume = (this.player_.volume() * 100).toFixed(2);
+    const volume = Math.round((this.player_.volume() * 100).toFixed(2));
```
gkatsev: no need to the `.toFixed()`, that's what adds the two decimal places. Maybe just a `.toString()`

BrandonOCasey:
I wasn't even aware that GitHub could auto-close issues. I will have to use the suggested format more often.

src/js/control-bar/volume-control/volume-bar.js
BrandonOCasey: changed to toString()

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -106,7 +106,7 @@ class VolumeBar extends Slider {
    */
   updateARIAAttributes(event) {
     // Current value of volume bar as a percentage
-    const volume = (this.player_.volume() * 100).toFixed(2);
+    const volume = (this.player_.volume() * 100).toString();
```
gkatsev: I think we'd still want Math.round or Math.floor to make sure we are getting an integer.

BrandonOCasey: Done

PR #4025, opened 2017-02-02
feat: modal dialog accessibility updates

gkatsev:
If the modal dialog was opened and the focus was preset inside the
player, move the focus to the modal dialog.
When the modal dialog is closed, move the focus back to the previously
active element.
When focus is inside the dialog, trap tab focus. This was inspired by https://github.com/gdkraus/accessible-modal-dialog and ally.js

Still needs:
* [x] unit tests

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1013,20 +1013,6 @@ class Component {
   }
 
   /**
-   * Set the focus to this component
-   */
-  focus() {
-    this.el_.focus();
-  }
-
-  /**
-   * Remove the focus from this component
-   */
-  blur() {
-    this.el_.blur();
-  }
-
-  /**
    * Emit a 'tap' events when touch event support gets detected. This gets used to
    * support toggling the controls through a tap on the video. They get enabled
    * because every sub-component would have extra overhead otherwise.
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1013,6 +1013,20 @@ class Component {
   }
 
   /**
+   * Set the focus to this component
+   */
+  focus() {
+    this.el_.focus();
+  }
+
+  /**
+   * Remove the focus from this component
+   */
+  blur() {
+    this.el_.blur();
+  }
+
+  /**
    * Emit a 'tap' events when touch event support gets detected. This gets used to
    * support toggling the controls through a tap on the video. They get enabled
    * because every sub-component would have extra overhead otherwise.
--- a/src/js/modal-dialog.js
+++ b/src/js/modal-dialog.js
@@ -4,6 +4,7 @@
 import * as Dom from './utils/dom';
 import * as Fn from './utils/fn';
 import Component from './component';
+import document from 'global/document';
 
 const MODAL_CLASS_NAME = 'vjs-modal-dialog';
 const ESC = 27;
@@ -187,6 +188,7 @@ class ModalDialog extends Component {
 
       player.controls(false);
       this.show();
+      this.conditionalFocus_();
       this.el().setAttribute('aria-hidden', 'false');
 
       /**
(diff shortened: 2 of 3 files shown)


gkatsev:
I think this PR needs to be updated to keep the focus trapped inside the modal dialog while it's open.

src/js/modal-dialog.js
```diff
@@ -399,6 +410,94 @@ class ModalDialog extends Component {
...
+      event.preventDefault();
+    }
+  }
+
+  /**
+   * get all focusable elements
+   *
+   * @private
+   */
+  focusableEls_() {
```
BrandonOCasey: Would this be something we could use elsewhere? Should it be in `Component`?

BrandonOCasey: Should this be called `focusableChildELs_`?

src/js/component.js
```diff
@@ -999,6 +999,20 @@ class Component {
   }
 
   /**
+   * Set the focus to this component
+   */
+  focus() {
+    this.el_.focus();
+  }
```
BrandonOCasey: I think we already have these in master? Does this PR need a rebase for some reason? https://github.com/videojs/video.js/blob/master/src/js/component.js#L1004

src/js/modal-dialog.js
BrandonOCasey: Is there a link that should be included in the comments here to explain how this works?

src/js/modal-dialog.js
```diff
@@ -399,6 +410,94 @@ class ModalDialog extends Component {
...
+   * conditionally blur the element and refocus the last focused element
+   *
+   * @private
+   */
+  conditionalBlur_() {
+    if (this.previouslyActiveEl_) {
+      this.previouslyActiveEl_.focus();
+      this.previouslyActiveEl_ = null;
+
+      this.off(document, 'keydown', this.handleKeyDown);
```
BrandonOCasey: Should we always turn off the listener for `keydown` regardless of wether there is a `previouslyActiveEl_`?

src/js/modal-dialog.js
```diff
@@ -399,6 +410,94 @@ class ModalDialog extends Component {
...
   }
+
+  /**
+   * conditionally focus the modal dialog if focus was previously on the player.
+   *
+   * @private
+   */
+  conditionalFocus_() {
+    const activeEl = document.activeElement;
+    const playerEl = this.player_.el_;
```
BrandonOCasey: I am wondering if we even need a const for this here, since it is only used in the if statement below. Doesn't matter either way though.

src/js/component.js
gkatsev: Yup, it needs to be rebased. I had two separate PRs that both needed this.

src/js/modal-dialog.js
gkatsev: Because using `instanceof` isn't great, I think it should stay here until and if we start needing it elsewhere.
`focusableChildEls_` is probably more accurate.

src/js/modal-dialog.js
gkatsev: Potentially, though, I think we only need to `off` it only if one exists because that's the only time we `on` it.

src/js/modal-dialog.js
gkatsev: I thought it was a bit cleaner than using the expanded form twice.

src/js/modal-dialog.js
BrandonOCasey: unless the previouslyActiveEl_ gets disposed/removed on the player for some reason between when this gets focused and blured

src/js/modal-dialog.js
BrandonOCasey: ok thats fine then

gkatsev:
Interesting. I thought I had it working right. I'll take a look at it.

PR #4027, opened 2017-02-02
fix: localize aria-labels

### LINKED ISSUE #2728: Components do not localize aria-label text
Several components of the UI have ARIA markup to improve compatibility of the player with assistive technologies such as screen readers. Some components use aria-label to add hidden text which is announced by the screen reader; unlike other component labeling, this text is not localized, so it may not be useful for screen reader users in non-English speaking locales.

All aria-labels in the source can be found with [this search](https://github.com/videojs/video.js/search?q=aria-label).

Examples include:
'volume level' in [volume-bar.js](https://github.com/videojs/video.js/blob/e58c2266de3325c30412c2dcd1ce8096d980b63e/src/js/control-bar/volume-control/volume-bar.js#L37)
'video progress bar' in [seek-bar.js](https://github.com/videojs/video.js/blob/e58c2266de3325c30412c2dcd1ce8096d980b63e/src/js/control-bar/progress-control/seek-bar.js#L38)

gkatsev:
Fix #2728


src/js/control-bar/text-track-controls/captions-button.js
```diff
@@ -26,7 +26,7 @@ class CaptionsButton extends TextTrackButton {
    */
   constructor(player, options, ready) {
     super(player, options, ready);
-    this.el_.setAttribute('aria-label', 'Captions Menu');
+    this.el_.setAttribute('aria-label', this.localize('Captions Menu'));
```
gkatsev: @OwenEdwards thoughts on whether we need to create a new item for this or if we can just re-use `Captions` and `Chapters` etc for these button labels?

OwenEdwards: @gkatsev see #4034 - that PR removes this `aria-label` attribute in all menus.

lang/en.json
```diff
@@ -1,4 +1,6 @@
 {
+  "audio player": "audio player",
+  "video player": "video player",
```
misteroneill: Should these be title-cased for consistency as long as we're pushing toward a major release?

src/js/control-bar/text-track-controls/captions-button.js
gkatsev: Cool. I'll rebase this after that one gets merged then.

lang/en.json
OwenEdwards: I agree, these and others (e.g. "progress bar") ought to be title-cased.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -48,7 +48,7 @@ class SeekBar extends Slider {
     return super.createEl('div', {
       className: 'vjs-progress-holder'
     }, {
-      'aria-label': 'progress bar'
+      'aria-label': this.localize('progress bar')
```
OwenEdwards: Doesn't this need to be in Title Case too?

gkatsev: whoops

PR #4028, opened 2017-02-02
Combined tracks control

mister-ben:
## Description
Puts captions and subtitles in a single control. Tidier UI and less confusing for those who don't generally distinguish between those types. Captions have '[CC]' appended - CSS currently, but this should be localisable.

For now, there's an example in /sandbox.

## Specific Changes proposed
- Added a 'TrackButton'
- Modified the track controls to have a `kinds` array internally so they can iterate through more than one kind when toggling other tracks' state.
- Doesn't break the exisiting single-kind controls (in fact fixes an issue that the subtitles menu doesn't update if a subtitle track was showing when captions are enabled)

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/control-bar/text-track-controls/tracks-button.js
```diff
@@ -0,0 +1,105 @@
...
+   *
+   * @param {Object} [options]
+   *        The key/value store of player options.
+   *
+   * @param {Component~ReadyCallback} [ready]
+   *        The function to call when this component is ready.
+   */
+  constructor(player, options, ready) {
+    super(player, options, ready);
+    this.el_.setAttribute('aria-label', 'Captions Menu');
```
OwenEdwards: This aria-label assignment should be removed after #4034 lands.

src/js/control-bar/text-track-controls/tracks-button.js
```diff
@@ -0,0 +1,105 @@
...
+ */
+TracksButton.prototype.kinds_ = ['captions', 'subtitles'];
+
+/**
+ * The text that should display over the `TracksButton`s controls. Added for localization.
+ *
+ * @type {string}
+ * @private
+ */
+TracksButton.prototype.controlText_ = 'Captions';
```
OwenEdwards: This text will need to change depending on North America/Rest Of the World, as with the icon for the control.

src/js/control-bar/control-bar.js
```diff
@@ -67,6 +68,7 @@ ControlBar.prototype.options_ = {
     'descriptionsButton',
     'subtitlesButton',
     'captionsButton',
+    'tracksButton',
```
OwenEdwards: I assume that the default Control Bar won't have `subtitlesButton`, `captionsButton` *and* `tracksButton` - it'll be either/or?

OwenEdwards:
To me, the name "tracksButton" is confusing because we have so many different kinds of tracks now, including (non-text) audio tracks. How about "subsCapsButton", something like that?

BrandonOCasey:
what about `textTracksButton`?

src/js/control-bar/control-bar.js
mister-ben: Left in for testing to make sure they don't break each other, but I don't see a reason for them to be used like that for real.

mister-ben:
Agreed, `tracksButton` is inadequate.  I haven't thought of anything better than `subCapsButton` that that's descriptive and accurate.

src/js/control-bar/text-track-controls/tracks-button.js
mister-ben: And a little more complex to localise if the string needs to be localised as captions or subtitles here, but specifically as captions on the captions-only control, which might be translated as "subtitles for the hard of hearing".

This would be easier if we used something like CAPTIONS_OR_SUBTITLES as keys in the translations json and included en.json in the player.

OwenEdwards:
Looking at the way Safari handles this for videos with text tracks and an audio description track, it all gets rolled together with the Subtitles icon for the menu, then options for "Audio Track" and "Subtitles".
![2017-02-08_14-09-41]. Thoughts on that?(https://cloud.githubusercontent.com/assets/2245205/22759729/5f11e980-ee08-11e6-8401-c8f9a48c0a44.png)

src/js/control-bar/text-track-controls/tracks-button.js
gkatsev: That PR was landed. This PR probably needs to be rebased against master.

src/js/control-bar/text-track-controls/off-text-track-menu-item.js
```diff
@@ -51,7 +58,7 @@ class OffTextTrackMenuItem extends TextTrackMenuItem {
     for (let i = 0, l = tracks.length; i < l; i++) {
       const track = tracks[i];
 
-      if (track.kind === this.track.kind && track.mode === 'showing') {
+      if ((this.track.kinds.indexOf(track.kind) > -1) && track.mode === 'showing') {
```
gkatsev: this line is erroring out for me. `this.track.kinds`doesn't get set.

gkatsev: should it be `this.options_.kinds`?

src/css/components/_tracks-button.scss
```diff
@@ -0,0 +1,31 @@
...
+.video-js .vjs-tracks-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-GB) .vjs-tracks-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-IE) .vjs-tracks-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-AU) .vjs-tracks-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-NZ) .vjs-tracks-button .vjs-icon-placeholder {
+  @extend .vjs-icon-subtitles;
+}
+
+// Todo
+.video-js .vjs-tracks-button .vjs-captions-menu-item::after {
```
gkatsev: we should continue using single colon for pseudo elements.

src/css/components/_tracks-button.scss
```diff
@@ -0,0 +1,31 @@
...
+  content: ' [CC]';
+}
+
+.video-js:lang(en) .vjs-tracks-button .vjs-icon-placeholder,
+.video-js:lang(fr-CA) .vjs-tracks-button .vjs-captions-menu-item::after {
+  content: none;
+}
+
+.video-js:lang(en-US) .vjs-tracks-button .vjs-subtitles-menu-item::after,
```
gkatsev: I think this should be `en`?

src/css/components/_tracks-button.scss
gkatsev: so, `captions` are always indicated as `[cc]` and subs are only indicated as such in `en` and `fr-CA` locales?

src/js/control-bar/control-bar.js
gkatsev: What about removing from here and updating the combined-tracks test page to include all the separate caption buttons in the control bar as a default config?

mister-ben:
Updated and rebased.

I've left the old controls in for the moment as several tests need to be modified to take them out of the default ControlBar.

To distinguish captions in the menu, I've added the CC icon rather than "[cc]" text now - thoughts on whether that's readable enough? The material cc icon isn't great at small sizes.

src/js/control-bar/control-bar.js
```diff
@@ -68,6 +69,7 @@ ControlBar.prototype.options_ = {
     'descriptionsButton',
     'subtitlesButton',
     'captionsButton',
+    'subCapsButton',
```
gkatsev: this should be `subsCapsButton` took me forever to figure out why it wasn't working for me, haha

src/css/components/_subs-caps.scss
```diff
@@ -0,0 +1,21 @@
...
+.video-js .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-GB) .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-IE) .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-AU) .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-NZ) .vjs-subs-caps-button .vjs-icon-placeholder {
+  @extend .vjs-icon-subtitles;
+}
+
+// Todo
+.video-js .vjs-subs-caps-button + .vjs-menu .vjs-captions-menu-item .vjs-menu-item-text:after {
```
gkatsev: this should be `vertical-align: baseline` so it's all on the same line.

src/js/control-bar/control-bar.js
mister-ben: Urgh, sorry.

src/css/components/_subs-caps.scss
mister-ben: `vertical-align: bottom` seems to work better.

mister-ben:
I think this is in a viable state now, with updated tests

src/js/control-bar/text-track-controls/subs-caps-button.js
```diff
@@ -0,0 +1,100 @@
...
+ * @extends TextTrackButton
+ */
+class SubsCapsButton extends TextTrackButton {
+
+  constructor(player, options = {}) {
+    super(player, options);
+
+    // Although North America uses "captions" in most cases for
+    // "captions and subtitles" other locales use "subtitles"
+    this.label_ = 'subtitles';
```
mister-ben: This gets concatenated into 'captions setttings' and 'captions off', which do get localised.

src/js/control-bar/text-track-controls/subs-caps-button.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * The text that should display over the `SubsCapsButton`s controls.
+     *
+     *
+     * @type {string}
+     * @private
+     */
+    SubsCapsButton.prototype.controlText_ = this.label_;
```
mister-ben: The controlText wasn't getting set at all, that approach fixes it.

src/css/components/_subs-caps.scss
```diff
@@ -0,0 +1,22 @@
...
+.video-js .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-GB) .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-IE) .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-AU) .vjs-subs-caps-button .vjs-icon-placeholder,
+.video-js.video-js:lang(en-NZ) .vjs-subs-caps-button .vjs-icon-placeholder {
+  @extend .vjs-icon-subtitles;
+}
+
+.video-js .vjs-subs-caps-button + .vjs-menu .vjs-captions-menu-item .vjs-menu-item-text:after {
+  content: " \f10d";
```
gkatsev: this is showing up in the screen reader. We need to add an icon placeholder icon.

OwenEdwards: Good catch @gkatsev !! I haven't had a chance to check this, but is there some `controlText` for the icon, so that something is read out by a screen reader instead of the icon? Otherwise, that's an #a11y violation too - information which is conveyed visually, but not to SR users.

OwenEdwards:
My one slight concern about merging this PR is that, while the feature looks great, the dummy WebVTT tracks in the `docs/examples/shared` folder are just or testing, and aren't really what they say they are. Also, the test HTML file in the `sandbox` is the only one with a `.html` extension; we discussed why they have the `.example` extension a while ago, and I'd love to change it, but we ought to go one way or another with this, for consistency.

mister-ben:
@OwenEdwards good point, I'll clean that up. The Oceans video doesn't really lend itself to a meaningful demo of captions and subtitles at all though.

gkatsev:
Maybe we should switch the default video to something like elephants dream which has all the tracks? At least for the sandbox?

sandbox/combined-tracks.html.example
```diff
@@ -38,17 +25,17 @@
...
-    <source src="http://vjs.zencdn.net/v/oceans.mp4" type="video/mp4">
-    <source src="http://vjs.zencdn.net/v/oceans.webm" type="video/webm">
-    <source src="http://vjs.zencdn.net/v/oceans.ogv" type="video/ogg">
-    <track kind="captions" src="../docs/examples/shared/example-captions.vtt" srclang="en" label="English"></track><!-- Tracks need an ending tag thanks to IE9 -->
-    <track kind="captions" src="../docs/examples/shared/example-captions-fr.vtt" srclang="fr" label="French"></track><!-- Tracks need an ending tag thanks to IE9 -->
-    <track kind="subtitles" src="../docs/examples/shared/example-subtitles-fr.vtt" srclang="fr" label="French"></track><!-- Tracks need an ending tag thanks to IE9 -->
-    <track kind="subtitles" src="../docs/examples/shared/example-subtitles-de.vtt" srclang="de" label="German"></track><!-- Tracks need an ending tag thanks to IE9 -->
+  <video id="vid1" class="video-js vjs-default-skin" lang="en" controls preload="auto" width="640" height="264" poster="http://vjs.zencdn.net/v/oceans.png">
+    <source src="//d2zihajmogu5jn.cloudfront.net/elephantsdream/ed_hd.mp4" type="video/mp4">
```
OwenEdwards: Nice! I guess I ought to move the Descriptions example to use this Cloudfront URL!

mister-ben: This is the content used at videojs.com/advanced

src/css/components/_subs-caps.scss
mister-ben: OK, this can't be done with CSS alone in that case.

src/js/control-bar/text-track-controls/subs-caps-button.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * @file tracks-button.js
```
OwenEdwards: Just noticed - this file name needs updating. (<Captain Pedantic, at you're service ;-)>)

mister-ben: Ah yes. (Not rising to "you're"...)

src/css/components/_subs-caps.scss
gkatsev: Yeah, I don't think this is possible to do in CSS. @mister-ben do you think you can get to it this week? Otherwise, I'm thinking we can merge it in as is and I or someone else can tackle it.

mister-ben: I probably won't have a chance to look at this now before the end of next week.

PR #4029, opened 2017-02-02
Disable time tooltips entirely in IE8

misteroneill:
Pretty self-explanatory. These are more trouble than they're worth for a browser we only semi-support.

There are also a few tiny code cleanup pieces in here. Nothing significant.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/progress-control/play-progress-bar.js
+++ b/src/js/control-bar/progress-control/play-progress-bar.js
@@ -2,6 +2,7 @@
  * @file play-progress-bar.js
  */
 import Component from '../../component.js';
+import {IE_VERSION} from '../../utils/browser.js';
 import formatTime from '../../utils/format-time.js';
 
 import './time-tooltip';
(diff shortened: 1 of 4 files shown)


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -177,12 +178,15 @@ class SeekBar extends Slider {
...
   children: [
     'loadProgressBar',
-    'mouseTimeDisplay',
     'playProgressBar'
   ],
   barName: 'playProgressBar'
 };
 
+if (!IE_VERSION || IE_VERSION > 8) {
+  SeekBar.prototype.options_.children.push('mouseTimeDisplay');
```
gkatsev: should we splice it in to maintain the ordering?

src/js/control-bar/progress-control/play-progress-bar.js
```diff
@@ -77,9 +78,7 @@ class PlayProgressBar extends Component {
  * @private
  */
 PlayProgressBar.prototype.options_ = {
-  children: [
-    'timeTooltip'
-  ]
+  children: IE_VERSION && IE_VERSION < 9 ? [] : ['timeTooltip']
```
BrandonOCasey: I think we may want make this the same style as the other if check

BrandonOCasey:
Booting up IE8 to test, code looks good

PR #4030, opened 2017-02-03
Filter the source given to `Player#src` before use

BrandonOCasey:
## Description
* Add a util that filters all possible `Tech~SourceObjects` into a valid `Tech~SourceObject` (see tests) 
* Remove `Player#sourceList_` as it is no longer in use

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2219,28 +2219,27 @@ class Player extends Component {
    *         The current video source when getting
    */
   src(source) {
+    // getter usage
+    if (typeof source !== 'undefined') {
+      return this.cache_.src;
+    }
     let src = source;
 
     // filter out invalid sources and turn our source into
     // an array of source objects
     src = filterSource(src);
 
-    // no valid sources, return the current source
+    // if a source was passed in then it is invalid because
+    // it was filtered to a zero length Array. So we have to
+    // show an error
     if (!src.length) {
-      // if a source was passed in then it is invalid because
-      // it was filtered to a zero length Array. So we have to
-      // show an error
-      if (typeof source !== 'undefined') {
-        this.setTimeout(function() {
-          this.error({ code: 4, message: this.localize(this.options_.notSupportedMessage) });
-        }, 0);
-        return;
-      }
-      return this.cache_.src;
+      this.setTimeout(function() {
+        this.error({ code: 4, message: this.localize(this.options_.notSupportedMessage) });
+      }, 0);
+      return;
     }
 
     this.cache_.sources = src;
-
     this.changingSrc_ = true;
     this.cache_.source = src[0];
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2221,24 +2221,18 @@ class Player extends Component {
   src(source) {
     let src = source;
 
-    // remove invalid values from here and turn the source into an object
+    // filter out invalid sources and turn our source into
+    // an array of source objects
     src = filterSource(src);
 
-    // if they pass a zero length array, or
-    // an undefined value, return the current source.
-    // aka make this function work as a getter
-    if ((Array.isArray(src) && !src.length) || !src) {
+    if (!src.length) {
       return this.cache_.src;
     }
 
-    this.changingSrc_ = true;
+    this.cache_.sources = src;
+    src = src[0];
 
-    if (Array.isArray(src)) {
-      this.cache_.sources = src;
-      src = src[0];
-    } else {
-      this.cache_.sources = [src];
-    }
+    this.changingSrc_ = true;
 
     this.cache_.source = src;
     this.currentType_ = src.type;
@@ -2249,8 +2243,6 @@ class Player extends Component {
       const err = this.src_(src_);
 
       if (err) {
-        // if there was an error with this src
-        // try another
         if (Array.isArray(source) && source.length > 1) {
           return this.src(source.slice(1));
         }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -28,7 +28,7 @@ import ModalDialog from './modal-dialog';
 import Tech from './tech/tech.js';
 import * as middleware from './tech/middleware.js';
 import {ALL as TRACK_TYPES} from './tracks/track-types';
-import {filterSource} from './utils/filter-source';
+import filterSource from './utils/filter-source';
 
 // The following imports are used only to ensure that the corresponding modules
 // are always included in the video.js package. Importing the modules will
@@ -2215,7 +2215,7 @@ class Player extends Component {
    * @param {Tech~SourceObject|Tech~SourceObject[]} [source]
    *        One SourceObject or an array of SourceObjects
    *
-   * @return {String}
+   * @return {string}
    *         The current video source when getting
    */
   src(source) {
@@ -2225,6 +2225,7 @@ class Player extends Component {
     // an array of source objects
     src = filterSource(src);
 
+    // no valid sources, return the current source
     if (!src.length) {
       return this.cache_.src;
     }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2220,7 +2220,7 @@ class Player extends Component {
    */
   src(source) {
     // getter usage
-    if (typeof source !== 'undefined') {
+    if (typeof source === 'undefined') {
       return this.cache_.src;
     }
     let src = source;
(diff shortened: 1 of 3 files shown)


src/js/player.js
```diff
@@ -2218,27 +2219,23 @@ class Player extends Component {
    *         The current video source when getting
    */
   src(source) {
-    if (source === undefined) {
```
gkatsev: why was this part removed?

gkatsev: Ah, I see. I think it's better to keep it. Why do extra work if we know for sure we can exit early here?

src/js/player.js
```diff
@@ -2218,27 +2219,23 @@ class Player extends Component {
...
-      src = {
-        src: source
-      };
+    // filter out invalid sources and turn our source into
+    // an array of source objects
+    src = filterSource(src);
 
-      this.cache_.sources = [src];
+    // no valid sources, return the current source
+    if (!src.length) {
```
gkatsev: Should we treat `src([])` the same as `src()`?

src/js/utils/filter-source.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @module filter-source
+ */
+import {isObject} from './obj';
+
+/**
+ * Filter out single bad source objects or multiple source objects in an
```
gkatsev: I don't think it's worth handling nested arrays of sources.

src/js/player.js
```diff
@@ -2218,27 +2219,23 @@ class Player extends Component {
...
+      return this.cache_.src;
     }
 
-    this.cache_.source = src;
+    this.cache_.sources = src;
+    src = src[0];
```
gkatsev: with this, we'd only ever be able to try running through the first source. See line 2247.

src/js/player.js
```diff
@@ -2218,27 +2219,23 @@ class Player extends Component {
...
+
+    this.changingSrc_ = true;
 
+    this.cache_.source = src;
     this.currentType_ = src.type;
```
gkatsev: just realized that we need to move/copy this down into line 2264

src/js/player.js
gkatsev: Or at least, we don't have the benefit of `filterSource` for subsequent iterations and the two arrays aren't necessary in sync.

src/js/player.js
BrandonOCasey: I think we want to exit early in the case where an empty source is passed, I could see logging an error, but not much else would be different.

src/js/utils/filter-source.js
BrandonOCasey: I will remove that from the doc comments/unit tests if you like, but the code will probably stay the same either way.

src/js/player.js
BrandonOCasey: yeah I guess we can keep it to do less work, I just though it would be better to have less code. I don't think that filterSource will do anything with it other than turn it into an empty array. I guess it depends on wether or not we want to log an error if an invalid source array is passed to us vs being used as a getter.

gkatsev: video.js throws the following errors:
```js
player.src([]); // No compatible source was found for this media
player.src({}); // The media could not be loaded, either because the server or network failed or because the format is not supported.
```

It looks like vjs 5 special cased empty arrays to have the custom object and with an empty object it basically tried to set an empty string and had it raise the error from the video element.

gkatsev: I think we should maintain this functionality. Though, having `player.src({})` produce the same error as `player.src([])` sounds fine too.

src/js/player.js
BrandonOCasey: is there any reason not to use the `cache_` object here?

src/js/player.js
BrandonOCasey: I think 2247 uses the original `source` object and not `src`

src/js/player.js
gkatsev: probably no reason not to ad that in. This was just how it used to be.

src/js/player.js
gkatsev: Yes, but previously `source` was the source of truth for a list of sources. With the `filterSource` functionality, the output of `filterSource` became the source of truth.

src/js/player.js
BrandonOCasey: Do we want to wrap the `this.error` call in a `setTimeout` like what was done in `sourceList_`?

src/js/player.js
BrandonOCasey: I see

src/js/player.js
BrandonOCasey: I am wondering two things now:
1. Do we need to change `this.cache_.source`, `this.cache_.src`, and `this.cache_.type` in the `middleware.setSource` callback and also above it? or an we just do it inside?
2. Should we even cache `this.cache_.src` and `this.cache_.type` or should we just cache `this.cache_.source` and use that when we need to get the `type` or the `src` url

gkatsev: If you have a source like this:
```js
{
  src: 'foo',
  type: 'video/foo'
}
```
With a middleware that returns an m3u8 for it, we want the following behavior:
```js
player.src();
// 'http://example.com/video.m3u8'

player.currentSrc();
// 'foo'

player.currentSource();
// {src: 'foo', type: 'video/foo'}
```

Let me double check what we talked about for `currentType` previously.

gkatsev: Oh, maybe we don't need to move `currentType` because we want it to match the type of the video returned by `player.currentSrc()`. So, in the above scenario, it'll be `video/foo`.

src/js/player.js
gkatsev: Probably, yes.

BrandonOCasey:
@gkatsev this has been updated with a lot of the things that you asked about

src/js/player.js
BrandonOCasey: done

src/js/player.js
```diff
@@ -2385,14 +2357,7 @@ class Player extends Component {
    *         The current source object
    */
   currentSource() {
-    const source = {};
-    const src = this.currentSrc();
```
BrandonOCasey: This code didn't make sense because `this.currentSrc` was looking at the same variable (`this.cache_.source`).

src/js/player.js
```diff
@@ -2218,37 +2219,39 @@ class Player extends Component {
    *         The current video source when getting
    */
   src(source) {
-    if (source === undefined) {
```
gkatsev: I really think this should be separate and at the top here. It makes a clear delineation between when it's being called as a getter vs when it's called as a setter.

src/js/player.js
```diff
@@ -2415,7 +2380,7 @@ class Player extends Component {
    *         The source MIME type
    */
   currentType() {
-    return this.currentType_ || '';
+    return this.currentSource() && this.currentSource().type || '';
```
gkatsev: any other references to `this.currentType_` that need to be removed?

src/js/player.js
BrandonOCasey: ok

src/js/player.js
BrandonOCasey: using `ack` and `ripgrep` returns no results for `this.currentType_` now, so I don't think so.

src/js/player.js
BrandonOCasey: added back in in the most recent commit

src/js/player.js
```diff
@@ -2264,11 +2266,26 @@ class Player extends Component {
       }
 
       this.changingSrc_ = false;
-      this.cache_.src = src_.src;
+      this.cache_.source = src[0];
+      this.cache_.src = src_;
```
gkatsev: this should be `src_.src`.

gkatsev:
Also, should we clear out the source caches if we error out? So, if everything worked but then you do `player.src([])`, `player.currentSources()` still returns the previously loaded sources.

BrandonOCasey:
Maybe we should reset the source cache at the top of the function if it is not being used as a getter?

src/js/player.js
BrandonOCasey: your right

gkatsev:
Making the executive decision to merge even though travis is being silly. Hopefully master will be fine.

PR #4034, opened 2017-02-06
Fix the structure of elements in menus to comply with ARIA requirements

OwenEdwards:
## Description
The previous (v5.x) menu structure violated an ARIA requirement, that actionable elements (e.g. the menu button) not have child elements which are also actionable (e.g. the menu items). This PR changes that structure to make "MenuButton" actually a `<div>` containing a `<button>` and a menu. The v5.x structure, and the use of `role="group"` on the Control Bar, were a workaround which made the menus work with most screen readers, but the structure might fail in future screen readers.

This change is very similar to the change to the combined volume from v5 -> v6.

## Specific Changes proposed
Restructure the HTML elements in a MenuButton & Menu; change the Javascript to handle this. Also, remove unnecessary `aria-label` attribute on the MenuButton, since the actual button used to open the menu is now correctly labeled by the `controlText`.

## Requirements Checklist

### DIFF
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -13,7 +13,7 @@ import document from 'global/document';
 /**
  * A `MenuButton` class for any popup {@link Menu}.
  *
- * @extends ClickableComponent
+ * @extends Component
  */
 class MenuButton extends Component {
 
@@ -45,10 +45,10 @@ class MenuButton extends Component {
 
     this.enabled_ = true;
 
-    this.menuButton_.on('tap', Fn.bind(this, this.handleClick));
-    this.menuButton_.on('click', Fn.bind(this, this.handleClick));
-    this.menuButton_.on('focus', Fn.bind(this, this.handleFocus));
-    this.menuButton_.on('blur', Fn.bind(this, this.handleBlur));
+    this.on(this.menuButton_, 'tap', this.handleClick);
+    this.on(this.menuButton_, 'click', this.handleClick);
+    this.on(this.menuButton_, 'focus', this.handleFocus);
+    this.on(this.menuButton_, 'blur', this.handleBlur);
 
     this.on('keydown', this.handleSubmenuKeyPress);
   }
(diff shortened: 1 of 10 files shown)


src/js/menu/menu-button.js
```diff
@@ -27,12 +29,27 @@ class MenuButton extends ClickableComponent {
...
+
+    this.addChild(this.menuButton_);
+
     this.update();
 
     this.enabled_ = true;
 
-    this.el_.setAttribute('aria-haspopup', 'true');
-    this.el_.setAttribute('role', 'menuitem');
+    this.menuButton_.on('tap', Fn.bind(this, this.handleClick));
```
gkatsev: I think you can change this to:
```js
this.on(this.menuButton_, 'tap', this.handleClick);
```

src/js/menu/menu-button.js
```diff
@@ -164,6 +204,24 @@ class MenuButton extends ClickableComponent {
   }
 
   /**
+   * Handle Focus - Add keyboard functionality to element
+   *
+   * @method handleFocus
+   */
+  handleFocus() {
```
gkatsev: not that we need to right now, but I think we need a keyhandler-mixin.

OwenEdwards:
I think these two commits fix all of the issues @gkatsev pointed out.

src/js/menu/menu-button.js
```diff
@@ -113,11 +130,34 @@ class MenuButton extends ClickableComponent {
...
     });
   }
 
   /**
+   * Allow sub components to stack CSS class names for the wrapper element
+   *
+   * @return {string}
+   *         The constructed wrapper DOM `className`
+   */
+  buildWrapperCSSClass() {
```
gkatsev: I just realized that the change to this `buildWrapperCSSClass` means that the elements lose out on having something like `vjs-captions-button` applied to the main wrapper element.

I think we want to return something like 
```js
return `${buttonClass} ${super.buildCSSClass()} ${this.buildCSSClass()}`;
```
instead, since the components are generally already set up with `buildCSSClass` to return the correct thing.

I'll open an issue for this.

PR #4050, opened 2017-02-09
feat: fix accessibility of the captions setting dialog

gkatsev:
This fixes a lot of the issues from https://github.com/videojs/video.js/issues/2746 by making the dialog inherit from our actual ModalDialog which now has tab focus trapping.
Todo copied from #2746 to track what has been fixed. #2746 should be used to track what has been verified as fixed.

The Captions Settings dialog has some accessibility issues:
- [x] Field labels and fields are not explicitly associated
- [x] Keyboard focus does not move into the dialog when it is opened
- [x] Keyboard focus is not trapped inside the dialog while it is open
- [x] Keyboard focus does not return to the control which opened the dialog when it is closed
- [x] The extent (top and bottom) of the dialog is not indicated to screen readers
- [x] The dialog cannot be closed with the Esc key
- [x] The meaning of '---' in the select fields is not clear
- [x] The control to close the dialog is labeled "Done" rather than "Close"
- [x] The purpose of the "Defaults" button may not be obvious, and its effect may not be apparent to screen reader users
- [x] Focus highlighting (outline) on the Default and Done buttons is *very* hard to see
- [x] Pressing the Done button doesn’t seem to have the same effect as pressing the Close (x) button; does it trigger the same focus movement
- [x] This requirement to move it back to the triggering element is tricky, since clicking on that item in the CC menu dismisses the CC menu. I need to think about this a little more - either the menu should open again, or the focus should go to the main CC Menu Button
- [x] The focus outline on the whole dialog goes too far to the left (all the way to the edge of the video window, not just to the edge of the dialog)


gkatsev:
@OwenEdwards I'm not sure how the "extent" is supposed to be specified.

OwenEdwards:
Here's the idea of the "extent" of the dialog:

The top and bottom of the dialog area are clearly marked for sighted users. Focus is constrained inside the dialog, but a screen reader user can use the up and down arrow keys to move the "virtual cursor" and read content which isn't focusable/actionable. If they move out of the top or bottom of the dialog and into other content in the DOM, they may not realize that this has happened, and may become confused/lost.

The preferred solution is to use `aria-hidden="true"` on all of the rest of the page *except* the dialog - this is tricky because `aria-hidden` doesn't nest, so you can't have `aria-hidden="false"` on the dialog element when it is a child of the main page which has `aria-hidden="true"`. So you need to mark everything *else* as aria-hidden, or have a structure like this:

````
<body>
    <div class="main" aria-hidden="true> ... </div>
    <div role="dialog" ... > .... </div>
</body>
````

Alternatively, you can use hidden (screen reader only) text to say "Start of modal dialog", "End of modal dialog" at exactly the place where the extent of the dialog is visibly marked in the DOM.

There has been *some* talk that marking an ARIA `region`, or even the `role="dialog"`, ought to be enough in itself, but screen readers don't always announce movement of the virtual cursor into/out of an  ARIA region.

**Comment by gkatsev** [comment]
It sounds like have a "start" and "end" elements that are only available to screen readers is what we want.

**Comment by OwenEdwards** [comment]
Right, and I forgot that we already have `Beginning of dialog window....`, so we only need `End of dialog window`.

**Comment by gkatsev** [comment]
@OwenEdwards this is ready for testing. Only issue is https://github.com/videojs/video.js/issues/4049.

**Comment by gkatsev** [comment]
@OwenEdwards I'm not really sure how to improve the highlighting of the `defaults` and `done` buttons. Both of those are using the default outlines when just using the keyboard or using voiceover.

The others can definitely be fixed and I've already fixed the last one.

**Comment by gkatsev** [comment]
I decided to update it so that it re-focuses on the captions button when it is closed.

File: src/js/tracks/text-track-settings.js [hunk 1]
```diff
@@ -290,21 +305,28 @@ class TextTrackSettings extends Component {
...
       createEl('label', {
+        id,
         className: 'vjs-label',
         textContent: this.localize(config.label)
       }, {
-        for: id
       }),
-      createEl('select', {id}, undefined, config.options.map(o => {
+      createEl('select', {}, {
+        'aria-labeledby': `${legendId} ${id}`
```
gkatsev: Yup, oops.

src/js/tracks/text-track-settings.js
```diff
@@ -290,21 +305,28 @@ class TextTrackSettings extends Component {
...
+      }, config.options.map(o => {
+        const optionId = id + '-' + o[1];
+
         return createEl('option', {
+          id: optionId,
           textContent: this.localize(o[1]),
           value: o[0]
+        }, {
+          'aria-labeledby': `${legendId} ${id} ${optionId}`
```
gkatsev: I was testing with voiceover on mac. Seems like Safari reads it out like `White, Text Color, menu item` but on chrome it reads it out only as `white` unless I add this `labelledby`.

gkatsev:
@misteroneill all the missing jsdocs are fixed by adding `@extends ModalDialog`.

src/js/tracks/text-track-settings.js
```diff
@@ -290,21 +306,28 @@ class TextTrackSettings extends Component {
...
+        'aria-labelledby': `${legendId} ${id}`
+      }, config.options.map(o => {
+        const optionId = id + '-' + o[1];
+
         return createEl('option', {
+          id: optionId,
           textContent: this.localize(o[1]),
           value: o[0]
+        }, {
+          'aria-labelledby': `${legendId} ${id} ${optionId}`
```
gkatsev: @OwenEdwards did we have an answer whether we should keep this because of VO and Chrome or remove it?

OwenEdwards: I'm a little concerned about how well this will work - there's a circular reference where the label for this element includes the `id` of two other elements, and then the **id of this element**. The behavior in this situation *is mentioned* in the W3C [Accessible Name Computation - Text Alternative Computation](https://www.w3.org/TR/accname-aam-1.1/#mapping_additional_nd_te):

> if computing a name, and the current node has a non-empty aria-labelledby attribute, _and the current node is not already part of an aria-labelledby traversal_, process its IDREFs in the order they occur: 

but it may not work correctly in all browsers. I guess we'll just have to fix it if that happens?

gkatsev: from the little I understand about this, I think it should be fine and it falls within what the spec allows.
I'm 👌 with leaving this and then fixing in the future if we find an issue.

src/js/tracks/text-track-settings.js
```diff
@@ -435,9 +461,11 @@ class TextTrackSettings extends Component {
    * @private
    */
   createElControls_() {
+    const defaultsDescription = this.localize('restore all settings to the default values');
     const defaultsButton = createEl('button', {
       className: 'vjs-default-button',
-      textContent: this.localize('Defaults')
+      title: defaultsDescription,
+      innerHTML: `${this.localize('Defaults')}<span class='vjs-control-text'> - ${defaultsDescription}</span>`
```
gkatsev: @OwenEdwards @mmcc brought up that maybe 'Reset' would be a better name for this button, thoughts?

OwenEdwards: I think that would work fine - if the button is marked 'Reset', then the hidden text can be ' all settings to the default values', and you don't need the ' - ' in the hidden text.

BrandonOCasey: Reset seems like a better name to me

src/js/tracks/text-track-settings.js
```diff
@@ -457,6 +485,10 @@ class TextTrackSettings extends Component {
    *         The element that was created.
    */
   createEl() {
+    return super.createEl();
```
BrandonOCasey: I don't know if we need this here, won't this be the default behavior?

src/js/tracks/text-track-settings.js
```diff
@@ -457,6 +485,10 @@ class TextTrackSettings extends Component {
...
+  }
+
+  content() {
```
BrandonOCasey: This function and a few of the funcitons below this need jsdoc comments

src/js/tracks/text-track-settings.js
gkatsev: Um... yeah, probably not needed.

src/js/tracks/text-track-settings.js
gkatsev: JSDoc gets them from the parent.

src/js/tracks/text-track-settings.js
misteroneill: +1

gkatsev: Updated.

PR #4060, opened 2017-02-10
feat: allow tokens in localize, localize progress bar time

### LINKED ISSUE #4024: Progress bar text for screen reader should include the duration, if available
## Description
The progress bar uses `aria-valuetext` to convey the current time to screen reader users. Visually, the current time is displayed as a point on the progress bar line where the _length of the bar_ represents the duration of the video. Screen reader users should also be presented with this information from the progress bar, for example as "01:23 of 12:34", rather than just "01:23" (and then needing to navigate through the control bar for the duration separately).

gkatsev:
Fix #4024.

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -222,28 +222,28 @@ class Component {
    * @return {string}
    *         The localized string or if no localization exists the english string.
    */
-  localize(string) {
+  localize(string, tokens) {
     const code = this.player_.language && this.player_.language();
     const languages = this.player_.languages && this.player_.languages();
+    const language = languages && languages[code];
+    const primaryCode = code && code.split('-')[0];
+    const primaryLang = languages && languages[primaryCode];
 
-    if (!code || !languages) {
-      return string;
-    }
-
-    const language = languages[code];
+    let localizedString = string;
 
     if (language && language[string]) {
-      return language[string];
+      localizedString = language[string];
+    } else if (primaryLang && primaryLang[string]) {
+      localizedString = primaryLang[string];
     }
 
-    const primaryCode = code.split('-')[0];
-    const primaryLang = languages[primaryCode];
-
-    if (primaryLang && primaryLang[string]) {
-      return primaryLang[string];
+    if (tokens) {
+      localizedString = localizedString.replace(/\{(\d+)\}/g, function(match, index) {
+        return tokens[index - 1];
+      });
     }
 
-    return string;
+    return localizedString;
   }
 
   /**
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -216,8 +216,37 @@ class Component {
   /**
    * Localize a string given the string in english.
    *
+   * If tokens are provided, it'll try and run a simple token replacement on the provided string.
+   * The tokens it loooks for look like `{1}` with the index being 1-indexed into the tokens array.
+   *
+   * If a `defaultValue` is provided, it'll use that over `string`,
+   * if a value isn't found in provided language files.
+   * This is useful if you want to have a descriptive key for token replacement
+   * but have a succinct localized string and not require `en.json` to be included.
+   *
+   * Currently, it is used for the progress bar timing.
+   * ```js
+   * {
+   *   "progress bar timing: currentTime={1} duration={2}": "{1} of {2}"
+   * }
+   * ```
+   * It is then used like so:
+   * ```js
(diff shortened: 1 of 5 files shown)


src/js/component.js
```diff
@@ -220,28 +220,28 @@ class Component {
...
     }
 
-    const primaryCode = code.split('-')[0];
-    const primaryLang = languages[primaryCode];
-
-    if (primaryLang && primaryLang[string]) {
-      return primaryLang[string];
+    if (tokens) {
+      localizedString = localizedString.replace(/\{(\d+)\}/g, function(match, index) {
+        return tokens[index - 1];
```
OwenEdwards: Do we need to check the validity of `index` here (is a number, is in-range for `tokens[]`, etc.)?

gkatsev: Yeah, we probably should. Should we just return the `match` if the index isn't available in `tokens[]`?
So, if you do `this.localize('{1} {2} {3}', ['a', 'b'])` you would get `a b {3}`.

OwenEdwards: Seems like a reasonable default behavior. Of course, we're also going to need to clearly document what `{1}, {2}` etc. are for each string.

OwenEdwards: Just looking at this, I realize that the visual display next to the progress bar is actually the _remaining time_, not the duration of the video. Maybe `{3}` should be remaining time, so that the text can be customized that way, too?

OwenEdwards:
Tested with IE11 + JAWS, and macOS 10.12.3 Safari + VoiceOver; other than the notes I've already made, LGTM

src/js/component.js
mister-ben: > Of course, we're also going to need to clearly document what {1}, {2} etc. are for each string

I've been planning to write a glossary explaining every string in en.json. The translations have grown a lot recently and there's a lot in there that's not all that clear out of context.

gkatsev: Another thing I can think of to help with this is having the translation string and then default be two different things. So, for this case I can do something like `this.localize('progress bar timing', [currentTime, duration], '{1} {2}')` and the en.json would have `progress bar timing: '{1} {2}'`. That way we have a default, that's described, but people who only look at the json would have a better idea about what's happening.
Not quite sure if there's the best approach but a thought.

gkatsev: To elaborate and update (after some discussion with @OwenEdwards).
The key should be `progress bar timing: currentTime={1}, duration={2}` and the value would be `{1} of {2}`. That way, as you're reading the JSON files, you have some idea of what the items are.
Then, we would want to add a 3rd argument to `localize`, a defaults argument, which would be `{1} of {2}` which is what should be used in the case a language file isn't included (and so we don't have to require en.json to be included).

src/js/component.js
```diff
@@ -220,28 +220,35 @@ class Component {
    * @return {string}
    *         The localized string or if no localization exists the english string.
    */
-  localize(string) {
+  localize(string, tokens, defaultValue = string) {
```
misteroneill: Don't forget to update the JSDoc block for this function.

gkatsev: oops, will do.

gkatsev: Updated.

BrandonOCasey: do we actually need "defaultValue" as an argument here? Can't we just make a copy of string in the function

gkatsev: yes, we need `defaultValue` as an argument. The `= string` just makes the default value of `defaultValue` to be string. But if a `defaultValue` is passed in to this function, it is used instead of the value of `string.

BrandonOCasey: Does it ever make sense to make the default value anything other than the string that was passed in for localization though?

misteroneill: Yeah, I'm confused by this, too.

gkatsev: I tried explaining it in the jsdoc, can you re-read it and let me know?

misteroneill: Ah, yes, makes sense to me. 👍

BrandonOCasey: Why wouldn't we just make "{1} of {2}" the string we are localizing here? Seems weird to make the translation key and the value have different wording.

gkatsev: So that the key in the lang file describes what it should be. If you see `{1} of {2}` how would you know what is `{1}` and `{2}` and also what should you be translating?

BrandonOCasey: hmmm ok LGTM

PR #4062, opened 2017-02-11
Update French translation

lionel-m:
## Description
French translation updated


lionel-m:
Casual is a typeface? I did not find any translation for this word...

mister-ben:
Yes. A translation of "handwriting" might be ok.

lionel-m:
@mister-ben Thanks for the clarification. PR updated

PR #4070, opened 2017-02-13
docs: Add MediaLoader to components list

mister-ben:
Updated list.

As an aside, should the loader's `div` have a class so it's obvious what it is when looking at the DOM?

### DIFF
--- a/docs/guides/components.md
+++ b/docs/guides/components.md
@@ -285,6 +285,7 @@ The default component structure of the Video.js player looks something like this
 
 ```tree
 Player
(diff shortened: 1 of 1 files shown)


gkatsev:
I wonder if we should just get rid of it altogether.

gkatsev:
I don't think a div is necessary. For example, in the React community, they've found a pretty nice pattern of having container and presentational components. container components usually don't have any UI associated with them while presentational components do. Though, both are React Classes. I would put the MediaLoader is a category similar to the container component, so, it doesn't require one.

PR #4075, opened 2017-02-13
fix: trap tab focus in modal when hitting s-tab

### LINKED ISSUE #4049: Can lose focus from ModalDialog when it is opened if shift-tab is hit
When the ModalDialog is opened, we focus on the ModalDialog element itself. We then trap the tab focus inside of it. This is works fine if we hit tab after the ModalDialog opens. However, we only check for tabbable items inside of the Modal, which works fine if you tab forward. However, tabbing back in this case doesn't bring you to the last element as it should. This probably needs to be special cased in the current logic.

gkatsev:
Pretend that the modal dialog is equivalent to the first
focusable/tabbable element when hitting shift-tab.

Fix #4049

### DIFF
--- a/src/js/modal-dialog.js
+++ b/src/js/modal-dialog.js
@@ -467,6 +467,10 @@ class ModalDialog extends Component {
       }
     }
 
+    if (document.activeElement === this.el_) {
+      focusIndex = 0;
+    }
+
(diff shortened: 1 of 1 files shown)


gkatsev:
Tests are passing locally, just timing out on browserstack.

PR #4086, opened 2017-02-14
Add support for withCredentials on TextTracks

schue:
## Description
This change adds support for fetching text tracks that require
security cookies to be passed in for access.

## Specific Changes proposed
The addition text-track.js checks for a crossorigin="use-credentials"
attribute on a given text track and, if defined, will set the
withCredentials option on the XHR fetch. We discovered this problem
when hosting closed caption data on AWS CloudFront which uses signed
cookies for access controls.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
I totally forgot about this PR. Meant to respond to it previously. The track element itself doesn't actually get a `crossorigin` attribute, only the video element and given how closely we tried to follow the spec, I'd be inclined to reject this PR. We probably should have the a crossorigin getter on the tech and the player and have text track's ask the tech whether crossorigin is enabled or not.

Also, I think the last two commits are what add the dist/ files, which we don't want.

gkatsev:
I think there's a lot to solve before we can accept a PR for this. We should have an issue to discuss it instead. Closing this PR for now.

PR #4088, opened 2017-02-15
fix: make mergeOptions behave the same across browsers

forbesjo:
## Description
`Object.keys()` will error on IE11 instead of default to an empty array, this will make sure that `videojs.mergeOptions()` has the same behavior across browsers.

## Specific Changes proposed
Made a `keys()` function that defaults to `[]` when given an non-object.

## Requirements Checklist
  - [ ] Docs/guides updated


forbesjo:
Will do

PR #4089, opened 2017-02-15
fix: videojs.hooks and videojs.hook should have consistent return values.

misteroneill:
I noticed that while `videojs.hook` simply calls `videojs.hooks`, it does not return the same value.

## Requirements Checklist

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -163,9 +163,12 @@ videojs.hooks = function(type, fn) {
  *
  * @param {Function|Function[]}
  *        The function or array of functions to attach.
+ *
+ * @return {Array}
+ *         an array of hooks, or an empty array if there are none.
  */
 videojs.hook = function(type, fn) {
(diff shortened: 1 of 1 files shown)


gkatsev:
I think the difference here is that `hook` is a setter, "add a hook", and `hooks` is a getter "get hooks for this type".
Because the difference is slight, a decision was made at the time to have the getter also act as a setter, rather than log an error/warning.
Should we change this so there's only one combined getter/setter?

BrandonOCasey:
that may make more sense

misteroneill:
Alternatively, rename `hooks` to `getHooks` and `hook` to `addHook`?

misteroneill:
Actually, this doesn't seem worth it. It was my mis-reading of the code that caused a misunderstanding. Might be something to tackle in a 6.x minor, but I don't want to add to confusion/chaos.

PR #4090, opened 2017-02-15
fix: make mergeOptions behave the same across browsers (5.x)

forbesjo:
## Description
Object.keys() will error on IE11 instead of default to an empty array, this will make sure that videojs.mergeOptions() has the same behavior across browsers.

## Specific Changes proposed
Made a keys() function that defaults to [] when given an non-object.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/obj.js
+++ b/src/js/utils/obj.js
@@ -30,6 +30,18 @@
  */
 const toString = Object.prototype.toString;
 
+/**
+ * Get the keys of an Object
+ *
+ * @param {Object}
+ *        The Object to get the keys from
+ *
+ * @return {string[]}
+ *         An array of the keys from the object. Returns an empty array if the
+ *         object passed in was invalid or had no keys.
+ *
+ * @private
+ */
 const keys = function(object) {
   return isObject(object) ? Object.keys(object) : [];
 };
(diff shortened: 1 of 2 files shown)


src/js/utils/obj.js
```diff
@@ -30,6 +30,10 @@
  */
 const toString = Object.prototype.toString;
 
+const keys = function(object) {
```
misteroneill: Can we get a JSDoc block for this function, with `@private`?

src/js/utils/obj.js
```diff
@@ -30,6 +30,17 @@
  */
 const toString = Object.prototype.toString;
 
+/*
+ * Gives keys of an object
+ *
+ * @param {Object} object
```
BrandonOCasey: The doc comments should probably look something like this to follow a similar format for the project. I more descriptive language as well.
```js
/**
 * Get the keys of an Object
 *  
 * @param {Object}
 *        The Object to get the keys from
 *
 * @return {string[]}
 *         An array of the keys from the object. Returns an empty array if the 
 *         object passed in was invalid or had no keys.
 * 
 * @private
 */
```

forbesjo: Thank you, shamelessly copy/pasted

PR #4097, opened 2017-02-16
fix: Do not create element for MediaLoader

mister-ben:
## Description
`MediaLoader` has a div that's unnecessary. See #4070

## Specific Changes proposed
- Do not create an element for `MediaLoader`
- Update `Component#dispose()` to check if it has an el before trying to remove it

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/tech/loader.js
+++ b/src/js/tech/loader.js
@@ -4,6 +4,7 @@
 import Component from '../component.js';
 import Tech from './tech.js';
 import toTitleCase from '../utils/to-title-case.js';
+import mergeOptions from '../utils/merge-options.js';
 
 /**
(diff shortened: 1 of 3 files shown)


gkatsev:
so *that's* where that div comes from.

src/js/tech/loader.js
```diff
@@ -26,7 +27,10 @@ class MediaLoader extends Component {
    *        The function that is run when this component is ready.
    */
   constructor(player, options, ready) {
-    super(player, options, ready);
+    // MediaLoader has no element
+    const options_ = mergeOptions({createEl: false}, options);
```
gkatsev: I totally didn't know that we had a `createEl` options.

Issue #4064, opened 2017-02-11
Clicking the mute button also sets the volume slider to 0, but not the ARIA value of the slider


OwenEdwards:
## Description
Clicking on the mute button changes its icon and also updates the volume slider to the 0 (leftmost) position. But using a screen reader, the volume level is still read out as its previous value.

Please include any additional information necessary here. Including the following:
#### videojs
v6.0.0-RC.2
#### browsers
IE11 & JAWS
#### OSes
Windows 7
### plugins
_(none)_

OwenEdwards:
This is related to, but distinct from, #4056.

gkatsev:
@kevinlitchfield can I ask you to take a look at this one as well?

kevinlitchfield:
@gkatsev Sure thing!

PR #4099, opened 2017-02-17
Muting with `MuteToggle` sets ARIA value of `VolumeBar` to 0

kevinlitchfield:
Currently, the ARIA value of `VolumeBar` tracks the value of `volume`. It should instead track the position of the slider on `VolumeBar`, which tracks `volume` except when the player is muted, in which case it's 0.

Addresses https://github.com/videojs/video.js/issues/4064.

### Question about testing

Clicking `MuteToggle` fires the `volumechange` event on the player, which in turn runs `VolumeBar#updateARIAAttributes`.

However, in the testing environment—at least as I have the tests set up!—clicking `MuteToggle` doesn't fire `volumechange` on the player, so I've had to manually call `VolumeBar#updateARIAAttributes` where it should have been run automatically in order to get the tests to pass. Of course, this needs to be fixed before merging.

Do any problems with my test configuration stick out, or is this likely a bug?

### DIFF
--- a/src/js/control-bar/volume-control/volume-bar.js
+++ b/src/js/control-bar/volume-control/volume-bar.js
@@ -107,9 +107,15 @@ class VolumeBar extends Slider {
   updateARIAAttributes(event) {
     // Current value of volume bar as a percentage
     const volume = Math.round((this.player_.volume() * 100)).toString();
+    const muted = this.player_.muted();
 
-    this.el_.setAttribute('aria-valuenow', volume);
-    this.el_.setAttribute('aria-valuetext', volume + '%');
+    if (muted) {
+      this.el_.setAttribute('aria-valuenow', 0);
+      this.el_.setAttribute('aria-valuetext', 0 + '%');
+    } else {
+      this.el_.setAttribute('aria-valuenow', volume);
+      this.el_.setAttribute('aria-valuetext', volume + '%');
+    }
   }
 
   /**
--- a/src/js/control-bar/volume-control/volume-bar.js
+++ b/src/js/control-bar/volume-control/volume-bar.js
@@ -105,17 +105,19 @@ class VolumeBar extends Slider {
    * @listens Player#volumechange
    */
   updateARIAAttributes(event) {
-    // Current value of volume bar as a percentage
-    const volume = Math.round((this.player_.volume() * 100)).toString();
-    const muted = this.player_.muted();
+    const ariaValue = this.player_.muted() ? 0 : this.volumeAsPercentage_();
 
-    if (muted) {
-      this.el_.setAttribute('aria-valuenow', 0);
-      this.el_.setAttribute('aria-valuetext', 0 + '%');
-    } else {
-      this.el_.setAttribute('aria-valuenow', volume);
-      this.el_.setAttribute('aria-valuetext', volume + '%');
-    }
+    this.el_.setAttribute('aria-valuenow', ariaValue);
+    this.el_.setAttribute('aria-valuetext', ariaValue + '%');
+  }
+
+  /**
+   * Returns the current value of the player volume as a percentage
+   *
+   * @private
+   */
+  volumeAsPercentage_() {
+    return Math.round((this.player_.volume() * 100)).toString();
   }
 
   /**
--- a/src/js/control-bar/volume-control/volume-bar.js
+++ b/src/js/control-bar/volume-control/volume-bar.js
@@ -117,7 +117,7 @@ class VolumeBar extends Slider {
    * @private
    */
   volumeAsPercentage_() {
-    return Math.round((this.player_.volume() * 100)).toString();
+    return Math.round(this.player_.volume() * 100);
   }
 
   /**
--- a/test/unit/controls.test.js
+++ b/test/unit/controls.test.js
@@ -8,8 +8,16 @@ import FullscreenToggle from '../../src/js/control-bar/fullscreen-toggle.js';
 import TestHelpers from './test-helpers.js';
 import document from 'global/document';
 import Html5 from '../../src/js/tech/html5.js';
+import sinon from 'sinon';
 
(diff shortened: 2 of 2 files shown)


src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -107,9 +107,15 @@ class VolumeBar extends Slider {
   updateARIAAttributes(event) {
     // Current value of volume bar as a percentage
     const volume = Math.round((this.player_.volume() * 100)).toString();
```
gkatsev: what about changing this `const` to a `let`, and then setting it to `0`, if `muted` is true? That way, we aren't duplicated the setting of the `aria-volumenow` and `aria-valuetext`

kevinlitchfield: Ah, yeah, that's much better! It occurred to me that perhaps using a conditional could make `updateARIAAttributes` even shorter and also, by extracting the statement that returns volume as a percentage into a named method, eliminate the need for an explanatory comment; please see https://github.com/videojs/video.js/pull/4099/commits/0c5f3cc5b8873c93774d885c6832152290bbfd59 and let me know what you think.

test/unit/controls.test.js
```diff
@@ -138,4 +147,34 @@ if (Html5.isSupported()) {
...
+
+    this.clock.tick(1);
+
+    assert.equal(player.volume(), 1, 'Volume is 1');
+    assert.equal(player.muted(), false, 'Muted is false');
+    assert.equal(volumeBar.el_.getAttribute('aria-valuenow'), 100, 'ARIA value of VolumeBar is 100');
+
+    muteToggle.handleClick();
+
+    volumeBar.updateARIAAttributes();
```
gkatsev: is this still required after the tick update?

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -105,11 +105,19 @@ class VolumeBar extends Slider {
...
+    this.el_.setAttribute('aria-valuenow', ariaValue);
+    this.el_.setAttribute('aria-valuetext', ariaValue + '%');
+  }
+
+  /**
+   * Returns the current value of the player volume as a percentage
+   *
+   * @private
+   */
+  volumeAsPercentage_() {
```
gkatsev: :+1:

kevinlitchfield:
@gkatsev Thanks for the suggestion about fake timers! That worked in two of the places I was calling `updateARIAAttributes` manually, but there's one place it didn't work: after `muteToggle.handleClick()` in the second test.

test/unit/controls.test.js
kevinlitchfield: Yeah, see comment below. The last assertion ("ARIA value is 0") fails with that line removed or replaced with another tick update.

gkatsev: Probably because it's async. One thing that might be more accurate is to trigger a `volumechange` on `player` manually, but probably not much of a difference.

src/js/control-bar/volume-control/volume-bar.js
```diff
@@ -105,11 +105,19 @@ class VolumeBar extends Slider {
    * @listens Player#volumechange
    */
   updateARIAAttributes(event) {
-    // Current value of volume bar as a percentage
-    const volume = Math.round((this.player_.volume() * 100)).toString();
+    const ariaValue = this.player_.muted() ? 0 : this.volumeAsPercentage_();
 
-    this.el_.setAttribute('aria-valuenow', volume);
-    this.el_.setAttribute('aria-valuetext', volume + '%');
+    this.el_.setAttribute('aria-valuenow', ariaValue);
```
OwenEdwards: Does it matter that you haven't converted `ariaValue` to a string by this point (where the previous code used the `.toString()` method on the numeric value)? According to [Using the aria-valuenow attribute - Accessibility | MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuenow_attribute), the value should be "_String representation of a number_". (The `aria-valuetext` will become a string thanks to the `+ '%'`)

kevinlitchfield: @OwenEdwards Thanks for the suggestion. I don't think this does matter, though. The number gets coerced to a string when being set on the element even without `+ '%'` there to more visibly do the coercion. Tried changing everything to strings just in case and there was no difference.

test/unit/controls.test.js
kevinlitchfield: Yeah, that's better. Have updated. Think there's any way around the async thing?

gkatsev: Nope, unless you want to change the QUnit test to be an async test, which is generally frowned upon. This should be fine, though.

kevinlitchfield: OK. I added a comment explaining the reason for the`player.trigger('volumechange')` line, let me know how that looks to you.

test/unit/controls.test.js
```diff
@@ -138,4 +147,37 @@ if (Html5.isSupported()) {
...
+
+    this.clock.tick(1);
+
+    assert.equal(player.volume(), 1, 'Volume is 1');
+    assert.equal(player.muted(), false, 'Muted is false');
+    assert.equal(volumeBar.el_.getAttribute('aria-valuenow'), 100, 'ARIA value of VolumeBar is 100');
+
+    muteToggle.handleClick();
+
+    // Because `MuteToggle#handleClick` is async, the `volumechange` event
```
gkatsev: It's the triggering of `volumechange` that is asynchronous, but otherwise, 👍 to the comment.

kevinlitchfield: Oops, thanks! Thoughts on update?

gkatsev: Sounds good.

PR #4100, opened 2017-02-17
fix: copy basic plugin properties onto the wrapper

BrandonOCasey:
## Description
Copy basic plugin properties to prevent breaking older basic plugins

## Requirements Checklist

### DIFF
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -312,6 +312,9 @@ class Plugin {
     if (name !== BASE_PLUGIN_NAME) {
       if (Plugin.isBasic(plugin)) {
         Player.prototype[name] = createBasicPlugin(name, plugin);
+        Object.keys(plugin).forEach(function(prop) {
+          Player.prototype[name][prop] = plugin[prop];
+        });
       } else {
         Player.prototype[name] = createPluginFactory(name, plugin);
       }
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -88,17 +88,25 @@ const markPluginAsActive = (player, name) => {
  * @returns {Function}
  *          A wrapper function for the given plugin.
  */
-const createBasicPlugin = (name, plugin) => function() {
-  const instance = plugin.apply(this, arguments);
+const createBasicPlugin = function(name, plugin) {
+  const basicPluginWrapper = function() {
+    const instance = plugin.apply(this, arguments);
 
-  markPluginAsActive(this, name);
+    markPluginAsActive(this, name);
 
-  // We trigger the "pluginsetup" event on the player regardless, but we want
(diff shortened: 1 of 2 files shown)


src/js/plugin.js
```diff
@@ -312,6 +312,9 @@ class Plugin {
     if (name !== BASE_PLUGIN_NAME) {
       if (Plugin.isBasic(plugin)) {
         Player.prototype[name] = createBasicPlugin(name, plugin);
+        Object.keys(plugin).forEach(function(prop) {
+          Player.prototype[name][prop] = plugin[prop];
+        });
```
misteroneill: I feel like this belongs in the `createBasicPlugin` function.

BrandonOCasey: I thoughts that too at first, but `Player.prototype[name]` does not exist until we get the return value from `createBasicPlugin`

BrandonOCasey: If we really want it in createBasicPlugin, then we have to do the `Player.prototype[name]` assign in there as well. I don't see any problems with that. Thoughts?

gkatsev: couldn't it just copy the props from `plugin` onto the `instance` that `createBasicPlugin` returns`?

gkatsev: `createBasicPlugin` is a function that [returns a function](https://github.com/videojs/video.js/blob/afa747c2fdeca36aa971f4705dc121f94549daf4/src/js/plugin.js#L91). It's a bit confusing.
But `instance` is inside the returned function, you'd want to do the property copying in the outer arrow function instead.

gkatsev: I almost thought just now that `createBasicPlugin` initialized basic plugins when called, but it doesn't do that.

BrandonOCasey: yeah calling the function that `createBasicPlugin` returns on player is actually what initializes. The question is do we want to copy the methods that are on the plugin right away (this is how it used to work in 5.x)? or only after the plugin has been initialized. 

Examples:

After initialization copy properties
```js
var basic = () => {};
basic.VERSION = '9.9.9';
videojs.registerPlugin('basic', basic);
...
console.log(player.basic.VERSION);
// undefined
player.basic();

console.log(player.basic.VERSION);
// 9.9.9
```

Before initialization copy properties
```js
var basic = () => {};
basic.VERSION = '9.9.9';
videojs.registerPlugin('basic', basic);
...
console.log(player.basic.VERSION);
// '9.9.9'
```

gkatsev: I think we want to mimic current behavior, which means that we want to copy it before init and the plugin would need to handle things after init, like they do now.

gkatsev:
Tests are fine except that IE8 is being weird on browserstack.

Issue #4093, opened 2017-02-15
Adding `vtt.js`'s exports to `window` breaks `addCue` in Chrome


justinanastos:
## Description
In [`tech.js#addWebVttScript_`](https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L515-L521), all of the exports from `vtt.js` are blindly added to `window`. This will break the ability to call Chrome's native `TextTrack.prototype.addCue()` because Chrome expects an instance of it's native `VTTCue` and not the `vtt.js` version.

This will effect all use cases of calling `textTrack.addCue()`. I found the issue while trying to restore text track functionality to `videojs-contrib-dash` in https://github.com/videojs/videojs-contrib-dash/pull/135. The details of the error can be seen specifically in this comment https://github.com/videojs/videojs-contrib-dash/pull/135#issuecomment-278740001.

### Reduced Test Case

http://jsbin.com/tomegu/7/edit?html,js,output

The console will show the error thrown.

### Expected
I expect `TextTrack.addCue(new VTTCue( ... ))` to add cues.

### Actual
Chrome will throw an error, shown below.

### Error output
```
Uncaught TypeError: Failed to execute 'addCue' on 'TextTrack': parameter 1 is not of type 'TextTrackCue'.
```

#### videojs
I haven't checked any versions besides `v5.16`, but the same code exists in `master` so I assume all of the version.

#### browsers
Chrome

#### OSes
Only tested on MacOS 10.12.3 (latest as of this writing)

### plugins
None

## Resolution

I propose that only values that do not exist on `window` are defined when iterating through `vtt.js`'s exports.

This reduced test case shows a cue being added as expected when using the native `VTTCue` in Chrome.

http://jsbin.com/yelizon/5/edit?html,js,output

gkatsev:
Yeah, that sounds like an issue. Historically, we haven't needed to handle this because either native tracks or emulated tracks would be used. The current proposed fix, unfortunately, will break emulated caption support, at least with the current version of vttjs that we're using. If it doesn't, let me know I'm wrong!
We already have an issue for updating vttjs (https://github.com/videojs/video.js/issues/3819) and we recently updated how vttjs is included. However, it seems like we need to spend some more time on this, potentially making vttjs be much more contained within tracks.

justinanastos:
[...]

PR #4115, opened 2017-02-21
Update vttjs

gkatsev:
Fixes https://github.com/videojs/video.js/issues/4093

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -336,14 +336,14 @@ class TextTrack extends Track {
    * @param {TextTrack~Cue} cue
    *        The cue to add to our internal list
    */
-  addCue(_cue) {
-    let cue = _cue;
+  addCue(originalCue) {
+    let cue = originalCue;
 
-    if (!(_cue instanceof window.vttjs.VTTCue)) {
-      cue = new window.vttjs.VTTCue(_cue.startTime, _cue.endTime, _cue.text);
+    if (!(originalCue instanceof window.vttjs.VTTCue)) {
+      cue = new window.vttjs.VTTCue(originalCue.startTime, originalCue.endTime, originalCue.text);
 
-      for (const prop in _cue) {
-        cue[prop] = _cue[prop];
+      for (const prop in originalCue) {
+        cue[prop] = originalCue[prop];
       }
     }
 
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -336,7 +336,17 @@ class TextTrack extends Track {
    * @param {TextTrack~Cue} cue
    *        The cue to add to our internal list
    */
-  addCue(cue) {
+  addCue(_cue) {
+    let cue = _cue;
+
+    if (!(_cue instanceof vttjs.VTTCue)) {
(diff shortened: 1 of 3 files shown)


justinanastos:
@gkatsev I find that this fixes the issue I'd seen previously in #4093. I see this is targeted at `master`, will this be applied to the 5.x branch as well?

gkatsev:
Yup, I'll make another PR against `5.x` when this is complete. Thanks for verifying that it works for you!

justinanastos:
You're welcome @gkatsev !

src/js/tracks/text-track.js
```diff
@@ -336,7 +336,17 @@ class TextTrack extends Track {
    * @param {TextTrack~Cue} cue
    *        The cue to add to our internal list
    */
-  addCue(cue) {
+  addCue(_cue) {
+    let cue = _cue;
+
+    if (!(_cue instanceof window.vttjs.VTTCue)) {
+      cue = new window.vttjs.VTTCue(_cue.startTime, _cue.endTime, _cue.text);
```
BrandonOCasey: would it be possible/better to expose this on videojs rather than window?

src/js/tracks/text-track.js
```diff
@@ -336,7 +336,17 @@ class TextTrack extends Track {
    * @param {TextTrack~Cue} cue
    *        The cue to add to our internal list
    */
-  addCue(cue) {
+  addCue(_cue) {
+    let cue = _cue;
```
BrandonOCasey: can we change this to `cueCopy` and change `_cue` back to just `cue`?

gkatsev: We could. I just thought it would be better to leave the rest of the code as is and keep the change relatively self-contained.

src/js/tracks/text-track.js
gkatsev: it's more or less exposed globally right now plus we've expected it to be exposed. For the 5.x version, we definitely should keep it exposed on window because it'll be easier to make it backwards compat.
For this PR, we could look into making a version of vttjs that doesn't bind itself to `window`. Though, I think it should be done as a separate PR rather than part of this work.

misteroneill: My feeling here is that vttjs is kind of its own thing - we just use it and it should stand alone more or less.

gkatsev: @misteroneill sort of. Given that we're using a custom fork, we can do what we want with it at this point.

The only thing I can imagine that would be tricky to get right to have vttjs not export itself globally is getting the novtt build working correctly.

src/js/tracks/text-track.js
BrandonOCasey: I think that it's more a self documenting code change that would be worth doing, as `_cue` doesn't really have any significance to tell you what it actually is.

src/js/tracks/text-track.js
BrandonOCasey: Im open to doing it in another PR

src/js/tracks/text-track.js
gkatsev: What about naming it `originalCue`?

BrandonOCasey: sure that works too

PR #4116, opened 2017-02-21
fix: addChild instance names should be toTitleCased

gkatsev:


gkatsev:
Probably.

gkatsev:
IE is timing out, tests pass locally.

PR #4117, opened 2017-02-21
fix: addChild instance names should be toTitleCased

gkatsev:


BrandonOCasey:
See the other PR for my comment on unit test

PR #4125, opened 2017-02-22
feat: add back MenuButton#controlText

BrandonOCasey:
## Description
In 5.x this function existed. I think that it was an oversight to remove it so, I have brought it back.

## Requirements Checklist


gkatsev:
IE is timing out, merging anyway.

PR #4134, opened 2017-02-24
fix: early play should wait for player ready, even if source is available

### LINKED ISSUE #4057: Trying to play the player too early before the tech is ready, won't play video when tech loads
Basically, it ends up getting into [this block](https://github.com/videojs/video.js/blob/master/src/js/player.js#L1615-L1624) and not actually playing when the player is ready and the video loads.

gkatsev:
Fixes #4057

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1616,7 +1616,7 @@ class Player extends Component {
       });
 
     // Only calls the tech's play if we already have a src loaded
-    } else if (this.src() || this.currentSrc()) {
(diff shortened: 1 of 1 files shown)


gkatsev:
I can take a look at adding a unit test for this. We probably can do it. I'll look into it after I finish fixing the current player.test.js tests.

Isabela96:
Looks great

gkatsev:
Should be rebased against master, should make tests run better.

PR #4139, opened 2017-02-28
fix: dom.getBoundingClientRect check that el is defined

BrandonOCasey:
## Description
Verify that `el` is defined so that we don't throw if we get to this function with an undefined `el`
 
## Requirements Checklist


gkatsev:
This makes sense to have but why does the usage end up being called without an el?

gkatsev:
Should be rebased against master, should make tests run better.

gkatsev:
Rebased

PR #4140, opened 2017-02-28
Don't throw when re-registering a plugin

misteroneill:
Currently, if a plugin is re-registered, it will throw an error. On the face, this makes sense, but it comes with two issues:

1. It will totally break in an unrecoverable way in a case where it might not actually be a break.
2. Nothing else has this behavior - components and techs do not throw.

## Requirements Checklist

### DIFF
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -5,6 +5,7 @@ import evented from './mixins/evented';
 import stateful from './mixins/stateful';
 import * as Events from './utils/events';
 import * as Fn from './utils/fn';
+import log from './utils/log';
 import Player from './player';
 
 /**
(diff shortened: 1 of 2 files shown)


src/js/plugin.js
```diff
@@ -305,8 +306,10 @@ class Plugin {
       throw new Error(`Illegal plugin name, "${name}", must be a string, was ${typeof name}.`);
     }
 
-    if (pluginExists(name) || Player.prototype.hasOwnProperty(name)) {
-      throw new Error(`Illegal plugin name, "${name}", already exists.`);
+    if (pluginExists(name)) {
+      log.warn(`A plugin named "${name}" already exists. You may want to avoid re-registering plugins!`);
+    } else if (Player.prototype.hasOwnProperty(name)) {
```
misteroneill: This is an `else if` because if a plugin exists, it will have a method on the `Player.prototype`, but if the plugin _does not exist_ **and** the name matches a `Player.prototype` method, it's an actual player method.

gkatsev: This sounds like a reasonable compromise.

gkatsev:
This should be rebased against master, will make tests run better.

src/js/plugin.js
```diff
@@ -305,8 +306,10 @@ class Plugin {
       throw new Error(`Illegal plugin name, "${name}", must be a string, was ${typeof name}.`);
     }
 
-    if (pluginExists(name) || Player.prototype.hasOwnProperty(name)) {
-      throw new Error(`Illegal plugin name, "${name}", already exists.`);
+    if (pluginExists(name)) {
+      log.warn(`A plugin named "${name}" already exists. You may want to avoid re-registering plugins!`);
```
gkatsev: @misteroneill when you get the chance, can you make another PR that silences this warning in the plugin tests?

PR #4141, opened 2017-02-28
chore(tests): make tests not print out errors

gkatsev:


gkatsev:
There is still one more error being logged in IE8 but I don't have any more time to track it down right now.

test/unit/tracks/html-track-element.test.js
```diff
@@ -27,23 +27,19 @@ QUnit.test('can create a html track element with various properties', function(a
   const kind = 'chapters';
   const label = 'English';
   const language = 'en';
-  const src = 'http://www.example.com';
```
gkatsev: It would be nice but it ends up causing an XHR which logs an error in IE8 and there isn't a good way to cancelling the track.
Text Tracks have a source is already tested elsewhere. I think not having an extra error logged for a failed XHR is better than testing something that is partially tested elsewhere anyway.

test/unit/tracks/text-tracks.test.js
```diff
@@ -365,6 +366,12 @@ QUnit.test('removes cuechange event when text track is hidden for emulated track
     mode: 'showing'
   });
 
+  const oldVttjs = window.vttjs;
```
gkatsev: this change probably should be removed from this PR in favor of https://github.com/videojs/video.js/pull/4148/files#diff-45ae6eaab623b37642b347fe7707682aR338

test/unit/tracks/html-track-element.test.js
BrandonOCasey: ok acceptable

test/unit/tracks/text-tracks.test.js
BrandonOCasey: 👍

gkatsev:
I'll wait for the master version of https://github.com/videojs/video.js/pull/4148 to be merged before merging this one so this can be updated with changes.

PR #4148, opened 2017-03-01
5.x fix cue-points with a startTime of 0

BrandonOCasey:
## Description
Wait for tech to be ready before worrying about timeupdate events

## Requirements Checklist

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -198,7 +198,9 @@ class TextTrack extends Track {
     });
 
     if (mode !== 'disabled') {
-      tt.tech_.on('timeupdate', timeupdateHandler);
+      tt.tech_.ready(() => {
+        tt.tech_.on('timeupdate', timeupdateHandler);
+      }, true);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/js/tracks/text-track.js
```diff
@@ -331,7 +335,7 @@ class TextTrack extends Track {
   addCue(originalCue) {
     let cue = originalCue;
 
-    if (!(originalCue instanceof window.vttjs.VTTCue)) {
+    if (window.vttjs && !(originalCue instanceof window.vttjs.VTTCue)) {
```
gkatsev: I was mocking vttjs out https://github.com/videojs/video.js/pull/4146/files#diff-3b4ce30c38112a433d137fb9bf1bcbc9R21 but this probably is a better choice.

test/unit/tracks/text-track.test.js
```diff
@@ -25,7 +27,7 @@ TrackBaseline(TextTrack, {
   mode: 'disabled',
   label: 'English',
   language: 'en',
-  tech: defaultTech
+  tech: new TechFaker()
```
gkatsev: should this be `this.tech`?

BrandonOCasey: its a suite of tests that runs on it's own, so it won't have access to `this.tech` as it is not in a `QUnit.test`

gkatsev: oh, I see. LGTM then.

PR #4152, opened 2017-03-02
fix: cues at startTime 0 do not fire

BrandonOCasey:
## Description
Previously `timeupdate` would fire before the video was playing, and the tech was not ready. This caused issues when `preload` was set to `auto`, because `cuechange` would fire before the video was even started for cues with a `startTime` of 0.

* Wait for tech to be ready before watching for `timeupdate`
* update unit tests to use TechFaker
* Add a unit test to verify that we wait for Tech to be ready.
## Specific Changes proposed

## Requirements Checklist

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -352,6 +352,9 @@ class TextTrack extends Track {
           cue[prop] = originalCue[prop];
         }
       }
+
+      // make sure that `id` is copied over
+      cue.id = originalCue.id;
     }
 
     const tracks = this.tech_.textTracks();
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -198,7 +198,9 @@ class TextTrack extends Track {
     });
 
     if (mode !== 'disabled') {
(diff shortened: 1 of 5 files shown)


src/js/tracks/text-track.js
```diff
@@ -339,14 +344,17 @@ class TextTrack extends Track {
   addCue(originalCue) {
     let cue = originalCue;
 
-    if (!(originalCue instanceof window.vttjs.VTTCue)) {
+    if (window.vttjs && !(originalCue instanceof window.vttjs.VTTCue)) {
       cue = new window.vttjs.VTTCue(originalCue.startTime, originalCue.endTime, originalCue.text);
 
       for (const prop in originalCue) {
         if (!(prop in cue)) {
```
BrandonOCasey: should we just do `|| prop === id` instead of copying it below? Doesn't really matter all that much I guess though.

gkatsev: I think since we're only copying the id, having it be separate with a comment makes it much nicer and explicit. Though, if we end up needing to force copy other stuff, I'd agree combining into here would make more sense.

PR #4164, opened 2017-03-04
Fix poster image to match the video in the 'combined-tracks.html' example in sandbox

OwenEdwards:
## Description
Remove poster image which doesn't match the video from the 'combined-tracks.html' example in sandbox, and also fixed a typo.

## Requirements Checklist

### DIFF
--- a/sandbox/combined-tracks.html.example
+++ b/sandbox/combined-tracks.html.example
@@ -25,7 +25,7 @@
     <pre>open http://localhost:9999/sandbox/index.html</pre>
   </div>
 
-  <video id="vid1" class="video-js vjs-default-skin" lang="en" controls preload="auto" width="640" height="264" poster="http://vjs.zencdn.net/v/oceans.png">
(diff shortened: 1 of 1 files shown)


sandbox/combined-tracks.html.example
```diff
@@ -25,7 +25,7 @@
     <pre>open http://localhost:9999/sandbox/index.html</pre>
   </div>
 
-  <video id="vid1" class="video-js vjs-default-skin" lang="en" controls preload="auto" width="640" height="264" poster="http://vjs.zencdn.net/v/oceans.png">
+  <video id="vid1" class="video-js vjs-default-skin" lang="en" controls preload="auto" width="640" height="264">
```
gkatsev: what about swapping to `//d2zihajmogu5jn.cloudfront.net/elephantsdream/poster.png`?

OwenEdwards: Oh, I didn't know that the poster for ED was on the CDN. I can make that change - thanks @gkatsev!

PR #4176, opened 2017-03-07
fix(vttjs): wait till tech el in DOM before loading vttjs

gkatsev:
The issue is that in Flash, `Flash.embed` wraps the `object` element in a div but that `object` element is what is referenced by `this.el()`, so, `parentNode` isn't null but we aren't actually in the DOM. Instead, just check to see whether the tech element is in the DOM or not via `node.contains()`.


forbesjo:
Looks good to me

PR #4186, opened 2017-03-10
fix: dont pause when the ErrorDisplay is opened

BrandonOCasey:
## Description
* Added a feature to make it optional for `ModalDialog` to `pauseOnOpen`
* Use `pauseOnOpen: false` in `ErrorDisplay` so that errors that can recover during playback will recover

## Requirements Checklist


BrandonOCasey:
ok I will mark it as a feat

PR #4200, opened 2017-03-13
fix links in docs

BrandonOCasey:
## Description
* Update the API doc build to fix links after generation
* Fix any currently broken API docs
* Used `linkdown` to make sure links are working

## Outstanding Issues (probably for another PR)
* The tech source handler mixin is currently not showing/broken in the docs
* videojs itself does not show most of its functions in the API docs
* Some things are using `Dom` when they should be using `module:dom`

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/build/fix-api-docs.js
+++ b/build/fix-api-docs.js
@@ -1,0 +1,20 @@
+var replace = require("replace");
+var path = require('path')
+var apiPath = path.join(__dirname, '..', 'docs', 'api');
+
+var replacements = [
+  {find: '\"/docs/guides/\"', replace: '"#"'},
+  {find: '\"/docs/guides/(.*)\.md\"', replace: '"tutorial-$1.html"'},
+  {find: '\"tutorial-tech.html\"', replace: '"tutorial-tech_.html"'}
+];
+
+
+replacements.forEach(function(obj) {
+  replace({
+    regex: obj.find,
+    replacement: obj.replace,
+    paths: [apiPath],
+    recursive: true,
+    silent: true
+  });
+});
--- a/build/fix-api-docs.js
+++ b/build/fix-api-docs.js
@@ -3,9 +3,9 @@ var path = require('path')
 var apiPath = path.join(__dirname, '..', 'docs', 'api');
 
 var replacements = [
-  {find: '\"/docs/guides/\"', replace: '"#"'},
-  {find: '\"/docs/guides/(.*)\.md\"', replace: '"tutorial-$1.html"'},
-  {find: '\"tutorial-tech.html\"', replace: '"tutorial-tech_.html"'}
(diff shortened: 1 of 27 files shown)


build/fix-api-docs.js
```diff
@@ -0,0 +1,20 @@
+var replace = require("replace");
+var path = require('path')
+var apiPath = path.join(__dirname, '..', 'docs', 'api');
+
+var replacements = [
+  {find: '\"/docs/guides/\"', replace: '"#"'},
+  {find: '\"/docs/guides/(.*)\.md\"', replace: '"tutorial-$1.html"'},
```
gkatsev: should this be the first item in the array? That is, do the most specific changes first and work towards more generic ones after?

build/fix-api-docs.js
```diff
@@ -0,0 +1,20 @@
...
+  {find: '\"tutorial-tech.html\"', replace: '"tutorial-tech_.html"'}
```
gkatsev: Why is this page rendered as `tech_`?

package.json
```diff
@@ -27,6 +27,7 @@
     "docs": "npm run docs:lint && npm run docs:api",
     "jsdoc": "jsdoc",
     "docs:api": "jsdoc -c .jsdoc.json",
+    "postdocs:api": "node ./build/fix-api-docs.js",
```
gkatsev: this runs immediately after `docs:api` and before the `prepublish` script finishes, as an example?

build/fix-api-docs.js
BrandonOCasey: I'm not entirely sure, I think it may have something to do with a file name tech already existing? Either that or the source handler weirdness, (jsdoc links to `-_Tech.html` on the Tech API docs)

build/fix-api-docs.js
BrandonOCasey: yeah that probably makes sense let me try it out

package.json
BrandonOCasey: Should run directly after `docs:api` and before anything else I would think. Would we rather use `npm-run-all` here so its more clear?

gkatsev: No, I think it's fine. Just wanted to verify that `prepublish` should work as we expect so the urls will get changed as part of that.

BrandonOCasey: manually verified that this will work by run `npm run prepublish`

build/fix-api-docs.js
```diff
@@ -0,0 +1,20 @@
+var replace = require("replace");
+var path = require('path')
+var apiPath = path.join(__dirname, '..', 'docs', 'api');
+
+var replacements = [
+  {find: /\/docs\/guides\/(.+)\.md/, replace: 'tutorial-$1.html'},
```
gkatsev: All of these need the global flag to apply properly to all occurrences in the file.

PR #4208, opened 2017-03-16
fix: removeCue with native/same cues

BrandonOCasey:
## Description
This PR fixes two issues:
1. If you add two of the same cues to the cues list they will both be removed by one call to `removeCue`
2. If you create an add a native cue to the cues list. You cannot use the native cue to remove the cue that was added to the cue list.

## Requirements Checklist

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -355,6 +355,7 @@ class TextTrack extends Track {
 
       // make sure that `id` is copied over
       cue.id = originalCue.id;
+      cue.originalCue_ = originalCue;
     }
 
     const tracks = this.tech_.textTracks();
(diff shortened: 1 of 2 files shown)


src/js/tracks/text-track.js
```diff
@@ -376,20 +377,17 @@ class TextTrack extends Track {
...
-    for (let i = 0, l = this.cues_.length; i < l; i++) {
+    while (i--) {
       const cue = this.cues_[i];
 
-      if (cue === removeCue) {
+      if (cue === removeCue || (cue.originalCue_ && cue.originalCue_ === removeCue)) {
         this.cues_.splice(i, 1);
-        removed = true;
+        this.cues.setCues_(this.cues_);
+        return;
```
gkatsev: probably more appropriate to use `break` here.

PR #4216, opened 2017-03-21
fix: default text tracks

BrandonOCasey:
## Description
There are currently two issues with `TextTrack`s that have `default` set to `true` when added. 
1. The TextTrackMenu has the `captions off` and the new `default` caption button both selected
2. The captions do not show up until after a re-selection

## Specific Changes proposed
1. is addressed by listening for player `loadstart`, rather than just `texttrackchange` as in most cases text tracks will be added before we start to listen for `texttrackchange`
2. Treat `addtrack` and `removetrack` similar to how we treat `change` now (watch for `cuechange` on any showing tracks and update the display) 

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/text-track-controls/text-track-menu-item.js
+++ b/src/js/control-bar/text-track-controls/text-track-menu-item.js
@@ -34,7 +34,6 @@ class TextTrackMenuItem extends MenuItem {
     super(player, options);
 
     this.track = track;
-
     const changeHandler = Fn.bind(this, this.handleTracksChange);
 
     player.on(['loadstart', 'texttrackchange'], changeHandler);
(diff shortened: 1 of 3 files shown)


src/js/control-bar/text-track-controls/text-track-menu-item.js
```diff
@@ -36,9 +36,11 @@ class TextTrackMenuItem extends MenuItem {
     this.track = track;
     const changeHandler = Fn.bind(this, this.handleTracksChange);
 
+    player.on(['loadstart', 'texttrackchange'], changeHandler);
     tracks.addEventListener('change', changeHandler);
     this.on('dispose', function() {
       tracks.removeEventListener('change', changeHandler);
+      player.off(['loadstart', 'texttrackchange'], changeHandler);
```
gkatsev: There are removed automatically.

PR #4223, opened 2017-03-22
update videojs-vtt.js to 0.12.3

mjneil:

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -551,7 +551,7 @@ class Tech extends Component {
       // passed in
       const script = document.createElement('script');
 
-      script.src = this.options_['vtt.js'] || 'http://vjs.zencdn.net/vttjs/0.12.3/vtt.min.js';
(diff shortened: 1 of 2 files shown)


src/js/tech/tech.js
```diff
@@ -551,7 +551,7 @@ class Tech extends Component {
       // passed in
       const script = document.createElement('script');
 
-      script.src = this.options_['vtt.js'] || 'https://cdn.rawgit.com/gkatsev/vtt.js/vjs-v0.12.1/dist/vtt.min.js';
+      script.src = this.options_['vtt.js'] || 'https://cdn.rawgit.com/gkatsev/vtt.js/vjs-v0.12.3/dist/vtt.min.js';
```
gkatsev: this file doesn't exist anymore since it isn't check in to github.

mjneil: What should this be updated to so the right file is loaded if options isn't passed in?

gkatsev: Looking into getting it published to the CDN.
Also, I realized that we probably want to update vttjs to get rid of `browser.js` and only use `browser-index.js`.

PR #4227, opened 2017-03-24
Minimum volume after unmuting

### LINKED ISSUE #4054: Minimum volume after unmuting
After implementation of [**`feat: unmute goes back to previously selected volume`**](https://github.com/videojs/video.js/commit/cb42fcfb02370fc7ed38499ac352b2c1630668d3), there are a couple edge cases in which unmuting will restore the volume from zero to a lower value than a user would probably anticipate or want:

1. **User drags volume to zero twice in a row.** In this situation, unmuting sets the volume to the last value it had before reaching zero on the second drag (i.e., something extremely close to zero).
1. **User steps volume to zero with keyboard after adjusting with mouse.** The keyboard steps the volume in increments of 0.1, so if the user sets the volume to 0.91 with the mouse, then steps to zero with the keyboard, then unmutes, the volume will be 0.01.

It's unlikely that a user who is trying to unmute by clicking `MuteToggle` would ever want to set the volume to such a low value as 0.01. Enforcing a minimum volume after unmuting would ensure that doesn't happen. Two important considerations occur to me in thinking about what the value of the minimum should be:

- In the situations where this is an issue, the user has actively chosen to set the volume very low before muting (or has actually muted before muting).
- If the minimum is higher than the volume a user has chosen for listening, they will no longer be able to toggle between their chosen volume and zero by clicking `MuteToggle`.

Given those things, I'd suggest 0.1 as the minimum.

Thoughts? If the maintainers agree with this approach, I'd be happy to make a PR.

kevinlitchfield:
When unmuting and `lastVolume` is less than 0.1, set volume to 0.1 instead of to `lastVolume`.

Fixes #4054.

## Before

less than 0 1

Volume set to **0.024390243902439025**.

## After

Volume set to **0.1**.

### DIFF
--- a/src/js/control-bar/mute-toggle.js
+++ b/src/js/control-bar/mute-toggle.js
@@ -57,7 +57,9 @@ class MuteToggle extends Button {
     const lastVolume = this.player_.lastVolume_();
 
     if (vol === 0) {
-      this.player_.volume(lastVolume);
+      const volumeToSet = lastVolume < 0.1 ? 0.1 : lastVolume;
+
+      this.player_.volume(volumeToSet);
       this.player_.muted(false);
     } else {
       this.player_.muted(this.player_.muted() ? false : true);
--- a/test/unit/controls.test.js
+++ b/test/unit/controls.test.js
@@ -138,4 +138,18 @@ if (Html5.isSupported()) {
     assert.equal(player.volume(), 0.5, 'volume is set to lastVolume');
     assert.equal(player.muted(), false, 'muted is set to false');
   });
+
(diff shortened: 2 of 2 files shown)


gkatsev:
Cool!

there seems to be a conflict in the test file, can you take a look at that?

kevinlitchfield:
@gkatsev Oops! Got it, thanks.

gkatsev:
Thanks @kevinlitchfield!

gkatsev:
Looks like IE9 rounding strikes again!
https://travis-ci.org/videojs/video.js/builds/214771473#L356

PR #4228, opened 2017-03-24
6.x build updates

gkatsev:


.npmignore
```diff
@@ -3,3 +3,4 @@
 !dist/**
 !es5/**
 !src/css/**
+!docs/api/**
```
gkatsev: make sure that docs are published to npm

package.json
```diff
@@ -32,8 +32,9 @@
     "docs:lint": "remark -- './**/*.md'",
     "docs:fix": "remark --output -- './**/*.md'",
     "babel": "babel src/js -d es5",
-    "prepublish": "not-in-install && npm run docs:api || in-install",
-    "prepush": "npm run lint -- --errors"
+    "prepublish": "not-in-install && run-p docs:api build || in-install",
```
gkatsev: run docs and the build in parallel if we're publishing.

package.json
```diff
@@ -32,8 +32,9 @@
...
+    "prepush": "npm run lint -- --errors",
+    "version": "node build/version.js && git add CHANGELOG.md"
```
gkatsev: run the changelog in version only if we're not in prerelease, afterwards, add it to git.

PR #4236, opened 2017-03-28
Tracks settings updates

gkatsev:
Make the font column of the text track settings dialog also use fieldset and legends like the colors column. Also, don't set a width on the label and legend elements.
This makes it so that the two columns line themselves up similarly.

### DIFF
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -255,9 +255,6 @@ class TextTrackSettings extends ModalDialog {
     options.temporary = false;
 
     super(player, options);
-
-    this.contentEl().className += '  vjs-caption-settings';
-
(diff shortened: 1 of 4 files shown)


src/js/tracks/text-track-settings.js
```diff
@@ -306,14 +306,14 @@ class TextTrackSettings extends ModalDialog {
    *         The DOM element that gets created.
    * @private
    */
-  createElSelect_(key, legendId = '') {
+  createElSelect_(key, legendId = '', type='label') {
```
misteroneill: Linter is complaining about the `=` in `type='label'` not having spaces. That's a hard one for me to follow, too, coming from some Python experience.

gkatsev: Oops, will update.

PR #4237, opened 2017-03-29
fix: remove focus ring from player itself

gkatsev:
The focus ring on the player itself isn't necessary and can be confusing and distracting to sighted users. This PR removes the focus ring from the player element itself but not from other elements.


BrandonOCasey:
Does this have to be done on 5.x?

gkatsev:
No, we don't have focus rings on 5.x.

PR #4255, opened 2017-04-05
feat: Add 'beforepluginsetup' event and named plugin setup events (e.g. 'pluginsetup:foo')

misteroneill:
This adds a `beforepluginsetup` event as well as `beforepluginsetup:$name` and `pluginsetup:$name` events.

The drive behind this is improving the ability for people to make cross-plugin dependencies in a more robust manner.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -86,7 +86,8 @@ const markPluginAsActive = (player, name) => {
  *         A plugin event hash.
  *
  * @param  {Boolean} [before]
- *         If true, modifies the events to be "before" events.
+ *         If true, prefixes the event name with "before". In other words,
+ *         use this to trigger "beforepluginsetup" instead of "pluginsetup".
  */
 const triggerSetupEvent = (player, hash, before) => {
   const eventName = (before ? 'before' : '') + 'pluginsetup';
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -76,6 +76,26 @@ const markPluginAsActive = (player, name) => {
 };
 
 /**
+ * Triggers a pair of plugin setup events.
+ *
+ * @private
+ * @param  {Player} player
+ *         A Video.js player instance.
+ *
+ * @param  {Plugin~PluginEventHash} hash
+ *         A plugin event hash.
+ *
+ * @param  {Boolean} [before]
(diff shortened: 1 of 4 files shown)


src/js/plugin.js
```diff
@@ -76,6 +76,26 @@ const markPluginAsActive = (player, name) => {
...
+ *
+ * @private
+ * @param  {Player} player
+ *         A Video.js player instance.
+ *
+ * @param  {Plugin~PluginEventHash} hash
+ *         A plugin event hash.
+ *
+ * @param  {Boolean} [before]
+ *         If true, modifies the events to be "before" events.
```
BrandonOCasey: I think it would be good to have an example, as this confused me at first. Maybe `I.E `pluginsetup` -> `beforepluginsetup``

src/js/plugin.js
```diff
@@ -91,15 +111,20 @@ const markPluginAsActive = (player, name) => {
...
 const createBasicPlugin = function(name, plugin) {
   const basicPluginWrapper = function() {
+
+    // We trigger the "beforepluginsetup" and "pluginsetup" events on the player
+    // regardless, but we want the hash to be consistent with the hash provided
+    // for advanced plugins.
+    //
+    // The only potentially counter-intuitive thing here is the `instance` in
+    // the "pluginsetup" event is the value returned by the `plugin` function.
+    triggerSetupEvent(this, {name, plugin, instance: null}, true);
```
BrandonOCasey: I'm up in the air about wether we should pass data along with the event. I understand why we would need to but, it still goes against most of the practices that we have for other events. Anyone else want to chime in here?

misteroneill: Well, it's already the case for `pluginsetup`.

BrandonOCasey: true, I guess this is just what we have to do

gkatsev: Also, I do know some plugins do return something from the method, so, providing it is nice.

misteroneill:
We could potentially add `after` events at some point, but, similar to the modal events, the "unprefixed" events are already kind of `after` events as it is.

misteroneill:
Also, I plan to add tests for this at some point this week.

PR #4258, opened 2017-04-05
feat(VttLoader): a better way of loading vtt.js

gkatsev:
This is the first commit in that feature. Still needs to be exposed
properly on video.js. Also, vtt.js needs to be updated so that the
WebVTT object is properly available on the vttjs object in the CDN
version of the file.
Afterwards, parseCues in TextTrack should be updated to use the new way
of getting vttjs.

This also means that if vttjs is being loaded via script, it won't load until a player is initialized and it is using a tech that requires emulated text tracks. So, if a player loads and only requests native text tracks, it won't ever try to load vttjs.
It also means that a user can disable loading of vttjs if using the novtt script by requesting that the component be unavailable.
```js
const player = videojs('vid', {
  vttLoader: false
});
```

It is now exported on `videojs` as an object with two properties, which are functions: `get` and `loaded`.
```js
if (videojs.vttjs.loaded()) {
  const vttjs = videojs.vttjs.get();
  /* use vttjs */
}
```

## Todos
* [x] expose vttjs on videojs object as `videojs.vttjs`
* [x] add videojs hooks for `vttjsloaded`
    this is so because vttjs isn't really a per-player thing. The only events aren't going away but should be considered deprecated.
* [x] if loaded vttjs via script, check to see if it has loaded previously.
* [ ] Tests!!!
* [ ] port to 5.x branch

@imbcmdth @gesinger @mjneil this is the start of exporting vttjs properly.

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -11,7 +11,7 @@ import { isCrossOrigin } from '../utils/url.js';
 import XHR from 'xhr';
 import merge from '../utils/merge-options';
 import * as browser from '../utils/browser.js';
-import {vttjsLoaded, getVttjs} from './vtt.js';
+import {vttjsLoaded} from './vtt.js';
 
 /**
  * Takes a webvtt file contents and parses it into cues
@@ -91,18 +91,18 @@ const loadTrack = function(src, track) {
     // NOTE: this is only used for the alt/video.novtt.js build
     if (vttjsLoaded) {
       parseCues(responseBody, track);
-    } else {
-      if (track.tech_) {
-        const loadHandler = () => parseCues(responseBody, track);
-
-        track.tech_.on('vttjsloaded', loadHandler);
-        track.tech_.on('vttjserror', () => {
-          log.error(`vttjs failed to load, stopping trying to process ${track.src}`);
-          track.tech_.off('vttjsloaded', loadHandler);
-        });
-      }
+      return;
     }
 
+    if (track.tech_) {
+      const loadHandler = () => parseCues(responseBody, track);
+
+      track.tech_.on('vttjsloaded', loadHandler);
+      track.tech_.on('vttjserror', () => {
+        log.error(`vttjs failed to load, stopping trying to process ${track.src}`);
+        track.tech_.off('vttjsloaded', loadHandler);
+      });
+    }
   }));
 };
 
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -11,6 +11,7 @@ import { isCrossOrigin } from '../utils/url.js';
 import XHR from 'xhr';
 import merge from '../utils/merge-options';
 import * as browser from '../utils/browser.js';
+import {vttjsLoaded, getVttjs} from './vtt.js';
 
 /**
  * Takes a webvtt file contents and parses it into cues
@@ -88,7 +89,9 @@ const loadTrack = function(src, track) {
 
     // Make sure that vttjs has loaded, otherwise, wait till it finished loading
     // NOTE: this is only used for the alt/video.novtt.js build
-    if (typeof window.WebVTT !== 'function') {
+    if (vttjsLoaded) {
+      parseCues(responseBody, track);
+    } else {
       if (track.tech_) {
         const loadHandler = () => parseCues(responseBody, track);
 
@@ -97,10 +100,7 @@ const loadTrack = function(src, track) {
           log.error(`vttjs failed to load, stopping trying to process ${track.src}`);
           track.tech_.off('vttjsloaded', loadHandler);
         });
-
       }
-    } else {
-      parseCues(responseBody, track);
     }
 
   }));
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -11,7 +11,7 @@ import { isCrossOrigin } from '../utils/url.js';
 import XHR from 'xhr';
 import merge from '../utils/merge-options';
 import * as browser from '../utils/browser.js';
-import {vttjsLoaded} from './vtt.js';
+import {vttjsLoaded, getVttjs} from './vtt.js';
 
 /**
  * Takes a webvtt file contents and parses it into cues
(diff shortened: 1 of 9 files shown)


misteroneill:
[test](https://trello.com/c/36nTSxgT/1-test)

gkatsev:
[Export vtt.js in a way that is consumable by all users of video.js](https://trello.com/c/4TeNd1uW/4-export-vtt-js-in-a-way-that-is-consumable-by-all-users-of-video-js)

gkatsev:
This requires vttjs updated with this PR: https://github.com/videojs/vtt.js/pull/12

src/js/video.js
```diff
@@ -147,6 +148,15 @@ videojs.hooks_ = {};
  */
 videojs.hooks = function(type, fn) {
   videojs.hooks_[type] = videojs.hooks_[type] || [];
+  if (fn && type === 'vttjsloaded') {
+    vttjsOnLoad((vttjs) => {
+      const vttjsLoadedHooks = videojs.hooks('vttjsloaded');
+
+      vttjsLoadedHooks.forEach((hookFn) => hookFn(vttjs));
+      // we notified the listeners, now clear them out
+      vttjsLoadedHooks.length = 0;
```
misteroneill: So this prevents the hooks from being called multiple times? Because it seems like otherwise you'd call each hook `vttjsLoadedHooks.length` times.

gkatsev: Yup, it's to make sue that each of the `vttjsloaded` hooks is only ever called the one time.

src/js/tracks/vtt.js
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * The `VttLoader` is the `Component` that decides how to load in vttjs
+ *
+ * @extends Component
+ * @fires VttLoader#vttjsloaded
+ * @fires VttLoader#vttjserror
+ */
+class VttLoader extends Component {
+  constructor(player, options, ready) {
+    const options_ = mergeOptions({createEl: false}, options);
```
BrandonOCasey: I think this is named `settings` in other places, this probably makes more sense, but for consistency we may want to do that.

src/js/tracks/vtt.js
```diff
@@ -0,0 +1,136 @@
...
+  triggerLoaded() {
+    vttjsLoaded = true;
+
+    /**
+     * Fired when vtt.js is loaded.
+     *
+     * @event VttLoader#vttjsloaded
+     * @type {EventTarget~Event}
+     */
+    this.trigger('vttjsloaded');
```
BrandonOCasey: Do we want to trigger this event here? seems linke we only listen for this on tech

src/js/tracks/vtt.js
```diff
@@ -0,0 +1,136 @@
...
+  triggerError() {
+    vttjsLoaded = false;
+
+    /**
+     * Fired when vtt.js was not loaded due to an error
+     *
+     * @event VttLoader#vttjserror
+     * @type {EventTarget~Event}
+     */
+    this.trigger('vttjserror');
```
BrandonOCasey: Do we want to trigger this event here? seems linke we only listen for this on tech

src/js/video.js
```diff
@@ -147,6 +148,15 @@ videojs.hooks_ = {};
  */
 videojs.hooks = function(type, fn) {
   videojs.hooks_[type] = videojs.hooks_[type] || [];
+  if (fn && type === 'vttjsloaded') {
+    vttjsOnLoad((vttjs) => {
```
BrandonOCasey: The onload function seems a bit weird in the vttjsloader module. Maybe we should just move all the logic for that here? I don't think that it would be used anywhere else.

src/js/tracks/text-track.js
```diff
@@ -88,21 +88,19 @@ const loadTrack = function(src, track) {
...
+    if (vttjsLoaded) {
       parseCues(responseBody, track);
+      return;
+    }
+    if (track.tech_) {
+      const loadHandler = () => parseCues(responseBody, track);
+
+      track.tech_.on('vttjsloaded', loadHandler);
+      track.tech_.on('vttjserror', () => {
+        log.error(`vttjs failed to load, stopping trying to process ${track.src}`);
```
gkatsev: Probably a good change, though, it's been around for a while now :P

src/js/tracks/vtt.js
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * The `VttLoader` is the `Component` that decides how to load in vttjs
+ *
+ * @extends Component
+ * @fires VttLoader#vttjsloaded
+ * @fires VttLoader#vttjserror
+ */
+class VttLoader extends Component {
+  constructor(player, options, ready) {
```
gkatsev: Having it be a component gives us a lot of things for free. For example, the ability to turn off the inclusion of vttjs when using the novtt build.
There's also precedent with the MediaLoader.

src/js/tracks/vtt.js
gkatsev: Makes sense to change. I just copied this from MediaLoader.

src/js/tracks/vtt.js
gkatsev: Yeah, we probably don't need it. I was thinking that maybe users can get the instance and listen to it but with the hooks feature it's redundant.

src/js/video.js
gkatsev: it can't be outside of vttjsloader because only that module knows if and when vttjs has loaded. Keeping it there, I think, it good because it contains all (except this little bit, but this mostly deals with hooks) the things that work with loading vttjs.

BrandonOCasey:
So fas this LGTM in my tests, we will have to release videojs-vtt.js and update the package.json here to get this to work.

BrandonOCasey:
So everything except for ie8 work, I don't know if we are doing work to support ie8 anymore anyway. This LGTM if thats the case. IE8 may have broken from another change, I can't even get videos to play.

gkatsev:
@BrandonOCasey @gesinger noticed that `grunt connect` isn't serving the swf up correctly anymore. If you can change the local sandbox to point to the cdn (or delete the override setting) it potentially should work, assuming that's the issue.

gkatsev:
This isn't going to go in any time soon, closing.

PR #4261, opened 2017-04-06
Add: Notes on Webpack Fixes in React Docs

MCDELTAT:
Note how to require the css file, as denoted in the FAQ sections and clarification on how to resolve the .eot files in the css file.

### DIFF
--- a/docs/guides/webpack.md
+++ b/docs/guides/webpack.md
@@ -1,12 +1,12 @@
 # Using Webpack with Video.js
 
-video.js, and the playback technologies such as videojs-contrib-hls all work in a Webpack based build environment. Here are several configuration changes specific to Webpack that will get you up and running. The following guide was written and tested for videojs v15.9.1 and videojs-contrib-hls v5.4.0
+video.js, and the playback technologies such as videojs-contrib-hls all work in a Webpack based build environment. Here are several configuration changes specific to Webpack that will get you up and running.
 
-### Video.js CSS:
+## Video.js CSS:
 To add the CSS that the player requires, simply add
 `require('!style-loader!css-loader!video.js/dist/video-js.css')` to the file where the player is also included or initialized.
 
-### Handling .eot files in Webpack
+## Handling .eot files in Webpack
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks @MCDELTAT, I wonder though if we should move it to a separate `webpack.md` file since the recommendations provided aren't react-specific. We'd also want to update the FAQ to point at that doc as well. Thoughts?

MCDELTAT:
Sure. I'll do that later today.

MCDELTAT:
In addition to the fixes already in place I also figured out how to use videojs-contrib-hls in te same setup. Should I create a webpack document there about the subject or combine all webpack related fixes in one doc?

gkatsev:
Keeping it here for contrib-hls is probably fine, though, we'd probably want to document it for HLS as well.

Also, is the change at the bottom of the FAQ just sorting of the links?

docs/guides/webpack.md
```diff
@@ -0,0 +1,37 @@
+# Using Webpack with Video.js
+
+video.js, and the playback technologies such as videojs-contrib-hls all work in a Webpack based build environment. Here are several configuration changes specific to Webpack that will get you up and running. The following guide was written and tested for videojs v15.9.1 and videojs-contrib-hls v5.4.0
```
gkatsev: I'm not sure it's necessary to specify the versions used. It should work with all versions

docs/guides/webpack.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Video.js CSS:
```
gkatsev: these headings should be `##`, but I can always change that before merging.

MCDELTAT:
Ok. Yes the bottom of the page is just sorting the links alphabetically. I guess my OCD kicked in. I took the time to verify all the links still worked as well. I'll change those to ## right now.

I think I left the versions in just because contrib-hls states that you need to add videojs-flash as well if you use v6.0. I can remove that if you want as well.

gkatsev:
Heh, it's fine. Just took me a while to figure out what was going on and wanted confirmation :D

PR #4277, opened 2017-04-12
fix: techOrder names can be camelCased.

gkatsev:
In the new middleware work, the way that new sources were loaded was refactored. We also recently made techs and components work either TitleCased or camelcased. There was one comparison that didn't do the proper check and cause the tech to be reloaded, even if the two techs were the same.

### DIFF
--- a/src/js/utils/to-title-case.js
+++ b/src/js/utils/to-title-case.js
@@ -22,6 +22,17 @@ function toTitleCase(string) {
 
 export default toTitleCase;
 
+/**
+ * Compares the TitleCase versions of the two strings for equality.
+ *
+ * @param {string} str1
+ *        The first string to compare
+ * @param {string} str2
+ *        The second string to compare
+ *
+ * @return {boolean}
+ *         Whether the TitleCase versions of the strings are equal
+ */
 export function titleCaseEquals(str1, str2) {
   return toTitleCase(str1) === toTitleCase(str2);
 }
--- a/src/js/utils/to-title-case.js
+++ b/src/js/utils/to-title-case.js
@@ -27,6 +27,7 @@ export default toTitleCase;
  *
  * @param {string} str1
  *        The first string to compare
+ *
  * @param {string} str2
  *        The second string to compare
  *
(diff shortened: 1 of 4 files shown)


src/js/utils/to-title-case.js
```diff
@@ -21,3 +21,7 @@ function toTitleCase(string) {
 }
 
 export default toTitleCase;
+
+export function titleCaseEquals(str1, str2) {
```
misteroneill: JSDoc?

BrandonOCasey:
should we update any other if checks to use the new titleCaseEquals function?

src/js/utils/to-title-case.js
gkatsev: Yes. Also, we need some tests, probably.

gkatsev:
@BrandonOCasey I think updates to current code should be done in another PR.

BrandonOCasey:
@gkatsev OK

src/js/utils/to-title-case.js
```diff
@@ -21,3 +21,18 @@ function toTitleCase(string) {
 }
 
 export default toTitleCase;
+
+/**
+ * Compares the TitleCase versions of the two strings for equality.
+ *
+ * @param {string} str1
+ *        The first string to compare
```
BrandonOCasey: I think we want a newline between this param and the other one for consistency

test/unit/utils/to-title-case.test.js
```diff
@@ -8,3 +8,10 @@ QUnit.test('should make a string start with an uppercase letter', function(asser
 
   assert.ok(foo === 'Bar');
 });
+
+QUnit.test('titleCaseEquals compares whether the TitleCase of two strings is equal', function(assert) {
+  assert.ok(titleCaseEquals('foo', 'foo'), 'foo equals foo');
+  assert.ok(titleCaseEquals('foo', 'Foo'), 'foo equals Foo');
+  assert.ok(titleCaseEquals('Foo', 'foo'), 'Foo equals foo');
+  assert.ok(titleCaseEquals('Foo', 'Foo'), 'Foo equals Foo');
```
BrandonOCasey: maybe some negative tests  with `notOK` too? `fOO` & `foo` etc

PR #4295, opened 2017-04-19
Persist caption/description choice over source changes

ldayananda:
## Description
This should allow a user's choice of subtitle or description to be respected across an entire session(if available).

## Specific Changes proposed
- Remember the user selection
- User selected choice is preferred over the first default track

## Requirements Checklist
- If necessary, more likely in a feature request than a bug fix
  - [x] Docs/guides updated (n/a)

### DIFF
--- a/src/js/control-bar/text-track-controls/text-track-button.js
+++ b/src/js/control-bar/text-track-controls/text-track-button.js
@@ -60,21 +60,41 @@ class TextTrackButton extends TrackButton {
     this.hideThreshold_ += 1;
 
     const tracks = this.player_.textTracks();
+    let preferredItem;
 
     for (let i = 0; i < tracks.length; i++) {
       const track = tracks[i];
+      const selectedLanguage = this.player_.cache_.selectedLanguage;
 
       // only add tracks that are of an appropriate kind and have a label
       if (this.kinds_.indexOf(track.kind) > -1) {
+
         const item = new TrackMenuItem(this.player_, {
           track,
           // MenuItem is selectable
           selectable: true
         });
 
+        // Find a track that matches at least the selectedLanguage's language
+        if (selectedLanguage) {
+          // Matches both language and kind
+          if (track.language === selectedLanguage.language &&
+            track.kind === selectedLanguage.kind) {
+            preferredItem = item;
+          // Matches just language
+          } else if (!preferredItem &&
+            track.language === selectedLanguage.language) {
+            preferredItem = item;
+          }
+        }
+
         item.addClass(`vjs-${track.kind}-menu-item`);
         items.push(item);
       }
+    }
+
+    if (preferredItem) {
+      preferredItem.track.mode = 'showing';
     }
 
     return items;
--- a/src/js/control-bar/text-track-controls/text-track-button.js
+++ b/src/js/control-bar/text-track-controls/text-track-button.js
@@ -60,11 +60,9 @@ class TextTrackButton extends TrackButton {
     this.hideThreshold_ += 1;
 
     const tracks = this.player_.textTracks();
-    let preferredItem;
 
     for (let i = 0; i < tracks.length; i++) {
       const track = tracks[i];
-      const selectedLanguage = this.player_.cache_.selectedLanguage;
 
       // only add tracks that are of an appropriate kind and have a label
       if (this.kinds_.indexOf(track.kind) > -1) {
@@ -75,26 +73,9 @@ class TextTrackButton extends TrackButton {
           selectable: true
         });
 
-        // Find a track that matches at least the selectedLanguage's language
-        if (selectedLanguage) {
-          // Matches both language and kind
-          if (track.language === selectedLanguage.language &&
-            track.kind === selectedLanguage.kind) {
-            preferredItem = item;
-          // Matches just language
-          } else if (!preferredItem &&
-            track.language === selectedLanguage.language) {
-            preferredItem = item;
-          }
-        }
-
         item.addClass(`vjs-${track.kind}-menu-item`);
         items.push(item);
       }
-    }
-
-    if (preferredItem) {
-      preferredItem.track.mode = 'showing';
     }
 
     return items;
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -115,22 +115,24 @@ class TextTrackDisplay extends Component {
       const trackList = this.player_.textTracks();
       let firstDesc;
       let firstCaptions;
+      let preferredTrack;
       let preferredDesc;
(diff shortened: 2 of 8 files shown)


test/unit/tracks/text-track-display.test.js
```diff
@@ -0,0 +1,279 @@
...
+    player.src({type: 'video/mp4', src: 'http://google.com'});
+    // manualCleanUp = true by default
+    player.addRemoteTextTrack(track1);
+    player.addRemoteTextTrack(track2);
+    // Force 'es' as user-selected track
+    player.cache_.selectedLanguage = 'es';
+    this.clock.tick(1);
+
+    // the spanish captions track should be shown
+    const englishTrack = getTrackByLanguage(player, 'en');
```
gkatsev: `player.addRemoteTextTrack` returns an `HTMLTrackElement` (or a shim) that has the track in a `.track` property. That might be another way of getting it rather than `getTrackByLanguage`, if that's the only places that `getTrackByLanguage` is used.

src/js/tracks/text-track-display.js
```diff
@@ -115,10 +115,24 @@ class TextTrackDisplay extends Component {
...
+          player.cache_.selectedLanguage === track.language) {
+          if (track.kind === 'descriptions' && !preferredDesc) {
+            preferredDesc = track;
+          } else if (track.kind in modes && !preferredCaptions) {
+            preferredCaptions = track;
+          }
+          break;
+        }
+
         if (track.default) {
```
gkatsev: should this be an `else if` for the above if?

test/unit/tracks/text-track-display.test.js
```diff
@@ -0,0 +1,279 @@
...
+    player.addRemoteTextTrack(track1);
+    player.addRemoteTextTrack(track2);
+
+    // Keep track of menu items
+    esCaptionMenuItem = getMenuItemByLanguage(captionsButton.items, 'es');
+    enCaptionMenuItem = getMenuItemByLanguage(captionsButton.items, 'en');
+
+    // The user chooses Spanish
+    player.play();
+    captionsButton.pressButton();
```
gkatsev: probably no need to press the captions button.

gkatsev:
Also, it's great that the non-code changes are pretty small and fairly localized!

test/unit/tracks/text-track-display.test.js
```diff
@@ -0,0 +1,279 @@
...
+  assert.strictEqual(fakeTTDSpy.callCount, 1, 'text track display was created');
+
+  Html5.TEST_VID = oldTestVid;
+  browser.IS_FIREFOX = oldIsFirefox;
+  Component.registerComponent('TextTrackDisplay', oldTextTrackDisplay);
+
+  player.dispose();
+});
+
+QUnit.test('no captions tracks, no captions are displayed', function(assert) {
```
ldayananda: You're right! Turns out there's a test like this in `text-track-controls.test.js` already which I will expand to check if `remoteTextTracks` is empty as well.

gkatsev: `remoteTextTracks` should never have stuff in it if `textTracks` is already empty.

test/unit/tracks/text-track-display.test.js
```diff
@@ -0,0 +1,279 @@
...
+      default: true
+    };
+    const track2 = {
+      kind: 'captions',
+      label: 'Spanish',
+      language: 'es',
+      src: 'es.vtt'
+    };
+
+    browser.IS_FIREFOX = true;
```
ldayananda: You're right, I don't actually need this if I check `supportsNativeTextTracks`

test/unit/tracks/text-track-display.test.js
ldayananda: oh nice, will just remove this one then.

src/js/tracks/text-track-display.js
ldayananda: Yes, will update

ldayananda:
Updated code and tests according to the feedback above. We decided that the reset method should continue to behave as it does currently and selectedLanguage will remain set if reset is called.

src/js/control-bar/text-track-controls/text-track-button.js
```diff
@@ -60,23 +60,43 @@ class TextTrackButton extends TrackButton {
...
+            preferredItem = item;
+          }
+        }
+
         item.addClass(`vjs-${track.kind}-menu-item`);
         items.push(item);
       }
     }
 
+    if (preferredItem) {
```
gkatsev: couldn't this have the side-effect of switching the selected track when a new track is added? Also, this seems like it'll try and enable a track per text-track-button.

ldayananda: Yes and yes. Back to the drawing board :)

ldayananda:
Updated once again, this time with more encapsulation

src/js/tracks/text-track-display.js
```diff
@@ -92,6 +92,7 @@ class TextTrackDisplay extends Component {
 
     player.on('loadstart', Fn.bind(this, this.toggleDisplay));
     player.on('texttrackchange', Fn.bind(this, this.updateDisplay));
+    player.on('loadstart', Fn.bind(this, this.preselectTrack));
```
gkatsev: did we decide on whether we wanted this to happen on `addtrack` as well?

src/js/tracks/text-track-list.js
```diff
@@ -61,6 +61,14 @@ class TextTrackList extends TrackList {
     track.addEventListener('modechange', Fn.bind(this, function() {
       this.trigger('change');
     }));
+
+    const nonLanguageTextTrackKind = ['metadata', 'chapters'];
+
+    if (nonLanguageTextTrackKind.indexOf(track.kind) === -1) {
+      track.addEventListener('modechange', Fn.bind(this, function() {
+        this.trigger('selectedlanguagechange');
```
gkatsev: just realized it's possible that the language doesn't change, so, the naming is a big weird, but maybe still ok.
For example, if you go from english captions to english subtitles, the language is the same but only the kind changed.

src/js/tracks/text-track-display.js
ldayananda: I believe that we discussed it and decided that it would be disorienting to preselect a new track if a track had already been selected by the user.

src/js/tracks/text-track-list.js
ldayananda: maybe `displayedtexttrackchange` would be a better name?

apadhye:
QA LGTM

Issue #4294, opened 2017-04-19
textTracks().on('change', handler) triggered for every caption


arski:
Hi, i dont know if this is intended behaviour but this seems wrong to me - the above event is triggered for every caption in the captions list when a caption is changed. E.g. I have 10 captions, i pick one, 10 events are triggered... This makes any kind of analytics tracking on my side very hard. Could you check if tjis is a bug or if there is a workaround? Thanks!

gkatsev:
This is "as designed", though, I guess more like "as implemented". It's a lot easier to just listen to each track's mode change and then re-trigger it.
Also, when we enable one track via the UI, we potentially disable all the other tracks, though, we probably don't need to do that and make sure that we only change the mode on the previously enabled track.

We probably should have it rollup multiple change events into one but it's probably tricky to do right.

arski:
> make sure that we only change the mode on the previously enabled track.

indeed that would make quite some sense if we sent 2 events, but not for those tracks that were already disabled..

oh well.. ill try to figure how to do this, but it sure is highly annoying to track and identify an actual change :(

gkatsev:
This has been addressed.

PR #4298, opened 2017-04-20
Only update track mode if changed

arski:
Hi there,

Would you consider this fix for triggering less track.modechange and subsequently trackList.change events - at the moment too many of those are triggered even if nothing has actually changed.

cc https://github.com/videojs/video.js/issues/4294

### DIFF
--- a/src/js/control-bar/text-track-controls/text-track-menu-item.js
+++ b/src/js/control-bar/text-track-controls/text-track-menu-item.js
@@ -101,7 +101,7 @@ class TextTrackMenuItem extends MenuItem {
       const track = tracks[i];
 
       if (track === this.track && (kinds.indexOf(track.kind) > -1)) {
-        if (track.mode !== 'showing) {
+        if (track.mode !== 'showing') {
           track.mode = 'showing';
         }
       } else if (track.mode !== 'disabled') {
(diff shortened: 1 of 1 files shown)


arski:
cc @gkatsev what do you think

src/js/control-bar/text-track-controls/text-track-menu-item.js
```diff
@@ -101,9 +101,13 @@ class TextTrackMenuItem extends MenuItem {
...
       if (track === this.track && (kinds.indexOf(track.kind) > -1)) {
-        track.mode = 'showing';
+        if (track.mode !== 'showing') {
+          track.mode = 'showing';
+        }
       } else {
-        track.mode = 'disabled';
+        if (track.mode !== 'disabled') {
+          track.mode = 'disabled';
+        }
```
misteroneill: The linter is complaining about this not being combined with the `else`, like so:

```js
} else if (track.mode !== 'disabled') {
```

gkatsev:
@arski thanks! Would you be able to make a similar PR against the 5.x branch? I can take care of that otherwise.

arski:
@gkatsev thanks! would you mind doing against the 5.x branch? Im not sure when ill get to it :( 

btw how frequently do you bump versions for the lib?

gkatsev:
Sure, I'll make a PR for it.

As for how often, it really just depends on whether there's things to release, we don't really have a strict release schedule right now, though, I want to try and aim for regular releases. 6.x with this fix has been released already.

arski:
Aha nice, good to know!

gkatsev:
Made a PR against 5.x in #4368.

PR #4316, opened 2017-04-29
Prevent dupe events on enabled ClickableComponents

### LINKED ISSUE #4312: ClickableComponent enable and disable method register events multiple times
## Description
i'm adding plugins to VideoJS version 5.19.2 and I created some buttons that can be disabled/enabled. When we call the enable() method and the button is already enabled, it will register the click event a second time.
 
~~~~
class TestButton extends Button {
        super(player, options);
        this.on(player, 'play', this.updateShowing);
    }
    buildCSSClass() {
        return `vjs-test-control ${super.buildCSSClass()}`;
    }
    handleClick(event) {
        //this will be called 1, 2, 4, 8, 16, 32, ... times
        this.player_.play();
    }
    updateShowing(event) {
        if (true) {
            this.enable();
        } else {
            this.disable();
        }
    }
}
~~~~
## Steps to reproduce
1. Create a button and call enable() multiple times.
2. Click on that button.

## Results
handleClick() will be called more than once.
### Expected
handleClick() is called once. The enable() call should have no effect if it's already enabled. Same for disable().

mister-ben:
## Description
Prevent ClickableComponents re-adding event listeners each time `enabled()` is called
Fixes #4312

## Specific Changes proposed
- Keeps track of enabled state (`this.enabled_`)
- `enable()` doesn't do anything if the component is enabled, so the event handlers are not re-added

## Requirements Checklist

### DIFF
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -141,28 +141,30 @@ class ClickableComponent extends Component {
    * Enable this `Component`s element.
    */
   enable() {
-    this.removeClass('vjs-disabled');
-    this.el_.setAttribute('aria-disabled', 'false');
-    if (typeof this.tabIndex_ !== 'undefined') {
-      this.el_.setAttribute('tabIndex', this.tabIndex_);
+    if (!this.enabled_) {
+      this.enabled_ = true;
+      this.removeClass('vjs-disabled');
+      this.el_.setAttribute('aria-disabled', 'false');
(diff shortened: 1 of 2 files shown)


gkatsev:
I guess we should fix this in 5.x as well.

mister-ben:
It's causing some tests in MenuButton and TextTracksControls to fail on 5.x, will need to dig into why.

PR #4318, opened 2017-05-02
Ensure existing text tracks are cleared when switching to a new source

tmm1:
## Description

Previously only audio/video tracks were cleared. This is a simple bug/oversight.

## Specific Changes proposed

Using the same pattern to clear tracks as elsewhere.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
thanks for the PR. However, text tracks were specifically not cleared here because that's how the browser works. When you swap videos in the browser, text tracks stick around between videos unless you clear them out. We do need to update our "manual cleanup" option of `addRemoteTextTrack`.

PR #3246, opened 2016-04-08
Blacklist Android Chrome for playback rate support

mister-ben:
## Description
Backlists Android Chrome when testing for playback rate change support, as this browser implements the playbackRate API but it is ineffective.
Fixes #3180

## Specific Changes proposed
Blacklist based on UA in `Html5.canControlPlaybackRate()`.
Should be revisited when Chrome implements playback rate control.

Plus some additional doc corrections.

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] Reviewed by Two Core Contributors

PR #4321, opened 2017-05-03
Remove playbackRate blacklist for recent Android Chrome

mister-ben:
## Description
Android Chrome now supports playbackRate properly, so removes the blacklist added in #3246 for newer Chrome versions.  

## Specific Changes proposed
Adds `browser.CHROME_VERSION` as a necessary evil.
No longer blacklists for Chrome 58+ -- this could possibly be fixed since 52, but 58 is all I've been able to test on and most users should keep Chrome up to date.

## Requirements Checklist

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -62,6 +62,14 @@ export const IS_NATIVE_ANDROID = IS_ANDROID && ANDROID_VERSION < 5 && appleWebki
 export const IS_FIREFOX = (/Firefox/i).test(USER_AGENT);
 export const IS_EDGE = (/Edge/i).test(USER_AGENT);
 export const IS_CHROME = !IS_EDGE && (/Chrome/i).test(USER_AGENT);
+export const CHROME_VERSION = (function() {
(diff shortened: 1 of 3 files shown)


src/js/utils/browser.js
```diff
@@ -62,6 +62,14 @@ export const IS_NATIVE_ANDROID = IS_ANDROID && ANDROID_VERSION < 5 && appleWebki
 export const IS_FIREFOX = (/Firefox/i).test(USER_AGENT);
 export const IS_EDGE = (/Edge/i).test(USER_AGENT);
 export const IS_CHROME = !IS_EDGE && (/Chrome/i).test(USER_AGENT);
+export const CHROME_VERSION = (function() {
+  const match = USER_AGENT.match(/Chrome\/(\d+)/);
+
+  if (match && match[1]) {
+    return match[1];
```
gkatsev: we should return `parseFloat(match[1])` here so that we can actually do number comparison. The tests would need to be updated in that case.

PR #4325, opened 2017-05-04
Allow 'playsinline' player option

alex-barstow:
### Description
Video.js players can accept a number of standard `<video>` element options (`autoplay`, `muted`, `loop`, etc), but not currently `playsinline`, which is now part of the [HTML spec](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-video-playsinline). We should add it to the list of `<video>` attributes that can be provided to the player as options.

### Specific Changes proposed
Update the `Html5` and `Player` classes with getter and setter methods for `playsinline` and make a few other additions to ensure that the `playsinline` option works the same way as the other `<video>` element options.

### Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -822,6 +822,7 @@ class Player extends Component {
       'textTracks': this.textTracks_,
       'audioTracks': this.audioTracks_,
       'autoplay': this.options_.autoplay,
+      'playsinline': this.options_.playsinline,
       'preload': this.options_.preload,
       'loop': this.options_.loop,
       'muted': this.options_.muted,
@@ -2397,7 +2398,7 @@ class Player extends Component {
       this.options_.playsinline = value;
       return this;
     }
-    return this.techGet_('playsinline', value);
+    return this.techGet_('playsinline');
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2394,6 +2394,8 @@ class Player extends Component {
    * @return {string|Player}
    *         - the current value of playsinline
    *         - the player when setting
+   *
+   * @see [Spec]{@link https://html.spec.whatwg.org/#attr-video-playsinline}
    */
   playsinline(value) {
     if (value !== undefined) {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2381,6 +2381,26 @@ class Player extends Component {
   }
 
   /**
+   * Get or set the playsinline attribute.
+   *
+   * @param {boolean} [value]
+   *        - true means that we should play inline in iOS Safari
+   *        - false means that we should not play inline in iOS Safari
+   *
+   * @return {string|Player}
+   *         - the current value of playsinline
+   *         - the player when setting
+   */
+  playsinline(value) {
+    if (value !== undefined) {
+      this.techCall_('setPlaysinline', value);
+      this.options_.playsinline = value;
+      return this;
+    }
+    return this.techGet_('playsinline', value);
+  }
+
+  /**
    * Get or set the loop attribute on the video element.
    *
    * @param {boolean} [value]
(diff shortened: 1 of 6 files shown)


alex-barstow:
I assume this will need a sister PR in `master` as well

src/js/player.js
```diff
@@ -2381,6 +2382,26 @@ class Player extends Component {
   }
 
   /**
+   * Get or set the playsinline attribute.
+   *
+   * @param {boolean} [value]
+   *        - true means that we should play inline in iOS Safari
```
gkatsev: I would change these two say `set the playsinline attribute` or `unset it`. And mention above that this probably only applies on ios right now.

src/js/tech/html5.js
```diff
@@ -1235,6 +1235,20 @@ Html5.resetMediaElement = function(el) {
   'autoplay',
 
   /**
+   * Get the value of `playsinline` from the media element. `playsinline` indicates
+   * that the media should play inline in iOS Safari.
```
gkatsev: iOS is an example of a browser that will only play inline if the attribute is present.

src/js/tech/html5.js
```diff
@@ -1518,6 +1532,19 @@ Html5.resetMediaElement = function(el) {
   'autoplay',
 
   /**
+   * Set the value of `playsinline` on the media element. `playsinline` indicates
```
gkatsev: Same here.

gkatsev:
Also, yes, a PR against master would be good. Can be done once this PR is approved, though.

gkatsev:
@alex-barstow While testing, I found that if I have a video embed with `playsinline` attribute, calling `player.playsinline()` after the player loads doesn't reflect that value.

Also, we should add an `@abstract` method in `tech.js` for this.

src/js/tech/html5.js
```diff
@@ -1431,6 +1431,27 @@ Html5.resetMediaElement = function(el) {
...
+ *
+ * @method Html5#playsinline
+ * @return {boolean}
+ *         - The value of `playsinline` from the media element.
+ *         - True indicates that the media should play inline.
+ *         - False indicates that the media should not play inline.
+ *
+ * @see [Spec]{@link https://html.spec.whatwg.org/#attr-video-playsinline}
+ */
+Html5.prototype.playsinline = function() {
```
gkatsev: you should move these up and use the class syntax

src/js/tech/html5.js
```diff
@@ -1553,6 +1574,30 @@ Html5.resetMediaElement = function(el) {
   };
 });
 
+// This setter is declared outside of the loop because there is no 'playsinline'
```
gkatsev: We can just describe this in the PR instead, don't think it's necessary, especially if we move it into the class syntax.

src/js/player.js
```diff
@@ -2381,6 +2382,29 @@ class Player extends Component {
   }
 
   /**
+   * Set or unset the playsinline attribute.
+   * Playsinline tells the browser that non-fullscreen playback is preferred.
+   *
```
gkatsev: Can we add a `@see` here as well?

imbcmdth:
@gkatsev @alex-barstow  Should we have any testing like https://github.com/videojs/video.js/blob/master/test/unit/player.test.js#L94 where we make sure `autoplay` is propagated from the video element?

gkatsev:
@imbcmdth Perhaps adding it would be useful, but the tests are already testing that if attributes are present they persist when videojs wraps the video element.
I guess we might want to test the setter as well.

gkatsev:
@alex-barstow would you be able to make the sister PR against master for this as well?

Issue #2141, opened 2015-05-12
Is there any way to jump to the next chapter in the video.


chaitanyapro:
So I've defined all the chapters and everything is working fine is there any api method to jump to the next chapter in the video.

space87:
Hey chaitanyapro,

you can use the built in chapters menu that appears as a icon on the control bar, or just take the start time of each chapter and store them and use the currentTime() method to set to the time of the chapter when a button is clicked

OwenEdwards:
Seems like a great opportunity for a simple plug-in feature!

tmm1:
I've written a simple plugin for this at https://github.com/fancybits/videojs-chapter-nav, but it requires #4326 to function reliably.

PR #4326, opened 2017-05-04
Trigger `change` event on ChaptersButton after update

tmm1:
I am writing a plugin to add prev/next chapter buttons, and need this event to be able to detect when the chapters menu changes.

cc #2141


gkatsev:
As it is, right now, it's not really that much different than listening to the TextTrackList events yourself (`addtrack`, `removetrack` and `change`) and then calling `chaptersButton.findChaptersTrack()` to find the currently selected track for chapters.
As it is now, I think that's probably the best bet actually.

tmm1:
The problem is that those events are not enough when dealing with remote tracks. ChaptersButton also schedules an `update()` from `setTrack()` once the remote track is loaded, so firing `change` from update and reacting to that simplifies things greatly.

PR #4336, opened 2017-05-08
Remove playbackRate blacklist for recent Android Chrome (v5)

mister-ben:
## Description
#4321 for v5

Android Chrome now supports playbackRate properly, so removes the blacklist added in #3246 for newer Chrome versions.  

## Specific Changes proposed
Adds `browser.CHROME_VERSION` as a necessary evil.
No longer blacklists for Chrome 58+ -- this could possibly be fixed since 52, but 58 is all I've been able to test on and most users should keep Chrome up to date.

## Requirements Checklist


gkatsev:
Interesting, just tried it out, and seems like tapping on the playback rate button, it just cycles through the items. Not something we should change here but was interesting to find.

PR #4347, opened 2017-05-10
test: fix playsinline test for IE8

gkatsev:
This test is more of an integration test and uses the html5 tech to make
sure the setter and getter work properly. This causes issues with IE8
which doesn't work well with the Html5 tech.

Instead, we wrap the test in an if that checks whether Html5 is
supported.


gkatsev:
cc @alex-barstow.

PR #4354, opened 2017-05-12
fix: only disable user-selection on sliders

gkatsev:


src/css/components/_modal-dialog.scss
```diff
@@ -18,6 +18,8 @@
 .vjs-modal-dialog .vjs-modal-dialog-content {
   @extend %fill-parent;
 
+  @include user-select(auto);
```
dmlap: Why wasn't this user selectable to being with?

gkatsev: We're [disabling user-selection on the player div](https://github.com/videojs/video.js/blob/master/src/css/components/_layout.scss#L24) so that it doesn't get highlighted if you're dragging the progress bar. For whatever reason, even though, it's not supposed to be inherited, it applies on this element.

dmlap: [mdn says](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#Syntax) it "applies to subelements", though it's not inherited. Is this something we need to worry about more broadly?

dmlap: A thought: what about if we just disabled user selection on the progress bar if that's what we're trying to guard against?

PR #4357, opened 2017-05-13
Prevent dupe events on enabled ClickableComponents (v5)

mister-ben:
## Description
Like #4316 for v6, this prevents multiple click handlers being added if an already enabled component is enabled. In the PR for six, `ClickableComponent` was tracking `this.enabled_`  but this is incompatible with v5 on a `MenuButton` which also changes the state of `this.enabled_`. 

## Specific Changes proposed
Removes event handlers in `enable()` before adding them. 

Do we want the same fix implemented in a different way in 5 and 6 though? 

## Requirements Checklist

### DIFF
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -147,15 +147,17 @@ class ClickableComponent extends Component {
    *         Returns itself; method can be chained.
    */
   enable() {
-    this.removeClass('vjs-disabled');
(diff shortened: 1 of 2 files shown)


gkatsev:
I don't think it's a huge deal if these are implemented different across the two releases lines. Though, consistency is nice.

PR #4363, opened 2017-05-16
chore: move imports up so bundlers wont trip up

gkatsev:
The order of imports in video.js is causing some issues with some bundlers from producing a working package. While we should try and produce a reduced test case for the bundlers to fix this issue, we should also just work around it since it is fairly easy to do so.

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -7,6 +7,8 @@
 
 import window from 'global/window';
 import document from 'global/document';
+import * as browser from './utils/browser.js';
+import * as Dom from './utils/dom.js';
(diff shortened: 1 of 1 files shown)


src/js/video.js
```diff
@@ -7,6 +7,8 @@
 
 import window from 'global/window';
 import document from 'global/document';
+import * as browser from './utils/browser.js';
```
gkatsev: this needs to be the first local import, otherwise, rollup puts Dom below browser in the output file and it breaks at run time.

PR #4368, opened 2017-05-18
fix: Only update text track mode if changed

gkatsev:
This is the 5.x version of https://github.com/videojs/video.js/pull/4298


arski:
Lgtm thx

PR #4383, opened 2017-05-30
fix: Update a translation key to match correct string

mister-ben:
## Description
The lang files were translating "all settings to the default values" instead of "[restore all settings to the default values](https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-settings.js#L461)".

## Specific Changes proposed
- Updated the key in all lang files already using it
- Updated the French translation and added some further missing German translations

@h0lysl4y3r the Slovakian translation probably needs an update now.
 
## Requirements Checklist

### DIFF
--- a/docs/translations-needed.md
+++ b/docs/translations-needed.md
@@ -60,7 +60,7 @@ This default value is hardcoded as a default to the localize method in the SeekB
 |                         | Script                                                                              |
 |                         | Small Caps                                                                          |
 |                         | Reset                                                                               |
-|                         | all settings to the default values                                                  |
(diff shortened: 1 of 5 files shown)


gkatsev:
I wonder if the string should just be `all settings to the default values`. I think it was supposed to be so that the screen reader would read: `Reset all settings to the default values` but if we add `restore` into the translations, it'll be doubled up and it'll read `Reset restore all settings to the default values`.

mister-ben:
Ah, but that wouldn't work for some languages because of the word order.

Issue #4378, opened 2017-05-25
IE10: Unexpected call to method or property access


jetlogs:
## Description
JS error on IE10 with HTML5 tech causing entire JS execution to break

## Steps to reproduce
1. ~~multiple videoJS instances on same page~~ reproducible even with a single player instance
2. when one video instance ends, JS error gets triggered

## Results
```
SCRIPT65535: Unexpected call to method or property access.
audio-track-list.js (44,1)
```
![](http://i.imgur.com/tS93zlA.jpg)

### Expected
```js
var disableOthers = function disableOthers(list, track) {
  for (var i = 0; i < list.length; i++) {
    if (track.id === list[i].id) {
      continue;
    }
    // another audio track is enabled, disable it
    list[i].enabled = false;
  }
};
```
in this function, list[i] would return an empty Object {} without an id property causing said issue
causes the rest of the JS execution to break when this triggers

any way to disable the disableOthers() function or fix this issue? I'm not using any audioTracks for the videos.

Thanks!

## Additional Information
IE10
videojs: 5.16.2

jetlogs:
update:
it looks like this happens even with only one instance of the player on IE10

jetlogs:
created a PR to fix this issue on v6.2
https://github.com/videojs/video.js/pull/4395

PR #4395, opened 2017-06-05
fix IE10 issue for disableOthers

jetlogs:
## Description
in IE10, the variable list for disableOthers returns "permission denied"
this causes an uncaught exception which breaks the rest of the javascript execution

issue described in #4378 

## Specific Changes proposed
directly invoking list[i].id triggers the permission denied error
however, if an object has all of its properties denied access to, Object.keys returns a length of 0

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/audio-track-list.js
+++ b/src/js/tracks/audio-track-list.js
@@ -19,7 +19,7 @@ import document from 'global/document';
  */
 const disableOthers = function(list, track) {
   for (let i = 0; i < list.length; i++) {
-    if (track.id === list[i].id) {
+    if (!Object.keys(list[i]).length || track.id === list[i].id) {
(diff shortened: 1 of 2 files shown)


gkatsev:
Hey @jetlogs. Thanks for the PR. This fix doesn't seem bad but I would like to have more background on why and when it's occurring? Would you be able to make a reduced test case? What type of source are you playing? In most of my testing, I can't get this to fail.

jetlogs:
Hi,

it looks like this issue is happening when IE10 restricts access to an object's properties so acessing them directly causes an exception.
If I have to take a guess, this might have been caused by loading a media file that's cross-domain (like a CDN for in our use case)

Issue #4424, opened 2017-06-20
Reset volume icon to reflect that muted(true) and vol = 0 are not the same


marguinbc:
## Description
Currently, the same icon is used for muted(true) and volume = 0. However, it is possible for volume to be 0 without the player being muted, for example by sliding the volume slider to 0.  This changes the icon to the muted icon. Toggling the muted icon then does nothing because the player is not muted, it just has a volume level of 0.  The only way to remove the muted icon is to move the volume slider to set a new volume level.

If we want to fix this, we would need to keep track of whether we muted the player by toggling the mute-toggle, via the muted attribute, or by moving the volume to 0 using the slider.

An easier solution would be to redefine the ranges for the volume icons so that:

- muted(true):  vjs-icon-volume-muted = vjs-vol-0
- vol 0:  vjs-icon-volume-low = vjs-vol-1 
- vol >0 && <= .5: vjs-icon-volume-mid = vjs-vol-2
- vol > .5: vjs-icon-volume-high = vjs-vol-3

Visually this also makes sense because:

- vjs-icon-volume-muted has a slash through it representing muted
- vjs-icon-volume-low has no slash and the speaker icon with no arcs/sound waves
- vjs-icon-volume-mid has one arc/sound ring which can visually be interpreted as half
- vjs-icon-volume-high has two arcs/sound rings which can visually be interpreted as half

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Mute the player. Note that player.muted() returns true, player.volume() should return whatever the volume is/was prior to muting - which initially should be 1
2.Un-mute the player tapping the mute-toggle button. Note that player.muted() returns false, player.volume() should return the same as before
3. Click and drag the volume slider to 0. Note that player.muted() is false and player.volume() returns 0 but the icon is the same as muted() true
4. Try and restore the volume by tapping the mute-toggle button. Nothing happens. The volume remains 0 and the icon remains crossed out.
5. Click and drag anywhere on the volume slider and note that the volume will change, removing the muted icon

### Expected
Sliding the volume slider to 0 sets the volume to 0 without setting muted(true) or showing the muted icon. When the muted-toggle button appears muted, tapping it should un-muted and restore the original volume level.

### Actual
Sliding the volume slider to 0 sets the volume to 0 sets the mute-toggle button icon to muted but does not mute the player. Tapping the mute-toggle button to try and un-mute does not work. Only sliding the volume slider removes the muted icon state and restores the volume.

### Error output
[...]

PR #4425, opened 2017-06-20
adjust volume ranges so muted(true) and vol=0 do not use the same icons

marguinbc:
## Description
Potential fix for https://github.com/videojs/video.js/issues/4424

## Specific Changes proposed
Change the volume levels used to select the videojs-icon-volume-* classes so muted(true) and volume(0) do not use the same icon.  Fixes an issue where sliding the volume slider to 0 has the appearance of setting muted(true) but toggling the mute-toggle/volumeMenuButton does not restore the volume.  Only setting the volume using the slider removes the muted icon appearance.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/mute-toggle.js
+++ b/src/js/control-bar/mute-toggle.js
@@ -83,9 +83,9 @@ class MuteToggle extends Button {
 
     if (this.player_.muted()) {
       level = 0;
-    } else if (vol === 0) {
+    } else if (vol < 0.33) {
       level = 1;
-    } else if (vol > 0 && vol <= 0.5) {
+    } else if (vol < 0.67) {
       level = 2;
     }
 
--- a/src/js/control-bar/mute-toggle.js
+++ b/src/js/control-bar/mute-toggle.js
@@ -81,11 +81,11 @@ class MuteToggle extends Button {
     const vol = this.player_.volume();
     let level = 3;
 
-    if (vol === 0 || this.player_.muted()) {
+    if (this.player_.muted()) {
       level = 0;
-    } else if (vol < 0.33) {
+    } else if (vol === 0) {
       level = 1;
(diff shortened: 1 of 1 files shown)


src/js/control-bar/mute-toggle.js
```diff
@@ -81,11 +81,11 @@ class MuteToggle extends Button {
     const vol = this.player_.volume();
     let level = 3;
 
-    if (vol === 0 || this.player_.muted()) {
+    if (this.player_.muted()) {
```
gkatsev: what about only making this change?
I think level 1 is supposed to imply that the volume is very low and not just 0. So, if muted, level 0, if <0.33, level 1, if less than .67 level 2, else level 3?

marguinbc: That works for me. Only reason I changed the other levels is because I thought no ring seems like a good indicator for 0 sound, 1 ring for up to 50% and two rings for over 50%

mister-ben:
I'm in two minds about this. It does make sense to have some UI change when muting a player with zero volume so it doesn't appear that nothing has happened, but it might cause confusion if a user thinks the slider isn't getting all the way to zero when they drag it to the left. It does illustrate that the v6 behaviour is a big improvement in UX.

gkatsev:
The fact that volume 0 and muted are separate is kind of weird from a UI perspective. Kind of makes sense from an implementation perspective, though. @marguinbc made a PR to make the swf behave like the video element does wrt muted and volume (https://github.com/videojs/video-js-swf/pull/230).
@mister-ben do you think we should not have this change? Unfortunately, I don't think we can really make any other changes to improve this in v5 because it may be a breaking change.

mister-ben:
I'd go with this on balance. I've been following the work on the swf, great improvement.

PR #4426, opened 2017-06-20
perf: Various small performance improvements.

misteroneill:
These are a few performance improvements.

- Use `textContent` instead of `innerHTML` for text content.
- Construct `TextTrackSettings` content with HTML strings where possible because it's a bit faster than manual DOM construction.
- Do minimal DOM updates to time controls. This reduces our `timeupdate` footprint from ~2-2.5ms to ~1ms!

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/time-controls/current-time-display.js
+++ b/src/js/control-bar/time-controls/current-time-display.js
@@ -25,8 +25,6 @@ class CurrentTimeDisplay extends Component {
    */
   constructor(player, options) {
     super(player, options);
-
-    this.formattedTime_ = '0:00';
     this.throttledUpdateContent = throttle(bind(this, this.updateContent), 25);
     this.on(player, 'timeupdate', this.throttledUpdateContent);
   }
@@ -67,7 +65,7 @@ class CurrentTimeDisplay extends Component {
     if (this.textNode_) {
       this.contentEl_.removeChild(this.textNode_);
     }
-    this.textNode_ = document.createTextNode(` ${this.formattedTime_}`);
+    this.textNode_ = document.createTextNode(` ${this.formattedTime_ || '0:00'}`);
     this.contentEl_.appendChild(this.textNode_);
   }
 
--- a/src/js/control-bar/time-controls/current-time-display.js
+++ b/src/js/control-bar/time-controls/current-time-display.js
@@ -1,8 +1,10 @@
 /**
  * @file current-time-display.js
  */
+import document from 'global/document';
 import Component from '../../component.js';
 import * as Dom from '../../utils/dom.js';
+import {bind, throttle} from '../../utils/fn.js';
 import formatTime from '../../utils/format-time.js';
 
 /**
@@ -24,7 +26,9 @@ class CurrentTimeDisplay extends Component {
   constructor(player, options) {
     super(player, options);
 
-    this.on(player, 'timeupdate', this.updateContent);
+    this.formattedTime_ = '0:00';
+    this.throttledUpdateContent = throttle(bind(this, this.updateContent), 25);
+    this.on(player, 'timeupdate', this.throttledUpdateContent);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/js/control-bar/time-controls/current-time-display.js
```diff
@@ -62,12 +82,11 @@ class CurrentTimeDisplay extends Component {
   updateContent(event) {
     // Allows for smooth scrubbing, when player can't keep up.
     const time = (this.player_.scrubbing()) ? this.player_.getCache().currentTime : this.player_.currentTime();
-    const localizedText = this.localize('Current Time');
     const formattedTime = formatTime(time, this.player_.duration());
 
     if (formattedTime !== this.formattedTime_) {
       this.formattedTime_ = formattedTime;
-      this.contentEl_.innerHTML = `<span class="vjs-control-text">${localizedText}</span> ${formattedTime}`;
+      this.updateTextNode_();
```
gkatsev: we should probably wrap this in a rAF as well.

src/js/control-bar/time-controls/duration-display.js
```diff
@@ -24,13 +26,18 @@ class DurationDisplay extends Component {
...
-    this.on(player, 'timeupdate', this.updateContent);
-    this.on(player, 'loadedmetadata', this.updateContent);
+    this.on(player, [
+      'durationchange',
+
+      // Also listen for timeupdate and loadedmetadata because removing those
+      // listeners could have broken dependent applications/libraries. These
+      // can likely be removed for 7.0.
+      'loadedmetadata',
+      'timeupdate'
```
gkatsev: Noooooo. Did we forget to remove `timeupdate` in 6.x? T__T

src/js/tech/html5.js
```diff
@@ -368,24 +368,26 @@ class Html5 extends Tech {
     // playback has started, which triggers the live display erroneously.
     // Return NaN if playback has not started and trigger a durationupdate once
     // the duration can be reliably known.
-    if (this.el_.duration === Infinity &&
```
gkatsev: is this section mostly formatting?

src/js/control-bar/time-controls/current-time-display.js
```diff
@@ -24,7 +26,9 @@ class CurrentTimeDisplay extends Component {
   constructor(player, options) {
     super(player, options);
 
-    this.on(player, 'timeupdate', this.updateContent);
+    this.formattedTime_ = '0:00';
+    this.throttledUpdateContent = throttle(bind(this, this.updateContent), 100);
```
gkatsev: how come here it's 100 and below it's 25?

src/js/control-bar/time-controls/current-time-display.js
gkatsev: And below as well.

src/js/control-bar/time-controls/duration-display.js
misteroneill: Yeah. 😢

src/js/tech/html5.js
misteroneill: Yeah. It's going from:

```js
if (condition) {
  if (condition) {
    // Do stuff.
  }
}
```
to:
```js
if (condition && condition) {
  // Do stuff.
}
```

src/js/control-bar/time-controls/current-time-display.js
misteroneill: Whoops. :)

src/js/tech/html5.js
misteroneill: Not strictly perf-related, but I noticed it in searching for `timeupdate` listeners.

src/js/control-bar/time-controls/current-time-display.js
```diff
@@ -39,19 +43,35 @@ class CurrentTimeDisplay extends Component {
...
+   * Updates the "current time" text node with new content using the
+   * contents of the `formattedTime_` property.
+   *
+   * @private
+   */
+  updateTextNode_() {
+    if (this.textNode_) {
+      this.contentEl_.removeChild(this.textNode_);
+    }
+    this.textNode_ = document.createTextNode(` ${this.formattedTime_}`);
```
gkatsev: `updateTextNode_` gets called from `createEl`, which in turn, gets called during component init which happens before we set `this.formattedTime_`. We should either have a default value for when `this.formattedTime_` isn't set, or we need to call `updateTextNode_` in the end of the constructor.
This applies to the currentTimeDisplay and duration display and whatever else uses this `updateTextNode_`.
undefined gif

misteroneill: I think the former is preferable. I'll update. Can't believe I somehow didn't notice this...

Issue #4315, opened 2017-04-29
 Setting TextTrack to defaults not working when text track and video player are created dynamically.


Pinank08:
## Description
I wan't to create a dynamic video player i.e. video src and caption src comes from different file on click event.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. I have created a dynamic video player in videojs.
2. Than to add text track I have used addRemoteTextTrack() function.
3. But I am not able to make the track added as default as it always selects subtitle off as default.

Here is the code that I have used. 
```js
var oldTracks = player.remoteTextTracks();
  var i = oldTracks.length;
  while (i--) {
    player.removeRemoteTextTrack(oldTracks[i]);
  }

var trackk = new function() {
    this.kind = 'subtitles';
    this.label = 'english';
    this.language = 'en';
    this.id='track1';
    this.src = url+'.vtt';
    this.default= 'true'; // i have tried default too
}
    player.addRemoteTextTrack(trackk, true);

/*  I have used 
player.TextTrack = 'default';
player.TextTracks.mode='showing';
and many more... but subtitles are always off bydefault
*/ 
```

### Expected

I Expect to see subtitles when video is loaded.   

### Actual

When Video is loaded subtitles are off.

jthomerson:
@Pinank08 please see my description in #4403. Are you using version 6.x? If so, please let us know because your problem is likely caused by #4403. Can you try the workarounds I show in that example to see if they fix your issue?

gkatsev:
Closing as potentially fixed and due to lack of activity. If we get more information, we can re-open.

partap:
I am having this problem with version 6.6.2.

Actually, I have the video and a default track loaded initially with HTML tags, and the track is selected by default...this works fine.  The page implements a very rudimentary subtitle editor...for now, any time a cue is edited, it deletes the entire TextTrack and reloads it using `Player.removeTrack()` and `Player.addRemoteTextTrack()`.  This is also working, except when the new track is loaded with `default: true` option set, it remains unselected in the player until I click it in the menu.  

The documentation is not clear as to how to activate a loaded TextTrack from Javascript, other than setting `default: true`

partap:
I just spent a while stepping through with devtools... it looks like the property that is affected by selecting the text track from the menu is `mode`, which toggles between `'disabled'` and `'showing'` 

[...]

PR #4434, opened 2017-06-22
fix: auto-removal remote text tracks being removed when not supposed to

### LINKED ISSUE #4403: Remote Text Tracks Deleted Immediately Because of Asynchronous Nature of `src`
## Description

If I call `.src({ src: '...' })` and then call `.addRemoteTextTrack({ src: '...' }, false)` right after it, my remote text tracks are deleted immediately, and do not appear on the video player. Why? Because the call to `.src(...)` is actually asynchronous, and passing the `false` means "auto cleanup remote text tracks the next time the source changes". See the example below for a more detailed flow description.

This did not happen in video.jx 5.x because calls to `.src(...)` were synchronous. You can verify that by replacing the CSS/JS includes below with:

 * http://vjs.zencdn.net/5.19.2/video-js.css
 * http://vjs.zencdn.net/5.19.2/video.js

I believe the source of the problem to be this line of code: https://github.com/videojs/video.js/blob/c31836c30cebe6e117edc341588d39a464570867/src/js/tech/middleware.js#L19

One possible fix would be to make the call to `addRemoteTextTrack` asynchronous as well. But that feels a bit like a hack.

## Steps to reproduce
Use this example page, linking the CSS and JS to a 6x version of video.js. (*side point: I can't find 6.x on a CDN, which is what stopped me from making this a jsfiddle*)

```html
<head>
   <link href="/dist/video-js.css" rel="stylesheet">
</head>

<body>
   <video id="my-video" class="video-js"></video>

   <script src="/dist/video.js"></script>
   <script
      src="https://code.jquery.com/jquery-3.2.1.min.js"
      integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4="
      crossorigin="anonymous"></script>
   <script>
      $(document).ready(function() {
            fluid: true,
            controls: true,
         });

         player.ready(function() {
            // the call to src is asynchronous under the covers
            player.src({ src: 'https://download-a.akamaihd.net/files/media_video/00/wsb_E_r720P.mp4' });
            // this says "delete this text track whenever the source next changes"
            // (in other words manual cleanup is off, auto cleanup is on)
            // but because the call to src(...) was asynchronous, and addRemoteTextTrack is synchronous, the order of
            // operations is:
            //    1. I call src - it sets a timeout for 1ms to set the actual source
            //    2. I call addRemoteTextTrack - it immediately adds the text track
            //    3. I'm done, and now (or 1ms from now), the source gets actually set
            //    4. as part of setting the source, my remote track gets cleaned up - deleted - gone - blown away
            // Nobody sees subtitles and everyone is sad.
            player.addRemoteTextTrack({ src: 'https://download-a.akamaihd.net/files/media_video/9a/wsb_E_r720P.vtt' }, false);

            /*
            // This would work, though:
            // Under the covers, ready must queue functions to happen in a setTimeout - haven't looked to be sure
            player.ready(function() {
               player.addRemoteTextTrack({ src: 'https://download-a.akamaihd.net/files/media_video/9a/wsb_E_r720P.vtt' }, false);
            });
            */

            /*
            // And this works, because our setTimeout function will get invoked after the previously-scheduled one:
            setTimeout(function() {
               player.addRemoteTextTrack({ src: 'https://download-a.akamaihd.net/files/media_video/9a/wsb_E_r720P.vtt' }, false);
            }, 1);
            */
         });
      });
   </script>
</body>
```

### Expected
Please describe what you expected to see.

Subtitles.

### Actual
Please describe what actually happened.

No subtitles.

### Error output

Not applicable.

## Additional Information
#### videojs

6.x

#### browsers
what browser are affected?

all

#### OSes

all

### plugins

none

gkatsev:
## Description
We added a feature so that remote text tracks can auto-removed when a source changes. However, in 6.x we changed the source behavior to be asynchronous meaning that some text tracks were accidentally being removed when they weren't supposed to be.
For example:
```js
var player = videojs('my-player');
player.src({src: 'video.mp4', type: 'video/mp4'});
 // set second arg to false so that they get auto-removed on source change
player.addRemoteTextTrack({kind: 'captions', src: 'text.vtt', srclang: 'en'}, false);
```
Now when the player loads, this captions track is actually missing because it was removed.

## Specific Changes proposed
Instead of adding auto-removal tracks immediately to the list, wait until we've selected a source before adding them in.
This probably needs a test.

Fixes #4403 and #4315.

### DIFF
--- a/test/unit/tracks/text-tracks.test.js
+++ b/test/unit/tracks/text-tracks.test.js
@@ -551,4 +551,26 @@ if (Html5.isSupported()) {
     this.clock.tick(1);
     assert.equal(player.textTracks().length, 1, 'we have one text track');
   });
+
+  QUnit.test('auto remove tracks added right before a source change will be cleaned up', function(assert) {
+    const player = TestHelpers.makePlayer({
+      techOrder: ['html5']
+    });
+
+    console.log(player.isReady_, player.tech_.isReady_);
+
+    const track = {
+      kind: 'kind',
+      src: 'src',
+      language: 'language',
+      label: 'label',
+      default: 'default'
+    };
+
+    player.addRemoteTextTrack(track, false);
+    player.src({src: 'example.mp4', type: 'video/mp4'});
+
+    this.clock.tick(1);
+    assert.equal(player.textTracks().length, 0, 'we do not have any tracks left');
+  });
 }
--- a/test/unit/tracks/text-tracks.test.js
+++ b/test/unit/tracks/text-tracks.test.js
@@ -528,3 +528,27 @@ QUnit.test('removeRemoteTextTrack should be able to take both a track and the re
               'the track element was removed correctly');
   player.dispose();
 });
+
+if (Html5.isSupported()) {
+  QUnit.test('auto remove tracks should not clean up tracks added while source is being added', function(assert) {
+    const player = TestHelpers.makePlayer({
+      techOrder: ['html5']
+    });
+
+    console.log(player.isReady_, player.tech_.isReady_);
+
+    const track = {
+      kind: 'kind',
+      src: 'src',
+      language: 'language',
+      label: 'label',
+      default: 'default'
+    };
+
+    player.src({src: 'example.mp4', type: 'video/mp4'});
+    player.addRemoteTextTrack(track, false);
+
+    this.clock.tick(1);
+    assert.equal(player.textTracks().length, 1, 'we have one text track');
+  });
+}
--- a/test/unit/tracks/text-tracks.test.js
+++ b/test/unit/tracks/text-tracks.test.js
@@ -532,7 +532,10 @@ QUnit.test('removeRemoteTextTrack should be able to take both a track and the re
 if (Html5.isSupported()) {
   QUnit.test('auto remove tracks should not clean up tracks added while source is being added', function(assert) {
     const player = TestHelpers.makePlayer({
-      techOrder: ['html5']
+      techOrder: ['html5'],
+      html5: {
+        nativeTextTracks: false
+      }
     });
 
     const track = {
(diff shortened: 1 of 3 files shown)


test/unit/tracks/text-tracks.test.js
```diff
@@ -528,3 +528,49 @@ QUnit.test('removeRemoteTextTrack should be able to take both a track and the re
...
   player.dispose();
 });
+
+if (Html5.isSupported()) {
+  QUnit.test('auto remove tracks should not clean up tracks added while source is being added', function(assert) {
+    const player = TestHelpers.makePlayer({
+      techOrder: ['html5']
+    });
+
+    console.log(player.isReady_, player.tech_.isReady_);
```
misteroneill: Linter is complaining about this.

test/unit/tracks/text-tracks.test.js
```diff
@@ -528,3 +528,49 @@ QUnit.test('removeRemoteTextTrack should be able to take both a track and the re
...
+    this.clock.tick(1);
+    assert.equal(player.textTracks().length, 1, 'we have one text track');
+  });
+
+  QUnit.test('auto remove tracks added right before a source change will be cleaned up', function(assert) {
+    const player = TestHelpers.makePlayer({
+      techOrder: ['html5']
+    });
+
+    console.log(player.isReady_, player.tech_.isReady_);
```
misteroneill: Linter is complaining about this.

gkatsev:
Closing this PR for the moment so I can re-open it via upstream so tests will run against browserstack.

gkatsev:
New PR: https://github.com/videojs/video.js/pull/4450

PR #4437, opened 2017-06-22
fix: Safari picture-in-picture triggers fullscreenchange

mister-ben:
## Description
When picture-in-picture mode is entered on Safari, `webkitbeginfullscreen` is triggered which results in a proxied `fullscreenchange` event and adding the fullscreen class to the player. That causes the tech element to collapse to zero height so that the "this video is playing in picture in picture" placeholder is not visible. 

## Specific Changes proposed
On `webkitbeginfullscreen` check whether the presentation mode is `picture-in-picture` before proxying a `fullscreenchange` event.

## Requirements Checklist

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -430,9 +430,12 @@ class Html5 extends Tech {
     };
 
     const beginFn = function() {
-      this.one('webkitendfullscreen', endFn);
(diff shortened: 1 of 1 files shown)


gkatsev:
I guess we'll want to also fix this in 5.x.

PR #4440, opened 2017-06-23
Use passive event listeners for touchstart/touchmove

### LINKED ISSUE #4432: Chrome Violation: Non-passive scroll handlers
## Description
Chrome is logging "Violations" for videojs's scroll handlers.

```
[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive.
external_embed-b5b3e31….js:936
[Violation] Added non-passive event listener to a scroll-blocking 'touchmove' event. Consider marking event handler as 'passive' to make the page more responsive.
```

Related Links:
https://developers.google.com/web/updates/2016/06/passive-event-listeners
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support

## Steps to reproduce
1. Open Chrome and dev tools
2. Set console level to "Verbose"
3. Go to http://videojs.com
4. Look in the console for "Violation" warnings.

### Expected
Videojs should add `passive: true` when adding event listeners, so Chrome (and others) can be more performant.  The Violation warnings should go away as well.

### Actual
Violation warnings in the dev tools console.

#### videojs
5.19.2
#### browsers
Chrome Version 59.0.3071.109 (Official Build) (64-bit)
#### OSes
I tested on mac
### plugins
No

mister-ben:
## Description
Not using passive event listeners now causes Chrome to log warnings to the console, and possibly has a performance impact
Fixes #4432

## Specific Changes proposed
Where supported, use passive event listeners for `touchmove` and `touchstart` 

## Requirements Checklist

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -207,17 +207,19 @@ export function fixEvent(event) {
  */
 let _supportsPassive = false;
 
-try {
-  const opts = Object.defineProperty({}, 'passive', {
-    get() {
-      _supportsPassive = true;
-    }
-  });
+(function() {
+  try {
+    const opts = Object.defineProperty({}, 'passive', {
+      get() {
+        _supportsPassive = true;
+      }
+    });
 
-  window.addEventListener('test', null, opts);
-} catch (e) {
-  // disregard
-}
+    window.addEventListener('test', null, opts);
+  } catch (e) {
+    // disregard
+  }
+})();
 
 /**
  * Touch events Chrome expects to be passive
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -220,6 +220,14 @@ try {
 }
 
 /**
+ * Touch events Chrome expects to be passive
+ */
+const passiveEvents = [
+  'touchstart',
+  'touchmove'
+];
+
+/**
  * Add an event listener to element
  * It stores the handler function in a separate cache object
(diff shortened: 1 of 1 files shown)


src/js/utils/events.js
```diff
@@ -273,7 +290,13 @@ export function on(elem, type, fn) {
 
   if (data.handlers[type].length === 1) {
     if (elem.addEventListener) {
-      elem.addEventListener(type, data.dispatcher, false);
+      let options = false;
+
+      if (_supportsPassive &&
+        ['touchstart', 'touchmove'].indexOf(type) > -1) {
```
misteroneill: I wonder if this array ought to be defined once in the parent scope rather than inline. Something like:

```js
const passiveEvents = [
  'touchstart',
  'touchmove'
];
```

src/js/utils/events.js
```diff
@@ -203,6 +203,31 @@ export function fixEvent(event) {
 }
 
 /**
+ * Whether passive event listeners are supported
+ */
+let _supportsPassive = false;
+
+try {
```
gkatsev: we should wrap this try/catch in an iife.

misteroneill:
Looks like this will fix #4432 (we should not that in the merge commit).

gkatsev:
We probably should make this against 5.x as well.

PR #4443, opened 2017-06-26
fix: player.duration() should return NaN if duration is not known

alex-barstow:
## Description
`player.duration()` currently returns 0 if the duration is not known, when it should return `NaN`. The problem is that if `NaN` is passed to `player.duration()` as an argument, [we set the duration to 0](https://github.com/videojs/video.js/blob/5.x/src/js/player.js#L1700). If `player.cache_.duration` was set to `NaN` by other means, `player.duration()` [would still return 0](https://github.com/videojs/video.js/blob/5.x/src/js/player.js#L1697). If approved, this will require a sister PR in the 5.x branch as well.

## Specific Changes proposed
Modify the player `duration()` method so that it will 1.) set the cached duration to `NaN` if it is passed in as an argument, and 2.) return the proper value when called without an argument.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1747,7 +1747,6 @@ class Player extends Component {
    */
   duration(seconds) {
     if (seconds === undefined) {
-      
       // return NaN if the duration is not known
       return this.cache_.duration !== undefined ? this.cache_.duration : NaN;
     }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1747,10 +1747,12 @@ class Player extends Component {
    */
   duration(seconds) {
     if (seconds === undefined) {
-      return this.cache_.duration || 0;
+      
+      // return NaN if the duration is not known
+      return this.cache_.duration !== undefined ? this.cache_.duration : NaN;
     }
 
-    seconds = parseFloat(seconds) || 0;
+    seconds = parseFloat(seconds);
 
     // Standardize on Inifity for signaling video is live
     if (seconds < 0) {


src/js/player.js
```diff
@@ -1747,10 +1747,11 @@ class Player extends Component {
    */
   duration(seconds) {
     if (seconds === undefined) {
-      return this.cache_.duration || 0;
+      // return NaN if the duration is not known
+      return this.cache_.duration !== undefined ? this.cache_.duration : NaN;
     }
 
-    seconds = parseFloat(seconds) || 0;
+    seconds = parseFloat(seconds);
```
gkatsev: I think in here we still want to do `|| 0` because if what is passed in, isn't a number we should just assume 0.

alex-barstow: I think the problem with leaving the `|| 0` is that if `player.techGet_('duration')` equals `NaN` when [handleTechDurationChange_()](https://github.com/videojs/video.js/blob/cc6e82442fcefb3b62a6e33a588d7f0b7152fd3d/src/js/player.js#L1368) is first called, then the initial value of `player.cache_.duration` is 0 rather than `NaN`. At least that was the case with the page I was testing with.

gkatsev: Ah, I see. 🤔
I guess maybe this is fine since `parseFloat()` will return `NaN` for bad values. So, we'd just want to make sure the rest of this method handles `NaN` correctly, if it doesn't already.

gkatsev: Yeah, looks like this is fine but I think the duration display's `updateDuration` method needs to be updated to account for `NaN` values. Specifically, what happens if we start with unknown duration, transitioning from unknown duration to a known duration, and transitioning from a known duration to an unknown duration. That last one in particular I think might be problematic, though, not sure how common it is.

alex-barstow: [This conditional](https://github.com/videojs/video.js/blob/c9022795fdfcda166a79bef51dfeab29c1f7c191/src/js/control-bar/time-controls/duration-display.js#L74) will be `false` if `player.duration() = NaN`, so off the top of my head I can't think of a problem going from a unknown duration to a known duration. If we are transitioning from a known duration to an unknown duration, I believe the duration display (as currently written) would keep it's previous 'known' state until the new duration becomes known and the display updates. Assuming I'm right about that, would this be acceptable behavior?

gkatsev: Yeah, that's probably fine. Just wanted someone to think about it in some more depth.

gkatsev:
I realized one thing that this is missing: tests :)

gkatsev:
@alex-barstow I pulled it in as is, but would be great to get some tests later, maybe when you get the 5.x PR ready.

PR #4454, opened 2017-06-28
docs: Fix Player#src API documentation.

misteroneill:
The API docs for `Player#src` are inaccurate/incomplete.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2247,17 +2247,16 @@ class Player extends Component {
   }
 
   /**
-   * The source function updates the video source
-   * There are three types of variables you can pass as the argument.
-   * **URL string**: A URL to the the video file. Use this method if you are sure
-   * the current playback technology (HTML5/Flash) can support the source you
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -2247,17 +2247,16 @@ class Player extends Component {
...
-   * the current playback technology (HTML5/Flash) can support the source you
-   * provide. Currently only MP4 files can be used in both HTML5 and Flash.
+   * Get or set the video source.
    *
-   * @param {Tech~SourceObject|Tech~SourceObject[]} [source]
-   *        One SourceObject or an array of SourceObjects
+   * @param {Tech~SourceObject|Tech~SourceObject[]|string} [source]
+   *        A SourceObject, an array of SourceObjects, or a string referencing
+   *        a URL to a media source. If not provided, this method acts as a
+   *        getter.
```
mister-ben: Is it worth recommending using an object rather than a string here so that source selection can take account of the source's type?

misteroneill: Done, thanks.

PR #4462, opened 2017-07-02
chore: switch to using chrome for testing PRs on travis

gkatsev:


gkatsev:
@dmlap the tests that were added in https://github.com/videojs/video.js/pull/4450 seem to break in travis firefox *only*. I tried really hard to reproduce locally but cannot. It works great locally directly in firefox or through karma. I even installed a linux VM to try out directly and via karma and it passes as well. Given that this blocks a few PRs, I think we should proceed with this change and then maybe we can re-enable firefox after debugging on travis directly.

PR #4471, opened 2017-07-07
docs: make jsdoc generate anchor names so ToC links work

gkatsev:


.jsdoc.json
```diff
@@ -35,6 +35,7 @@
   },
   "plugins": ["plugins/markdown"],
   "markdown": {
-    "tags": ["example"]
+    "tags": ["example"],
+    "idInHeadings": true
```
gkatsev: The way that marked/jsdoc add anchor links isn't quite compatible with GFM, so, it warranted the changes in `fix-api-docs.js` below.

build/fix-api-docs.js
```diff
@@ -5,7 +5,26 @@ var apiPath = path.join(__dirname, '..', 'docs', 'api');
 var replacements = [
   {find: /\/docs\/guides\/(.+)\.md/g, replace: 'tutorial-$1.html'},
   {find: /tutorial-tech.html/g, replace: 'tutorial-tech_.html'},
-  {find: /\/docs\/guides\//g, replace: '#'}
+  {find: /\/docs\/guides\//g, replace: '#'},
+  {find: /(\<h[1-6] id="(?:.*)?)video-js(.*)?"\>/g, replace: '$1videojs$2">'},
+  {find: /(\<h[1-6] id="(?:.*)?)don-t(.*)?"\>/g, replace: '$1dont$2">'},
+  {find: /(\<h[1-6] id="(?:.*)?)node-js(.*)?"\>/g, replace: '$1nodejs$2">'},
+  {find: /(\<h[1-6] id="(?:.*)?)vtt-js(.*)?"\>/g, replace: '$1vttjs$2">'},
```
gkatsev: headings with a period get a dash in jsdoc but not in GFM

build/fix-api-docs.js
```diff
@@ -5,7 +5,26 @@ var apiPath = path.join(__dirname, '..', 'docs', 'api');
...
+  {find: /(\<h[1-6] id=")-(.*)("\>)/g, replace: '$1$2$3'},
+  {find: /(\<h[1-6] id=")(.*)-("\>)/g, replace: '$1$2$3'},
```
gkatsev: headings with backticks get a dash in there. Need two rules to get those that start and end with them.

build/fix-api-docs.js
```diff
@@ -5,7 +5,26 @@ var apiPath = path.join(__dirname, '..', 'docs', 'api');
...
+  {find: /(\<h[1-6] id=".*)-docs-guides-.*-md("\>)/g, replace: '$1$2'},
```
gkatsev: In the main page, a lot of headings are linked and jsdoc adds the link pieces into the heading but not GFM.

build/fix-api-docs.js
```diff
@@ -5,7 +5,26 @@ var apiPath = path.join(__dirname, '..', 'docs', 'api');
...
+  {find: '<h2 id="q-how-can-i-hide-the-links-to-my-video-subtitles-audio-tracks">',
+   replace: '<h2 id="q-how-can-i-hide-the-links-to-my-videosubtitlesaudiotracks">'},
+  {find: '<h3 id="dispose-http-docs-videojs-com-player-html-dispose">',
+   replace: '<h3 id="dispose">'},
+  {find: '<h4 id="effect-on-player-width-and-player-height">',
+   replace: '<h4 id="effect-on-playerwidth-and-playerheight">'},
+  {find: '<h4 id="i-want-to-have-a-single-source-and-dont-care-about-live-adaptive-streaming">',
+   replace: '<h4 id="i-want-to-have-a-single-source-and-dont-care-about-liveadaptive-streaming">'},
+  {find: '<h2 id="api-docs-api">', replace: '<h2 id="api-docs">'},
+  {find: '<h2 id="guides-docs-guides">', replace: '<h2 id="guides">'}
```
gkatsev: various specific changes that are hard to target programmatically and are just changed directly.

build/fix-api-docs.js
gkatsev: I probably should move some of these comments directly into the code.

misteroneill: I don't think this needs changes, but I feel obliged to post this... https://stackoverflow.com/a/1732454

build/fix-api-docs.js
misteroneill: I guess my only question with this section is whether it would make more sense to parse the HTML into a DOM and loop through all the heads, looking at their `id` attributes? It might be less error/typo-prone than this approach.

gkatsev: Probably best option would be to make a PR to jsdoc to expose the marker renderer so we can override the headings portion to our likings. For example, we're still missing a chain icon thing near each heading to click to be taken to that section.

PR #4481, opened 2017-07-12
Add GA note to primary readme

mmcc:
We already added a similar note to the [getting started guide](http://videojs.com/getting-started/) on the website, so it should probably be here too for the folks that don't make it off Github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,20 @@ Thanks to the awesome folks over at [Fastly][fastly], there's a free, CDN hosted
 ```
 > For the latest version of video.js and URLs to use, check out the [Getting Started][getting-started] page on our website.
 
-> In the CDN-hosted versions we include a [stripped down Google Analytics pixel](https://github.com/videojs/cdn/blob/master/src/analytics.js) that tracks a random sampling (currently 1%) of players loaded from the CDN. This allows us to see (roughly) what browsers are in use in the wild, along with other useful metrics such as OS and device. If you'd like to disable analytics, you can simply include the following global before including Video.js via the free CDN:
+> In the `vjs.zencdn.net` CDN-hosted versions of Video.js we include a [stripped down Google Analytics pixel](https://github.com/videojs/cdn/blob/master/src/analytics.js) that tracks a random sampling (currently 1%) of players loaded from the CDN. This allows us to see (roughly) what browsers are in use in the wild, along with other useful metrics such as OS and device. If you'd like to disable analytics, you can simply include the following global before including Video.js via the free CDN:
 >
 > ```html
 > <script>window.HELP_IMPROVE_VIDEOJS = false;</script>
+> ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -25,9 +25,14 @@ Thanks to the awesome folks over at [Fastly][fastly], there's a free, CDN hosted
 <link href="//vjs.zencdn.net/5.19/video-js.min.css" rel="stylesheet">
 <script src="//vjs.zencdn.net/5.19/video.min.js"></script>
 ```
-
 > For the latest version of video.js and URLs to use, check out the [Getting Started][getting-started] page on our website.
 
+> We include a [stripped down Google Analytics pixel](https://github.com/videojs/cdn/blob/master/src/analytics.js) that tracks a random sampling (currently 1%) of players loaded from the CDN. This allows us to see (roughly) what browsers are in use in the wild, along with other useful metrics such as OS and device. If you'd like to disable analytics, you can simply include the following global before including Video.js via the free CDN:
```
mister-ben: Just to be extra explicit that this is the CDN version how about "Only the CDN version of Video.js includes this analytics pixel. If you'd like to use the CDN hosted version but would like to disable analytics, you can simply..."

mmcc: Sorry, I've been in the African bush (literally) and just saw this 😬 I'll update the PR when I get to more reliable internets.

mmcc:
@mister-ben @OwenEdwards Made the update to be more explicit, let me know what you all think.

README.md
```diff
@@ -25,9 +25,14 @@ Thanks to the awesome folks over at [Fastly][fastly], there's a free, CDN hosted
...
 ```
-
 > For the latest version of video.js and URLs to use, check out the [Getting Started][getting-started] page on our website.
 
+> In the CDN-hosted versions we include a [stripped down Google Analytics pixel](https://github.com/videojs/cdn/blob/master/src/analytics.js) that tracks a random sampling (currently 1%) of players loaded from the CDN. This allows us to see (roughly) what browsers are in use in the wild, along with other useful metrics such as OS and device. If you'd like to disable analytics, you can simply include the following global before including Video.js via the free CDN:
+>
+> ```html
+> <script>window.HELP_IMPROVE_VIDEOJS = false;</script>
+> ```
+
```
OwenEdwards: Suggest that we add something like "Alternatively, you can [include Video.js in your website using npm](http://videojs.com/getting-started/#download-npm) or [Bower](http://videojs.com/getting-started/#download-bower) - those distributions **do not** include the Google Analytics tracking." here, too.

mister-ben:
"CDN hosted" could be more specific to make it clear it's the version hosted on vjs.zencdn.net only, and not anything you'd find on unpkg, cdnjs etc.

thijstriemstra:
> "CDN hosted" could be more specific to make it clear it's the version hosted on vjs.zencdn.net only,

And, for full disclosure?, that it contains google analytics code.

gkatsev:
@mmcc @mister-ben @OwenEdwards I made a PR against this branch with changes: https://github.com/mmcc/video.js/pull/10

gkatsev:
And it's been merged here.

gkatsev:
Merged this. If we think there are other additions, please make a PR or post here and we can update it again.

PR #4496, opened 2017-07-17
Fix: issue where tracks are disabled and cuepoints are cleared in iOS native player

alex-barstow:
## Description
This PR addresses a behavior in the iOS Safari native player: if a captions text track has its mode set to `showing`, Safari will set the mode of any other present tracks to `disabled`, including metadata tracks. In iOS Safari, specifically, this also unfortunately means that the `cues` and `activeCues` arrays for those disabled tracks are set to `null`. This is a problem for developers who wish to have functional cue points in iOS fullscreen video playback.

## Specific Changes proposed
Add a private `Html5` tech method that snapshots the state of all metadata tracks leading up to a `'fullscreenchange'` event, then retroactively restores the metadata tracks to their pre-fullscreen state after the first track `'change'` event that occurs during native playback. Although it isn't ideal, we are presuming that the first change event will be the one initiated by Safari. The inevitable (but perhaps necessary) downside to this is that it removes the guarantee that all metadata tracks can be programmatically disabled at any point in the video. However, functional cue points in the iOS native player is almost certainly a more important feature to developers/publishers than the guaranteed ability to disable metadata tracks during iOS native playback.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -131,7 +131,7 @@ class Html5 extends Tech {
 
         // listen for first 'change' event only while in fullscreen
         if (data.isFullscreen) {
-          const textTracks = myPlayer.textTracks();
+          const textTracks = this.textTracks();
 
           textTracks.one('change', () => {
             for (let i = 0; i < textTracks.length; i++) {
@@ -141,9 +141,9 @@ class Html5 extends Tech {
                 track.mode = 'hidden';
               }
             }
-          })
+          });
         }
-      })
+      });
     }
   }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -90,7 +90,9 @@ class Html5 extends Tech {
     }
 
     // prevent iOS Safari from disabling metadata text tracks during native playback
-    this.reEnableMetadataTracksInIOSNativePlayer_();
+    if (browser.IS_IOS) {
+      this.saveAndRestoreMetadataTrackModeInIOSNativePlayer_();
+    }
 
     // Determine if native controls should be used
     // Our goal should be to get the custom controls on mobile solid everywhere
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -115,6 +120,59 @@ class Html5 extends Tech {
...
+    this.on('fullscreenchange', (event, data) => {
+      if (data.isFullscreen) {
+        // stop updating the snapshot now that we are in fullscreen
+        textTracks.off('change', takeMetadataTrackSnapshot);
+
+        textTracks.one('change', () => {
+          for (let i = 0; i < metadataTracksPreFullscreenState.length; i++) {
+            const storedTrack = metadataTracksPreFullscreenState[i];
+
+            storedTrack.track.mode = storedTrack.storedMode;
```
alex-barstow: I think I should wrap this in a `if (storedTrack.track.mode === 'disabled') {...}` conditional so that the track mode is only updated if it is currently disabled. That way devs can still programmatically _enable_ metadata tracks during native iOS playback without us forcibly restoring them to their previous `disabled` state. Thoughts?

gkatsev: Sounds good. Also, we should check to see that the `storedMode` is different from the current `mode`, so, we're only updating it if necessary.

alex-barstow: @gkatsev made the changes.

alex-barstow:
Also, if approved this will need a sister in `5.x`

src/js/tech/html5.js
```diff
@@ -115,6 +120,61 @@ class Html5 extends Tech {
...
+      }
+    };
+
+    // snapshot each metadata track's initial state, and update the snapshot
+    // each time there is a track 'change' event
+    takeMetadataTrackSnapshot();
+    textTracks.on('change', takeMetadataTrackSnapshot);
+
+    // when we enter fullscreen playback, restore all track modes to their initial state
+    this.on('fullscreenchange', (event, data) => {
```
gkatsev: Optionally, if you listen to the `webkitenterfullscreen` and `webkitexitfullscreen` events, you wouldn't need to do an extra `isFullscreen` check but also you'd be able to remove the user agent check above.

src/js/tech/html5.js
```diff
@@ -115,6 +120,61 @@ class Html5 extends Tech {
...
+      if (data.isFullscreen) {
+        // stop updating the snapshot now that we are in fullscreen
+        textTracks.off('change', takeMetadataTrackSnapshot);
+
+        textTracks.one('change', () => {
+          for (let i = 0; i < metadataTracksPreFullscreenState.length; i++) {
+            const storedTrack = metadataTracksPreFullscreenState[i];
+
+            if (storedTrack.track.mode === 'disabled' && storedTrack.track.mode !== storedTrack.storedMode) {
+              storedTrack.track.mode = storedTrack.storedMode;
```
alex-barstow: It's worked in all the tests I've done. I tried to handle the track snapshotting the same way it's handled in contrib-ads ([storing](https://github.com/videojs/videojs-contrib-ads/blob/master/src/snapshot.js#L61-L69) and [restoration](https://github.com/videojs/videojs-contrib-ads/blob/master/src/snapshot.js#L97-L107))

src/js/tech/html5.js
alex-barstow: I like that better. Change made.

src/js/tech/html5.js
```diff
@@ -115,6 +118,64 @@ class Html5 extends Tech {
...
+    // when we enter fullscreen playback, stop updating the snapshot and
+    // restore all track modes to their pre-fullscreen state
+    this.on('webkitbeginfullscreen', () => {
+      textTracks.removeEventListener('change', takeMetadataTrackSnapshot);
+      textTracks.addEventListener('change', restoreTrackMode);
+    });
+
+    // start updating the snapshot again after leaving fullscreen
+    this.on('webkitendfullscreen', () => {
+      textTracks.addEventListener('change', takeMetadataTrackSnapshot);
```
gkatsev: we probably should remove this handle first and then re-add it, just in case we somehow get two `webkitendfullscreen` event without a `webkitbeginfullscreen` event.

misteroneill:
@alex-barstow Can you make a sister PR for 5.x as well? Thanks!

PR #4497, opened 2017-07-17
fix: only change focus from BPB if not a mouse click

gkatsev:
This is a simple change to not change the focus to the play toggle from the Big Play Button when clicking on it with the mouse.
There's still work to be done with regards to the focus ring and mouse vs key events but that's for another time and would probably be more involved.

We cancel out focus changing if the mouse was used to click the button, or, as far as we could tell via the `mousedown` handler. Unfortunately, using VoiceOver's functionality (via control-option space), we also get a `mousedown` event. However, those events do not add `clientX` and `clientY` to the `click` event like a true mouse event does.

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -19,7 +19,6 @@ class BigPlayButton extends Button {
     this.on('mousedown', this.handleMouseDown);
   }
 
-
   /**
    * Builds the default DOM `className`.
    *
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -11,6 +11,14 @@ import Component from './component.js';
  * @extends Button
  */
 class BigPlayButton extends Button {
+  constructor(player, options) {
+    super(player, options);
+
+    this.mouseused_ = false;
+
+    this.on('mousedown', this.handleMouseDown);
+  }
+
 
   /**
    * Builds the default DOM `className`.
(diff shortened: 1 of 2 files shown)


src/js/button.js
```diff
@@ -116,6 +124,10 @@ class Button extends ClickableComponent {
     // Pass keypress handling up for unsupported keys
     super.handleKeyPress(event);
   }
+
+  handleMouseDown(event) {
```
misteroneill: Linter will probably complain about missing JSDoc here.

src/js/button.js
```diff
@@ -12,6 +12,13 @@ import {assign} from './utils/obj';
  * @extends ClickableComponent
  */
 class Button extends ClickableComponent {
+  constructor(player, options) {
+    super(player, options);
+
+    this._mouseused = false;
```
misteroneill: Is there a reason this property name doesn't follow other conventions? Most of the time, it's camel case with a trailing underscore (i.e. `mouseUsed_`).

gkatsev: I didn't remember where we put the `_`. I'll move it to the end.

src/js/big-play-button.js
```diff
@@ -36,6 +36,11 @@ class BigPlayButton extends Button {
   handleClick(event) {
     const playPromise = this.player_.play();
 
+    // exit early if clicked via the mouse
+    if (this._mouseused && event.clientX && event.clientY) {
```
BrandonOCasey: Could we just use `clientX` or `clientY` without `mouseused` here? could be a terrible idea but I figured I would throw it out there.

src/js/button.js
BrandonOCasey: Is there any reason not stick this in `BigPlayButton` instead of the `Button` class?

gkatsev: You're right. That would be better since I want this as self contained as possible.

PR #4517, opened 2017-07-25
chore: add a release script to makes things easier

gkatsev:
I've been using this to release the 5.x branch for a while. I've also just ran it through http://www.shellcheck.net/

### DIFF
--- a/build/bin/release-next.sh
+++ b/build/bin/release-next.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 echo "Release Type?"
-read -p "> " rtype
+read -rp "> " rtype
 
 echo grunt version:${rtype}
 grunt version:${rtype}
--- a/build/bin/release-next.sh
+++ b/build/bin/release-next.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+echo "Release Type?"
+read -p "> " rtype
+
+echo grunt version:${rtype}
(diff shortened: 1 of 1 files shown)


build/bin/release-next.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+echo "Release Type?"
+read -rp "> " rtype
+
+echo grunt version:${rtype}
+grunt version:${rtype}
+version=$(./build/bin/version)
```
BrandonOCasey: We may want to use something like this here:
```sh
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
version="$("$SCRIPT_DIR"/version)"
```

This will allow you to use the script outside of the root directory of video.js without issues

misteroneill: I like this idea - I use something similar in my horrible bash scripts. 😉

gkatsev: I'm not sure we want to allow that, though. I think some of the grunt scripts depend on being in the root.

BrandonOCasey: I guess we could just have the script cd to the video.js root dir when we start it?

gkatsev: I'm also a bit weary of changing this script since I know it works as is. I would prefer to just document that you need to run this from the root of the project and we can fix it later if this becomes a problem.

PR #4523, opened 2017-07-26
fix: only change focus from BPB if not a mouse click

gkatsev:
5.x version of https://github.com/videojs/video.js/pull/4497/

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -19,7 +19,6 @@ class BigPlayButton extends Button {
     this.on('mousedown', this.handleMouseDown);
   }
 
-
   /**
    * Builds the default DOM `className`.
    *
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -56,9 +56,13 @@ class BigPlayButton extends Button {
       return;
     }
 
-    this.setTimeout(function() {
-      playToggle.focus();
-    }, 1);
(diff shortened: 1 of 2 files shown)


src/js/big-play-button.js
```diff
@@ -44,9 +56,23 @@ class BigPlayButton extends Button {
       return;
     }
 
-    this.setTimeout(function() {
-      playToggle.focus();
-    }, 1);
+    if (playPromise) {
+      playPromise.then(() => playToggle.focus());
+    } else {
+      this.setTimeout(function() {
```
BrandonOCasey: maybe store this function since it is used above? or just use the shorthand again?

src/js/big-play-button.js
```diff
@@ -44,9 +56,23 @@ class BigPlayButton extends Button {
       return;
     }
 
-    this.setTimeout(function() {
-      playToggle.focus();
-    }, 1);
+    if (playPromise) {
```
misteroneill: Should this duck-type the play promise (e.g. `if (playPromise && playPromise.then) {`)?

gkatsev: This was ported over from master, it's probably safer

src/js/big-play-button.js
gkatsev: yeah, might make sense. Ported this from master where it's written this way.

PR #4555, opened 2017-08-09
chore: improve dev and beginner experience

gkatsev:
Makes the output of dev rollup and babel watch more quiet.
Add a task to copy .example files over to non .example files if the non .example files are *not* present.

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -466,7 +466,7 @@ module.exports = function(grunt) {
         }
       },
       babel: {
-        command: 'npm run babel -- --watch',
+        command: 'npm run babel -- --watch --quiet',
         options: {
           preferLocal: true
         }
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
LGTM but I have a question/alternative approach. Is there any reason not to move `.example` files to `.html` in the repo rather than on `dev`?

gkatsev:
It's because sandbox is in gitingore but the files are force-added to the git repo. That means that if you modify that file and use `git add -a` those changes will end up in your commit. It's probably a bit less of an issue currently since when you make a PR the default is to allow maintainers access to that branch but it could be a problem still.

BrandonOCasey:
couldn't we just move them, keep them under gitignore, and then force add them for this PR?

gkatsev:
Yes but changes to forcefully added files still show up as changes made, which is the issue at hand.

PR #4559, opened 2017-08-10
Update playback rate menu for the new player layout

rafaelgaspar:
## Description
Bug: the font-size and line-height were not following the new player layout.

## Specific Changes proposed
Change both the rate value and the menu to use the new player layout with font-size of 1.0em and line-height of 3.0em.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
I think the design of the playback rate button was made specifically to make that number large to look more inline with the font-icons than the other text in the control bar.
@mmcc do you remember the design decisions around it?

Issue #4547, opened 2017-08-07
Subs Caps Button isn't sized fully in v6


gkatsev:
## Description
The new Subs Caps Button in v6 has full functionality, however, when you click on it, you can see that the focus outline is sized to only the font-icon rather than the entire button itself.
![screen shot 2017-08-07 at 17 17 37](https://user-images.githubusercontent.com/535884/29046242-91dd43fe-7b94-11e7-97ec-1f8a2739e958.png)
Where as it should look like so:
![screen shot 2017-08-07 at 17 20 14](https://user-images.githubusercontent.com/535884/29046263-b33779b6-7b94-11e7-912c-4f2194b0f263.png)

## Steps to reproduce
1. Create a player
2. Add some captions or subtitles
3. Play the video
4. Click on the subs caps button.

### Expected
Expected to outline the entire button
![screen shot 2017-08-07 at 17 20 14](https://user-images.githubusercontent.com/535884/29046263-b33779b6-7b94-11e7-912c-4f2194b0f263.png)

### Actual
The outline only covers the font icon itself
![screen shot 2017-08-07 at 17 17 37](https://user-images.githubusercontent.com/535884/29046242-91dd43fe-7b94-11e7-97ec-1f8a2739e958.png)

## Proposed solution
In the [`button.scss`](https://github.com/videojs/video.js/blob/master/src/css/components/_button.scss) file, add a new selector that targets `.vjs-button` class inside of a `.vjs-control` class and sizes it to be `100%`
```css
.vjs-control .vjs-button {
  width: 100%;
  height: 100%;
}
```

To rebuild the styles locally, you can then run `grunt skin` or `npm run grunt -- skin`.

agiridharan:
HI I would like to help with this bug issue! Is there anyone else working on it right now?

gkatsev:
@agiridharan feel free to grab this. No one is working on it as far as I know. Thanks!

agiridharan:
Hi @gkatsev I'm not understanding how to setup the video.js. I have run the build command, but I'm stuck on what I should do after that.

gkatsev:
[...]

PR #4548, opened 2017-08-08
Fix: Subs caps button size issue

Kishan08:
Fix the subs caps button size issue [(#4547)](https://github.com/videojs/video.js/issues/4547)

PR #4560, opened 2017-08-10
fix width and height of cc button

mugendin:
add as proposed 100% width and height and reorder css properties according to https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/ fixies issue #4547

## Description
Bug fix of issue #4547
Add width 100% and height 100%
also change styles order in button.scss as in https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

Issue:
https://github.com/videojs/video.js/issues/4547

## Requirements Checklist
- [ 1] Feature implemented / Bug fixed
  - [ 1] Change has been verified in an actual browser (Chome, Firefox, IE) 
  Tested in Chrome and Firefox
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_button.scss
+++ b/src/css/components/_button.scss
@@ -1,10 +1,12 @@
 .video-js button {
-  background: none;
-  border: none;
-  color: inherit;
   display: inline-block;
-
   overflow: visible; // IE8
+  border: none;
+  width: 100%;
+  height: 100%;
+  background: none;
+  color: inherit;
+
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks. Unfortunately, this will cause issues for buttons that aren't in inside of a vjs-control. For example, buttons in the captions settings dialog
screen shot 2017-08-10 at 15 16 40.

This has already been fixed by #4548.

Issue #4351, opened 2017-05-11
'muted' option does not mute the player except in one specific case


alex-barstow:
## Description
Providing the `"muted": true` option does add the `muted` attribute to the `<video>` element, but does not actually mute content audio unless all of the following conditions are true:
- the option is passed into the `videojs()` function rather than the `data-setup` attribute
- the video.js CDN link is included in the `head` of the document instead of the `body`
- the player is loaded in Chrome

In all other cases, the player is not muted despite the presence of the `muted` attribute because the value of the `<video>` element `muted` property is `false`.

#### Test pages
Option provided via `data-setup`:
- [CDN link in head](http://jsbin.com/fukatapela/1/edit?html,output)
- [CDN link in body](http://jsbin.com/jifedupoho/1/edit?html,output)

Option provided via `videojs()` function:
- [CDN link in head](http://jsbin.com/xodesehuwa/1/edit?html,output)
- [CDN link in body](http://jsbin.com/sonadevuco/1/edit?html,output)

## Steps to reproduce
1. Open above test links in Chrome, Firefox, and Safari.
2. Begin playback and inspect `<video>` element

### Expected
- `muted` attribute is present on the `<video>` element
- `MuteToggle` button is in it's muted state
- audio should be muted

### Actual
- (pass) `muted` attribute **is** present on the `<video>` element in all cases
- (fail) `MuteToggle` button is **not** in it's muted state and content audio is **not** muted, except for the 3rd test link in Chrome

## Additional Information
I have not tested this behavior with video.js installed via `npm`
#### videojs
5.19.2
#### OSes
Tested on MacOS Sierra v.10.12.3

KevinBrogan:
I can confirm this.

```
player = videojs('videoPlayer',
{
	muted: false
})
```

does not work except on chrome, ie10 and ie11, if the video tag has the muted attribute.

A workaround is to do this

```
player = videojs('videoPlayer',
{
	muted: muteState
},
function()
{
	// regarding 'muted: muteState,' in the above setup,
	// muted is not being set properly by videojs on firefox edge or safari. set it here.
	player.el_.children[0].muted = muteState;
});
```

gkatsev:
Working on it! See #4562.

PR #4562, opened 2017-08-11
fix: make boolean attributes set and check both the associated property and the attribute

gkatsev:
I don't think this is quite complete yet but this is the initial implementation for making boolean attributes like `muted` set the property `video.muted = {{boolean}}` and the attribute `video.setAttribute('muted', 'muted')` when changing the value.
Also, in html5's createEl, we were setting/updating attributes for various properties but we were only setting the attributes and not also the properties but also autoplay was happening first rather than last which caused autoplay to fail because `muted` and `playsinline` weren't necessary set by that time.

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -375,7 +375,7 @@ export function getAttributes(tag) {
   // known boolean attributes
   // we can check for matching boolean properties, but older browsers
   // won't know about HTML5 boolean attributes that we still read from
-  const knownBooleans = ',' + 'autoplay,controls,loop,muted,default' + ',';
+  const knownBooleans = ',' + 'autoplay,controls,playsinline,loop,muted,default' + ',';
 
   if (tag && tag.attributes && tag.attributes.length > 0) {
     const attrs = tag.attributes;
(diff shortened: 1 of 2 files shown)


gkatsev:
Fixes https://github.com/videojs/video.js/issues/4351

src/js/utils/dom.js
```diff
@@ -375,7 +375,7 @@ export function getAttributes(tag) {
   // known boolean attributes
   // we can check for matching boolean properties, but older browsers
   // won't know about HTML5 boolean attributes that we still read from
-  const knownBooleans = ',' + 'autoplay,controls,loop,muted,default' + ',';
+  const knownBooleans = ',' + 'autoplay,controls,playsinline,loop,muted,default' + ',';
```
misteroneill: Should `default` here be `defaultMuted`?

gkatsev: I think this is supposed to be `default` because `<track>` els use a `default` attribute. Going to add `defaultMuted` in here.

misteroneill: Sounds good.

Issue #424, opened 2013-04-03
iOS: Fullscreen does not activate until after first play


dmlap:
On iOS devices, the fullscreen button is a no-op until after the video has been played for the first time. With the default video.js skin, this isn't an issue because the fullscreen button is not visible until the video has been played for the first time. It could be fixed, though, by making the first attempt to enter fullscreen mode implicitly trigger play or at least kick off the video load process.
### Steps to reproduce

Create a video.js skin that displays the fullscreen button before the video has been played (e.g. the "YouTube" skin)
Click the fullscreen button
### Expected

The video enters fullscreen mode
### Actual

Nothing happens

PR #4582, opened 2017-08-21
fix: rely on browser or tech to handle autoplay

gkatsev:
## Description
When setting the source, we were calling `play` if `autoplay` was set. We shouldn't do that and instead we should rely on the browser or tech (like Flash) to handle it properly.
I'm unsure whether this should be a `fix` or a `feat`, thoughts?

There are two more places that we should consider removing but they are probably OK because they're specific/bug fixes rather than general "if autoplay, call play" kind of deals.
There's one in `enterFullscreen` on the html5 tech, which should only be called on iOS devices.
https://github.com/videojs/video.js/blob/fe63992bd1e0d6e4eb7783584bce17c2796c4a2c/src/js/tech/html5.js#L548-L557

In the player, in `handleTechReady_`, there's another call to `play`, which is a fix for older chrome and older safari. This one seems more of a candidate for removal than the html5 one above.
https://github.com/videojs/video.js/blob/fe63992bd1e0d6e4eb7783584bce17c2796c4a2c/src/js/player.js#L1068-L1079

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1076,7 +1076,6 @@ class Player extends Component {
       } catch (e) {
         log('deleting tag.poster throws in some browsers', e);
       }
-      this.play();
     }
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2363,10 +2363,6 @@ class Player extends Component {
         this.load();
       }
 
-      if (this.options_.autoplay) {
-        this.play();
-      }
-
     // Set the source synchronously if possible (#2326)
     }, true);
 


gkatsev:
@videojs/core-committers any thoughts on the removal of those two `play()` calls?

In the `player` case, we can either delete the entire block or just the call to `play`. The try/catch+delete https://github.com/videojs/video.js/commit/4e0325013e04b73b8c8022cc62f55083d194ca6a and the whole block was added here: https://github.com/videojs/video.js/commit/e5595b1e38d0aa7c26bb2494e95995afe8ab83e9#diff-3b0266ff1c037b289ec624ab25b0272eR400

Seems like the html5 one was added 4 years ago for iOS 6 support: https://github.com/videojs/video.js/commit/7d67abdae8c6b611d6e2bb6cb2aab37deb20a8c1

Seems like removing these pieces could potentially cause older browsers to misbehave.

heff:
Could there be people relying on this behavior when switching videos? If so we might want this to be a major version change, unless it's breaking something now.

gkatsev:
It's theoretically possible but I find it unlikely. I think the reason the code was there was for flash but also because we did a bad job at setting attributes/properties on the video element, which was recently fixed in https://github.com/videojs/video.js/pull/4562.
Autoplay should continue to work assuming that the underlying tech supports the autoplay option. I guess we should survey to see if things like the youtube tech would break with this change or not.

misteroneill:
I think "fix" is appropriate here. If our goal is emulating/mirroring the standard behavior, this gets us closer to that in a way it wasn't before.

For `enterFullscreen` on the `Html5` tech, if it's platform-specific, maybe it should have that as a condition? What happens without that call to `play()`?

Yeah, I think that can probably be safely removed, based on my assumptions about browser usage.

ldayananda:
Although "fix" is appropriate for this project, it does mean a change to the behavior of the underlying tech/player which is concerning and I think would push this towards a "feat" so users have the ability to test and update their integrations. 

I agree with updating `enterFullScreen` to be platform specific if iOS is the only platform it is intended for.

`handleTechReady_` does seem like a good candidate for removal.

gkatsev:
> For enterFullscreen on the Html5 tech, if it's platform-specific, maybe it should have that as a condition? What happens without that call to play() in iOS?

`enterFullscreen` gets called if the tech says it supports fullscreen but native FS APIs aren't present. This is basically only the case for iOS. It seems like the issue is clicking the fullscreen button before ever hitting play on iOS used to be a no-op (https://github.com/videojs/video.js/issues/424).

Removing `play` from `handleTechReady_` seems to not cause any issues.

-------------

Also, I tested this change with videojs-youtube and it autoplays. Looking at a selection of other techs from the [wiki](https://github.com/videojs/video.js/wiki/Plugins#playback-techs-support-media-from-different-sources).

gkatsev:
I tested it with a few others but either couldn't get them to work at all or they don't autoplay regardless of this change (like the soundcloud tech).

mister-ben:
An advantage of using `play()` to "autoplay" would be to use the promise returned by some browsers to know whether autoplay succeeded. We don't do that currently, but being able to track plays that resulted from autoplay distinctly from manual could be useful.

PR #4591, opened 2017-08-28
improvement: PlayProgressBar Smooth Transition

### LINKED ISSUE #4526: How do I make the play progress transition smoother for VideoJS plugin?
I am trying to make the play progress transition smoother for VideoJS plugin. I've tried to edit the CSS but it causes a delay when the video is paused or stopped. Please help.

## Steps to reproduce
1. Play a short video that's about 5 seconds long.
2. You will notice that the slider is jumpy.
3. If you play a video on Youtube it is smooth even if it is a short video.

I am using the latest version of VideoJS.

vhmth:
Instead of tying the play progress update lifecycle to the timeupdate
event, the update happens every 15ms. This makes it so the play progress
transitions smoothly.

Fixes #4526

## Description

Here's a video!

https://www.useloom.com/share/d578023adf3545e8b97a065d22d2ea92

## Specific Changes proposed

Instead of relying on the `timeupdate` event exclusively to update the play progress bar, we set an interval on `play` and clear this interval on `pause` and `ended`.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -15,6 +15,9 @@ import './mouse-time-display.js';
 // The number of seconds the `step*` functions move the timeline.
 const STEP_SECONDS = 5;
 
+// The interval at which the bar should update as it progresses.
+const UPDATE_REFRESH_INTERVAL = 30;
+
 /**
  * Seek bar and container for the progress bars. Uses {@link PlayProgressBar}
  * as its `bar`.
@@ -35,10 +38,30 @@ class SeekBar extends Slider {
   constructor(player, options) {
     super(player, options);
 
-    this.update = Fn.throttle(Fn.bind(this, this.update), 50);
+    this.update = Fn.throttle(Fn.bind(this, this.update), UPDATE_REFRESH_INTERVAL);
+
     this.on(player, 'timeupdate', this.update);
     this.on(player, 'ended', this.handleEnded);
 
+    // when playing, let's ensure we smoothly update the play progress bar
+    // via an interval
+    this.updateInterval = null;
+
+    this.on(player, ['playing'], () => {
+      this.clearInterval(this.updateInterval);
+
+      this.updateInterval = this.setInterval(() =>{
+        this.requestAnimationFrame(() => {
+          this.update();
+        });
+      }, UPDATE_REFRESH_INTERVAL);
+    });
+
+    this.on(player, ['ended', 'pause', 'waiting', 'stalled'], () => {
+      this.clearInterval(this.updateInterval);
+    });
+
+    this.on(player, ['timeupdate', 'ended'], this.update);
   }
 
   /**
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -57,7 +57,7 @@ class SeekBar extends Slider {
       }, UPDATE_REFRESH_INTERVAL);
     });
 
-    this.on(player, ['ended', 'pause', 'waiting', 'stalled'], () => {
+    this.on(player, ['ended', 'pause', 'waiting'], () => {
       this.clearInterval(this.updateInterval);
     });
 


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -34,7 +33,23 @@ class SeekBar extends Slider {
...
+
+    this.on(player, ['play'], () => {
+      this.clearInterval(this.updateInterval);
+
+      this.updateInterval = this.setInterval(() =>{
+        this.update();
+      }, 15);
+    });
+
+    this.on(player, ['ended', 'pause'], () => {
```
misteroneill: This list might not be sufficient. I could potentially see including things like `abort`, `emptied`, `error`, and maybe others. Additionally, I'd be curious to know what happens when including contrib-ads and its event redispatch behavior (such as, would it need to be updated to clear this interval in its own cases).

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -34,7 +33,23 @@ class SeekBar extends Slider {
    */
   constructor(player, options) {
     super(player, options);
-    this.update = Fn.throttle(Fn.bind(this, this.update), 50);
+
+    // when playing, let's ensure we smoothly update the play progress bar
+    // via an interval
+    this.updateInterval = null;
+
+    this.on(player, ['play'], () => {
```
misteroneill: I think this would probably need to be `playing` as opposed to `play`.

src/js/control-bar/progress-control/seek-bar.js
vhmth: Agreed here. Probably more cases I can cover. I'd like to test these plus others that can be found in `Html5.events` (or something like that - I was grep-happy last night and it came up on my radar).

src/js/control-bar/progress-control/seek-bar.js
vhmth: Tying to `playing` will result in the same issue we saw before. We can't tie the update flow to these HTML5 video events since they don't trigger enough to allow for smooth play progress.

vhmth:
@misteroneill agreed on throttling. I can add back throttling the update function to ensure it only gets fired at most once every 15ms. However, we can't throttle it more than that if we want the play progress bar to appear as though it's progressing smoothly (at least if we do it in JS).

Another option here (that I'm not opposed to) is to make use of CSS transitions. It would require a bit more work, and I noticed janky issues when clicking around (we have to do logic where we remove the transition when someone clicks around the `ProgressBar`). I'd like to hear from a couple core contributors on this matter. I can write the code no problem, but there seem to be a few gotchas.

vhmth:
Also, tell the Brightcove peeps Loom says hi @misteroneill! 👋  :-)

mawenmaywin:
My version of VideoJS doesn't seem to have seperate .js files. How can I implement this with video.dev.js file? Are you sure we are talking about the same plugin here?

vhmth:
@mawenmaywin this PR modifies the source code of video.js, which is only present when you clone the repo, not when you install via `npm`. You can clone my repo [here](https://github.com/loomhq/video.js/tree/vinay/smooth_play_progress), checkout the `vinay/smooth_play_progress` branch, and then build the dist files. Then copy the `dist` files to your `node_modules` directory in the application you're testing out locally to see this work.

Alternatively, these changes are pretty small. You can also just modify the `dist` files directly to add this interval and check it yourself.

mawenmaywin:
I'm sorry I am sort of a newb when it comes to plugins and Javascript. I don't understand what you're talking about but I think I will read up on node_modules and npm's.

src/js/control-bar/progress-control/seek-bar.js
heff: `playing` is only fired once when the video actually starts playing after a `play` and shouldn't make any difference in the update speed, so I agree with @misteroneill that it seems more right to use `playing`. Maybe try it out and if you do see a difference I'd be really interested to know why.

src/js/control-bar/progress-control/seek-bar.js
heff: I think the video element is pretty good about firing a pause whenever it's 'not playing', even on ended, but that's worth verifying. Contrib-ads is an interesting question where I have no idea, though this seems like a pretty basic use of play/pause, so if there's a problem with this then contrib-ads probably breaks other things too.

src/js/control-bar/progress-control/seek-bar.js
vhmth: Ah actually you guys are completely correct. It seems like`playing` fires only once. No idea why I saw it fire multiple times. I was probably mistaking this for another event. 

https://stackoverflow.com/a/14270041/696130

Seems like I should definitely check for `waiting` too when clearing the interval.

src/js/control-bar/progress-control/seek-bar.js
vhmth: Not sure what `contrib-ads` is, but should be an easy search against the codebase. Will update y'all here.

vhmth: Seems like switching to `playing` and adding `waiting` and `stalled` is the way to go. I've updated the PR to include those events.

vhmth: https://html.spec.whatwg.org/multipage/media.html#event-media-playing

vhmth: https://stackoverflow.com/a/14270041/696130

vhmth: @misteroneill @heff do either of y'all know of an easy way to test the redispatch behavior with [contrib-ads](https://github.com/videojs/videojs-contrib-ads/blob/master/src/redispatch.js)? It seems like the only events that get special treatment are `playing`, `ended`, `loadstart`, `loadedata` and `loadedmetadata`. Even though the 2 of that list that we're using is `playing` and `ended`, they seem to be pretty straightforward. It would be nice to do a quick sanity check on my end though.

gkatsev: It shouldn't be an issue because the player controls bind the event listeners before contrib-ads's redispatch runs.

vhmth:
Given @gkatsev's update on not needing to test contrib-ads and my updates to the events that this timer binds to, what am I missing here? Looking at the unit tests, it doesn't seem like this really fits into any of them. Happy to add any unit tests y'all think are appropriate though.

On another note, doesn't seem like this should result in any updates to the docs.

I'll be here and dedicated to getting this over the finish line as soon as I get replies. :-)

gkatsev:
I'm still a bit worried about having a setInterval that runs every 15ms. Though, I do agree with you it'll be nice to get this fixed. Maybe a larger interval would be ok? like 50ms or 100ms?
I know some projects use linear interpolation to calculate where the line should be but that's tricky too.

gkatsev:
Yup, I don't think we need any docs updated for this change, thanks for thinking of that :)

vhmth:
@gkatsev 

https://www.useloom.com/share/2c0f11312b4e48de8a7a1ca1d968e50f

vhmth:
@gkatsev just lmk what you wanna go with. I'm cool with bumping it down to `50`.

vhmth:
Hey y'all bump here?

vhmth:
It seems like the main concern might be that the interval is left sitting around when it shouldn't be. Are there any tests you'd like me to write to put everyone's mind at ease? I'd be happy to do that if it helps get this landed faster.

paulius005:
Any updates here?

gkatsev:
Hey, thanks for your patience, we went to [demuxed](http://demuxed.com/) (and I even [spoke there](https://www.youtube.com/watch?v=5AO2ih9-xbI)) and [FOMS](http://www.foms-workshop.org/foms2017/) in early october and that took a lot of energy.

@vhmth thanks for the looms, they were great. Maybe 30ms would be a good compromise? it being 2x of 15 and 30fps is a decent goal :). Thoughts @misteroneill and @heff?

Also, I think bring the throttling back would be good, and of course, it should match the internal we're using. Another thought is to wrap the `update` call in the interval a `requestAnimationFrame`. Do you think that makes sense?

vhmth:
@gkatsev no worries. Dope! Good to put a face to the avatar.

Just to be clear, action items:

1. 30ms timeout
2. Add a throttle back to the update func that's at 30ms as well
3. Call `requestAnimationFrame` within the timeout (seems reasonable)

vhmth:
I can do that very quickly, will wait to hear from others before I update though

misteroneill:
Yeah, that sounds good.

I'm not 100% sure `stalled` is needed here - I'm not sure it means playback has stopped.

heff:
Yeah, `stalled` here is the network, not playback. Otherwise 👍

gkatsev:
@vhmth any updates? :) looks like there's some conflicts now due to this pr (https://github.com/videojs/video.js/pull/4648).

vhmth:
@gkatsev updated and rebased! Still looking good locally and tests are passing still. :-)

gkatsev:
Awesome, thanks @vhmth!

vhmth:
Sweet!

PR #4599, opened 2017-09-04
Add an Audio Description icon to an audio track name in the track menu if it is "main-desc" kind.

OwenEdwards:
## Description
Indicate in the Audio Tracks menu if an audio track contains Audio Description.

## Specific Changes proposed
Add an icon after Audio Description audio tracks in the audio tracks menu, in the same way that the `subs-caps` menu adds an icon after either captions or subtitles tracks.

Note that [videojs-contrib-hls](https://github.com/videojs/videojs-contrib-hls ) added support for [indicating that an audio track contains Audio Description](https://github.com/videojs/videojs-contrib-hls/commit/bfa1fbfe1d5e20d24ebd0f18ffbabe7b5bcea9bd) (by setting its `kind` to `main-desc`) in v5.4.0.

## Requirements Checklist
  - [ ] Docs/guides updated


OwenEdwards:
@gkatsev @mister-ben can you take a look at this - the additional icon in the audio menu is causing the menu to have scrollbars or the icon gets clipped:



I can't figure out how to avoid that. Any thoughts?

mister-ben:
I haven't dug into this yet but this is a problem for the subs/caps menu too - it's just the labels are generally short enough that I hadn't noticed.

gkatsev:
We probably want to truncate the text if there's an icon and set the `title` attribute to the fulltext, so, visually it'll be truncated, but will be available to screen readers or on hover.

OwenEdwards:
Another alternative would be to move Audio Description text tracks into a menu which contains those tracks *and* subtitles and captions, like Safari does:

prozak915:
Thx guys iam relally aproshieitet fot your help u save my life a havenr sklep for 2 data i on tnie phone its really słów to do everything

misteroneill:
Was there ever a solution for the overflow issue?

OwenEdwards:
@misteroneill yes, the "Change CSS based on PR #5228 to fix menu layout with icons" pulls in the solution from that PR which means the AD icon wraps to the next line if necessary.

PR #4607, opened 2017-09-11
fix: issue where changing focus is causing unwanted scrolling behavior on iOS

alex-barstow:
## Description
Changing focus to a sub-menu (ex. by tapping the captions menu button) when a player is embedded in an iframe is resulting in unwanted scrolling behavior on iOS. @gkatsev is planning to refactor all focus handling soon, but this will serve as a temporary fix in the meantime. 

## Specific Additions
Modify the MenuButton `pressButton()` method to only change focus to a sub-menu on platforms other than iOS

### DIFF
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -343,7 +343,7 @@ class MenuButton extends Component {
 
       // set the focus into the submenu, except on iOS where it is resulting in
       // undesired scrolling behavior when the player is in an iframe
-      if (!IS_IOS) {
+      if (!IS_IOS && !Dom.isPlayerInFrame()) {
         this.menu.focus();
       }
     }
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -343,7 +343,7 @@ class MenuButton extends Component {
 
       // set the focus into the submenu, except on iOS where it is resulting in
       // undesired scrolling behavior when the player is in an iframe
-      if (!IS_IOS && !Dom.isPlayerInFrame()) {
+      if (!IS_IOS && !Dom.isInFrame()) {
         this.menu.focus();
       }
     }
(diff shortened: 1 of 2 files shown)


src/js/menu/menu-button.js
```diff
@@ -8,6 +8,7 @@ import * as Dom from '../utils/dom.js';
 import * as Fn from '../utils/fn.js';
 import * as Events from '../utils/events.js';
 import toTitleCase from '../utils/to-title-case.js';
+import IS_IOS from '../utils/browser.js';
```
gkatsev: `IS_IOS` needs to be wrapped in curly braces when you want a named import
```js
import { IS_IOS } from '../utils/browser.js';
```

gkatsev: This is also why the build is failing.

alex-barstow: _facepalm_. Fixed.

priyankapaladi:
I checked the fix on iOS 10.2.1 (iPad), using master branch of socrates and fix-menu-focus-ios branch of video.js. It works fine, I don't see it scroll abruptly anymore.
QA-LGTM

gkatsev:
@alex-barstow so, since this only happens when the player is embedded in an iframe, maybe we should add that to the check? `if (ios && in-iframe) { dontFocus() }`. Thoughts?

alex-barstow:
@gkatsev Changes made! I added the iframe check as a new utils function in dom.js

src/js/utils/dom.js
```diff
@@ -86,6 +86,23 @@ export function isEl(value) {
 }
 
 /**
+ * Determines if the player is embedded in an iframe.
+ *
+ * @return {boolean}
+ *
+ */
+export function isPlayerInFrame() {
```
misteroneill: I propose changing this to `isInIframe` more generally. Might be pedantic or whatever, but that seems like a better name.

Issue #4611, opened 2017-09-12
Disable right click seek videojs


kocoten1992:
I want to disable right-click seek video, https://github.com/videojs/video.js/issues/4120 doesn't help disable right click seek

gkatsev:
How do you right-click to seek? But we might want to update our code that handles seek to make sure that we only consider left clicks.

kocoten1992:
I'm using v5.20, doesn't do anything special, it was right clickable seek by default

gkatsev:
Oh, when I tried it earlier it didn't work but was able to get it to work after trying again.
Yup, sounds like we need to update our slider component to only work with the left click rather than any click event. Would you be interested and able to work on this, @kocoten1992?

kocoten1992:
ye, I'll fix it, my project need that too

kocoten1992:
Fixed :+1:

asd29696:
Chuong <notifications@github.com> 於 2017年11月17日 週五 15:36 寫道：

PR #4613, opened 2017-09-13
Fix only left click draggable bar

kocoten1992:
hi @gkatsev , this can't be fix from slider-component alone, some logic was on seek-bar and volume-bar, best preventing from there. #4611 

I'm not familiar with git, do I need to do this again for v5.x ?


kocoten1992:
don't merge that yet, there are potential bug with with volume control with that code

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -113,6 +113,11 @@ class SeekBar extends Slider {
    * @listens mousedown
    */
   handleMouseDown(event) {
+    // only allow left-click seeking
+    if (event.buttons !== 1) {
```
gkatsev: shouldn't this be [`0`](https://w3c.github.io/uievents/#dom-mouseevent-button)? `1` is the auxiliary button.

kocoten1992: for some reason ```button``` not work, then I tried ```buttons``` and it work, then there is funny issue with touchscreen, I'll do a careful test and might delete this pull, it getting messy

kocoten1992:
Will test on real device tomorrow

kocoten1992:
@gkatsev , thank you :+1: 

Allowed me to have few more days, need to be test on blackberry and android, and there is a very subtle bug (may not need to be fix), will report in few days

kocoten1992:
It work on blackberry, android, and ios, think I nailed it.

```css
.video-js .vjs-slider:focus {
    text-shadow: 0em 0em 1em white;
    -webkit-box-shadow: 0 0 1em #fff;
    -moz-box-shadow: 0 0 1em #fff;
    box-shadow: 0 0 1em #fff;
}
```

Previously vjs-slider never have a chance to show up, but now it does, it pop a white box-shadow, I'll leave this for vjs team.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -113,6 +113,15 @@ class SeekBar extends Slider {
...
    */
   handleMouseDown(event) {
+    // only allow left-click seeking
+    if (typeof event.button === 'undefined' && typeof event.buttons === 'undefined') {
+      // touch screen, not apply, do nothing
+    } else if (event.button === 0 && typeof event.buttons === 'undefined') {
+      // touch screen, safari on ios
+    } else if (event.button !== 0 || event.buttons !== 1) {
+      return;
+    }
```
misteroneill: I find this series of `if` blocks confusing. Can it be simplified to:

```js
if ((event.button !== undefined && event.button !== 0) || (event.buttons !== undefined && event.buttons !== 1)) {
  return;
}
```

Does that not work?

kocoten1992: I don't know :/, what I did was trial by combat, didn't have device to test right now, think I will stick with what work, maybe change to that later if I have device to test

kocoten1992: Beside, I think if we do the current style, we have a chance to handle/explain what is going on with particular device, OS or stuff

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -113,6 +113,15 @@ class SeekBar extends Slider {
...
    */
   handleMouseDown(event) {
+    // only allow left-click seeking
+    if (event.button === undefined && event.buttons === undefined) {
+      // touch screen, not apply, do nothing
+    } else if (event.button === 0 && event.buttons === undefined) {
+      // touch screen, safari on ios
+    } else if (event.button !== 0 || event.buttons !== 1) {
+      return;
+    }
```
gkatsev: You could potentially do something like:
```js
  // only allow left-click seeking
if (
  // touch screen, not apply, do nothing
  !(event.button === undefined && event.buttons === undefined) &&
  // touch screen, safari on ios
  !(event.button === 0 && event.buttons === undefined) &&
  (event.button !== 0 || event.buttons !== 1)
) {
 return;
}
```

Also, we probably should pull this out into a tester function because keeping this copied between the two places is bound to cause issues down the line. Maybe moved to dom.js as a `isNotLeftClick(event)` or something like that and can be used something like so:
```js
if (isNotLeftClick(event)) {
  return;
}
```

kocoten1992:
I'm continue this, will
- [ ] rermove my old code in volume and seek bar
- [ ] create isNotLeftClick in dom.js
- [ ] use isNotLeftClick in both handleMouseDown and handleMouseMove of both volume and seek bar

src/js/utils/dom.js
```diff
@@ -728,6 +728,52 @@ export function insertContent(el, content) {
...
+ *
+ * @param {EventTarget~Event} event
+ *        Event object
+ *
+ * @return {boolean}
+ *         - True if a left click
+ *         - False if not a left click
+ */
+export function isSingleLeftClick(event) {
+  /* Note: if you create something draggable, be sure to
```
misteroneill: Minor stylistic concern, but could we stick to using `/* */` comments for JSDoc only and use `//` for all inline comments (even if they span multiple lines)?

PR #4614, opened 2017-09-14
fix: reset to a play/pause button when seeking after ended

BrandonOCasey:
## Description
When seeking backwards after the video has ended, the `vjs-ended` class will remain on the video until play is pushed. This means the the play button will still show as a replay button, which seems incorrect to me.

## Specific Changes proposed
This PR removes `vjs-ended` on any `seeked` event on the player and resets the button into a play/pause state depending upon the player state.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -74,21 +74,11 @@ class PlayToggle extends Button {
       return;
     }
 
-    // if it has set us back to a play/pause button
-    // depending on the current state
+    // remove the ended class
     this.removeClass('vjs-ended');
 
-    if (this.player_.paused()) {
-      if (!this.hasClass('vjs-paused')) {
-        this.addClass('vjs-paused');
-      }
-      this.controlText('Pause');
-    } else {
-      if (!this.hasClass('vjs-playing')) {
-        this.addClass('vjs-playing');
-      }
-      this.controlText('Play');
-    }
+    // and set us back to showing the play button
+    this.handlePause(event);
   }
 
   /**
@@ -102,7 +92,10 @@ class PlayToggle extends Button {
   handlePlay(event) {
     this.removeClass('vjs-ended');
     this.removeClass('vjs-paused');
-    this.addClass('vjs-playing');
+
+    if (!this.hasClass('vjs-playing')) {
+      this.addClass('vjs-playing');
+    }
     // change the button text to "Pause"
     this.controlText('Pause');
   }
@@ -117,7 +110,10 @@ class PlayToggle extends Button {
    */
   handlePause(event) {
     this.removeClass('vjs-playing');
-    this.addClass('vjs-paused');
+
+    if (!this.hasClass('vjs-paused')) {
+      this.addClass('vjs-paused');
+    }
     // change the button text to "Play"
     this.controlText('Play');
   }
@@ -128,7 +124,10 @@ class PlayToggle extends Button {
    */
   handleEnded(event) {
     this.removeClass('vjs-playing');
-    this.addClass('vjs-ended');
+
+    if (!this.hasClass('vjs-ended')) {
+      this.addClass('vjs-ended');
+    }
     // change the button text to "Replay"
     this.controlText('Replay');
   }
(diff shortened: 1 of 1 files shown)


src/js/control-bar/play-toggle.js
```diff
@@ -58,6 +59,39 @@ class PlayToggle extends Button {
...
+    // check to see if the video has ended yet
+    if (!this.hasClass('vjs-ended')) {
+      return;
+    }
+
+    // if it has set us back to a play/pause button
+    // depending on the current state
+    this.removeClass('vjs-ended');
+
+    if (this.player_.paused()) {
```
gkatsev: if we're ended, we should already be paused, so, I think we can probably simply this. Maybe we can just call the `handlePause` handler so we don't duplicate work?

BrandonOCasey: I was thinking about it, but it seemed a bit weird to be calling a function made for being called by events manually. If you don't see a problem with it. I can go ahead and do it though.

gkatsev: probably ok? Can always pull it out into a separate function that both methods can call.

BrandonOCasey:
@gkatsev updated

src/js/control-bar/play-toggle.js
```diff
@@ -58,6 +59,29 @@ class PlayToggle extends Button {
...
+   * of a video
+   *
+   * @param {EventTarget~Event} [event]
+   *        The event that caused this function to run.
+   *
+   * @listens Player#seeked
+   */
+  handleSeeked(event) {
+    // check to see if the video has ended yet
+    if (!this.hasClass('vjs-ended')) {
```
gkatsev: oh, we probably should use `this.ended()` instead of checking the class.

BrandonOCasey: Yeah I guess that will work, should be `this.player_.ended()` right?

gkatsev: oops, yes, it's on the player

src/js/control-bar/play-toggle.js
```diff
@@ -58,6 +58,26 @@ class PlayToggle extends Button {
...
+   * @param {EventTarget~Event} [event]
+   *        The event that caused this function to run.
+   *
+   * @listens Player#seeked
+   */
+  handleSeeked(event) {
+    // remove the ended class
+    this.removeClass('vjs-ended');
+
+    if (this.player_.paused()) {
```
gkatsev: why would we not be paused when we're ended?

src/js/control-bar/play-toggle.js
```diff
@@ -66,9 +86,11 @@ class PlayToggle extends Button {
    * @listens Player#play
    */
   handlePlay(event) {
-    this.removeClass('vjs-ended');
     this.removeClass('vjs-paused');
-    this.addClass('vjs-playing');
+
+    if (!this.hasClass('vjs-playing')) {
```
gkatsev: why not just keep this as unconditional?

src/js/control-bar/play-toggle.js
```diff
@@ -83,20 +105,33 @@ class PlayToggle extends Button {
...
+   * @param {EventTarget~Event} [event]
+   *        The event that caused this function to run.
+   *
+   * @listens Player#ended
    */
   handleEnded(event) {
     this.removeClass('vjs-playing');
-    this.addClass('vjs-ended');
+
+    if (!this.hasClass('vjs-ended')) {
```
gkatsev: why not just keep this as unconditional?

src/js/control-bar/play-toggle.js
```diff
@@ -83,20 +105,33 @@ class PlayToggle extends Button {
    */
   handlePause(event) {
     this.removeClass('vjs-playing');
-    this.addClass('vjs-paused');
+
+    if (!this.hasClass('vjs-paused')) {
```
gkatsev: why not just keep this as unconditional?

src/js/control-bar/play-toggle.js
BrandonOCasey: When you click the button, it will cause a play and then a seek, which would cause the play button to show, even though the video is playing.

src/js/control-bar/play-toggle.js
BrandonOCasey: moved them back to unconditional as this is already a check in addClass.

src/js/control-bar/play-toggle.js
gkatsev: Ah, I see.

PR #4620, opened 2017-09-15
feat: add remainingTimeDisplay method

### LINKED ISSUE #4608: Rounding errors with player.remainingTime()
## Description
Demo of the problem here: [http://jsbin.com/caderis/1/edit?js,output](http://jsbin.com/caderis/1/edit?js,output)

It seems the best solution would be to make `player.remainingTime` take an argument for the number of significant figures, or even a rounding function to use. Then it can properly round the remainingTime to be consistent with other time labels. This is important any time you're displaying remainingTime, as there's currently no way to use `remainingTime` anywhere in our app and have it be consistent with the other (pre-rounded) time labels.

## Steps to reproduce
1. Display the `.vjs-time-control` component
2. Note that currentTime + remainingTime !== duration (visually)

### Expected
I expect time labels to appear consistent to the user.

### Actual
Rounding errors cause time labels to be inconsistent at times.

#### videojs
6.2.5

Happy to help and submit a PR if you can give me some direction as to how you'd like to address this. Something like this would probably be most versatile:

```
  /**
   * Calculates how much time is left in the video. Not part
   * of the native video API.
   *
   * @param {function}
   *        - an optional rounding function
   *
   * @return {number}
   *         The time remaining in seconds
   */
  remainingTime(roundFn) {
    const round = typeof roundFn === "function" ? roundFn : (x) => x; // noop (default)

    return round(this.duration()) - round(this.currentTime());
  }
```

```
// now if we need the remaining time to display visually, we can do this:
const remainingTime = player.remainingTime(Math.floor);
```

But simply fixing the value shown in the `vjs-remaining-time-display` component would be an improvement.

BrandonOCasey:
## Description
Allow remainingTime to be given a rounding function during use. Add a function which should be used as the default when displaying remaining time.

Fixes #4608 

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/time-controls/remaining-time-display.js
+++ b/src/js/control-bar/time-controls/remaining-time-display.js
@@ -49,7 +49,7 @@ class RemainingTimeDisplay extends TimeDisplay {
       return;
     }
 
-    this.updateFormattedTime_(this.player_.remainingTime());
+    this.updateFormattedTime_(this.player_.remainingTimeDisplay());
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1787,6 +1787,17 @@ class Player extends Component {
     return this.duration() - this.currentTime();
   }
 
+  /**
+   * A remaining time function that is intented to be used when
+   * the time is to be displayed directly to the user.
+   *
+   * @return {number}
+   *         The rounded time remaining in seconds
+   */
+  remainingTimeDisplay() {
+    return Math.floor(this.duration()) - Math.floor(this.currentTime());
+  }
+
   //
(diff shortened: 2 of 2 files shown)


gkatsev:
We'd want to update the remainingTimeDisplay to pass in Math.floor (or something else) as well.

gkatsev:
Also, I'm still not sure this is the direction we want as it requires the user to know how format time works to know what to pass in if the user wants to use it for displaying themselves.

BrandonOCasey:
After some more exhaustive testing http://jsbin.com/nohawin/1/edit?js,output it seems that floor/floor is the only possible way for this to work.

PR #4622, opened 2017-09-17
fix typo in plugins guide

thijstriemstra:
Based on the docs it should be `this` instead of `player`.

vjsstandard also found issue in my plugin:

```
/home/foo/projects/videojs-wavesurfer/src/videojs.wavesurfer.js:30:30: "player" is not defined.
```

### DIFF
--- a/docs/guides/plugins.md
+++ b/docs/guides/plugins.md
@@ -251,7 +251,7 @@ class Advanced extends Plugin {
   }
 
   updateState() {
-    this.setState({playing: !player.paused()});
+    this.setState({playing: !this.paused()});
   }
 
   logState(changed) {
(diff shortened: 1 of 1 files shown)


docs/guides/plugins.md
```diff
@@ -251,7 +251,7 @@ class Advanced extends Plugin {
   }
 
   updateState() {
-    this.setState({playing: !player.paused()});
+    this.setState({playing: !this.paused()});
```
gkatsev: this probably needs to be `this.player().paused()`

thijstriemstra: according to docs (and my experience) it's `player`:

> With advanced plugins, the value of `this` is the _instance of the plugin class_. The player is passed to the plugin constructor as its first argument (and is automatically applied to the plugin instance as the `player` property) and any further arguments are passed after that.

gkatsev: Yup, it's a property, was thinking of regular components.

PR #4635, opened 2017-09-28
fix: don't throttle duration change updates

BrandonOCasey:
## Description
Right now the `durationchange` event is throttled with the other two events, `timeupdate` and `loadedmetadata`. This means that only one of those events will trigger an update if they all occur within 25ms of each other. 

This functionality makes sense for `timeupdate` or `loadedmetadata` as those should not indicate a time update (even though they often do). 

For `durationchange` however it will always indicate a change in the duration, and we want to always update the display when it happens. Here is a scenario of how we could show duration incorrectly right now:

User is playing a video that has a postroll ad at the end. After the postroll ad their will be a `timeupdate`, and then a `durationchange` to signify that we are back in content. Then `ended` will fire, and no more events will happen.

The player will still show the duration of the ad, as the `durationchange` was eaten by the `timeupdate` that happened at nearly the same time.

## Requirements Checklist

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1239,7 +1239,9 @@ class Component {
     fn = Fn.bind(this, fn);
 
     const timeoutId = window.setTimeout(fn, timeout);
-    const disposeFn = () => this.clearTimeout(timeoutId);
+    const disposeFn = function() {
+      this.clearTimeout(timeoutId);
+    };
 
     disposeFn.guid = `vjs-timeout-${timeoutId}`;
 
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1239,9 +1239,7 @@ class Component {
     fn = Fn.bind(this, fn);
 
     const timeoutId = window.setTimeout(fn, timeout);
-    const disposeFn = function() {
-      this.clearTimeout(timeoutId);
-    };
+    const disposeFn = () => this.clearTimeout(timeoutId);
 
     disposeFn.guid = `vjs-timeout-${timeoutId}`;
 
@@ -1302,9 +1300,7 @@ class Component {
 
     const intervalId = window.setInterval(fn, interval);
 
-    const disposeFn = function() {
-      this.clearInterval(intervalId);
-    };
+    const disposeFn = () => this.clearInterval(intervalId);
 
     disposeFn.guid = `vjs-interval-${intervalId}`;
 
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1239,9 +1239,7 @@ class Component {
     fn = Fn.bind(this, fn);
 
     const timeoutId = window.setTimeout(fn, timeout);
-    const disposeFn = function() {
-      this.clearTimeout(timeoutId);
-    };
+    const disposeFn = () => this.clearTimeout(timeoutId);
 
     disposeFn.guid = `vjs-timeout-${timeoutId}`;
 
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -48,7 +48,7 @@ class TimeDisplay extends Component {
       'aria-live': 'off'
     }, Dom.createEl('span', {
       className: 'vjs-control-text',
-      textContent: this.localize(this.contentText_)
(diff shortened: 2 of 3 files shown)


src/js/control-bar/time-controls/time-display.js
```diff
@@ -48,7 +48,7 @@ class TimeDisplay extends Component {
       'aria-live': 'off'
     }, Dom.createEl('span', {
       className: 'vjs-control-text',
-      textContent: this.localize(this.contentText_)
+      textContent: this.localize(this.controlText_)
```
BrandonOCasey: this was a typo, I could arguable pull this into its own pr if we want, but this is one of the causes of the broken unit tests.

src/js/control-bar/time-controls/time-display.js
```diff
@@ -63,7 +63,7 @@ class TimeDisplay extends Component {
    * @private
    */
   updateTextNode_() {
-    if (this.textNode_) {
+    if (this.textNode_ && this.contentEl_ && this.contentEl_.contains(this.textNode_)) {
       this.contentEl_.removeChild(this.textNode_);
```
BrandonOCasey: IE11 throws in request animation frames when trying to remove a DOM node from a container when the container does not have the node. Might only be an issue for our unit tests, but I fixed it here.

src/js/control-bar/time-controls/time-display.js
gkatsev: This is what's causing the issue on ie8? https://travis-ci.org/videojs/video.js/jobs/290639178#L1937-L1941

BrandonOCasey: im actually not sure what is causing it

src/js/component.js
```diff
@@ -1239,9 +1239,7 @@ class Component {
     fn = Fn.bind(this, fn);
 
     const timeoutId = window.setTimeout(fn, timeout);
-    const disposeFn = function() {
-      this.clearTimeout(timeoutId);
-    };
+    const disposeFn = () => this.clearTimeout(timeoutId);
```
BrandonOCasey: This seemed a bit weird to me in a regular function we wouldn't have access to `this.clearTimeout`

src/js/component.js
```diff
@@ -1302,9 +1300,7 @@ class Component {
 
     const intervalId = window.setInterval(fn, interval);
 
-    const disposeFn = function() {
-      this.clearInterval(intervalId);
-    };
+    const disposeFn = () => this.clearInterval(intervalId);
```
BrandonOCasey: fixed here as well

src/js/component.js
gkatsev: we call the `dispose` handler with context as the component instance, so, it should have it.

gkatsev: at least, it should be. this is ok too.

src/js/control-bar/time-controls/time-display.js
```diff
@@ -63,7 +63,12 @@ class TimeDisplay extends Component {
    * @private
    */
   updateTextNode_() {
-    if (this.textNode_) {
+    if (!this.contentEl_) {
+      return;
+    }
+
+    // ie8/9 don't support `el.contains` inside of setTimeout call.
```
gkatsev: weird

src/js/control-bar/time-controls/time-display.js
```diff
@@ -63,7 +63,12 @@ class TimeDisplay extends Component {
...
+    if (this.textNode_ && this.textNode_.parentNode === this.contentEl_) {
```
gkatsev: what about just removing everything inside of `contentEl_`? It'll also fix an issue where if the page is translated in chrome using google translate, the text node will be wrapped in another element and this method will fail.
This could also just happen in a separate PR.

src/js/component.js
```diff
@@ -1239,9 +1239,7 @@ class Component {
     fn = Fn.bind(this, fn);
 
     const timeoutId = window.setTimeout(fn, timeout);
```
gkatsev: I almost commented saying that this should be changed to `this.setTimeout` before realizing what this line is for 😆

PR #4639, opened 2017-10-02
Fix safari play() after source change

alex-barstow:
## Description
Programmatically calling `player.play()` immediately following a source change will fail in Safari (desktop and iOS) if the video element's `preload` attribute is set to `none`.

Open the following links in Safari, start playback, then press the "Play new source" button. These samples demonstrate the problem using a vanilla `<video>` element.
- [play() FAILS following source change](http://jsbin.com/qiyoqiluhi/1/edit?html,output)
- [play() SUCCEEDS after calling load() following source change](http://jsbin.com/jokikadeke/1/edit?html,output)

**Expected behavior:** the source changes and starts to play.
**Actual behavior:**  playback stalls

One observed effect of this problem is with the implementation of playlists. With `preload: "none"`, playlists, whether configured to autoadvance or click-to-play, do not work in Safari because playback stalls when switching sources.

**Note:** This bug has apparently been fixed in Safari 11.

## Specific additions
Modify the `player.play()` method so that if it is called while the content source is still changing in Safari, `load()` will be called before `play()`.

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -85,6 +85,14 @@ export const IE_VERSION = (function() {
 
 export const IS_SAFARI = (/Safari/i).test(USER_AGENT) && !IS_CHROME && !IS_ANDROID && !IS_EDGE;
 export const IS_ANY_SAFARI = IS_SAFARI || IS_IOS;
+export const DESKTOP_SAFARI_VERSION = (function() {
+  const match = USER_AGENT.match(/Version\/(\d+)/);
+
+  if (match && match[1] && IS_SAFARI && !IS_IOS) {
+    return parseFloat(match[1]);
+  }
+  return null;
+}());
 
 export const TOUCH_ENABLED = Dom.isReal() && (
   'ontouchstart' in window ||
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1624,8 +1624,8 @@ class Player extends Component {
       this.ready(function() {
 
         // Safari struggles to play immediately following a source change
-        // without first calling load()
-        if (browser.IS_ANY_SAFARI) {
+        // without first calling load() if preload="none", except in desktop Safari 11
+        if (this.preload() === 'none' && (browser.IS_IOS || browser.DESKTOP_SAFARI_VERSION < 11)) {
           this.techCall_('load');
         }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1622,6 +1622,13 @@ class Player extends Component {
   play() {
     if (this.changingSrc_) {
       this.ready(function() {
+
+        // Safari struggles to play immediately following a source change
+        // without first calling load()
+        if (browser.IS_ANY_SAFARI) {
+          this.techCall_('load');
+        }
+
         const retval = this.techGet_('play');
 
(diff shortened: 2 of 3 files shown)


serv:
I have a similar problem that's happening with Safari 11 for video.js `v5.20.3`.

In Safari 11 > Preferences > Websites, I have the website set to `Never Auto-Play`.

screen shot 2017-10-03 at 4 06 01 pm

When I try to play any content, the videojs pauses the content. I can check this by seeing that videojs triggers `pause` event.

Do you think this is the same issue? Or should I file another issue.?

gkatsev:
@serv that's another unrelated issue that's already been addressed as much as we can in video.js. See https://github.com/videojs/video.js/pull/4562 and https://github.com/videojs/video.js/pull/4582. Also, https://www.brightcove.com/en/blog/2017/09/autoplay is a good overview, I think of the issues.

serv:
Sorry to resurface this issue one more time.

After reading, https://webkit.org/blog/6784/new-video-policies-for-ios/, the problem I am having is coming from this.

`If a <video> element gains an audio track or becomes un-muted without a user gesture, playback will pause.`

Wouldn't clicking a button, that triggers `.play()` be sufficient for a `user gesture`? 

I think there's definitely something funky happening with video.js still because jumping to a middle of the track using `.currentTime()` plays the track successfully, while simply `.play()` a track immediately triggers a pause.

alex-barstow:
@misteroneill Yup, minimizing the scope makes sense. I'll make the changes.

src/js/utils/browser.js
```diff
@@ -85,6 +85,14 @@ export const IE_VERSION = (function() {
 
 export const IS_SAFARI = (/Safari/i).test(USER_AGENT) && !IS_CHROME && !IS_ANDROID && !IS_EDGE;
 export const IS_ANY_SAFARI = IS_SAFARI || IS_IOS;
+export const DESKTOP_SAFARI_VERSION = (function() {
```
gkatsev: Maybe this should just be `SAFARI_VERSION`? on iOS it'll be equivalent to `IOS_VERSION`?
I.e., if you return `IOS_VERSION` instead of `null` from this iife.

gkatsev:
Looks like tests are still failing.

src/js/utils/browser.js
alex-barstow: Personally I would find it clearer to have `IOS_VERSION` solely responsible for mobile and `DESKTOP_SAFARI_VERSION` solely responsible for desktop, rather than a multi-purpose `SAFARI_VERSION`, but that's just me. Happy to make this change if you want me to.

gkatsev: @videojs/core-committers any thoughts?

ldayananda: I think having a separate Desktop Safari version would be more clear.

BrandonOCasey: Yeah I think having desktop is better as well

alex-barstow:
[Recent changes](https://github.com/videojs/video.js/pull/4743) to Player#play have addressed the underlying issue here, so this PR is no longer necessary. Closing.

PR #4640, opened 2017-10-03
feat: allow embeds via <video-js> element

gkatsev:
## Description
Add the ability to initialize Video.js with an element named `video-js`. This is because sometimes, seeing the native element is undesirable, plus, it's cool to have our own element.
Can be used just like the `video` embed. For IE9-only we have a `vjs-source` element that's copied to a `source` element because IE9 is strict about keeping `source` elements inside of `video` elements.

## TODO
* [x] tests need to be updated
* [x] fix tests in IE8
* [x] Limit this feature to IE10+ so we won't need a `vjs-source` element?
* [x] documentation
* [x] we should potentially update `Player.getTagSettings` instead of my workaround for sources to work properly with `vjs-source`.
* [x] We may need a `vjs-track` as well for IE9.

### DIFF
--- a/docs/guides/embeds.md
+++ b/docs/guides/embeds.md
@@ -1,9 +1,9 @@
 # How to Embed the Video.js player
 
-Video.js is meant to be an ehnacement to the video element in HTML5 so for years, it's embed code has been just a `<video>` element.
+Video.js is meant to be an enhacement to the video element in HTML5 so for years, its embed code has been just a `<video>` element.
 Video.js then wraps the video element in a div that is used for us to place controls and anything else that's required for the player.
 
-For a long time this was enough. In 2016, a "div ingest" which allows the developer to give Video.js a player div to use instead of making it's own.
+For a long time this was enough. In 2016, "div ingest" was added, it allows the developer to give Video.js a player div to use instead of making it's own.
 This is partly to help with content reflow but also to help with iOS where you sometimes need to prime the video element and we re-create the video element when we create the player div.
 However, this is kind of weird to have a `<video>` element embed with a `<div>` wrapped around it. So, we built out a new embed, a `<video-js>` embed.
 
@@ -81,5 +81,5 @@ These all work in all browsers that Video.js supports, though, there are some ca
 - IE9 doesn't support having `source` elements outside of the `video` element, thus, the `video-js` embed will not work there. Though, if the source is set later, it should still work.
 
 ## data-setup
-This is an ease-of-use method for having Video.js set up the player automatically. It is an html attribute and it takes a JSON string representation of the [player options](/docs/guides/options.md) as the value.
+This is an ease-of-use method for having Video.js set up the player automatically. It is an HTML attribute and it takes a JSON string representation of the [player options](/docs/guides/options.md) as the value.
 Using the programmatic approach is probably preferable.
--- a/docs/guides/embeds.md
+++ b/docs/guides/embeds.md
@@ -59,18 +59,20 @@ This is useful for all the things that the player div ingest is useful for and i
 
 ```html
 <!-- via data-setup -->
-<video-js id="vid1" class="video-js" data-setup='{}'>
+<video-js id="vid1" data-setup='{}'>
   <source src="//vjs.zencdn.net/v/oceans.mp4">
 </video-js>
 
 <!-- via code -->
-<video-js id="vid1" class="video-js">
+<video-js id="vid1">
   <source src="//vjs.zencdn.net/v/oceans.mp4">
 </video-js>
 ```
 ```js
 const player = videojs('vid1', {});
 ```
+
+Adding `class="video-js"` with this embed is no longer necessary as it will automatically add the class `video-js` if missing.
 
 #### Custom Elements
 Native Custom Elements support is relativly small according to [Can I Use](http://caniuse.com/#feat=custom-elementsv1) and because we didn't want to include a polyfill we're going with just an element called `video-js` rather than a full blown custom element.
--- a/docs/guides/embeds.md
+++ b/docs/guides/embeds.md
@@ -1,0 +1,83 @@
+# How to Embed the Video.js player
+
+Video.js is meant to be an ehnacement to the video element in HTML5 so for years, it's embed code has been just a `<video>` element.
+Video.js then wraps the video element in a div that is used for us to place controls and anything else that's required for the player.
+
+For a long time this was enough. In 2016, a "div ingest" which allows the developer to give Video.js a player div to use instead of making it's own.
+This is partly to help with content reflow but also to help with iOS where you sometimes need to prime the video element and we re-create the video element when we create the player div.
+However, this is kind of weird to have a `<video>` element embed with a `<div>` wrapped around it. So, we built out a new embed, a `<video-js>` embed.
+
+Below, the three kinds of embeds are detailed.
+
+## Embeds
+### `<video>` embed
+The classic Video.js embed. You can then initialize it via `data-setup` or via the `videojs` method.
+
+```html
+<!-- via data-setup -->
+<video id="vid1" class="video-js" data-setup='{}'>
+  <source src="//vjs.zencdn.net/v/oceans.mp4">
+</video>
+
+<!-- via code -->
+<video id="vid1" class="video-js">
+  <source src="//vjs.zencdn.net/v/oceans.mp4">
+</video>
+```
+```js
+const player = videojs('vid1', {});
+```
+
+### Player div ingest
+The enhanced classic embed. You can also initialize it via `data-setup` or via `videojs` method.
+
+```html
+<!-- via data-setup -->
+<div data-vjs-player>
+  <video id="vid1" class="video-js" data-setup='{}'>
+    <source src="//vjs.zencdn.net/v/oceans.mp4">
+  </video>
+</div>
+
+<!-- via code -->
+<div data-vjs-player>
+  <video id="vid1" class="video-js">
+    <source src="//vjs.zencdn.net/v/oceans.mp4">
+  </video>
+</div>
+```
+```js
+const player = videojs('vid1', {});
+```
+
+As you can see, it isn't much different from the classic `<video>` embed. It also does make it easier to work with [React](/docs/guides/react.md).
+
+### `<video-js>` embed
+This is the [I Can't Believe It's Not Custom Elements](https://developers.google.com/web/fundamentals/web-components/customelements) embed code.
+It looks very much like the `<video>` embed but instead of `video` it's a `video-js` embed.
+This is useful for all the things that the player div ingest is useful for and it matches our library name!
+
+```html
+<!-- via data-setup -->
+<video-js id="vid1" class="video-js" data-setup='{}'>
+  <source src="//vjs.zencdn.net/v/oceans.mp4">
+</video-js>
+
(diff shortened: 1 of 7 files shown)


src/js/setup.js
```diff
@@ -31,6 +31,7 @@ const autoSetup = function() {
   // through each list of elements to build up a new, combined list of elements.
   const vids = document.getElementsByTagName('video');
   const audios = document.getElementsByTagName('audio');
+  const divs = document.getElementsByTagName('div');
```
gkatsev: this allows for using `data-setup` with divs. Though, this seems a bit generic, should we use a different trigger with divs?

sandbox/embeds.html.example
```diff
@@ -0,0 +1,165 @@
+<!DOCTYPE html>
```
gkatsev: this file will get autocopied into `sandbox/embeds.html` on the first `npm start` run by the user.
It has 18 different video.js embeds grouped by whether data-setup is used or not, how the source is added, and the 3 different embed types: video, video+player div wrapper, just div.

src/js/player.js
```diff
@@ -508,34 +509,56 @@ class Player extends Component {
...
       el = this.el_ = super.createEl('div');
     }
 
+    // Copy over all the attributes from the tag, including ID and class
+    // ID will now reference player box, not the video tag
+    const attrs = Dom.getAttributes(tag);
+
+    if (divEmbed) {
+      el = this.el_ = tag;
+      tag = this.tag = document.createElement('video');
```
gkatsev: we probably should try to be smarted about this. Maybe we need an audio element instead but we might not have a source at this point. Probably more than good enough for first release.

gkatsev:
So, other than adding *a lot* of tests, we should decide if we're ok with using `data-setup` on the div elements. Otherwise, this is pretty much good to go, I think.

sandbox/embeds.html.example
```diff
@@ -0,0 +1,165 @@
...
+    .source-js.data-setup { background: green; }
+    .options-src.data-setup { background: blue; }
+
+    .video-js {
+      height: 150px;
+      width: 300px;
+    }
+
+    .wrapper {
+      display: grid;
```
gkatsev: yeah, grid!
screen shot 2017-10-12 at 16 07 17

gkatsev:
In a grooming discussion, we talked about the `div` and `data-setup` issue. Because `div` is a lot more used and `data-setup` is a relatively generic name, we are at risk of trying to initialize a div that was not given to video.js. One suggestion that was brought up is to rename `data-setup` to be `data-vjs-setup` and keep `data-setup` for video/audio elements and perhaps deprecate it altogether.
What are your thoughts @videojs/core-committers?

heff:
Is the div embed meant to take over as the primary embed method or is it an optional feature for React-like situations?

`data-setup` is definitely a convenience thing. It's specifically useful in situations where a user doesn't have access to add javascript but can add HTML (some CMSs). Other than that, if this is an optional feature I'd be in favor of dropping data-setup for the div and see if anyone misses it. Otherwise `data-vjs-setup` seems fine too.

What I'd really love to see is a videojs web component instead of a div, if that at all makes sense to do here.

gkatsev:
Haven't thought about whether this should be the default yet. The current pain-point is that if you use the video tag, you end up quickly flashing the default video element before video.js gets initialized. So, this will solve it. However, this will also work better for the react-like situations compared to what we currently have with the "ingested player div".

I do like the idea of a web-component, but I think that isn't necessarily equivalent to a div embed. Maybe just a custom element? However, native custom element support is pretty small right now and I would prefer not to have to include a polyfill for it.

src/js/player.js
ldayananda: Isn't it possible that the div will contain a video element already?

ldayananda:
I'm in favor of using `data-vjs-setup` and deprecating `data-setup`

mister-ben:
`data-vjs-setup` sounds like a sensible way to go.

gkatsev:
The broken PR was interesting because it revealed that IE9 doesn't allow using a source element outside of a video element. We'd want to make sure that this is documented well before this PR is merged.

gkatsev:
At the video.js TSC meeting, we talked about whether we can just use a `video-js` element and a `vjs-source` element. At first glance is that we can but with the `div` embed we can just re-use that as the player div and create a tech video element inside of it. However, it seems like we won't be able to do that with the `video-js` element. Will investigate further.

heff:
I thought custom elements were basically just renamed divs unless you
subclassed a native element.

gkatsev:
Oh, I think I figured out my issue. I hadn't updated the code to look for `video-js` over `div` and then I didn't actually copy the child elements over properly.

src/js/player.js
```diff
@@ -374,6 +375,8 @@ class Player extends Component {
 
     this.el_ = this.createEl();
 
+    this.options_.sources = Player.getTagSettings(this.tag).sources || this.options_.sources;
```
gkatsev: we might want to just update `getTagSettings to know about `vjs-source` as well, instead of this, though, this seems to work.

src/js/setup.js
```diff
@@ -31,6 +31,7 @@ const autoSetup = function() {
   // through each list of elements to build up a new, combined list of elements.
   const vids = document.getElementsByTagName('video');
   const audios = document.getElementsByTagName('audio');
+  const divs = document.getElementsByTagName('video-js');
```
gkatsev: we now look for `video-js` elements for `data-setup`, and so we *definitely* can keep `data-setup` now.

gkatsev:
@videojs/core-committers Should we only allow `video-js` in browsers newer than IE9? Then we won't even need to have `vjs-source` support.

src/js/player.js
gkatsev: This breaks some tests, so, removing and updating `getTagSettings`.

gkatsev:
Also, do we want the elements to be called `video-js` and `vjs-source` or just `videojs` and `vjssource`?

mister-ben:
Custom element names [should contain a hyphen for namespacing](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name).

`video-*` is probably too likely to conflict with something eventually. `videojs-player` might be safer.

gkatsev:
Yup, probably best to keep the hyphen, though it isn't technically a real custom element.
I don't think that any potential conflicts that `video-js` can have are big enough and having it match the name is super nice, which is why I was asking about whether we should just go with `videojs` at the element name.

docs/guides/embeds.md
```diff
@@ -0,0 +1,83 @@
+# How to Embed the Video.js player
```
gkatsev: after I added this page I remember that we have the `setup.md` file. This page can probably stay as is but `setup.md` probably should be updated a bit as well and point to this page.

gkatsev: Linked from `setup.md` now.

src/js/player.js
gkatsev: Missed this previously. Yes but I will say it's misconfigured in that case.

docs/guides/embeds.md
```diff
@@ -0,0 +1,85 @@
...
+
+Video.js is meant to be an ehnacement to the video element in HTML5 so for years, it's embed code has been just a `<video>` element.
```
misteroneill: Typo `ehnacement` and `it's` should be `its`

docs/guides/embeds.md
```diff
@@ -0,0 +1,85 @@
...
+Video.js then wraps the video element in a div that is used for us to place controls and anything else that's required for the player.
+
+For a long time this was enough. In 2016, a "div ingest" which allows the developer to give Video.js a player div to use instead of making it's own.
```
misteroneill: Second sentence is incomplete. Maybe needs `was added` somewhere?

docs/guides/embeds.md
```diff
@@ -0,0 +1,85 @@
...
+As you can see, it isn't much different from the classic `<video>` embed. It also does make it easier to work with [React](/docs/guides/react.md).
+
+### `<video-js>` embed
+This is the [I Can't Believe It's Not Custom Elements](https://developers.google.com/web/fundamentals/web-components/customelements) embed code.
+It looks very much like the `<video>` embed but instead of `video` it's a `video-js` embed.
+This is useful for all the things that the player div ingest is useful for and it matches our library name!
+
+```html
+<!-- via data-setup -->
+<video-js id="vid1" class="video-js" data-setup='{}'>
```
misteroneill: Should you need the `class="video-js"`? Could we add that if it's missing?

docs/guides/embeds.md
```diff
@@ -0,0 +1,85 @@
...
+#### Custom Elements
+Native Custom Elements support is relativly small according to [Can I Use](http://caniuse.com/#feat=custom-elementsv1) and because we didn't want to include a polyfill we're going with just an element called `video-js` rather than a full blown custom element.
+
+#### Browser support
+These all work in all browsers that Video.js supports, though, there are some caveats for some older browsers.
+- IE8 requires running `document.createElement('video-js')` before using the `video-js` embed code.
+- IE9 doesn't support having `source` elements outside of the `video` element, thus, the `video-js` embed will not work there. Though, if the source is set later, it should still work.
+
+## data-setup
+This is an ease-of-use method for having Video.js set up the player automatically. It is an html attribute and it takes a JSON string representation of the [player options](/docs/guides/options.md) as the value.
```
misteroneill: `html` should be all caps.

docs/guides/embeds.md
gkatsev: Could make sense to add it for the `video-js` element. Adding it unconditionally could potentially cause stylistic bugs.

pagarwal123:
QA-LGTM

docs/guides/embeds.md
gkatsev: oops, yes, something like that.

src/js/player.js
```diff
@@ -508,34 +509,59 @@ class Player extends Component {
...
+    if (divEmbed) {
+      el = this.el_ = tag;
+      tag = this.tag = document.createElement('video');
+      while (el.children.length) {
+        tag.appendChild(el.firstChild);
+      }
+
+      if (el.className.indexOf('video-js') === -1) {
+        el.className += ' video-js';
+      }
```
misteroneill: Should we use `dom.hasClass()` here? Wouldn't this leave room for false positives like if it had `video-js-xyz`?

misteroneill: And `addClass()` maybe?

gkatsev: oh, yeah, I forgot we had helpers for this. Didn't use the player methods because it isn't quite available yet.

gkatsev:
I think the IE8 shenanigans might be an actual issue where if you call `videojs()` on a previously made player with `video-js`, we end up creating the player again. or something. Will investigate tomorrow.

gkatsev:
So, if you've seen the last ~15 or so commits, I've been testing and trying to figure out why tests are broken on IE8. Turns out it's a slightly memory leak and it's causing the IE8 on browserstack to blow up. The tests pass locally just fine. My proposal is that these new tests are wrapped in an `if (IS_IE8)` to unblock this PR. I'll then spend time tracking down the memory leaks that cause this so we can continue writing tests and having them run on IE8.

PR #4643, opened 2017-10-03
fix: Make sure time displays use correctly-formatted time.

misteroneill:
The recent refactoring of time displays is great, but it seems there was a bug where it would always prepend times with `-` even when we want a positive number (i.e. everywhere except remaining time display)!

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -66,8 +66,23 @@ class TimeDisplay extends Component {
     if (this.textNode_) {
       this.contentEl_.removeChild(this.textNode_);
     }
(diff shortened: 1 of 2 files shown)


src/js/control-bar/time-controls/time-display.js
```diff
@@ -66,11 +66,26 @@ class TimeDisplay extends Component {
     if (this.textNode_) {
       this.contentEl_.removeChild(this.textNode_);
     }
-    this.textNode_ = document.createTextNode(` -${this.formattedTime_ || '0:00'}`);
+    this.textNode_ = document.createTextNode(this.formattedTime_ || '0:00');
```
misteroneill: This was the relevant mistake. Not all sub-classes wanted to display negative formatted times.

PR #4648, opened 2017-10-05
feat: set the seek bar to 100% on ended

BrandonOCasey:
## Description
As with the other prs that just went in, we want to make it seem like a video has ended from a user perspective, even if internally e know something is off. (currentTime !== duration, but the video has ended).

## Specific Changes proposed
A small refactor of `seekBar` to allow this

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -35,8 +35,6 @@ class SeekBar extends Slider {
   constructor(player, options) {
     super(player, options);
 
-    // @deprecated, we should change this to throttleUpdate in v7
-    // as modifying functions so that they always throttle is a bit weird
     this.update = Fn.throttle(Fn.bind(this, this.update), 50);
     this.on(player, 'timeupdate', this.update);
     this.on(player, 'ended', this.handleEnded);
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -34,8 +34,10 @@ class SeekBar extends Slider {
    */
   constructor(player, options) {
     super(player, options);
-    this.update = Fn.throttle(Fn.bind(this, this.update), 50);
-    this.on(player, ['timeupdate', 'ended'], this.update);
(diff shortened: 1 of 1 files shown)


misteroneill:
I'd like to do some testing, but overall this looks solid.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -53,53 +55,88 @@ class SeekBar extends Slider {
...
+   *
+   * @private
    */
-  getPercent() {
-
-    // Allows for smooth scrubbing, when player can't keep up.
-    const time = (this.player_.scrubbing()) ?
+  getCurrentTime_() {
+    return (this.player_.scrubbing()) ?
       this.player_.getCache().currentTime :
```
gkatsev: huh, we have a `getCache()` method?

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -34,8 +34,10 @@ class SeekBar extends Slider {
    */
   constructor(player, options) {
     super(player, options);
-    this.update = Fn.throttle(Fn.bind(this, this.update), 50);
-    this.on(player, ['timeupdate', 'ended'], this.update);
+    this.throttleUpdate = Fn.throttle(Fn.bind(this, this.update), 50);
```
gkatsev: I'd be inclined to keep this `this.update`. If people call the `update` method they won't be throttled with this change which could be unexpected.

src/js/control-bar/progress-control/seek-bar.js
BrandonOCasey: yeah haha

src/js/control-bar/progress-control/seek-bar.js
BrandonOCasey: Would you be okay with adding a `@deprecated` comment here so that it can be changed in a major version bump? I don't think we want to modify functions directly, so that they can be called without a throttle.

gkatsev: internally, with your change, `update_` is available to be called unthrottled, no?

BrandonOCasey: true, yeah lest leave it at that

PR #4649, opened 2017-10-05
feat: allow progress controls to be disabled

BrandonOCasey:
## Description
Sometimes you do not want the user to be able to use the progress controls, such as during an ad. It would be quite difficult to do this outside of video.js due to the complexity of it. I think that we should allow this within video.js.

## Specific Changes proposed
Add functions to disable/enable progress controls

## Requirements Checklist

### DIFF
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -8,6 +8,10 @@
   min-width: 4em;
 }
 
+.video-js .vjs-progress-control.disabled {
+  cursor: default;
+}
+
 .vjs-live .vjs-progress-control {
   display: none;
 }
@@ -36,7 +40,7 @@
 
 // This increases the size of the progress holder so there is an increased
 // hit area for clicks/touches.
-.video-js .vjs-progress-control:hover .vjs-progress-holder {
+.video-js .vjs-progress-control:hover .vjs-progress-holder.enabled {
   font-size: 1.666666666666666666em;
 }
 
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -45,7 +45,7 @@
 }
 
 .video-js .vjs-progress-control:hover .vjs-progress-holder.disabled {
-  font-size: inherit;
+  font-size: 1em;
 }
 
 // .vjs-play-progress / PlayProgressBar and .vjs-load-progress / LoadProgressBar
@@ -139,6 +139,10 @@
   // Ensure that we maintain a font-size of ~10px.
   font-size: 0.6em;
   visibility: visible;
+}
+
+.video-js .vjs-progress-control.disabled:hover .vjs-time-tooltip {
+  font-size: 1em;
 }
 
 // .vjs-mouse-display / MouseTimeDisplay
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -40,9 +40,14 @@
 
 // This increases the size of the progress holder so there is an increased
 // hit area for clicks/touches.
-.video-js .vjs-progress-control:hover .vjs-progress-holder.enabled {
+.video-js .vjs-progress-control:hover .vjs-progress-holder {
   font-size: 1.666666666666666666em;
 }
+
+.video-js .vjs-progress-control:hover .vjs-progress-holder.disabled {
+  font-size: inherit;
+}
+
 
 // .vjs-play-progress / PlayProgressBar and .vjs-load-progress / LoadProgressBar
 //
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -48,7 +48,6 @@
   font-size: inherit;
 }
 
-
 // .vjs-play-progress / PlayProgressBar and .vjs-load-progress / LoadProgressBar
 //
 // These are bars that appear within the progress control to communicate the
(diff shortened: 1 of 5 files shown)


thijstriemstra:
>  It would be quite difficult to do this outside of video.js due to the complexity of it.

Here's what I used:

```js
if (this.player.options_.controls) {
    // progress control isn't used by this plugin
    this.player.controlBar.progressControl.hide();
}
```

gkatsev:
Hm... I'm not sure we want this. Disabling built-ins like this seems really weird to me. Since I can see cases where you'd want the progress control but just disable seeking but right now this disables it completely, if I understand it correctly.
One thing that I was thinking off middleware is the ability to cancel a chain, this is something that's currently missing and I think might fit better for the specific usecase.

BrandonOCasey:
Well the real issue here is @gkatsev there isn't a good way to disable dragging on the progress bar when you want to. Disabling clicks is easy enough. All that this PR does is disable interacting with the progress bar. It is still there and works as intended, you just cant use it to seek.

gkatsev:
Yeah, maybe this isn't that bad. We should either keep listening to `mousemove` or we should hide the mouse time display when disabled. Also, we probably want to change the cursor to regular when the control bar is disabled.

Also, probably should just be called `disable` and `enable`.

gkatsev:
Maybe just hiding the mouse time display is best as another indicator that this is a "read-only" control would be better.

BrandonOCasey:
All good ideas, I will implement them when I get some time.

src/css/components/_progress.scss
```diff
@@ -8,6 +8,10 @@
   min-width: 4em;
 }
 
+.video-js .vjs-progress-control.disabled {
+  cursor: default;
```
BrandonOCasey: we don't really want a pointer when its disabled

src/css/components/_progress.scss
```diff
@@ -36,7 +40,7 @@
 
 // This increases the size of the progress holder so there is an increased
 // hit area for clicks/touches.
-.video-js .vjs-progress-control:hover .vjs-progress-holder {
+.video-js .vjs-progress-control:hover .vjs-progress-holder.enabled {
   font-size: 1.666666666666666666em;
```
BrandonOCasey: we only want to increase the size on hover when enabled

src/js/slider/slider.js
```diff
@@ -31,17 +31,74 @@ class Slider extends Component {
...
+    this.off('touchstart', this.handleMouseDown);
+    this.off('focus', this.handleFocus);
+    this.off('blur', this.handleBlur);
+    this.off('click', this.handleClick);
+    this.off(this.player_, 'controlsvisible', this.update);
+    this.off(doc, 'mousemove', this.handleMouseMove);
+    this.off(doc, 'mouseup', this.handleMouseUp);
+    this.off(doc, 'touchmove', this.handleMouseMove);
+    this.off(doc, 'touchend', this.handleMouseUp);
+    this.removeAttribute('tabindex');
```
BrandonOCasey: Don't allow tabbing to the slider when it is disabled

src/css/components/_progress.scss
```diff
@@ -36,7 +40,7 @@
 
 // This increases the size of the progress holder so there is an increased
 // hit area for clicks/touches.
-.video-js .vjs-progress-control:hover .vjs-progress-holder {
+.video-js .vjs-progress-control:hover .vjs-progress-holder.enabled {
```
gkatsev: do we need an `enabled` class? Can that just be the default and also have a `disabled` class?

src/js/slider/slider.js
```diff
@@ -31,17 +31,74 @@ class Slider extends Component {
...
 
-    this.on(player, 'controlsvisible', this.update);
+    this.on(this.player_, 'controlsvisible', this.update);
+
+    if (this.playerEvent) {
+      this.on(this.player_, this.playerEvent, this.update);
+    }
+
+    this.removeClass('disabled');
+    this.addClass('enabled');
```
gkatsev: I don't think we need an enabled class, like mentioned earlier

src/js/slider/slider.js
gkatsev: Good call.

src/css/components/_progress.scss
BrandonOCasey: sure we can do that

src/css/components/_progress.scss
```diff
@@ -40,6 +44,10 @@
   font-size: 1.666666666666666666em;
 }
 
+.video-js .vjs-progress-control:hover .vjs-progress-holder.disabled {
```
gkatsev: I don't think this block is necessary and it causes the time display not to look right
screen shot 2017-10-26 at 11 25 00

Though, now that I think about it, this is probably on purpose so that the progress holder/seek handle isn't as large?

nicu-chiciuc:
Didn't want to open a new issue since this one seems to be quite close. Basically based on [Slider.js:78](https://github.com/videojs/video.js/blob/master/src/js/slider/slider.js#L78) I should be able to call `player.controlBar.progressControl.seekBar.disable()` and the user should not be able to seek anywhere.

Although the method is called and all the events are set `off`, the events still persist.

I'm currently trying to adapt [danielcebrian/rangeslider-videojs](https://github.com/danielcebrian/rangeslider-videojs) to my needs so it will take some time to come up with some jsfiddle that shows this in action.

PR #4657, opened 2017-10-10
chore(gh-release): no console log on success

gkatsev:
We've been printing the resulting value from a successful call to the ghrelease, this was useful for debugging but isn't necessary anymore.
Add the result to the error conditional with some headings.

### DIFF
--- a/build/gh-release.js
+++ b/build/gh-release.js
@@ -36,9 +36,8 @@ if (args.prerelease || npmargs.some(function(arg) { return /next/.test(arg); }))
 ghrelease(options, function(err, result) {
   if (err) {
     console.log('Unable to publish release to github');
-    console.log(err);
+    console.log('err:', err, '\nresult:', result);
   } else {
     console.log('Publish release to github!');
-    console.log(result);
   }
 });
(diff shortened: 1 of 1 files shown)


build/gh-release.js
```diff
@@ -36,9 +36,8 @@ if (args.prerelease || npmargs.some(function(arg) { return /next/.test(arg); }))
 ghrelease(options, function(err, result) {
   if (err) {
     console.log('Unable to publish release to github');
-    console.log(err);
+    console.log('err:', err, '\nresult:', result);
```
BrandonOCasey: Maybe we should switch this to `console.error` not that it really matters that much. It also seems weird to have an explicit new line here. We should probably just use another `console` statement.

build/gh-release.js
```diff
@@ -35,10 +35,10 @@ if (args.prerelease || npmargs.some(function(arg) { return /next/.test(arg); }))
 
 ghrelease(options, function(err, result) {
   if (err) {
-    console.log('Unable to publish release to github');
-    console.log(err);
+    console.error('Unable to publish release to github');
+    console.error('err:', err);
+    console.error('\nresult:', result);
```
BrandonOCasey: do we still want the `\n` here? should we just use a blank console.error?

PR #4660, opened 2017-10-12
feat: sourceset

BrandonOCasey:
## Description
For a long time there has been talk about knowing when a source has changed via a `sourcechange` event. 

I think I finally got the implementation right here, but i'm sure that I missed some scenarios. 

> Note: Currently this will not work for IE < 11 because MutationObserver is not supported. I am going to look into that.

## Specific Changes proposed
* Watch for `src` attribute change on the video element.
* Trigger a source change on the second source set

## Requirements Checklist


src/js/player.js
```diff
@@ -312,6 +312,34 @@ class Player extends Component {
...
+    }
+    const triggerSourceChange = () => {
+      if (!sourceSet) {
+        sourceSet = true;
+        return;
+      }
+      this.trigger('sourcechange');
+    };
+
+    if (window.MutationObserver) {
```
BrandonOCasey: for now we use `MutationObserver`, but there has got to be another way that is more portable across all the browsers.

test/unit/source-change.test.js
```diff
@@ -0,0 +1,341 @@
...
+import document from 'global/document';
+import window from 'global/window';
+
+const testSrc = {
+  src: 'http://vjs.zencdn.net/v/oceans.mp4',
+  type: 'video/mp4'
+};
+
+QUnit.module('sourcechange', function() {
+  QUnit.module('videojs sources', {
```
BrandonOCasey: I think this is all the ways to set sources, feel free to tell me I need more here though

gkatsev: Basically 2 main methods: via a src attribute or source element. And these have corresponding vjs options like the `sources` option and also just calling .src() on the player. So, guess it makes 4 total ways of doing it.

src/js/player.js
gkatsev: we can definitely say it's only available on browsers where MutationObserver is supported for this new feature given the costs of supporting it for all browsers.

src/js/player.js
```diff
@@ -312,6 +312,44 @@ class Player extends Component {
...
+      this.videoSrcObserver_ = new window.MutationObserver(triggerSourceChange);
+      this.videoSrcObserver_.observe(tag, {attributes: true, attributeFilter: ['src']});
+    } else {
+      Object.defineProperty(tag, 'src', {
```
gkatsev: 🤔  I'm really not a fan of modifying built-ins like this...

BrandonOCasey: Well we could just skip that part, and only do `sourcechange` for browsers that support `MutationObserver`, but I think that would limit the usefulness of this feature.

BrandonOCasey:
Newest update is verified to work on Andriod 6-8, IOS 7-9, Safari 8-11, IE 9-11, Edge, FireFox, and Chrome

src/js/player.js
```diff
@@ -312,6 +312,37 @@ class Player extends Component {
     // Run base component initializing with new options
     super(null, options, ready);
 
+    // we don't want a source change for the first source set
```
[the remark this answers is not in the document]
BrandonOCasey: Well technically the source won't have changed as it is initially being set, I could see us changing the event to `sourceset` though. Then we could listen for that and determine if the source has changed.

src/js/player.js
```diff
@@ -312,6 +312,9 @@ class Player extends Component {
     // Run base component initializing with new options
     super(null, options, ready);
 
+    // was a source set intially? used to determine if we should fire a sourcechange
+    this.initialSourceSet_ = tag.src ? true : false;
```
forbesjo: why not `!!tag.src`?

BrandonOCasey: that would work too

BrandonOCasey: seems like it doesn't work as intended as tag.src returns weird things on IE

BrandonOCasey:
Some questions that have come up from this:

* Should the initial source set fire a `sourcechange`?
* Should this just be changed to `sourceset`?
* Should this be changed to something more general? changing the source often "reloads" the player

gkatsev:
I like `sourceset`, consistency across all cases is nice. However, I'm really against modifying the built-in prototypes like this unless it's for polyfilling broken behavior (like we did for `canPlayType` on old android devices). After thinking about it, I don't have a problem with MutationObservers for this usecase even if it still feels a bit weird.
Also, I added this to the TSC agenda so we can talk about it.

gkatsev:
In the TSC discussion, we all liked `sourceset` more. Using MediaObservers would be cool but given that they're asynchronous, it might mean that we could get the event but the `src` is still set to the old value. We should investigate, if it works great, it would be better than modifying the attributes on the video element we created/used. Though, if it doesn't work if overriding the `src` attribute on the element we create works, that might be OK too.

BrandonOCasey:
Got something working that supports any browser above IE 8, the only current issue is implementing the old mutation observer so that we can know when the video element changed out from under us.

src/js/player.js
```diff
@@ -823,6 +833,151 @@ class Player extends Component {
...
+      this.videoTagObserver_ = new window.MutationObserver((muts) => {
+        muts.forEach((mut) => {
+          const addedNodes = Array.prototype.filter.call(mut.addedNodes, (node) => (/video|audio/i).test(node.tagName));
+
+          if (!addedNodes.length) {
+            return;
+          }
+          setupSrcObserver();
+        });
+      });
```
BrandonOCasey: We need to implement the old mutation event handler here

src/js/player.js
```diff
@@ -823,6 +833,151 @@ class Player extends Component {
...
+    if (!sources.length && !src) {
+      return;
+    }
+
+    if (el.getAttribute('src') === null) {
+      src = null;
+    }
+
+    if (src) {
+      src = {src, type: 'video/mp4'};
```
BrandonOCasey: This is hardcoded for now, we can either use a MIME_TYPE list, or just specify it as unknown. I am also open to suggestions.

gkatsev:
The old mutation observer meaning mutation events? If so, I think we should avoid them. They were removed from the standard and cause huge performance issues when enabled.

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,499 @@
...
+let qunitFn = 'module';
+
+if (videojs.browser.IS_IE8) {
+  qunitFn = 'skip';
+}
+
+const Html5 = videojs.getTech('Html5');
+const oldMovingMedia = Html5.prototype.movingMediaElementInDOM;
+
+const validateSource = function(assert, player, sources, checkVideoSource = true) {
```
BrandonOCasey: All the tests validate that the source is correct, although some have to skip the "video" source, which can technically be empty string/null if we are using a source element.

BrandonOCasey:
Okay so since we are now modifying the properties (which has much more support) and those properties are copied by us when the media element changes in the tech on iOS, we don't even need mutation observers. I updated the tests and tested on all platforms again. Seems like IE8 wont be supported and iOS7 or less won't be supported. Every other platform works!

BrandonOCasey:
The last question is what we do about the type of a source when we only have a `mediaEl.src` string. Do we want to have a list of MIME_TYPES? just make it unknown? something else?

src/js/player.js
```diff
@@ -823,6 +825,179 @@ class Player extends Component {
...
+
+      Object.defineProperty(el, 'src', {
+        get() {},
+        set() {}
+      });
+    } catch (e) {
+      return;
+    }
+
+    const el = this.el().getElementsByTagName('video')[0] || this.el().getElementsByTagName('audio')[0];
```
misteroneill: This could be simplified to `this.$('video') || this.$('audio')`.

src/js/player.js
```diff
@@ -823,6 +825,179 @@ class Player extends Component {
...
+
+      this.ready(() => this.trigger('sourceset'), true);
+
+      return retval;
+    };
+
+    this.on('dispose', () => {
+      if (!el) {
+        return;
+      }
```
misteroneill: When would this condition be `true`?

src/js/player.js
```diff
@@ -823,6 +825,179 @@ class Player extends Component {
...
+  }
+
+  /**
+   * Update the internal source cache based on the `sourceset` event.
+   *
+   * @param {EventTarget~Event} e
+   *        The sourceset event that triggered this function to run.
+   */
+  handleSourceSet_(e) {
+    const el = this.el().getElementsByTagName('video')[0] || this.el().getElementsByTagName('audio')[0];
```
misteroneill: Same comment as above about using `$`.

src/js/player.js
```diff
@@ -823,6 +825,179 @@ class Player extends Component {
...
+    this.on('sourceset', this.handleSourceSet_);
+  }
+
+  /**
+   * Update the internal source cache based on the `sourceset` event.
+   *
+   * @param {EventTarget~Event} e
+   *        The sourceset event that triggered this function to run.
+   */
+  handleSourceSet_(e) {
```
misteroneill: I feel like this function could use some more comments for future maintainers.

BrandonOCasey: added a bit more comments

src/js/player.js
BrandonOCasey: I don't think it ever will be, but I just assume anything is possible these days haha. I don't feel too strongly about removing it, but I would prefer to just comment that we don't think it will run.

BrandonOCasey:
Moved the source cache updates to a new pr: https://github.com/videojs/video.js/pull/4717

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,506 @@
...
+}
+
+const Html5 = videojs.getTech('Html5');
+const oldMovingMedia = Html5.prototype.movingMediaElementInDOM;
+
+const validateSource = function(assert, player, sources, checkMediaElSource = true) {
+  const tech = player.tech_;
+  const mediaEl = tech.el();
+
+  // assert.deepEqual(player.currentSource(), sources[0], 'currentSource is correct');
```
BrandonOCasey: Once we actually change the source cache based on a sourceset we will do this.

src/js/player.js
```diff
@@ -823,6 +825,112 @@ class Player extends Component {
...
+    if (typeof srcDescriptor.enumerable === 'undefined') {
+      srcDescriptor.enumerable = true;
+    }
+
+    Object.defineProperty(el, 'src', {
+      get: srcDescriptor.get.bind(el),
+      set: (v) => {
+        const retval = srcDescriptor.set.call(el, v);
+
+        this.ready(() => this.trigger('sourceset'), true);
```
gkatsev: should this `ready` be synchronous?

src/js/player.js
```diff
@@ -823,6 +825,112 @@ class Player extends Component {
...
+
+        return retval;
+      },
+      configurable: true,
+      enumerable: srcDescriptor.enumerable
+    });
+
+    const oldSetAttribute = el.setAttribute;
+
+    el.setAttribute = (n, v) => {
```
gkatsev: does setting the `src` via `setAttribute` actually change the source of the element? if not, we could then skip this.

src/js/player.js
```diff
@@ -823,6 +825,112 @@ class Player extends Component {
...
+
+      return retval;
+    };
+
+    const oldLoad = el.load;
+
+    el.load = () => {
+      const retval = oldLoad.call(el);
+
+      this.ready(() => this.trigger('sourceset'), true);
```
gkatsev: it's possible that `load` is called multiple times. It won't necessarily mean that a new source was set.

src/js/player.js
```diff
@@ -823,6 +825,112 @@ class Player extends Component {
   }
 
   /**
+   * Modify the video/audio element so that we can detect when
+   * the source is changed. Fires `sourceset` just after the source has changed
+   *
+   * @fires Player#sourceset
+   */
+  watchForSourceSet_() {
```
gkatsev: This change is basically HTML5 only and should probably be moved into the html5 tech. However, this leaves us with the question of what to do for other techs like Flash?
Also, what to do if we're on a browser like iOS 7? Should we fallback and trigger `sourceset` at the earliest opportunity that we know the source has changed otherwise? For example on a `loadstart` if we haven't already triggered a `sourceset`? It's possible that this also answers the Flash-tech issue.

src/js/player.js
BrandonOCasey: It wont, but the video element will reload as if a source was set. Basically its the difference between `sourcechange` and `sourceset`

src/js/player.js
BrandonOCasey: setAttribute does actually change the source

src/js/player.js
BrandonOCasey: this is difficult because we want the intial `sourceset` to not fire until the player is ready, so that users have a chance to listen to it. But we want all others to fire as soon as the source has been changed.

src/js/player.js
BrandonOCasey: Yeah this does seem like it should be in HTML5. I would think that we should try to support sourceset as well as we can in older browsers and other techs, but I think we may want to do that in another PR as this is already quite big.

gkatsev: That sounds reasonable. As long as we have an idea of a path forward.

BrandonOCasey:
OK this should be ready for re-review! Now everything is in Html5 and Tech.

BrandonOCasey:
rebased on master

pagarwal123:
QA LGTM

gkatsev:
I'd still like to take a look but it's good that QA passed!

gkatsev:
Looks like some issues on IE11: https://travis-ci.org/videojs/video.js/builds/325606889#L3448

BrandonOCasey:
strange that it doesn't cause the tests to fail, and doesn't seem to happen when I test it manually.

BrandonOCasey:
seems like the issues were mostly from some strange test mocking

src/js/tech/tech.js
```diff
@@ -989,6 +989,14 @@ Tech.prototype.featuresPlaybackRate = false;
  */
 Tech.prototype.featuresProgressEvents = false;
 
+/**
+ * Boolean indicating wether the `Tech` supports the `sourceset` event.
+ *
+ * @type {boolean}
+ * @default
+ */
+Tech.prototype.featuresSourceset = false;
```
gkatsev: so by default, techs will not trigger `sourceset` but a particular tech can override it and trigger a `sourceset`?

src/js/player.js
```diff
@@ -83,6 +83,24 @@ const TECH_EVENTS_RETRIGGER = [
    */
   'abort',
 
+  /**
+   * Fired when the source is set on the tech causing the media element
+   * to reload.
+   *
+   * @private
+   * @method Player#handleTechSourceset_
```
gkatsev: is this a holdover from before extracting out https://github.com/videojs/video.js/pull/4717?

src/js/tech/html5.js
```diff
@@ -119,6 +121,125 @@ class Html5 extends Tech {
...
+
+      return retval;
+    };
+
+    const oldLoad = el.load;
+
+    el.load = () => {
+      const retval = oldLoad.call(el);
+
+      this.triggerSourceSet_();
```
gkatsev: so, because we're now calling this event `sourceset` and the [html5 spec](https://html.spec.whatwg.org/multipage/media.html#dom-media-load) says that everything needs to be aborted and the resource selection algorithm is re-run, we want to trigger a `sourceset` here?
This is definitely something we'll want to document.

test/unit/sourceset.test.js
gkatsev: Is this change made in https://github.com/videojs/video.js/pull/4717? Can we remove it from here then?

test/unit/tech/html5.test.js
```diff
@@ -686,7 +688,7 @@ QUnit.test('Html5#reset calls Html5.resetMediaElement when called', function(ass
   Html5.resetMediaElement = oldResetMedia;
 });
 
-QUnit.test('When Android Chrome reports Infinity duration with currentTime 0, return NaN', function(assert) {
+test('When Android Chrome reports Infinity duration with currentTime 0, return NaN', function() {
```
gkatsev: why these changes?

test/unit/tech/html5.test.js
```diff
@@ -388,9 +388,11 @@ if (Html5.supportsNativeTextTracks()) {
       addEventListener: (type, fn) => events.push([type, fn]),
       removeEventListener: (type, fn) => events.push([type, fn])
     };
-    const el = document.createElement('div');
+    const el = document.createElement('video');
```
gkatsev: is this related to test failures?

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,532 @@
...
+        }
+        this.hook = (player) => player.on('sourceset', () => this.sourcesets++);
+        videojs.hook('setup', this.hook);
+
+        this.sourcesets = 0;
+        this.fixture = document.getElementById('qunit-fixture');
+
+        if ((/audio/i).test(testName)) {
+          this.mediaEl = document.createElement('audio');
+          this.testSrc = {
```
gkatsev: fwiw, we can reuse the same `testSrc` for both element types.

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,532 @@
...
+        if ((/audio/).test(testName)) {
+          this.sourceOne = {src: 'http://example.com/one.mp3', type: 'audio/mpeg'};
+          this.sourceTwo = {src: 'http://example.com/two.mp3', type: 'audio/mpeg'};
+        }
+
+        this.mediaEl.className = 'video-js';
+        this.fixture.appendChild(this.mediaEl);
+      },
+      afterEach(assert) {
+        const done = assert.async();
```
gkatsev: you can make `afterEach` async?
Also, is there a particular reason for making it async? I would prefer if we can make it synchronous using sinon's timers.

gkatsev: I won't bother commenting on each async test below, but it applies there too :)

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,532 @@
...
+
+        if (typeof this.totalSourcesets === 'undefined') {
+          this.totalSourcesets = 1;
+        }
+
+        window.setTimeout(() => {
+          assert.equal(this.sourcesets, this.totalSourcesets, 'no additional sourcesets');
```
gkatsev: running assertions in the `afterEach` just feels wrong to me.

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,532 @@
...
+        assert.equal(this.sourcesets, 0, 'no sourceset');
+        done();
+      }, wait);
+    });
+  }));
+
+  QUnit.module('source change', (subhooks) => testTypes.forEach((testName) => {
+    QUnit.module(testName, {
+      beforeEach(assert) {
+        sinon.stub(log, 'error');
```
gkatsev: is there much difference in this `beforeEach` and the one above? Could probably pull out to one function above and reference both in the object here.

test/unit/sourceset.test.js
BrandonOCasey: yeah I think we can remove it from here

src/js/tech/html5.js
```diff
@@ -119,6 +121,125 @@ class Html5 extends Tech {
...
+
+      return retval;
+    };
+
+    this.on('dispose', () => {
+      if (!el) {
+        return;
+      }
+
+      if (oldLoad) {
```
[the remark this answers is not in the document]
BrandonOCasey: well if we throw out the video element no, maybe we can just add a comment that we don't need to.

src/js/tech/tech.js
BrandonOCasey: I would think so, since its not really part of the spec, but I could see it being really helpful to have all techs do it, maybe in a major version bump?

src/js/player.js

src/js/tech/html5.js
BrandonOCasey: Exactly. I think we could use a comment here and also a separate doc pr

test/unit/tech/html5.test.js
BrandonOCasey: It is. I am not sure about the assert below anymore, but it was causing some weird error in IE11 of all things, could be worth re-investigating the assert again. This change seems harmless though.

test/unit/sourceset.test.js
BrandonOCasey: If we want to mock how an actual video element acts we could, but this is more of a test against the real video element.

BrandonOCasey: Also yeah you can make any step async.

test/unit/sourceset.test.js
BrandonOCasey: Well we could add logic to each to to make sure we don't get extra source sets, or a function that can be re-used, not sure which route to go.

test/unit/sourceset.test.js
BrandonOCasey: I think we could pull it out.

src/js/tech/tech.js
gkatsev: Most techs that aren't html5 can't get a source set directly like you can with the video element, so, might be ok to just trigger it in the `tech`'s `setSource()` method? But I think this is fine as a way forward for other techs.

test/unit/sourceset.test.js
gkatsev: might be fine to just leave as is

src/js/tech/html5.js
```diff
@@ -119,6 +121,125 @@ class Html5 extends Tech {
...
   }
 
+  /**
+   * A special function to trigger source set in a way that
+   * will allow player to re-trigger if player/html5 are not ready
+   * yet.
+   *
+   * @fires Tech#sourceset
+   */
+  triggerSourceSet_() {
```
gkatsev: should we trigger `sourceset` asynchronously? Also, I'm thinking maybe moving this method to the `tech` object so that other techs could just call `this.triggerSourceSet_()` whenever necessary if/when they build out support for this.

BrandonOCasey: I could see moving triggerSourceSet out and I think that makes sense

BrandonOCasey: I am not sure if that would work, I guess we could include the source that was set in the event and it would be fine. But at one point it was said that we want something like this to work:
```this.on('sourceset', (e) => {
  console.log(player.currentSource())
});

// should log `{type: 'fake1', src: 'test1'}` above
player.src({type: 'fake1', src: 'test1'});

// should log `{type: 'fake2', src: 'test2'}` above
player.src({type: 'fake2', src: 'test2'});

// should log `{type: 'fake3', src: 'test3'}` above
player.src({type: 'fake3', src: 'test3'});

```


or maybe this

```this.one('sourceset', (e) => {
  // should be fake1, test1
  console.log(player.currentSource());

  this.one('sourceset', () => {
     // should be fake2, test2
     console.log(player.currentSource());
  })
  player.src({type: 'fake2', src: 'test2'});
});

player.src({type: 'fake1', src: 'test2'});
```

If it was async I am not sure that it would

src/js/tech/tech.js
BrandonOCasey: Yeah Just triggering in there may make sense, and setting this to true also makes sense

test/unit/tech/html5.test.js
BrandonOCasey: IE11 was dying here somehow, maybe it was related to the memory issues we fixed and now all this can be reverted

test/unit/tech/html5.test.js
BrandonOCasey: This may have also been due to the memory issues during testing, maybe this can all be reverted

gkatsev: Ah, this is because we're getting the `src` getter from the media element and applying it to the div which safari doesn't like. So, unless we update setting the src getter code, I guess this piece is fine to leave.

src/js/player.js
gkatsev: and I just realized that this was for exposing the doc for it publicly. So, we just need to reword this one and remove the `@private` from another.

gkatsev:
Here's how the current docs for it look: https://deploy-preview-4660--videojs-docs.netlify.com/player#event:sourceset

src/js/player.js
```diff
@@ -664,11 +687,14 @@ class Player extends Component {
...
    *
+   * @param {boolean} [skipListeners]
+   *        Skip the playerresize event trigger
+   *
    * @return {number}
    *         The current width of the `Player` when getting.
    */
-  width(value) {
-    return this.dimension('width', value);
+  width(value, skipListeners) {
```
BrandonOCasey: hmmm why is this different then master?

gkatsev: huh, weird. Maybe my local master checkout had an old version or something when I merged. I'll take a look.

gkatsev: Nope, I think the merge is failing and thinking that we want to keep this change when we reverted it in master. Lame. I don't really want to run a rebase. I'll probably cherry-pick the revert or something.

test/unit/sourceset.test.js
gkatsev: looks like there's a lot of overlap but there are some differences. I'll probably pull out the main bulk that's the same into a function.

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,479 @@
...
+  const mediaEl = tech.el();
+
+  if (checkMediaElSource) {
+    assert.equal(mediaEl.src, sources[0].src, 'mediaEl.src is correct');
+    assert.equal(mediaEl.getAttribute('src'), sources[0].src, 'mediaEl attribute is correct');
+    assert.equal(tech.src(), sources[0].src, 'tech is correct');
+  }
+};
+
+const setupEnv = function(env, testName) {
```
gkatsev: this is most of the shared code from the `beforeEach`

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,479 @@
...
+    type: 'video/mp4'
+  };
+  env.sourceOne = {src: 'http://example.com/one.mp4', type: 'video/mp4'};
+  env.sourceTwo = {src: 'http://example.com/two.mp4', type: 'video/mp4'};
+
+  env.mediaEl.className = 'video-js';
+  env.fixture.appendChild(env.mediaEl);
+};
+
+const setupAfterEach = function(totalSourcesets) {
```
gkatsev: this is all the code from the `afterEach`, everything was the same except the value set for `totalSourcesets`, so, just made a HOF.

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,479 @@
...
+
+  if (testName === 'change video el' || testName === 'change audio el') {
+    Html5.prototype.movingMediaElementInDOM = false;
+  }
+
+  env.sourcesets = 0;
+  env.hook = (player) => player.on('sourceset', () => env.sourcesets++);
+  videojs.hook('setup', env.hook);
+
+  if ((/audio/i).test(testName)) {
```
gkatsev: good to test creating with `audio` element but there's no need to set mp3 urls, plus the mp3 urls were creating a lot of 404 errors in the console, so, just using a single mp4 test source and only mp4 sources

test/unit/sourceset.test.js
```diff
@@ -0,0 +1,479 @@
+/* eslint-env qunit */
+import videojs from '../../src/js/video.js';
+import document from 'global/document';
+import window from 'global/window';
+import log from '../../src/js/utils/log.js';
+import sinon from 'sinon';
+
+const Html5 = videojs.getTech('Html5');
+const wait = 1;
```
gkatsev: @BrandonOCasey do you remember why this was set to 100 initially? Setting it to 1 halved the testing time for this module.

BrandonOCasey: one of the IE versions was too slow, and tests would time out and fail

gkatsev: well, everything seems to have passed on the first run. We can always set it to a slightly larger number later on for IEs or something.

BrandonOCasey:
LGTM, but its my pr so 🤷‍♂️

src/js/tech/tech.js
```diff
@@ -155,6 +155,30 @@ class Tech extends Component {
...
+   * A special function to trigger source set in a way that will allow player
+   * to re-trigger if the player or tech are not ready yet.
+   *
+   * @fires Tech#sourceset
+   */
+  triggerSourceset() {
+    if (!this.isReady_) {
+      // on initial ready we have to trigger source set
+      // 1ms after ready so that player can watch for it.
+      this.one('ready', () => this.setTimeout(this.triggerSourceset, 1));
```
gkatsev: to verify: if we have multiple source changes when we're not ready, when we get `sourceset`, would `player.currentSrc()` return the last source or each source?
Things that would affect it: how setting the source works when we're not ready. A potential workaround would be to add the source to the event object.

BrandonOCasey: hmm I think adding the source to the event would make sense instead of waiting for `ready`. We would just have to make sure that there is a way to do this:
```
var player = videojs(...);
player.on('sourceset', (e) => console.log('the first source was set');
```

I think that the source may be set before a listener can be added

gkatsev: I can confirm that if there are two source changes when the tech isn't ready, and we check the source, we get the latest version, currently.
```js
player.on('sourceset', () => console.log(player.currentSource()));
player.src({src: './oceans.webm', type: 'video/webm'});
player.src({src: './oceans.mp4', type: 'video/mp4'});
```
we'd get two logs of the oceans.mp4.

gkatsev: I think the only real solution is to add the source to the event and then document that if you want to know for sure what source was set when the event was triggered, you need to look at the event object, to know the latest one, look on the player, sort like how `currentTime` exists on the event object but no one uses it. We can also mention that this should only really be a problem for multiple sets around tech switching.

gkatsev: Since removing this waiting and the timeout means that a user cannot do something like:
```js
player.ready(() => player.on('sourceset', () =>  console.log(player.currentSrc())))
```
or
```js
```js
player.on('sourceset', () => player.on('sourceset', () =>  console.log(player.currentSrc())))
```

gkatsev: This is a really fun one. Turns out you'll end up getting `oceans.webm` after getting the `oceans.mp4` source because when `oceans.webm` is being triggered, the tech isn't ready yet but by the time that `oceans.mp4` is set, the tech is ready and so it ends up triggering immediately and then `oceans.webm` gets triggered shortly after.
So, you'll see something like:
```js
var player = videojs('vid', {
  techOrder: ['flash', 'html5']
});
// so that we'll change techs when we set a new source
player.src({ src: './oceans.flv', type: 'video/flv'});

player.on('sourceset', (e) => console.log(e.src, player.currentSrc()));
player.src({src: './oceans.webm', type: 'video/webm'});
player.src({src: './oceans.mp4', type: 'video/mp4'});
//> oceans.mp4, oceans.mp4
//> oceans.webm, oceans.mp4
```

gkatsev:
https://deploy-preview-4660--videojs-docs.netlify.com/player#event:sourceset

BonBonSlick:
I do not understand, it says "merged to master", but no such event in docs "sourceChange"

PR #4661, opened 2017-10-12
fix: Make sure we remove vjs-ended from the play toggle in all appropriate cases.

misteroneill:
There are cases where we might not see a `seeked` event when replaying a video. This will make this code more reliable.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -67,7 +67,6 @@ class PlayToggle extends Button {
    * @listens Player#seeked
    */
   handleSeeked(event) {
-    // remove the ended class
     this.removeClass('vjs-ended');
 
     if (this.player_.paused()) {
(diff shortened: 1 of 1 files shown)


src/js/control-bar/play-toggle.js
```diff
@@ -101,6 +99,7 @@ class PlayToggle extends Button {
    * @listens Player#pause
    */
   handlePause(event) {
+    this.removeClass('vjs-ended');
```
gkatsev: not sure we want this. If you're ended and the player is told to pause, it should look the same.

misteroneill:
Removed the one comment because it was kind of redundant.

src/js/control-bar/play-toggle.js
thijstriemstra: agreed, this ended thing is starting to look messy (see extra releases just to fix things related to this). +1 for not making the play-toggle more complex (which trickles down to plugins).

PR #4663, opened 2017-10-14
Update for Russian translation

estim:
Update for Russian translation


lang/ru.json
```diff
@@ -1,13 +1,18 @@
 {
+  "Audio Player": "Аудио проигрыватель",
+  "Video Player": "Видео проигрыватель",
   "Play": "Воспроизвести",
-  "Pause": "Приостановить",
+  "Pause": "Пауза",
```
gkatsev: Isn't Пауза a noun where-as Приостановить a verb, so, Приостановить would be more appropriate?

estim: Yes, I think that's a good point to have here a verb instead of a noun.

gkatsev: @estim can you revert this line change then? I can do it otherwise before merging.
Thanks.

PR #4665, opened 2017-10-16
fix: Fix cases where the player reports that it is ready during source selection, but before tech selection.

misteroneill:
In working on some issues with asynchronous middleware and playlists, I realized that there is a period of time during the source and tech setting process where a new source and tech have not been selected, but the player still considers itself "ready".

The effect of this is to cause any callbacks to `ready()` during source selection to be called immediately instead of waiting for the source/tech to be selected:

```js
player.src({src: 'foo.mp4', type: 'video/mp4'});
player.ready(() => {
  player.play();
});
```

I think most people would expect, in that case, `foo.mp4` to begin playback. However, if there is asynchronous middleware happening, what you may get is some playback of the current source and then a source change to `foo.mp4` (which will pause the player).

The essence of this change is to separate the player's readiness from the tech's readiness. The tech can be ready when the player is not because the player is still resolving middleware, etc.

This should be backward-compatible and make `ready()` more reliable and better match user expectations.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -24,6 +24,7 @@ import MediaError from './media-error.js';
 import safeParseTuple from 'safe-json-parse/tuple';
 import {assign} from './utils/obj';
 import mergeOptions from './utils/merge-options.js';
+import {silencePromise} from './utils/promise';
 import textTrackConverter from './tracks/text-track-list-converter.js';
 import ModalDialog from './modal-dialog';
 import Tech from './tech/tech.js';
@@ -459,8 +460,6 @@ class Player extends Component {
 
     this.on('fullscreenchange', this.handleFullscreenChange_);
     this.on('stageclick', this.handleStageClick_);
-
-    this.changingSrc_ = false;
   }
 
   /**
@@ -1623,35 +1622,33 @@ class Player extends Component {
   }
 
   /**
-   * start media playback
+   * Begin playback.
    *
    * @return {Promise|undefined}
-   *         Returns a `Promise` if the browser returns one, for most browsers this will
-   *         return undefined.
+   *         Returns a `Promise` only if the browser returns one and the player
+   *         is ready to begin playback. For most browsers and non-ready
+   *         situations, this will return undefined.
    */
   play() {
-    if (this.changingSrc_) {
-      this.ready(function() {
-        const retval = this.techGet_('play');
 
-        // silence errors (unhandled promise from play)
-        if (retval !== undefined && typeof retval.then === 'function') {
-          retval.then(null, (e) => {});
-        }
+    // If the player is not ready, queue up a call to the tech's `play()`
+    // method for when it _is_ ready.
+    if (!this.isReady_) {
+      this.ready(() => {
+        silencePromise(this.techGet_('play'));
       });
 
-    // Only calls the tech's play if we already have a src loaded
-    } else if (this.isReady_ && (this.src() || this.currentSrc())) {
+    // If the player is ready and there is a source, call the tech's `play()`
+    // method.
+    } else if (this.src() || this.currentSrc()) {
       return this.techGet_('play');
-    } else {
-      this.ready(function() {
-        this.tech_.one('loadstart', function() {
-          const retval = this.play();
 
-          // silence errors (unhandled promise from play)
-          if (retval !== undefined && typeof retval.then === 'function') {
-            retval.then(null, (e) => {});
-          }
+    // Finally, if the player is ready, but we don't have a source, wait for
+    // one to be set.
+    } else {
+      this.ready(() => {
+        this.tech_.one('loadstart', () => {
+          silencePromise(this.techGet_('play'));
         });
       });
     }
@@ -2304,7 +2301,7 @@ class Player extends Component {
 
     // intial sources
     this.cache_.sources = sources;
-    this.changingSrc_ = true;
+    this.isReady_ = false;
 
     // intial source
     this.cache_.source = sources[0];
@@ -2325,17 +2322,22 @@ class Player extends Component {
           this.error({ code: 4, message: this.localize(this.options_.notSupportedMessage) });
         }, 0);
 
-        // we could not find an appropriate tech, but let's still notify the delegate that this is it
-        // this needs a better comment about why this is needed
+        // We could not find an appropriate tech, but let's still notify the
+        // delegate that this is it. This needs a better comment about why this
+        // is needed.
         this.triggerReady();
 
         return;
       }
 
-      this.changingSrc_ = false;
+      // Now that we have a source (and a tech, thanks to `src_`) we can
+      // tell any subscribers that the player is ready again.
+      this.triggerReady();
+
       // video element listed source
       this.cache_.src = middlewareSource.src;
 
+      // Notify middlewares of a new tech.
       middleware.setTech(mws, this.tech_);
     });
   }
@@ -2361,7 +2363,6 @@ class Player extends Component {
     }
 
     if (!titleCaseEquals(sourceTech.tech, this.techName_)) {
-      this.changingSrc_ = true;
 
       // load this technology with the chosen source
       this.loadTech_(sourceTech.tech, sourceTech.source);
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1546,27 +1546,27 @@ class Player extends Component {
   }
 
   /**
-   * Pass values to the playback tech
+   * Call a method on the playback tech with an optional argument.
    *
-   * @param {string} [method]
-   *        the method to call
+   * Waits for the tech to be ready before executing the call.
    *
-   * @param {Object} arg
-   *        the argument to pass
+   * @param  {string} method
+   *         The name of the method to call on the tech.
+   *
+   * @param  {Object} [arg]
+   *         An argument to pass to the tech.
    *
    * @private
    */
   techCall_(method, arg) {
-    // If it's not ready yet, call method when it is
-
-    this.ready(function() {
+    this.tech_.ready(() => {
       if (method in middleware.allowedSetters) {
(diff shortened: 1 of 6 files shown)


src/js/player.js
```diff
@@ -1620,30 +1618,29 @@ class Player extends Component {
...
-        const retval = this.techGet_('play');
 
-        // silence errors (unhandled promise from play)
-        if (retval !== undefined && typeof retval.then === 'function') {
-          retval.then(null, (e) => {});
-        }
-      });
+    // Given a return value for a tech's `play()` method, silences "unhandled
+    // promise from play" errors.
+    const silencePromise = (value) => {
```
gkatsev: we probably want a promise helper file that contains `isPromise` and `silencePromise`, since we silence promises in a few places.

src/js/player.js
```diff
@@ -1620,30 +1618,29 @@ class Player extends Component {
...
-          // silence errors (unhandled promise from play)
-          if (retval !== undefined && typeof retval.then === 'function') {
-            retval.then(null, (e) => {});
-          }
-        });
-      });
+    // Finally, if the player is ready, but we don't have a source, wait for
+    // one to be set.
+    } else {
+      this.ready(() => this.tech_.one('loadstart', () => silencePromise(this.play())));
```
gkatsev: This will fail because the function gets called with the context of the player and not the tech as relied on previously.

src/js/player.js
```diff
@@ -2315,17 +2312,22 @@ class Player extends Component {
           this.error({ code: 4, message: this.localize(this.options_.notSupportedMessage) });
         }, 0);
 
-        // we could not find an appropriate tech, but let's still notify the delegate that this is it
-        // this needs a better comment about why this is needed
+        // we could not find an appropriate tech, but let's still notify the
+        // delegate that this is it this needs a better comment about why this
```
gkatsev: can you revert this change or put a new line in? Otherwise the two statements run in together.

src/js/player.js
misteroneill: D'oh.

src/js/player.js
misteroneill: Ah, yes, it is hard to distinguish sentences without capitals and punctuation. ;)

test/unit/tracks/text-tracks.test.js
```diff
@@ -574,7 +574,7 @@ if (Html5.isSupported()) {
     player.addRemoteTextTrack(track, false);
     player.src({src: 'example.mp4', type: 'video/mp4'});
 
-    this.clock.tick(1);
+    this.clock.tick(2);
```
misteroneill: Calling `src()` now results in two async functions being queued up in sequence. First, `middleware.setSource` runs async. Then, its callback invokes `this.triggerReady()`, which queues up a flush of the ready queue.

That's why this changed from `1` to `2`.

test/unit/player.test.js
```diff
@@ -1601,14 +1601,22 @@ QUnit.test('src_ does not call loadTech is name is titleCaseEquals', function(as
...
-    ready() {},
+    // ready() {},
+    load() {},
     loadTech_() {
       loadTechCalled++;
     }
   };
 
-  Player.prototype.src_.call(playerProxy);
+  Player.prototype.src_.call(playerProxy, {});
```
misteroneill: The changes to this test were necessary because `src_()` no longer calls `ready()`. Instead, it immediately loads the source into the tech because the only cases where this is possible is if the tech did not change; so, it should already be ready.

src/js/player.js
```diff
@@ -2274,58 +2271,77 @@ class Player extends Component {
...
     this.cache_.sources = sources;
-    this.changingSrc_ = true;
-
-    // intial source
     this.cache_.source = sources[0];
 
-    // middlewareSource is the source after it has been changed by middleware
+    // Set the player to a non-ready state while middleware asynchronously
+    // determines the final source.
+    this.isReady_ = false;
```
misteroneill: This is the key fix: the player should not be "ready" while middleware is resolving a source and deciding if it needs to load a new tech or set the resolved source on the existing tech.

test/unit/tracks/text-tracks.test.js
BrandonOCasey: Im just wondering if this is going to break plugin tests, I don't really see an issue other that that.

src/js/player.js
```diff
@@ -1613,35 +1612,33 @@ class Player extends Component {
...
    *
    * @return {Promise|undefined}
-   *         Returns a `Promise` if the browser returns one, for most browsers this will
-   *         return undefined.
+   *         Returns a `Promise` only if the browser returns one and the player
+   *         is ready to begin playback. For most browsers and non-ready
+   *         situations, this will return undefined.
    */
   play() {
-    if (this.changingSrc_) {
```
BrandonOCasey: Having this function sometimes return a promise just sucks, not really something we can/should do about it in this PR though.

misteroneill: I dunno. I mean, it does model the native `play` method. The alternative would be to include a `Promise` polyfill.

test/unit/tracks/text-tracks.test.js
misteroneill: It might.

src/js/player.js
gkatsev: I do think making a Promise is what we'd want to do eventually. Though, we'd probably ask users to provide a polyfill rather than include it in video.js. Then if Promises exists or was passed in we'll make one.
Either way, outside the scope of this PR :)

gkatsev:
@videojs/core-committers any thoughts or comments regarding this change?

gkatsev:
removing confirmed until this is tested so we don't merge it in early.

gkatsev:
I just tested this with videojs-playlist and autoadvance is broken with this but works with the released version 6.3.3. We might want to re think about this taking the `sourceset` PR into account.

alex-barstow:
@gkatsev I saw the same playlist behavior when re-testing to see if https://github.com/videojs/video.js/pull/4639 is still necessary. Autoadvance didnt work unless I wrapped [silencePromise(this.techGet_('play'))](https://github.com/videojs/video.js/pull/4665/files#diff-3b0266ff1c037b289ec624ab25b0272eR1628) in a timeout of some arbitrary amount (can't recall exact wait time), so perhaps the player is still reporting it is ready too early.

misteroneill:
@alex-barstow is correct. What's happening is that we're calling `triggerReady` when Video.js has updated its source, but the tech/video element still has the old source.

misteroneill:
I'm not sure this is going to work, but I'm going to gather some thoughts here, partly to get my head on straight.

The current and previous assumption within Video.js has been that the player's ready state is identical to the tech's ready state.

This PR aims to adjust that assumption such that the tech could be ready, but the player may still be resolving a source asynchronously via middleware; so, the player _should not_ be ready.

Why? Because, I think it's fair to assume as a user of Video.js that if I queue a ready callback on a player, I should be able to interact with it once that callback fires (e.g., calling `play()`). This is not currently the case.

However, the `src_` function uses `techCall_` to set the source, which relies on the tech/player being in a ready state. This leads to a chicken or egg situation that's not easy to solve. Changing `techCall_` to use `tech_.ready` instead causes a raft of weird side effects, such as... video not playing. 😆 

I'm going to spend some time today tinkering with this, but it may not be possible to work around without breaking something.

misteroneill:
Alright, turns out most of the issues with changing `techCall_` to wait for the tech, specifically, to be ready, were very minor.

src/js/player.js
```diff
@@ -2344,15 +2350,19 @@ class Player extends Component {
...
 
-      // Notify middlewares of a new tech.
-      middleware.setTech(mws, this.tech_);
+        // Cache the source URL.
+        this.cache_.src = middlewareSource.src;
+
+        // At this point, we've found a compatible tech and source and set the
+        // source both on the tech and in the internal cache. It's safe to tell
+        // any subscribers that the player is ready again.
+        this.triggerReady();
```
gkatsev: do we even need to do this? Should we be getting ready from the tech for the source change in the first place?

src/js/player.js
```diff
@@ -2324,10 +2330,10 @@ class Player extends Component {
       this.middleware_ = mws;
 
       // Attempt to set the source on a tech.
-      const foundTech = this.src_(middlewareSource);
+      const didNotFindTech = this.src_(middlewareSource);
```
gkatsev: that is a better name

src/js/player.js
misteroneill: Well, the whole idea here is that tech ready != player ready - this is the latter. The tech is ready before the source change is complete. Maybe I'm not following the line of questions.

gkatsev: Well, when the player gets the tech's `ready` event, it already triggers ready itself: https://github.com/videojs/video.js/blob/c7ad7b31f0d3a15a8d6cf82dfaf0033264fadd6b/src/js/player.js#L1055-L1061
So, we would end up getting two `ready` events for the same source unless we change the way that `handleTechReady` functions.

misteroneill:
@dmlap brought up an interesting question related to this. If you are using `on('ready')` will you get more ready events than you expect?

I wondered more about this possibly being a subtle breaking change and whether the alternative approach should be to suggest waiting for `canplay` before calling `play()`, leaving the `ready` as a tech crutch...

```js
player.src({src: 'foo.mp4', type: 'video/mp4'});
player.ready(() => {
  player.one('canplay', () => player.play());
});
```

gkatsev:
@misteroneill sounds like @dmlap and I had similar questions though worded differently.

misteroneill:
I wonder if a better fix is just to have `play()` wait for `loadstart` if it's changing sources?

misteroneill:
I don't think this is the right path forward. I'm going back to the drawing board.

Issue #3100, opened 2016-02-10
isCrossOrigin test Fail on IE 11


Naouak:
When trying to build on windows 10, I get this output with Karma tests:
````
IE 11.0.0 (Windows 10 0.0.0) url isCrossOrigin can identify cross origin urls FAILED
        //google.com from http://google.com is not cross origin
        Expected: true
        Actual: false
           at Anonymous function (/video.js/node_modules/qunitjs/qunit/qunit.js:2145:5)
           at Anonymous function (***/Temp/eaf8d6bff56c5e86c49b7c014c3af2eb.browserify:34838:3 <- ..\unit\utils\test\unit\utils\url.test.js:86:2)
           at runTest (/video.js/node_modules/qunitjs/qunit/qunit.js:903:4)
           at Test.prototype.run (/video.js/node_modules/qunitjs/qunit/qunit.js:888:4)
           at Anonymous function (/video.js/node_modules/qunitjs/qunit/qunit.js:1030:6)
           at process (/video.js/node_modules/qunitjs/qunit/qunit.js:691:4)

.........
IE 11.0.0 (Windows 10 0.0.0): Executed 285 of 285 (1 FAILED) (6.589 secs / 6.506 secs)
````

gkatsev:
Sounds like we should just remove that particular check then. I did try and make sure it worked across all browsers and it passes locally.
What version of IE11 are you using? I wonder if it's related to being Windows 10 as well.
We should try to fix it if possible but otherwise just remove that assertion because it's not worth it otherwise.

Naouak:
Karma launch IE and not Edge (don't really know why).
I got IE 11.63.10586.0 / 11.0.26 (KB3104002).

I'm not sure what the test is use for as it is a security enforced by the browser and the player shouldn't really have to check for it.

gkatsev:
It probably isn't launching edge because we don't have the karma edge launcher [installed](https://github.com/videojs/video.js/blob/084bcf692ced2848973fe8b0c4fad6d745231b4c/package.json#L73-L81). As for the security aspect, this is a test to make sure that the `isCorssOrigin` method is returning the correct results. It isn't testing whether the browser is enforcing cross origin. The way that the check is written, though, makes it hard to test in a completely browser-independent way since it relies on a browser feature to do the check.

gkatsev:
Also, you have a newer version of IE11 than I do, which may or may not be the issue.

gkatsev:
[...]

PR #4669, opened 2017-10-18
Fixing an issue where patching functions into Object.prototype makes the event removal crash.

MarcAMo:
## Description
I found a problem, where adding custom functions to Object.prototype made the event-off function crash upon iterating over all properties of an object.
I fixed it by checking for ```hasOwnProperty```. ~~As the tests failed for me on Win10 IE11 I also fixed issue #3100 that addressed this.~~

## Specific Changes proposed
~~I added an if-clause to url.js‘s ```parseUrl``` to add the current window.location.protocol if the protocol of the given url is empty (or ```//```).~~

~~I also added 2 if statements to events.js‘s ```off``` in the "type is falsy" if to check that data and data.handler is present (occasionally was undefined with me) and within the loop check if data.handler has t as it‘s own property~~.
I added an if statement to events.js‘s off in the ```if (!type)``` to check that data.handler is still present and has ```t``` as its own property while running the for loop. This way it won‘t try to remove monkey-patched properties.

To make the loop only run in the case of an omitted type parameter the surrounding if was corrected from ```(!type)``` to ```(type === undefined)```

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -351,11 +351,9 @@ export function off(elem, type, fn) {
 
   // Are we removing all bound events?
   if (!type) {
-    if (!!data && !!data.handlers) {
-      for (const t in data.handlers) {
-        if (!!data.handlers && !!data.handlers.hasOwnProperty && data.handlers.hasOwnProperty(t)) {
-          removeType(t);
-        }
+    for (const t in data.handlers) {
+      if (data.handlers) {
+        removeType(t);
       }
     }
     return;
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -351,8 +351,12 @@ export function off(elem, type, fn) {
 
   // Are we removing all bound events?
   if (!type) {
-    for (const t in data.handlers) {
-      removeType(t);
+    if(!!data && !!data.handlers) {
+      for (var t in data.handlers) {
+        if(!!data.handlers && !!data.handlers.hasOwnProperty && data.handlers.hasOwnProperty(t)) {
+          removeType(t);
+        }
+      } 
     }
     return;
   }
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -350,9 +350,9 @@ export function off(elem, type, fn) {
   };
 
   // Are we removing all bound events?
-  if (!type) {
+  if (type === undefined) {
     for (const t in data.handlers) {
-      if (data.handlers) {
+      if (Object.prototype.hasOwnProperty.call(data.handlers || {}, t)) {
         removeType(t);
       }
     }
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -351,12 +351,12 @@ export function off(elem, type, fn) {
 
   // Are we removing all bound events?
   if (!type) {
-    if(!!data && !!data.handlers) {
-      for (var t in data.handlers) {
-        if(!!data.handlers && !!data.handlers.hasOwnProperty && data.handlers.hasOwnProperty(t)) {
+    if (!!data && !!data.handlers) {
+      for (const t in data.handlers) {
+        if (!!data.handlers && !!data.handlers.hasOwnProperty && data.handlers.hasOwnProperty(t)) {
           removeType(t);
         }
-      } 
+      }
     }
     return;
   }
--- a/src/js/utils/url.js
+++ b/src/js/utils/url.js
@@ -85,10 +85,6 @@ export const parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
-  if (details.protocol === '') {
-    details.protocol = window.location.protocol;
-  }
-
   return details;
 };
 
--- a/src/js/utils/url.js
+++ b/src/js/utils/url.js
@@ -85,6 +85,10 @@ export const parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
+  if (details.protocol === '') {
+    details.protocol = window.location.protocol;
+  }
+
   return details;
 };
 


src/js/utils/events.js
```diff
@@ -351,8 +351,12 @@ export function off(elem, type, fn) {
 
   // Are we removing all bound events?
   if (!type) {
-    for (const t in data.handlers) {
-      removeType(t);
+    if (!!data && !!data.handlers) {
```
gkatsev: Why do this check here? If `data.handlers` doesn't exist, we should already have exited on line 338.

src/js/utils/events.js
```diff
@@ -351,8 +351,12 @@ export function off(elem, type, fn) {
...
+      for (const t in data.handlers) {
+        if (!!data.handlers && !!data.handlers.hasOwnProperty && data.handlers.hasOwnProperty(t)) {
```
gkatsev: does this mean that if `hasOwnProperty` is missing for whatever reason, we'll end up not actually removing the handler?

src/js/utils/url.js
```diff
@@ -85,6 +85,10 @@ export const parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
+  if (details.protocol === '') {
```
gkatsev: seems like this belongs in separate PR.

gkatsev:
Thanks for the PR! I made some comments.

src/js/utils/events.js
MarcAMo: You are right. It appears as if I was overly cautious when looking for the reason of ```data.handlers``` being ```undefined```.

src/js/utils/events.js
MarcAMo: I just went in and reconstructed the situation.
I called ```player.off("")``` to get into the ```if (!type)``` statement.
It had ```data.handlers``` filled up and in the loop removed one child after another. It seems like after the last own property is removed data.handler gets removed, as when it starts processing the prototype properties ```data``` actually is ```{}```. 

So in the end 
```JS
if (!type) {
  for (const t in data.handlers) {
    if (data.handlers) {
      removeType(t);
    }
  }
  return;
}
```
would totally do the trick.

src/js/utils/url.js
MarcAMo: You are right. This is a separate topic.
It's my first time committing to a bigger project with such a refined build tool chain and I did not want to (and I believe it didn't let me) push failing code. Deemed that more important than singling out a secondary fix which is closed due to inactivity anyway and wait for it to be accepted before committing this PR.

If you prefer I can revert this here and add it to a separate PR.

gkatsev: Thanks for contributing!
Making multiple PRs is generally better because then a single topic isn't blocking another topic that has been approved and the discussion can focus on the thing issue at hand.

src/js/utils/events.js
gkatsev: Is `player.off('')` what causing issues here? Are there other examples?

MarcAMo: I hit this problem at work in a more complex code base. I cannot exactly find the instance/code point that caused it.

```player.off('')``` was the most condensed trigger I could find to fit that code. It appears any falsy input would trigger the loop.

File: src/js/utils/url.js (hunk 3 above)
Comment (MarcAMo) [reply src/js/utils/url.js]: Ok. I will move it out from here.

**Comment by gkatsev** [comment]
So, I think I figured out the root cause of the issue. Video.js has a feature where on a component or the player if you call `off()` (like `player.off()`) without arguments, it will remove all bound events. However, an empty string is an argument but it's falsey, so, line 353 thinks that we didn't pass anything in and goes on to remove all the handlers. Seems like a proper solution would be to specifically check that `type` is set to `undefined`. So, something like `if (type === undefined)` instead of `if (!type)`. Thoughts?

**Comment by MarcAMo** [comment]
You are right in that ```if (undefined === type)``` (or other way around - your call) would be more precise than ```if (!type)```.

But that does not affect the problem I was aiming to fix. It still fails on ```off()``` as soon as there are custom properties in ```Object.prototype```.
So we still need 
```JS
for (const t in data.handlers) {
  if (data.handlers) {
    removeType(t);
  }
}
```

gkatsev:
Oh, I see.
Not exactly sure how the extra `data.handlers` check helps, though.
However, just doing an `if (Object.prototype.hasOwnProperty.call(data.handlers, t) { removeType(t); }` is probably best.
And if you can do the `type === undefined` check that would be great too.

MarcAMo:
I did the change locally and ```if (Object.prototype.hasOwnProperty.call(data.handlers, t))``` still fails if ```data.handlers``` is ```null``` or ```undefined```.

The two obvious solutions would be 
```JS
if (data.handlers && Object.prototype.hasOwnProperty.call(data.handlers, t))
``` 
and 
```JS
if (Object.prototype.hasOwnProperty.call(data.handlers || {}, t))
```

which one do you prefer?

gkatsev:
Probably the second one but I don't understand why `data.handlers` would ever be `null` at that point. Can you try and make a test case for this?

MarcAMo:
That would be, because line 30 of events.js deletes the current ```t``` from ```data.handlers``` and then line 43-47 deltes ```data.handlers```, ```data.dispatcher``` and ```data.disabled``` if ```data.handlers``` has no own properties left.

So once the last own property is removed ```_cleanUpEvents``` kills ```data.handlers```.

gkatsev:
Oh, I see. I was able to reproduce it locally. So, the issue here is really that `for..in` doesn't return only *own* enumerable properties. So, adding the `if (Object.prototype.hasOwnProperty(...))` works, another solution is to iterate over `Object.keys(data.handlers)` instead.
But this should be good to go then. Thanks for bearing with me while I try to understand the issue :)

MarcAMo:
No problem.
I ran into that problem a couple of times in different places in my own code and didn't think of explaining that part of the problem to you.
Sorry for that on my part.
Thanks for taking your time with me.

MarcAMo:
Are we still missing some step that I should perform or is anything else still off here?
Or is this simply sticking around as nobody took the time merging this, yet?

gkatsev:
Yup, we hadn't merged it yet. We'll be merging things in this week :)

PR #4671, opened 2017-10-18
feat: Add a noConflict method.

misteroneill:
## Description
Add a `noConflict` method, like jQuery offers. This can help in those rare cases where a user embeds multiple copies of Video.js of different versions on the same page.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -132,6 +132,23 @@ function videojs(id, options, ready) {
   return player;
 }
 
+// Cache the previous videojs global in case it's overwritten.
+const previousVideojs = window.videojs;
+
+/**
+ * Removes this `videojs` global from the global namespace and returns it, so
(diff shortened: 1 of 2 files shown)


misteroneill:
I guess we could say Flash will not work properly if you use `noConflict` and warn if we see someone using both?

misteroneill:
Alternatively, we could check for `window.videojs.Flash` and ensure it never goes away even if `noConfict` is used.

gkatsev:
Yeah, it should warn or throw or no-op if Flash is being used. Not exactly sure what should happen. We should probably have a decision before we merge this in.

gkatsev:
I'm going to close this while it's on-hold, we can re-open at a later date when we revisit the no-conflict functionality.

PR #4672, opened 2017-10-18
feat: Add videojs.hookOnce method to allow single-run hooks.

misteroneill:
## Description
This adds a `videojs.hookOnce()` method, which allows single-run hooks.

Also, cleaned up the guide for readability and grammar/spelling/typos.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -140,8 +140,8 @@ videojs.hooks_ = {};
  * @param {string} type
  *        the lifecyle to get hooks from
  *
- * @param {Function} [fn]
- *        Optionally add a hook to the lifecycle that your are getting.
+ * @param {Function|Function[]} [fn]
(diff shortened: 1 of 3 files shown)


src/js/video.js
```diff
@@ -140,8 +140,8 @@ videojs.hooks_ = {};
  * @param {string} type
  *        the lifecyle to get hooks from
  *
- * @param {Function} [fn]
- *        Optionally add a hook to the lifecycle that your are getting.
+ * @param {Function|Function[]} [fn]
```
gkatsev: I don't think we should document or officially support multiple handlers in a single hooks call. It just encourages bad code.

misteroneill: It's already documented for the other function. Was trying to be consistent.

gkatsev: So it is. Carry on.

PR #4673, opened 2017-10-18
Bugfix for issue #3100

MarcAMo:
## Description
url.js ```parseUrl``` drops the protocol from the return object on IE if the url starts with "//:". Chrome returns the current ```window.location.protocol``` value in that case. This causes the test mentioned in #3100 to fail on IE, but not on Chrome or FF.

## Specific Changes proposed
I added an if-clause to url.js‘s ```parseUrl``` to add the current window.location.protocol if the protocol of the given url is empty (or ```//```).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/url.js
+++ b/src/js/utils/url.js
@@ -81,12 +81,12 @@ export const parseUrl = function(url) {
     details.host = details.host.replace(/:443$/, '');
   }
 
-  if (addToBody) {
-    document.body.removeChild(div);
+  if (!details.protocol) {
+    details.protocol = window.location.protocol;
   }
 
-  if (details.protocol === '') {
-    details.protocol = window.location.protocol;
+  if (addToBody) {
+    document.body.removeChild(div);
   }
 
   return details;
--- a/src/js/utils/url.js
+++ b/src/js/utils/url.js
@@ -85,6 +85,10 @@ export const parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
+  if (details.protocol === '') {
+    details.protocol = window.location.protocol;
+  }
+
   return details;
 };
 


gkatsev:
This is probably fine. I'll have to take a look and see if there might be any issues with this.

src/js/utils/url.js
```diff
@@ -85,6 +85,10 @@ export const parseUrl = function(url) {
     document.body.removeChild(div);
   }
 
+  if (details.protocol === '') {
+    details.protocol = window.location.protocol;
+  }
+
```
misteroneill: Overally, seems like a good change/fix. I wonder if it ought to be simply testing `!details.protocol` instead?

Also, maybe it should be moved above the `if (addToBody) {}` block to keep the `details.protocol` conditions together?

MarcAMo: There should be no protocol that is as well legal as falsy. So ```!details.protocol``` sounds good.

Do you refer only to readability with your second suggestion or should there be a functional difference I don't see?

I guess you see it still below the addToBody declaration, don't you?

misteroneill: Yep, exactly. That'd be great, thanks!

MarcAMo:
Here you are. Thanks for the constructive feedback!

MarcAMo:
Did I miss some needed action here? As far as I understood everything is fine. Is "needs:testing" still applicable? At the end of the day this fixed the existing tests' failing. So I don't think that ```Unit Tests updated or fixed```,  ```Docs/guides updated``` or ```Example created (starter template on JSBin)``` is applicable, is it?

Or are we simply waiting for a 2nd core contributor approval?

PR #4680, opened 2017-10-22
refactor player.hasStarted()

kocoten1992:
update doc on hasStarted()

I feel 'request' are more suitable than hasStarted (as a variable) because when I do "if (request === undefined)", it mean if you don't have any request, then we just return current status, it a bit more clear

## Requirements Checklist
- [x] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1139,7 +1139,7 @@ class Player extends Component {
       // act as getter, if we have no request to change
       return this.hasStarted_;
     }
-    
+
     if (request === this.hasStarted_) {
       return;
     }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1135,8 +1135,13 @@ class Player extends Component {
    *         the boolean value of hasStarted_
    */
   hasStarted(request) {
-    if (request === undefined || request === this.hasStarted_) {
+    if (request === undefined) {
+      // act as getter, if we have no request to change
       return this.hasStarted_;
+    }
+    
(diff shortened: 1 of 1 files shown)


kocoten1992:
Hi @gkatsev , does this style of writing clearer than the previous one.

Cause I'm running through the source and see a lot of js write at old style, if this pull request accepted, I might refactor some more ?

src/js/player.js
```diff
@@ -1124,29 +1127,28 @@ class Player extends Component {
...
-          this.trigger('firstplay');
-        } else {
-          this.removeClass('vjs-has-started');
-        }
-      }
-      return;
+   *         the boolean value of hasStarted_
+   */
+  hasStarted(request) {
+    if (request === undefined || request === this.hasStarted_) {
```
gkatsev: A comment explaining that this is used as a getter or if the value isn't changing, we should just return what we have would be good

src/js/player.js
```diff
@@ -1124,29 +1127,28 @@ class Player extends Component {
...
+    this.hasStarted_ = request;
+
+    if (this.hasStarted_) {
+      this.addClass('vjs-has-started');
+      this.trigger('firstplay');
+    } else {
+      this.removeClass('vjs-has-started');
+    }
+
+    return this.hasStarted_;
```
gkatsev: Our setters generally do not return anything.

misteroneill: As @gkatsev said, this is my only concern with the PR.

PR #4693, opened 2017-10-26
refactor component.ready()

kocoten1992:
### Requirements Checklist
- [x] refactor component.ready()

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -598,18 +598,21 @@ class Component {
    *         Returns itself; method can be chained.
    */
   ready(fn, sync = false) {
-    if (fn) {
-      if (this.isReady_) {
-        if (sync) {
-          fn.call(this);
-        } else {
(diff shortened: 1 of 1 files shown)


src/js/component.js
```diff
@@ -598,18 +598,21 @@ class Component {
...
+    }
+
+    if (!this.isReady_) {
+      this.readyQueue_ = this.readyQueue_ || [];
+      this.readyQueue_.push(fn);
+      return;
+    }
+
+    if (sync) {
+      fn.call(this);
```
BrandonOCasey: You could return after this and get rid of the `else` here, I think that it would fit better with the new flow of this function.

kocoten1992: Uhm no, the point was to easy read/debug, abusing that style in only 4 line of code would make thing harder

But feel free to edit anyway

Issue #4697, opened 2017-10-27
Log a warning if the element given to Video.js isn't in the DOM


gkatsev:
## Description
Video.js can accept a video element (or soon a `video-js` element #4640) as the first argument and it will build the player around the video element or inside that div. However, if the element given to Video.js isn't actually in the DOM nothing will happen and the developer could end up confused as to what is happening.
We should log a warning if the element that is passed to the Video.js method isn't currently included in the DOM.

More information on reproducing is below but to fix it, it will basically involve doing a call to [`document.body.contains()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains), which has full browser support, with the element that we received in `videojs` around [here](https://github.com/videojs/video.js/blob/master/src/js/video.js#L96-L100). Then we can log a warning with `videojs.log.warn()`.

*If you'd like to work on this, please let us know and please don't work on this if someone else has already claimed it!*
 
## Steps to reproduce
1. Create a video element in JavaScript
2. Pass that video element to Video.js
3. Notice that the page is still blank

```js
// This is the *broken* scenario
var vid = document.createElement('video');
var player = videojs(vid, {});
// We should see a warning logged here
```
```js
// This scenario should work and not lead to a warning log
var vid = document.createElement('video');
document.body.appendChild(vid);
var player = videojs(vid, {});
```

### Expected
If we are passed an element that is not in the DOM, we should log a warning

### Actual
Nothing currently happens, which could be confusing.

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
Any
#### browsers
Any
#### OSes
Any
### plugins
None or any

odisei369:
Very nice description. I'll do this.

gkatsev:
Merged the initial PR for the feature but it would be good to add a test to it. Info regarding the test can be found here: https://github.com/videojs/video.js/pull/4698#issuecomment-340815154

WaliHassanKhan:
Allow me to contribute in it

gkatsev:
Oops, sorry, we forgot to close this when the change was merge in. Unfortunately, it's already been done @WaliHassanKhan. thanks for your interest, I hope to have more of these in the future.

WaliHassanKhan:
That's ok

WilliamTsao:
When working with shadowDOM, this warning is thrown because document.body.contains cannot find element within shadowRoot.

brendon:
[...]

PR #4698, opened 2017-10-28
add warning, if the element given to Video.js is not in the DOM

odisei369:
## Description
Issue #4697 

## Specific Changes proposed
Add warning, if the element given to Video.js is not in the DOM

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -99,6 +99,11 @@ function videojs(id, options, ready) {
     throw new TypeError('The element or ID supplied is not valid. (videojs)');
   }
 
+  // Check if element is included in the DOM
+  if (Dom.isEl(tag) && !document.body.contains(tag)) {
+    log.warn('The element supplied is\'t included in the DOM');
+  }
+
   // Element may have a player attr referring to an already created player instance.
   // If so return that otherwise set up a new player below
   if (tag.player || Player.players[tag.playerId]) {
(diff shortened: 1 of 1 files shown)


src/js/video.js
```diff
@@ -99,6 +99,11 @@ function videojs(id, options, ready) {
     throw new TypeError('The element or ID supplied is not valid. (videojs)');
   }
 
+  // Check if element is included in the DOM
+  if (Dom.isEl(tag) && !document.body.contains(tag)) {
+    log.warn('The element supplied is\'t included in the DOM');
```
gkatsev: This works, though, we tend to switch to double quotes (`"`) if we're including single quotes (`'`) in the text, that way we don't need to escape it :)

odisei369: @gkatsev Thank you. Of course it would look better.
Now I also can see misprint in "isN't".
What should I do with that? Make another commit?

mister-ben: If you push another commit to your branch the PR will update.

gkatsev:
@odisei369 hey, would you be up for adding a test for this? You can use [this test](https://github.com/videojs/video.js/blob/master/test/unit/video.test.js#L56-L92) as an example. Basically, it'll involve mocking `log.warn`, creating a new player with a random video element, asserting that `log.warn` was called with the value we want, and then disposing the player we created. If you have questions or concerns, please let me know.

odisei369:
@gkarsev OK, I'll add the test. With example it would be easy.

gkatsev:
@odisei369 thanks! I just merged this PR, so, please make another one when you do :)

gkatsev:
And again, if you have questions, feel free to post them there or stop by and ask on slack http://slack.videojs.com :)

PR #4699, opened 2017-10-28
refactor options.language (player.constructor)

kocoten1992:
move down this line, so we don't have to think too much while in the loop
```js
options.language = element.getAttribute('lang');
```

#### Requirements Checklist
- [x] refactor options.language (player.constructor)


kocoten1992:
Oops, read it all wrong !!

PR #4700, opened 2017-10-28
refactor options.language (player.constructor)

kocoten1992:
move line this down to reduce thinking while in the loop
```js
options.language = element.getAttribute('lang');
```

#### Requirements Checklist
- [x] refactor options.language (player.constructor)

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -290,22 +290,23 @@ class Player extends Component {
 
     // If language is not set, get the closest lang attribute
     if (!options.language) {
-      if (typeof tag.closest === 'function') {
-        const closest = tag.closest('[lang]');
 
-        if (closest) {
-          options.language = closest.getAttribute('lang');
-        }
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -290,22 +290,23 @@ class Player extends Component {
 
     // If language is not set, get the closest lang attribute
     if (!options.language) {
-      if (typeof tag.closest === 'function') {
-        const closest = tag.closest('[lang]');
```
gkatsev: I think here particularly, we should avoid calling `closest` twice, given that it's a DOM operation.

kocoten1992: This is a hard one, no early return allowed, but about DOM operation, I think it only slow when involve with render html (this case is only compare).

I'll close this, been thinking but there is no good solution in this case, even without performance, 2 if is bit gross to write.

PR #4708, opened 2017-10-31
fix: aria-labelledby attribute has an extra space

### LINKED ISSUE #4688: Video.js aria labelledby and ids contain spaces
## Description
There's seems to be odd spaces between aria labels in the video-js outputted markup inside the vjs track setting components.

## Steps to reproduce
I've created a codepen in which demonstrates this issue
https://codepen.io/HannahG/pen/xXovgx

Annoyingly it's in an iframe but if you do a search for "captions-font-size-video_component_373" in dev tools you can see the issue. You get the below:
` aria-labelledby=" captions-font-size-video_component_373"`

This is an issue with the current build 6.2.8 and fails accessibility on the aria-labels.

Is there anything we can do to fix this?

## Additional Information
We are using this for a government digital service and need AAA Accessibility if possible.
#### videojs
6.2.8
#### browsers
Chrome Version 61.0.3163.100 (Official Build) (64-bit)
#### OSes
Windows 
### plugins
none

knilob:
## Description
The `aria-labelledby` attribute on the `fontPercent`, `edgeStyle`, and
`fontFamily` select options includes an extra space since there is
no `ledgendId` variable being set on the `createElFont_()` method.

[Fixes #4688](https://github.com/videojs/video.js/issues/4688)

## Specific Changes proposed
This fix adds a check to see if the `legendId` value is set or not inside
the `createElSelect_()` method. This should keep the extra space
from appearing on the select tags created by the `createElFont_()`
method.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -312,14 +312,14 @@ class TextTrackSettings extends ModalDialog {
       `<${type} id="${id}" class="${type === 'label' ? 'vjs-label' : ''}">`,
       this.localize(config.label),
       `</${type}>`,
-      `<select aria-labelledby="${legendId} ${id}">`
+      `<select aria-labelledby="${legendId !== '' ? legendId + ' ' : ''}${id}">`
     ].
       concat(config.options.map(o => {
         const optionId = id + '-' + o[1];
 
         return [
           `<option id="${optionId}" value="${o[0]}" `,
-          `aria-labelledby="${legendId} ${id} ${optionId}">`,
+          `aria-labelledby="${legendId !== '' ? legendId + ' ' : ''}${id} ${optionId}">`,
           this.localize(o[1]),
           '</option>'
         ].join('');


OwenEdwards:
@knilob did you see any symptom of this extra space? Was a screen reader (JAWS, NVDA, VoiceOver, TalkBack) having issues with it? I'm just wondering because, like CSS classes, aria-labelledby takes a space-separated list of values (in this case IDs), so it really *shouldn't* have any impact. If it does, it'd be worth knowing exactly what and where. Thanks!

gkatsev:
@OwenEdwards see #4688, seems like it's being flagged by the WAVE tool from webaim

gkatsev:
Also, looks like we have a bug in the generation of the aria-labelledby for the font families since the values themselves have spaces when we create the idea for it, it ends up messing up. For example, we end up with an id `captions-font-family-vid1_component_423-Proportional Sans-Serif` instead of something like `captions-font-family-vid1_component_423-Proportional-Sans-Serif` and then when we reference it it's wrong.

OwenEdwards:
@gkatsev seems like a component ought to have a general-purpose toID() method (or maybe it should be part of DOM) that only allows valid ID labels to be created; anything that's not a-z, A-Z, 0-9, - or _ gets converted (technically a period is allowed too, but it gets tricky for CSS selectors).

I'm not convinced that having a space at the start or end of aria-labelledby (or aria-describedby) is invalid, but if WAVE is flagging it, it seems like any time an aria-labelledby is created it could just be `trim()`'ed.

gkatsev:
Yeah, I don't think this causes issues and making the code remove it isn't a big burden. `trim()` is probably better and now that I noticed we need a slightly better refactor, we probably should switch to that.

PR #4711, opened 2017-11-01
fix: default cached src to empty string

forbesjo:
## Description
`player.src()` returns `undefined` for a new player.

## Specific Changes proposed
Default to `''` if there is no source cached.

## Requirements Checklist


gkatsev:
@BrandonOCasey yup, plus, we already do that for `currentSrc()`.

PR #4714, opened 2017-11-02
feat: add a version function to all advanced plugin instances

BrandonOCasey:
## Description
Having to use videojs.getPluginVersion(<pluginInstance>.constructor.name) is a bit weird when you already have an instance. I would even say we should get rid of the getPluginVersion function or deprecate it.

## Requirements Checklist

### DIFF
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -220,6 +220,13 @@ class Plugin {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


gkatsev:
Oh, totally forgot about that part in the main comment. And yeah, this should be purely additive definitely usecase for the method as @misteroneill said.

BrandonOCasey:
makes sense

PR #4716, opened 2017-11-03
refactor player.userActive()

kocoten1992:
#### Requirements Checklist
- [x] refactor player.userActive()

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2781,6 +2781,8 @@ class Player extends Component {
       return this.userActive_;
     }
 
+    bool = !!bool;
+
     if (bool === this.userActive_) {
       return;
     }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -318,6 +318,9 @@ class Player extends Component {
     // Init state hasStarted_
     this.hasStarted_ = false;
 
+    // Init state userActive_
+    this.userActive_ = false;
+
     // if the global option object was accidentally blown away by
     // someone, bail early with an informative error
     if (!this.options_ ||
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -2774,53 +2777,43 @@ class Player extends Component {
...
-           * @type {EventTarget~Event}
-           */
-          this.trigger('userinactive');
-        }
-      }
+    if (bool === this.userActive_) {
+      return;
+    }
+
+    this.userActive_ = bool;
```
gkatsev: we should cast this to a boolean like previously to make sure that things don't break silently
`!!bool`

src/js/player.js
```diff
@@ -2774,53 +2777,45 @@ class Player extends Component {
...
-        if (bool) {
-          // If the user was inactive and is now active we want to reset the
-          // inactivity timer
-          this.userActivity_ = true;
-          this.removeClass('vjs-user-inactive');
-          this.addClass('vjs-user-active');
-          /**
-           * @event Player#useractive
-           * @type {EventTarget~Event}
-           */
```
misteroneill: We need to preserve these JSDoc blocks.

src/js/player.js
```diff
@@ -2774,53 +2777,45 @@ class Player extends Component {
...
+    if (bool === undefined) {
+      return this.userActive_;
+    }
 
-          this.removeClass('vjs-user-active');
-          this.addClass('vjs-user-inactive');
-          /**
-           * @event Player#userinactive
-           * @type {EventTarget~Event}
-           */
```
misteroneill: We need to preserve these JSDoc blocks.

PR #4717, opened 2017-11-03
Feature: Use `sourceset` to update the current source cache

BrandonOCasey:
## Description
Logic to update the source on sourceset

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -569,6 +569,9 @@ class Player extends Component {
       el.appendChild(tag);
 
       playerElIngest = this.playerElIngest_ = el;
+      Object.keys(el).forEach((k) => {
+        tag[k] = el[k];
+      });
     }
 
     // set tabindex to -1 so we could focus on the player element
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -31,6 +31,7 @@ import Tech from './tech/tech.js';
 import * as middleware from './tech/middleware.js';
 import {ALL as TRACK_TYPES} from './tracks/track-types';
 import filterSource from './utils/filter-source';
+import getMimeType from './utils/get-mime-type';
 
 // The following imports are used only to ensure that the corresponding modules
 // are always included in the video.js package. Importing the modules will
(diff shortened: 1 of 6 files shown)


src/js/player.js
```diff
@@ -823,6 +825,213 @@ class Player extends Component {
...
+            type = source.type;
+            break;
+          }
+        }
+      }
+
+      // otherwise we set the type to unknown,
+      // or in the case where we know the type based on the file extension
+      // we use that
+      if (!type) {
```
BrandonOCasey: As mentioned in the previous PR we need a better way to find the type here. Here are some of the questions I have:

* Should we reference a list of common source types and make it `unknown` when we don't know the type?
* Should we only update player.src?
* Should we use a `HEAD` request to get the `content-type` header for the source?

gkatsev:
we probably should rebase this against master at this point.
Also, my main concern would be this overriding the current cache work that does the right thing with types with the value from sourceset, even when the video.js API was used.

BrandonOCasey:
TODO: rebase, test with `video-js` el

src/js/player.js
```diff
@@ -568,6 +569,9 @@ class Player extends Component {
       el.appendChild(tag);
 
       playerElIngest = this.playerElIngest_ = el;
+      Object.keys(el).forEach((k) => {
+        tag[k] = el[k];
```
BrandonOCasey: Not sure if we want this or not, but it seems like we did. This brings properties that are set in js from the video-js object over to the video object. In my case I was using `mediaEl.src = ...;` (`mediaEl` being the `video-js` element) in js. Then initializing the player. This would cause the `src` property to be lost which I do not think is correct.

src/js/player.js
```diff
@@ -2435,10 +2557,12 @@ class Player extends Component {
     }
 
     // intial sources
+    this.cache_.previousSources = this.cache_.sources;
```
BrandonOCasey: We have to store `previousSource` and `previousSources` due to source setting being async in here. Without it, it was possible for the type of the source to be lost. I think we may eventually want to remove the source cache from here and only update it on `sourceset`.

```
player.src('src1.mp4');
player.src('src2.mp4');
```

src/js/tech/html5.js
```diff
@@ -172,7 +172,9 @@ class Html5 extends Tech {
       set: (v) => {
         const retval = srcDescriptor.set.call(el, v);
 
-        this.triggerSourceset(v);
+        // we use the getter here to get the actual value set on
+        // src
+        this.triggerSourceset(el.src);
```
BrandonOCasey: This is important for relative sources

src/js/tech/html5.js
```diff
@@ -186,7 +188,7 @@ class Html5 extends Tech {
       const retval = oldSetAttribute.call(el, n, v);
 
       if (n === 'src') {
-        this.triggerSourceset(v);
+        this.triggerSourceset(el.getAttribute('src'));
```
BrandonOCasey: This is important for relative sources

src/js/tech/html5.js
```diff
@@ -202,7 +204,30 @@ class Html5 extends Tech {
       // source elements will be used but implementing the source selection algorithm
       // is laborious and asynchronous, so,
       // instead return an empty string to basically indicate source may change
-      this.triggerSourceset(el.src || '');
+      let src = el.src || '';
+
+      // if their is only one source
```
BrandonOCasey: Realized that we can actually know what `<source>` it will choose if there isn't a `src` attribute, and there is only one non-duplicate `<source>` url.

BrandonOCasey:
closing in favor of https://github.com/videojs/video.js/pull/5040

PR #4719, opened 2017-11-04
refactor: player.listenForUserActivity_()

kocoten1992:
### Description
Change setTimeout min to 4 as https://www.w3.org/TR/2011/WD-html5-20110525/timers.html#timers

#### Requirements Checklist
- [x] refactor: player.listenForUserActivity_()

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2881,31 +2881,38 @@ class Player extends Component {
 
     this.setInterval(function() {
       // Check to see if mouse/touch activity has happened
-      if (this.userActivity_) {
-        // Reset the activity tracker
-        this.userActivity_ = false;
-
-        // If the user state was inactive, set the state to active
-        this.userActive(true);
-
-        // Clear any existing inactivity timeout to start the timer over
-        this.clearTimeout(inactivityTimeout);
-
-        const timeout = this.options_.inactivityTimeout;
-
-        if (timeout > 0) {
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -2881,31 +2881,38 @@ class Player extends Component {
...
+      this.userActive(true);
+
+      // Clear any existing inactivity timeout to start the timer over
+      this.clearTimeout(inactivityTimeout);
+
+      const timeout = this.options_.inactivityTimeout;
+
+      // Min time for setTimeout is 4ms regardless any less number
+      // https://www.w3.org/TR/2011/WD-html5-20110525/timers.html#timers
+      if (timeout <= 4) {
```
gkatsev: I think we want to keep the `0` check here because while the timer does resolve to 4ms, the value `0` has the special meaning of stopping the `inactivityTimeout`. A number that isn't zero means that you want a timeout. I think it would be confusing if you set the `inactivityTimeout` to 2 and then have it not function.

kocoten1992: Hi, does that mean ```timeout === 0``` good enough or it should be at normal ```timeout < 0```, I'm in flavor the first if just 0 if treat it as special

gkatsev: unfortunately, we should keep the same check because it could break people. We can correct this in a major update.

PR #4737, opened 2017-11-11
Update fa.json

EhsanCh:
## Description
Optimize translate


lang/fa.json
```diff
@@ -1,26 +1,26 @@
...
+  "Non-Fullscreen": "غیر تمام‌صفحه",
   "Mute": "بی صدا",
-  "Unmute": "بهمراه صدا",
+  "Unmute": "با صدا",
   "Playback Rate": "سرعت پخش",
   "Subtitles": "زیرنویس",
   "subtitles off": "بدون زیرنویس",
-  "Captions": "عنوان",
+  "Captions": "عنوان‌ها",
   "captions off": "بدون عنوان",
```
mister-ben: Should this also change like "captions"?

EhsanCh: Yes i think. i will review again and add all strings and will submit it again.
thanks.

PR #4743, opened 2017-11-13
fix: Wait for loadstart when changing sources, instead of just ready.

misteroneill:
The core goal here is to make sure the following works in light of some middleware process that makes setting the source more async than next tick:

```js
player.src('...');
player.ready(() => player.play());
```

In fact, given this change, we should even be able to do:

```js
player.src('...');
player.play();
```

Unlike #4665, which would have clarified/changed the meaning of "ready", it remains a reflection of the tech's state and we make better use of the ability to queue things on that state and on the middleware `setSource` process.

## Requirements Checklist

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -3,6 +3,7 @@
  */
 import Button from './button.js';
 import Component from './component.js';
+import {isPromise} from './utils/promise';
 
 /**
  * The initial play button that shows before the video has played. The hiding of the
@@ -58,10 +59,8 @@ class BigPlayButton extends Button {
 
     const playFocus = () => playToggle.focus();
 
-    if (playPromise && playPromise.then) {
-      const ignoreRejectedPlayPromise = () => {};
-
-      playPromise.then(playFocus, ignoreRejectedPlayPromise);
+    if (isPromise(playPromise)) {
+      playPromise.then(playFocus, () => {});
     } else {
       this.setTimeout(playFocus, 1);
     }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -24,6 +24,7 @@ import MediaError from './media-error.js';
 import safeParseTuple from 'safe-json-parse/tuple';
 import {assign} from './utils/obj';
 import mergeOptions from './utils/merge-options.js';
+import {silencePromise} from './utils/promise';
 import textTrackConverter from './tracks/text-track-list-converter.js';
 import ModalDialog from './modal-dialog';
 import Tech from './tech/tech.js';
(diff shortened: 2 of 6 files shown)


src/js/player.js
```diff
@@ -1623,37 +1624,31 @@ class Player extends Component {
...
 
-        // silence errors (unhandled promise from play)
-        if (retval !== undefined && typeof retval.then === 'function') {
-          retval.then(null, (e) => {});
-        }
-      });
-
-    // Only calls the tech's play if we already have a src loaded
-    } else if (this.isReady_ && (this.src() || this.currentSrc())) {
+    // If the tech is not ready, queue up another call to `play()` for when it
```
misteroneill: The key change in all this is that we kind of have a "checklist" of conditions before we can attempt playback.

First, we need the tech to be ready before anything else. When that's satisfied, `play()` gets called again.

Second, we need to make sure middleware isn't changing sources. When that's satisfied by a `loadstart`, we call `play()` again.

Finally, both conditions should be satisfied, so we can attempt playback if there's a source.

src/js/player.js
```diff
@@ -1623,37 +1624,51 @@ class Player extends Component {
...
+    if (!this.isReady_) {
+
+      // Bail out if we're already waiting for `ready`!
+      if (this.playWaitingForReady_) {
+        return;
+      }
+
+      this.playWaitingForReady_ = true;
+      this.ready(() => {
+        this.playWaitingForReady_ = false;
```
gkatsev: can we set this in the constructor to a default value as well?

src/js/player.js
```diff
@@ -1623,37 +1624,51 @@ class Player extends Component {
...
+      this.playWaitingForLoadstart_ = true;
+      this.tech_.one('loadstart', () => {
+        this.playWaitingForLoadstart_ = false;
+        silencePromise(this.play());
       });
+
+    // Finally, if we have a source, we can attempt playback for real.
+    } else if (this.src() || this.currentSrc()) {
+      return this.techGet_('play');
     }
```
gkatsev: is it possible that we end up in an `else` condition?

src/js/big-play-button.js
```diff
@@ -58,10 +59,8 @@ class BigPlayButton extends Button {
 
     const playFocus = () => playToggle.focus();
 
-    if (playPromise && playPromise.then) {
-      const ignoreRejectedPlayPromise = () => {};
-
-      playPromise.then(playFocus, ignoreRejectedPlayPromise);
+    if (isPromise(playPromise)) {
+      playPromise.then(playFocus, () => {});
```
gkatsev: why not import `silencePromise` and use that?

misteroneill: Copy/paste from my other PR, good call.

src/js/player.js
misteroneill: I've been trying to think of one that isn't captured by the previous two statements, but I haven't been able to. It probably makes sense to reverse this conditional block so the `ready` call comes in the `else` block.

misteroneill: Actually, yeah, the only condition that we'd have that would be an else would be if the tech is ready, we're not changing sources, and we don't have a source.

src/js/player.js
misteroneill: Done.

src/js/big-play-button.js
misteroneill: Oh, actually, because it's not silencing it.

gkatsev: It isn't? What the difference between `ignoreRejectedPlayPromise` and `silencePromise`?

Oh, I see. We are silencing the playPromise but we're also adding a `playFocus` handler.

gkatsev: we could potentially do something like:
```js
if (isPromise(playPromise)) {
  silencePromise(playPromise);
  playPromise.then(playFocus);
}
```

misteroneill: Seems like overkill...

gkatsev: yeah, either way. using `silencePromise` is probably extra explicit/clear but either is fine.

test/unit/utils/promise.test.js
```diff
@@ -0,0 +1,19 @@
...
+  if (!window.Promise) {
+    assert.expect(0);
+    return;
+  }
+
+  assert.ok(promise.isPromise(new window.Promise((resolve) => resolve())));
+});
+
+QUnit.test('can identify a Promise-like object', function(assert) {
+  assert.notOk(promise.isPromise({}));
```
gkatsev: assertion messages?

test/unit/player-play.test.js
```diff
@@ -0,0 +1,79 @@
+/* eslint-env qunit */
+import sinon from 'sinon';
+import {silencePromise} from '../../src/js/utils/promise';
+import TestHelpers from './test-helpers';
+
+QUnit.module('Player#play', {
+
+  beforeEach() {
+    this.clock = sinon.useFakeTimers();
+    this.player = TestHelpers.makePlayer({techOrder: ['html5']});
```
gkatsev: if the html5 tech is required for testing, we'd want to wrap this file in a `!IS_IE8`.

misteroneill: I don't think it actually is. It can probably use the fake tech.

test/unit/utils/promise.test.js
misteroneill: I guess... 😛

PR #4745, opened 2017-11-13
perf: null out els on dispose to minimize detached els

gkatsev:
## Description
We have a bunch of references to DOM elements we were retaining even after we had disposed the player. This cleans up a bunch of them.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -148,6 +148,9 @@ class Component {
       DomData.removeData(this.el_);
       this.el_ = null;
     }
+
+    // remove reference to the player after disposing of the element
+    this.player_ = null;
   }
 
   /**
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -84,8 +84,11 @@ class Component {
       this.el_ = this.createEl();
     }
 
-    // Make this an evented object and use `el_`, if available, as its event bus
-    evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+    // if evented is anything except false, we want to mixin in evented
+    if (options.evented !== false) {
+      // Make this an evented object and use `el_`, if available, as its event bus
+      evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+    }
     stateful(this, this.constructor.defaultState);
 
     this.children_ = [];
--- a/src/js/control-bar/text-track-controls/text-track-menu-item.js
+++ b/src/js/control-bar/text-track-controls/text-track-menu-item.js
@@ -33,8 +33,12 @@ class TextTrackMenuItem extends MenuItem {
     super(player, options);
 
     this.track = track;
-    const changeHandler = (...args) => this.handleTracksChange.apply(this, args);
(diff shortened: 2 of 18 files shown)


gkatsev:
Yeah, I think that's a bit too magic and will just mean that if we do something else we will think we're safe when we're not.

src/js/component.js
```diff
@@ -84,8 +84,11 @@ class Component {
       this.el_ = this.createEl();
     }
 
-    // Make this an evented object and use `el_`, if available, as its event bus
-    evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+    // if evented is anything except false, we want to mixin in evented
+    if (options.evented !== false) {
```
gkatsev: this PR is still a perf but adds this feature to disable auto `evented` mixin.

src/js/mixins/evented.js
```diff
@@ -366,7 +367,12 @@ function evented(target, options = {}) {
   Obj.assign(target, EventedMixin);
 
   // When any evented object is disposed, it removes all its listeners.
-  target.on('dispose', () => target.off());
+  target.on('dispose', () => {
+    target.off();
+    window.setTimeout(() => {
+      target.eventBusEl_ = null;
```
gkatsev: we want to null out this reference so that the reference isn't retained when a component is disposed.

src/js/utils/events.js
```diff
@@ -344,16 +344,16 @@ export function off(elem, type, fn) {
   }
 
   // Utility function
-  const removeType = function(t) {
+  const removeType = function(el, t) {
     data.handlers[t] = [];
-    _cleanUpEvents(elem, t);
```
gkatsev: makes it more likely that `elem` isn't accidentally kept in the closure.

gkatsev:
Looks like the latest set of changes broke some tests.

gkatsev:
IE8 tests can actually finish! even if they might run out of memory on a run, it'll be fine the next run. Will look at the broken tests soon.

src/js/control-bar/text-track-controls/text-track-menu-item.js
```diff
@@ -34,13 +33,14 @@ class TextTrackMenuItem extends MenuItem {
     super(player, options);
 
     this.track = track;
-    const changeHandler = Fn.bind(this, this.handleTracksChange);
-    const selectedLanguageChangeHandler = Fn.bind(this, this.handleSelectedLanguageChange);
+    const changeHandler = (...args) => this.handleTracksChange.apply(this, args);
```
gkatsev: these two methods were binding the prototype methods so if any TextTrackMenuItem was disposed, all the event handlers were removed.

gkatsev:
YES. All tests passed and with just one IE8 run!

src/js/component.js
misteroneill: Good idea. 👍

PR #4752, opened 2017-11-16
test: cleanup test warnings

BrandonOCasey:
## Description
Currently we get logs in the tests due to a recent change. This PR fixes that. It also fixes a case were we would log that the dom element isn't in the dom, when it really is:

Here is an example (taken from the test that showed me it)
```js
const fixture = document.getElementById('qunit-fixture');

fixture.innerHTML += '<video id="test_vid_id"></video>';

const tag = document.getElementById('test_vid_id');

const player = videojs(tag)

// this would log a warning about tag not being in the dom. Technically its not because
// of the changes we made to it. I would even say that this is a bad way to get the player again.
// but I think that warning that the players tag is not in the dom on the first creation is enough, and // warning that it isn't in the dom because the tag has been changed is a bit weird.

// returns the same player as above.
playerAgain(tag);

```

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -100,15 +100,15 @@ function videojs(id, options, ready) {
     throw new TypeError('The element or ID supplied is not valid. (videojs)');
   }
 
-  // Check if element is included in the DOM
-  if (Dom.isEl(tag) && !document.body.contains(tag)) {
-    log.warn('The element supplied is not included in the DOM');
-  }
-
   // Element may have a player attr referring to an already created player instance.
   // If so return that otherwise set up a new player below
   if (tag.player || Player.players[tag.playerId]) {
     return tag.player || Player.players[tag.playerId];
+  }
+
+  // Check if element is included in the DOM
+  if (Dom.isEl(tag) && !document.body.contains(tag)) {
+    log.warn('The element supplied is not included in the DOM');
   }
 
   options = options || {};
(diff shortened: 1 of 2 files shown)


src/js/video.js
```diff
@@ -100,17 +100,17 @@ function videojs(id, options, ready) {
...
-    log.warn('The element supplied is not included in the DOM');
-  }
-
   // Element may have a player attr referring to an already created player instance.
   // If so return that otherwise set up a new player below
   if (tag.player || Player.players[tag.playerId]) {
     return tag.player || Player.players[tag.playerId];
   }
 
+  // Check if element is included in the DOM
```
gkatsev: this change would make this be `fix` rather than a `test`

gkatsev: should we have this be a separate PR from the test changes?

test/unit/player.test.js
```diff
@@ -1107,13 +1107,22 @@ QUnit.test('play promise should resolve to native value if returned', function(a
...
+    const tag = TestHelpers.makeTag();
+
+    fixture.appendChild(tag);
+
+    videojs(tag);
   }, 'a falsey techOrder should throw');
 
   videojs.options.techOrder = techOrder;
+
+  while (fixture.firstChild) {
```
gkatsev: I don't think we need to do this. qunit should be taking care to empty out qunit-fixture element after a test is run.

BrandonOCasey: good point

src/js/video.js
BrandonOCasey: Yeah sure

BrandonOCasey:
See https://github.com/videojs/video.js/pull/4755 for the other PR

PR #4753, opened 2017-11-16
test: update tests to use qunit 2 assert format

BrandonOCasey:
## Description
We cannot change the package name from `qunitjs` to `qunit` at this time as `karma-qunit` does not support `qunit` only the `qunitjs` package name.

See: https://github.com/karma-runner/karma-qunit/issues/98

Previously we stopped using an updated qunit as it did not support ie8. This may still be the case and we may not want to update qunit here at all.

This only came up because I wanted to use `Qunit.timeout` in some `sourceset` tests, but it isn't supported pre v2.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -3,12 +3,12 @@
 <head>
   <meta charset="utf-8">
   <title>VideoJS Tests</title>
-  <link rel="stylesheet" href="../node_modules/qunitjs/qunit/qunit.css">
+  <link rel="stylesheet" href="../node_modules/qunit/qunit/qunit.css">
   <link rel="stylesheet" href="../build/temp/video-js.min.css">
 </head>
 <body>
   <div id="qunit"></div>
-  <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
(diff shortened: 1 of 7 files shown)


gkatsev:
Yup, still broken. Not sure why one of the travises passed.
Oh, it thought no changes were made...

test/index.html
```diff
@@ -3,12 +3,12 @@
...
   <meta charset="utf-8">
   <title>VideoJS Tests</title>
-  <link rel="stylesheet" href="../node_modules/qunitjs/qunit/qunit.css">
+  <link rel="stylesheet" href="../node_modules/qunit/qunit/qunit.css">
   <link rel="stylesheet" href="../build/temp/video-js.min.css">
 </head>
 <body>
   <div id="qunit"></div>
-  <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
+  <script src="../node_modules/qunit/qunit/qunit.js"></script>
```
gkatsev: should these urls be reverted as well?

BrandonOCasey: yes

BrandonOCasey:
ok everything regarding the version update/name change should be reverted now.

PR #4754, opened 2017-11-16
Fixes TimeUI update on seek

shahlabs:
## Description
TimeUpdate ui does not fire on seek when paused. 

## Specific Changes proposed
By manually triggering a timeupdate we force the ui to not use the cache.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -246,7 +246,8 @@ class SeekBar extends Slider {
     this.player_.scrubbing(false);
 
     /**
-     * Triggered to force timeui to not use cache time.
+     * Trigger timeupdate because we're done seeking and the time has changed.
+     * This is particularly useful for if the player is paused to time the time displays.
      *
      * @event Tech#timeupdate
      * @type {EventTarget~Event}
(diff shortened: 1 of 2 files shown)


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -244,6 +244,8 @@ class SeekBar extends Slider {
     super.handleMouseUp(event);
 
     this.player_.scrubbing(false);
+    // forces timeui to not use cache
+    this.player_.tech_.trigger('timeupdate');
```
gkatsev: is it necessary to trigger on the tech or can we trigger it on the player itself?

Also, we should trigger this event like we do here: https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L304-L310
and we should copy over and update the jsdoc as well like there.

shahlabs: It is necessary to trigger on the tech, I had tried `this.player_.trigger`, it didn't work. I will update the trigger event.

gkatsev: cool, thanks @shahlabs!

shahlabs: Retested with this.player_.trigger and it worked, so updating the trigger on player @gkatsev

gkatsev:
Also, I updated the jsdoc comment.

Issue #4725, opened 2017-11-07
PopupButton and Popup component undefined


kocoten1992:
### Description
http://jsbin.com/qodopuzona/edit?html,console

gkatsev:
Thanks for the PR @asronihidayat.
Though, looking into this, it was only used in 5.x with the old VolumeMenuButton component which no longer exists. I'm thinking we can just remove these files until/unless they're needed?
Thoughts? @OwenEdwards would be good to have your take as well.

asronihidayat:
Actually,i use popup and popup button for custom component videojs, but if you really want to remove both of them, i can refactor my code, so, it will compatible with videojs future updated.

kocoten1992:
Hi, this is not total relevant but while we on the topic, think we are missing a something like ```ComponentList```

I'm implementing something like resolution list like youtube (1080, 720, 480...), that is a list only know length at runtime (when manifest fetched).

Our component currently is just a single component, thing like ```handleClick```, ```handleMouseDown``` will point to the root element instead of child element.

In my app I manually add event to child element, that rather cumbersome, but I'm not sure this idea 
 useful either, what your though ?

gkatsev:
@asronihidayat what about if we release the popup stuff as a separate plugin?

gkatsev:
I'm leaning towards leaving it out of core but publishing it as a separate plugin. In v6 we already moved videojs-flash out of core and core doesn't use the popup buttons anymore, so, including them seems a bit odd to me.

gkatsev:
@asronihidayat so, after some thinking about this, we aren't going to be including PopupButton and Popup back in Video.js by default. Feel free to grab the files though and use them in your project. But be aware that we stopped using PopupButtons because of accessibility issues around our old volume menu button. See some info here: http://blog.videojs.com/feature-spotlight-accessibility/.

PR #4762, opened 2017-11-22
fix: PopupButton and Popup component undefined #4725

asronihidayat:
i just try to fix this issue, https://github.com/videojs/video.js/issues/4725.


gkatsev:
Answered in #4725.

PR #4763, opened 2017-11-22
Edge 15 fix for closest.getAttribute

emkayy:
## Description
If the HTML element doesn't have a lang attribute, Edge throws an exception when calling closest.getAttribute because tag.closest('[lang]') returns an empty object (which will be coerced to true) instead of null.

## Specific Changes proposed
Just test for getAttribute on closest.

## Requirements Checklist
- [x ] Bug fixed


gkatsev:
Looks like it's fixed in Edge 16, but we can pull it in.

Issue #4609, opened 2017-09-12
npm package too big for jsDelivr


bharod:
I'm trying to load this package from [jsDelivr CDN](https://www.jsdelivr.com/package/npm/video.js) but it fails because the package is over 50 MB. Are there any files that could be removed to reduce the package size?

MartinKolarik:
50 MB out of the total 57 MB is the documentation under `docs/`  which should probably excluded from npm anyway.

thijstriemstra:
+1 for cutting docs out of npm package.

gkatsev:
I think it's good to have the documentation with the release and I think it's weird that jsdelivr makes such an exception when other CDNs have no issues (cdnjs, unpkg.com). But I get where you're coming from.
Currently, we rely on the docs folder in the release to deploy our docs site but we'll consider removing it when we change how the docs site is deployed.

MartinKolarik:
@gkatsev the limit is there because when serving a file from a package for the first time, we need to download and unzip the full package. The bigger it is, the higher the chance it will take too long and user's request will fail, which is why we used this limit (in most cases, even a 150 MB package would download fast enough, but "most cases" is not good enough).

I understand why you would want the documentation in a release that people download manually, e.g. from GitHub, but npm is mostly used for automated installation and IMHO should only have the necessary code.

gkatsev:
Btw, this should be fixed by #4774.

MartinKolarik:
@gkatsev 👍 I forgot to mention it here but I added an exception for this package so it already works, but removing the docs is still a good idea.

gkatsev:
Well, in that case, NEVERMIND! ;)
yeah, it'll also be good for the docs to auto-update, since we're now 3 semver-minor versions behind :)

PR #4774, opened 2017-11-29
docs: deploy docs using netlify

gkatsev:
## Description
This adds in the legacy docs into the repo and enables automatic deployment via netlify. The `netlify-docs.js` script will error out the build on netlify on master if we're not on a tagged commit so that it won't redeploy the docs unless there's a new release. If we're not on master or on master with a tagged commit, it will process with the deploy.
Also, this removes the API docs from being published with npm, fixing #4609.


gkatsev:
Once this is merged and published, we can change `docs.videojs.com` from pointing at the github pages to netlify.

BrandonOCasey:
Do legacy docs have to be in the repo? or could we pull them down on netlify before deploy?

build/netlify-docs.js
```diff
@@ -0,0 +1,19 @@
...
+
+const output = sh.exec(GIT_CONTAINS, {async: false, silent:true}).stdout;
+
+// if we're on master branch and not on a tagged commit,
+// error the build so it doesn't redeploy the docs
+if (process.env.BRANCH === 'master' && output === '')
+  process.exit(1);
+
+// if we're on any other branch, we can regenerate docs
+else {
```
BrandonOCasey: seems like we should have { and } here

gkatsev: um... yes. Forgot to do so.

gkatsev:
The legacy docs don't strictly have to be here but I think it's easier to just include them and copy them over during the deploy rather than trying to figure out how to download it.

BrandonOCasey:
I just think it will add a lot of bulk to the repo during `git clone` other than that I don't see why we couldn't keep it in the repo. I guess either way is fine, but maybe we should get other opinions.

misteroneill:
I don't have a problem with this. It will add something to the `git clone` cost, but I don't think it's a lot. And it's a one-time cost versus adding complexity to the versioning process to download/generate old docs.

PR #4790, opened 2017-12-04
Doc fix

thijstriemstra:
Text tracks are not video-only and can be used for audio as well.

### DIFF
--- a/docs/guides/text-tracks.md
+++ b/docs/guides/text-tracks.md
@@ -1,6 +1,6 @@
 # Text Tracks
 
-Text tracks are a feature of HTML5 for displaying time-triggered text to the user. Video.js offers a cross-browser implementation of text tracks.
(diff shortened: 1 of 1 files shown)


docs/guides/text-tracks.md
```diff
@@ -1,6 +1,6 @@
 # Text Tracks
 
-Text tracks are a feature of HTML5 video for displaying time-triggered text to the viewer. Video.js offers a cross-browser implementation of text tracks.
+Text tracks are a feature of HTML5 for displaying time-triggered text to the user. Video.js offers a cross-browser implementation of text tracks.
```
gkatsev: makes to remove `video` there, viewer does imply video but I'm not sure how well user works. Maybe end-user? Since user often implies the user of Video.js, aka, the developer.

gkatsev:
Thanks @thijstriemstra

PR #4792, opened 2017-12-05
chore: remove unused popup classes

### LINKED ISSUE #4725: PopupButton and Popup component undefined

gkatsev:
In v6, we moved away from popup due to accessibility reasons. It was
only used in the old VolumeMenuButton.

Fixes #4725.


ldayananda:
There's some references to these components in the CSS files that should also be cleaned up :)

gkatsev:
@ldayananda the MenuButton still has a concept of a popup and I think the CSS there is used by it.

PR #4797, opened 2017-12-07
feat: add throttle on adding vjs-waiting class

keymnsk:
In some browsers (chrome on windows) when playing high fps video (.mp4 60 fps) there is blinking loader that keeps showing all the time while playing, despite buffer size. 

I suggest to add a short throttle to avoid this.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1233,7 +1233,7 @@ class Player extends Component {
    * @private
    */
   handleTechWaiting_() {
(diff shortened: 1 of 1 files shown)


gkatsev:
One thing that I've been thinking for this is maybe we add a delay using css? idk if it's easily doable with how the spinner is set up currently, though.

keymnsk:
That`s a good idea. I will make another PR with css only edits.

Issue #4501, opened 2017-07-18
Seeking to negative values


MekliCZ:
## Description
There is possibility to seek to negative time using player.currentTime(), which causes weird behaviour of the player.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Seek to negative value using currentTime() function

### Expected
It shouldn't be possible to seek to negative time (my opinion is that it would be best to just seek to 0 if negative number was used as a parameter of currentTime())

### Actual
In combination with hls.js, player behaviour is weird - it seeks to end of the video

## Additional Information
Please include any additional information necessary here. Including the following:
### plugins
videojs-hls.js (integration of hls.js to video.js)

gkatsev:
Video.js itself seems to handle seeking to a negative number just fine. It could be an issue with hls.js or the its integration with video.js.

MekliCZ:
But even if seeking to negative values works with pure HTML5 video element, there is a lot of libraries using MSE which can be broken by seeking to negative values. Not only hls.js.

gkatsev:
True, we can potentially make it better but it won't solve the root cause that the MSE libs aren't handling it correctly :).
Best case, this should be fixed in both places.

MekliCZ:
+1 for fixing it both in video.js and MSE libs :)

gkatsev:
Making this a "help wanted" "first timers only" issue.
The fix is pretty easy, it just needs to clamp the negative value to 0 in `player.currentTime()`. Basically, inside of [this if statement](https://github.com/videojs/video.js/blob/master/src/js/player.js#L1791), if `seconds` are less than 0, we should set them to zero.
Please comment here if you'd like to work on it so multiple people don't work on it at once.

PR #4799, opened 2017-12-08
#4501 Seek to 0 if attempt is made to seek to negative value

CharlesRyan:
## Description
Fix for bug #4501 

## Specific Changes proposed
Added if statement to player.currentTime() that sets argument 'seconds' to 0 if negative value is passed to the function

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1789,6 +1789,9 @@ class Player extends Component {
    */
   currentTime(seconds) {
     if (typeof seconds !== 'undefined') {
+      if (seconds < 0) {
+        seconds = 0;
+      }
       this.techCall_('setCurrentTime', seconds);
       return;
     }


misteroneill:
I wonder if it might be more defensive and standards-compliant to use seekable ranges to enforce the bounds of `currentTime`? In general, this will give us a lower bound of `0`, but would help with possible edge cases.

> The earliest possible position is not explicitly exposed in the API; it corresponds to the start time of the first range in the seekable attribute's TimeRanges object, if any, or the current playback position otherwise.

Thoughts @gkatsev and/or others?

gkatsev:
Techs really should be the ones in charge of forcing the seeks to proper ranges. This is just a compromise to make sure that we don't seek to negative numbers.

misteroneill:
Sounds good to me.

PR #4800, opened 2017-12-10
feat: player resize event

### LINKED ISSUE #4629: Add an event to signal that the player was resized using the API e.g. player.dimensions()
## Description
When using the API to resize the player, ie: `player.dimensions(width, height)`, there is no event triggered to notify one that the player has been resized.  This need surfaced when working on a third-party plugin that needed to be informed of the player size changing so that underlying VPAID ads could resize in response.  

I propose we ad a `vjs-player-resize` event that is triggered when the player dimensions are changed so that developers can listen for this event to resize VPAID ads, overlays, position controls, etc. in response to the dimensions changing.

sivapalan:
## Description
Fixes #4629, and also handles multi-dimensional resizing correctly. I.e. `playerresize` event is only triggered once when width and height is changed at once (using `Component#dimensions`).

## Specific Changes proposed
Trigger a `playerresize` event in `Player#dimension` if `Player.isReady_` and `skipListeners` is falsy.

`skipListeners`  is added as an optional parameter to the following methods: `Player#width`, `Player#height`, and `Player#dimension`. This parameter can be used to skip the `playerresize` event trigger.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -659,11 +659,14 @@ class Player extends Component {
    * @param {number} [value]
    *        The value to set the `Player`'s width to.
    *
+   * @param {boolean} [skipListeners]
+   *        Skip the playerresize event trigger
+   *
    * @return {number}
    *         The current width of the `Player` when getting.
    */
-  width(value) {
-    return this.dimension('width', value);
+  width(value, skipListeners) {
+    return this.dimension('width', value, skipListeners);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


gkatsev:
Given that we're adding a new event, this would be a `feat`.

gkatsev:
Also, thanks for adding tests!

PR #4804, opened 2017-12-11
fix: force autoplay in Chrome

### LINKED ISSUE #4720: Autoplay not working on chrome, videojs 6.4.0 / chrome 62.0.3202.75 (Works on Firefox / Edge)
## Description
Using the latest 6.4.0 js and css the autoplay doesn't work only on chrome, tested on firefox / edge and it work. Also videojs 5 autoplay does work, so it seems to be a problem related to version 6.

## Steps to reproduce
1. Load latest JS / CSS 6.4.0
2. Print the player
(example)             <video id='video' class="video-js vjs-fluid" data-setup='{ "controls": true, "autoplay": true, "preload": "auto" }' width="600" height="337" poster="<?=$config->thumbUrl?>/<?=$video->filename?>.jpg?i=<?php echo time(); ?>">
3. try all 3 browsers
-Chrome
-Edge
-Firefox

## Results
Chrome - No Autoplay
Edge - Autoplays
Firefox - Autoplays
### Expected
Chrome - Autoplays
Edge - Autoplays
Firefox - Autoplays

### Actual
autoplay doesn't work

### Error output
none

## Additional Information
Video js version 5 does autoplay correctly
#### videojs
6.4.0 / 6.2.8 tested
#### browsers
chrome
#### OSes
windows
### plugins
none

gkatsev:
Chrome has started pausing autoplaying video elements when they are
moved in the DOM. Here we need to make sure that if the video started
autoplaying, after we move the element in the DOM we call play again.

Fixes #4720.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -475,7 +475,10 @@ class Player extends Component {
     // chrome started pausing the video when moving in the DOM
     // causing autoplay to not continue due to how Video.js functions.
     // See #4720 for more info.
-    if (this.paused() && this.options_.autoplay && browser.IS_CHROME) {
+    if (this.paused() &&
+        this.options_.autoplay &&
+        browser.IS_CHROME &&
+        !browser.IS_ANDROID) {
       this.play();
     }
   }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -472,15 +472,7 @@ class Player extends Component {
     this.playWaitingForReady_ = false;
     this.playOnLoadstart_ = null;
 
-    // chrome started pausing the video when moving in the DOM
-    // causing autoplay to not continue due to how Video.js functions.
-    // See #4720 for more info.
-    if (this.paused() &&
-        this.options_.autoplay &&
-        browser.IS_CHROME &&
-        !browser.IS_ANDROID) {
-      this.play();
-    }
+    this.forceAutoplayInChrome_();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/js/tech/html5.js
```diff
@@ -329,6 +333,7 @@ class Html5 extends Tech {
       }
     }
 
+    this.hasStartedPlayback_ = el.paused;
```
mister-ben: `el.paused`, is also true if it's never played.

gkatsev: yup, this should actually be `!el.paused`. Also @andriybohdan tested it and found that it just autoplayed everything, probably because of this.

mister-ben: I think this might also be too late to check anyway. The element would be moved and apparently paused at https://github.com/videojs/video.js/blob/master/src/js/player.js#L643.

What is the disadvantage to creating a new video el as we do for iphone?

gkatsev: Recreating the element in all conditions could potentially be a breaking change.
However, it'll be something to consider if this type of workaround wouldn't work.

gkatsev: Also, I think this place is fine because we set the `autoplay` attribute a few lines above. Seems like `!el.paused` seems to work here but I guess we'd want to do more tests before pulling it in.

gkatsev:
Looks like getting the conditional right also fixed the tests :)

gkatsev:
Seems like doing the check here isn't good enough. Seems like there are two options:
1. call play if autoplay is set, or
1. recreate the video element.
Option 1 isn't great because we had just removed code like this. Might be ok if we check for chrome and autoplay and we aren't playing already at the end of video.js initialization.
Option 2 isn't great because it's potentially a breaking change. I've seen too much code create video.js players but then proceed to call play and what not on the video element itself.

pagarwal123:
QA-LGTM

PR #4806, opened 2017-12-12
feat(css): add a delay before showing loading spinner

keymnsk:
In some browsers (chrome on windows) when playing high fps video (.mp4 60 fps) there is blinking loader that keeps showing all the time while playing, despite buffer size.

I suggest add a delay (0.3s) before showing it.


misteroneill:
I like this idea. I feel like we regularly have people bring up minor complaints about the spinner appearing for a split second now and then. Hopefully this improves the UX. Thanks for the PR! 👍

PR #4813, opened 2017-12-13
chore: switch to node 8

gkatsev:


gkatsev:
I think we should wait for the patch and get this merged as part of the next minor. Don't want to change deployment process as part of a patch.

PR #4823, opened 2017-12-14
Fix: use correct logic for menu focus

### LINKED ISSUE #4821: When the player is embed in an iframe, keyboard-only users are unable to select a caption track.
## Description
When the player is embed in an iframe, keyboard-only users are unable to select a caption track.

Example: 
https://mfairchild365.github.io/notes/vjs-caption-button-bug.html

## Steps to reproduce
1. Using Google Chrome 
2 .go to the above example
3. Press the `tab` key until you reach the captions button and press `enter`
4. the captions list shows, but you are unable to select a track by pressing arrow keys.

### Expected
[WCAG 2.1.1: Users should be able to select a captions track by using a keyboard only](https://www.w3.org/WAI/WCAG20/quickref/#qr-keyboard-operation-keyboard-operable)

### Actual
Focus was never sent to the sub-menu (captions list), and thus was inoperable. This causes a failure of WCAG 2.0 Level A

### Error output
none

## Additional Information

This appears to be a result of https://github.com/videojs/video.js/pull/4607 

However, I believe the logic in that should have been `if (ios && in-iframe) { dontFocus() }` instead of `if (!IS_IOS && !isInFrame()) { doFocus()}` (this is causing the bug in all browsers).

Additionally, I don't see the behavior that the PR was trying to fix when I test with IOS and remove that condition. Is it even needed anymore? 

## Proposal

Revert https://github.com/videojs/video.js/pull/4607

mfairchild365:
This was resulting in the menu not being focused for all browsers/devices (if the player was in a frame), but it should target only IOS. This fixes #4821

### DIFF
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -345,7 +345,8 @@ class MenuButton extends Component {
       // set the focus into the submenu, except on iOS where it is resulting in
       // undesired scrolling behavior when the player is in an iframe
       if (IS_IOS && Dom.isInFrame()) {
-        return; //Return early so that the menu isn't focused
+         // Return early so that the menu isn't focused
+        return;
       }
 
       this.menu.focus();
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks @mfairchild365 !

PR #4835, opened 2017-12-21
Use CSS grid for caption Setting overlay to avoid cutoff

RevinKey:
## Description
Currently the Caption settings menu has two issues:
1. extra characters added to the classname here: https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-settings.js#L437
2. When the player is smaller than ~450px the menu is squished together and some items are cut off. 

## Enhancements/ Changes
1. Removed the extra characters from the class name so the CSS will target correctly
2. Switch to CSS Grid with minmax column row size to make sure the menu is responsive and looks uniform at all sizes. 
**Before:**
Large Player:
hiddenbuttons

Small Player: 
screenshot2

**After:**
Large Player:
screen shot 2017-12-21 at 12 56 48 pm

Small Player: 
screen shot 2017-12-21 at 12 56 39 pm

## Requirements Checklist
  - [ ] Docs/guides updated


src/css/components/_captions-settings.scss
```diff
@@ -2,23 +2,32 @@
...
   background-color: rgba($primary-background-color, 0.75);
   color: $primary-foreground-color;
-  height: 70%;
+  height: 100%;
 }
 
 // Layout divs
 .vjs-text-track-settings .vjs-modal-dialog-content {
-  display: table;
+  display: grid;
```
gkatsev: we can't just switch to CSS grid because it will break all older browsers that do not support it.

gkatsev: We probably should keep what we have now and add a `@supports(grid)` (or whatever it is) for progressive enhancement. That way, older browsers will continue with what we have but newer browsers will get the grid.

RevinKey: What are the browsers we need to support? CSS grid is supported in all modern browsers and partial support in ie10/11

src/css/components/_captions-settings.scss
```diff
@@ -4,33 +4,49 @@
...
-}
-.vjs-text-track-settings .vjs-track-settings-controls {
-  grid-column: 2;
-  grid-row: 2;
+  .vjs-text-track-settings .vjs-track-settings-controls {
+    grid-column: 2;
+    grid-row: 2;
+  }
 }
+@supports not (display: grid) {
```
gkatsev: `@supports` doesn't work on the browsers that we want to target that doesn't have the grid. We need to keep the regular CSS without any check and then reset it and make it a grid inside the `@supports` block. The `@supports` block will be completely ignored on older browsers that don't support `@supports`.

gkatsev: Video.js needs to support or at least not break all the way back to IE8!

RevinKey: So have the current table CSS then below it the `@supports` so if it does support grid it will override?

gkatsev: yup, more or less.

RevinKey:
Just updated/tested to have the table CSS default and it uses Grid when supported.

misteroneill:
I guess my only concern with this would be the possibility that skins have been relying on overriding the default style and could break with these changes due to CSS rule changes. But it may be unlikely enough that we can move forward.

RevinKey:
Hey Pat, thanks for taking a look, I haven't added/renamed any CCS classes, I think if people are targeting the classes in-order to customize they shouldn't have issues. IMHO grid provides a much better responsive design and allows for more sane customizations to be added, rather than with table and flex design.

gkatsev:
I think I'm good to move forward with these changes unless anyone has strong objections. If it causes issues, we can revert and think about it some more.
I'd also take it for a spin before confirming.

gkatsev:
Testing this out, if your player is smaller than 360px in height, when you open up the dialog it will have a scroll bar but there's maybe 200-250px of content.
I think we only need a 2x2 grid rather than the current 3x3 grid, though, maybe we want a 2x3 grid where the middle row flexes so that the buttons can be bottom aligned.

After playing around a bit with it, I think that switching to these values would be good:
```css
  grid-template-columns: 1fr 1fr;
  grid-template-rows: 1fr auto;
```

RevinKey:
CR wanted: Updated this and was able to get the whole menu inside the player without the need for scrolling now.

src/css/components/_captions-settings.scss
```diff
@@ -2,9 +2,8 @@
   background-color: $primary-background-color;
   background-color: rgba($primary-background-color, 0.75);
   color: $primary-foreground-color;
-  height: 70%;
+  height: 100%;
```
gkatsev: oh, I just noticed this. We want to keep this to 70% because we use the currently set captions as a preview (probably not ideal but that's how it is).

src/css/components/_captions-settings.scss
```diff
@@ -21,7 +20,50 @@
...
+    grid-row: 1;
+  }
+
+  .vjs-text-track-settings .vjs-track-settings-controls {
+    grid-column: 2;
+    grid-row: 2;
+    text-align: unset;
+    vertical-align: unset;
+  }
+  .vjs-text-track-settings fieldset span.vjs-opacity{
```
gkatsev: there are a few places where a space should go before the opening brace.

PR #4836, opened 2017-12-21
feat: Add `getPlayer` method to Video.js.

misteroneill:
There are cases where you want to get a player by an ID or an element, but not necessarily call `videojs()` to do it - because you don't want to create a player where one is absent.

## Requirements Checklist
  - [x] Docs/guides updated


gkatsev:
test: ☑️

PR #4842, opened 2017-12-24
feat: Add `videojs.getAllPlayers` to get an array of players.

misteroneill:
When working on another PR, I noticed that there were tests somewhere not completely cleaning up after themselves. Then, I thought, wouldn't it be nice to be able to delegate method calls to _all_ players, so we could dispose them?

This new function makes that easier:

```js
videojs.getAllPlayers().forEach(p => p.dispose());
```

Previously, you'd have to do something like:

```js
Object.keys(videojs.players).map(k => videojs.players[k]).filter(Boolean).forEach(p => p.dispose());
```

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -286,36 +286,19 @@ videojs.getPlayer = (id) => {
 };
 
 /**
- * Either return an array of all players or call a function on all players.
- *
- * @param  {string} [method]
- *         If given, calls a method on all players.
- *
- * @param  {...Mixed} [args]
- *         Subsequent arguments will be passed through to the method being
- *         called.
+ * Returns an array of all current players.
  *
  * @return {Array}
- *         If a `method` argument was passed, returns an array of the return
- *         values for all players. Otherwise, if no `method` argument was
- *         passed, returns an array of all players.
- *
- *         Either way, the array will be in the order that `Object.keys`
- *         provides, which could potentially vary between JavaScript engines.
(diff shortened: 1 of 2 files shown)


gkatsev:
I'm for adding these but against this API. Having it return an array of players when called via `.all()` and then call the players when called via `.all('dispose')` is confusing and inconsistent.
What about keeping the calling functionality for calling methods and moving the returning the array to a `.getAllPlayers()` method? Also, potentially move the calling to something like `.playersDo()`, which has precedent in vim and similar functionality like `:tabdo` and `:bufdo` where you do something for each tab.

misteroneill:
Yeah, I don't disagree. I had similar thoughts.

I like `getAllPlayers` for getting an array, but `playersDo` seems confusing to me. Maybe `callAllPlayers` or `delegateAllPlayers`?

Alternatively, the `callAllPlayers` could be ignored entirely and you could simply do:

```js
videojs.getAllPlayers().forEach(p => p.dispose());
```

Another alternative approach would be to have `getAllPlayers` return an array-like object with `Player` prototype methods mixed in:

```js
const playerCollection = videojs.getAllPlayers();

playerCollection.dispose();

const playerArray = videojs.getAllPlayers().toArray();
```

gkatsev:
I'm partial to `callAllPlayers` but yeah, maybe leave it out for now until we have a real use-case or see people wanting to use it more?

Issue #4849, opened 2017-12-28
Add callback to "addRemoteTextTrack" ?


kocoten1992:
Hi there,

I was trying to do this:
```js
player.addRemoteTextTrack({src: "www.example.com"}, false)
```

The above code is async,  but we don't have a callback :sweat_smile: 

Also I have a question, why ```loadTrack``` and ```parseCues``` in text-track.js is private, if we make it part of class TextTrack then people can do thing like this:

```js
var x = new TextTrack
x.loadTrack(src, callback) // callback doesn't exist yet
```

Is there a special reason for private those ?

gkatsev:
For better or worse, we currently rely on native-like behavior for the loading of the text tracks. `addRemoteTextTrack` returns either a native or an emulated HTMLTrackElement. On that object, you can then listen to the `load` event to know when the track has loaded. Seems like we're lacking in documentation of this.
Here's an example:
```js
const trackEl = player.addRemoteTextTrack({src: 'en.vtt'}, false);
trackEl.addEventLisetner('load', function() {
  // the object has a reference to the TextTrack object via the `track` property
  console.log(trackEl.track.cues);
});
```

gkatsev:
Thanks for the docs!

kocoten1992:
sorry, my internet is very laggy, two main under sea cable in my country AAG and APG got problem at the same time, I was thinking do the docs as you said next week, but anyway thanks for merge that

gkatsev:
If you get around to it, that would be great!

PR #4855, opened 2018-01-03
docs: callback alternative to addRemoteTextTrack

kocoten1992:
## Description
Add docs to guide new user to use callback alternative of function ```player.addRemoteTextTrack```

issue: https://github.com/videojs/video.js/issues/4849

### DIFF
--- a/docs/guides/text-tracks.md
+++ b/docs/guides/text-tracks.md
@@ -229,7 +229,7 @@ As mentioned [above](#a-note-on-remote-text-tracks), remote text tracks represen
   **Note**: If you need a callback, instead of a callback you could use the technique below:
   ```js
   const trackEl = player.addRemoteTextTrack({src: 'en.vtt'}, false);
-  trackEl.on('load', function() {
+  trackEl.addEventListener('load', function() {
     // your callback go here
   });
   ```
--- a/docs/guides/text-tracks.md
+++ b/docs/guides/text-tracks.md
@@ -225,6 +225,14 @@ As mentioned [above](#a-note-on-remote-text-tracks), remote text tracks represen
 * `Player#addRemoteTextTrack(Object options)`
 
   Available options are the same as the [available `track` attributes](#track-attributes). And `language` is a supported option as an alias for the `srclang` attribute - either works here.
+  
+  **Note**: If you need a callback, instead of a callback you could use the technique below:
+  ```js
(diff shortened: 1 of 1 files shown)


docs/guides/text-tracks.md
```diff
@@ -225,6 +225,14 @@ As mentioned [above](#a-note-on-remote-text-tracks), remote text tracks represen
 * `Player#addRemoteTextTrack(Object options)`
 
   Available options are the same as the [available `track` attributes](#track-attributes). And `language` is a supported option as an alias for the `srclang` attribute - either works here.
+  
+  **Note**: If you need a callback, instead of a callback you could use the technique below:
+  ```js
+  const trackEl = player.addRemoteTextTrack({src: 'en.vtt'}, false);
+  trackEl.on('load', function() {
```
gkatsev: we should use `addEventListener` because if native tracks are used `trackEl` will reference the native object.

gkatsev:
Also, thanks for getting the ball rolling on this.

kocoten1992:
@gkatsev , after some testing, this is how thing play out:

```player.addTextTrack('kind', 'label', 'lang')``` it fully synchronous, so we don't have to wait for load of any kind, even we add ```load``` it just don't work, example below **don't work**:

```js
var trackEl = player.addTextTrack('captions', 'text1', 'en')

trackEl.addEventListener('load', function () { console.log('pleaseeee work!') }) // never work, because it already loaded - it fully sync
```

Only thing left is ```player.addRemoteTextTrack({src: 'someurl.tld'}, false)```, and since it async, and is the only player in town, should we change ```addEventListener``` to ```on```.

If that oke, I'll do a PR for this, maybe we don't need a section for this, because ```addRemoteTextTrack``` is the only one need callback ?

gkatsev:
Yeah, `addTextTrack` is for programmatic usage, so, there's no need for a `load` event, especially since it returns a TextTrack object and not a TextTrackElement object.

The main thing I think it would be good for adding a "how to wait for tracks to load" section is that it's decoupled from the "API docs" section but yeah, it probably isn't required.
The problem with using `on` is that if you get a native element object it won't have `.on` but you will if you get our emulated element. So, just using `addEventListener` you'll always have it work and it won't break in some cases. Currently, Video.js runs emulated text tracks by default on all browsers except Safari and iOS.

Issue #2964, opened 2016-01-03
Progress bar hover bug


kokororin:
When there is a 'word-break:all' in my style, the progress bar hover will also work breaked.
like this:
![3d6024a0-c141-476f-b9d0-fec1bcf1d039](https://cloud.githubusercontent.com/assets/10093992/12078397/9b87d6da-b24b-11e5-914a-5400459719a6.png)
To solve the problem,I added `word-break: normal!important;`in `video-js.css`
code like this:

``` css
.video-js .vjs-big-play-button:before, .video-js .vjs-control:before, .video-js .vjs-modal-dialog, .vjs-modal-dialog .vjs-modal-dialog-content {
  position: absolute;
  top: 0;
  left: 0;
  width: 100%;
  height: 100%;
  word-break: normal!important; }
```

gkatsev:
We should reset `word-break` inside of the [video-js element](https://github.com/videojs/video.js/blob/master/src/css/components/_layout.scss#L1). Specifically, we should set it to `word-break: initial;`.

rishabh92:
Can i work on this one?

gkatsev:
@rishabh92 you got it!

PR #4859, opened 2018-01-05
fix progress bar bug by adding word-break

rishabh92:
## Description
Add word-break to fix progress bar bug

fixes: #2964


gkatsev:
Thanks, @rishabh92. I updated the PR with a short comment.

PR #4861, opened 2018-01-05
fix: let the tech preload auto on its own

BrandonOCasey:
## Description
This was found due to the work done in #4660. Basically we reload the video element twice on every source with preload set to auto. This can potentially cause the same data to be downloaded twice.

## Specific Changes proposed
* remove players call to load, as tech should handle preload on its own, and this is harmful for the html5 tech.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2504,11 +2504,7 @@ class Player extends Component {
         this.techCall_('src', source.src);
       }
 
-      if (this.options_.preload === 'auto') {
-        this.load();
-      }
-
-    // Set the source synchronously if possible (#2326)
+      // Set the source synchronously if possible (#2326)
     }, true);
 
     return false;
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2492,6 +2492,7 @@ class Player extends Component {
     }
 
     // wait until the tech is ready to set the source
+    // and set it synchronously if possible (#2326)
     this.ready(function() {
 
       // The setSource tech method was added with source handlers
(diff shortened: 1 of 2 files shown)


BrandonOCasey:
See https://github.com/videojs/video.js/blob/d7f45bab89790fe84f8fcc162df2b9e2ff6d265b/test/unit/sourceset.test.js#L166

src/js/player.js
```diff
@@ -2471,11 +2471,7 @@ class Player extends Component {
         this.techCall_('src', source.src);
       }
 
-      if (this.options_.preload === 'auto') {
```
gkatsev: what about skipping this is `enableSourceset` is set?

src/js/player.js
```diff
@@ -2471,11 +2471,7 @@ class Player extends Component {
...
-        this.load();
-      }
-
-    // Set the source synchronously if possible (#2326)
+      // Set the source synchronously if possible (#2326)
```
gkatsev: this indent is interesting. I think it was dedented so that it lines up with the this.ready call, not sure if it's necessary or not. Maybe we want to make a new line for the arg and have the comment directly above the arg?

BrandonOCasey:
Since `sourceset` is now in master I made the skipped tests run again.

BrandonOCasey:
I think we may be better of testing if this causes breakage with some QA, but if we just want to do it later we could put it behind the `enableSourceset` flag.

pagarwal123:
QA-LGTM

PR #4864, opened 2018-01-08
feat: playerresize event in all cases

gkatsev:
Use [ResizeObserver](https://github.com/WICG/ResizeObserver) when available for better and more performant resizing information, otherwise, fall back to a throttled `resize` event on an iframe that's the size of the player.
Allows a video.js user to disable this by setting `resizeManager: false` as an option since the component will not be initialized.
This probably needs to revert #4800 (https://github.com/videojs/video.js/commit/e0ed0b5cd762f83dc54a7b56e8774718485e924d) because we end up getting two `playerresize` events with the dimension methods now.

TODO:
* [x] Revert #4800 
* [x] fix current tests
* [x] make RO polyfill possible
* [x] allow RO to be disabled
* [x] debounce calls instead of throttle
* [x] Write tests
* [x] documentation
* [x] make sure things are disposed properly
  * [x] events
  * [x] objects


gkatsev:
One thing to consider as well here is that we're going to get multiple `playerresize` events when changing into fullscreen, depending on what the throttle is, it can be anywhere from 1 to 3 events. Currently, the ResizeObserver isn't throttled but the iframe `resize` event is throttled to 50ms. It's probable that we want to debounce rather than throttle.

test/unit/resize-manager.test.js
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+
+    let playerresizeCalled = 0;
+    const rm = new ResizeManager(this.player, {ResizeObserver: MyResizeObserver});
+
+    this.player.on('playerresize', function() {
+      playerresizeCalled++;
+    });
+    rm.resizeObserver_.observer();
```
ldayananda: `observe`?

ldayananda: actually looks like you don't need this line at all

test/unit/resize-manager.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+    this.clock.tick(100);
+
+    assert.equal(playerresizeCalled, 1, 'playerresize was triggered');
+
+    rm.dispose();
+  });
```
ldayananda: one good test to add would be a test that makes sure a resizeObserver is created when one is available on the window

src/js/player.js
```diff
@@ -730,17 +720,6 @@ class Player extends Component {
...
-
-    // skipListeners allows us to avoid triggering the resize event when setting both width and height
-    if (this.isReady_ && !skipListeners) {
-      /**
-       * Triggered when the player is resized.
-       *
-       * @event Player#playerresize
-       * @type {EventTarget~Event}
-       */
-      this.trigger('playerresize');
```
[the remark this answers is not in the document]
gkatsev: We are falling back to using an iframe for playerresize events when ResizeObserver isn't available, currently, that's basically everywhere, though, chrome will be shipping ResizeObserver soon.

src/js/resize-manager.js
```diff
@@ -0,0 +1,79 @@
...
+import * as Events from './utils/events.js';
+import mergeOptions from './utils/merge-options.js';
+import Component from './component.js';
+
+class ResizeManager extends Component {
+  constructor(player, options) {
+    let RESIZE_OBSERVER_AVAILABLE = options.ResizeObserver || window.ResizeObserver;
+
+    if (options.ResizeObserver === null) {
+      RESIZE_OBSERVER_AVAILABLE = false;
```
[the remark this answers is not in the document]
gkatsev: Yup, the idea is if you wanted to disable the ResizeObserver for whatever reason you could pass in `null`.

test/unit/resize-manager.test.js
gkatsev: It would be nice but mocking out `window` is a pain :)

test/unit/resize-manager.test.js
gkatsev: I set this in the constructor of MyResizeObserver, calling it calls the observe handler which will trigger `playerresize` which we can then test. Maybe I should just make it a local variable to be more obvious what is going on.

src/js/resize-manager.js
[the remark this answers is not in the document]
ldayananda: Gotcha. I'd suggest a comment over this block so it's clear for future readers 😄

test/unit/resize-manager.test.js
ldayananda: Ah, I see. I agree that making it a local variable will make it more clear

src/js/resize-manager.js
```diff
@@ -0,0 +1,79 @@
...
+
+      this.on('load', this.loadListener_);
+    }
+  }
+
+  createEl() {
+    return super.createEl('iframe', {
+      className: 'vjs-resize-manager'
+    });
+  }
```
misteroneill: I believe the element will be created even when it's not needed. Should we suppress the creation of it when `ResizeObserver` is available?

pagarwal123:
QA-LGTM

Issue #4431, opened 2017-06-21
Horizontal Volume Slider doesn't auto-hide after mute + adjust volume (version 6.1)


Equipement:
Hello,

Horizontal Volume Slider doesn't auto-hide after mute + adjust volume (in VideoJs version 6.1)

Steps to reproduce the problem :
- Play video
- Click on the Volume Button icon (should mute video)
- Now click on any mid-Volume Level in the Horizontal Slider
- Wait
- Volume Slider still there (and is under the progress bar)

Equipement:
I have see that :
a) Click on the Volume Button icon (should mute video) => the Volume Button icon has the focus
b) Click on the Volume Button icon (should mute video) then click on any mid-Volume Level in the Horizontal Slider => the Volume Button icon has the focus (and not the Horizontal Slider)

Equipement:
In the file videojs.css, with this update (4 lines are disabled), it seems to work :

```
  .video-js .vjs-volume-panel:hover .vjs-volume-control,
  .video-js .vjs-volume-panel:active .vjs-volume-control,
  .video-js .vjs-volume-panel:focus .vjs-volume-control,
  .video-js .vjs-volume-panel .vjs-volume-control:hover,
  .video-js .vjs-volume-panel .vjs-volume-control:active,
/*  .video-js .vjs-volume-panel .vjs-volume-control:focus, */
  .video-js .vjs-volume-panel .vjs-mute-control:hover ~ .vjs-volume-control,
  .video-js .vjs-volume-panel .vjs-mute-control:active ~ .vjs-volume-control,
/*  .video-js .vjs-volume-panel .vjs-mute-control:focus ~ .vjs-volume-control, */
  .video-js .vjs-volume-panel .vjs-volume-control.vjs-slider-active {
    visibility: visible;
    opacity: 1;
    position: relative;
    -webkit-transition: visibility 0.1s, opacity 0.1s, height 0.1s, width 0.1s, left 0s, top 0s;
    -moz-transition: visibility 0.1s, opacity 0.1s, height 0.1s, width 0.1s, left 0s, top 0s;
    -ms-transition: visibility 0.1s, opacity 0.1s, height 0.1s, width 0.1s, left 0s, top 0s;
    -o-transition: visibility 0.1s, opacity 0.1s, height 0.1s, width 0.1s, left 0s, top 0s;
    transition: visibility 0.1s, opacity 0.1s, height 0.1s, width 0.1s, left 0s, top 0s; }
[...]

PR #4866, opened 2018-01-09
fix: Hide horizontal volume panel while the slider is not active and mute toggle button is on focus

mrdtron:
## Description
Show horizontal volume panel only when the volume slider is active or volume button is hovered but not when focused. This addresses [#4431](https://github.com/videojs/video.js/issues/4431)

## Requirements Checklist

### DIFF
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -53,10 +53,8 @@
   &:focus  .vjs-volume-control,
   & .vjs-volume-control:hover  ,
   & .vjs-volume-control:active ,
-  & .vjs-volume-control:focus  ,
   & .vjs-mute-control:hover  ~ .vjs-volume-control,
   & .vjs-mute-control:active ~ .vjs-volume-control,
-  & .vjs-mute-control:focus  ~ .vjs-volume-control,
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks for the PR and your patience. Code-wide, this looks good. Though, I'd like to take it for a spin before merging.

gkatsev:
Just took a look, this is pretty great. Thanks!
One thing that I noticed is that if you tab to the mute button the volume slider doesn't show, though, not sure if that's a big deal or not. Any thoughts @OwenEdwards?

mrdtron:
@gkatsev Thanks for reviewing this! I'd personally think that slider should appear only when hovering on the mute toggle button or tabbing into it from the keyboard — a good example of the behaviour is the Youtube player — but I'm open to suggestions. 

One issue I noticed just now is that once you mute/unmute volume through keyboard volume slider flickers. Fixing that now.

gkatsev:
Yeah, I guess it's fine not to open the volume slider via the keyboard when tabbing into the mute toggle, looks fine in youtube.

gkatsev:
And I confirmed no volume slider flicker when toggling mute.

PR #4868, opened 2018-01-09
feat: add mediator middleware type for play()

ldayananda:
## Description
This will allow middleware to interact with calls to play() from the tech. This will require a method of indicating to middleware previously run that a middleware down the chain has terminated or stopped execution. 

## Specific Changes proposed
- Adds middleware `mediator` method that runs middleware from the player to the tech and a second time back up to the player. This category was created because `play` is both a setter(changes the playback state) and a getter(gets a native play promise if available). This also has the ability to tell whether a middleware has terminated before reaching the tech.
- Adds a `middleware.TERMINATOR` sentinel value that is available on the `videojs` object
- Adds `play` to the allowedMediators
- Adds `paused` to the allowedGetters
- Adds a sandbox example of a play mediator middleware

## Checklist
- [x] unit tests added
- [x] documentation added
- [x] guide added/updated
- [x] verified in browser
- [x] example created

### DIFF
--- a/docs/guides/middleware.md
+++ b/docs/guides/middleware.md
@@ -5,14 +5,15 @@ Middleware is a Video.js feature that allows interaction with and modification o
 ## Table of Contents
 
 * [Understanding Middleware](#understanding-middleware)
+  * [Termination and Mediators](#termination-and-mediators)
 * [Using Middleware](#using-middleware)
 * [setSource](#setsource)
 
 ## Understanding Middleware
 
-Middleware are functions that return an object with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable` and `played`.
+Middleware are functions that return an object with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable`, `played`, `play` and `pause`.
 
-These allowed methods are split into two categories: `getters` and `setters`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`.
+These allowed methods are split into three categories: `getters`, `setters` and `mediators`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`.
 
 ```
 +----------+                      +----------+
@@ -23,6 +24,31 @@ These allowed methods are split into two categories: `getters` and `setters`. Se
 |          |  getter middleware   |          |
 +----------+                      +----------+
 ```
+
+### Termination and Mediators
+
+Mediators are the thrid category of allowed methods. These are methods that not only change the state of the Tech, but also return some value back to the Player. Currently, these are `play` and `pause`.
+
+Mediators make a round trip: starting at the `Player`, mediating to the `Tech` and returning the result to the `Player` again. A `call{method}` method must be supplied by the middleware which is used when mediating to the `Tech`. For example: `callPlay`. On the way back to the `Player`, the `{method}` will be called instead, with 2 arguments: `terminated`, a Boolean indicating whether a middleware terminated during the mediation to the tech portion, and `value`, which is the value returned from the `Tech`.
+
+```
+               mediate to tech
+               +------------->
+
++----------+                      +----------+
+|          |                      |          |
+|          +-----call{method}+---->          |
+|  Player  |                      |   Tech   |
+|          <-------{method}-------+          |
+|          |                      |          |
++----------+                      +----------+
+
+              <---------------+
+              mediate to player
+
+```
+
+Middleware termination occurs when a middleware method decides to stop mediating to the Tech. We'll see more examples of this in the next section.
 
 ## Using Middleware
 
@@ -56,6 +82,23 @@ var myMiddleware = function(player) {
 videojs.use('*', myMiddleware);
 ```
 
+Mediator methods can terminate, by doing the following:
+
+```javascript
+var myMiddleware = function(player) {
+  return {
+    callPlay: function() {
+      // Terminate by returning the middleware terminator
+      return videojs.middleware.TERMINATOR;
+    },
+    play: function(terminated, value) {
+      // the terminated argument should be true here.
+    }
+  };
+};
+
+videojs.use('*', myMiddleware);
+```
 
 ## setSource
 
@@ -71,4 +114,3 @@ videojs.use('*', function(player) {
   };
 });
 ```
-
--- a/docs/guides/middleware.md
+++ b/docs/guides/middleware.md
@@ -7,11 +7,12 @@ Middleware is a Video.js feature that allows interaction with and modification o
 * [Understanding Middleware](#understanding-middleware)
   * [Termination and Mediators](#termination-and-mediators)
 * [Using Middleware](#using-middleware)
+  * [Terminating Mediator Methods](#terminating-mediator-methods)
 * [setSource](#setsource)
 
 ## Understanding Middleware
 
-Middleware are functions that return an object with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable`, `played`, `play` and `pause`.
+Middleware are functions that return an object with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable`, `played`, `play`, `pause` and `paused`.
 
 These allowed methods are split into three categories: `getters`, `setters` and `mediators`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`.
 
@@ -28,8 +29,6 @@ These allowed methods are split into three categories: `getters`, `setters` and 
 ### Termination and Mediators
 
 Mediators are the third category of allowed methods. These are methods that not only change the state of the Tech, but also return some value back to the Player. Currently, these are `play` and `pause`.
-
-Mediators make a round trip: starting at the `Player`, mediating to the `Tech` and returning the result to the `Player` again. A `call{method}` method must be supplied by the middleware which is used when mediating to the `Tech`. For example: `callPlay`. On the way back to the `Player`, the `{method}` will be called instead, with 2 arguments: `terminated`, a Boolean indicating whether a middleware terminated during the mediation to the tech portion, and `value`, which is the value returned from the `Tech`.
 
 ```
                mediate to tech
@@ -48,7 +47,25 @@ Mediators make a round trip: starting at the `Player`, mediating to the `Tech` a
 
 ```
 
-Middleware termination occurs when a middleware method decides to stop mediating to the Tech. We'll see more examples of this in the next section.
+Mediators make a round trip: starting at the `Player`, mediating to the `Tech` and returning the result to the `Player` again. A `call{method}` method must be supplied by the middleware which is used when mediating to the `Tech`. On the way back to the `Player`, the `{method}` will be called instead, with 2 arguments: `terminated`, a Boolean indicating whether a middleware terminated during the mediation to the tech portion, and `value`, which is the value returned from the `Tech`. A barebones example of a middleware with Mediator methods is:
+
+```
+var myMiddleware = function(player) {
+  return {
+    callPlay: function() {
+      // mediating to the Tech
+      ...
+    },
+    pause: function(terminated, value) {
+      // mediating back to the Player
+      ...
+    },
+    ...
+  };
+};
+```
+
+Middleware termination occurs when a middleware method decides to stop mediating to the Tech. We'll see more examples of this in the [next section](#terminating-mediator-methods).
 
 ## Using Middleware
 
@@ -81,6 +98,8 @@ var myMiddleware = function(player) {
 
 videojs.use('*', myMiddleware);
 ```
+
+### Terminating Mediator Methods
 
 Mediator methods can terminate, by doing the following:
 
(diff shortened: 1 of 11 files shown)


src/js/player.js
```diff
@@ -1632,6 +1632,9 @@ class Player extends Component {
     this.ready(function() {
       if (method in middleware.allowedSetters) {
         return middleware.set(this.middleware_, this.tech_, method, arg);
+
+      } else if (method in middleware.allowedMediators) {
```
ldayananda: `player.play()` seems to be done via `techGet` for the html5 tech, but it's conceivable that a future mediator will use `techCall`

src/js/tech/middleware.js
```diff
@@ -2,6 +2,8 @@ import { assign } from '../utils/obj.js';
 
 const middlewares = {};
 
+export const TERMINATOR = 'TERMINATOR';
```
gkatsev: we should make this be an object instance rather than a string since objects only equal themselves. Then the string will also be a valid value.

src/js/player.js
gkatsev: maybe we should just make a `techCallAndGet` for stuff like play?

src/js/tech/middleware.js
ldayananda: what do you mean by "the string will also be a valid value"

src/js/player.js
ldayananda: That might make sense; I don't really strongly about it one way or another.

src/js/tech/middleware.js
ldayananda: Did you mean change this to `export const TERMINATOR = () => { return 'TERMINATOR'; }` and then here:

https://github.com/ldayananda/video.js/blob/52011c3bc175897e735ec5018159de222db0980c/src/js/video.js#L752

change it to be `videojs.TERMINATOR = TERMINATOR()` ?

gkatsev: nope, literally just `TERMINATOR = {}`

ldayananda: Oh, you meant "the string will also be a valid value" to mean that an allowed method could return the string 'TERMINATOR'! 👍  will do

src/js/video.js
```diff
@@ -749,5 +749,7 @@ videojs.dom = Dom;
  */
 videojs.url = Url;
 
+videojs.TERMINATOR = TERMINATOR;
```
misteroneill: If we have access to `Object.defineProperty` we should probably use it here to prevent overwriting this value. Others may think that's overkill, though.

ldayananda: Seems reasonable, I'll do that

src/js/video.js
```diff
@@ -749,7 +749,9 @@ videojs.dom = Dom;
  */
 videojs.url = Url;
 
-videojs.TERMINATOR = TERMINATOR;
+Object.defineProperty(videojs, 'TERMINATOR', {
```
ldayananda: @misteroneill does this seem right?

src/js/video.js
```diff
@@ -749,5 +749,9 @@ videojs.dom = Dom;
  */
 videojs.url = Url;
 
+Object.defineProperty(videojs, 'TERMINATOR', {
```
gkatsev: this will break in IE8 :)

Also, I think it's reasonable to make it enumerable and it might be nice to make writeable false be explicit.

gkatsev: should this be directly on `videojs`? Maybe namespaced `videojs.middleware.TERMINATOR`?

src/js/tech/middleware.js
```diff
@@ -31,20 +34,41 @@ export function set(middleware, tech, method, arg) {
...
+  played: 1,
+  paused: 1
 };
 
 export const allowedSetters = {
   setCurrentTime: 1
 };
 
+export const allowedMediators = {
+  play: 1
```
gkatsev: are we adding `pause` here as well?

src/js/tech/middleware.js
```diff
@@ -53,6 +77,14 @@ function middlewareIterator(method) {
   };
 }
 
+function executeRight(mws, method, value, terminated) {
+  for (let i = mws.length - 1; i >= 0; i--) {
+    const mw = mws[i];
+
+    mw[method](value, terminated);
```
gkatsev: can we swap the argument order so it's similar to a node-style callback?

src/js/player.js
gkatsev: Let's get everything else ready and then come back to this.

src/js/big-play-button.js
```diff
@@ -59,7 +59,7 @@ class BigPlayButton extends Button {
 
     const playFocus = () => playToggle.focus();
 
-    if (isPromise(playPromise)) {
+    if (playPromise && isPromise(playPromise)) {
```
gkatsev: `isPromise` should be doing a null check as well, no, this shouldn't be strictly necessary.

ldayananda: I did notice that `isPromise` was checking for `undefined` specifically. I thought that was for a specific reason, but maybe I was wrong?

src/js/tech/middleware.js
ldayananda: Oh, I should. Thanks for catching that

src/js/tech/middleware.js
ldayananda: Good idea, we can do that.

src/js/big-play-button.js
gkatsev: if `play()` is called and no promise was returned, we end up with `undefined`. Though, we'll never go into this if statement if `playPromise` doesn't duck-type into a promise.
I'm ok with leaving it if you want, but it shouldn't be necessary.

src/js/video.js
ldayananda: Is there an example somewhere else in the project of how to do this for IE8 as well? I tried to find one earlier and came up empty.

🆗  explicit `writeable: false` and `enumerable: true` sound good to me

ldayananda: I like the idea of it being namespaced to `videojs.middleware.TERMINATOR`

gkatsev: something like
```js
if (Object.defineProperty && !browser.IS_IE8) {
  Object.defineProperty(...);
} else {
  videojs.TERMINATOR = TERMINATOR;
}
```

src/js/big-play-button.js
ldayananda: Ah, while I was testing, if the play was terminated then I saw an error coming from this `if` statement. Perhaps the right way is to return something other than `null` in the cases where a middleware terminates.

gkatsev: Ah, I see. Maybe we should just loosen the check in `isPromise`

src/js/video.js
```diff
@@ -751,8 +751,8 @@ videojs.url = Url;
 
 // Object.defineProperty is not available in IE8
 if (!browser.IS_IE8 && Object.defineProperty) {
-  Object.defineProperty(videojs, 'middleware.TERMINATOR', {
-    value: TERMINATOR,
+  Object.defineProperty(videojs, 'middleware', {
```
gkatsev: you'd want defineProperty each step because now you could overwrite `videojs.middleware.TERMINATOR = false`

ldayananda: Good point

docs/guides/middleware.md
```diff
@@ -0,0 +1,116 @@
+# Middleware
```
ldayananda: This will have to be rebased once the current middleware guide is merged

gkatsev:
Looking at this version of the middleware guide (https://deploy-preview-4868--videojs-docs.netlify.com/tutorial-middleware.html), I realized we forgot to document the "special" `setTech` method. Also, looks like we say we have to return an object with these methods but an instance of a class or prototype or w/e will work too. But we can do this as part of a separate PR, probably.
Also, the package-lock has conflicts, yay! (sorry)

sandbox/middleware-play.example
```diff
@@ -0,0 +1,184 @@
+<!DOCTYPE html>
```
gkatsev: this file should be `middleware-play.html.example`

sandbox/middleware-play.example
```diff
@@ -0,0 +1,184 @@
...
+  <script>
+    videojs.options.flash.swf = '../build/temp/video-js.swf';
+  </script>
+
+  <style>
+    .terminate-btn {
+      margin: 2em 1em;
+    }
+
+    .vjs-progress-control.vjs-control .vjs-play-progress.vjs-slider-bar.terminated {
```
gkatsev: why not look for the `terminated` class on the player instead?

src/js/player.js
gkatsev: eh, I think we can just leave it like this.

ldayananda: Yeah, doesn't feel like there's a pressing need to change this yet

sandbox/middleware-play.example
gkatsev:
```css
.terminated .vjs-process-control .vjs-play-progress
```
Then below (line 44), you can just do `player.addClass('terminated')` instead of going through the entire child heirarchy to get the proper element. Or something like that.

ldayananda:
@gkatsev I think I've now addressed all your comments. I agree that the corrections to the middleware guide can be done in a separate PR. And I removed the package-lock so we should be good now 😄

src/js/tech/middleware.js
```diff
@@ -26,15 +26,28 @@ export function setTech(middleware, tech) {
   middleware.forEach((mw) => mw.setTech && mw.setTech(tech));
 }
 
+/**
+ * Calls a getter on the tech first, through each middleware
+ * from right to left to the player.
+ **/
```
gkatsev: nit: only one `*` for closing :)

docs/guides/middleware.md
```diff
@@ -5,14 +5,15 @@ Middleware is a Video.js feature that allows interaction with and modification o
...
+Middleware are functions that return an object with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable`, `played`, `play` and `pause`.
 
-These allowed methods are split into two categories: `getters` and `setters`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`.
+These allowed methods are split into three categories: `getters`, `setters` and `mediators`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`.
```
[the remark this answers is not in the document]
gkatsev: Probably. Would be nice to get a one-sentence description, though, if we can.

src/js/video.js
```diff
@@ -333,6 +333,31 @@ videojs.registerTech = Tech.registerTech;
...
+    enumerable: true
+  });
+
+  Object.defineProperty(videojs.middleware, 'TERMINATOR', {
+    value: TERMINATOR,
+    writeable: false,
+    enumerable: true
+  });
+} else {
+  videojs.middleware.TERMINATOR = TERMINATOR;
```
gkatsev: this will fail because `videojs.middleware` isn't an object.
Should be `videojs.middleware = {TERMINATOR}`

gkatsev:
Oh, the commit message should probably be more like `feat: add mediator middleware type for play()`

docs/guides/middleware.md
```diff
@@ -5,14 +5,16 @@ Middleware is a Video.js feature that allows interaction with and modification o
...
 
-These allowed methods are split into two categories: `getters` and `setters`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`.
+These allowed methods are split into three categories: `getters`, `setters` and `mediators`. Setters will be called on the `Player` first and run through middleware(from left to right) before calling the method, with its arguments, on the `Tech`. Getters are called on the `Tech` first and are run though middleware(from right to left) before returning the result to the `Player`. Mediators are called on the `Player` first, run through middleware from left to right, then called on the `Tech` and the result is returned to the `Player` unchanged, while calling the middleware from right to left. For more information, check out the [mediator section](#termination-and-mediators).
```
ldayananda: @gkatsev Added a sentence for Mediators, but it's more confusing in my opinion than reading the full section

gkatsev: 🤔  I think this is ok. Maybe @videojs/core-committers has any other thoughts?

misteroneill: I think this paragraph should be split up. I propose breaking it into three distinct sections. Also, made some grammatical improvements.

----------

These allowed methods are split into three categories: getters, setters, and mediators.

### Middleware Setters
Setters will be called on the `Player` first and run through middleware (from left to right) before calling the method, with its arguments, on the `Tech`.

### Middleware Getters
Getters are called on the `Tech` first and are run though middleware (from right to left) before returning the result to the `Player`.

### Middleware Mediators
Mediators are called on the `Player` first, run through middleware (from left to right), then called on the `Tech`. The result is returned to the `Player` unchanged, while calling the middleware from right to left. For more information on mediators, check out the [mediator section](#termination-and-mediators).

src/js/utils/promise.js
```diff
@@ -9,7 +9,7 @@
  *         Whether or not the object is `Promise`-like.
  */
 export function isPromise(value) {
-  return value !== undefined && typeof value.then === 'function';
+  return value !== undefined && value !== null && typeof value.then === 'function';
```
misteroneill: Mabye we should just do `return Boolean(value) && typeof value.then === 'function';`

src/js/video.js
```diff
@@ -325,8 +325,39 @@ videojs.getTech = Tech.getTech;
...
+ *
+ * @type {object}
+ * @memberOf {videojs}
+ * @property {object} middleware.TERMINATOR
+ */
+// Object.defineProperty is not available in IE8
+if (!browser.IS_IE8 && Object.defineProperty) {
+  Object.defineProperty(videojs, 'middleware', {
+    value: {},
+    writeable: false,
```
misteroneill: I believe `writable: false` is the default, so this shouldn't be necessary.

ldayananda: We talked about this elsewhere, but since having `writeable: false` is very explicit, I plan to keep it 😄

src/js/utils/promise.js
ldayananda: it's nice to be explicit here too 🤔

PR #4874, opened 2018-01-11
docs: Update COLLABORATOR_GUIDE.md to include label meanings

ldayananda:


COLLABORATOR_GUIDE.md
```diff
@@ -30,6 +30,20 @@ Any issue and PR can be closed if they are not relevant, when in doubt leave it
...
+There are labels that are useful to include on issues and PRs. A few of them are defined below:
+
+| Label | Issue or PR | Description |
+|-------|-------------|-------------|
+| confirmed | Issue and PR | Issue: marks as reproducible. PR: marks as ready to be merged|
+| 5.x | PR | Marks as a change to the 5.x branch only |
+| bug | Issue | Marks as a confirmed bug |
+| first-timers-only | Issue | Marks as a good bug or enhancement for first time contributors only |
+| minor, patch, major | PR | Marks PR with the expected semver classification of the change |
+| needs: LGTM | PR | Marks PR to be reviewed by a collaborator |
```
mister-ben: is it worth mentioning the `needs: more info` and `needs: reduced test case here` labels here or is there somewhere better for that?

ldayananda: I was considering putting those in the CONTRIBUTING.md but wasn't sure which file they fit into better. Any preference?

gkatsev: we probably want to have them here and in contributing.md with different explanations for the different audiences. In here with respect to what as a maintainer I should use and when and there with respect to when I'm contributing, what does it mean when something I opened or PRed was given a particular label.

ldayananda:
Added the labels section to `CONTRIBUTING.md` as well! 😄

CONTRIBUTING.md
```diff
@@ -74,6 +75,20 @@ A good bug report should be as detailed as possible, so that others won't have t
...
+### Labels
+
+There are a few labels that might be added to your issue or PR by a maintainer. Here's a quick rundown of what they mean:
+
+| Label | Issue or PR | Description |
+|-------|-------------|-------------|
+| confirmed | Issue and PR | Issue: marks as reproducible by a maintainer. PR: marked by a maintainer as ready to be merged|
+| 5.x | PR | Marks as a change to the 5.x branch only |
+| bug | Issue | Marks as a confirmed bug by a maintainer |
+| first-timers-only | Issue | Marked as a good bug or enhancement for first time contributors only |
```
OwenEdwards: @gkatsev what's the difference between `first-timers-only` and  `good first issue`? I thought `good first issue` was what you were flagging "a good bug or enhancement for first time contributors", but I'm not entirely clear why an issue would be for "first timers **only**". Are you suggesting that people who aren't "first timers" **shouldn't** work on those issues? That seems like it's much less important that "a first timer is encouraged to work on this issue rather than on more complicated ones", so I'd expect to see `good first issue` described here rather than `first-timers-only`.

gkatsev: Maybe this is something to elaborate on but `first-timers-only` should be super simple issues that are strongly encouraged to be done by people who are coming into open source for the first time (or well, recently) and well-established people should try and not address those issues. `good first issue` can be directed at anyone as a good point to get into the project.
This was started by Kent C Dodds and he has a blog post about it: https://blog.kentcdodds.com/first-timers-only-78281ea47455
so far, I think using it has turned out pretty well.

ldayananda: thanks for the question @OwenEdwards! I've updated both guides to be more clear on the meanings of these two labels.

PR #4885, opened 2018-01-17
fix: Do not patch canPlayType on Android Chrome

mister-ben:
## Description
Patching `canPlayType` on Android shouldn't be necessary on Android Chrome since it reports its ability to player HLS correctly.

## Specific Changes proposed
Skip patching if UA is Android Chrome.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1051,7 +1051,8 @@ const mp4RE = /^video\/mp4/i;
 Html5.patchCanPlayType = function() {
 
   // Android 4.0 and above can play HLS to some extent but it reports being unable to do so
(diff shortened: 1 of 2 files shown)


mister-ben:
I think there's a vanishingly small chance that anyone's still watching video on Andorid 2.2 but as it would be a breaking change ought it stay for now?

gkatsev:
I guess we should leave it in.

gkatsev:
@mister-ben also, should we do this as a fix or would it be better to release as a minor?

gkatsev:
Actually, since I'm going to merge this and the other features, it'll be released as a minor.

PR #4903, opened 2018-01-25
test: add project and build names to browserstack

gkatsev:
videojs/http-streaming is starting to run tests in browserstack so we should name our runs better.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -1,7 +1,8 @@
 module.exports = function(config) {
   var browserstackName = process.env.TRAVIS_BUILD_NUMBER + ' ' + process.env.TRAVIS_BRANCH;
+  console.log(process.env);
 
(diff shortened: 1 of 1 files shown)


gkatsev:
Looks like 
screen shot 2018-01-25 at 18 23 09

Previously it would just say `7907`

forbesjo:
This might be out of the original scope of the PR but you could also turn off `video` to disable video capture and, I think, speed up VM startup. The recorded video isn't particularly useful since none of the tests have a visual output.

gkatsev:
Turning it off sounds good to me, though, we do have the qunit output enabled and it has been helpful for IE8 debugging, but we can always add it in manually when needed and re-run the build.

gkatsev:
Looks like turning it off doesn't really make much of a difference in our build times so I'm going to leave them on.

PR #4912, opened 2018-01-29
chore: generate a test example on netlify on PRs

gkatsev:
Sometimes, we get a small PR that has code that's good but testing it means that we have to check out the code locally. This should generate a `test-example` folder on the docs site for the PR with a built version of Video.js from the PR.
Hopefully, this PR will be the the very first example of it too.

### DIFF
--- a/build/generate-example.js
+++ b/build/generate-example.js
@@ -17,7 +17,7 @@ export function cleanupExample() {
   sh.rm('-rf', path.join(dest, 'node_modules'));
 }
 
-export default function generateExample({skipBuild}) {
+export default function generateExample({skipBuild} = {}) {
   // run the build
   if (!skipBuild) {
     sh.exec('npm run build');
--- a/build/generate-example.js
+++ b/build/generate-example.js
@@ -1,0 +1,33 @@
+import path from 'path';
+import sh from 'shelljs';
+import klawSync from 'klaw-sync';
+
+const dest = 'docs/api/';
+const distDest = path.join(dest, 'dist');
+const exampleDest = path.join(dest, 'test-example');
+
+export function cleanupExample() {
+  sh.rm('-rf', distDest);
+  sh.rm('-rf', exampleDest);
+}
+
+export default function generateExample({skipBuild}) {
+  // run the build
+  if (!skipBuild) {
+    sh.exec('npm run build');
+  }
+
+  // make sure that the example dest is available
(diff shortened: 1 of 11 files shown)


gkatsev:
IT WORKED! https://deploy-preview-4912--videojs-docs.netlify.com/test-example/
All the sandbox files are available, for example https://deploy-preview-4912--videojs-docs.netlify.com/test-example/plugin.html

build/grunt.js
```diff
@@ -449,7 +448,6 @@ module.exports = function(grunt) {
     'cssmin',
 
     'copy:fonts',
-    'copy:swf',
```
gkatsev: our switch to node 8/npm 5 meant that packages were flattened, so, the swf snuck back into the dist folder.

build/netlify-docs.js
```diff
@@ -16,4 +17,9 @@ if (process.env.BRANCH === 'master' && semver.valid(output.trim()) === null) {
 
   // copy the legacy docs over
   sh.cp('-R', 'docs/legacy-docs', 'docs/api/docs');
+
+  if (process.env.BRANCH !== 'master') {
```
gkatsev: there's no need or want to generate this in the master branch currently, I don't think.

build/generate-example.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  // make sure that the example, flash, and swf dests are available
+  sh.mkdir('-p', exampleDest);
+  sh.mkdir('-p', vjsFlashDest);
+  sh.mkdir('-p', swfDest);
+
+  // copy the `dist` dir
+  sh.cp('-R', 'dist', path.join(dest, 'dist'));
+  sh.rm(path.join(dest, 'dist', `video-js-${pkg.version}.zip`));
```
gkatsev: we don't need the zip file published here.

sandbox/icons.html.example
```diff
@@ -4,7 +4,7 @@
   <meta charset="utf-8" />
   <title>Video.js Icons Sandbox</title>
 
-  <link href="../build/temp/video-js.css" rel="stylesheet" type="text/css">
+  <link href="../build/video-js.css" rel="stylesheet" type="text/css">
```
misteroneill: Is this one intentionally using `build` instead of `dist`?

gkatsev: nope. Probably got confused about it, heh.

PR #4913, opened 2018-01-30
 fix(middleware): do a null check in mediator methods

gkatsev:


test/unit/tech/middleware.test.js
```diff
@@ -394,3 +394,18 @@ QUnit.test('setSource will select all middleware of a given type, until src chan
...
+  let pauseCalled = 0;
+  const myMw = {};
+  const mwFactory = () => myMw;
+  const mwFactory2 = () => ({
+    pause() {
+      pauseCalled++;
+    }
+  });
+
+  middleware.mediate([mwFactory(), mwFactory2()], {pause: () => {}}, 'pause');
```
gkatsev: This will throw and thus fail the test without the change in `middleware.js`.

Issue #4902, opened 2018-01-24
Let's discuss a text alternative for the loading spinner


misteroneill:
I'm particularly interested in perspective from @OwenEdwards.

Currently, there is no text alternative for the loading spinner to reflect that a `waiting` event has occurred. It seems like this could be a useful accessibility measure, but the precise implementation of it is an open question. Here are some ideas to get discussion going:

1. Add a visually-hidden `role="alert"` element to the player that should announce that the player is loading/buffering.
1. Add a visually-hidden status element immediately before the control bar that contains the current status of the player if it's interesting.
1. Any other ideas?

The loading spinner shows on seeking, but I don't think we should interrupt seeking when buffering is expected in that case. At least, not until seeking is finished and the player is buffering at the seeked-to point.

Jad3z:
+1

gkatsev:
I think the discussion here is a bit larger than the loading spinner specifically but rather a more general "how do we notify the user, particularly, one that is using a screen reader, of state changes in the player"? For sighted users, a loading spinner is generally a good indicator that something is happening wrt loading of the video but currently our loading spinner has no text alternative for screen readers. Adding a text alternative to it, I think, is the first step to this and should be done regardless of how we choose to address the notification issue.

gkatsev:
Just merged #4916 which adds a text alternative to the loading spinner. This doesn't fully solve the issues here but gets us one step closer to making things better.

gkatsev:
So, circling back to this, I don't think we want or need an alternative. Instead, we want to have the loading spinner announce the state to the screen reader, if the user is focused inside the player. Then, if the spinner is continuously up for more than a certain time, it should announce again, for example, every 10 seconds.

We don't have many other things that we need to provide status for to screen readers that don't already do so themselves.

OwenEdwards:
This example is probably the way to go:

http://pauljadam.com/demos/loading-indicator.html

OwenEdwards:
Here's another reference on this:

https://codeburst.io/how-to-create-a-simple-css-loading-spinner-make-it-accessible-e5c83c2e464c

Although we should note that in some cases it makes sense to announce to a screen reader user that something in the page is loading, but it *doesn't* make sense if they don't have the video player (or part of it) focused *and* they are interacting with or reading another part of the page.

PR #4916, opened 2018-01-31
fix: Add alternate text in an ARIA alert to the loading spinner.

misteroneill:
Partial fix for #4902 

Verified in VoiceOver.

## Requirements Checklist

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -80,5 +80,6 @@
   "Done": "Done",
   "Caption Settings Dialog": "Caption Settings Dialog",
   "Beginning of dialog window. Escape will cancel and close the window.": "Beginning of dialog window. Escape will cancel and close the window.",
-  "End of dialog window.": "End of dialog window."
+  "End of dialog window.": "End of dialog window.",
+  "{1} is loading.": "{1} is loading."
 }
--- a/src/js/loading-spinner.js
+++ b/src/js/loading-spinner.js
@@ -28,8 +28,6 @@ class LoadingSpinner extends Component {
     const el = super.createEl('div', {
       className: 'vjs-loading-spinner',
       dir: 'ltr'
-    }, {
-      role: 'alert'
     });
 
     el.appendChild(controlText);
--- a/src/js/loading-spinner.js
+++ b/src/js/loading-spinner.js
@@ -2,6 +2,7 @@
  * @file loading-spinner.js
  */
 import Component from './component';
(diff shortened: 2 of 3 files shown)


src/js/loading-spinner.js
```diff
@@ -17,10 +18,20 @@ class LoadingSpinner extends Component {
...
   createEl() {
+    const isAudio = this.player_.isAudio();
+    const playerType = this.localize(isAudio ? 'Audio Player' : 'Video Player');
+
     return super.createEl('div', {
       className: 'vjs-loading-spinner',
       dir: 'ltr'
-    });
+    }, {
+      role: 'alert'
```
gkatsev: I'm not sure we want this, at least not yet.

src/js/loading-spinner.js
```diff
@@ -17,10 +18,20 @@ class LoadingSpinner extends Component {
...
+    }, [
+      dom.createEl('div', {
```
gkatsev: I'm not seeing the `innerHTML` in the live example: https://deploy-preview-4916--videojs-docs.netlify.com/test-example/

Also, I think it would be better to use a `span` here as it also patches other `vjs-control-text` uses.

lang/en.json
```diff
@@ -80,5 +80,6 @@
   "Done": "Done",
   "Caption Settings Dialog": "Caption Settings Dialog",
   "Beginning of dialog window. Escape will cancel and close the window.": "Beginning of dialog window. Escape will cancel and close the window.",
-  "End of dialog window.": "End of dialog window."
+  "End of dialog window.": "End of dialog window.",
+  "{1} is loading.": "{1} is loading."
```
gkatsev: can you regenerate `grunt check-translations`?

gkatsev:
Also, this could definitely be considered a `fix`.

src/js/loading-spinner.js
misteroneill: Yeah, me either. I wanted to see how it behaved.

lang/en.json
misteroneill: Done

misteroneill:
This morning, we discussed the potential for `role="alert"` to be disruptive. One idea thrown around was to only add it when the player gains focus.

BrandonOCasey:
I would think that we can add it if any part of the player is currently in focus, or the player gains focus

PR #4918, opened 2018-02-01
Fix the video continue player when user drag in timeline

199911:
## Description
When user seek by dragging the area below the timeline, the video continue playing.
Video should stop when user is seeking
pause-when-seek-before

## Specific Changes proposed
Reuse the mouse up and mouse down event handler of seekBar. So the video is paused during seek
pause-when-seek-after-2

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/progress-control.js
+++ b/src/js/control-bar/progress-control/progress-control.js
@@ -157,6 +157,9 @@ class ProgressControl extends Component {
    */
   handleMouseDown(event) {
     const doc = this.el_.ownerDocument;
+    const seekBar = this.getChild('seekBar');
+
+    seekBar.handleMouseDown(event);
 
     this.on(doc, 'mousemove', this.throttledHandleMouseSeek);
     this.on(doc, 'touchmove', this.throttledHandleMouseSeek);
@@ -175,6 +178,9 @@ class ProgressControl extends Component {
    */
   handleMouseUp(event) {
     const doc = this.el_.ownerDocument;
+    const seekBar = this.getChild('seekBar');
+
+    seekBar.handleMouseUp(event);
 
     this.off(doc, 'mousemove', this.throttledHandleMouseSeek);
     this.off(doc, 'touchmove', this.throttledHandleMouseSeek);
--- a/src/js/control-bar/progress-control/progress-control.js
+++ b/src/js/control-bar/progress-control/progress-control.js
@@ -55,22 +55,25 @@ class ProgressControl extends Component {
    */
   handleMouseMove(event) {
     const seekBar = this.getChild('seekBar');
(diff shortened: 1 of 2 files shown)


gkatsev:
Hey, thanks for making the PR. We haven't done anything to change this behavior, though, I definitely can reproduce this locally with 6.7.0 and cannot reproduce it with 6.6.0. However, I *cannot* reproduce this with 6.7.1. Were you using 6.7.0? If so, can you retest with 6.7.1? I think I may want to deprecate version 6.7.0 itself and just stick with 6.7.1.

199911:
I can reproduce the problem on 6.7.1
I am using Chrome 63.0.3239.132 in mac OS.
Here is my steps on building a local copy of 6.7.1 player

```bash
git checkout master
git pull upstream master
rm -rf dist
cp sandbox/index.html.example sandbox/index.html
npm run build
npm run start
# open http://localhost:9999/sandbox/
```

PS: the bug only happen when dragging the thin area between the white timeline and control bar, but not the white timeline

gkatsev:
Oh, I see. The PS is the important part here :)
I can now reproduce even in 6.7.1, I was probably accidentally scrubbing in different areas.

gkatsev:
Just tried it here (https://deploy-preview-4918--videojs-docs.netlify.com/test-example/) and looks like this is causing an issue when dragging on the seekbar itself now. Seeking in the progress control area will pause and play correctly but now when seeking in the seekbar, it will stay paused when stopped scrubbing.

199911:
@gkatsev 
Thanks for your feedback. The bug is should be caused by `handleMouseDown()` is called twice.

When we drag the seek bar, the `handleMouseDown()` record the video player is *playing*.
And then the mouse down event is propagate to the progress control panel, `handleMouseDown()` is called again and record the video player is *paused before drag*. 
So the player does not resume after drag.

My fix is to stop the event propagate from seek bar to progress control panel, if it is handled by seek bar already.

src/js/control-bar/progress-control/progress-control.js
```diff
@@ -157,6 +157,9 @@ class ProgressControl extends Component {
    */
   handleMouseDown(event) {
     const doc = this.el_.ownerDocument;
+    const seekBar = this.getChild('seekBar');
```
ldayananda: Just in case the seekBar was removed from the player through the player options, you'll want to double check the seekBar component was found before calling `seekBar.handleMouseDown()`

199911: That's true, I have also checked other function in this component to make sure seekBar exist.

gkatsev:
definitely doesn't hurt to do the check, though, disabling the seekbar as makes the component useless :)

Issue #4910, opened 2018-01-28
Techs can change the player poster only once


LoveIsGrief:
## Description

New techs are unable to integrate into the poster lifecycle and can switch a poster only once which triggers a `posterchange` event.

[Example](https://codepen.io/anon/pen/bLbmxL?editors=0010)

The region of code that seems to be responsible is [handleTechPosterChange_](https://github.com/videojs/video.js/blob/acf4153963640c51e21187bb6d3eb3d6e38289c7/src/js/player.js#L2594)

## Steps to reproduce

Explain in detail the exact steps necessary to reproduce the issue.
1. Create a tech that triggers a `posterchange` event and changes the poster
2. Implement the `poster` method on the new tech
3. Change the poster and trigger a `posterchange `event

### Expected

A new poster

### Actual

The same old poster.

### Error output

None

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
5 + 6, most likely older versions as well

#### browsers

firefox, chrome,

#### OSes

tested on Kubuntu 16.04

### plugins

None

gkatsev:
Interesting. That's definitely an unintended side-effect of the code. The idea was that if the user sets a poster the tech shouldn't override it but the way that the code accepts a poster from the tech means that it only gets set once. It totally makes sense for the techs to be able to change their poster as much as possible assuming we don't change the behavior that a user-set poster isn't overridden. Would you be able to and interested in submitting a fix?
Thanks.

LoveIsGrief:
Yeah, sure. I wouldn't mind. There'd be a few things to clear up first, since I see a few possibilities here:

**Introduce techs that have the power to override any poster**

Could add a static field to the tech that is checked in the posterchange handler e.g

```js
if( (!this.poster_ || this.tech_.canOverridePoster) && this.tech_.poster ){
  this.poster_ = this.tech_.poster();
  this.trigger('posterchange'); 
}
```

**Introduce boolean option in player to allow techs to override the poster**

```js
if( (!this.poster_ || this.options_.techCanOverridePoster) && this.tech_.poster ){
  this.poster_ = this.tech_.poster();
  this.trigger('posterchange'); 
}
```

**Track when poster has been set by `player.setPoster`**

This could be done with a boolean in the `player.setPoster` and setting the poster to an empty string will allow techs to override the poster once again. Maybe setting it to `false` will indicate that it should remain unset.

--------------

[...]

PR #4921, opened 2018-02-01
Allow techs to change poster if player option `techCanOverridePoster` is set

LoveIsGrief:
## Description

The option for the player `techCanOverridePoster` is introduced in this commit. It allows techs to update the post whenever they like. `isPosterFromTech_` is introduced as a private player field in order to track when a poster was set by a tech. This allows us to clear the poster whenever the tech is disposed of by the player.

Additionally, attempting to set the same poster more than once will have no effect / no changes will be made, since the poster is the same. This was done in order to stop triggering multiple *posterchange* events when calling `player.poster(aPoster)` with `techCanOverridePoster` set to true.

Related to #4910 

## Specific Changes proposed

 - `techCanOverridePoster` option
 - `isPosterFromTech_` player field
 - `player.poster(aPoster)` triggers a single *posterchange* event
 - `poster.poster(aPoster) && poster.poster(aPoster)` triggers a single *posterchange* event

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -927,7 +927,8 @@ class Player extends Component {
       'poster': this.poster(),
       'language': this.language(),
       'playerElIngest': this.playerElIngest_ || false,
-      'vtt.js': this.options_['vtt.js']
+      'vtt.js': this.options_['vtt.js'],
+      'canOverridePoster': !!this.options_.techCanOverridePoster
     };
 
     TRACK_TYPES.names.forEach((name) => {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -317,6 +317,9 @@ class Player extends Component {
     // Run base component initializing with new options
     super(null, options, ready);
 
+    // Tracks when a tech changes the poster
+    this.isPosterFromTech_ = false;
+
     // Turn off API access because we're loading a new tech that might load asynchronously
     this.isReady_ = false;
 
@@ -513,6 +516,8 @@ class Player extends Component {
 
     if (this.tech_) {
       this.tech_.dispose();
+      this.isPosterFromTech_ = false;
+      this.poster_ = '';
     }
 
     if (this.playerElIngest_) {
@@ -1017,6 +1022,13 @@ class Player extends Component {
     this.tech_.dispose();
 
     this.tech_ = false;
+
+    if (this.isPosterFromTech_) {
+      this.poster_ = '';
+      this.trigger('posterchange');
+    }
+
+    this.isPosterFromTech_ = false;
   }
 
   /**
@@ -2655,11 +2667,17 @@ class Player extends Component {
       src = '';
     }
 
+    if (src === this.poster_) {
+      return;
+    }
+
     // update the internal poster variable
     this.poster_ = src;
 
     // update the tech's poster
     this.techCall_('setPoster', src);
+
+    this.isPosterFromTech_ = false;
 
     // alert components that the poster has been set
     /**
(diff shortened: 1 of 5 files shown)


LoveIsGrief:
I don't know which docs to update or where the options are documented, but I wouldn't mind adding them if someone points me to them.

gkatsev:
For docs, I'd think that [options](https://github.com/videojs/video.js/blob/master/docs/guides/options.md) and [tech](https://github.com/videojs/video.js/blob/master/docs/guides/tech.md) docs get updated. Not sure if there are other things that would need to be updated or not. Updating the docs would be awesome!

src/js/player.js
```diff
@@ -513,6 +516,8 @@ class Player extends Component {
 
     if (this.tech_) {
       this.tech_.dispose();
+      this.isPosterFromTech_ = false;
```
gkatsev: why only set and remove these inside of this if statement? Maybe we just want to do it unconditionally?

src/js/player.js
```diff
@@ -1017,6 +1022,13 @@ class Player extends Component {
     this.tech_.dispose();
 
     this.tech_ = false;
+
+    if (this.isPosterFromTech_) {
```
gkatsev: unset the poster if the poster was set via the tech? Interesting. That sounds like a great idea.

test/unit/player.test.js
```diff
@@ -1709,3 +1709,93 @@ QUnit.test('player.duration() sets the value of player.cache_.duration', functio
...
+
+QUnit.test('setPoster in tech with `techCanOverridePoster` in player should override poster', function(assert) {
+  const player = TestHelpers.makePlayer({
+    techCanOverridePoster: true
+  });
+  const posterchangeSpy = sinon.spy();
+  const firstPosterUrl = 'https://wherever.test/test.jpg';
+  const techPosterUrl = 'https://somewhere.text/my/image.png';
+
+  assert.equal(player.options_.techCanOverridePoster, true);
```
gkatsev: can we add a description to all the assertions?

LoveIsGrief: Do you mean as comments or here in the PR?

gkatsev: in the PR, the third argument to the assert calls.

src/js/player.js
LoveIsGrief: I didn't want to be greedy and just wanted to limit the effects of my changes. But I suppose they can unconditionally be set.

I believe a `posterchange` event and an update of the `PosterImage` component would be useless here too. Do you concur?

test/unit/player.test.js
LoveIsGrief: Ah, I see. On it!

src/js/player.js
gkatsev: Yup, no need to do that during cleanup.

src/js/player.js
```diff
@@ -922,7 +927,8 @@ class Player extends Component {
       'poster': this.poster(),
       'language': this.language(),
       'playerElIngest': this.playerElIngest_ || false,
-      'vtt.js': this.options_['vtt.js']
+      'vtt.js': this.options_['vtt.js'],
+      'canOverridePoster': !!this.options_.techCanOverridePoster
```
gkatsev: seems reasonable.

LoveIsGrief:
Hi, is there something more that needs to be done in order to get this merged?

gkatsev:
Hey, thanks for your patience @LoveIsGrief. I kept it unmerged because we had some fixes that we wanted to merge in before doing another minor release. We're also in the process of doing some tests of this PR as well. I think we can probably get a new minor out this or next week. Nothing more is required on your end, thanks!

LoveIsGrief:
Thanks for the response @gkatsev. I was just curious :smiley: 

Cheers!

pagarwal123:
QA-LGTM

gkatsev:
Thanks @LoveIsGrief!

PR #4926, opened 2018-02-02
docs(middleware): updating the middleware guide with setTech and other corrections

ldayananda:


docs/guides/middleware.md
```diff
@@ -5,28 +5,43 @@ Middleware is a Video.js feature that allows interaction with and modification o
...
   * [Middleware Mediators](#middleware-mediators)
   * [Termination and Mediators](#termination-and-mediators)
 * [Using Middleware](#using-middleware)
   * [Terminating Mediator Methods](#terminating-mediator-methods)
-* [setSource](#setsource)
 
 ## Understanding Middleware
 
-Middleware are functions that return an object with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable`, `played`, `play`, `pause` and `paused`.
+Middleware are functions that return an object, a class instance, a prototype, e.c.t, scoped to the Player with methods matching those on the `Tech`. There are currently a limited set of allowed methods that will be understood by middleware. These are: `buffered`, `currentTime`, `setCurrentTime`, `duration`, `seekable`, `played`, `play`, `pause` and `paused`. These allowed methods are split into three categories: [getters](#middleware-getters), [setters](#middleware-setters), and [mediators](#middleware-mediators).
```
incompl: Is e.c.t a correct stylization? I usually see etc.

docs/guides/middleware.md
```diff
@@ -5,28 +5,43 @@ Middleware is a Video.js feature that allows interaction with and modification o
...
+  * [setTech](#setTech)
   * [Middleware Setters](#middleware-setters)
   * [Middleware Getters](#middleware-getters)
   * [Middleware Mediators](#middleware-mediators)
   * [Termination and Mediators](#termination-and-mediators)
 * [Using Middleware](#using-middleware)
   * [Terminating Mediator Methods](#terminating-mediator-methods)
-* [setSource](#setsource)
 
 ## Understanding Middleware
```
incompl: I suggest starting out with a purpose statement for readers who have arrived at this page with no context. Something like "Middleware allows you to override the behavior of basic player methods."

docs/guides/middleware.md
ldayananda: whoops, good catch. I often make this mistake

docs/guides/middleware.md
ldayananda: That's actually at the top of the file 😄

incompl: Doh, missed that. Looks good!

docs/guides/middleware.md
```diff
@@ -38,28 +53,44 @@ Mediators are called on the `Player` first, run through middleware (from left to
...
+|          +----+call{method}+---->          |
+|  Player  |                      |   Tech   |
+|          <------+{method}+------+          |
+|          |                      |          |
++----------+                      +----------+
 
 ```
 
-Mediators make a round trip: starting at the `Player`, mediating to the `Tech` and returning the result to the `Player` again. A `call{method}` method must be supplied by the middleware which is used when mediating to the `Tech`. On the way back to the `Player`, the `{method}` will be called instead, with 2 arguments: `terminated`, a Boolean indicating whether a middleware terminated during the mediation to the tech portion, and `value`, which is the value returned from the `Tech`. A barebones example of a middleware with Mediator methods is:
+A skeleton of a middleware with Mediator methods is given below:
```
gkatsev: I noticed that this example uses `callPlay` and `pause`. Should the `pause` be `play`?

docs/guides/middleware.md
```diff
@@ -104,45 +135,78 @@ var myMiddleware = function(player) {
...
+          .then(function() {
+            console.log('The play succeeded!')
+          })
+          .catch(function (err) {
+            console.log('The play was rejected', err);
+          });
+      }
     },
-    ...
+    setSource: function(srcObj, next) {
```
gkatsev: can we move `setSource` to be the first property of the object? Since it's required and it's important to know which source it's selected with, having it first, I think, is useful.

gkatsev: below as well

docs/guides/middleware.md
ldayananda: good catch

Issue #4851, opened 2017-12-29
Audio tags not working


calvincorreli:
## Description

This seems almost too obvious to not be intentional, but I couldn't see anything about it in either the 6.0 release notes or migration guide.

It looks like the audio-only player isn't working, and hasn't been since 6.0.

## Steps to reproduce

Here's a JSfiddle with 6.5.0:

https://jsfiddle.net/q5avhfLf/

Click the play button on the video.

### Actual

Nothing happens.

### Expected

Expecting to hear the audio and see the control bar chugging along.

This one uses 5.19 and is otherwise identical, and it does exactly that:

https://jsfiddle.net/ym5vvjzd/1/

If you remove videojs, it also plays fine.

I've tested in Chrome, Safari, and Firefox on Mac.

No error messages on the console.

Did something change in 6.0 that I didn't see documented somewhere?

kocoten1992:
The first link https://jsfiddle.net/q5avhfLf/,  videojs was 5.19, please update it to 6.5.0, seem to be a bug

calvincorreli:
So sorry, updated the 6.5.0 fiddle above.

Also here:

https://jsfiddle.net/calvincorreli/k66zok6o/2/

misteroneill:
Thanks for the report and the reduced test case. 👍

This definitely appears to be broken. The browser can natively play the MP3 if I disable the call to `videojs()` and I tried all the way back to 6.0.0 - playback didn't work and the player behaves as if it doesn't have a source.

calvincorreli:
That's exactly what I saw, too. 

It first I was sure the bug had to be in my code. So strange that nobody seems to have noticed since the release all the way back in April!

Any ideas on what might be the cause?

misteroneill:
Not sure, we'll take a look at it when everyone's back online tomorrow, but any investigative help is appreciated.

I produced another test case just as a sanity check: http://output.jsbin.com/wihiyuj/3

gkatsev:
Looks like there's some disconnect between the player and the tech in the audio case. If I call play directly on the tech: `player.tech().play()` it works. I can then pause but not play again.

gkatsev:
Ah, looks like the issue is with the `src` attribute. Seems to have broken. If I move to use a source element, it works: http://jsbin.com/yavovakecu/edit?html,output
I think we've always had issues with the `src` attribute not being fully supported. I would definitely encourage using the source element rather than the attribute. Though, we should still look into having it work with the src attribute.

[...]

Issue #4469, opened 2017-07-06
Bad behaviour when no type specified on source


davidgg:
## Description
Right now, when using a video with source tag but without a type attribute, video.js takes the extension of the file and checks if the browser is able to play it. You can see it here:
https://github.com/videojs/video.js/blob/b63666379d1a16e4995d08ea62b29a52f7607a26/src/js/tech/html5.js#L1672

```
Html5.nativeSourceHandler.canHandleSource = function(source, options) {

  // If a type was provided we should rely on that
  if (source.type) {
    return Html5.nativeSourceHandler.canPlayType(source.type);

  // If no type, fall back to checking 'video/[EXTENSION]'
  } else if (source.src) {
    const ext = Url.getFileExtension(source.src);

    return Html5.nativeSourceHandler.canPlayType(`video/${ext}`);
  }

  return '';
};
```

There are some valid formats that are not working because of this, some samples:

.opus -> video/ogg
.mov -> video/mp4 
.aac -> audio/mp4
...

## Steps to reproduce
1. Create a new video or audio tag and use a .mov, .opus or .aac file without type
2. Video.js shows an error message

### Expected
Legit audio/video formats are accepted.

### Actual
Some legit audio/video formats cause an error

## Proposal
Maintain a hash with popular extensions and MIME types, check against this hash the extensions to get the MIME type.

misteroneill:
Your proposal would be relatively easy to implement, I think. I wonder if there's a historical reason why Video.js does not maintain such a mapping.

Any thoughts from @videojs/core-committers?

gkatsev:
Having a full mime-type mapping will greatly increase the size of the player. If we trim it, how do we decide which to include and which to exclude? `.mp4` is probably one of the only ones that's pretty reliable. In addition to this, source urls can and often come without a filetype, so, you can't even reliably detect the type based on the url and even worth is if you have a url like akamai where it has stuff like .mp4 and .m3u8 in the same url.
I'm not inherently opposed but unfortunately, it isn't as simple as just including known mimetypes for video.

davidgg:
Hi @gkatsev thank you for the reply.

I'm talking only about some popular types, not a full mime-type list. The samples I said (.opus, .mov, .aac) are quite popular and are not the only. Right now video.js is breaking them, we have to answer if this is a better situation than having some types in the code. 

All the scenarios you are describing are happening now, if you check the code you can see that it is using the extension when no type is provided, so it's not a problem related with the solution suggested.

[...]

PR #4947, opened 2018-02-11
feat: add mimetype type to source object when possible (#4469)

davidgg:
File mimetype is filled when the file extension is known and type is not provided.

## Description
Feature to support some bad behaviours like #4469 and #4851.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/filter-source.js
+++ b/src/js/utils/filter-source.js
@@ -57,27 +57,14 @@ const filterSource = function(src) {
  *        src Object with known type
  */
 function checkMimetype(src) {
-  const type = extractType(src.src);
+  const ext = Url.getFileExtension(src.src);
+  const mimetype = MimetypesKind[ext.toLowerCase()];
 
-  if (!src.type && type) {
-    src.type = type;
+  if (!src.type && mimetype) {
+    src.type = mimetype;
   }
 
   return src;
-}
-
-/**
- * Extracts video mimetype
- *
- * @param {string} src
- *        The src whose type will be extracted
- * @return {string}
- *        Extracted Mimetype for src
- */
-function extractType(src) {
-  const ext = Url.getFileExtension(src);
-
-  return MimetypesKind[ext.toLowerCase()];
 }
 
 export default filterSource;
--- a/src/js/utils/filter-source.js
+++ b/src/js/utils/filter-source.js
@@ -2,6 +2,8 @@
  * @module filter-source
  */
 import {isObject} from './obj';
+import {MimetypesKind} from './mimetypes';
+import * as Url from '../utils/url.js';
 
 /**
  * Filter out single bad source objects or multiple source objects in an
@@ -34,10 +36,10 @@ const filterSource = function(src) {
     src = newsrc;
   } else if (typeof src === 'string' && src.trim()) {
     // convert string into object
-    src = [{src}];
+    src = [checkMimetype({src})];
   } else if (isObject(src) && typeof src.src === 'string' && src.src && src.src.trim()) {
     // src is already valid
-    src = [src];
+    src = [checkMimetype(src)];
   } else {
     // invalid source, turn it into an empty array
     src = [];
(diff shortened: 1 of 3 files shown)


gkatsev:
Looks like compared to master, we are gaining ~0.15KB of minified gzipped filesize:
on [this branch](https://travis-ci.org/videojs/video.js/builds/340192167#L3691)
```
┌────────────────────────┬───────────┬───────────┐
│ filename               │      size │   gzipped │
├────────────────────────┼───────────┼───────────┤
│ alt/video.novtt.js     │ 648.37 KB │ 144.99 KB │
├────────────────────────┼───────────┼───────────┤
│ alt/video.novtt.min.js │ 169.26 KB │  43.32 KB │
├────────────────────────┼───────────┼───────────┤
│ video.cjs.js           │ 637.07 KB │ 141.65 KB │
├────────────────────────┼───────────┼───────────┤
│ video.es.js            │ 636.84 KB │ 141.59 KB │
├────────────────────────┼───────────┼───────────┤
│ video.js               │ 701.28 KB │ 159.38 KB │
├────────────────────────┼───────────┼───────────┤
│ video.min.js           │ 188.47 KB │  49.73 KB │
└────────────────────────┴───────────┴───────────┘
```
and [on master](https://travis-ci.org/videojs/video.js/builds/340272195#L3823):
```
┌────────────────────────┬───────────┬───────────┐
│ filename               │      size │   gzipped │
├────────────────────────┼───────────┼───────────┤
│ alt/video.novtt.js     │ 647.29 KB │ 144.61 KB │
├────────────────────────┼───────────┼───────────┤
│ alt/video.novtt.min.js │ 168.89 KB │  43.18 KB │
├────────────────────────┼───────────┼───────────┤
│ video.cjs.js           │ 635.99 KB │ 141.27 KB │
├────────────────────────┼───────────┼───────────┤
│ video.es.js            │ 635.76 KB │ 141.22 KB │
├────────────────────────┼───────────┼───────────┤
│ video.js               │  700.2 KB │ 159.01 KB │
├────────────────────────┼───────────┼───────────┤
│ video.min.js           │  188.1 KB │  49.58 KB │
└────────────────────────┴───────────┴───────────┘
```

davidgg:
Hi, I've changed the code to remove some mimetypes. This hash should contain only popular mimetypes when extension differs from `video/${extension}` template. Saying 'popular' is easy to understand but not to mantain, so you probably will have to add some extensions. I implemented the ones I considered popular. 

I've changed also the filter-sources method to lighted it up a little. More optimization is posible but compromising code readability.

hartman:
Small observation of this patch: since matroska and mp4 are listed... what about webm videos ?

gkatsev:
Probably makes sense to add webm, though, I this wasn't meant as a definitive or anything and that we can add or remove things if needed.

davidgg:
@gkatsev @hartman That's not necessary.  As I said, only files whose mimetype differs from `video/${extension}` template should appear in the hash. That is not the case for webm, as its mimetype is `video/webm` (https://www.webmproject.org/docs/container/). Right now video.js is already using a fallback to know if the browser can handle a video, that's the reason we need the hash. It uses that template when src has not a type property, as you can see here: 

https://github.com/videojs/video.js/blob/master/src/js/tech/html5.js#L1772-L1777

This fallback covers most of the cases but sometimes is wrong so we are using the mimetypes hash to fill type property for known problematic extensions.

**Additional**.[ Is the Video source type tag optional? If it is should I put the src inline rather than nested](https://stackoverflow.com/questions/34506704/is-the-video-source-type-tag-optional-if-it-is-should-i-put-the-src-inline-rath)?

gkatsev:
Ah, that's great @davidgg, I totally didn't know this :)

I think with this PR we already cover the case of a `source` element without a type attribute. Since each source in an array gets run through `filterSource`. Though, I guess would be good to verify.

davidgg:
I didn't manually check it but you can pass an array of sources to filter, that case is automatically covered in the new tests. Every item in the array is processed.

gkatsev:
Yeah, source elements are turned into an array of source objects when the player is initialized, so, we're probably all set. At the same time, I haven't seen anyone use a source element and not provide a type yet, even if it is technically optional :)

gkatsev:
Also, this is good to go but there are a few patch fixes I'd like to get in before doing another minor release, so, letting it hang out a bit.

gkatsev:
Thanks @davidgg!

PR #4962, opened 2018-02-19
feat: Add an option to override format-time (#2931)

twosmalltrees:
## Description
Addresses #2931, approximately following the suggestions of @gkatsev in the discussion on #3433.  Adds named export `setFormatTime`  to **utils/format-time.js**. Allows overwriting the standard format-time implementation with a custom function, which will be called with seconds and guide as arguments. Also exposes setFormatTime on **videojs**.

Example:

```javascript
videojs.setFormatTime((seconds, guide) => `${seconds}, ${guide}`));
```

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/format-time.js
+++ b/src/js/utils/format-time.js
@@ -16,7 +16,7 @@
  * @return {string}
  *         Time formatted as H:MM:SS or M:SS
  */
-function formatTime(seconds, guide = seconds) {
+function implementation(seconds, guide) {
   seconds = seconds < 0 ? 0 : seconds;
   let s = Math.floor(seconds % 60);
   let m = Math.floor(seconds / 60 % 60);
@@ -44,4 +44,11 @@ function formatTime(seconds, guide = seconds) {
   return h + m + s;
 }
 
-export default formatTime;
+// override format-time with a custom implementation
+export function setFormatTime (customImplementation) {
+  implementation = customImplementation;
+}
+
+export default function (seconds, guide = seconds) {
+  return implementation(seconds, duration)
+};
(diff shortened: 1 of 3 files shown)


gkatsev:
Hey, thanks for your patience. I think this looks good for me. I'll spend a bit more time looking at the code and testing today or tomorrow.

src/js/video.js
```diff
@@ -810,5 +810,10 @@ videojs.dom = Dom;
  */
 videojs.url = Url;
 
+/**
+ * Replaces format-time with a custom implementation
```
gkatsev: can the JSdoc be elaborated here? This is the main doc that Video.js users will see for this method, probably.
Also, we should mention that if an item, like the duration, has already formatted itself, setting a new format time won't update that item.

src/js/utils/format-time.js
```diff
@@ -3,20 +3,7 @@
...
- *        Number of seconds to be turned into a string
- *
- * @param {number} guide
- *        Number (in seconds) to model the string after
- *
- * @return {string}
- *         Time formatted as H:MM:SS or M:SS
- */
-function formatTime(seconds, guide = seconds) {
+let implementation = function(seconds, guide) {
```
gkatsev: do we need a way to get back to the original implementation?

src/js/video.js
gkatsev: can this be moved to be below where we set `formatTime` so the two things sit together?

pagarwal123:
QA-LGTM

twosmalltrees:
No problems at all, thanks for taking the time to review the changes! I've added in `resetFormatTime` to reset back to the default implementation, added to the jsDocs and re-arranged as per your feedback

gkatsev:
Thanks @twosmalltrees! It's great to finally get this in!

Issue #4964, opened 2018-02-20
Error when setting textTrackSettings to false


ookami125:
To put it simply I created a new videojs object with the propery textTrackSettings: false and I added a remote text track. After that I click on the CC button and click on the remote track that was just added then error below pops up.
```
video.js:498 VIDEOJS: ERROR: TypeError: Cannot read property 'getValues' of undefined
    at TextTrackDisplay.updateForTrack (http://vjs.zencdn.net/6.6.3/video.js:12225:52)
    at TextTrackDisplay.updateDisplay (http://vjs.zencdn.net/6.6.3/video.js:12203:12)
    at Player.bound (http://vjs.zencdn.net/6.6.3/video.js:2345:15)
    at HTMLDivElement.bound (http://vjs.zencdn.net/6.6.3/video.js:2345:15)
    at HTMLDivElement.data.dispatcher (http://vjs.zencdn.net/6.6.3/video.js:1971:31)
    at trigger (http://vjs.zencdn.net/6.6.3/video.js:2100:25)
    at Player.trigger$$1 [as trigger] (http://vjs.zencdn.net/6.6.3/video.js:2908:12)
    at Player.(anonymous function) [as handleTechTexttrackchange_] (http://vjs.zencdn.net/6.6.3/video.js:24266:17)
    at Html5.bound (http://vjs.zencdn.net/6.6.3/video.js:2345:15)
    at HTMLVideoElement.bound (http://vjs.zencdn.net/6.6.3/video.js:2345:15)
```
videojs version: 6.6.3
browsers: chrome, firefox
os: windows 10
what platforms (operating systems and devices) are affected?
plugins:
I'm using a custom plugin that I'm developing but the only thing it does is adds remote text tracks. So i believe this in itself doesn't matter. If it turns out it does I'll commit it to github so you can look it over.

ldayananda:
Hmm, seems like this is due to accessing [textTrackSettings without checking if it was included](https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-display.js#L273). Would you be willing to make a PR to fix this?

PR #4974, opened 2018-02-26
Small addition to an if check to fix an error

ookami125:
## Description
Bug fix for issue https://github.com/videojs/video.js/issues/4964

## Specific Changes proposed
added one condition to early out to check if this.player_.textTrackSettings is false if so return as the later code isn't needed.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -278,7 +278,7 @@ class TextTrackDisplay extends Component {
 
     window.WebVTT.processCues(window, cues, this.el_);
 
-    if (this.player_.textTrackSettings === false) {
+    if (!this.player_.textTrackSettings) {
       return;
     }
 
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -266,7 +266,7 @@ class TextTrackDisplay extends Component {
    *        Text track object to be added to the list.
    */
   updateForTrack(track) {
-    if (typeof window.WebVTT !== 'function' || !track.activeCues) {
+    if (typeof window.WebVTT !== 'function' || !track.activeCues || this.player_.textTrackSettings === false) {
       return;
     }
 
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -266,11 +266,10 @@ class TextTrackDisplay extends Component {
    *        Text track object to be added to the list.
    */
   updateForTrack(track) {
-    if (typeof window.WebVTT !== 'function' || !track.activeCues || this.player_.textTrackSettings === false) {
+    if (typeof window.WebVTT !== 'function' || !track.activeCues) {
       return;
     }
 
-    const overrides = this.player_.textTrackSettings.getValues();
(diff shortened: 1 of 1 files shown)


src/js/tracks/text-track-display.js
```diff
@@ -266,7 +266,7 @@ class TextTrackDisplay extends Component {
    *        Text track object to be added to the list.
    */
   updateForTrack(track) {
-    if (typeof window.WebVTT !== 'function' || !track.activeCues) {
+    if (typeof window.WebVTT !== 'function' || !track.activeCues || this.player_.textTrackSettings === false) {
```
gkatsev: instead of exiting this function entirely, what about moving the check to be around `overrides` below?
something like
```js
let overrides = {};
if (this.player_.textTrackSettings) {
  overrides = this.player_.textTrackSettings.getValues();
}
```
That way captions still load properly it just doesn't bother getting overrides.

ookami125: Actually what if we move overrides down after processCues that way that function is called, but then do the early out after that just to avoid the while loop. I mean nothing in the while loop will be run due to overrides being empty so why not skip it all together.

gkatsev: That should be fine as well.

ookami125: what should I do in this case? close the pull request an make a new one? I dont see a way to update the commits of the current one.

ookami125:
Well that should be it.

src/js/tracks/text-track-display.js
```diff
@@ -279,6 +278,12 @@ class TextTrackDisplay extends Component {
 
     window.WebVTT.processCues(window, cues, this.el_);
 
+    if (this.player_.textTrackSettings === false) {
```
ldayananda: I gave this a try and `this.player_.textTrackSettings` is undefined and not false when the TextTrackSettings component is not added. You might want to use something like [player.getChild('TextTrackSettings')](http://docs.videojs.com/component#getChild) to check for it.

gkatsev:
Yeah, the menu item should probably not show up. I'm OK with that happening in another PR not to hold this up.

src/js/tracks/text-track-display.js
gkatsev: oh, thanks for pointing this out. We generally do loose checking so, just 
```js
if (!this.player_.textTrackSettings) {
  // ...
}
```
should be fine

ldayananda: Yep, either works for me

gkatsev:
@ookami125 hey, any updates on this?

gkatsev:
I went ahead and made the change so that it could go in to the next patch release

gkatsev:
Thanks @ookami125!

ookami125:
Sorry for the lack of a reply, I've been a little busy and haven't checked my email until now.

gkatsev:
No worries, the final request was tiny so I just made it to get it into the patch release. It's now in 6.7.4.

PR #4985, opened 2018-03-01
WIP - refactor: webpack based build system

BrandonOCasey:
## Description
With the upcoming release of video.js@7 we may need to switch to webpack for our build system. Awhile ago I had worked on something to do just that. This should help when that work is started.

## What's missing
* A way to build css 
* A way to build all the alternative js files that we currently support:
  * `video.cjs.js` - I think webpack can do this
  * `video.es.js` - I think webpack can do this
  *  the `alt` folder


gkatsev:
Thanks for the inspiration, if we were to go forward with this, it'll be in #5033.

PR #4986, opened 2018-03-01
fix: Fix an issue where disabling the progress control would throw an error.

misteroneill:
I noticed a small issue when the progress control is disabled, this code throws an error because it gets called without an event object.

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -247,7 +247,9 @@ class SeekBar extends Slider {
     super.handleMouseUp(event);
 
     // Stop event propagation to prevent double fire in progress-control.js
-    event.stopPropagation();
+    if (event) {
+      event.stopPropagation();
+    }
     this.player_.scrubbing(false);
 
     /**


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -247,7 +247,9 @@ class SeekBar extends Slider {
     super.handleMouseUp(event);
 
     // Stop event propagation to prevent double fire in progress-control.js
-    event.stopPropagation();
+    if (event) {
```
gkatsev: why do we end up calling this function without an `event`?

misteroneill: https://github.com/videojs/video.js/blob/270a23196a524045db726e099ba1d713026236c2/src/js/control-bar/progress-control/progress-control.js#L130

misteroneill: And then: https://github.com/videojs/video.js/blob/270a23196a524045db726e099ba1d713026236c2/src/js/control-bar/progress-control/progress-control.js#L191

gkatsev: ah, I was trying to find where this could happen but only found cases that pass an `event` object through, but I guess there's a possibility of it not existing.

PR #4990, opened 2018-03-05
Add missing question and fix non-matching links and headers in FAQ

bperel:
This PR brings the following documentation fixes :
- Add missing question in FAQ
- Fix non-matching links and headers in FAQ (fixes the fact that clicking on one of these 2 links doesn't do anything in [http://docs.videojs.com/tutorial-faq.html](http://docs.videojs.com/tutorial-faq.html))

### DIFF
--- a/docs/guides/faq.md
+++ b/docs/guides/faq.md
@@ -23,6 +23,7 @@
 * [Q: Can I turn off Video.js logging?](#q-can-i-turn-off-videojs-logging)
 * [Q: What is a plugin?](#q-what-is-a-plugin)
 * [Q: How do I make a plugin for Video.js?](#q-how-do-i-make-a-plugin-for-videojs)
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks for noticing those two links. I think we would prefer to stick to the new "play" language for those two.

We use `remark` to update our TOCs, guess we forgot to run it after updating the last one. You can run 
```sh
`npm bin`/remark --output -- 'docs/guides/faq.md'
```
to update the faq (or `npm run docs:fix` to fix up all the docs, though, seems like there's a lot of changes, so, only doing one doc at a time would be better).
I did notice that we have a broken piece of markdown in the bottom of the faq page (https://github.com/videojs/video.js/blame/master/docs/guides/faq.md#L348) (to fix just remove `[cors]:`).

gkatsev:
I'm going to close this PR for now, if you update it, we can re-open and merge it in. (You can also open a new PR if you'd like).

PR #4997, opened 2018-03-06
Use CSS grid for caption Setting overlay to avoid cutoff

RevinKey:
## Description
Currently the Caption settings menu has two issues:
1. extra characters added to the classname here: https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-settings.js#L437
2. When the player is smaller than ~450px the menu is squished together and some items are cut off. 

## Enhancements/ Changes
1. Removed the extra characters from the class name so the CSS will target correctly
2. Switch to CSS Grid with minmax column row size to make sure the menu is responsive and looks uniform at all sizes. 
**Before:**
Large Player:
hiddenbuttons

Small Player: 
screenshot2

**After:**
Large Player:
screen shot 2017-12-21 at 12 56 48 pm

Small Player: 
screen shot 2017-12-21 at 12 56 39 pm

### DIFF
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -43,11 +43,15 @@
   .vjs-text-track-settings .vjs-track-settings-controls {
     grid-column: 2;
     grid-row: 2;
-    text-align: unset;
-    vertical-align: unset;
+
   }
+
   .vjs-text-track-settings fieldset span.vjs-opacity {
     margin-left: unset;
+  }
+
+  .vjs-track-setting > select {
+    margin-right: 5px;
   }
 }
 
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -45,17 +45,13 @@
     grid-row: 2;
 
   }
-
-  .vjs-text-track-settings fieldset span.vjs-opacity {
-    margin-left: unset;
-  }
-
-  .vjs-track-setting > select {
-    margin-right: 5px;
-  }
 }
 
 // Form elements
+.vjs-track-setting > select {
+  margin-right: 5px;
+}
+
 .vjs-text-track-settings fieldset {
   margin: 5px;
   padding: 3px;
@@ -64,7 +60,7 @@
 
 .vjs-text-track-settings fieldset span {
   display: inline-block;
-  margin-left: 5px;
+  margin-left: 0px;
 }
 
 .vjs-text-track-settings legend {
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -60,7 +60,6 @@
 
 .vjs-text-track-settings fieldset span {
   display: inline-block;
-  margin-left: 0px;
 }
 
 .vjs-text-track-settings legend {
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -90,13 +90,13 @@
 }
 
 .vjs-track-settings-controls button:hover {
-  color: rgba(#2B333f, 0.75);
+  color: rgba(#2B333F, 0.75);
 }
 
 .vjs-track-settings-controls button {
   background-color: $primary-foreground-color;
   background-image: linear-gradient(-180deg, $primary-foreground-color 88%, $secondary-background-color 100%);
-  color: #2B333f;
+  color: #2B333F;
   cursor: pointer;
   border-radius: 2px;
 }
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -2,7 +2,7 @@
   background-color: $primary-background-color;
   background-color: rgba($primary-background-color, 0.75);
   color: $primary-foreground-color;
-  height: 100%;
+  height: 70%;
 }
 
 // Layout divs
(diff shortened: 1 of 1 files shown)


RevinKey:
This is a new PR for #4835 since I somehow broke my branch

gkatsev:
Just tried this out on the live test page: https://deploy-preview-4997--videojs-docs.netlify.com/test-example/
and the right column isn't aligned correctly.

src/css/components/_captions-settings.scss
```diff
@@ -21,6 +21,48 @@
   vertical-align: bottom;
 }
 
+// code that will only run if CSS Grid is supported by the browser
+@supports (display: grid) {
+  .vjs-text-track-settings .vjs-modal-dialog-content {
+    display: grid;
+    grid-template-columns: 1fr auto;
```
gkatsev: the second auto should be `1fr` here

src/css/components/_captions-settings.scss
```diff
@@ -21,6 +21,48 @@
...
+
+  .vjs-text-track-settings .vjs-track-settings-font {
+    grid-column: 2;
+    grid-row: 1;
+  }
+
+  .vjs-text-track-settings .vjs-track-settings-controls {
+    grid-column: 2;
+    grid-row: 2;
+    text-align: unset;
```
gkatsev: why are these two unset?

src/css/components/_captions-settings.scss
```diff
@@ -35,7 +77,7 @@
 
 .vjs-text-track-settings legend {
   color: $primary-foreground-color;
-  margin: 0 0 5px 0;
+  margin: 0 0 5px;
```
gkatsev: can you revert this change and the ones below it? It'll keep the PR focused on just necessary code changes.

src/css/components/_captions-settings.scss
gkatsev: not unsetting them keeps the buttons visually at the bottom right where they should be.

src/css/components/_captions-settings.scss
```diff
@@ -21,6 +21,48 @@
...
+    grid-row: 2;
+    text-align: unset;
+    vertical-align: unset;
+  }
+  .vjs-text-track-settings fieldset span.vjs-opacity {
+    margin-left: unset;
+  }
+  .video-js {
+    @media (max-width: 365px) {
+      .vjs-modal-dialog .vjs-modal-dialog-content {
```
gkatsev: we shouldn't set generic modal-dialog properties inside this file. We should target the text-track-settings specifically.

gkatsev: Also, this doesn't seem to be getting applied?

src/css/components/_captions-settings.scss
RevinKey: Yup my code beautifier messed up a few things like these. Working right now to get it back to how it should be

src/css/components/_captions-settings.scss
RevinKey: I had this in there to resize the font if the player was a certain size. but your are correct doesn't seem to work, nor make a beneficial difference in appearance when I just tested manually

src/css/components/_captions-settings.scss
RevinKey: I thought it gave it a weird 3 column effect. But i can revert this if you'd like.

gkatsev: yeah, I think I'd prefer to keep it as is, we can change it up a bit later on if that was wanted.

src/css/components/_captions-settings.scss
```diff
@@ -21,6 +21,36 @@
...
+  }
+
+  .vjs-text-track-settings .vjs-track-settings-controls {
+    grid-column: 2;
+    grid-row: 2;
+    text-align: unset;
+    vertical-align: unset;
+  }
+  .vjs-text-track-settings fieldset span.vjs-opacity {
+    margin-left: unset;
```
gkatsev: unsetting the margin makes the dropdowns too close. Should probably be removed too.

RevinKey: just updated it to give the margin-right on the <select> and keep the left margin unset. this way it will have a gap between them still and be aligned vertically on small players

src/css/components/_captions-settings.scss
```diff
@@ -21,6 +21,40 @@
...
+  }
+
+  .vjs-text-track-settings .vjs-track-settings-controls {
+    grid-column: 2;
+    grid-row: 2;
+
+  }
+
+  .vjs-text-track-settings fieldset span.vjs-opacity {
+    margin-left: unset;
```
gkatsev: instead of having this and the margin-right, should we just update the default styling below to have a margin-right?

RevinKey: Just tried it out and it works good in both the grid and table moving this to the default style.

src/css/components/_captions-settings.scss
```diff
@@ -30,7 +60,7 @@
 
 .vjs-text-track-settings fieldset span {
   display: inline-block;
-  margin-left: 5px;
+  margin-left: 0px;
```
gkatsev: I think this line is unnecessary now.

RevinKey: yes you are correct, just pushed to remove this.

bcbclifford:
This is a step in the right direction but more is needed for smaller players. Items are still cropped at ~450px for a 16:9 player. We could reduce padding, switch to a single column and show the scrollbar.

gkatsev:
@bcbclifford those are good changes, I'm not sure we want to make them as part of this PR though.

PR #5013, opened 2018-03-14
fix: load sourceset src

BrandonOCasey:
## Description
Currently `sourceset` on `load()` does not have the current source on the event object. This seeks to do the source selection algorithm to determine what source will be selected.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -194,19 +194,19 @@ class Html5 extends Tech {
 
     const chooseSourceEl = (sources) => {
       for (let i = 0; i < sources.length; i++) {
-        if (!sources[0].src) {
+        if (!sources[i].src) {
           continue;
         }
 
-        if (sources[0].type && !Html5.canPlayType(sources[0].type)) {
+        if (sources[i].type && !Html5.canPlayType(sources[i].type)) {
           continue;
         }
 
-        if (sources[0].media) {
+        if (sources[i].media) {
           // ignored, as it isn't used for videos
         }
 
-        return sources[0].src;
+        return sources[i].src;
       }
 
       return '';
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -192,10 +192,39 @@ class Html5 extends Tech {
 
     const oldLoad = el.load;
 
+    const chooseSourceEl = (sources) => {
+      for (let i = 0; i < sources.length; i++) {
+        if (!sources[0].src) {
+          continue;
+        }
+
+        if (sources[0].type && !Html5.canPlayType(sources[0].type)) {
+          continue;
+        }
+
+        if (sources[0].media) {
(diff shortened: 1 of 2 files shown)


src/js/tech/html5.js
```diff
@@ -192,10 +192,39 @@ class Html5 extends Tech {
...
+
+        if (sources[i].type && !Html5.canPlayType(sources[i].type)) {
+          continue;
+        }
+
+        if (sources[i].media) {
+          // ignored, as it isn't used for videos
+        }
+
+        return sources[i].src;
```
gkatsev: the problem is something like:
```html
  <video >
    <source src="oceans.flv">
    <source src="oceans.mp4">
  </video>
```
which the browser will select oceans.mp4 but here we'll end up selecting oceans.flv.

BrandonOCasey: `Html5.canPlayType` will return `maybe` or `possibly` for flv?

gkatsev: we don't have a type for the source.

gkatsev: And we can't really rely on the extension either because it is possible it won't exist.
```
  <video >
    <source src="./oceansflv">
    <source src="./oceansmp4">
  </video>
```
will also have the second source chosen.

BrandonOCasey: yeah your right, let me see what the browser does in some of these cases

gkatsev: it requests the first chunk of each file until it gets one that it can play.
screen shot 2018-03-14 at 11 50 15

gkatsev: even better, seems like browsers ignore the type anyway, or at least verify after the fact:
```
  <video >
    <source src="./oceansflv" type="video/mp4">
    <source src="./oceansmp4" type="video/mp4">
  </video>
```
still loads `oceansmp4`

BrandonOCasey: So what if we do it like this:
1. Get all sources
2. Loop through them, if we find one with a type that we can play return that one
3. If we find one without a type before one we can play return empty sting

gkatsev: using the src attribute might be ok but the issue is that we can't be sure which source will be used because types are unreliable, as seen in the example in the previous comment.

gkatsev: So, I spoke with @dmlap and @incompl about this. I think what we'd want to do is use the src attribute if it's available and otherwise have it set to the empty string.
Also, I'm thinking of marking this as the first experimental API we have so we are at liberty to fiddle with it in minor releases.

gkatsev:
Closed in favor of https://github.com/videojs/video.js/pull/5016

src/js/tech/html5.js
BrandonOCasey: ok sounds good to me, the more I look at it the more I think thats the way we should go.

PR #5016, opened 2018-03-14
fix(sourceset): set evt.src to empty string or src attr from load

gkatsev:
If `.load()` is called, we don't know what the new source will be because of the asynchronous nature of the resource selection algorithm. However, we can know if the src attribute is set on the video element because that's what will be chosen. So, set the src property on the event object to be the src attribute or be the empty string. This matches how `currentSrc` works where it is an empty string when no source is set. So, when `sourceset` is triggered and `src` is `""` in the event object, it means that the source is changing.
In addition, we're going to be releasing `sourceset` as an *experimental* feature so we could improve the API and make changes in minor releases.

TODO:
* [x] More documentation for this change

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -195,7 +195,10 @@ class Html5 extends Tech {
     el.load = () => {
       const retval = oldLoad.call(el);
 
-      this.triggerSourceset(el.src || el.currentSrc);
+      // if `el.src` is set, that source will be loaded
+      // otherwise, we can't know for sure what source will be set, so,
+      // instead return an empty string to basically indicate source may change
+      this.triggerSourceset(el.src || '');
 
       return retval;
     };
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -196,7 +196,9 @@ class Html5 extends Tech {
       const retval = oldLoad.call(el);
 
       // if `el.src` is set, that source will be loaded
-      // otherwise, we can't know for sure what source will be set, so,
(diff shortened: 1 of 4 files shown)


gkatsev:
@videojs/core-committers should we make "experimental" features like this available by default or should it be opt-in via some option?

src/js/tech/html5.js
```diff
@@ -195,7 +195,10 @@ class Html5 extends Tech {
     el.load = () => {
       const retval = oldLoad.call(el);
 
-      this.triggerSourceset(el.src || el.currentSrc);
+      // if `el.src` is set, that source will be loaded
+      // otherwise, we can't know for sure what source will be set, so,
```
BrandonOCasey: it could be good to mention that: `<source>` elements will be used if the `src` attribute isn't set, and we cannot track that because that happens async without a good way of interception.

BrandonOCasey:
I think that I would want it in by default, but I don't know if I have a bias here.

gkatsev:
This is how the sourcset docs will look like currently: https://deploy-preview-5016--videojs-docs.netlify.com/player#event:sourceset

misteroneill:
I don't have a strong feeling about it being an explicit opt-in, but that does give us more leeway in changing it in non-major releases.

ldayananda:
I think it's safer for it to be opt-in while it's still in the experimental stage.

gkatsev:
Merged this in, but we still need to iron out the experimental stuff before a release can be made.

PR #5024, opened 2018-03-15
Queue playback events when the playback rate is zero

squarebracket:
## Description
SourceHandlers that use MSE have a problem: if they push a segment into a `SourceBuffer` and then seek close to the end, playback will stall and/or there will be a massive downswitch in quality. The general approach to fixing this that was discussed on slack was by setting the playback rate of the player to zero, buffering all that was required, and then restoring the previous playback rate. In my implementation, I've done this in the source handler (see: videojs/videojs-contrib-hls#1374).

From the video.js perspective, it should ensure that the UI reflects the buffering status and that the player API behaves like you'd expect -- that is to say, that it will fire `seeking` immediately after a call to `currentTime`, and it will fire `seeked`, `canplay`, `canplaythrough`, and `playing` when everything is buffered.

## Specific Changes proposed
- When the playback rate of the tech is zero, queue up callbacks for the `seeked`, `canplay`, `canplaythrough`, and `playing` events
- When the playback rate is changed to something non-zero and its previous playback rate was zero, fire the queued callbacks

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -229,6 +229,8 @@ const TECH_EVENTS_RETRIGGER = [
 ];
 
 // events to queue when playback rate is zero
+// this is a hash for the sole purpose of mapping non-camel-cased event names
+// to camel-cased function names
 const TECH_EVENTS_QUEUE = {
   canplay: 'CanPlay',
   canplaythrough: 'CanPlayThrough',
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -966,15 +966,15 @@ class Player extends Component {
     });
 
     Object.keys(TECH_EVENTS_QUEUE).forEach((event) => {
-      this.on(this.tech_, event, () => {
+      this.on(this.tech_, event, (eventObj) => {
         if (this.tech_.playbackRate() === 0 && this.tech_.seeking()) {
           this.queuedCallbacks_.push({
             callback: this[`handleTech${TECH_EVENTS_QUEUE[event]}_`].bind(this),
-            event
+            event: eventObj
           });
           return;
         }
-        this[`handleTech${TECH_EVENTS_QUEUE[event]}_`](event);
+        this[`handleTech${TECH_EVENTS_QUEUE[event]}_`](eventObj);
       });
     });
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -387,6 +387,9 @@ class Player extends Component {
 
     this.el_ = this.createEl();
 
+    // Set default value for lastPlaybackRate
+    this.cache_.lastPlaybackRate = this.defaultPlaybackRate();
+
     // Make this an evented object and use `el_` as its event bus.
     evented(this, {eventBusKey: 'el_'});
 
@@ -1297,11 +1300,11 @@ class Player extends Component {
    * @listens Tech#ratechange
    */
   handleTechRateChange_() {
-    if (this.tech_.playbackRate() > 0 && this.previousPlaybackRate === 0) {
+    if (this.tech_.playbackRate() > 0 && this.cache_.lastPlaybackRate === 0) {
       this.queuedCallbacks_.forEach((queued) => queued.callback(queued.event));
       this.queuedCallbacks_ = [];
     }
-    this.previousPlaybackRate = this.tech_.playbackRate();
+    this.cache_.lastPlaybackRate = this.tech_.playbackRate();
     /**
      * Fires when the playing speed of the audio/video is changed
      *
@@ -3101,12 +3104,14 @@ class Player extends Component {
    */
   playbackRate(rate) {
     if (rate !== undefined) {
+      // NOTE: this.cache_.lastPlaybackRate is set from the tech handler
+      // that is registered above
       this.techCall_('setPlaybackRate', rate);
       return;
     }
 
     if (this.tech_ && this.tech_.featuresPlaybackRate) {
-      return this.techGet_('playbackRate');
+      return this.cache_.lastPlaybackRate || this.techGet_('playbackRate');
     }
     return 1.0;
   }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -244,6 +244,13 @@ const TECH_EVENTS_RETRIGGER = [
   'texttrackchange'
 ];
 
+const TECH_EVENTS_QUEUE = {
+  canplay: 'CanPlay',
+  canplaythrough: 'CanPlayThrough',
+  playing: 'Playing',
+  seeked: 'Seeked'
+};
+
 /**
  * An instance of the `Player` class is created when any of the Video.js setup methods
  * are used to initialize a video.
@@ -319,6 +326,9 @@ class Player extends Component {
 
     // Tracks when a tech changes the poster
     this.isPosterFromTech_ = false;
+
+    // TODO: Docs
+    this.queuedCallbacks_ = [];
 
     // Turn off API access because we're loading a new tech that might load asynchronously
     this.isReady_ = false;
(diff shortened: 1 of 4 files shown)


asd29696:
2018-03-16 6:54 GMT+08:00 Chuck Wilson <notifications@github.com>:

src/js/player.js
```diff
@@ -985,6 +1002,7 @@ class Player extends Component {
     this.on(this.tech_, 'loadedmetadata', this.updateStyleEl_);
     this.on(this.tech_, 'posterchange', this.handleTechPosterChange_);
     this.on(this.tech_, 'textdata', this.handleTechTextData_);
+    this.on(this.tech_, 'ratechange', this.handleTechRateChange_);
```
gkatsev: if we want to run special handling, it needs to be removed from https://github.com/videojs/video.js/blob/master/src/js/player.js#L196 and the documentation needs to be moved to the handleTechRateChange_.

src/js/player.js
```diff
@@ -967,15 +977,22 @@ class Player extends Component {
     TECH_EVENTS_RETRIGGER.forEach((event) => {
       this.on(this.tech_, event, this[`handleTech${toTitleCase(event)}_`]);
     });
+
+    Object.keys(TECH_EVENTS_QUEUE).forEach((event) => {
+      this.on(this.tech_, event, () => {
+        if (this.tech_.playbackRate() === 0) {
+          this.queuedCallbacks_.push(this[`handleTech${TECH_EVENTS_QUEUE[event]}_`].bind(this));
```
gkatsev: the event object should be passed through as well.

src/js/player.js
```diff
@@ -967,15 +977,22 @@ class Player extends Component {
...
+          return;
+        }
+        this[`handleTech${TECH_EVENTS_QUEUE[event]}_`]();
```
gkatsev: event object should be passed through as well.

src/js/player.js
```diff
@@ -967,15 +977,22 @@ class Player extends Component {
     TECH_EVENTS_RETRIGGER.forEach((event) => {
       this.on(this.tech_, event, this[`handleTech${toTitleCase(event)}_`]);
     });
+
+    Object.keys(TECH_EVENTS_QUEUE).forEach((event) => {
+      this.on(this.tech_, event, () => {
+        if (this.tech_.playbackRate() === 0) {
```
gkatsev: I think this would be a breaking change. Potentially, doing it when `isSeeking()` is true *and* `playbackRate` is `0`, might be OK.

squarebracket: Could you elaborate what would/could be breaking? A tech or source handler would have to both set the playback rate to zero and rely on these events being dispatched by the _player_ (the tech itself will still dispatch them) right after the seek. Perhaps there's some use case I'm not thinking of.

I don't think that would work. Correct me if I'm wrong but I think `seeking()` is only true between the `seeking` and `seeked` events, and this needs to queue everything after and including the `seeked` event.

gkatsev: Because people could still expect these events even when `playbackRate` is `0`.
Maybe we need to allow techs/source handlers dictate when seeking is happening? Looks like `seeking()` method just asks the video element directly:
https://github.com/videojs/video.js/blob/d7f45bab89790fe84f8fcc162df2b9e2ff6d265b/src/js/tech/html5.js#L1552-L1564
https://github.com/videojs/video.js/blob/d7f45bab89790fe84f8fcc162df2b9e2ff6d265b/src/js/tech/html5.js#L1700-L1704

but we could modify it so that sourcehandlers can handle it as well like `duration` and `seekable`? https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L1166-L1173

squarebracket: > Looks like seeking() method just asks the video element directly

Yeah, that's why I said it wouldn't work. I like the idea of deferring `seeking()` to the source handlers. In fact, that should probably be done anyway to make the API correct -- i.e. you'd expect `seeking()` to be true up to the point a `seeked` is fired but that's not the case with the current code. I'll make that change and gate the event queueing to it.

Any other APIs you can think of that might need to be deferred to spoof things properly?

gkatsev: Not sure, but we've been deferring them as needed, we just haven't really needed any others until now, apparently.
Do you think that deferring to SH and then only queueing the events if `seeking()` works?

squarebracket: Yeah, that should work.

If you want to go with override-only behaviour, there's potentially another possibility. We could rejigger player.js so that most (all?) of the event handlers are dispatched on the player event rather than the tech event. And then have something like `tech.delegateEvent(event)`. An undelegated event would then just be retriggered on the player, whereas a delegated event would be triggered on a source handler -- and expected to eventually be fired on the player manually be the source handler. I did a quick test and it worked as expected, but you obviously have a better idea about the bigger implications of such a change.

gkatsev: I think that deferring `seeking()` makes sense in this case.

I'm not sure I follow this 100%. If I understand correctly, the player would not listen for tech events but rather the tech would trigger the events directly on the player when necessary? If that is the case, I see two issues.
1. the tech and SH aren't supposed to have a reference to the player, so, they can't trigger an event on it.
2. triggering the event on the player directly would mean that the player doesn't necessarily have a chance to do something related to the event before others would receive the event, plus, if the player would to re-trigger the event, it would end up in an event loop.

This also does lend itself a bit to what middleware strives to be and we've been thinking/wanting to have events be able to be intercepted by mw as well, so, maybe we can do that eventually. I think that deferring `seeking()` is the quick solution and a decent solution to this.

squarebracket: Yeah, I was talking about removing `this.trigger` to avoid the event loop, but you got the gist of what I was going for.

gkatsev: Cool. Yeah, I think that's a bit more complicated and would require a lot of though to get it right.

gkatsev:
Also, this probably needs to be changed to be a PR against 6.x, master will soon be a 7.x-only. Or you can leave it against master and we can backport it later on.

squarebracket:
I'll have to open the PR against any VJS branch that contrib-hls/VHS supports, since the PR there will require the change in video.js.

src/js/player.js
```diff
@@ -967,15 +964,25 @@ class Player extends Component {
     TECH_EVENTS_RETRIGGER.forEach((event) => {
       this.on(this.tech_, event, this[`handleTech${toTitleCase(event)}_`]);
     });
+
+    Object.keys(TECH_EVENTS_QUEUE).forEach((event) => {
+      this.on(this.tech_, event, () => {
+        if (this.tech_.playbackRate() === 0 && this.tech_.seeking()) {
+          this.queuedCallbacks_.push({
+            callback: this[`handleTech${TECH_EVENTS_QUEUE[event]}_`].bind(this),
+            event
```
gkatsev: the event that we want to cache is actually event object that's passed to the handler (the argument that would be passed into this arrow function), not the event name

test/unit/player.test.js
```diff
@@ -1364,6 +1364,49 @@ QUnit.test('Remove waiting class on timeupdate after tech waiting', function(ass
   player.dispose();
 });
 
+QUnit.test('Queues playing events when playback rate is zero', function(assert) {
```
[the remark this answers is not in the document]
squarebracket: Might as well. Added tests.

src/js/player.js
```diff
@@ -1276,6 +1285,32 @@ class Player extends Component {
...
+   * @method Player#handleTechRateChange_
+   * @fires Player#ratechange
+   * @listens Tech#ratechange
+   */
+  handleTechRateChange_() {
+    if (this.tech_.playbackRate() > 0 && this.previousPlaybackRate === 0) {
+      this.queuedCallbacks_.forEach((queued) => queued.callback(queued.event));
+      this.queuedCallbacks_ = [];
+    }
+    this.previousPlaybackRate = this.tech_.playbackRate();
```
misteroneill: Should this be pseudo-private (i.e. `previousPlaybackRate_`)?

gkatsev: totally missed that, yes, it should. Could even potentially be stored on the cache_ object as well/instead.

squarebracket: Changed to `this.cache_.lastPlaybackRate`

src/js/player.js
```diff
@@ -389,6 +387,9 @@ class Player extends Component {
 
     this.el_ = this.createEl();
 
+    // Set default value for lastPlaybackRate
+    this.cache_.lastPlaybackRate = this.defaultPlaybackRate();
```
squarebracket: Is it appropriate to use `defaultPlaybackRate` here? i.e. will everything be set up properly at this point? Since the element exists, I made the assumption that it would.

src/js/player.js
squarebracket: Moved the documentation.

squarebracket:
LMK if there's somewhere in the docs these changes should be reflected.

src/js/player.js
gkatsev: some techs, namely flash, don't support playbackrate. However, if it isn't supported or the tech isn't ready, it'll just return the default value of 1.0, so, this is probably fine. https://github.com/videojs/video.js/blob/4e79a04579d8b3748ddd4c42707fd5a8bf68cce8/src/js/player.js#L3090-L3099

gkatsev:
It doesn't seem like we have any docs for source handlers, so, not really any place to put it currently.

gkatsev:
@forbesjo brought up that when contrib-hls is updated with this, what happens if a user is using a version of Video.js without this change? Would this make the contrib-hls update a breaking change or can we make the contrib-hls backwards compatible?

squarebracket:
I figured it would require a bump in the dependency version. So we'd wait until this change is merged in and released, and then the changes for contrib-hls would be merged in with a change to its package.json.

Is that a problem? I imagine that if you're updating contrib-hls, it's probably not that big of a deal to update video.js as well.

Otherwise we'll have to do some feature detection. A `tech.supportsProperSeek` or some such thing.

squarebracket:
Care to elaborate what you mean by making the choice explicitly?

gkatsev:
As in, we should decide what to do in this case, from making a breaking change on contrib-hls, or just upping the minimum requires video.js version in a minor, to making contrib-hls work with or without this change (with degraded behavior without this change). What we shouldn't do is just merge and release and let the cards fall where they may. Though, choosing that could be an option too.

forbesjo:
Upon further discussion we'll make it very clear in the contrib-hls README and package.json which version of video.js this change is compatible with.

gkatsev:
Thanks @squarebracket!

PR #5027, opened 2018-03-16
fix: remove button outlines

Michael77:
## Description
This removes the button outlines that look out of place around the play and volume controls (and perhaps others)

## Specific Changes proposed
Removes the button outlines for the vjs-control class

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_control.scss
+++ b/src/css/components/_control.scss
@@ -8,8 +8,8 @@
   padding: 0;
   height: 100%;
   width: 4em;
+  outline: 0;
   @include flex(none);
-
 }
 .vjs-button > .vjs-icon-placeholder:before {
   font-size: 1.8em;


gkatsev:
Thanks for the PR. However, removing these outlines will make the player less accessible and that isn't what we're striving for. Instead, we recommend you implement this yourself if you do not want the outlines in your player. However, for the sake of accessibility, I would urge you not to do so.

Michael77:
In what way would it make it less accessible? Youtube and Vimeo don't use button outlines around their player controls, which is worth mentioning.

mister-ben:
It's important for keyboard control. See e.g. http://www.outlinenone.com

It's not so much we can't remove it as we shouldn't without an alternative. We could have nicer focus styles. YouTube's approach is to only style for focus when there's been keyboard interaction, which seems a reasonable compromise between aesthetics and accessibility.

gkatsev:
We do want to make it so that it is only shown for keyboard usage but, unfortunately, we haven't gotten a chance to get that in yet.

Michael77:
Thank you for the informative replies, I will read into this topic more.

gkatsev:
Thanks @mister-ben for replying. I should've been more informative about this in my initial reply.

PR #5031, opened 2018-03-19
feat: require enableSourceset option for event

gkatsev:
While sourceset is experimental, require the option `enableSourceset`.
This can easily be reversed in a minor update.

Going forward with this because most of the core committers preferred a flag for this experimental feature and this gets it out the door sooner and can be flipped easily in the future.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -34,7 +34,9 @@ class Html5 extends Tech {
   constructor(options, ready) {
     super(options, ready);
 
-    this.setupSourcesetHandling_();
+    if (options.enableSourceset) {
+      this.setupSourcesetHandling_();
+    }
 
     const source = options.source;
     let crossoriginTracks = false;
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -163,6 +163,10 @@ class Tech extends Component {
    * @param {string} src The source string at the time of the source changing.
    */
   triggerSourceset(src) {
+    if (!this.options_.enableSourceset) {
+      return;
+    }
+
     if (!this.isReady_) {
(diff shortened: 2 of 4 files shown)


src/js/tech/tech.js
```diff
@@ -163,6 +163,10 @@ class Tech extends Component {
    * @param {string} src The source string at the time of the source changing.
    */
   triggerSourceset(src) {
+    if (!this.options_.enableSourceset) {
```
BrandonOCasey: I don't know if we should do this here. It makes manually using `sourceset` after init a bit weird. I guess we won't expect users to call `setupSourcesetHandling_()` on their own, but I think it is something that I would do. This would require setting an option and calling that function.

src/js/tech/html5.js
```diff
@@ -34,7 +34,9 @@ class Html5 extends Tech {
   constructor(options, ready) {
     super(options, ready);
 
-    this.setupSourcesetHandling_();
+    if (options.enableSourceset) {
```
BrandonOCasey: I am wondering what you think about moving the `featuresSourceSet` check here instead of inside of the function? seems closer to how our other feature checks work.

gkatsev: all the features are set more or less statically, then there's a secondary check as to whether we want to use the feature or not.

src/js/tech/tech.js
gkatsev: Yeah, we don't expect and not recommend people call private/suffixed methods like that, though, I'd be ok with removing this bit regardless.

src/js/tech/html5.js
BrandonOCasey: ok

PR #5037, opened 2018-03-21
refactor: move sourceset code out of tech

BrandonOCasey:
## Description
The `sourceset` code is growing a bit large, and some of the parts are going to be `re-usable` in the prs that follow this one. I think it makes sense to split it into another file.


gkatsev:
also, as we merge these, we'd want to cherry-pick or make a new PR against the 6.x branch as well.

PR #5038, opened 2018-03-21
Fix/sourceset append source

BrandonOCasey:
## Description
In Chrome/Firefox/Safari appending a `<source>` element when the media element has no source, causes a `sourceset`, but our code does not.
## Specific Changes proposed
1. If there is not a `src` or `<source>` elements when adding `sourceset` support. Patch `innerHTML`, `append`, `appendChild` and `insertAdjacentHTML` so that they call the internal function, but also check if a source was set.
1. On the first `sourceset` un-patch these funcitons.


src/js/tech/setup-sourceset.js
```diff
@@ -1,21 +1,121 @@
...
 
-  const el = tech.el();
+  /**
+   * Since there isn't a src property on the media element, source elements will be used for
+   * implementing the source selection algorithm. This happens asynchronously and
+   * for most cases were there is more than one source we cannot tell what source will
+   * be loaded, without re-implementing the source selection algorithm. At this time we are not
+   * going to do that. There are three special cases that we do handle here though:
+   *
+   * 1. If there are no sources, do not fire `sourceset`.
```
BrandonOCasey: This changed the behavior on `load` slightly, but I think the comment does a good job explaining how.

src/js/tech/setup-sourceset.js
```diff
@@ -42,42 +161,175 @@ const setupSourceset = function(tech) {
...
+  const el = tech.el();
+  const srcDescriptor = getSrcDescriptor(el);
+  const oldSetAttribute = el.setAttribute;
+  const oldLoad = el.load;
+
+  // we need to fire sourceset when the player is ready
+  // if we find that the media element had a src when it was
+  // given to us and that tech element is not in a stalled state
+  if (el.src || el.currentSrc && el.initNetworkState_ !== 3) {
+    tech.triggerSourceset(el.src || el.currentSrc);
```
gkatsev: should we just call `sourcesetLoad` here instead?

src/js/tech/setup-sourceset.js
```diff
@@ -42,42 +161,175 @@ const setupSourceset = function(tech) {
...
+    el.firstSourceWatch_ = false;
+    el.appendChild = oldAppendChild;
+    if (oldAppend) {
+      el.append = oldAppend;
+    }
+    if (oldInsertAdjacentHTML) {
+      el.insertAdjacentHTML = oldInsertAdjacentHTML;
+    }
+
+    Object.defineProperty(el, 'innerHTML', {
```
gkatsev: can't we just do `Object.defineProperty(el, 'innerHTML', innerDescriptor)`?

src/js/tech/setup-sourceset.js
```diff
@@ -42,42 +161,175 @@ const setupSourceset = function(tech) {
...
+
+      return retval;
+    },
+    configurable: true,
+    enumerable: innerDescriptor.enumerable
+  });
+
+  // on the first sourceset, we need to revert
+  // our changes
+  tech.one('sourceset', (e) => {
```
gkatsev: and because all of the things above trigger sourceset, we'd reset this once the first source was set?

test/unit/sourceset.test.js
```diff
@@ -276,15 +300,263 @@ QUnit[qunitFn]('sourceset', function(hooks) {
...
+        this.source.type = this.testSrc.type;
+
+        this.player = videojs(this.mediaEl, {enableSourceset: true});
+
+        this.player.one('sourceset', (e) => {
+          validateSource(assert, this.player, [this.testSrc], false);
+
+          this.player.one('sourceset', (e2) => {
+            // should be the same event
+            validateSource(assert, this.player, [this.testSrc], false);
```
gkatsev: is this a test that we want to assert based on the value of `e.src` and `e2.src`? because otherwise this doesn't doesn't do anything beyond checking that two sourceset events fired.

test/unit/sourceset.test.js
```diff
@@ -443,6 +715,7 @@ QUnit[qunitFn]('sourceset', function(hooks) {
     });
 
     QUnit.test('mediaEl.load()', function(assert) {
+      const done = assert.async();
```
gkatsev: this shouldn't be necessary since everything happens synchronously.

test/unit/sourceset.test.js
```diff
@@ -471,15 +745,17 @@ QUnit[qunitFn]('sourceset', function(hooks) {
 
     QUnit.test('mediaEl.load() x2 at the same time', function(assert) {
       const source = document.createElement('source');
+      const done = assert.async();
```
gkatsev: same here, everything should be happening async

test/unit/sourceset.test.js
```diff
@@ -471,15 +745,17 @@ QUnit[qunitFn]('sourceset', function(hooks) {
...
 
       source.src = this.sourceOne.src;
       source.type = this.sourceOne.type;
 
       this.player.one('sourceset', (e1) => {
-        assert.equal(e1.src, '', 'we got a sourceset with an empty src');
+        assert.equal(e1.src, this.sourceOne.src, 'we got a sourceset with an empty src');
```
gkatsev: the test message needs an update

src/js/tech/setup-sourceset.js
```diff
@@ -42,42 +161,175 @@ const setupSourceset = function(tech) {
...
+  // calls `load` internally right away. We need to handle that.
+  if (!el.src && !el.currentSrc && !tech.$$('source').length) {
+    firstSourceWatch(tech);
+  }
+
   Object.defineProperty(el, 'src', {
     get: srcDescriptor.get.bind(el),
     set: (v) => {
-      const retval = srcDescriptor.set.call(el, v);
+      const retval = srcDescriptor.set.call(tech.el(), v);
```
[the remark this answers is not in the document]
BrandonOCasey: I'm not sure, I will revert this

src/js/tech/setup-sourceset.js
BrandonOCasey: yeah I think we should

src/js/tech/setup-sourceset.js
BrandonOCasey: yes

test/unit/sourceset.test.js
BrandonOCasey: yeah your right I will revert this. This was changed in the `update player cache` branch, and validateSource actually takes the event in that one. I will put this back

test/unit/sourceset.test.js
BrandonOCasey: agreed

src/js/tech/setup-sourceset.js
```diff
@@ -42,42 +161,175 @@ const setupSourceset = function(tech) {
...
+  // we need to fire sourceset when the player is ready
+  // if we find that the media element had a src when it was
+  // given to us and that tech element is not in a stalled state
+  if (el.src || el.currentSrc && el.initNetworkState_ !== 3) {
+    tech.triggerSourceset(el.src || el.currentSrc);
+  }
+
+  // for some reason adding a source element when a mediaElement has no source
+  // calls `load` internally right away. We need to handle that.
+  if (!el.src && !el.currentSrc && !tech.$$('source').length) {
```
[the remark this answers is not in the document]
BrandonOCasey: OK so I tried this and I do think that it would make sense, but it has a bad interaction with:
https://github.com/videojs/video.js/blob/378d98ee87b6ac48536d0b8f263bee1f76c68aab/src/js/tech/html5.js#L49

and I don't think we want to change that code right now. So I will be keeping everything as is and open another pr after this one to make this change.

test/unit/sourceset.test.js
```diff
@@ -565,38 +835,27 @@ QUnit[qunitFn]('sourceset', function(hooks) {
...
-            assert.equal(e3.src, sourceOne.src, 'the third sourceset is the first source');
-            sourcesets++;
-
-            assert.equal(sourcesets, 2, 'two sourcesets');
-            player.dispose();
-            delete Tech.techs_.FakeFlash;
-            done();
-          });
-        });
+      // the second sourceset ends up being the second source because when the first source is set
```
gkatsev: looks like I messed up this comment.
I believe the first sourceset will be the second source so it should be:
> the *first* sourceset ends up being the second source

src/js/tech/setup-sourceset.js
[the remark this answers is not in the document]
gkatsev: making it in another PR sounds OK.

test/unit/sourceset.test.js
```diff
@@ -276,14 +300,260 @@ QUnit[qunitFn]('sourceset', function(hooks) {
...
+        const done = assert.async();
+
+        this.source = document.createElement('source');
+        this.source.src = this.testSrc.src;
+        this.source.type = this.testSrc.type;
+
+        this.player = videojs(this.mediaEl, {enableSourceset: true});
+
+        this.player.one('sourceset', (e) => {
+          validateSource(assert, this.player, [this.testSrc], false);
```
gkatsev: I'm concerned that these tests that don't assert inside of `validateSource` (the last false being passed in), don't actually verify things we want verified, at least if that's the only call to `validateSource` without another assertion.

test/unit/sourceset.test.js
```diff
@@ -276,14 +300,260 @@ QUnit[qunitFn]('sourceset', function(hooks) {
...
+        this.source.type = this.sourceOne.type;
+
+        this.source2 = document.createElement('source');
+        this.source2.src = this.sourceTwo.src;
+        this.source2.type = this.sourceTwo.type;
+
+        this.player = videojs(this.mediaEl, {enableSourceset: true});
+
+        this.player.one('sourceset', (e) => {
+          validateSource(assert, this.player, [this.sourceOne], false);
```
gkatsev: here as well

PR #5039, opened 2018-03-21
Fix: `video-js` el sourceset

BrandonOCasey:
## Description
Currently properties from `video-js` elements are not copied over to the `video` element when we do that. This means doing things like `videojsEl.src = ...;` will not carry over if they are done before player init. This also add `video-js` testing to our `sourceset` tests.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -568,6 +568,13 @@ class Player extends Component {
       el.appendChild(tag);
 
       playerElIngest = this.playerElIngest_ = el;
+      // move properties over from our custom `video-js` element
+      // to our new `video` element. This will move things like
+      // `src` or `controls` that were set via js before the player
+      // was initialized.
+      Object.keys(el).forEach((k) => {
+        tag[k] = el[k];
+      });
     }
 
     // set tabindex to -1 so we could focus on the player element
--- a/test/unit/sourceset.test.js
+++ b/test/unit/sourceset.test.js
@@ -34,7 +34,7 @@ const validateSource = function(assert, player, sources, checkMediaElSource = tr
 const setupEnv = function(env, testName) {
   env.fixture = document.getElementById('qunit-fixture');
 
-  if (testName === 'change video el' || testName === 'change audio el') {
+  if ((/^change/i).test(testName)) {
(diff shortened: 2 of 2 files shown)


src/js/player.js
```diff
@@ -568,6 +568,13 @@ class Player extends Component {
       el.appendChild(tag);
 
       playerElIngest = this.playerElIngest_ = el;
+      // move properties over from our custom `video-js` element
+      // to our new `video` element. This will move things like
+      // `src` or `controls` that were set via js before the player
+      // was initialized.
+      Object.keys(el).forEach((k) => {
```
gkatsev: actually, I wonder how much of this we want to do. I can see some benefit in having it but I can also see moving stuff like autoplay and src causing issues and having us enter into the `lateInit` part of the html5 tech without wanting to.
We should already be setting the source properly on the tech, are we not?

BrandonOCasey: So this fix basically makes things like this work:
```js
var vid = document.querySelector('video-js');

// this will only set the property on the element and not the attribute.
vid.src = 'some-src.mp4';
vid.autoplay = true;

videojs(vid);
```

Currently we only bring over attributes from the `<video-js>` element. This means that the above code will have no `src` and not `autoplay`. This is because we only move over attributes and the `<video-js>` element does not set attributes when properties are set like a real video element does.

gkatsev:
yeah, my main concern is around the src attribute but haven't had a chance to try out an example to see whether what I'm thinking is actually an issue or not

BrandonOCasey:
we could check if any of the attributes exist while doing this, and in that case ignore the property. We would also want to warn if the property exists but is different than the attribute and we are going to use the attribute.

gkatsev:
so, I thought this was going to cause more of an issue but in investigating it, it just as bad as it is currently. We should at some point, not as part of this, have separate support for the src attribute.
The concern was basically that if we copy over the src attribute to the video element, the browser will try to load the video and the source handlers will not be able to intercept it due to the "late init" work in the tech but that's basically equivalent to what video.js currently will do for the src attribute of the video element. So, I guess this is good to go as well.

PR #5040, opened 2018-03-21
feat: update the players source cache on sourceset

BrandonOCasey:
## Description
Update the players source cache on `sourceset`

Orignal PR is https://github.com/videojs/video.js/pull/4717


src/js/player.js
```diff
@@ -1213,6 +1342,16 @@ class Player extends Component {
    * @private
    */
   handleTechSourceset_(event) {
+    this.updateSourceCaches_(event.src);
```
gkatsev: my main concern here is that we accidentally set the source cache to the wrong value.
Specifically, around behavior like https://github.com/videojs/video.js/blob/master/test/unit/sourceset.test.js#L534-L601 where we end up with the first source triggering the second sourceset

Maybe we just need more unit/integration tests specifically around this?

gkatsev:
Also, maybe we should only change the source cache from a sourceset if it wasn't changed via the video.js API?

src/js/player.js
BrandonOCasey: Yeah I have a lot of unit test changes, I just don't want to port them over until the other big pr is in: #5038

src/js/player.js
```diff
@@ -2426,23 +2554,24 @@ class Player extends Component {
...
+      this.updateSourceCaches_(middlewareSource);
       const err = this.src_(middlewareSource);
 
       if (err) {
         if (sources.length > 1) {
           return this.src(sources.slice(1));
         }
 
+        this.updateSourceCaches_();
+        this.changingSrc_ = false;
```
BrandonOCasey: if there was an error and there are no more sources, we should empty the cache and set changingSrc_ to false

src/js/player.js
```diff
@@ -2505,6 +2633,7 @@ class Player extends Component {
         this.techCall_('src', source.src);
       }
 
+      this.changingSrc_ = false;
```
BrandonOCasey: changingSrc should be set false after the source is actual set on the tech using the src method

test/unit/player.test.js
```diff
@@ -164,13 +166,16 @@ QUnit.test('should get current sources from source tag', function(assert) {
   const tag = document.getElementById('example_1');
   const player = TestHelpers.makePlayer({}, tag);
 
+  this.clock.tick(1);
```
BrandonOCasey: I got these to work because the source cache is now updated when the source is selected. We can't actually know what the currentSource will be until then. I am not sure if we want to cache the sources when we get them, then update the cache when a source is actually set.

src/js/player.js
```diff
@@ -1177,6 +1178,109 @@ class Player extends Component {
...
+    // update `currentSource` cache always
+    this.cache_.source = {src, type};
+
+    const matchingSources = this.cache_.sources.filter((s) => s.src && s.src === src);
+    const sourceElSources = [];
+    const sourceEls = this.$$('source');
+    const matchingSourceEls = [];
+
+    for (let i = 0; i < sourceEls.length; i++) {
+      const sourceObj = {src: sourceEls[i].src, type: sourceEls[i].type};
```
misteroneill: What if the `source` has no `type` attribute? It doesn't seem to be required.

BrandonOCasey: It should return undefined or empty string in that case and we may want to omit it in that case. I will test this.

src/js/player.js
gkatsev: I guess this makes sense.

gkatsev: Looks like the video element doesn't update `video.currentSrc` when a new `src` fails to load properly.

BrandonOCasey: yeah `video.currentSrc` remains the previous value but `src` will obviously have changed. I guess we should change the `src` cache but leave the `currentSource` cache on the previous value here?

src/js/player.js
```diff
@@ -1177,6 +1178,112 @@ class Player extends Component {
...
+   * find the mime type of a given source string url if possible
+   *
+   * @param {string} src
+   *        The source string
+   *
+   * @return {string}
+   *         The type that was found
+   *
+   */
+  findMimeType_(src) {
```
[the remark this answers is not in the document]
BrandonOCasey: Its only used in one place so we could just make the function internal to `updateSourceCaches_` if we want.

BrandonOCasey:
I guess I have a general question for the remainder of the comments about when we should update the cache. I think we want the `src` cache to be updated the same as `mediaEl.src` but when do we want the `currentSource(s)` cache to be updated? should it be similar to `mediaEl.currentSrc` or should it be more like `mediaEl.src` with a type?

src/js/player.js
```diff
@@ -1225,6 +1288,30 @@ class Player extends Component {
    * @private
    */
   handleTechSourceset_(event) {
+    // only update the source cache when the source
+    // was not updated using the player api
+    if (!this.changingSrc_) {
```
gkatsev: we could exist early if `this.changingSrc_` instead.

src/js/player.js
```diff
@@ -2464,23 +2551,29 @@ class Player extends Component {
     }
 
     // intial sources
-    this.cache_.sources = sources;
     this.changingSrc_ = true;
 
-    // intial source
-    this.cache_.source = sources[0];
```
gkatsev: this is still not being set unless `this.updateSourceCaches_` sets it?

src/js/player.js
BrandonOCasey: yeah that looks nicer

src/js/player.js
BrandonOCasey: `updateSourceCaches_` will set it.

src/js/player.js
BrandonOCasey: actually that will cause us to have to duplicate the `trigger` that happens below this block

src/js/player.js
BrandonOCasey: https://github.com/videojs/video.js/pull/5040/files#diff-3b0266ff1c037b289ec624ab25b0272eR1224

src/js/player.js
gkatsev: oh, I see, missed that piece.

PR #5050, opened 2018-03-23
test: update karma browser OS versions

forbesjo:
## Description
This PR updates the karma tests to use more modern and popular versions of Windows/Mac OS

## Requirements Checklist

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -151,21 +151,21 @@ function getCustomLaunchers(){
       base: 'BrowserStack',
       browser: 'chrome',
       os: 'Windows',
-      os_version: '8.1'
+      os_version: '10'
     },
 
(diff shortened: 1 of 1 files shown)


test/karma.conf.js
```diff
@@ -151,21 +151,21 @@ function getCustomLaunchers(){
...
       os: 'Windows',
-      os_version: '8.1'
+      os_version: '10'
     },
 
     safari_bs: {
       base: 'BrowserStack',
       browser: 'safari',
       os: 'OS X',
-      os_version: 'El Capitan'
```
gkatsev: we may want to keep this as el capitan so that it tests against safari 9 so we don't break it. We could add high sierra as a separate test run as well, though.

PR #5054, opened 2018-03-26
fix: several source setting inconsistencies and slight refactor

BrandonOCasey:
## Description
* `load` should always fire sourceset
  * because `mediaEl.src` -> `mediaEl.removeAttribute('src')` -> `load()` should fire a sourceset
* initial sourceset should fire the correct source
* handle empty string sources
* sourceset should always fire an absolute source
* player cache should be absolute urls
* reset souceset before dispose and init sourceset after initial `sourceset`

## TODO
* probably one more pr as some browsers support what I am calling "error recovery", which means that they will start watching for a source append after an error if there is currently no source.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2612,9 +2612,11 @@ class Player extends Component {
 
     if (!titleCaseEquals(sourceTech.tech, this.techName_)) {
       this.changingSrc_ = true;
-
       // load this technology with the chosen source
       this.loadTech_(sourceTech.tech, sourceTech.source);
+      this.tech_.ready(() => {
+        this.changingSrc_ = false;
+      });
       return false;
     }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -35,10 +35,6 @@ class Html5 extends Tech {
   constructor(options, ready) {
     super(options, ready);
 
-    if (options.enableSourceset) {
-      this.setupSourcesetHandling_();
-    }
-
     const source = options.source;
     let crossoriginTracks = false;
 
@@ -50,6 +46,11 @@ class Html5 extends Tech {
       this.setSource(source);
     } else {
       this.handleLateInit_(this.el_);
+    }
+
+    // setup sourceset after late sourceset/init
+    if (options.enableSourceset) {
+      this.setupSourcesetHandling_();
     }
 
     if (this.el_.hasChildNodes()) {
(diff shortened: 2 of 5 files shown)


src/js/tech/html5.js
```diff
@@ -46,7 +46,7 @@ class Html5 extends Tech {
     // 1) Check if the source is new (if not, we want to keep the original so playback isn't interrupted)
     // 2) Check to see if the network state of the tag was failed at init, and if so, reset the source
     // anyway so the error gets fired.
-    if (source && (this.el_.currentSrc !== source.src || (options.tag && options.tag.initNetworkState_ === 3))) {
```
gkatsev: I think removing the `initNetworkState_` check breaks `if not, we want to keep the original so playback isn't interrupted`

src/js/player.js
```diff
@@ -2612,8 +2612,6 @@ class Player extends Component {
     }
 
     if (!titleCaseEquals(sourceTech.tech, this.techName_)) {
-      this.changingSrc_ = true;
```
gkatsev: why remove this? Aren't we changing sources here?

src/js/tech/setup-sourceset.js
```diff
@@ -1,6 +1,18 @@
 import window from 'global/window';
 import document from 'global/document';
 import mergeOptions from '../utils/merge-options';
+import {getAbsoluteURL} from '../utils/url';
+
+const getSrc = (el) => {
+  // We use the attribute to check if a source is set because when
+  // the source for the element is set to a blank string. The attribute will
+  // return '' and the property will return window.location.href.
```
gkatsev: however, this is the actual url that the browser tried to load, why hide it?

src/js/utils/filter-source.js
```diff
@@ -55,13 +56,15 @@ const filterSource = function(src) {
...
  */
-function checkMimetype(src) {
+function fixSource(src) {
   const mimetype = getMimeType(src.src);
 
   if (!src.type && mimetype) {
     src.type = mimetype;
   }
 
+  src.src = getAbsoluteURL(src.src);
```
gkatsev: why change the source to be an absolute url?

BrandonOCasey: I don't think we should, I think this was leftover from some testing I was doing with absolute vs relative sources.

gkatsev:
Since this merges into #5040, we should wait until that one is approved before merging this in.

PR #5055, opened 2018-03-27
Use time format function to provide a default consistent with the tim…

guided1:
## Specific Changes proposed
Use the timeFormat function to provide a consistent default rather than hardcode the string.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -78,7 +78,7 @@ class TimeDisplay extends Component {
       this.contentEl_.removeChild(this.contentEl_.firstChild);
     }
 
-    this.textNode_ = document.createTextNode(this.formattedTime_ || '0:00');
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks @guided1!

guided1:
Question... will this get into 6.x or is this only for 7?

gkatsev:
We'll most likely move it over to 6.x as well. Though, the easiest way to get that in would be to open a PR against the `6.x` branch with the same change.
Been busy on https://github.com/videojs/video.js/pull/5096 and other related 7 stuff but I'll probably break away from that early next week to do a 6.x release and what not.

gkatsev:
@guided1 btw, I ported all these stuff from master to 6.x branch and they're released as part of 6.9.

guided1:
Thank you very much!!!

PR #5057, opened 2018-03-28
feat: built-in HLS playback support

forbesjo:
## Description
This PR is to add HLS playback support built into video.js for 7.0 via videojs-http-streaming (shorthand VHS). VHS is the next major version of videojs-contrib-hls that removes HLS Flash playback and includes some experimental DASH support (hence the rename). The purpose is to improve the out-of-the-box experience for video.js and allow cross browser HLS compatibility.

The proposed changes are to have the standard video.js browser and module scripts contain VHS and provide an alternate video.js browser script that does not contain VHS.

Requires https://github.com/videojs/http-streaming/pull/69

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/build/rollup.js
+++ b/build/rollup.js
@@ -55,12 +55,39 @@ const primedBabel = babel({
   plugins: ['external-helpers']
 });
 
+const noVhsEs = {
+  options: {
+    entry: 'src/js/video.js',
+    plugins: [
+      json(),
+      primedBabel,
+      args.progress ? progress() : {},
+      filesize()
+    ],
+    onwarn(warning) {
+      if (warning.code === 'UNUSED_EXTERNAL_IMPORT' ||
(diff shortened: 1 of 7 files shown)


forbesjo:
The node require test will fail due to https://github.com/gkatsev/rollup-plugin-bundle-worker/blob/master/src/workerhelper.js#L3

build/rollup.js
```diff
@@ -76,6 +78,36 @@ const es = {
   banner: compiledLicense(Object.assign({includesVtt: true}, bannerData)),
   useStrict: false,
   format: 'es',
```
forbesjo: Are we sure that we want to use the `es` format and not `cjs` for better compatibility?

forbesjo:
Opened https://github.com/gkatsev/rollup-plugin-bundle-worker/pull/1 for fixing the node require test

PR #5062, opened 2018-03-30
queue playback events when playback rate is zero and seeking is true (vjs5)

squarebracket:
This is a clone of PR #5024 made against the 5.x branch.

Needed because the [related PR for contrib-hls](https://github.com/videojs/videojs-contrib-hls/pull/1374) will also link against vjs5.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -223,6 +223,16 @@ const TECH_EVENTS_RETRIGGER = [
   'texttrackchange'
 ];
 
+// events to queue when playback rate is zero
+// this is a hash for the sole purpose of mapping non-camel-cased event names
+// to camel-cased function names
+const TECH_EVENTS_QUEUE = {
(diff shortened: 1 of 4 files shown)


gkatsev:
Maybe we drop vjs 5 support with this PR for contrib-hls? thoughts @forbesjo @OshinKaramian @mjneil?
I'd prefer not having to do a minor release of the 5.x line.

forbesjo:
I don't think we should update video.js 5 for this however as noted in the other PR (https://github.com/videojs/video.js/pull/5024#issuecomment-378820010) we should make sure that the change in videojs-contrib-hls is backwards compatible

forbesjo:
Upon further discussion we'll make it very clear in the contrib-hls README and package.json which version of video.js this change is compatible with.

gkatsev:
Going to close this until/if we decide that we want to make this change.

PR #5072, opened 2018-04-03
fix: fire sourceset on initial source append (#5038)

BrandonOCasey:
6.x version of #5038


BrandonOCasey:
fixed IE9 by skipping `innerHTML` tests. IE9 cannot add a `<source>` through `innerHTML`

PR #5074, opened 2018-04-04
Add option to override native audio and video to html5 tech

OshinKaramian:
## Description
Adding an option to override native audio/video tracks.  This is being added to ease enabling and disabling it within https://github.com/videojs/http-streaming.  This is partly to make enabling/disabling native playback easier when integrating videojs/http-streaming into this project.

I'm in the process of adding tests.  I'd like eyes on this while I'm doing that to ensure that the implementation makes sense/if I'm missing anything that might break playback that I'm unaware of.

## Specific Changes proposed
Adds 2 flags on the Html5 tech, `overrideNativeAudioTracks` and `overrideNativeVideoTracks` that can be set at any time.  The change will take effect any time the source is set.

One question I'm unclear on... what should the version compatibility be with video.js with this change?  I can modify things to be backwards compatible (checking to see if the APIs exist and handling things one way or another).  Do we care about this?  Or are we okay removing this (since leaving it in for backwards compatibility purposes would add a little bit of tech debt).

## Requirements Checklist
  - [x] Change has been verified in an actual browser (Verified in Edge with VHS)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -204,7 +204,7 @@ class Html5 extends Tech {
    * Attempt to force override of native audiio/video tracks.
    *
    * @param {Boolean} override - If set to true native audio/video will be overridden,
-   * otherwise native video will potentially be used.
+   * otherwise native audio/video will potentially be used.
    */
   overrideNativeTracks(override) {
     // If there is no behavioral change don't add/remove listeners
@@ -257,6 +257,7 @@ class Html5 extends Tech {
           !elTracks.addEventListener) {
         return;
       }
+
       const listeners = {
         change(e) {
           techTracks.trigger({
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -200,10 +200,22 @@ class Html5 extends Tech {
     });
   }
 
+  /**
+   * Attempt to force override of native audio tracks.
+   *
+   * @param {Boolean} override - If set to true native audio will be overridden,
+   * otherwise native video will potentially be used.
+   */
   overrideNativeAudioTracks(override) {
     this.forceNativeAudioOverride = override;
   }
 
+  /**
+   * Attempt to force override of native video tracks.
+   *
+   * @param {Boolean} override - If set to true native video will be overridden,
+   * otherwise native video will potentially be used.
+   */
   overrideNativeVideoTracks(override) {
     this.forceNativeVideoOverride = override;
   }
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -208,7 +208,7 @@ class Html5 extends Tech {
    */
   overrideNativeTracks(override) {
     // If there is no behavioral change don't add/remove listeners
-    if (override !== (this.featuresNativeAudioTracks && this.featuresNativeAudioTracks)) {
+    if (override !== (this.featuresNativeAudioTracks && this.featuresNativeVideoTracks)) {
       return;
     }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -200,6 +200,14 @@ class Html5 extends Tech {
     });
   }
 
+  overrideNativeAudioTracks(override) {
+    this.forceNativeAudioOverride = override;
+  }
+
+  overrideNativeVideoTracks(override) {
+    this.forceNativeVideoOverride = override;
+  }
+
   /**
    * Proxy all native track list events to our track lists if the browser we are playing
    * in supports that type of track list.
@@ -217,22 +225,36 @@ class Html5 extends Tech {
           !elTracks.addEventListener) {
         return;
       }
-      const listeners = {
-        change(e) {
+
+      const listeners = {};
+
+      listeners.change = (e) => {
+        if (!this[`forceNative${props.capitalName}Override_`]) {
           techTracks.trigger({
             type: 'change',
             target: techTracks,
             currentTarget: techTracks,
             srcElement: techTracks
           });
-        },
-        addtrack(e) {
+        }
+      };
+
+      listeners.addtrack = (e) => {
+        if (this[`forceNative${props.capitalName}Override_`] && elTracks.addTrack) {
+          elTracks.addTrack(e.track);
+        } else if (!this[`forceNative${props.capitalName}Override_`]) {
           techTracks.addTrack(e.track);
-        },
-        removetrack(e) {
+        }
+      };
+
+      listeners.removetrack = (e) => {
+        if (this[`forceNative${props.capitalName}Override_`] && elTracks.addTrack) {
+          elTracks.removeTrack(e.track);
+        } else if (!this[`forceNative${props.capitalName}Override_`]) {
           techTracks.removeTrack(e.track);
         }
       };
+
       const removeOldTracks = function() {
         const removeTracks = [];
 
@@ -605,12 +627,22 @@ class Html5 extends Tech {
    * @deprecated Since version 5.
    */
   src(src) {
+    this.forceNativeAudioOverride_ = this.forceNativeAudioOverride;
+    this.forceNativeVideoOverride_ = this.forceNativeVideoOverride;
+
     if (src === undefined) {
       return this.el_.src;
     }
 
     // Setting src through `src` instead of `setSrc` will be deprecated
     this.setSrc(src);
+  }
+
+  setSrc(src) {
+    this.forceNativeAudioOverride_ = this.forceNativeAudioOverride;
+    this.forceNativeVideoOverride_ = this.forceNativeVideoOverride;
+
+    super.setSrc(src);
   }
 
   /**
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -208,7 +208,7 @@ class Html5 extends Tech {
    */
   overrideNativeTracks(override) {
     // If there is no behavioral change don't add/remove listeners
-    if (override === !(this.featuresNativeAudioTracks && this.featuresNativeAudioTracks)) {
+    if (override !== (this.featuresNativeAudioTracks && this.featuresNativeAudioTracks)) {
       return;
     }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -267,16 +267,13 @@ class Html5 extends Tech {
             srcElement: techTracks
           });
         },
-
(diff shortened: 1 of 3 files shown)


src/js/tech/html5.js
```diff
@@ -200,6 +200,14 @@ class Html5 extends Tech {
     });
   }
 
+  overrideNativeAudioTracks(override) {
```
forbesjo: Needs a jsDoc explaining the intent of these functions and how it will take affect on source change

forbesjo:
We should make sure these options are compatible with the player options https://github.com/videojs/video.js/blob/master/docs/guides/options.md#html5

src/js/tech/html5.js
```diff
@@ -605,6 +639,9 @@ class Html5 extends Tech {
    * @deprecated Since version 5.
    */
   src(src) {
+    this.forceNativeAudioOverride_ = this.forceNativeAudioOverride;
```
gkatsev: why the two separate flags?

src/js/tech/html5.js
```diff
@@ -217,22 +237,36 @@ class Html5 extends Tech {
...
             currentTarget: techTracks,
             srcElement: techTracks
           });
-        },
-        addtrack(e) {
+        }
+      };
+
+      listeners.addtrack = (e) => {
+        if (this[`forceNative${props.capitalName}Override_`] && elTracks.addTrack) {
```
gkatsev: why would `elTracks` not have an `addTrack` method?

gkatsev:
So, the idea here is that VHS would call `overrideNativeVideoTracks` and `overrideNativeVideoTracks` with `true` to tell the the tech not to use native stuff?
If I'm remembering proxyNativeTracks_ correctly, would the best solution be to just remove the listeners when we disable native track support? Though, that may require a larger refactor.

Also, the override methods should probably be added to `tech.js` as empty methods.

src/js/tech/html5.js
```diff
@@ -217,22 +237,36 @@ class Html5 extends Tech {
...
+        } else if (!this[`forceNative${props.capitalName}Override_`]) {
           techTracks.addTrack(e.track);
-        },
-        removetrack(e) {
+        }
+      };
+
+      listeners.removetrack = (e) => {
+        if (this[`forceNative${props.capitalName}Override_`] && elTracks.addTrack) {
+          elTracks.removeTrack(e.track);
```
gkatsev: so, I think we don't want to do this part of the if statement for `addTrack` and `removeTrack`, since the event comes from the native elTracks object already.
I think best case scenario is that we should be removing the listeners when we disable native support.

src/js/tech/html5.js
```diff
@@ -200,6 +200,26 @@ class Html5 extends Tech {
...
   }
 
+  /**
+   * Attempt to force override of native audio tracks.
+   *
+   * @param {Boolean} override - If set to true native audio will be overridden,
+   * otherwise native video will potentially be used.
+   */
+  overrideNativeAudioTracks(override) {
+    this.forceNativeAudioOverride = override;
```
gkatsev: should we just override `featuresNativeAudioTracks` and `featuresNativeVideoTracks`? While a lot of the docs language around it says it's around "support" it also means whether the tech will be using that feature or not.

gkatsev: we may also need to do something like call `proxyNativeTracks_` after the fact if this is called with `false` so that that process kicks in, especially if we originally start with them disabled.

src/js/tech/html5.js
OshinKaramian: It seemed more flexible to have two flags over one flag to rule them all.  I can merge them together if that's preferred.

src/js/tech/html5.js
OshinKaramian: I'm not sure, it seemed safer to check to see if a method existed on something before using it.

OshinKaramian:
Unfortunately I can't comment on single comments, so re: removing listeners:

I explored this initially but had also determined it would be a pretty large refactor, I was attempting to go in with less work/more of a scalpel.  I can go down this route, but I have a feeling this is going to be a bit more work to ensure things do not break.

src/js/tech/html5.js
```diff
@@ -260,6 +298,16 @@ class Html5 extends Tech {
...
+
+        if (!this.trackListeners) {
+          this.trackListeners = [];
+        }
+
+        if (!this.trackListeners[props.capitalName]) {
+          this.trackListeners[props.capitalName] = [];
+        }
+
+        this.trackListeners[props.capitalName].push({ eventName, listener });
```
gkatsev: what about instead of making an array of stuff, do something like the following outside of this Object.keys().forEach:
```js
this[props.getterName + 'Listeners_'] = listeners;
```
then to remove these, outside of this method we can do something like:
```js
Object.keys(this.audioTracksListeners_).forEach((eventName) => {
  elTracks.removeEventListener(eventName, this.audioTracksListeners_[eventName]
});
Object.keys(this.videoTracksListeners_).forEach((eventName) => {
  elTracks.removeEventListener(eventName, this.videoTracksListeners_[eventName]
});
```

src/js/tech/html5.js
```diff
@@ -200,6 +200,40 @@ class Html5 extends Tech {
     });
   }
 
+  /**
+   * Attempt to force override of native video tracks.
+   *
+   * @param {Boolean} override - If set to true native video will be overridden,
+   * otherwise native video will potentially be used.
+   */
+  overrideNativeTracks(override) {
```
gkatsev: might be good to make this a no-op if we aren't changing the override.

src/js/tech/html5.js
```diff
@@ -200,6 +200,40 @@ class Html5 extends Tech {
...
+      this.trackListeners = [];
+    }
+
+    removeTracks('video');
+    removeTracks('audio');
+
+    this.trackListeners.Video = [];
+    this.trackListeners.Audio = [];
+
+    this.proxyNativeTracks_();
```
gkatsev: so basically, everything above will remove listeners and then we call proxyNativeTracks_ so that if we turn native back on it'll run but if we turned it off, proxyNativeTracks_ will exit early and not run?

src/js/tech/html5.js
```diff
@@ -200,6 +200,40 @@ class Html5 extends Tech {
     });
   }
 
+  /**
+   * Attempt to force override of native video tracks.
+   *
+   * @param {Boolean} override - If set to true native video will be overridden,
+   * otherwise native video will potentially be used.
```
gkatsev: would be good to update this to mention that this is just for audio and video tracks

src/js/tech/html5.js
gkatsev: we should add either an empty method or a default implementation in tech.js for this.

src/js/tech/html5.js
OshinKaramian: That sounds right to me.

src/js/tech/html5.js
```diff
@@ -260,6 +301,9 @@ class Html5 extends Tech {
         const listener = listeners[eventName];
 
         elTracks.addEventListener(eventName, listener);
+
+        this[props.getterName + 'Listeners_'] = listeners;
```
gkatsev: this can move outside of the forEach (line 300).

src/js/tech/html5.js
```diff
@@ -200,6 +200,46 @@ class Html5 extends Tech {
...
 
+  /**
+   * Attempt to force override of native audiio/video tracks.
+   *
+   * @param {Boolean} override - If set to true native audio/video will be overridden,
+   * otherwise native audio/video will potentially be used.
+   */
+  overrideNativeTracks(override) {
+    // If there is no behavioral change don't add/remove listeners
+    if (override === this.override_) {
```
gkatsev: could we compare this against the values of `featuresNativeVideoTracks` and `featuresNativeAudioTracks` instead? that way we don't have another flag to worry about.

OshinKaramian: Given a brief chat with @gkatsev I'll set this to check against `featuresNativeVideoTracks && featuresNativeAudioTracks`.

src/js/tech/html5.js
```diff
@@ -200,6 +200,43 @@ class Html5 extends Tech {
...
 
+  /**
+   * Attempt to force override of native audiio/video tracks.
+   *
+   * @param {Boolean} override - If set to true native audio/video will be overridden,
+   * otherwise native audio/video will potentially be used.
+   */
+  overrideNativeTracks(override) {
+    // If there is no behavioral change don't add/remove listeners
+    if (override !== (this.featuresNativeAudioTracks && this.featuresNativeAudioTracks)) {
```
forbesjo: copy/paste

OshinKaramian: That took me a minute to spot, good 👀 s

src/js/tech/html5.js
```diff
@@ -200,6 +200,43 @@ class Html5 extends Tech {
...
+        const elTracks = this.el().videoTracks;
+
+        elTracks.removeEventListener(eventName, this.videoTracksListeners_[eventName]);
+      });
+    }
+
+    this.featuresNativeVideoTracks = !override;
+    this.featuresNativeAudioTracks = !override;
+
+    this.audioTracksListeners_ = [];
```
gkatsev: should be set to `null` probably.

src/js/tech/html5.js
```diff
@@ -200,6 +200,43 @@ class Html5 extends Tech {
     });
   }
 
+  /**
+   * Attempt to force override of native audiio/video tracks.
```
BrandonOCasey: audiio -> audio

src/js/tech/html5.js
```diff
@@ -200,6 +200,43 @@ class Html5 extends Tech {
...
+   *
+   * @param {Boolean} override - If set to true native audio/video will be overridden,
+   * otherwise native audio/video will potentially be used.
+   */
+  overrideNativeTracks(override) {
+    // If there is no behavioral change don't add/remove listeners
+    if (override !== (this.featuresNativeAudioTracks && this.featuresNativeVideoTracks)) {
```
BrandonOCasey: should audio and video tracks be overridden in a separate check?

gkatsev: at this time, we're assuming they'll always match, we can always change it later if necessary.

PR #5080, opened 2018-04-09
Use valid json in translation example

thijstriemstra:
It's currently printed red in the github markdown pages, aka syntax error:

https://github.com/videojs/video.js/blob/master/docs/guides/languages.md#json-format


thijstriemstra:
>  I think originally it was to show that this isn't a complete Spanish translation 

I understand but it just looked silly, especially because it was being printed in red.

PR #5091, opened 2018-04-12
fix: avoid black borders if poster doesn't fit 100% to ratio

axten:
## Description
sometimes there are rounding issues or the poster is not sized 100% correctly leading to black borders at top/bottom or left/right.

## Specific Changes proposed
make the poster cover the available space

## Requirements Checklist
- [x] Bug fixed
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_poster.scss
+++ b/src/css/components/_poster.scss
@@ -30,3 +30,8 @@
 .vjs-using-native-controls .vjs-poster {
   display: none;
 }
+
+// avoid black borders if poster doesn't fit 100% to ratio
+.vjs-fluid .vjs-poster {
+  background-size: cover;
+}


gkatsev:
my main concern here is that cover allows the image to be cropped. This is probably fine if the poster is the same aspect ratio as the video so it only gets cropped by like a pixel to count for rounding with fluid mode. If the aspect ratio in the same, though, large portions of the poster could be cropped.

axten:
your right, thats a kind of breaking change.
but I think that ok for fluid mode. Imho nobody should apply a 4:3 poster to a 16:9 video...

gkatsev:
they shouldn't, but I've definitely seen it before.

@videojs/core-committers thoughts on this change?

axten:
basically both cases can be fixed with additional css from the users so imho you should go with the solution that fits for the greater group of users

gkatsev:
Yeah, I think this is too risky to move forward with. However, maybe it's something we could add in docs somewhere.
Thanks for the PR, though, @axten.

PR #5096, opened 2018-04-13
WIP: chore: update rollup and uglify and the build process

gkatsev:
This is a WIP PR that upgrades rollup and uglify to latest versions. It also rewrites the rollup config to take advantage of the newest features of rollup. Minification is also split up into a separate process, mostly to speed up the build. The total build type can go down from about 2.5 minutes to around 1 minute for all the generated javascript files.

TODO:
- [ ] hook up `npm run build`
- [ ] hook up `npm start`
- [ ] remove old `build/rollup.js` file
- [ ] verify tests
- [ ] verify builds


gkatsev:
When using `rollup --watch`, [rollup sets up an alternative screen and hides all other output](https://github.com/rollup/rollup/issues/1804). I have made a PR to fix it https://github.com/rollup/rollup/pull/2125, hopefully, with the current `clearScreen` `watch` option, when my PR is merged, just updating to it will fix it.

BrandonOCasey:
So I guess its faster to minify outside of rollup here?

gkatsev:
yup, because it doesn't need to do a lot of work yet another time plus we're sharing the name cache between builds. But the total build time from old stuff to new stuff is maybe half.

gkatsev:
This has benefit of using a default config for rollup and makes the config a lot more maintainable without weird custom code.

PR #5097, opened 2018-04-13
Moving seekbar event handler bindings into a function

guided1:
## Description
Move all eventHandler bindings into a function.

## Specific Changes proposed
This allows for easier customisation of the the seekbar component.
If you extend the seekbar component you can choose to override `setEventHandlers_` or keep it default.

If you do override `setEventHandlers_`  and call `super()` in the constructor, you get your own event handlers + the functionality provided by `Slider`

Currently extending slider and wanting different bindings, means either copy pasting from the  existing slider or writing a function to undo the bindings (which is not possible in this case since they are anonymous functions).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -38,7 +38,21 @@ class SeekBar extends Slider {
    */
   constructor(player, options) {
     super(player, options);
+    this.setEventHandlers_(player, options);
+  }
 
+  /**
+   * Sets the event handlers
+   *
+   * @param {Player} player
+   *        The `Player` that this class should be attached to.
+   *
+   * @param {Object} [options]
+   *        The key/value store of player options.
+   *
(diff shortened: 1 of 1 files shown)


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -38,7 +38,21 @@ class SeekBar extends Slider {
...
+   *
+   * @param {Player} player
+   *        The `Player` that this class should be attached to.
+   *
+   * @param {Object} [options]
+   *        The key/value store of player options.
+   *
+   * @private
+   */
+  setEventHandlers_(player, options) {
```
gkatsev: I would consider removing passing in the player and options and instead update the code to use `this.player_` and `this.options_`

guided1: sure thing.

PR #5123, opened 2018-04-24
feat: add 'autoSetup' option

### LINKED ISSUE #5094: Feature: would be great to have an option to disable autoSetup
## Description
since we init players manually there is no need to run the autosetup loop.

maybe its possible to implement with global setting `videojs.options.autoSetup = false;`
in the same tick as the video.js script gets excuted.

axten:
fixes #5094

## Description
this must be set globally with `videojs.options.autoSetup = false` in the same tick as videojs source is loaded to take effect

## Requirements Checklist
  - [x] Docs/guides updated


axten:
@gkatsev is something missing?

gkatsev:
nope, last week as hackweek at work, so, I didn't really look at Video.js issues and PRs.
I'll take a look this week.

src/js/setup.js
```diff
@@ -18,7 +18,7 @@ let videojs;
 const autoSetup = function() {
 
   // Protect against breakage in non-browser environments.
-  if (!Dom.isReal()) {
+  if (!Dom.isReal() || !videojs.options.autoSetup) {
```
gkatsev: while this would work, it still runs the initial timer and makes the comment above incorrect.

axten: ok I will change this

axten:
so changes are done... but now it's impossible to set this option, isn't it?

axten:
Yes, confirm. I will revert changes...

gkatsev:
@axten do you want this to go into Video.js 6 or will just having it in v7 be ok?

gkatsev:
went ahead and ported it over.

PR #5148, opened 2018-05-07
feat: add double-click handler to toggle fullscreen

bcdarius:
## Description
This PR adds double-click handler to toggle player fullscreen.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [x] Example created ([starter template on JSBin](http://output.jsbin.com/masuyin))


src/js/player.js
```diff
@@ -1519,6 +1519,42 @@ class Player extends Component {
...
+
+    // We want to handle single and double click events separately
+    // because they perform different actions.
+    // A single-click will pause/resume playback.
+    // A double-click will enter/exit fullscreen.
+
+    // If a subsequent click doesn't occur after a single click,
+    // within a short amount of time, trigger a single-click handler.
+    if (event.detail === 1) {
+      this.singleClickTimer_ = setTimeout(function() {
```
gkatsev: this shouldn't be necessary because there's a native `dblclick` event we can use instead.

src/js/player.js
```diff
@@ -1536,6 +1572,33 @@ class Player extends Component {
...
+   */
+  handleTechDoubleClick_(event) {
+    if (!this.controls_) {
+      return;
+    }
+
+    const inAllowedEls = Array.prototype.some.call(
+        this.$$('.vjs-control-bar, .vjs-modal-dialog'),
+        function(el) {
+          el.contains(event.target);
```
gkatsev: you want to return from this method or change the `function` to an arrow function.

src/js/player.js
bcdarius: I initially looked into utilising `dblclick`. However, the problem is that there's already `mousedown` listener on the tech which is triggered before `dblclick`. So using it would result in the player being paused/resumed when entering/exiting fullscreen.

gkatsev: it should be fine because that should receive two events so it should pause and then unpause. I think it would still be better than trying to implement our own double click handler.

bcdarius: Yep, that's what's happening — first click pauses, second click resumes playback — when entering/exiting fullscreen. While this is not a huge issue, it makes the experience somehow less seamless when toggling fullscreen. Firstly, the playback is visibly paused/resumed. The play toggle button also can be seen changing states. Furthermore,  in such cases unnecessary `play`, `pause` events are generated.

gkatsev: So, looking at youtube and vimeo, they don't even up pausing and playing the video during double clicks and as far as I can tell they both use dblclick handlers. I think play toggle handling is just incorrect and needs to be re-worked because trying to implement it ourselves will just lead to various weird corner cases and issues. For example, I can get it to switch playback state and fullscreen state if I take a short while between the two clicks.

bcdarius: The event order in `dblclick` is as follows: `mousedown -> mouseup -> click -> mousedown -> mouseup -> click -> dblclick`. If we look at re-working handling playback toggle and using native `dblclick`, we will still need some approach of cancelling the first click when `dblclick` occurs as having `click` — or in this particular case `mousedown` — and `dblclick` handler on the media element would trigger both handlers.

gkatsev:
500ms is too long of a wait. 250ms or 300ms might be reasonable.

Personally, I still think just leaving the regular behavior as is isn't really an issue.

Also, I was looking into this a bit and it seems like there's a [`detail` property](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail) on event objects that tells us how many clicks are happening which could simplify usage via just click events rather than using both click and dblclick, didn't know about it previously.

bcdarius:
Going beyond 500ms seem to introduce an issue where short double-clicks in some instances switch playback and fullscreen state at the same time. The same occurs when using regular `mousedown` (or `click`) and `dblclick` handlers.  I agree, though, having such a delay introduces a quite visible latency for pausing/resuming playback on single clicks. I'm looking into a more reasonable solution.

bcdarius:
I looked at `event.detail`  originally — that's what I initially used to track clicks (see the very first commit on this PR).   Out of the box, however, it doesn't help that much in this scenario because of how the event order occurs —  the event sequence for `dblclick` is `mousedown -> mouseup -> click -> mousedown -> mouseup -> click -> dblclick` — so you still need to have some mechanism to cancel out a single-click handler when double-click occurs.

src/js/player.js
```diff
@@ -1643,6 +1645,34 @@ class Player extends Component {
...
+    if (!this.controls_) {
+      return;
+    }
+
+    const inAllowedEls = Array.prototype.some.call(
+        this.$$('.vjs-control-bar, .vjs-modal-dialog'),
+        el => el.contains(event.target)
+      );
+
+    if (!inAllowedEls) {
```
misteroneill: This seems confusing. If it's not in allowed elements it's prohibited?

gkatsev: maybe we just need a comment? Basically, we shouldn't try toggling fullscreen if we double click inside of the control bar or the modal dialog.

gkatsev:
Can the package-lock changes be reverted?

gkatsev:
Also, this is much simpler and if we get issues around it, we can always fix it forward then!

src/js/player.js
bcdarius: Added a comment.

campones:
yes it doesn't work with flash. I ve implemented your solution since we spoke about it back in february and the only channel where double click doesn't work is the one that use the flash tech instead of html5

PR #5154, opened 2018-05-08
Remove unnecessary ARIA role on the Control Bar. Fixes #5134

### LINKED ISSUE #5134: control bar has a role of group but no label, could cause confusion
## Description
The control bar has a role of group but no label. This could cause confusion when a screen reader user enters the control bar since nothing would be announced beyond it being a group.
However, a potential issue with adding a label to the control bar is that if individual buttons would start announcing that label as part of their announcements, it would be a worse user experience.

OwenEdwards:
## Description
Remove unnecessary ARIA role on the Control Bar. Fixes #5134

## Requirements Checklist
  - [x] Change has been verified in an actual browser (Firefox+NVDA, IE+JAWS)

### DIFF
--- a/src/js/control-bar/control-bar.js
+++ b/src/js/control-bar/control-bar.js
@@ -40,9 +40,6 @@ class ControlBar extends Component {
       className: 'vjs-control-bar',
       dir: 'ltr'
     }, {
-      // The control bar is a group, but we don't aria-label it to avoid
-      //  over-announcing by JAWS
-      role: 'group'
     });
   }
 }
--- a/src/js/control-bar/control-bar.js
+++ b/src/js/control-bar/control-bar.js
@@ -39,7 +39,6 @@ class ControlBar extends Component {
     return super.createEl('div', {
       className: 'vjs-control-bar',
       dir: 'ltr'
-    }, {
     });
   }
 }


src/js/control-bar/control-bar.js
```diff
@@ -40,9 +40,6 @@ class ControlBar extends Component {
       className: 'vjs-control-bar',
       dir: 'ltr'
     }, {
```
gkatsev: we can just get rid of this object entirely, if we aren't using it.

gkatsev:
So you think that the control bar should just not have a role? I guess it makes sense since you never interact with the control bar itself.

OwenEdwards:
@gkatsev I did some testing, and there just doesn't seem to be any point in indicating the Control Bar as a separate object - it's all just part of the video player. Perhaps in the future, if there's content which isn't part of the control bar that needs to be clearly distinguishable from the control bar, then it would make sense, but right now you don't interact with the control bar and it doesn't convey any concept of grouping. For example, is it semantically important that the BPB isn't part of the control bar? I don't think it is.

To answer your other question, yes - this should also be ported to master; I wasn't sure if it was better to start there and back-post, or vice versa. Let me know if you want me to switch it.

gkatsev:
Makes sense.

Either way is fine. We can easily port for master when merged.

PR #5157, opened 2018-05-09
Fix time-display.js to restore hidden label text for screen readers. Fixes #5135

### LINKED ISSUE #5135: time displays only have the time and no description or context for screen readers
```
<div class="vjs-current-time vjs-time-control vjs-control">
  <div class="vjs-current-time-display" aria-live="off">0:01</div>
</div>
<div class="vjs-time-control vjs-time-divider">
  <div><span>/</span></div>
</div>
<div class="vjs-duration vjs-time-control vjs-control">
  <div class="vjs-duration-display" aria-live="off">0:46</div>
</div>
```

OwenEdwards:
## Description
Fix time-display.js to restore hidden label text for screen readers. Fixes #5135

## Specific Changes proposed
Also, fix a typo in translations-needed.md, and add a space in the hidden label for live-display, so that it doesn't run together with the visible "LIVE" indication.

## Requirements Checklist


lang/en.json
```diff
@@ -5,7 +5,7 @@
   "Pause": "Pause",
   "Replay": "Replay",
   "Current Time": "Current Time",
-  "Duration Time": "Duration Time",
```
gkatsev: So, looking at the translations, most of them just translated 'Duration', so, I wonder whether we should just leave this as is? Or add Duration as a second option and then where we use it in the code we first try Duration localization and then Duration Time localization.

src/js/control-bar/live-display.js
```diff
@@ -42,7 +42,7 @@ class LiveDisplay extends Component {
 
     this.contentEl_ = Dom.createEl('div', {
       className: 'vjs-live-display',
-      innerHTML: `<span class="vjs-control-text">${this.localize('Stream Type')}</span>${this.localize('LIVE')}`
+      innerHTML: `<span class="vjs-control-text">${this.localize('Stream Type')}&nbsp;</span>${this.localize('LIVE')}`
```
gkatsev: we should use the unicode representation (`\u00a0`) for xhtml safety https://github.com/videojs/video.js/pull/4825

src/js/control-bar/time-controls/current-time-display.js
```diff
@@ -70,12 +70,12 @@ class CurrentTimeDisplay extends TimeDisplay {
 }
 
 /**
- * The text that should display over the `CurrentTimeDisplay`s controls. Added to for localization.
+ * The text that is added to the `CurrentTimeDisplay` for screen reader users.
  *
  * @type {string}
  * @private
  */
-CurrentTimeDisplay.prototype.controlText_ = 'Current Time';
```
gkatsev: I'm thinking we should leave the controlText_ property in case anyone is relying on it but internally ignore it and only use labelText_

src/js/control-bar/time-controls/remaining-time-display.js
```diff
@@ -47,6 +47,7 @@ class RemainingTimeDisplay extends TimeDisplay {
    * @private
    */
   formatTime_(time) {
+    // TODO: The "-" should be decorative, and not announced by a screen reader
```
gkatsev: this is because the screen reader would announce something like "remaining time minus 5:00" as it is currently?

src/js/control-bar/time-controls/time-display.js
```diff
@@ -38,18 +38,16 @@ class TimeDisplay extends Component {
   createEl(plainName) {
     const className = this.buildCSSClass();
     const el = super.createEl('div', {
-      className: `${className} vjs-time-control vjs-control`
+      className: `${className} vjs-time-control vjs-control`,
+      innerHTML: `<span class="vjs-control-text">${this.localize(this.labelText_)}&nbsp;</span>`
     });
 
-    this.contentEl_ = Dom.createEl('div', {
+    this.contentEl_ = Dom.createEl('span', {
```
gkatsev: this is probably fine to make, but I'd like to take a look at the output first.

OwenEdwards: If we use a `<div>`, JAWS treats the content label and the time as two separate things, so you need to press the arrow key to move between them in Virtual Cursor mode. With a `<span>`, JAWS treats them as a single thing (hence the need to be careful about adding a space in certain places with spans).

src/js/control-bar/time-controls/remaining-time-display.js
OwenEdwards: Exactly. The "-" is a visual convention, but the actual amount of time remaining is not negative. It's a small detail, which is why I didn't fix it now, but it ought to be fixed in the future.

lang/en.json
OwenEdwards: "Duration Time" sounds like something poorly translated into English ;-)

gkatsev: yup, makes sense, just thinking about how to reduce need for re-working all the translations.

src/js/control-bar/time-controls/remaining-time-display.js
gkatsev: we should probably open an issue so we don't forget

src/js/control-bar/time-controls/time-display.js
gkatsev: yeah, I don't see any issues with this change once trying it out.

src/js/control-bar/time-controls/current-time-display.js
OwenEdwards: It's confusing, because these components don't extend `Button` or `ClickableComponent` which does have that concept of `controlText_`, and because these *aren't* controls, they're information displays.

It looks like `controlText_` was added when the common `time-display.js` was created. Which version was that? If it wasn't 6.0, then it doesn't seem like anyone ought to be "relying on it", since it's private.

lang/en.json
OwenEdwards: Could we just change it to "Duration" in this and all the other language files, and then note some of the translations may be incorrect? It seems like translations get fixed as much as they get added to?

gkatsev: yeah, maybe that's just the best choice.

src/js/control-bar/time-controls/remaining-time-display.js
OwenEdwards: Done: see #5168.

lang/en.json
OwenEdwards: Done.

OwenEdwards:
@gkatsev is this what you were thinking as far as leaving the `controlText_` attribute but not using it?

gkatsev:
yup, thanks @OwenEdwards!

Issue #5023, opened 2018-03-15
NVDA Windows Chrome & Firefox - Play button is repeated multiple times when clicked.


waughcbw:
## Description
Once NVDA focus is on the play button and the button is clicked, NVDA repeats the state of the button multiple times (Chrome 6 times, Firefox 3 times).

## Steps to reproduce
With NVDA running, go to videojs.com and click the play button, either through enter key or mouse click.

### Expected
The state of the play button is stated only once when clicked.

### Actual
The state of the play button is stated 6 times in Windows Chrome and 3 times in Windows Firefox when clicked. e.g. When the play button is clicked in Windows Chrome, NVDA will read "Play, Play, Play, Play, Play, Play".

### Error output
N/A

#### videojs
Latest version on video.js website.
#### browsers
Chrome, Firefox
#### OSes
Windows 7 & 10

jiangtiandao:
The interesting thing is what is announced is "Play Play lay Play Play Play"

jiangtiandao:
http://jsbin.com/piqulisose/2/edit?html,js,console

jiangtiandao:
I found a workaround by removing original play control button and implement one instead.
```HTML
<button class="vjs-play-control vjs-control vjs-button" type="button" aria-live="polite" title="Play" aria-disabled="false"><span aria-hidden="true" class="vjs-icon-placeholder"></span><span class="vjs-control-text">Play</span></button>
```
If we remove inner span will reduce to 4 times in Chrome.
If we remove the title attribute will reduce to 2 times  in Chrome.

gkatsev:
Hey, thanks for the issue. We knew this has been happening but haven't had a chance to take a look at it. The title attribute was originally added so that you can visually read what a button that only had an icon but this is clearly causing issues. The plan is to go with removing the title attribute and add our own tooltip.

alex-barstow:
This is reproducible without Video.js as well: http://solutions.brightcove.com/abarstow/BC-39639/nvda-test.html

gkatsev:
This should be fixed as part of https://github.com/videojs/video.js/pull/5158

PR #5158, opened 2018-05-09
Reduce the multiple-announcement by screen readers of the new name of a button when its text label changes. Partial fix to #5023

OwenEdwards:
## Description
Reduce the multiple-announcement by screen readers of the new name of a button when its text label changes. Partial fix to #5023

## Specific Changes proposed
Move the `aria-live` attribute to the control text element within buttons, rather than on the whole button, so it is not affected by the change of the `title` attribute, only by the change of the control text.

## Requirements Checklist

### DIFF
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -41,10 +41,7 @@ class Button extends ClickableComponent {
     attributes = assign({
 
       // Necessary since the default button type is "submit"
-      'type': 'button',
-
-      // let the screen reader user know that the text of the button may change
-      'aria-live': 'polite'
+      type: 'button'
     }, attributes);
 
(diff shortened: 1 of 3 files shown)


gkatsev:
Interesting that this test is failing: https://travis-ci.org/videojs/video.js/builds/376684471#L3660-L3669

OwenEdwards:
Fixed the failing test.

It seems like having `aria-live` on the button itself means that JAWS and NVDA announce the button both when the button text changes and when the `title` attribute changes. NVDA speaks the new label more times because it announces the button text as the label and the title as the description, so it says, for example, "_pause button pause_". JAWS doesn't appear to do this, so it doesn't repeat it as many times.

Issue #3663, opened 2016-10-03
Should Chapters Buttons be "selectable"?


gkatsev:
## Description
As part of https://github.com/videojs/video.js/pull/3472, we're making chapter menu items "selectable" so that they show up correctly in the UI. However, this is strictly the correct thing to do from the perspective of accessibility and screen readers.
We should think about what the control tries to achieve and what the correct accessibility implications for this are. Then, we should try and address those as best we can.

OwenEdwards:
The `selectable` option for menu items controls whether they are given the ARIA role of `menuitemcheckbox`, or simply `menuitem`; it then affects whether the item can be marked as `selected` with CSS styling (`vjs-selected`) and the `aria-checked` attribute.

This is analogous to the highlighting of selected menu options in this ARIA example: http://oaa-accessibility.org/example/26/ (note that this example contains some out-of-date ARIA implementation, and is included just to illustrate the visual effect of a menu item being selected).

In fact, there needs to be some further consideration of whether the selectable menu items are strictly `menuitemcheckbox` (multiple independently selectable options, like checkboxes), or `menuitemradiobutton` (selecting one option unselects all other options, like radio buttons) - in the case of the captions menu and the subtitles menu, the latter would seem to be the case.

However, the `Chapters` menu is a somewhat hybrid control, more similar to the progress bar. A user may select one item from the menu (like clicking on the progress bar), but as playback progresses, the chapter which is currently playing will change. Additionally, the WebVTT format for Chapters explicitly allows nested chapters (although video.js may not support them currently). So the concept of `selected` for Chapters doesn't really map to what a screen reader user, or sighted user, would expect, especially if the `selected` item doesn't change as playback progresses.

PR #5171, opened 2018-05-15
Change menu items to ARIA role of "menuitemradio" when only one item can be selected

### LINKED ISSUE #5136: text track menus use role of menuitemcheckbox, menuitemradio more appropriate
Text Track Menus currently use role of menuitemcheckbox for the menu items. This isn't strictly correct because we only allow one text track at a time. Instead, it would be more accurate to use menuitemradio.
Chapters menu is also a weird one because while using it, it's acts like a menuitemradio, but it also gets updated automatically like a progress bar as you play the video. menuitemradio is probably most appropriate for it.

OwenEdwards:
## Description
Change menu items to ARIA role of "menuitemradio" when only one item can be selected can be selected (i.e. in the Captions, Subtitles, Descriptions, Chapters, Audio Tracks, and Rate menus). Fixes #5136.

## Specific Changes proposed
Add an option of `multiSelectable` on `MenuItem` component; if `true`, the item is marked as a checkbox (`role="menuitemcheckbox"`), otherwise the item is marked as a radio button (`role="menuitemradio"`). All existing menu items, except the Captions Settings Menu, will be marked as radio buttons (the Settings Menu item has `role="menuitem"`), as will any components which extend the existing components and don't set 'multiSelectable'.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/audio-track-controls/audio-track-button.js
+++ b/src/js/control-bar/audio-track-controls/audio-track-button.js
@@ -62,7 +62,9 @@ class AudioTrackButton extends TrackButton {
       items.push(new AudioTrackMenuItem(this.player_, {
         track,
         // MenuItem is selectable
-        selectable: true
+        selectable: true,
+        // MenuItem is NOT multiSelectable (i.e. only one can be marked "selected" at a time)
+        multiSelectable: false
       }));
     }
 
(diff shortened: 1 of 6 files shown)


prozak915:
What's now??

OwenEdwards:
Also fixes #3663

gkatsev:
Is the decision regarding Chapters is that it's fine for them to be selectable so no need to keep that issue open?

gkatsev:
Ah, nice. I thought it already did that? Maybe it went missing at some point.

OwenEdwards:
I haven't verified it with any SRs, so the Issue is more of a note to check it, and perhaps add some hidden text which clarifies that it's a little more complicated than just radio buttons. We also need to verify that it doesn't announce via an SR when the menu isn't focused/opened, like I noted for other menus in #5172.

Issue #4583, opened 2017-08-21
JAWS & IE11: initial "Big Play Button" (BPB) and Play button is silent when tabbing to it


serenanubi:
## Description
Big play button is silent upon tabbing to it in IE11 with JAWS 18.

## Steps to reproduce
Go to videojs.com and use the tab button until the focus gets to the BPB.

### Expected
Once the focus is on the BPB, JAWS should say "Play Video button".

### Actual
JAWS is silent.

### Error output
N/A

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
Latest version (v6.2.5).
#### browsers
IE11
#### OSes
Windows 7
### plugins
Not sure..

gkatsev:
@OwenEdwards any ideas?

waughcbw:
@gkatsev Any updates?

OwenEdwards:
@gkatsev not sure what's going on here, but I see this behavior on the http://videojs.com/ site with IE11 & JAWS18: if you use arrow keys to move to the BPB, JAWS announces it but says something before "Play Video" (it sounds like "key"); if you Tab to the BPB, JAWS doesn't say anything.

But looking at the HTML, the `<span>` with `aria-hidden="true"` for the icon that you put in is missing. Did it somehow get dropped from the latest version of video.js?

serenanubi:
@gkatsev @OwenEdwards All the a11y improvements are great. It'd be really nice to get this fixed for JAWS & IE11. Any movement on this would be appreciated. I know IE11 has had issues with aria-live, so I was thinking that may have been the issue. 

I only mentioned the example on http://videojs.com, but this is affecting our users as well.

OwenEdwards:
@serenanubi I looked into this a little, and as I noted in #4601, there's some confusion because the http://videojs.com site is actually using v5.x, not v6.x of video.js. So I wanted to confirm that, when you say it's affecting your users, that you're using v6.2.5 as you mentioned in the original issue report?

serenanubi:
Actually we're using v5.x, which has the issue, but I was hoping v6.2.5 would have a fix, so I was testing our codebase with v6.2.5 instead to verify if the issue had been resolved, but it seems to still be a problem. Sorry for the confusion (and assumption on my part).

OwenEdwards:
@serenanubi is this still an issue in v6.x releases? If not, can we close this issue?

serenanubi:
Yes, this is still an issue in v6.x releases.

[...]

PR #5173, opened 2018-05-15
Workaround for #4583 (JAWS+IE doesn't announce BPB or play button)

OwenEdwards:
## Description
Workaround for #4583 (JAWS+IE doesn't announce BPB or play button)

## Specific Changes proposed
Add ARIA `role="application"` to the `<video>` element if the browser is IE, as recommended by @stevefaulkner in https://github.com/FreedomScientific/VFO-standards-support/issues/78#issuecomment-388327926.

## Requirements Checklist
- [x] Change has been verified in an actual browser (IE11 & JAWS 2018)

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -32,6 +32,7 @@ import * as middleware from './tech/middleware.js';
 import {ALL as TRACK_TYPES} from './tracks/track-types';
 import filterSource from './utils/filter-source';
 import {findMimetype} from './utils/mimetypes';
+import {IE_VERSION} from './utils/browser';
 
 // The following imports are used only to ensure that the corresponding modules
 // are always included in the video.js package. Importing the modules will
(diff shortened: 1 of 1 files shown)


OwenEdwards:
@gkatsev the problem with that is that focus will still move to that dummy button (if not, it doesn't fix the problem), and that dummy button will be unlabeled, which isn't good.

Hopefully Freedom Scientific/JAWS will fix this issue soon, but even then it will still work with this workaround since it's a workaround that they proposed themselves, so they'd be really shooting themselves in the foot if they fixed the problem in JAWS and afterward it didn't work with their own workaround!

stevefaulkner:
@OwenEdwards 

> but even then it will still work with this workaround since it's a workaround that they proposed themselves, so they'd be really shooting themselves in the foot if they fixed the problem in JAWS and afterward it didn't work with their own workaround!

There is no reason why the 'workaround' will have any negative effects if/when the JAWS+IE issues is resolved. Use of [`role=application` is conforming](https://www.w3.org/TR/html-aria/#video) for the `video` element in HTML.
Also note I am not a freedom scientific employee, so its not a workaround that freedom proposed.

PR #5041, opened 2018-03-21
refactor: remove IE8 specific changes

gkatsev:
BREAKING CHANGE: remove IE8 specific JavaScript and CSS code. Remove Android 2.3 workaround.

TODO:
* [x] remove weird track workarounds for defineProperty
* [x] use defineProperties where appropriate
* [ ] update vtt.js to remove weird workaround for defineProperty
* [ ] update vtt.js to use defineProperties
* [x] remove IS_OLD_ANDROID canPlayType monkeypatching.
* [x] remove es3 preset
* [x] remove IE8 CSS hacks and workarounds.
* [x] remove videojs-ie8 dep
* [x] update sandbox files appropriately.
* [x] remove IE8 specific tests
* [x] remove IE8 checking in tests and just run them always.
* [x] fix broken tests
* [ ] update videojs-fonts to remove ie8 stuff

PR #5194, opened 2018-05-22
fix(player): video-js embed missing video-js class

gkatsev:
When having a video-js embed with a class attribute, as part of the
changes to remove old IE support (#5041), we overwrote our addition of
the video-js class when it was missing. Instead, we want to make sure
that we don't override the class names again since they are already set
up correctly.

Fixes videojs/http-streaming#100

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -589,7 +589,7 @@ class Player extends Component {
     Object.getOwnPropertyNames(attrs).forEach(function(attr) {
       // don't copy over the class attribute to the player element when we're in a div embed
       // the class is already set up properly in the divEmbed case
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -587,7 +587,12 @@ class Player extends Component {
     tag.removeAttribute('height');
 
     Object.getOwnPropertyNames(attrs).forEach(function(attr) {
-      el.setAttribute(attr, attrs[attr]);
+      // don't copy over the class attribute to the player element when we're in a div embed
+      // the class is already set up properly in the divEmbed case
+      // and we want to make sure that `video-js` class doesn't get lost
```
gesinger: the

PR #5200, opened 2018-05-23
refactor: removed old bug work-around code

BrandonOCasey:
## Description
Found some code that looks like it was added for safari 5 and Chrome 15 support. We should probably test to make sure we don't need it, but it would seem like we don'


gkatsev:
We will not be porting this over to the 6.x release line.

Issue #3927, opened 2017-01-09
Return a Promise From play() when possible


brandonocasey:
## Description
There is currently a PR (#3907) that returns the native value of play to the user. In some cases (such a newer version of Chrome) this will be a Promise. 

I think that we should try to always return a Promise if the browser supports them or a Promise library is provided (`window.Promise || this.options_.Promise`). This means that we will return the native Promise from `play()` if it is returned and return a "psuedo-native" Promise if `play()` does not return a Promise. 

We will also have to be mindful of browsers that don't support promises at all and the [specification for Promises from play](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-play).

misteroneill:
This was actually a question at the meetup yesterday: would we support a polyfill or creating our own promise. So, that's a vote in this direction, I'd say.

gkatsev:
Yup, I don't want to write our own promise implementation :)

misteroneill:
Sorry, I meant we'd return a Promise as @BrandonOCasey is suggesting - not "create our own". Poor phrasing. 😆

shikkaba:
Is there any update on this?

gkatsev:
@shikkaba in Video.js v6, we already are returning a Promise from `play()` if the browser supports this. We have not made progress on making our own promise for the times that the browser doesn't support it.

shikkaba:
@gkatsev Thank you for the reply. I'm just wondering because in Chrome I'm seeing the following:
`VIDEOJS: ERROR: (CODE:4 MEDIA_ERR_SRC_NOT_SUPPORTED) The media could not be loaded, either because the server or network failed or because the format is not supported. [object Object]
 /absorica/js/video.js:489:47 logByType()
 /absorica/js/video.js:607:10 Function.log.error()
 /absorica/js/video.js:22620:11 Player.error()`

KevinBrogan:
videojs 7 does not return the native result from play

gkatsev:
Actually, re-opening this one for the follow-up work for IE11 and browsers that don't return a `promise` from `play()`.

Puritanic:
@gkatsev Any update on this? :)

gkatsev:
So, looks like this issue should be closed. At some point we did add promise support for older browsers. However, you must either provide videojs with a promise polyfill or include a promise polyfill globally on those platforms.

@Puritanic can I ask you to open a new issue with more info filled out in the issue template? Like video.js versions and preferably a reduced test case.

PR #5209, opened 2018-05-24
feat: manual autoplay

BrandonOCasey:
Due to some of the policy changes around autoplay we are going to add a new option. The name of the option and the parameters that the option can be can change in this pr, but I think the general functionality is going to remain the same. Basically you have three paths

1. Try to autoplay using `play` and if that doesn't work do nothing, mostly the same as using the `autoplay` attribute right now.
2. Try to mute and then autoplay using `play`, this will work in most cases
3. Try to autoplay using `play` and if that doesnt work try to mute and then autoplay, which we know will work.

Questions:
* Should we remove option 2 and just recommending using autoplay in that case?
* Should we default to using option 3?
* Should we just use the `autoplay` attribute and accept a string?

Coming Soon
* Tests
* Documentation updates


src/js/player.js
```diff
@@ -1158,6 +1158,49 @@ class Player extends Component {
...
     this.handleTechDurationChange_();
 
+    // if manual auto play is set to muted, any, or any truthy value
+    // we will try to auto play manually, rather than using the attribute
+    //
+    // - muted: mutes the media element and then calls play
+    // - truthy value: will just call play
+    // - any: Will call play and if the promise fails, it will try to mute
+    //        and then play
+    if (this.options_.manualAutoplay) {
```
gkatsev: I still think that maybe `autoplay` is best and have these options:
- `muted` - will mute the video and then play
- `manual` - will call play manually
- `any` - (or maybe another name?) try to play unmuted and then mute if it fails
- any other value do current behavior

then the autoplay getter and setter can also handle this whenever it gets called and the autoplay getter can know whether this option was set and which option.

then this line will just turn to something like `this.autoplay(this.options_.autoplay)` or something.

gkatsev: I realize that this overloads the values for the `autoplay` option, but it really is the best name and we can always deprecate the old values and go forward with just the new stuff.

gkatsev: also, we could theoretically, have the setter return the play promise in that way if called later on

BrandonOCasey: OK we are going to go this route, let me re-work this pr

marguinbc:
Late to the comment party but I feel like the 'new autoplay' should have states/settings that correspond to the browser options in those that allow users to choose autoplay options.  Something like:
- never (results in click-to-play)
- always (same as `any` - tries w/ sound, then muted)
- muted (same as `muted` above)

I'm not sure about `manual` because I feel `always` and `muted` would call `play()` manually anyway but I see no harm in keeping it and having it return the play promise
- manual (calls `play()` and returns play promise?)

And any other value defaults to current behavior...

misteroneill:
@marguinbc `always` doesn't make sense, though, because we can never guarantee that autoplay will work. I don't see why having a `never` option would make sense - that's just the default (no autoplay of any type).

Also, we can't return the play promise with these because it happens in a `loadstart` listener.

marguinbc:
I suppose `any` is a better name than `always` then as they are basically doing the same thing and `always` implies we can always autoplay, vs `any` implies "any autoplay that will work".

I see `never` as being a way to override the `autoplay` attribute or rather explicitly set it to `none` as @gkatsev's comment earlier said 

> any other value do current behavior

Which implies none/default = current behavior and would be dependent on whether the `autoplay` attribute is set or not.

Fair point on the promise... should have looked a the code.

src/js/player.js
```diff
@@ -379,6 +379,13 @@ class Player extends Component {
     tag.controls = false;
     tag.removeAttribute('controls');
 
+    // the attribute overrides the option
+    if (tag.hasAttribute('autoplay')) {
+      this.options_.autoplay = true;
```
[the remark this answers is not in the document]
BrandonOCasey: yeah, basically we would `autoplay`: `true` here and not `autoplay`: `play`

src/js/player.js
```diff
@@ -926,13 +933,17 @@ class Player extends Component {
...
+    const autoplay = typeof this.autoplay() === 'string' ? false : this.autoplay();
+
     // Grab tech-specific options from player options and add source and parent element to use.
     const techOptions = {
       source,
       'nativeControlsForTouch': this.options_.nativeControlsForTouch,
       'playerId': this.id(),
       'techId': `${this.id()}_${titleTechName}_api`,
-      'autoplay': this.options_.autoplay,
+      autoplay,
```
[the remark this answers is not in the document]
BrandonOCasey: ok

src/js/player.js
```diff
@@ -2747,22 +2803,48 @@ class Player extends Component {
...
+    // getter usage
+    if (value === undefined) {
+      return this.options_.autoplay;
+    }
+
+    let techAutoplay;
+
+    // if the value is a valid string set it to that
+    if (typeof value === 'string' && (/(any|play|muted)/).test(value)) {
       this.options_.autoplay = value;
```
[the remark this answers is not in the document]
BrandonOCasey: I think that autoplay is actually and option, and every time we have to switch techs we have to pass it to the tech. While I see the cache as more of a cache of the techs value on our end. I think that this is more of an option, but it has more possible values than what the tech has.

test/unit/player.test.js
```diff
@@ -1904,3 +1904,253 @@ QUnit.test('disposing a tech that dit NOT set a poster, should keep the poster',
 
   player.dispose();
 });
+
+QUnit.module('Player autoplay', {
```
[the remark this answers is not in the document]
BrandonOCasey: yeah I will do that.

src/js/player.js
```diff
@@ -2747,22 +2803,48 @@ class Player extends Component {
...
+      techAutoplay = false;
+      // if the value is a boolean set it to that
+    } else if (typeof value === 'boolean') {
+      this.options_.autoplay = value;
+      // any other value sets autoplay to true
+    } else {
+      this.options_.autoplay = true;
+    }
+
+    techAutoplay = techAutoplay || value;
```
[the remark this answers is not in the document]

src/js/player.js
```diff
@@ -2747,22 +2803,48 @@ class Player extends Component {
...
+    if (typeof value === 'string' && (/(any|play|muted)/).test(value)) {
       this.options_.autoplay = value;
-      return;
+      techAutoplay = false;
+      // if the value is a boolean set it to that
+    } else if (typeof value === 'boolean') {
+      this.options_.autoplay = value;
+      // any other value sets autoplay to true
+    } else {
+      this.options_.autoplay = true;
```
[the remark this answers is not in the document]
BrandonOCasey: Let me see what the video element does, I think any value on autoplay is treated as true.

src/js/player.js
```diff
@@ -2747,22 +2803,48 @@ class Player extends Component {
...
+    }
+
+    techAutoplay = techAutoplay || value;
+
+    // if we don't have a tech do not queue up
+    // a setAutoplay call. This option will be passed
+    // in the constructor for the tech
+    if (this.tech_) {
+      this.techCall_('setAutoplay', techAutoplay);
```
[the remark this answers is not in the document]
BrandonOCasey: I think I did this because in the case where autoplay is currently `true` on the tech, we want to set it to `false`, maybe that doesn't remove the attribute like I think that it should, I will test this out as well.

docs/guides/options.md
```diff
@@ -53,11 +53,34 @@ Each of these options is also available as a [standard `<video>` element attribu
 
 ### `autoplay`
 
-> Type: `boolean`
+> Type: `boolean|string`
+> NOTE: At this point the autoplay attribute and option are NOT a guarantee that your video will autoplay.
```
misteroneill: * `At this point` should be followed by a comma.

docs/guides/options.md
```diff
@@ -53,11 +53,34 @@ Each of these options is also available as a [standard `<video>` element attribu
...
+> NOTE2: If there is an attribute on the media element the option will be ignored.
+> NOTE3: You cannot pass a string value in the attribute, you must pass it in the videojs options
+
+Instead of using the `autoplay` attribute you should pass an `autoplay` option to videojs. The following values
```
misteroneill: I think `videojs` should have backticks here as it refers to the function rather than the library.

docs/guides/faq.md
```diff
@@ -134,38 +134,30 @@ techs/plugins made available to Video.js. For more information on media formats
...
-```
+Due to recent changes in autoplay behavior we no longer recommend using the `autoplay` attribute
+on the `video` element. It's still supported by Video.js but, many browsers, including Chrome, are changing their
+`autoplay` attribute behavior.
 
-```js
-var player = videojs('my-video', {
-  autoplay: true
-});
+For more information on the autoplay changes see our blog post: https://blog.videojs.com/autoplay-best-practices-with-video-js/
```
mister-ben: I think this section flows better with this line at the end.

docs/guides/faq.md
```diff
@@ -134,38 +134,30 @@ techs/plugins made available to Video.js. For more information on media formats
...
 
-// or
-
-player.autoplay(true);
-```
+Instead we recommend using the `autoplay` option rather than the `autoplay` attribute, for more information on using that.
+see the [autoplay option][autoplay-option] in the Video.js options guide.
```
mister-ben: no . after that

src/js/player.js
```diff
@@ -2747,22 +2803,53 @@ class Player extends Component {
...
-   * @return {string}
+   * @return {boolean|string}
    *         The current value of autoplay when getting
    */
   autoplay(value) {
-    if (value !== undefined) {
-      this.techCall_('setAutoplay', value);
+    // getter usage
+    if (value === undefined) {
+      return this.options_.autoplay;
```
mister-ben: This returns `undefined` rather than `false` if not configured.

BrandonOCasey: good call

BrandonOCasey:
hmmm yeah it could be a good idea to have something like `manual-autoplay-success` and `manual-autoplay-failure`. Thoughts @gkatsev ?

BrandonOCasey:
On thinking about this some more, I think we may want to do this for `play` in general. We might want to have `play-fulfilled`, `play-rejected`, and `play-pending`. Which is all based on states named in the promise specification.

misteroneill:
I think that idea has merit for sure. Maybe a property as well, e.g. `lastPlayState()` that returns a pseudo-symbol like `videojs.[FULFILLED|REJECTED|PENDING]`? Thinking out loud here.

src/js/player.js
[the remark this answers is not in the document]
gkatsev: basically, I think we should keep `this.options_` set to whatever is the initial set of options passed it and use other places to cache transient or instance properties. However, I don't think this is necessarily something for this PR.

src/js/player.js
```diff
@@ -2747,22 +2803,53 @@ class Player extends Component {
...
-      this.techCall_('setAutoplay', value);
+    // getter usage
+    if (value === undefined) {
+      return this.options_.autoplay || false;
+    }
+
+    let techAutoplay;
+
+    // if the value is a valid string set it to that
+    if (typeof value === 'string' && (/(any|play|muted)/).test(value)) {
```
gkatsev: should we call `this.manualAutoplay_` in here? Use case would be something like:
```js
var player = videojs('vid');

player.autoplay('any');
```
I'd expect the player to autoplay in this case. This would also match current behavior (in browsers that support autoplay). Main risk is that we'd want to be able to do something like the following too
```js
var player = videojs('vid', { autoplay: 'any' );

player.autoplay(false);
```
and have it not autoplay.

Also, should we have a string state "off" and deprecate boolean values for it?

gkatsev: Maybe just do this in a followup PR?

gkatsev:
Autoplay events probably make sense. We probably can have just two events `autoplay-success` and `autoplay-failure` and then have extra data on the event object or the data object for events (secondary arg) provide info on autoplay configuration.

I'm hesitant to map the play promise into events unless we have an actual use-case. Most of what we talked about it has been theoretical so far. This is especially since we have moved towards recommending people call `play` themselves and thus they will generally have a Promise available to them directly. We probably want to resurrect https://github.com/videojs/video.js/issues/3927 and actually implement it so that `player.promise` *always* returns a promise.

src/js/player.js
mister-ben: Should `player.autoplay('any')` also return a promise?

BrandonOCasey: I think we probably just want events rather than a promise

BrandonOCasey:
See https://github.com/videojs/video.js/pull/5227

src/js/player.js
```diff
@@ -1208,6 +1221,46 @@ class Player extends Component {
...
+      return;
+    }
+
+    playPromise.then(() => {
+      this.trigger('autoplay-success');
+    }).catch((e) => {
+      // if we are muted or not set to any here there is nothing we can do
+      if (type !== 'any' || this.muted()) {
+        this.trigger('autoplay-failure');
+        return;
```
mister-ben: If we've muted the player and the autoplay still fails, shouldn't we unmute again so a subsequent play triggered by a human has sound?

src/js/player.js
```diff
@@ -1208,6 +1221,46 @@ class Player extends Component {
...
+    // default to an object so that we can check
+    // if "then" is a property on it more easily
+    const playPromise = this.play();
+
+    if (!playPromise || !playPromise.then || !playPromise.catch) {
+      return;
+    }
+
+    playPromise.then(() => {
+      this.trigger('autoplay-success');
```
gkatsev: should we add the `autoplay` value to the event or are we expecting users to call `player.autoplay()` to get the current value?

src/js/player.js
gkatsev: This sounds reasonable to me.

Issue #5199, opened 2018-05-23
Memory leak in Component.requestAnimationFrame


Ismael:
## Description
Component.requestAnimationFrame leaks functions. Take a look at the code here:

https://github.com/videojs/video.js/blob/954f3d979f5514337d7b6195130da96dcb939c3b/src/js/component.js#L1373-L1388

Every time this is called, a disposeFn is attached to this.on('dispose'). The problem is this function is never freed. So you end up with an ever growing list of disposeFns for long-gone frames.

Only if you actually call cancelAnimationFrame a this.off('dispose', disposeFn) is run and the fn is freed.

I don't know how to fix this because I don't understand why you need to save that function.

## How to reproduce the issue
Load a video in Chrome. In dev tools, select Memory tab. Take a heap snapshot. Wait a few minutes, take another snapshot. Select latest snapshot, click "Comparison" (left of "class filter") and compare to previous snapshot.
Look for "(concatenated string)", if there's a bunch of "vjs-raf-2222" (the functions guids), then you are leaking functions. They also appear in "(closure)", but it's a bit harder to identify them there.

gkatsev:
That makes sense, we probably want to clear out this dispose function when no longer necessary. The reason for the disposeFn is so that if the component or player get disposed before the rAF has triggered, we don't want to trigger the callback method because it could produce errors, in some cases.

I guess we probably want to wrap use a separate callback that nulls out the disposeFn and then calls `fn` with the arguments.

Ismael:
Something like this?

``` javascript
const id = window.requestAnimationFrame((t) => {
        if (this.player_) {
          fn(t);
        }
      });
```
And then removing all the code for disposeFn.

I thought of checking this.player_ since it seems to be the most "fundamental" thing that should be alive. But maybe checking this.el_ makes more sense?

gkatsev:
Yeah, checking .el_ is probably better, could even check both.
Also, I was actually event thinking of doing the same thing as in cancelAnimationFrame for removing the dispose handler
https://github.com/videojs/video.js/blob/954f3d979f5514337d7b6195130da96dcb939c3b/src/js/component.js#L1410-L1414

Ismael:
Want me to do a pr?

gkatsev:
That would be great @Ismael!

Ismael:
I don't understand why the tests fail. It requires deeper knowledge of what's going on...

PR #5212, opened 2018-05-25
fix: memory leak on Component.requestAnimationFrame and Component.set…

Ismael:
…Timeout (#5199)

The disposeFn functions created by both methods were never freed, creating a memory leak.
Instead check if the player and element are alive when running the fn.

## Description

The tests are not passing. I don't understand the failure.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1222,13 +1222,11 @@ class Component {
    * Creates a function that runs after an `x` millisecond timeout. This function is a
    * wrapper around `window.setTimeout`. There are a few reasons to use this one
    * instead though:
-   * 1. It gets cleared via  {@link Component#clearTimeout} when
-   *    {@link Component#dispose} gets called.
+   * 1. It checks the player and element are still valid when it's run
+   *
    * 2. The function callback will gets turned into a {@link Component~GenericCallback}
    *
-   * > Note: You can use `window.clearTimeout` on the id returned by this function. This
-   *         will cause its dispose listener not to get cleaned up! Please use
-   *         {@link Component#clearTimeout} or {@link Component#dispose}.
+   * > Note: You can use `window.clearTimeout` on the id returned by this function.
    *
    * @param {Component~GenericCallback} fn
    *        The function that will be run after `timeout`.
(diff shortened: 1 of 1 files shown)


Ismael:
This way you do not need to keep track of the functions at all. It seemed simpler.

Also, how would you give it the id? it's only known after the rAF call.

Ismael:
The tests fails, and I don't understand why. We should fix that before merging.

gkatsev:
Thanks for your patience. Taking a look now and I'm kind of confused as to why this is broken. My current working theory is some weird compilation issue that happened with because of the changes in this PR.

gkatsev:
Oops, ignore what I said earlier. The issue is that there are Components that do not have a `player` property on them. Specifically, the Tech.
I didn't realize that `setTimeout` also changed when I first looked.

gkatsev:
Removing the `this.player_` check in the `if`s seems to make this pass.

also, should we make a similar change in `setInterval`?

src/js/component.js
```diff
@@ -1247,21 +1245,18 @@ class Component {
   setTimeout(fn, timeout) {
     fn = Fn.bind(this, fn);
 
-    const timeoutId = window.setTimeout(fn, timeout);
```
gkatsev: another option is to do something like:

```js
const timeoutId = window.setTimeout(() => {
  this.off('dispose', disposeFn);
  fn();
});
```

I think this would make sure that we don't keep the disposeFn around and we don't get into a situation where we accidentally don't call `fn` because of expectation changes for whether `this.player_` or `this.el_` exists.

Ismael: But disposeFn needs the timeoutId that's only available after setTimeout. I don't think you can get disposeFn into that closure after it was declared.

gkatsev: maybe we need to declare them as `var`s rather than lets so things are available properly in scope with the value being set properly due to how those vars work.

Maybe there are other ways of doing it, not sure.

gkatsev:
@Ismael thanks for getting this started.
I'll be fixing this in #5294.

PR #5227, opened 2018-06-04
feat: always return a promise from play, if supported

### LINKED ISSUE #3927: Return a Promise From play() when possible

BrandonOCasey:
## Description
Always return a Promise from play when we have a Promise object.

Fixes videojs/video.js#3927

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1873,12 +1873,37 @@ class Player extends Component {
    * Attempt to begin playback at the first opportunity.
    *
    * @return {Promise|undefined}
-   *         Returns a `Promise` only if the browser returns one and the player
-   *         is ready to begin playback. For some browsers and all non-ready
-   *         situations, this will return `undefined`.
+   *         Returns a promise if the browser supports Promises (or one
+   *         was passed in as an option). This promise will be resolved on
+   *         the return value of play. If this is undefined it will fulfill the
+   *         promise chain otherwise the promise chain will be fulfilled when
+   *         the promise from play is fulfilled.
    */
   play() {
+    const promise = this.options_.Promise || window.Promise;
 
+    if (promise) {
+      return new Promise((resolve) => {
+        this.play_(resolve);
+      });
+    }
+
+    return this.play_();
+  }
+
+  /**
+   * The actual logic for play, takes a callback that will be resolved on the
+   * return value of play. This allows us to resolve to the play promise if there
+   * is one on modern browsers.
+   *
+   * @private
+   * @param {Function} [callback]
+   *        The callback that should be called when the techs play is actually called
+   */
+  play_(callback) {
+    if (!callback) {
+      callback = () => {};
+    }
     // If this is called while we have a play queued up on a loadstart, remove
     // that listener to avoid getting in a potentially bad state.
     if (this.playOnLoadstart_) {
@@ -1898,12 +1923,13 @@ class Player extends Component {
       this.playWaitingForReady_ = true;
       this.ready(() => {
         this.playWaitingForReady_ = false;
-        silencePromise(this.play());
+        callback(this.play());
       });
 
     // If the player/tech is ready and we have a source, we can attempt playback.
     } else if (!this.changingSrc_ && (this.src() || this.currentSrc())) {
-      return this.techGet_('play');
+      callback(this.techGet_('play'));
+      return;
 
     // If the tech is ready, but we do not have a source, we'll need to wait
     // for both the `ready` and a `loadstart` when the source is finally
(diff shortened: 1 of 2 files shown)


test/unit/player.test.js
```diff
@@ -1148,7 +1149,18 @@ QUnit.test('play promise should resolve to native value if returned', function(a
   player.tech_.play = () => 'foo';
   const p = player.play();
 
-  assert.equal(p, 'foo', 'play returns foo');
+  const finish = (v) => {
```
BrandonOCasey: now that play will always return a promise if it is supported we have to use `then` to check the resolved value.

src/js/player.js
```diff
@@ -1873,12 +1873,37 @@ class Player extends Component {
...
   play() {
+    const promise = this.options_.Promise || window.Promise;
 
+    if (promise) {
+      return new Promise((resolve) => {
+        this.play_(resolve);
+      });
+    }
+
+    return this.play_();
```
BrandonOCasey: ie11 does not support promises, so that should be the only platform this happens on.

src/js/player.js
```diff
@@ -1873,12 +1873,34 @@ class Player extends Component {
...
+   *         was passed in as an option). This promise will be resolved on
+   *         the return value of play. If this is undefined it will fulfill the
+   *         promise chain otherwise the promise chain will be fulfilled when
+   *         the promise from play is fulfilled.
    */
   play() {
+    const promise = this.options_.Promise || window.Promise;
 
+    if (promise) {
+      return new Promise((resolve) => {
```
gkatsev: I wonder if it's possible to have `play_` return a promise itself or use a promise we give it. Maybe a bit tricky since our `play_` method has some event listeners.

BrandonOCasey: whats really tricky is that IE 11 does not have promise support, so we can't technically return a promise all the time.

gkatsev: yeah, IE11 aside, I mean.
We can require a polyfill for the best behavior here, as long as we don't fail if one doesn't provided.

BrandonOCasey:
Two questions for you @gkatsev.
1. By " sure that this is set up to adding rejecting the promise properly as well" Do you mean that we should make sure to silence the play promise? Or do you mean that we should just make sure that the promise we return will be rejected on a rejected play promise? If its number two I think we already will do that because we resolve our promise on the play promise. Then that play promise will actually determine if we are rejected/resolved.
2. Why do we want to return a promise from `play_`? I added `play_` because returning a promise from `play` was really messy when their wasn't always a PromiseClass to use. We could return a fake promise in those cases which would solve it, but I thought it would be better to not do that, so I opted for `play_` to take a callback which will be the `resolve` function when working with promises, or a noop when there are no promises.

gkatsev:
I guess I never responded here.

1. The idea is that we want to shim the `play` promise behavior for when the `play()` doesn't return a promise natively. Like on IE11. So, any change we do now, hopefully shouldn't be too complex that would require us to rewrite a lot of it as part of that work.
I think that what we have now works well if there is a native `play` promise and a step towards our goals. Though, maybe we should even check if just `window.Promise` exists right now because otherwise IE11 will return a promise that *never* resolves. Or maybe even do an IE11-specific check as well.
2. Looking at the code more, my suggestion is probably not helpful/doesn't make much sense based on the current implementation. I'm mostly trying to reduce usage of Promise constructor and passing around the inner `resolve`/`reject` methods.

BrandonOCasey:
So if we want to try and shim the play promise behavior on IE 11. Does that mean we want to:
1. Create a shim Promise that is returned from `Player#play()` when we do not have a real promise to use.
2. Check if the tech returns a Promise from it's `play` method and if it does not return, resolve the `Player#play` promise on the `play` event.

gkatsev:
Yeah, more or less. It gets a bit trickier because you'd want to reject the promise if you get a pause before play and stuff like that but essentially if a native play promise isn't provided we want to make our own promise.

gkatsev:
My main worry was that on IE11 if a Promise Polyfill was on the page, we'd end up returning a forever-pending promise, but I think that we'd get a promise that gets resolved once `play_` does its work, it would just not be rejected and that's probably good enough for this PR.

BrandonOCasey:
awesome, also make sure in the next pr to add something here: https://github.com/videojs/video.js/pull/5227/files#diff-3b0266ff1c037b289ec624ab25b0272eR1928

that will check to see if there is a play promise and if not it will fallback to using `play`/`pause`

Issue #4758, opened 2017-11-20
Captions menu doesn't show up in IE when having longer caption labels


mrdtron:
## Description
When using longer caption labels, the captions/subtitles menu doesn't render correctly in IE/Edge so a user can't select them. 

Sample test case: https://codepen.io/ttom/pen/rYJBrM

## Steps to reproduce
1. Load the player in IE/Edge
2. Start playback and click on captions/subtitle button in the control bar

### Expected
Captions/subtitles is displayed menu with a list of captions 

### Actual
An empty menu with scrollbars appear http://take.ms/MGe1W

#### videojs
6.x/7.x
#### browsers
IE11/Edge
#### OSes
Windows 7/8.1/10

OwenEdwards:
@mrdtron if you look in the Console window when you use the CodePen sample test case (especially on Firefox or Chrome), you'll see that loading of the text tracks is being [blocked because of cross-origin permission issues](http://docs.videojs.com/tutorial-text-tracks.html#text-tracks-from-another-domain). I don't know if this is the exact cause of what you're seeing, but it'll be hard to debug any further without fixing that issue.

mrdtron:
@OwenEdwards CORS error you observed would prevent loading VTTs. However, the issue here has nothing to do with loading text tracks. The problem here is that captions menu gets out of wack in IE when adding more than 4 tracks.  NB Firefox or Chrome don't exhibit this, only IE is affected.

## Chrome or Firefox (4+ caption tracks)
http://take.ms/qwjWM

## IE (4+ caption tracks)
http://take.ms/ginDJ

## IE (4 or less caption tracks)
http://take.ms/9tq96

Applying the following fixes the issue:
```
 .vjs-subs-caps-button .vjs-menu-content{
  overflow: hidden
}
```
https://codepen.io/anon/pen/RjQqgv

OwenEdwards:
Ahh, well that's great, you found the solution! I was thinking that the differences in the CORS error handling might be causing problems with IE knowing the names of the tracks which couldn't be loaded, but apparently, it's just a CSS issue. Can you submit a PR?

gkatsev:
The problem with `hidden` is that it prevents scrolling, so, if you end up having a lot of items, you won't be able to see them all. Example: https://jsfiddle.net/c4ho5whm/.
The menu currently uses `overflow: auto` which creates the scroll bar if the content is too big to fit. Example: https://jsfiddle.net/c4ho5whm/1/.
[...]

PR #5228, opened 2018-06-05
fix: menu sizing when using longer caption labels

bcdarius:
## Description
Proposed solution for #4758 

Currently, if longer captions labels are used the SubsCaps menu gets all messed up in IE11/Edge. This occurs due to the menu size is not adjusted correctly not taking into consideration the CC icon. This, as a result, causes the captions items not to fit in the menu and because they end up too wide and the menu has `overflow: auto` style,  the horizontal scrollbars show up and cause the menu not to render correctly.

## Specific Changes proposed
Do not position the CC icon absolutely. 

## Requirements Checklist
  - [ ] Docs/guides updated
  - [x] Example created ([starter template on JSBin](http://output.jsbin.com/xaducaq))

### DIFF
--- a/src/css/components/_subs-caps.scss
+++ b/src/css/components/_subs-caps.scss
@@ -15,7 +15,8 @@
 }
 
 .video-js .vjs-subs-caps-button + .vjs-menu .vjs-captions-menu-item .vjs-menu-item-text .vjs-icon-placeholder {
-  position: absolute;
+  vertical-align: middle;
+  display: inline-block;
 }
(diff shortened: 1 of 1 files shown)


gkatsev:
This actually works, though, it's not quite aligned properly with the text anymore for a short label, adding `bottom-margin: -0.1em;` seems to fix it. Thoughts?

bcdarius:
Haven't noticed this until you pointed out. Aligning the icon with negative `margin-bottom` value sounds a reasonable solution to me.

gkatsev:
@OwenEdwards hey can you take a look at this? Especially wrt your PR #4599.

OwenEdwards:
This change, rolled into #4599, fixes that PR completely! So this LGTM.

PR #5239, opened 2018-06-08
feat: run autoprefixer on css

BrandonOCasey:
## Description
`autoprefixer` cuts out any vendor prefixes that we don't need. It would probably be better to run this on the source files, but we cannot since we are working with sass. So for now we can run it as a "minification" step. 

## Results
* New size: `9514` bytes gziped
* Old size: `9829` bytes gziped

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -196,6 +196,14 @@ module.exports = function(grunt) {
         dest: 'dist/video-js-' + version.full + '.zip'
       }
     },
+    postcss: {
+      options: {
+        processors: [require('autoprefixer')({browsers: ['defaults', 'ie 11']})]
+      },
+      dist: {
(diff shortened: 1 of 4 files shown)


gkatsev:
We should try and not add stuff to grunt. We probably should add an npm script and then use grunt shell to call it.

build/grunt.js
```diff
@@ -398,9 +406,10 @@ module.exports = function(grunt) {
     'shell:lint',
     'clean:build',
 
-    'shell:rollupall',
+    // 'shell:rollupall',
```
gkatsev: this probably shouldn't be commented out.

BrandonOCasey: woops yeah, I was testing with it commented out.

gkatsev:
This probably doesn't need to be ported to 6.x.

PR #5255, opened 2018-06-15
fix: Allow evented objects, such as components and plugins, to listen to the window object in addition to DOM objects.

misteroneill:
I don't think there's an issue for this (I looked), but the bug is that objects using the new-ish `evented` mixin cannot listen to the `window` object, preventing things like:

```js
component.on(window, 'scroll', throttledListener);
```

This fixes that so anything that's a native `EventTarget` works.

## Specific Changes proposed
Allow any native `EventTarget` to be a listening target.

## Requirements Checklist
  - [ ] ~Docs/guides updated~
  - [ ] ~Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))~

### DIFF
--- a/src/js/mixins/evented.js
+++ b/src/js/mixins/evented.js
@@ -18,7 +18,9 @@ import EventTarget from '../event-target';
  * @return {boolean}
  *         Whether or not an object supports native events.
  */
-const supportsNativeEvents = (object) => object instanceof window.EventTarget;
+const supportsNativeEvents = (object) =>
+  (typeof window.EventTarget === 'function' && object instanceof window.EventTarget) ||
+  typeof object.addEventListener === 'function';
 
 /**
  * Returns whether or not an object has had the evented mixin applied.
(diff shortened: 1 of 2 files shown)


src/js/mixins/evented.js
```diff
@@ -9,6 +9,19 @@ import * as Fn from '../utils/fn';
...
+ *
+ * @param  {Object} object
+ *         An object to test.
+ *
+ * @return {boolean}
+ *         Whether or not an object supports native events.
+ */
+const supportsNativeEvents = (object) =>
+  (typeof window.EventTarget === 'function' && object instanceof window.EventTarget) ||
+  typeof object.addEventListener === 'function';
```
gkatsev: what about only just doing the addEventListener check?

gkatsev:
Confirmed that this works via https://deploy-preview-5255--videojs-docs.netlify.com/test-example/ and
```js
player.on(window, 'click', (e) => console.log(e.type))
```

PR #5262, opened 2018-06-20
feat: include iOS Chrome UA pattern when detecting Chrome in browser module

bcdarius:
## Description
`IS_CHROME` in browser module currently returns `false` for iOS Chrome. 

## Specific Changes proposed
Include iOS Chrome UA string pattern when detecting Chrome. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1010,7 +1010,7 @@ Html5.canOverrideAttributes = function() {
  *         - False otherwise
  */
 Html5.supportsNativeTextTracks = function() {
-  return browser.IS_ANY_SAFARI;
+  return browser.IS_ANY_SAFARI || (browser.IS_IOS && browser.IS_CHROME);
 };
 
 /**
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -61,10 +61,10 @@ export const IS_FIREFOX = (/Firefox/i).test(USER_AGENT);
 export const IS_EDGE = (/Edge/i).test(USER_AGENT);
 export const IS_CHROME = !IS_EDGE && ((/Chrome/i).test(USER_AGENT) || (/CriOS/i).test(USER_AGENT));
 export const CHROME_VERSION = (function() {
-  const match = USER_AGENT.match(/Chrome\/(\d+)/);
+  const match = USER_AGENT.match(/(Chrome|CriOS)\/(\d+)/);
 
-  if (match && match[1]) {
-    return parseFloat(match[1]);
+  if (match && match[2]) {
+    return parseFloat(match[2]);
   }
   return null;
 }());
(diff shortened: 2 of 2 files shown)


gkatsev:
Would this mean that this would identify as both chrome and safari?
Does the chrome version get filled out properly?

bcdarius:
Nope, just Chrome. Safari's UA string doesn't include `CriOS`. I completely overlooked CHROME_VERSION on iOS, thanks for pointing that out.

BrandonOCasey:
After testing this it seems to work as intended. I did notice that `IS_ANY_SAFARI` will be `true` for this, which seems un-intended. 


I also have some questions about this. Since Chrome on iOS is basically just a skin over safari with some chrome features:
  * Do we want `IS_SAFARI` and `IS_ANY_SAFARI` to be `true`  here? 
  * Do we even want chrome to register as chrome in this case? Seems like most of these are used to check browser features and chrome iOS will only have the features of safari.

bcdarius:
@BrandonOCasey That's a fair point. Although while iOS Chrome technically uses the same rendering engine as Safari, there are still some differences in capabilities between these two browsers in iOS. For example, Safari on iOS (since v11.2)  supports FairPlay,  Chrome, however, doesn't. Safari is also typically on a more recent Webkit build than Chrome which would, in some instances, have even differences in rendering. So, perhaps, it isn't quite accurate to call Chrome just a skinned Safari with some feature tacked on top. Considering this, I would say `IS_SAFARI` should be false in Chrome on iOS. 

I'm not entirely sure about `IS_ANY_SAFARI`. Within VideoJS core it appears to be used to check native `TextTrack` support only which is somehow confusing.

BrandonOCasey:
OK. So with that information I would say that we should make `IS_ANY_SAFARI`  equal to `false` for this pr and it should be good to go.

src/js/tech/html5.js
```diff
@@ -1010,7 +1010,7 @@ Html5.canOverrideAttributes = function() {
  *         - False otherwise
  */
 Html5.supportsNativeTextTracks = function() {
-  return browser.IS_ANY_SAFARI;
+  return browser.IS_ANY_SAFARI || (browser.IS_IOS && browser.IS_CHROME);
```
gkatsev: does chrome for iOS handle text tracks like iOS safari does? Are there caption settings available at the browser level or does it use the OS-level settings?

bcdarius: Right now, Chrome and Safari on iOS handles text tracks in the same way  (Chrome, same as Safari, uses OS level settings for captions and subtitles).

gkatsev: cool wanted to confirm that's how it goes.

uulaal:
The naming here caught me off guard. When I saw `IS_ANY_SAFARI`, I thought "any Safari engine" would include Chrome on iOS since it's just a wrapper around Safari. The JSDoc description confirms this expectation even more: "Whether or not this is _**any flavor of Safari**_ - including iOS". Only after testing it (unluckily, already in production) I realized that it excludes Chrome. Maybe renaming it or removing `!IS_CHROME` would better reflect what it actually checks?

PR #5273, opened 2018-06-22
feat: allow plugin deregistration from videojs

BrandonOCasey:
Currently you have to get the base plugin class and then deregister from there, which is a bit weird since you can directly register a plugin from `videojs`. This pr allows plugin de-registration from `videojs` as well.

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -452,6 +452,20 @@ videojs.bind = Fn.bind;
 videojs.registerPlugin = Plugin.registerPlugin;
 
 /**
+ * Deregister a Video.js plugin.
+ *
+ * @borrows plugin:deregisterPlugin as videojs.deregisterPlugin
+ * @method deregisterPlugin
+ *
+ * @param  {string} name
+ *         The name of the plugin to be deregistered. Must be a string and
+ *         must match an existing plugin or a method on the `Player`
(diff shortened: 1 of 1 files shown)


gkatsev:
My main concern is about the weird state we can get into with the way deregistration is currently set up: if you have a registered plugin, invoke it, and then deregister. Should we dispose active instances of the plugin? or just let it be?

gkatsev:
Also, I didn't realize that the deregisterPlugin was available on the Plugin object, I would've preferred it not be visible at all

BrandonOCasey:
I think `deregisterPlugin` should only stop future players from getting said plugin. Plugins that are already on a player should probably be dealt with manually.

gkatsev:
Maybe that's fine? Let's leave this to marinate for a bit.

thijstriemstra:
Im not native english speaker but `unregister` instead of `deregister` sounds better to me.

gkatsev:
They're synonyms but as this person outlines https://english.stackexchange.com/questions/25931/unregister-vs-deregister
deregister has the slight connotation that something was previously registered where-as unregistered can imply that it was never registered in the first place.

Also, given that `Plugin.deregisterPlugin` already exists, I would say we should just stick with it.

PR #5274, opened 2018-06-22
feat: roadmap

gkatsev:
So, this is more or less the roadmap I have. I don't think we want to merge this PR as I'd like to turn each item into an issue if it isn't already but making it a PR makes it easy to make comments on.
This is also more or less the raw list with some detail. I'd still want to go through and put more details into each out as part of turning them into issues. When I make the new issue I'll also close the several open issues and redirect those to the new one.
Also, we probably need to prioritize these so we know which should be worked on sooner and which aren't as important.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,5 +1,7 @@
 - Promise Play
   In addition to making sure that a Promise is always returned from play(), if Promises are supported or polyfilled, we should be able to shim a play() promise if a promise isn't provided to us natively.
+  https://github.com/videojs/video.js/pull/5227
+  https://github.com/videojs/video.js/issues/3927
 - vjs-fill being first class
   vjs-fluid is a first-class option now but vjs-fill is just as useful. We should have a fill() method and option just like fluid
 - loadMedia: https://github.com/videojs/video.js/issues/4342
@@ -42,3 +44,4 @@
   It's possible there are still cases where the native control bar shows up under ours in iOS, we should investigate and fix it
 - Make our language files work with bundlers
 - Investigate `type=module` script support for Video.js
+- Fix the rest of the Video.js issues
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,44 @@
+- Promise Play
+  In addition to making sure that a Promise is always returned from play(), if Promises are supported or polyfilled, we should be able to shim a play() promise if a promise isn't provided to us natively.
+- vjs-fill being first class
+  vjs-fluid is a first-class option now but vjs-fill is just as useful. We should have a fill() method and option just like fluid
+- loadMedia: https://github.com/videojs/video.js/issues/4342
+- refocus docs on handwritten docs rather than jsdoc using docusaurus or docz
+- turn video.js into a monorepo and break it up into many modules
+  especially with the addition of VHS, a lot of times you want just a piece of Video.js. We should slice video.js up into pieces and publish them as separate packages. This may make the need for "custom builds" obsolete
+  For example, `@videojs/core`, `@videojs/tracks`, `@videojs/controls`, `@videojs/utils`.
+- Focus ring changes
+  This may not actually require changes to Video.js and may just be a documentation thing for how to use focus-visible polyfill
+  https://github.com/videojs/video.js/issues/2993#issuecomment-387626945
+- Make framework test pages and test apps
+  Make react, angular, ember, phonegap test apps
+- Custom builds
+  May not be necessary if we split Video.js into packages. The idea is to make sure that it's possible to compile away pieces that are not wanted by the user.
+- Retry after error
+  This can get pretty involved if you consider any error but I think the MVP here is to retry after a MEDIA_ERR_SRC_NOT_SUPPORTED to basically waterfall the sources
+- Menu and control bar usability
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,47 @@
...
+- Responsive sizing
+  We already have some support for responsive sizing and we have an "adaptive" stylesheet but we still can do a much better job at handling different player sizing. Especially with the `playerresize` event.
+  Things we can do:
+    - make sure captions resize properly in relation to player size
+    - hide and show control bar buttons as necessary when the player resizes
+    - change the control bar sizes in relation to player size, for example, in fullscreen make the controls a lot larger
+- Audio skin
+  Currently, we support the audio element but it's feature support for it is confusing. The HTML spec differentiates between video elements and audio elements by whether they have a "canvas" to use for visual information, for things like the poster image, captions, and video. When using an audio element, we should make sure that we have just a control bar and no visual area around it.
+  However, if audio is played with the video element, we should be smart and keep the poster showing along with the control bar and allow for captions to be shown as well.
```
OwenEdwards: If we support audio playback in a video element with a poster, which allows captions to be shown over the poster, we need to provide an API into videojs-vtt.js which allows the developer to specify that captions are shown over the *whole* of the poster, not just the bottom portion as we do for video.

ROADMAP.md
```diff
@@ -0,0 +1,47 @@
...
+- loadMedia: https://github.com/videojs/video.js/issues/4342
+- refocus docs on handwritten docs rather than jsdoc using docusaurus or docz
+- turn video.js into a monorepo and break it up into many modules
+  especially with the addition of VHS, a lot of times you want just a piece of Video.js. We should slice video.js up into pieces and publish them as separate packages. This may make the need for "custom builds" obsolete
+  For example, `@videojs/core`, `@videojs/tracks`, `@videojs/controls`, `@videojs/utils`.
+- Focus ring changes
+  This may not actually require changes to Video.js and may just be a documentation thing for how to use focus-visible polyfill
+  https://github.com/videojs/video.js/issues/2993#issuecomment-387626945
+- Make framework test pages and test apps
+  Make react, angular, ember, phonegap test apps
```
ldayananda: should this be part of the handwritten docs issue?

ROADMAP.md
```diff
@@ -0,0 +1,47 @@
...
+- Custom builds
```
ldayananda: seems like a dup of monorepo

ROADMAP.md
```diff
@@ -0,0 +1,47 @@
...
+  - make it a first class node/bundler citizen
+- ClickableComponent
+  Figure out what is the status of it currently and either mark it for removal and undeprecate it
+- vdata errors
+  We often have issues come up for errors relating to a `vdata12345` error in the console. We should put up a page explaining this issue and then log a warning with a link to it when it happens rather than crashing.
+- iOS and multiple controls
+  It's possible there are still cases where the native control bar shows up under ours in iOS, we should investigate and fix it
+- Make our language files work with bundlers
+- Investigate `type=module` script support for Video.js
+- Fix the rest of the Video.js issues
```
ldayananda: this one's kind of vague 😄

ldayananda: how do you prioritize a catch all? Have a set number of issues to work on per month/every 4 months?

gkatsev: something like that or just look at individual issues. Basically, just so that the other issues don't get lost.

thijstriemstra:
I'd put this in a wiki page. It'll always be outdated and a pita when it lives in the repository.

gkatsev:
@thijstriemstra yeah, I'm not going to merge this in. Just wanted to put it out there as quickly as possible so people can take a look. Ultimately, each one should end up as an issue or something like that.

If you have any thoughts about the roadmap items, for example which should come first, I'd really welcome your feedback.

BrandonOCasey:
speaking of which we should probably update the wiki

ROADMAP.md
```diff
@@ -0,0 +1,47 @@
+- Promise Play
+  In addition to making sure that a Promise is always returned from play(), if Promises are supported or polyfilled, we should be able to shim a play() promise if a promise isn't provided to us natively.
+  https://github.com/videojs/video.js/pull/5227
+  https://github.com/videojs/video.js/issues/3927
+- vjs-fill being first class
+  vjs-fluid is a first-class option now but vjs-fill is just as useful. We should have a fill() method and option just like fluid
+- loadMedia: https://github.com/videojs/video.js/issues/4342
+- refocus docs on handwritten docs rather than jsdoc using docusaurus or docz
+- turn video.js into a monorepo and break it up into many modules
+  especially with the addition of VHS, a lot of times you want just a piece of Video.js. We should slice video.js up into pieces and publish them as separate packages. This may make the need for "custom builds" obsolete
```
forbesjo: Should VHS source code be then moved into the video.js monorepo?

thijstriemstra: can someone point me to a video.js document describing this 'VHS'? I keep seeing it but not sure what it is.

gkatsev: heh, it's what we call the replacement of contrib-hls https://github.com/videojs/http-streaming

gkatsev:
Some of this is going to be planned over the next 3 months. I'm going to close this PR for now but I have a task to port these over into proper github issues.

Issue #5159, opened 2018-05-09
TextTracks change event fired multiple times


dtranBC:
## Description
The "change" event of the TextTracksis triggered multiple times. You can see a page to reproduce the issue here: https://cs1.brightcodes.net/dtran/BC-40804.html

## Steps to reproduce
Load the sample provided
1. Click Play
2. Change closed caption option
3. Observe the Message Log

**Test Scenario: Choose English then work your way down the list**
Expected Result:

> Language Switched: en
> ************
> Language Switched: fr
> ************
> Language Switched: es
> ************
> Language Switched: ja
> ************
> Language Switched: 

Actual Result:

> Language Switched: en
> Language Switched: en
> ************
> Language Switched: fr
> ************
> Language Switched: es
> ************
> Language Switched: ja
> ************
> Language Switched: 

**Test Scenario 2: Starting from cc1 and work your way up the list**
Expected Result:
> Language Switched: ja
> Language Switched: es
> Language Switched: fr
> Language Switched: en

Actual Result:
> Language Switched: ja
> Language Switched: ja
> Language Switched: es
> Language Switched: es
> Language Switched: fr
> Language Switched: fr
> Language Switched: en
> Language Switched: en

## Additional Information
The results are actually different when you are testing in Safari as the duplicate record is occurring when you switch in any order. With Chrome and Firefox, the event is fired multiple times only when you are going backward from the bottom of the list.
On IE Edge, it actually produces triplets event fired.

The issue is reproducible also on mobile playback.

gkatsev:
It's working as expected currently. You get two change events per change because of the way it's implemented, it's non-trivial to make it only fire one. I think the output looks a bit weird because of how you're reading the tracks list and how Video.js updates the tracks when changes are made and it's basically an issue of ordering.
Previously, we should fire a change event for each track that's in the list but now we don't fire one if the mode of the track didn't change. Now, it just fires per track that's changed and given that we only allow one to be showing, we get two `change` events. Ideally, we'd only fire one change event.

dtranBC:
@gkatsev  - Just to clarify, the multiple events are as expected, however, my only confusion is that the multiple events are only fired if you move up the list rather than down the list.

For example, if you change from English to French, the event only fires once. But if you move back from French to English, the events are fired twice.

Should I not also expect the events to fire twice when I move down the list as well? Either direction, the behavior of the event should be the same right?

dtranBC:
[...]

PR #5278, opened 2018-06-26
feat: new text track event - ttChange

dtranBC:
## Description
As discussed in #5159 , the current event "change" is fired when modechange is triggered. This means for every text track change, there are multiple events fired and hence requires manual filtering of the events to have the single event. Ideally, when a text track is changed, there should only be one event fired.

## Specific Changes proposed
The change proposed is to add a new event as to not affect or change the behavior of text-tracks functions which is dependent on modechange. To reduce the event to only fire for the text track which is showing, the new event is triggered only when the textTrackMode is 'showing' which would only fire once for each text track change.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-list.js
+++ b/src/js/tracks/text-track-list.js
@@ -31,6 +31,14 @@ class TextTrackList extends TrackList {
       this.trigger('change');
     }));
 
+    /**
+     * @listens TextTrack#textTrackChange
+     * @fires TrackList#ttChange
+     */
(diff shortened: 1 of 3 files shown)


gkatsev:
This seems like a workaround for #5159 that could be outside of core Video.js. It doesn't seem to address the core of the issue.
Adding non-spec events isn't inherently a bad thing but we shouldn't add those unless we're sure there's no better way of doing it.

PR #5283, opened 2018-06-28
fix: autoplay throws 'undefined promise' error on some browsers.

ayamamori:
## Description
Handling possible exception when calling `player.autoplay()` so that the video-playing fallback implementation should work.

## Specific Changes proposed
If the source is not loaded, `player.play()` returns `undefined` on
Firefox and Edge.
This causes an exception when `player.autoplay()` called.
With this fix, player will not make an exception and the fallback
(Retrying video-play after loading video) will be enabled.

### Reproduce

Reproduced on Firefox(Mac) and Edge.
Not reproduced on Chrome(Mac) and Safari

- Reproducing code
```html
<!DOCTYPE html>
<html>
  <head> 
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
    <link href="http://vjs.zencdn.net/7.0/video-js.min.css" rel="stylesheet">
    <script src="./video.js"></script>
  </head>
  <body>
    <video id="test-video" class="video-js vjs-default-skin" controls>
      <source src="https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8" type='application/vnd.apple.mpegurl' />
      <!-- <source id="test-source" src="https://www.sample-videos.com/video/mp4/240/big_buck_bunny_240p_1mb.mp4" type='video/mp4' /> -->

    </video>
    <script>
      var main = function(){
        player = videojs("test-video");
        player.on(player, 'loadstart', function(){player.autoplay('any');});
      }
      main();
    </script>
  </body>
</html>
```

- Expected

Video is played as either muted or unmuted, without clicking the video element.

- Actual

video was not played, until clicking the video element. 

- Log

No related console logs were generated.

- Other

`player.autoplay()` is a very nice and comfortable feature! 
I hope it will release as soon as possible.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1259,7 +1259,6 @@ class Player extends Component {
     }
 
     if (!promise || !promise.then || !promise.catch) {
-      this.trigger({type: 'autoplay-failure', autoplay: type});
       return;
     }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1258,6 +1258,10 @@ class Player extends Component {
       promise = this.play();
     }
 
+    if (!promise || !promise.then || !promise.catch) {
+      this.trigger({type: 'autoplay-failure', autoplay: type});
+      return;
+    }
     return promise.then(() => {
       this.trigger({type: 'autoplay-success', autoplay: type});
     }).catch((e) => {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1262,6 +1262,7 @@ class Player extends Component {
       this.trigger({type: 'autoplay-failure', autoplay: type});
       return;
     }
+
     return promise.then(() => {
       this.trigger({type: 'autoplay-success', autoplay: type});
     }).catch((e) => {


gkatsev:
Good catch @mister-ben. I guess we should just return if there is no promise. Then as part of #3927, we can make it handle things properly if there's no play promise.

ayamamori:
I agree that changing `play()` to always-returning-Promise function should resolve the issue.

I tried this approach and I could resolve the issue itself, but some other unrelated tests, I think, were getting failed.
I couldn't figure out that these tests were actually broken, or tests were outdated and should be revised.
Can someone investigate this? 



...
Ah, wait. Already working on it in #5227? If so and #5227 is merged, please close this PR.

gkatsev:
There's some more work required for #5227 before it can be merged. I think updating this PR to just return if a promise isn't given will fix any errors and moves us closer to where we want to be.

ayamamori:
So, what kind of updates are needed for this PR?
You mean 
```js
// src/js/player.js#L1262
+      this.trigger({type: 'autoplay-failure', autoplay: type});
```
should be removed?

mister-ben:
Yes, just that.

ayamamori:
Done!

PR #5294, opened 2018-07-02
perf: setTimeout and requestAnimationFrame memory leak

### LINKED ISSUE #5199: Memory leak in Component.requestAnimationFrame

gkatsev:
Our setTimeout and requestAnimationFrame methods added dispose handlers so that they get cancelled if the component is disposed before they get a chance to run. However, we were only clearing out these dispose handlers if we cleared the timeout or the rAF manually. Instead make sure that we remove the dispose handler when it is no longer needed.

Fixes #5199.

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1313,18 +1313,11 @@ class Component {
    * @see [Similar to]{@link https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/setInterval}
    */
   setInterval(fn, interval) {
-    // declare as variables so they are properly available in interval function
-    // eslint-disable-next-line
-    var intervalId, disposeFn;
-
     fn = Fn.bind(this, fn);
 
-    intervalId = window.setInterval(() => {
-      this.off('dispose', disposeFn);
-      fn();
-    }, interval);
+    const intervalId = window.setInterval(fn, interval);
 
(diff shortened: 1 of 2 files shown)


gkatsev:
Waiting for tests to fail before removing the revert commit and also removing the changes to setInterval.

PR #5295, opened 2018-07-03
feat(middleware): make setSource be optional

gkatsev:
Adding tests first and implementation will be available after tests fail.

setSource is useful if you care to be fiddling with the source or doing some work depending on what source is set. However, sometimes, you don't need a setSource and want the middleware to always be selected.
Now, if setSource is missing, it will implicitly be included in the middleware chain.

### DIFF
--- a/docs/guides/middleware.md
+++ b/docs/guides/middleware.md
@@ -25,7 +25,9 @@ There are a few special methods that affect middleware: `setSource` and `setTech
 
 This method will setup the routing between a specific source and middleware and eventually sets the source on the `Tech`.
 
-If your middleware is not manipulating, redirecting or rejecting the source, you can pass along the source by doing the following:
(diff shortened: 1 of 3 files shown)


mister-ben:
Looks good, but docs/guides/middleware.md will need to be updated.

gkatsev:
Good call, I'll update docs tomorrow.

docs/guides/middleware.md
```diff
@@ -37,6 +38,8 @@ videojs.use('*', function(player) {
 });
 ```
 
+Alternative, in versions greater than 7.0.5, you may omit `setSource` entirely and the middleware will get selected implicitly.
```
ldayananda: I would just put this above the explanation of how to pass along the source, since it'll be default behavior from now on

Issue #4757, opened 2017-11-17
Browser version checks returning `null` is problematic


alex-barstow:
## Description
Browser version checks (`IOS_VERSION`, `ANDROID_VERSION`, `IE_VERSION`, `CHROME_VERSION`) return `null` if the user agent doesn't match the eponymous platform ex. `IOS_VERSION`is null on Android, `CHROME_VERSION` is null in Firefox, etc.

Since null gets [coerced into a 0](http://www.ecma-international.org/ecma-262/7.0/index.html#sec-tonumber) in numerical comparisons, this can lead to some confusing scenarios. For example, `IOS_VERSION < 11` will be `true` on Chrome (and every non-iOS platform). Currently, there _are_ a few workarounds to avoid trouble when using these version checks, such as:
1. Precede a version check with a platform check:`if (IS_IOS && IOS_VERSION < 11) {...}` 
2. Precede a version check with a null check:`if (IOS_VERSION && IOS_VERSION < 11) {...}` 

Additionally, there isn't really a concern when it comes to `browser_version > x` comparisons since it's probably safe to say that `0 > version number` will be false. _Regardless_, I think the behavior of these checks isn't as intuitive as it could be and makes it too easy to introduce bugs if one isn't careful.

## Possible Solution
The simplest solution might be to just have these IIFEs return `undefined` instead of `null` in cases where there is no user agent match found. Since `undefined` gets coerced into `NaN`, any numerical comparison would return `false`, which is more intuitive than the current behavior.

PR #5298, opened 2018-07-05
feat: Browser module improvements

### LINKED ISSUE #4917: IOS_VERSION returns major version and not major.minor.patch
## Description
`videojs.browser.IOS_VERSION` returns only the major iOS version and not the full version as I expected.

I have worked around my issue by using feature detection instead but it would be worth documenting this API's behavior.

## Steps to reproduce
1. On mobile Safari run `videojs.browsers.IOS_VERSION`

### Expected
The string `'11.2.0'` or the number `11.2` (similar to `ANDROID_VERSION`)

### Actual
The string `'11'`

bcdarius:
## Description
Closes #4917 and #4757 

## Specific Changes proposed
- Return `Major.Minor` version for `IOS_VERSION` if `Minor` is available. If `Minor` isn't available return Major only.
- If the User Agent doesn't match the platform, return `undefined` rather than `null` for *_VERSION constants.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -26,12 +26,24 @@ export const IS_IPOD = (/iPod/i).test(USER_AGENT);
 export const IS_IOS = IS_IPHONE || IS_IPAD || IS_IPOD;
 
 export const IOS_VERSION = (function() {
-  const match = USER_AGENT.match(/OS (\d+)_/i);
+  // This matches iOS Major.Minor versions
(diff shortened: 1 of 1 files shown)


gkatsev:
Unfortunately, these changes are potentially breaking and we're not at all ready to have a new major version of Video.js. When we get back to making a Video.js 8, we'll take a look back at this.

PR #5309, opened 2018-07-11
Copy all tag attributes after the tags attributes are modified

featheast:
## Description
In the existing code, the attributes of the tag is grabbed copied over first, but then a few modifications are applied afterwards (tabindex, role, width, height). The will make the attrs variable missing all the modifications had been applied.

## Specific Changes proposed
Only copy over the attributes once all modifications are done

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -562,10 +562,6 @@ class Player extends Component {
       el = this.el_ = super.createEl('div');
     }
 
-    // Copy over all the attributes from the tag, including ID and class
-    // ID will now reference player box, not the video tag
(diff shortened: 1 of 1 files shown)


gkatsev:
You're right that we want to make sure that we keep the other attribute modifications. However, we can't just move it down because otherwise, we can lose information depending on the [different embed types of Video.js](http://docs.videojs.com/tutorial-embeds.html).

Since it's possible that the `tag` changes in the `if (divEmbed)` statement. We probably want to update the attrs object as we modify the `tag`s attributes.

PR #5311, opened 2018-07-11
Added all missing translation for CZ_cs

Akxe:
Don't know how to translate caption and description, they are now noted with exclamation point as this: !caption!. Upon explanation I can hopefully translate them too.


gkatsev:
Captions are a specific subset of Subtitles. Subtitles, for example, only have spoken word, but captions include other auditory cues like sound effects or music.
Descriptions are basically subtitles that go beyond even further and will actually describe the video in detail. Like what the characters are doing or the scenery.
@OwenEdwards did you have anything to add?

Akxe:
Added those two, description is in czech translated as subtitles for blind, but I wasn't sure about captions... maybe some day, someone will translate it better...

Akxe:
Also, a bit offside, but I am trying to translate subtitle settings window, but none of the values translates, was it added in version 7 of VideoJS, or am I missing something? (using lastest v6)

gkatsev:
captions is often called subtitles for the hard of hearing.
descriptions as subtitles for the blind is probably good enough.

the subtitle settings has been around for a long time. Make sure you aren't forcing native subtitles and it is only available for our non-native subtitles.

lang/cs.json
```diff
@@ -1,26 +1,85 @@
...
+  "progress bar timing: currentTime={1} duration={2}": "{1} z {2}",
   "Fullscreen": "Celá obrazovka",
   "Non-Fullscreen": "Zmenšená obrazovka",
   "Mute": "Ztlumit zvuk",
   "Unmute": "Přehrát zvuk",
   "Playback Rate": "Rychlost přehrávání",
   "Subtitles": "Titulky",
-  "subtitles off": "Titulky vypnuty",
+  "subtitles off": "Bez titulků",
   "Captions": "Popisky",
```
OwenEdwards: Using Google Translate for "Subtitles for the Deaf" or "Subtitles for the Hard of Hearing", I get "Titulky pro neslyšící", whereas "Popisky" gets translated back as "labels".

lang/cs.json
```diff
@@ -1,26 +1,85 @@
...
   "captions off": "Popisky vypnuty",
   "Chapters": "Kapitoly",
+  "Descriptions": "Popisky",
```
OwenEdwards: Again, using Google Translate, "Descriptions for the Blind" comes back as "Popisy pro nevidomé" - does that work?

Akxe: "Popisky" for captions is perfect! With description it works for both, "popisek" is more like more like describe, so it may make more sense.

Akxe:
This should be final, but I am still not seeing my translations on subtitle settings window, more in #5316.

gkatsev:
Thanks @Akxe!

PR #5313, opened 2018-07-11
chore(welcome bot): add welcome bot config

gkatsev:


.github/config.yml
```diff
@@ -0,0 +1,29 @@
...
+  To help make it easier for us to investigate your issue, please follow the [contributing guidelines](https://github.com/videojs/video.js/blob/master/CONTRIBUTING.md) and [issue template](https://github.com/videojs/video.js/blob/master/.github/ISSUE_TEMPLATE.md).
+
+# Configuration for new-pr-welcome - https://github.com/behaviorbot/new-pr-welcome
+
+# Comment to be posted to on PRs from first time contributors in your repository
+newPRWelcomeComment: |
+  💖 Thanks for opening this pull request! 💖
+  Things that will help get your PR across the finish line:
+
+  - Run `npm run lint -- --errors` locally to catch formatting errors earlier.
```
BrandonOCasey: maybe this should just mention that `npm run test` passes locally?

.github/config.yml
```diff
@@ -0,0 +1,29 @@
...
+  - Include tests when adding/changing behavior.
+  - Include screenshots and animated GIFs whenever possible.
```
BrandonOCasey: Do they need to do this? Won't netlify do this for us most of the time? Should we mention that here somehow?

.github/config.yml
```diff
@@ -0,0 +1,29 @@
+# Configuration for welcome - https://github.com/behaviorbot/welcome
+# Heavily inspired by https://github.com/electron/electron/blob/master/.github/config.yml
+
+# Configuration for new-issue-welcome - https://github.com/behaviorbot/new-issue-welcome
+
+# Comment to be posted to on first time issues
+newIssueWelcomeComment: |
+  👋 Thanks for opening your first issue here! If you're reporting a 🐞 bug, please make sure you include steps to reproduce it. We get a lot of issues on this repo, so please be patient and we will get back to you as soon as we can.
```
BrandonOCasey: You may also want remove the part about the steps to reproduce here and just mention that they should follow the issue template (which includes the steps to reproduce).

.github/config.yml
```diff
@@ -0,0 +1,29 @@
...
+  To help make it easier for us to investigate your issue, please follow the [contributing guidelines](https://github.com/videojs/video.js/blob/master/CONTRIBUTING.md) and [issue template](https://github.com/videojs/video.js/blob/master/.github/ISSUE_TEMPLATE.md).
```
BrandonOCasey: Maybe we should just link here instead of the issue template at all. https://github.com/videojs/video.js/blob/master/CONTRIBUTING.md#filing-issues s

.github/config.yml
BrandonOCasey: it could also be worth mentioning that they should keep an eye on travis to make sure tests pass

.github/config.yml
gesinger: Maybe move the location of "If you're reporting a 🐞 bug, please make sure you include steps to reproduce it." a bit.

```
Thanks for opening your first issue here! We get a lot of issues on this repo, so please be patient and we will get back to you as soon as we can.
To help make it easier for us to investigate your issue, please follow the [contributing guidelines](https://github.com/videojs/video.js/blob/master/CONTRIBUTING.md) and [issue template](https://github.com/videojs/video.js/blob/master/.github/ISSUE_TEMPLATE.md). And if you're reporting a 🐞 bug, please make sure you include steps to reproduce it.
```

gkatsev: I think it would be good to keep because it's less likely to get ignored as it could be if it's in a link.

gkatsev: Only issue I see with this is that if it's in the end, people may drop off before they get there :)

.github/config.yml
gkatsev: I think it makes sense, particularly for UI changes. Netlify gives us a live example but we still can't be sure what the intent and broken behavior was without it.
Also, screenshots and gifs can help debug stuff that aren't directly UI changes but rendering issues and what not.

.github/config.yml
gkatsev: yeah, maybe just contributing.md is enough. electron only links there.

.github/config.yml
gkatsev: Maybe, I guess the question is how much we want to mention in this comment. We do already mention tests as well in the next bullet point, so, maybe that's enough reminder.

.github/config.yml
BrandonOCasey: OK, yeah your right.

.github/config.yml
```diff
@@ -0,0 +1,30 @@
+# Configuration for welcome - https://github.com/behaviorbot/welcome
+# Heavily inspired by https://github.com/electron/electron/blob/master/.github/config.yml
+
+# Configuration for new-issue-welcome - https://github.com/behaviorbot/new-issue-welcome
+
+# Comment to be posted to on first time issues
+newIssueWelcomeComment: |
+  👋 Thanks for opening your first issue here! If you're reporting a 🐞 bug, please make sure you include steps to reproduce it. We get a lot of issues on this repo, so please be patient and we will get back to you as soon as we can.
```
BrandonOCasey: Maybe this should be similar to the pr text `👋 Thanks for opening your first issue here! 👋` and then the rest on the next lines

.github/config.yml
```diff
@@ -0,0 +1,30 @@
...
+
+# Configuration for new-pr-welcome - https://github.com/behaviorbot/new-pr-welcome
+
+# Comment to be posted to on PRs from first time contributors in your repository
+newPRWelcomeComment: |
+  💖 Thanks for opening this pull request! 💖
+
+  Things that will help get your PR across the finish line:
+
+  - Run `npm run lint -- --errors` locally to catch formatting errors earlier.
```
BrandonOCasey: Actually since we have a git hook for this, it shouldn't be possible to push with linting errors

gkatsev: you can turn it off with `--no-verify`

BrandonOCasey: yeah but then its pretty much your own fault haha

Issue #5140, opened 2018-05-04
Player reset() function causes text track issues


DougGore:
## Description
The `reset()` function in the Player class appears to have seemingly intentional but undesired side effects. As part of the unloading and reloading of the tech plugin the unloadTech, loadTech functions take a JSON snapshot of the associated text tracks prior to the reset and restores them post reset. Unfortunately when using video.js with plugins like DASH and HLS which usually manage their own text tracks they are removing them during the tech plugin dispose stage and having them subsequently restored afterwards without the plugin itself being aware of this.

I have reproduced this in a simplified [test case here](http://jsbin.com/widotiwile/1/edit?html,output), simply play the video and press "Click me". You will see an accumulation of subtitle tracks in the subtitle selector.

In the case of videojs-contrib-dash you may need to pull this change in as well: https://github.com/videojs/videojs-contrib-dash/pull/251

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Have a video with subtitle loaded
2. Invoke player.reset()
3. Load another stream or the same stream again.

### Expected
The expected outcome is that only the subtitle tracks associated with the new video should be displayed.

### Actual
The subtitle tracks from the original video and additionally the new video will both be displayed.

### Error output
No errors seen, this appears to be intentional behaviour. 

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
6.8
#### browsers
Tested with DASH on Chrome and native HLS on Safari
#### OSes
Test with Chrome on Windows 10 and Safari on OSX
### plugins
videojs-contrib-dash (for DASH)
None for HLS

gkatsev:
Having reset remove text tracks sounds like the correct behavior. Would you be interested in making a PR for that? I'd be happy to help you figure out what needs to happen.

DougGore:
Sure, I can pick this up early next week. From studying the source I think the text track backup/restore code is only used as part of the unloadTech and loadTech functions so this should be mostly a case of deleting code.

gkatsev:
Maybe, though, we do want to make sure that text tracks are kept around during tech switches outside of `player.reset()`. If you have any questions or for more real-time chat, drop in by [our slack](http://slack.videojs.com) when you're working on this.

gstrat88:
There is a bigger problem regarding hls, reset and text tracks.
[...]

PR #5327, opened 2018-07-18
Player reset text track unwanted copy

gstrat88:
Add a isResetting flag to loadTech, and unloadTech, so not to copy text tracks.

## Description
When reset is called, text tracks are being kept. By adding the appropriate flag, we keep text tracks, but not when we are resetting the player.
https://github.com/videojs/video.js/issues/5140

## Specific Changes proposed
Added a isResetting flag to the loadTech_  and unloadTech_ methods which controls text tracks copy.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1058,8 +1058,7 @@ class Player extends Component {
         this[props.privateName] = this[props.getterName]();
       });
       this.textTracksJson_ = textTrackConverter.textTracksToJson(this.tech_);
-    } 
-    
+    }
     this.isReady_ = false;
 
     this.tech_.dispose();
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1056,7 +1056,7 @@ class Player extends Component {
       this[props.privateName] = this[props.getterName]();
     });
     this.textTracksJson_ = textTrackConverter.textTracksToJson(this.tech_);
-    
+
     this.isReady_ = false;
 
     this.tech_.dispose();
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks for the PR!

Why not just remove the text tracks in reset before `unloadTech` is called?

gstrat88:
Yeap this is better for sure (and this is my current workaround in my reset wrapper function). 
What about adding a player reset event? In my use case as the player is in fluid mode I set height and  width as zero to reset size, should I proceed to also handle this? In case a player reset event is added this could be done using the event.

gkatsev:
As in trigger a `reset` event when this method runs?

gstrat88:
No forget about the method.
There are two things;

1. Should a player reset event exist? (I believe yes)
2. Should flud dimensions reset when resetting the player?

gkatsev:
It should probably do everything, at least that's the intended behavior. Though, maybe it's worth opening a new issue for discussing this?

src/js/player.js
```diff
@@ -2763,6 +2763,9 @@ class Player extends Component {
    * and calls `reset` on the tech`.
    */
   reset() {
+    if (this.tech_) {
+      this.tech_.clearTracks(['text']);
```
mister-ben: Nitpicking, but this could save a couple of characters since `clearTracks()` puts a string argument into an array.

https://github.com/videojs/video.js/blob/237041698c38177b99de5995b13c298c837ed4a9/src/js/tech/tech.js#L391-L392

PR #5332, opened 2018-07-20
feat: async `change` events in TextTrackList with EventTarget#queueTrigger

### LINKED ISSUE #5159: TextTracks change event fired multiple times

gkatsev:
Trigger the `change` event on the next tick. This means that multiple changes to a track's mode will only result in a single `change` event on its associated TextTrackList rather than 3 events as it may be currently.

Fixes videojs/video.js#5159.

### DIFF
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -158,4 +158,23 @@ EventTarget.prototype.trigger = function(event) {
  */
 EventTarget.prototype.dispatchEvent = EventTarget.prototype.trigger;
 
+const EVENT_MAP = new Map();
+
+EventTarget.prototype.queueTrigger = function(event) {
+  const type = event.type || event;
+  let map = EVENT_MAP.get(this);
+
+  if (!map) {
+    map = new Map();
+    EVENT_MAP.set(this, map);
+  }
+
+  const oldTimeout = map.get(type);
+
+  window.clearTimeout(oldTimeout);
+
+  const timeout = window.setTimeout(() => this.trigger(event), 0);
+  map.set(type, timeout);
+};
+
 export default EventTarget;
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -187,7 +187,7 @@ EventTarget.prototype.queueTrigger = function(event) {
       EVENT_MAP.delete(this);
     }
 
-    this.trigger(event)
+    this.trigger(event);
   }, 0);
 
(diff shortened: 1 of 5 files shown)


gkatsev:
Also, want to make sure we clear out the Map as necessary.

gkatsev:
Tested in IE11 and it's 👌

gkatsev:
I also tested with VHS and sintel and bipbop and captions and they work 👌 too

src/js/event-target.js
```diff
@@ -158,4 +159,38 @@ EventTarget.prototype.trigger = function(event) {
  */
 EventTarget.prototype.dispatchEvent = EventTarget.prototype.trigger;
 
+let EVENT_MAP;
+
+EventTarget.prototype.queueTrigger = function(event) {
```
misteroneill: Love this.

gkatsev:
We've also tested this against current build of contrib-dash to verify that changing tracks is working properly. Thanks @squarebracket!

PR #5335, opened 2018-07-21
feat(player): add playerreset event

gstrat88:
## Description
Added a reset event trigger to reset method. Use this event to reset a fluid player.

## Specific Changes proposed
In reset() method a 'vjs_reset' event is being triggered. In fluid() method event listener is being initialialised. 
Because the first time fluid() runs player is not "evented" yet, the listener is initialised after player is evented. Maybe it would be better to add the event as "delayed" and after player is "evented" init all "delayed" events. 

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/player.js
```diff
@@ -2763,8 +2773,14 @@ class Player extends Component {
    * and calls `reset` on the tech`.
    */
   reset() {
+    if (this.tech_) {
+      this.tech_.clearTracks(['text']);
+    }
     this.loadTech_(this.options_.techOrder[0], null);
     this.techCall_('reset');
+    if (isEvented(this)) {
```
gkatsev: I don't think this check is necessary because the Player will always has event handling.

src/js/player.js
```diff
@@ -768,8 +772,14 @@ class Player extends Component {
 
     this.fluid_ = !!bool;
 
+    if (isEvented(this)) {
```
gkatsev: same here, isEvented check isn't needed.

src/js/player.js
```diff
@@ -2763,8 +2773,14 @@ class Player extends Component {
    * and calls `reset` on the tech`.
    */
   reset() {
+    if (this.tech_) {
+      this.tech_.clearTracks(['text']);
+    }
     this.loadTech_(this.options_.techOrder[0], null);
     this.techCall_('reset');
+    if (isEvented(this)) {
+      this.trigger('vjs_reset');
```
gkatsev: I think the best name for this may be `playerreset`. We already have precedent with the [`playerresize` event](https://github.com/videojs/video.js/pull/4864).

src/js/player.js
```diff
@@ -404,6 +405,9 @@ class Player extends Component {
     // Make this an evented object and use `el_` as its event bus.
     evented(this, {eventBusKey: 'el_'});
 
+    if (this.fluid_) {
+      this.on('vjs_reset', this.updateStyleEl_);
```
gkatsev: so, the idea is to rerun `updateStyleEl_` when reset is called? Makes sense to me.

gkatsev: so, `fluid` gets called in `createEl` so, I'm not sure we actually need this here. We can probably get away with just the piece inside the `fluid` method.

src/js/player.js
gstrat88: Fluid is called first time before the player is evented.

src/js/player.js
gstrat88: ok

src/js/player.js
gstrat88: Fluid is called first time before createEl, player is not evented yet. So no listener. Thats why the listener need to be added after createEl. I will take a look if first time fluid can move after player evented, but a delayed listener also seems a viable solution for me.

gkatsev: I see, 🤔

src/js/player.js
gstrat88: Player starts html5 tech and then resets test fails without it. It does not make a lot of sense though

gkatsev:
The test failures are unrelated, merging anyway. We will not do a release if tests continue failing.

gkatsev:
Passed in master https://travis-ci.org/videojs/video.js/builds/451039377

Issue #5347, opened 2018-07-26
VideoJS does not update the duration display when the duration is NaN or 0.


fketchakeu:
## Description
VideoJS does not update the duration display when the duration is NaN or 0.
Here is a [reduced test case](https://codepen.io/fketchakeu/pen/eVGggG) (Case A).
The issue also occurs when changing the player source and `preload` = 'none'.
Here is a [reduced test case](https://codepen.io/fketchakeu/pen/ZrXpoZ) (Case B)

## Steps to reproduce Case A
Go to [Case A](https://codepen.io/fketchakeu/pen/eVGggG) .
After the video has loaded, note the duration display has been updated.
click on **setDuration(0)** or **setDuration(NaN)**.

### Expected
When clicking on  **setDuration(10)** or  **setDuration(30)** or  **setDuration(0.0001)**, the duration display UI updated.
Thus when clicking on **setDuration(0)** or **setDuration(NaN)** we expect the duration display to update to "0:00”,
or perhaps something like '-:-', in the case of NaN.

### Actual
Clicking on **setDuration(0)** or **setDuration(NaN)** does not update the duration display.

## Steps to reproduce Case B
Go to [Case B](https://codepen.io/fketchakeu/pen/ZrXpoZ)
Click on the "swap SRC" button and observe the duration display.

### Expected
The duration display should reset to 0:00 or something like '-:-'.

### Actual
The duration display doesn't update when the source is changed.

## Additional Information
With plain html5 video, changing the source does indeed reset the duration,
in harmony with [Spec](https://www.w3.org/TR/2011/WD-html5-20110113/video.html#media-element-load-algorithm) (see step 10.4 of media element load algorithm).
e.g. [Html5 equivalent of Case B](https://codepen.io/fketchakeu/pen/gvGMgY)
When the source is changed, in Firefox and Chrome, the player UI is updated to reflect the duration change.

#### videojs
This is occurring in the latest version as of this post (v7.1.0)

gkatsev:
I do see the duration change in case b when the new source is loaded. I see it change from 21s to 30s.

As for case A, what's the usecase for setting a duration of 0 or NaN? Infinity does work and it puts the player into a live state as that's what a duration of infinity represents. I'd except you to only call `player.duration()` if you have a known/better duration than what the media returns.

fketchakeu:
I removed the reference to Infinity, since it doesn't apply here.

[...]

PR #5348, opened 2018-07-26
fix: duration reset, and allow duration NaN or 0 for duration display

fketchakeu:
## Description
This resets the duration display when a new media item is loaded with `preload` set to 'none', and allows the duration NaN or 0 to update the duration display.

Fixes [videojs/video.js#3927](https://github.com/videojs/video.js/issues/5347)

### DIFF
--- a/src/js/control-bar/time-controls/duration-display.js
+++ b/src/js/control-bar/time-controls/duration-display.js
@@ -28,6 +28,11 @@ class DurationDisplay extends TimeDisplay {
     // it has changed
     this.on(player, 'durationchange', this.updateContent);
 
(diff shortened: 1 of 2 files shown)


thijstriemstra:
I'm not a fan of these event listeners inside the controls. My plugins manually update the currentime and duration displays and I manually disable these listeners, and with this PR yet another listener would be added that I'd have to disable.

gkatsev:
Having the controls listening to these events make them self contained instead of having something else be reaching in and updating them. In Video.js 6 with middleware, you can more easily make these controls output whatever you want instead.

PR #5361, opened 2018-08-03
Update notice about tracking pixel in readme

thijstriemstra:
Afaik the tracking pixel has been removed so this message can be removed from the readme.

Also bumped version nrs and added vjs-default-skin class to example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,25 +23,19 @@
 Thanks to the awesome folks over at [Fastly][fastly], there's a free, CDN hosted version of Video.js that anyone can use. Add these tags to your document's `<head>`:
 
 ```html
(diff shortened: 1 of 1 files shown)


gkatsev:
I wonder if it's better to leave it in but stress that it's only in older versions? Otherwise, yes, Video.js 7 doesn't include the tracking pixel.

OwenEdwards:
Video.js v6.x is still supported, so the information needs to stay, but it could be re-written to change the emphasis.

thijstriemstra:
This looks like a good additional description: https://blog.videojs.com/video-js-7-is-here/#Google-Analytics

gkatsev:
@thijstriemstra want to update the README with the snippet from the blogpost?

PR #5369, opened 2018-08-09
fix: call component dispose in resize manager to fix leak

BrandonOCasey:
Currently most event listeners that are being added are not removed because we don't trigger dispose in the ResizeManager. I think that it makes sense to just call the parents dispose here instead though.


gkatsev:
We also probably want to turn off touch activity for the resize manager.

PR #5371, opened 2018-08-10
fix(sourceset): ignore blob urls when updating source cache

gkatsev:
It's possible for us to get a blob url for sourceset, but when we do, we
shouldn't update the source caches with that information.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1291,6 +1291,12 @@ class Player extends Component {
       src = srcObj.src;
       type = srcObj.type;
     }
+
+    // if we are a blob url, don't update the source cache
+    if (/^blob:/.test(src)) {
+      return;
+    }
+
     // make sure all the caches are set to default values
     // to prevent null checking
     this.cache_.source = this.cache_.source || {};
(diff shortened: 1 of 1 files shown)


flipace:
hi - i'm not sure why exactly this change was necessary since i don't know the video.js internals, but this just burned us 🙈 

we use video.js dynamically like so: ```videojs($element, opts)``` 
in opts, we have a key `sources` which is an array of objects with type and src values.
the src values are blob urls.

what happened now for us - since this change: when triggering the video.js internal ```play()``` function it always led to videojs not finding a src/currentSrc and therefor not playing the video.

it worked when we directly invoked `play()` on the video element though.
i dug into this issue and found that videojs checks its internal source cache on this.cache_.src for a current src, but this has not been set in our case (seemingly because the return statement which was introduced in this pr prevent the update source cache function from ever reaching the point where it would set this...).

i commented out this part of the code and it started to work as expected again.

i'm not sure whether we might have to change how we initialize videojs and this is expected behavior, or this is actually unintended?

gkatsev:
I have to say that this is a pretty unexpected usecase. Would you be able to open a new issue with an example for us to take a look? We may be able to make fix to limit this change to just `sourceset` related stuff but would like to know more about your usecase before doing any more changes.
Also, sorry that this bit you.

flipace:
@gkatsev sure - i hope i'll come around to make a small repro case for you soon

PR #5381, opened 2018-08-14
DNM 7.2.1 backport test

gkatsev:


gkatsev:
Test is a failure. Changes to ResizeManager cause the test to fail for some reason. Trying to see a stack trace causes it to pass. I'll call it a day and merge everything else instead and maybe revisit in the future.

PR #5385, opened 2018-08-17
docs: Update JSDoc comments, so core API docs for the videojs function are accurate.

misteroneill:
**Reviewers:** `?w=1` is recommended. 😄 

In helping someone with an issue, I realized our API docs, especially for properties of the `videojs` function, were really bad. And this is extra-bad because that's the API surface that people actually use in 99% of cases.

There is some limitation here due to the fact that we're using `videojs` (and other things) both as a function and a namespace; so, while the docs are not perfect after these changes, we actually have the majority of the public API documented. While functions-as-namespaces [is supported in JSDoc 3.5](https://github.com/jsdoc3/jsdoc/issues/955#issuecomment-313829149), our template does not seem to be able to handle it.

### DIFF
--- a/src/js/extend.js
+++ b/src/js/extend.js
@@ -40,19 +40,17 @@ const _inherits = function(subClass, superClass) {
  * Function for subclassing using the same inheritance that
  * videojs uses internally
  *
- * @static
- * @const
+ * @function
+ * @param    {Object} superClass
+ *           The class to inherit from
  *
- * @param {Object} superClass
- *        The class to inherit from
+ * @param    {Object} [subClassMethods={}]
+ *           The class to inherit to
  *
- * @param {Object} [subClassMethods={}]
- *        The class to inherit to
- *
- * @return {Object}
- *         The new object with subClassMethods that inherited superClass.
+ * @return   {Object}
+ *           The new object with subClassMethods that inherited superClass.
  */
-const extendFn = function(superClass, subClassMethods = {}) {
+const extend = function(superClass, subClassMethods = {}) {
   let subClass = function() {
     superClass.apply(this, arguments);
   };
@@ -80,4 +78,4 @@ const extendFn = function(superClass, subClassMethods = {}) {
   return subClass;
 };
 
-export default extendFn;
+export default extend;
--- a/src/js/extend.js
+++ b/src/js/extend.js
@@ -37,18 +37,25 @@ const _inherits = function(subClass, superClass) {
 };
 
 /**
- * Function for subclassing using the same inheritance that
- * videojs uses internally
+ * Used to subclass an existing class by emulating ES subclassing using the
+ * `extends` keyword.
  *
  * @function
- * @param    {Object} superClass
+ * @example
+ * var MyComponent = videojs.extend(videojs.getComponent('Component'), {
+ *   myCustomMethod: function() {
+ *     // Do things in my method.
+ *   }
+ * });
+ *
+ * @param    {Function} superClass
  *           The class to inherit from
  *
- * @param    {Object} [subClassMethods={}]
- *           The class to inherit to
+ * @param    {Object}   [subClassMethods={}]
+ *           Methods of the new class
  *
- * @return   {Object}
- *           The new object with subClassMethods that inherited superClass.
+ * @return   {Function}
+ *           The new class with subClassMethods that inherited superClass.
  */
 const extend = function(superClass, subClassMethods = {}) {
   let subClass = function() {
--- a/src/js/tech/middleware.js
+++ b/src/js/tech/middleware.js
@@ -11,12 +11,35 @@ const middlewareInstances = {};
 export const TERMINATOR = {};
 
 /**
- * Define a middleware.
+ * A middleware object is a plain JavaScript object that has methods that
+ * match the {@link Tech} methods found in the lists of allowed
+ * {@link module:middleware.allowedGetters|getters},
+ * {@link module:middleware.allowedSetters|setters}, and
+ * {@link module:middleware.allowedMediators|mediators}.
+ *
+ * @typedef {Object} MiddlewareObject
+ */
+
+/**
+ * A middleware factory function that should return a
+ * {@link module:middleware~MiddlewareObject|MiddlewareObject}.
+ *
+ * This factory will be called for each player when needed, with the player
+ * passed in as an argument.
+ *
+ * @callback MiddlewareFactory
+ * @param {Player} player
+ *        A Video.js player.
+ */
+
+/**
+ * Define a middleware that the player should use by way of a factory function
+ * that returns a middleware object.
  *
  * @param  {string} type
- *         The mime type to match or `"*"` for all mime types.
+ *         The MIME type to match or `"*"` for all MIME types.
  *
- * @param  {Function} middleware
+ * @param  {MiddlewareFactory} middleware
  *         A middleware factory function that will be executed for
  *         matching types.
  */
@@ -29,7 +52,7 @@ export function use(type, middleware) {
  * Gets middlewares by type (or all middlewares).
  *
  * @param  {string} type
- *         The mime type to match or `"*"` for all mime types.
+ *         The MIME type to match or `"*"` for all MIME types.
  *
  * @return {Function[]|undefined}
  *         An array of middlewares or `undefined` if none exist.
@@ -43,7 +66,9 @@ export function getMiddleware(type) {
 }
 
 /**
- * Asynchronously sets a source using middleware.
+ * Asynchronously sets a source using middleware by recursing through any
+ * matching middlewares and calling `setSource` on each, passing along the
+ * previous returned value each time.
  *
  * @param  {Player} player
  *         A {@link Player} instance.
@@ -59,7 +84,7 @@ export function setSource(player, src, next) {
 }
 
 /**
- * Sets the tech using middeware.
+ * When the tech is set, passes the tech to each middleware's `setTech` method.
  *
  * @param {Object[]} middleware
  *        An array of middleware instances.
--- a/src/js/tech/middleware.js
+++ b/src/js/tech/middleware.js
@@ -1,3 +1,7 @@
+/**
+ * @file middleware.js
+ * @module middleware
+ */
 import { assign } from '../utils/obj.js';
 import toTitleCase from '../utils/to-title-case.js';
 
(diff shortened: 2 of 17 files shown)


src/js/extend.js
```diff
@@ -40,19 +40,17 @@ const _inherits = function(subClass, superClass) {
...
+ * @param    {Object} [subClassMethods={}]
+ *           The class to inherit to
  *
- * @return {Object}
- *         The new object with subClassMethods that inherited superClass.
+ * @return   {Object}
+ *           The new object with subClassMethods that inherited superClass.
  */
-const extendFn = function(superClass, subClassMethods = {}) {
+const extend = function(superClass, subClassMethods = {}) {
```
misteroneill: This mis-match in name between what was exposed on `videojs` and what was here in the module looked as if it was confusing. This should be a non-functional change.

src/js/tech/middleware.js
```diff
@@ -1,3 +1,7 @@
+/**
+ * @file middleware.js
+ * @module middleware
+ */
```
misteroneill: This was hardly documented at all. Tsk, tsk. 😛 Please review this file closely for accuracy.

src/js/utils/computed-style.js
```diff
@@ -8,21 +8,19 @@ import window from 'global/window';
...
+ *           The property name you want
  *
- * @see https://bugzilla.mozilla.org/show_bug.cgi?id=548397
- *
- * @static
- * @const
+ * @see      https://bugzilla.mozilla.org/show_bug.cgi?id=548397
  */
-export default function computedStyle(el, prop) {
+function computedStyle(el, prop) {
```
misteroneill: This happens in a few places. Having `export default function name` seemed to confuse things when we got to the rendered docs. Moving the `export default` made it better.

src/js/utils/fn.js
```diff
@@ -47,13 +51,14 @@ export const bind = function(context, fn, uid) {
  * Wraps the given function, `fn`, with a new function that only invokes `fn`
  * at most once per every `wait` milliseconds.
  *
- * @param  {Function} fn
- *         The function to be throttled.
+ * @function
```
misteroneill: In cases where functions are anonymously assigned to variables, `@function` is needed.

src/js/video.js
```diff
@@ -45,21 +46,83 @@ import { use as middlewareUse, TERMINATOR } from './tech/middleware.js';
...
+ * The `videojs()` function doubles as the main function for users to create a
+ * {@link Player} instance as well as the main library namespace.
+ *
+ * It can also be used as a getter for a pre-existing {@link Player} instance.
+ * However, we _strongly_ recommend using `videojs.getPlayer()` for this
+ * purpose because it avoids any potential for unintended initialization.
+ *
+ * Due to [limitations](https://github.com/jsdoc3/jsdoc/issues/955#issuecomment-313829149)
+ * of our JSDoc template, we cannot properly document this as both a function
+ * and a namespace, so its function signature is documented here.
```
misteroneill: As you see here, any function-as-namespace is manually documented in its description. This is a hack to work around an incompatibility in our template. However, I felt it was better to do this in a couple places than have most of the public API of `videojs` go undocumented.

src/js/video.js
```diff
@@ -45,21 +46,83 @@ import { use as middlewareUse, TERMINATOR } from './tech/middleware.js';
...
+ * @borrows Plugin.getPlugin as getPlugin
+ * @borrows Plugin.getPluginVersion as getPluginVersion
+ * @borrows Tech.getTech as getTech
+ * @borrows Tech.registerTech as registerTech
+ * @borrows TextTrack as TextTrack
+ * @borrows module:time-ranges.createTimeRanges as createTimeRange
+ * @borrows module:time-ranges.createTimeRanges as createTimeRanges
+ * @borrows module:url.isCrossOrigin as isCrossOrigin
+ * @borrows module:url.parseUrl as parseUrl
+ * @borrows VideoTrack as VideoTrack
```
misteroneill: These `@borrows` statements were peppered throughout this file incorrectly and, so, did nothing. Having these in a working location means that we can remove a ton of redundant documentation blocks below!

src/js/video.js
```diff
@@ -542,264 +485,43 @@ videojs.addLanguage = function(code, data) {
...
-
-/**
- * export the AudioTrack class so that source handlers can create
- * AudioTracks and then add them to the players AudioTrackList
- *
- * @borrows AudioTrack as videojs.AudioTrack
- */
 videojs.AudioTrack = AudioTrack;
-
-/**
```
gkatsev: we should verify that we didn't lose any info all of these removals.

src/js/extend.js
BrandonOCasey: 👍

src/js/utils/browser.js
```diff
@@ -55,11 +100,49 @@ export const ANDROID_VERSION = (function() {
...
+ * Whether or not this is Microsoft Edge.
+ *
+ * @static
+ * @const
+ * @type {Boolean}
+ */
 export const IS_EDGE = (/Edge/i).test(USER_AGENT);
+
+/**
+ * Whether or not this is Google Chrome.
```
BrandonOCasey: It could be good to mention that this will also include IOS chrome which may have different support then normal.

src/js/utils/format-time.js
```diff
@@ -64,6 +68,26 @@ export function resetFormatTime() {
   implementation = defaultImplementation;
 }
 
-export default function(seconds, guide = seconds) {
+/**
+ * Delegates to either the default time formatting function or a custom
+ * function supplied via `setFormatTime`.
+ *
+ * Formats seconds as a time string (H:MM:SS or M:SS). Supplying a
```
BrandonOCasey: Could be good to have an example here.

src/js/extend.js
BrandonOCasey: We should probably think about just getting rid of this in some future version too.

src/js/tech/middleware.js
```diff
@@ -6,11 +10,30 @@ const middlewareInstances = {};
 
 export const TERMINATOR = {};
 
+/**
+ * Define a middleware.
+ *
+ * @param  {string} type
+ *         The mime type to match or `"*"` for all mime types.
+ *
+ * @param  {Function} middleware
```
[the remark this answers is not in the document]
BrandonOCasey: I think the description would be more clear as: Define a middleware that a player should use for a certain source type.

src/js/video.js
```diff
@@ -386,90 +418,32 @@ Object.defineProperty(videojs.middleware, 'TERMINATOR', {
...
- * compatibility with 4.x, but deprecated. Use `videojs.browser.TOUCH_ENABLED`
- * instead going forward.
+ * Use {@link module:browser.TOUCH_ENABLED|browser.TOUCH_ENABLED} instead; only
+ * included for backward-compatibility with 4.x.
  *
- * @deprecated since version 5.0
+ * @deprecated Since version 5.0, use {@link module:browser.TOUCH_ENABLED|browser.TOUCH_ENABLED instead.
  * @type {boolean}
  */
 videojs.TOUCH_ENABLED = browser.TOUCH_ENABLED;
```
BrandonOCasey: Seems like this should have been removed awhile ago, heh. I think we need some kind of deprecation list.

src/js/video.js
```diff
@@ -386,90 +418,32 @@ Object.defineProperty(videojs.middleware, 'TERMINATOR', {
...
 
-/**
- * Subclass an existing class
```
BrandonOCasey: I think this description should probably be added into the extend file, as the extend file has a very technical description.

src/js/video.js
```diff
@@ -386,90 +418,32 @@ Object.defineProperty(videojs.middleware, 'TERMINATOR', {
...
- * Mimics ES6 subclassing with the `extend` keyword
- *
- * @borrows extend:extendFn as videojs.extend
- */
-videojs.extend = extendFn;
-
-/**
- * Merge two options objects recursively
- * Performs a deep merge like lodash.merge but **only merges plain objects**
- * (not arrays, elements, anything else)
```
BrandonOCasey: I think that this description is much more clear than the one in the file for mergeOptions.

src/js/video.js
```diff
@@ -386,90 +418,32 @@ Object.defineProperty(videojs.middleware, 'TERMINATOR', {
...
+videojs.extend = extend;
 videojs.mergeOptions = mergeOptions;
-
-/**
- * Change the context (this) of a function
- *
- * > NOTE: as of v5.0 we require an ES5 shim, so you should use the native
- * `function() {}.bind(newContext);` instead of this.
- *
- * @borrows fn:bind as videojs.bind
```
BrandonOCasey: Another deprecation... hmmm

src/js/video.js
```diff
@@ -386,90 +418,32 @@ Object.defineProperty(videojs.middleware, 'TERMINATOR', {
...
 videojs.registerPlugin = Plugin.registerPlugin;
-
-/**
- * Deregister a Video.js plugin.
- *
- * @borrows plugin:deregisterPlugin as videojs.deregisterPlugin
- * @method deregisterPlugin
- *
- * @param  {string} name
- *         The name of the plugin to be deregistered. Must be a string and
```
BrandonOCasey: Thes description for the name parameter has much more information here than it does in the file.

src/js/video.js
BrandonOCasey: ➕

src/js/video.js
BrandonOCasey: Do we want a borrows for the functions contained on `videojs.dom` and `videojs.url`?

src/js/extend.js
misteroneill: Don't we still need it for allowing people to extend built-in classes? Maybe once `class` is supported universally we won't need it.

src/js/video.js
misteroneill: I tried, but couldn't get it working.

src/js/video.js
misteroneill: Makes sense.

src/js/utils/computed-style.js
misteroneill: That sounds like a good idea, but I think it's better to have useful documentation.

src/js/tech/middleware.js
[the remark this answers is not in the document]
misteroneill: @gkatsev That seems to only show up as `function` in the rendered docs. Not sure where that syntax comes from. It's not in the JSDoc documentation for `@param`.

src/js/extend.js
gkatsev: Oh, sorry, I meant that `extend` was `extendFn` probably because of IE8 issues. Yeah, makes sense that we still need it.

src/js/video.js
misteroneill: I think for 8, we should remove a bunch of deprecated stuff.

misteroneill: But I think a lot of things were deprecate late in 5, so nobody was comfortable removing them in 6. Then 7 was all about browser support changes and nothing else.

src/js/video.js
misteroneill: Don't think it's actually deprecated?

src/js/video.js
misteroneill: The only things I know we do lose are the top-level DOM function aliases that are deprecated. I may be able to restore them without manually documenting each, but it seemed like a better thing to have the non-deprecated API documented.

src/js/video.js
misteroneill: At least, it doesn't have any official notice of it.

src/js/video.js
BrandonOCasey: ok

PR #5390, opened 2018-08-20
tests: CI resize manager test issues

brandonocasey:


src/js/resize-manager.js
```diff
@@ -59,21 +59,22 @@ class ResizeManager extends Component {
     this.resizeObserver_ = null;
     this.debouncedHandler_ = debounce(() => {
       this.resizeHandler();
-    }, 100, false, player);
+    }, 100, false, this);
```
BrandonOCasey: seemed weird to put this into the players context, but then call a function in our context.  This probably isn't related to the bug though.

src/js/resize-manager.js
```diff
@@ -59,21 +59,22 @@ class ResizeManager extends Component {
...
 
     if (RESIZE_OBSERVER_AVAILABLE) {
       this.resizeObserver_ = new this.ResizeObserver(this.debouncedHandler_);
       this.resizeObserver_.observe(player.el());
 
     } else {
       this.loadListener_ = () => {
-        if (this.el_.contentWindow) {
-          Events.on(this.el_.contentWindow, 'resize', this.debouncedHandler_);
+        if (!this.el_ || this.el_.contentWindow) {
```
BrandonOCasey: Check `this.el_` too. Again probably not related to the bug.

src/js/resize-manager.js
```diff
@@ -59,21 +59,22 @@ class ResizeManager extends Component {
...
+          return;
         }
-        this.off('load', this.loadListener_);
+
+        Events.on(this.el_.contentWindow, 'resize', this.debouncedHandler_);
       };
 
-      this.on('load', this.loadListener_);
+      this.one('load', this.loadListener_);
```
BrandonOCasey: This may be a fix, I doubt we are seeing two load events, but if we were. This would have caused two debounce handlers to be added for window `resize`.

src/js/resize-manager.js
```diff
@@ -114,15 +125,10 @@ class ResizeManager extends Component {
...
 
-    if (this.debouncedHandler_) {
-      this.debouncedHandler_.cancel();
-    }
-
     this.ResizeObserver = null;
     this.resizeObserver = null;
     this.debouncedHandler_ = null;
     this.loadListener_ = null;
-    super.dispose();
```
BrandonOCasey: We don't really need to `super.dispose` since we turned off touch event reporting.

PR #5407, opened 2018-08-31
Fixed typos in cs translation

Akxe:
Lang file edit only


lang/cs.json
```diff
@@ -67,19 +67,19 @@
...
   "Reset": "Obnovit",
   "restore all settings to the default values": "Vrátit nastavení do výchozího stavu",
   "Done": "Hotovo",
-  "Caption Settings Dialog": "Okno s nastevením !caption!",
-  "Beginning of dialog window. Escape will cancel and close the window.": "Začátek dialogového okna. Klávesa Escape okno zavře.",
+  "Caption Settings Dialog": "Okno s nastavením titulků",
+  "Beginning of dialog window. Escape will cancel and close the window.": "Začátek dialogového okna. Klávesa Esc okno zavře.",
   "End of dialog window.": "Konec dialogového okna.",
-  "{1} is loading.": "{1} se načítá."
+  "{1} is loading.": "{1} se načítá.",
```
misteroneill: Trailing `,` is invalid JSON.

Akxe: Why is that? It would be better for diffs, no?

misteroneill: It's just how JSON is. With the trailing comma, `JSON.parse()` will throw an error. It comes from the bad old days when certain JS engines would choke on trailing commas.

PR #5416, opened 2018-09-06
fix: make sure all attrs are updated before applying to tag

gkatsev:
Fixes videojs/video.js#5309

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -591,17 +591,27 @@ class Player extends Component {
 
     // set tabindex to -1 to remove the video element from the focus order
     tag.setAttribute('tabindex', '-1');
+    attrs.tabindex = '-1';
+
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -591,18 +591,28 @@ class Player extends Component {
...
+      attrs.role = 'application';
     }
 
     // Remove width/height attrs from tag so CSS can make it 100% width/height
     tag.removeAttribute('width');
     tag.removeAttribute('height');
 
+    if ('width' in attrs) {
+      delete attrs.width;
+    }
```
misteroneill: Generally, it should be safe to `delete` without condition, but after looking it is possible that if `width` or `height` were an own, non-enumerable property, that it couldn't be `delete`d, so 👍

PR #5418, opened 2018-09-07
feat: createLogger for easier logging in individual modules

gkatsev:
This is the most naive implementation of transforming what we have into providing a create logger functionality.
Basically, right now, `log.js` is changed into a `create-logger.js` which is a function that given a name returns a log function. `log.js` now uses `create-logger` to `createLogger('VIDEOJS')` which is the default logger.
`videojs.createLogger` is exposed for users of the library.

Currently, logByType is also exposed directly on the log object but that's probably not wanted.
Other tweaks we can consider: a shared/central history. This will allow us to make it be in one place and we could also provide some search functionality via the "main" logger. For example, searching for all things with the name of `VHS` or something.

Additionally, we may want to expose `createLogger` on the `log` itself rather than on `videojs` and have the logs inherit each time one is created:
```js
var log = videojs.log.createLogger('foo');
log('hello');
// > VIDEOJS: foo: hello
```


misteroneill:
> Currently, logByType is also exposed directly on the log object but that's probably not wanted.

Agree. I think it should be a candidate for deprecation (maybe not in this PR).

> shared/central history. This will allow us to make it be in one place and we could also provide some search functionality via the "main" logger. For example, searching for all things with the name of VHS or something.

Love this idea.

The only concern I have with logging history is the potential for memory leaks. For example, one might log a player object, which would immediately create a potential memory leak without them knowing.

A history refactor could be involve solutions to that problem.

- It could keep a history with only primitive values.
- It could remove references to players in the history when they are disposed (could be a performance concern, iterating over the whole history, though).

> Additionally, we may want to expose createLogger on the log itself rather than on videojs and have the logs inherit each time one is created.

I like the idea of inheritance. The prefixes generate a kind of path allowing using to more easily filter the console.

## Player-Scoped Logger
Another useful thing would be player-specific loggers:

```js
player.log('hello');

// > VIDEOJS: ${playerId}: hello
```

These could then be disposed with the player making some of the memory concerns with history a bit cleaner.

gkatsev:
We should just consider turning off history by default. Maybe even as part of this PR, I'm not sure that anyone is really using it and yeah, could be a memory issue.

Having a player-scoped logger is a cool idea.

gkatsev:
Looks like the only `logByType` that's used outside is in a test and that can be updated. Then I can just remove it from being exposed publicly.

gkatsev:
`logByType` has been made module private and the tests were updated for it as well.

Next is shared history, chainable loggers, and player log.

gkatsev:
Adding a `player.log` was surprisingly easy.
Going to work on refactoring to a shared history but also probably making a Log class instead.

gkatsev:
Not sure why travis isn't showing up but the two builds have succeeded there: [#9475](https://travis-ci.org/videojs/video.js/builds/426810470), [#9746](https://travis-ci.org/videojs/video.js/builds/426810476)

docs/guides/debugging.md
```diff
@@ -44,6 +49,23 @@ Similar to the `console`, any number of mixed-type values can be passed to `vide
 videojs.log('this is a string', {butThis: 'is an object'});
 ```
 
+### Creating new Loggers
+
+Sometimes, you want to make a new module or plugin and log messages with a label. Kind of how all these logs are prepended with `VIDEOJS:`. You can do that via the `createLog` method. It takes a name and gives you back a log object like `videojs.log`. Here's an example:
```
misteroneill: `createLog` should be `createLogger`

src/js/utils/create-logger.js
```diff
@@ -0,0 +1,232 @@
...
+
+  if (!fn && type === 'debug') {
+    // Certain browsers don't have support for console.debug. For those, we
+    // should default to the closest comparable log.
+    fn = window.console.info || window.console.log;
+  }
+
+  // Bail out if there's no console or if this type is not allowed by the
+  // current logging level.
+  if (!fn || !lvl || !lvlRegExp.test(type)) {
```
misteroneill: I don't recall which versions it is, but there are some IEs that don't allow using `apply`/`call` on `console` methods. Probably not the case in IE11, but might be worth checking.

gkatsev: This code wasn't changed from what was around previously.

misteroneill: Ah, then, it's probably safe. I could be showing my age again. 😄

Issue #5360, opened 2018-08-03
Better way to initialise plugins


Akxe:
## Description
VideoJS should have better way to initialising plugin than `player.myPlugin(...options)`, this is a problem if you encounter TypeScript. 

### Actual
`player.myPlugin(...options)`, this is fine for JS but TS must cast player to any.

### New proposal
`player.usePlugin(name: string, ...options: any[]): any` find plugin by name and apply options.
`player.usePlugin(plugin: videojs.Plugin)` use plugin
`player.usePlugin<T, K>(name: string, (...options: T[]) => K, ...options: T[]): K` Create plugin with name and use it immediately.

1. Useful, as it will mitigate use on `(player as any).myPlugin(...options)`.
2. I personally never used this, but it might be useful to some.
3. Most convenient to single player scenarios as well as for TS since we have options in one function and therefor it can be strongly typed.

gkatsev:
Why do you have to cast it to `any`? Is it because the typings for the player don't include plugin methods?

Akxe:
Indeed player don't have it typed unless you extend the player by yourself... Also as I think about it, if some public variables, maybe `player.acces(name, property)` but that's an other thing...

misteroneill:
Hmm, interesting. It would be nice if Video.js were easy to use with TypeScript.

I think the `usePlugin` idea has more potential for this specific use-case. I think something like this:

```js
player.usePlugin('foo', options);
```

is a nicer interface than this:

```js
player.access('foo')(options);
```

Though, I could see the latter generic property getter being a useful thing in some cases. Similar to `getattr` in Python, for example.

Akxe:
That was exactly what it was meant to be, `player.acces(name, attr)`. I 
would love that, just keep in mind to edit type definitions too. And many 
thanks for your time on making it accessible to typescript 

---------- Původní zpráva ----------
Od: misteroneill 
Datum: 4. 8. 2018 v 18:57:39
Předmět: Re: [videojs/video.js] Better way to initialise plugins (#5360)


Hmm, interesting. It would be nice if Video.js were easy to use with 
TypeScript.

I think the usePlugin idea has more potential for this specific use-case. I 
think something like this:


<span>player</span>.<span>usePlugin</span>(<span><span>'</span>foo<span>'</span></span>, options);


is a nicer interface than this:


<span>player</span>.<span>access</span>(<span><span>'</span>foo<span>'</span></span>)(options);


Though, I could see the latter generic property getter being a useful thing 
[...]

PR #5426, opened 2018-09-10
player.usePlugin - shortcut to register & init

Akxe:
## Description
videojs/video.js#5360 Added new function `usePlugin` to Player prototype. It will register the plugin as well as initialise it with given parameters.

## Motivation behing
When `registerPlugin` and its initialisation is merged into one function, many will be satisfied as usually only one player exists per page. Also this will help TS as it would make it possible to strongly check plugin's options.

## Requirements Checklist
- [x] Feature implemented
    - [ ] Test is working
  - [ ] Docs/guides updated
  - [ ] Updated TypeScript type declaration file

Does it need example? it could potentially replace current plugin example set up.

PS: Testing by hand works, but unit fails. May I get some help?

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3892,13 +3892,13 @@ TECH_EVENTS_RETRIGGER.forEach(function(event) {
  */
 
 /**
- * Register and immediately use plugin on Player.
+ * Apply plugin to player.
  *
  * @method Player#usePlugin
  * @param   {string} name
- *          The name of the plugin to be registered. Must be a string and
- *          must not match an existing plugin or a method on the `Player`
- *          prototype.
+ *          The name under witch the plugin will be saved. Under this same
+ *          name `player.usingPlugin(name)`. Must be a string and must not
+ *          match an existing plugin or a method on the `Player` prototype.
  *
  * @param   {Function} plugin
  *          A sub-class of `Plugin` or a function for basic plugins.
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -8,7 +8,7 @@ import {version} from '../../package.json';
 import document from 'global/document';
 import window from 'global/window';
 import tsml from 'tsml';
-import evented from './mixins/evented';
+import evented from './mixins/evented';i
 import * as Events from './utils/events.js';
 import * as Dom from './utils/dom.js';
 import * as Fn from './utils/fn.js';
@@ -3892,7 +3892,7 @@ TECH_EVENTS_RETRIGGER.forEach(function(event) {
  */
 
 /**
- * Register and imediatly use plugin on Player.
+ * Register and immediately use plugin on Player.
  *
  * @method Player#usePlugin
  * @param   {string} name
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -8,7 +8,7 @@ import {version} from '../../package.json';
 import document from 'global/document';
 import window from 'global/window';
 import tsml from 'tsml';
-import evented from './mixins/evented';i
+import evented from './mixins/evented';
 import * as Events from './utils/events.js';
 import * as Dom from './utils/dom.js';
 import * as Fn from './utils/fn.js';
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3896,14 +3896,14 @@ TECH_EVENTS_RETRIGGER.forEach(function(event) {
  *
  * @method Player#usePlugin
  * @param   {string} name
- *          The name under witch the plugin will be saved. Under this same
+ *          The name under which the plugin will be saved. Under this same
  *          name `player.usingPlugin(name)`. Must be a string and must not
  *          match an existing plugin or a method on the `Player` prototype.
  *
  * @param   {Function} plugin
  *          A sub-class of `Plugin` or a function for basic plugins.
  *
- * @param   {...any} options
+ * @param   {...any} args
  *          Options that the plugin should be called with.
  */
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3892,6 +3892,22 @@ TECH_EVENTS_RETRIGGER.forEach(function(event) {
  */
 
 /**
+ * Register and imediatly use plugin on Player.
+ *
+ * @method Player#usePlugin
+ * @param   {string} name
+ *          The name of the plugin to be registered. Must be a string and
+ *          must not match an existing plugin or a method on the `Player`
+ *          prototype.
+ *
+ * @param   {Function} plugin
+ *          A sub-class of `Plugin` or a function for basic plugins.
+ *
+ * @param   {...any} options
+ *          Options that the plugin should be called with.
+ */
+
+/**
  * Reports whether or not a player has a plugin available.
  *
  * This does not report whether or not the plugin has ever been initialized
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -456,7 +456,7 @@ Plugin.registerPlugin(BASE_PLUGIN_NAME, Plugin);
  *
  * @ignore
  */
-Player.prototype.usePlugin = function(name, plugin, ...options) {
+Player.prototype.usePlugin = function(name, plugin, ...args) {
   let wrappedPlugin;
 
   if (name !== BASE_PLUGIN_NAME) {
@@ -465,7 +465,7 @@ Player.prototype.usePlugin = function(name, plugin, ...options) {
     } else {
       wrappedPlugin = createPluginFactory(name, plugin);
     }
-    wrappedPlugin(...options);
+    wrappedPlugin(...args);
   } else {
     throw new Error('Please help with nice error...');
   }
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -456,6 +456,16 @@ Plugin.registerPlugin(BASE_PLUGIN_NAME, Plugin);
  *
  * @ignore
  */
+Player.prototype.usePlugin = function(name, plugin, ...options) {
+  Plugin.registerPlugin(name, plugin);
+  this[name](...options);
+};
+
+/**
+ * Documented in player.js
+ *
+ * @ignore
+ */
 Player.prototype.usingPlugin = function(name) {
   return !!this[PLUGIN_CACHE_KEY] && this[PLUGIN_CACHE_KEY][name] === true;
 };
--- a/src/js/plugin.js
+++ b/src/js/plugin.js
@@ -457,8 +457,18 @@ Plugin.registerPlugin(BASE_PLUGIN_NAME, Plugin);
  * @ignore
  */
 Player.prototype.usePlugin = function(name, plugin, ...options) {
-  Plugin.registerPlugin(name, plugin);
-  this[name](...options);
+  let wrappedPlugin;
+
+  if (name !== BASE_PLUGIN_NAME) {
+    if (Plugin.isBasic(plugin)) {
+      wrappedPlugin = createBasicPlugin(name, plugin);
+    } else {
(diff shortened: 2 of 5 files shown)


src/js/player.js
```diff
@@ -3891,6 +3891,22 @@ TECH_EVENTS_RETRIGGER.forEach(function(event) {
  * @type {EventTarget~Event}
  */
 
+/**
+ * Register and imediatly use plugin on Player.
```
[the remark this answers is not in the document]
Akxe: Typo fixed

src/js/plugin.js
```diff
@@ -451,6 +451,16 @@ Plugin.BASE_PLUGIN_NAME = BASE_PLUGIN_NAME;
 
 Plugin.registerPlugin(BASE_PLUGIN_NAME, Plugin);
 
+/**
+ * Documented in player.js
+ *
+ * @ignore
+ */
+Player.prototype.usePlugin = function(name, plugin, ...options) {
+  Plugin.registerPlugin(name, plugin);
```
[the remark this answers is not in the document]
gkatsev: I am against this as well. This method should *only* be a shorthand for calling the plugin manually.

When you register a plugin, it registers it with videojs, so, a plugin will be available to be initialized on all players and not just the player you called `usePlugin` on. For example:
```js
player.usePlugin('foo', function() {});

typeof player2.foo;
// function
```
That would be unexpected to me since I called `usePlugin` on `player`, why does `player2` have the option of initializing `foo`?

src/js/video.js
```diff
@@ -431,6 +431,26 @@ videojs.mergeOptions = mergeOptions;
...
+ *         prototype.
+ *
+ * @param  {Function} plugin
+ *         A sub-class of `Plugin` or a function for basic plugins.
+ *
+ * @return {Function}
+ *         For advanced plugins, a factory function for that plugin. For
+ *         basic plugins, a wrapper function that initializes the plugin.
+ */
+videojs.usePlugin = Plugin.usePlugin;
```
gkatsev: This doesn't make sense since plugins are initialized on players and `videojs` isn't an instance of a player.

src/js/plugin.js
```diff
@@ -451,6 +451,16 @@ Plugin.BASE_PLUGIN_NAME = BASE_PLUGIN_NAME;
...
+  this[name](...options);
```
gkatsev: We'd want to do a null-check around `this[name]` being a method.

Akxe: In this scenario, is it possible to be null? Wouldn't the `registerPlugin` throw instead?

src/js/plugin.js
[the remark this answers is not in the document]
Akxe: Well, but having to register and then call the plugin, break the type inherence, that is one of main reason behind it... Is there any way to do this without registering plugin "globally"?

But I do agree, that registering on one player and accessing on the other would lead to confusion.

src/js/player.js
```diff
@@ -3891,6 +3891,22 @@ TECH_EVENTS_RETRIGGER.forEach(function(event) {
  * @type {EventTarget~Event}
  */
 
+/**
+ * Apply plugin to player.
+ *
+ * @method Player#usePlugin
+ * @param   {string} name
+ *          The name under witch the plugin will be saved. Under this same
+ *          name `player.usingPlugin(name)`. Must be a string and must not
```
Akxe: `Must be a string`, is this true? It cannot be a symbol nor number? Not that I care for numbers, but I would be for symbols in future. Also, storing plugin in `WeakMap`, if available, might be preferable.

src/js/plugin.js
```diff
@@ -451,6 +451,26 @@ Plugin.BASE_PLUGIN_NAME = BASE_PLUGIN_NAME;
...
+  if (name !== BASE_PLUGIN_NAME) {
+    if (Plugin.isBasic(plugin)) {
+      wrappedPlugin = createBasicPlugin(name, plugin);
+    } else {
+      wrappedPlugin = createPluginFactory(name, plugin);
+    }
+    wrappedPlugin(...options);
+  } else {
+    throw new Error('Please help with nice error...');
+  }
```
misteroneill: Given the necessary removal of the `plugin` argument, the body of this function can be simplified to:

```js
const plugin = getPlugin(name);

if (plugin) {
  plugin.apply(this, args);
} else {
  throw new Error(`Could not use plugin "${name}" as it does not exist.`);
}
```

src/js/player.js
Akxe: You can't replicate symbol, therefor they would be a safe option for internal plugins. Other than that, no... It is just a primitive for this kind of things..

Akxe: You can't replicate symbol, therefor they would be a safe option for internal plugins. Other than that, no... It is just a primitive for this kind of things.

Akxe: You can't replicate symbol, therefor they would be a safe option for internal plugins. Other than that, no... It is just a primitive for this kind of things.

src/js/plugin.js
Akxe: Requiring registration of plugin in separate function removes the most, if not all, of the reason to have it. The point is to have type anotations for plugin options and that is possible only, if they are in the same context.

src/js/plugin.js
```diff
@@ -451,6 +451,26 @@ Plugin.BASE_PLUGIN_NAME = BASE_PLUGIN_NAME;
 
 Plugin.registerPlugin(BASE_PLUGIN_NAME, Plugin);
 
+/**
+ * Documented in player.js
+ *
+ * @ignore
+ */
+Player.prototype.usePlugin = function(name, plugin, ...options) {
```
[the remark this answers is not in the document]
Akxe: Renaming options to args is very reasonable. For removal of plugin see below.

Akxe:
[Typescript playground demo](http://www.typescriptlang.org/play/#src=interface%20Player%20%7B%0D%0A%09usePlugin%3CT%20extends%20any%5B%5D%2C%20K%20extends%20T%2C%20R%3E(name%3A%20string%2C%20plugin%3A%20(this%3A%20Player%2C%20...args%3A%20T)%20%3D%3E%20R%2C%20...args%3A%20K)%3A%20R%0D%0A%09registerPlugin%3CT%3E(name%3A%20string%2C%20plugin%3A%20(this%3A%20Player%2C%20...args%3A%20any%5B%5D)%20%3D%3E%20T)%3A%20T%0D%0A%09__usePlugin(name%3A%20string%2C%20...args%3A%20any%5B%5D)%3A%20any%0D%0A%7D%0D%0A%0D%0Adeclare%20const%20player%3A%20Player%3B%0D%0A%0D%0A%2F%2F%20%20Try%20catch%20is%20only%20to%20save%20implementation%20of%20player%0D%0Atry%20%7B%0D%0A%09let%20a%3A%20string%3B%0D%0A%09a%20%3D%20player.usePlugin('myPlugin'%2C%20(name%3A%20string%2C%20count%3A%20number%2C%20options%3A%20%7B%20complexity%3A%20string%20%7D)%20%3D%3E%20%7B%0D%0A%09%09%2F%2F%20%20plugin%20code...%0D%0A%09%09return%20'hello'%3B%0D%0A%09%7D%2C%20'name'%2C%20123%2C%20%7B%20complexity%3A%20'high'%20%7D)%3B%0D%0A%0D%0A%09%2F%2F%09Demostrate%20wrong%20argument%0D%0A%09player.usePlugin('myPlugin'%2C%20(name%3A%20string%2C%20count%3A%20number%2C%20options%3A%20%7B%20complexity%3A%20string%20%7D)%20%3D%3E%20%7B%0D%0A%09%09%2F%2F%20%20plugin%20code...%0D%0A%09%7D%2C%20'name'%2C%20'123'%2C%20%7B%20complexity%3A%20'high'%20%7D)%3B%0D%0A%0D%0A%09%2F%2F%09Demostrate%20missing%20argument%0D%0A%09player.usePlugin('myPlugin'%2C%20(name%3A%20string%2C%20count%3A%20number%2C%20options%3A%20%7B%20complexity%3A%20string%20%7D)%20%3D%3E%20%7B%0D%0A%09%09%2F%2F%20%20plugin%20code...%0D%0A%09%7D%2C%20'name'%2C%20%7B%20complexity%3A%20'high'%20%7D)%3B%0D%0A%0D%0A%09%2F%2F%09Demonstrate%20return%20type%20of%20plugin%0D%0A%09a%20%3D%20player.usePlugin('myPlugin'%2C%20(name%3A%20string%2C%20count%3A%20number%2C%20options%3A%20%7B%20complexity%3A%20string%20%7D)%20%3D%3E%20%7B%0D%0A%09%09%2F%2F%20%20plugin%20code...%0D%0A%09%09return%20123%3B%0D%0A%09%7D%2C%20'name'%2C%20123%2C%20%7B%20complexity%3A%20'high'%20%7D)%3B%0D%0A%0D%0A%09%2F%2F%0D%0A%09%2F%2F%09In%20two%20separate%20functions%0D%0A%09%2F%2F%0D%0A%0D%0A%09player.registerPlugin('myPlugin'%2C%20(name%3A%20string%2C%20count%3A%20number%2C%20options%3A%20%7B%20complexity%3A%20string%20%7D)%20%3D%3E%20%7B%0D%0A%09%09%2F%2F%20%20plugin%20code...%0D%0A%09%09return%20123%3B%0D%0A%09%7D)%3B%0D%0A%0D%0A%09%2F%2F%09It%20will%20simply%20not%20know%20how%20to%20help.%0D%0A%09player.__usePlugin('myPlugin'%2C%20'name'%2C%20'123')%3B%0D%0A%09player.__usePlugin('myPlugin')%3B%0D%0A%09a%20%3D%20player.__usePlugin('myPlugin'%2C%20'name'%2C%20'123'%2C%20%7B%20complexity%3A%20'string'%20%7D)%3B%0D%0A%0D%0A%7D%20catch%20(e)%20%7B%20%7D%0D%0A%0D%0A), of how would both solution look like.

gkatsev:
I don't think the complexity tradeoff here is worth it. Allowing the `usePlugin` method plugin and autoregister may help out typescript users but it makes it more confusing and harder to use for anyone who isn't using typescript.

Akxe:
What would be confusing? User has player, and want to use it to use a plugin he wrote. If he want more than one copy of that he might want to use `registerPlugin` and make use of the prototype, but other than that, it is more confusing the other way round... When I first registered my plugin, I was confused on why it isn't there, upon a bit of reading I found out it is on prototype of Player somehow. Don't forget that you are used to this way and not the other way round.

src/js/plugin.js
```diff
@@ -451,6 +451,26 @@ Plugin.BASE_PLUGIN_NAME = BASE_PLUGIN_NAME;
 
 Plugin.registerPlugin(BASE_PLUGIN_NAME, Plugin);
 
+/**
+ * Documented in player.js
+ *
+ * @ignore
+ */
+Player.prototype.usePlugin = function(name, plugin, ...args) {
```
gkatsev: while this no longer registers it still keeps an inconsistent API.
You can't use this method on plugins that were already registered elsewhere and the way that plugins used via this method work is different than other plugins.

gkatsev:
Thanks for your patience, it's been a while.
I agree that the way that plugins currently work isn't necessarily ideal and that there are scenarios that are hard to work with. However, this change makes an inconsistent API that we'll have to live with and ignores the bigger picture.
Currently, plugins are registered globally and then called on the particular player. I wouldn't want to introduce an API that handles things different without thinking about the implications of that. Without thinking of the usecases and functionality of plugins that work otherwise. I don't want us to merge something quickly and then down the line as we refactor how plugins work to support more usecases have to fight with an old API.

PR #5446, opened 2018-09-24
fix(text-tracks): cuechange handler not triggering correctly

### LINKED ISSUE #5308: hidden captions textTrack not triggering "cuechange" events
## Description
When setting a 'captions' text track to 'hidden' mode and adding a cuechange event, the events are not being fired.

## Steps to reproduce
1. set the captions track to hidden
2. add cuechange event listener
3. send some cue events

### Expected
cuechange event triggered with the data

### Actual
event is not being triggered, but works when setting the track mode to 'showing' 

### Error output
no error output

#### videojs
7.11
#### browsers
Chrome Latest
#### OSes
OSX

gkatsev:
We were only triggering cuechange events if a metadata track started out as not disabled or only when setting the mode to 'showing'

Fixes #5308

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -230,7 +230,7 @@ class TextTrack extends Track {
             this.tech_.ready(() => {
               this.tech_.on('timeupdate', timeupdateHandler);
             }, true);
-          } else if (mode === 'disabled') {
+          } else {
             this.tech_.off('timeupdate', timeupdateHandler);
           }
           /**
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -226,11 +226,12 @@ class TextTrack extends Track {
             return;
           }
           mode = newMode;
-          if (mode === 'showing') {
-
+          if (mode !== 'disabled') {
             this.tech_.ready(() => {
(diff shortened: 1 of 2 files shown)


src/js/tracks/text-track.js
```diff
@@ -226,11 +226,12 @@ class TextTrack extends Track {
             return;
           }
           mode = newMode;
-          if (mode === 'showing') {
-
+          if (mode !== 'disabled') {
             this.tech_.ready(() => {
               this.tech_.on('timeupdate', timeupdateHandler);
             }, true);
+          } else if (mode === 'disabled') {
```
misteroneill: Wouldn't `} else {` make sense here?

gkatsev: Potentially, let me see what browsers do for non-valid values.

gkatsev: Chrome, Firefox, and Chrome will not let you set the value to anything that isn't hidden, showing, and disabled.

gkatsev: Looks like we already ignore non-valid values, so, just else here is fine.

gkatsev:
Can be verified here https://deploy-preview-5446--videojs-docs.netlify.com/

apadhye:
QA LGTM

PR #5478, opened 2018-10-04
feat(fill): make vjs-fill a player mode

gkatsev:
Like fluid mode, you can enable it with the class or by calling the fill
method. Calling fill() will turn off fluid mode and calling fluid() will
turn off fill mode.

Fluid mode takes precedence over fill mode currently.


gkatsev:
Good call

Issue #5474, opened 2018-10-03
Prevent buttons from being focused on mouse click


gjanblaszczyk:
## Description
Every time I clicked on the control bar buttons. I saw that the clicked button was focused.
However, this behavior is fine but sometimes I want to have a possibility to disable this.
The outline should be visible when the user uses the keyboard keys to navigate through control bar buttons.
Can we add the new videojs option to enable/disable the control buttons from being focused on mouse click?
I want to have similar behavior like youtube player has now.
![videojs-play-outline](https://user-images.githubusercontent.com/5346546/46407383-1117a800-c70f-11e8-9ee5-e83ea8be6124.png)

## Steps to reproduce

1. Use mouse device to click on the control bar button for example: play or fullscreen.
2. The clicked button will be focused and you will see the outline around it.

### Expected
The button is focused but I don't see the visual outline. Unless I used keyboard 'tab' key to navigate through control bar controls.

### Actual
The button is focused and the outline is shown after the mouse click.

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
lates
#### browsers
any
#### OSes
any

gkatsev:
Yeah, this is something we've been thinking about a lot. Specifically, there hasn't been a good way to hide it for mouse users but keep it for keyboard users. Starting with v6, we opted that it's better to have the outlines for mouse users to make keyboard and screen reader user's experience better.

It's possible that we should just add something like the following, then when browsers implement it it'll go away or we can direct users to the focus-visible polyfill.
```css
:focus:not(:focus-visible) { outline: none; }
```

gjanblaszczyk:
Thanks for the information and this focus-visible polyfill is a really good idea. 
btw @gkatsev Do you know any other solution for this problem?

gkatsev:
There isn't a good way, unfortunately. One way I think we may be able to do it is by storing whether the component was interacted with with the mouse or not. Then, in the focus event, we can preventDefault if it was the mouse. But that's less than ideal.

OwenEdwards:
@gjanblaszczyk you can see a fully implemented version of the `:focus-visible` polyfill being used at:

https://github.com/OwenEdwards/video.js/commit/47f25a2d29be18fe7b87024239c8851117f5e1f2

It's a very small change, and `:focus-visible` is on track to be a W3C recommendation.

[...]

PR #5479, opened 2018-10-04
Fix: prevent buttons from being focused on mouse click

gjanblaszczyk:
## Description

A proposed fix for #5474
It's based on preventDefault. 
It is not perfect but at least works on all browsers.

## Requirements Checklist
  - [ ] Docs/guides updated


OwenEdwards:
I *strongly* oppose accepting this PR into the core of video.js. As I described in https://github.com/videojs/video.js/issues/5474#issuecomment-427478027, `:focus-visible` is a completely workable solution, it's available now, and it's most likely that that will be the *preferred* solution in the future if that gets adopted by W3C.

Integrating this hack into the core of video.js (so that it affects all users of video.js, not just @gjanblaszczyk ) would cause accessibility barriers for some existing users (such as low-vision users who use both mouse and keyboard in a blended way), and might break compatibility with people who *do* chose to use the `:focus-visible` polyfill, which would be exactly the opposite direction from where we should be going.

PR #5481, opened 2018-10-05
docs(layout): document fluid and fill mode

gkatsev:
~~Depends on https://github.com/videojs/video.js/pull/5478~~

### DIFF
--- a/docs/guides/layout.md
+++ b/docs/guides/layout.md
@@ -1,0 +1,101 @@
+# Layout
+
+Video.js generally lays out the player to the dimensions that are set as attributes or via CSS, like other DOM elements. However, we provide a few ways to make the player be more fluid.
+
+## Fluid mode
+
+Video.js has a fluid mode that keeps the player sized to a particular aspect ratio.
+By default, fluid mode will use the intrinsic size of the video once loaded but you can change it with classes or with the `aspectRatio` option.
+Enabling fluid mode will disable fill mode. If both are enabled, fluid mode takes precedence, currently.
+
+You can enable fluid one in a few ways:
+
+- Add `vjs-fluid`, `vjs-16-9`, or `vjs-4-3` as a class to the player element.
+- pass `fluid` option to the player.
+- call `player.fluid(true)`.
+- pass `aspectRatio` option to the player.
+- call `player.aspectRatio('16:9)`.
+
+### Classes
+
+There are three classes associated with fluid mode, `vjs-fluid`, `vjs-16-9`, and `vjs-4-3`.
+`vjs-fluid` turns on the general fluid mode which will wait for the video to load to calculate the aspect ratio of the video.
+Alternatively, because 16:9 and 4:3 aspect ratios are so common, we provided them as classes by default for you to use if you know that your videos are 16:9 or 4:3.
+
+### fluid option
+
+You can pass in the `fluid` option to the player or call `player.fluid(true)`. This will enable the generic fluid mode.
+
+```js
+var player = videojs('vid1', {
+  fluid: true
+});
+```
+
+```js
+var player = videojs('vid2');
+
+player.fluid(true);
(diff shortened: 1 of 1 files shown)


docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
+# Layout
+
+Video.js generally lays out the player to the dimensions that are set as attributes or via CSS, like other DOM elements. However, we provide a few ways to make the player be more fluid.
+
+## Fluid mode
+
+Video.js has a fluid mode that keeps the player sized to a particular aspect ratio.
+By default, fluid mode will use the intrinsic size of the video once loaded but you can change it with classes or with the `aspectRatio` option.
+Enabling fluid mode will disable fill mode. If both are enabled, fluid mode takes precedence, currently.
```
BrandonOCasey: I think you can remove currently from the end of this sentence

docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
...
+
+You can enable fluid one in a few ways:
```
BrandonOCasey: remove "one" here

docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
...
+
+- Add `vjs-fluid`, `vjs-16-9`, or `vjs-4-3` as a class to the player element.
+- pass `fluid` option to the player.
+- call `player.fluid(true)`.
+- pass `aspectRatio` option to the player.
+- call `player.aspectRatio('16:9)`.
```
BrandonOCasey: needs another single quote

docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Classes
+
+There are three classes associated with fluid mode, `vjs-fluid`, `vjs-16-9`, and `vjs-4-3`.
+`vjs-fluid` turns on the general fluid mode which will wait for the video to load to calculate the aspect ratio of the video.
+Alternatively, because 16:9 and 4:3 aspect ratios are so common, we provided them as classes by default for you to use if you know that your videos are 16:9 or 4:3.
+
+### fluid option
+
+You can pass in the `fluid` option to the player or call `player.fluid(true)`. This will enable the generic fluid mode.
```
BrandonOCasey: I think it would make sense to move rename this section `enabling fluid mode` and move the list from the section above down here.

docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
...
+
+You can disable fluid mode by remove the associated classes or by calling passing in `false` to the method.
+
+```js
+player.fluid(false);
+```
+
+### aspectRatio
+
+You can specify an aspect ratio for us to use if you don't want to use the intrinsic values from the video element or if you have a specific ratio in mind. It works as either a method call or an option to the player.
```
BrandonOCasey: I would think that this would also go into `Enabling Fluid Mode`

docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
...
+
+// make a square video
+player.aspectRatio('1:1');
+```
+
+## Fill mode
+
+Fill mode will make the player fit and fill out its container. This is often useful if you have a responsive website and already have a container for Video.js that resizes properly to your design. It can be set either via a class or an option.
+
+If fill is enabled, it'll turn off fluid mode.
```
BrandonOCasey: Could be good to add the disclaimer that fluid mode takes precedence here as well.

docs/guides/layout.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Class
+
+There's just one class for this one: `vjs-fill`. When available, Video.js will enter fill mode.
+
+## fill option
```
BrandonOCasey: Again it may make more sense to have a section called `Enabling fill mode`

PR #5483, opened 2018-10-05
Prevent focus on click v2

gjanblaszczyk:
## Description

A proposed fix for #5474
It based on :focus-visible selector.
For now, it works only on Chrome. You have to enable "Experimental Web Platform features" under the Chrome settings (chrome://flags/) to be an able test this solution. 
The PR doesn't break other browsers. 

## Requirements Checklist
  - [ ] Docs/guides updated


OwenEdwards:
"The PR doesn't break other browsers" - is that true? Doesn't this PR disable focus outlines if `:focus-visible` is not present, so it will disable the outline for **all** browsers **except** Chrome with _"Experimental Web Platform features" under the Chrome settings (chrome://flags/)"_ enabled?

We can't do that and break accessibility across all browsers except Chrome with an Experimental flag enabled. I thought this PR was going to integrate the `":focus-visible"` polyfill.

gjanblaszczyk:
Hi Owen, Thanks for the comment. 
For your question: "The PR doesn't break other browsers" - is that true?  
Yes, it shouldn't break any browsers at least according to some comments from focus-visible polyfill project.
See here:
https://github.com/WICG/focus-visible/blame/master/README.md#L134
I used the same CSS rule:
`/*
  Remove default focus styles for mouse users ONLY if
  :focus-visible is supported on this platform.
*/
button:focus:not(:focus-visible) {
  …
}`
I think also @gkatsev suggesting using: 
`:focus:not(:focus-visible) { outline: none; }`
https://github.com/videojs/video.js/issues/5474#issuecomment-426797628

asd29696:
Owen Edwards <notifications@github.com> 於 2018年10月8日週一 6:44 寫道：

gkatsev:
What about a general one in the main scss file instead for each component?
```css
.video-js *:focus:not(:focus-visible) {
  outline: none;
}
```

@OwenEdwards older browsers that don't support that pseudo-selector will ignore the entire block.

OwenEdwards:
@gjanblaszczyk and @gkatsev: "_older browsers that don't support that pseudo-selector will ignore the entire block._" - ahh, thanks for that clarification, I misunderstood that.

Then I agree with @gkatsev's suggestion above, making it generic for all of video.js. Additionally, I'd suggest that we add the `:focus-visible` polyfill into one of the HTML examples in `docs` or `sandbox`, and refer to it in some of the documentation - that will help steer people in that direction if they're concerned about this, rather than them submitting just a univeral `outline: none;` PR again (like #5027).

gkatsev:
I hadn't thought of that. Firefox's default focus ring is just so bad, heh.

gjanblaszczyk:
Thanks for comments,
@gkatsev I added `focus-visible` selector to the main file. 
Do we want to add focus-visible polyfill to the sandbox demo? If Yes I suggest adding the new focus-visible.html demo to the sandbox directory. any objections?

gkatsev:
Adding a `focus-visible.html.example` to the sandbox sounds reasonable to me.

gkatsev:
You can go to this URL directly `chrome://flags/#enable-experimental-web-platform-features`

gkatsev:
@gjanblaszczyk hey, did you get a chance to add the example page using the polyfill?

sandbox/focus-visible.html.example
```diff
@@ -0,0 +1,52 @@
...
+  <!-- Set the location of the flash SWF -->
+  <script>
+    videojs.options.flash.swf = '../node_modules/videojs-flash/node_modules/videojs-swf/dist/video-js.swf';
+  </script>
+  <style>
+    /*
+      This will hide the focus indicator if the element receives focus via the mouse,
+      but it will still show up on keyboard focus.
+    */
+    .js-focus-visible :focus:not(.focus-visible) {
```
gkatsev: is this piece necessary since we have a similar one in Video.js now?

sandbox/focus-visible.html.example
```diff
@@ -0,0 +1,52 @@
...
+      outline: none;
+    }
+  </style>
+</head>
+<body>
+  <div style="background-color:#eee; border: 1px solid #777; padding: 10px; margin-bottom: 20px; font-size: .8em; line-height: 1.5em; font-family: Verdana, sans-serif;">
+    <p>You can use /sandbox/ for writing and testing your own code. Nothing in /sandbox/ will get checked into the repo, except files that end in .example (so don't edit or add those files). To get started make a copy of index.html.example and rename it to index.html.</p>
+    <pre>cp sandbox/focus-visible.html.example sandbox/focus-visible.html</pre>
```
gkatsev: we really need to update this text, `npm start` will automatically copy these files over from .example if they don't already exist.

sandbox/focus-visible.html.example
```diff
@@ -0,0 +1,52 @@
...
+    <pre>npm run start</pre>
+    <pre>open http://localhost:9999/sandbox/focus-visible.html</pre>
+  </div>
+
+  <video id="vid1" class="video-js vjs-default-skin" controls preload="auto" width="640" height="264"
+      poster="http://vjs.zencdn.net/v/oceans.png"
+      data-setup='{"controlBar": {"volumePanel": {"inline": false}}}'>
+    <!-- <source src="http://wms.shared.streamshow.it/canale8/canale8/playlist.m3u8" type="application/x-mpegURL"> -->
```
gkatsev: can this line be removed?

src/css/video-js.scss
```diff
@@ -52,3 +52,7 @@
   border: none;
   visibility: hidden;
 }
+
+.video-js *:focus:not(.focus-visible) {
```
gkatsev: does the polyfill require the usage of the `.focus-visible` class? We probably want to have both selectors in that case.

gjanblaszczyk: Yes, according to the polyfill doc. This class is needed.

gkatsev: Ah, cool, didn't realize. Yeah, let's just do both, then eventually we can remove this one. Thanks!

gkatsev:
Actually, it's weird but I'm not seeing the outline when using the keyboard which is kind of weird. I'd expect to still see it then.

gjanblaszczyk:
Hmm, I guess https://deploy-preview-5483--videojs-docs.netlify.com/node_modules/focus-visible/dist/focus-visible.min.js polyfill was not deployed.

gkatsev:
Aah, makes sense. Totally missed that. I guess either swap it out for
```html
<script src="https://unpkg.com/focus-visible"></script>
```
or edit the build file https://github.com/videojs/video.js/blob/master/build/generate-example.js

The updating the url is probably easiest.

gkatsev:
Loading in the polyfill manually made it work beautifully.

gjanblaszczyk:
Great but maybe I should change the focus-visible polyfill script src to https://unpkg.com/focus-visible on the focus-visible example page and remove focus-visible polyfill from the node package.json file. I guess better to have less dependency. What do you think?

gkatsev:
Sounds good.

src/css/video-js.scss
OwenEdwards: If you're going to do this (which makes sense because the focus-visible polyfill can only apply a `.focus-visible` **class**, not a `:focus-visible` **pseudo-class**, I think you need to include the syntax I had in my example:

````css
.js-focus-visible .video-js *:focus:not(.focus-visible) {
  outline: none;
}
````

Without the `.js-focus-visible` it's going to remove the focus outline if the developer **doesn't** include the polyfill, which is exactly what we want to avoid.

OwenEdwards: ^^^ That comment should have been a review

gkatsev:
Oh, yeah, I'll add it in when I merge if @gjanblaszczyk doesn't get to it first.

gjanblaszczyk:
np @gkatsev

PR #5489, opened 2018-10-09
docs(fixup): fixup docs

gkatsev:


COLLABORATOR_GUIDE.md
```diff
@@ -35,18 +41,17 @@ If issues or PRs are very short and don't contain much information, ask for more
...
+| confirmed                | Issue and PR | Issue: marks as reproducible. PR: marks as ready to be merged              |
+| 5.x                      | PR           | Marks as a change to the 5.x branch only                                   |
+| bug                      | Issue        | Marks as a confirmed bug                                                   |
+| good first issue         | Issue        | Marks as a good bug or enhancement for first time contributors to Video.js |
+| first-timers-only        | Issue        | Marks as a good bug or enhancement to be done by a newcomer to open source |
+| minor, patch, major      | PR           | Marks PR with the expected semver classification of the change             |
+| needs: LGTM              | PR           | Marks PR to be reviewed by a collaborator                                  |
+| needs: more info         | Issue        | Marks as needing more information from the issue reporter                  |
+| needs: reduced test case | Issue        | Marks as needing a reduced test case from the issue reporter               |
```
misteroneill: Thank you. 😆

docs/translations-needed.md
```diff
@@ -13,6 +13,7 @@ This default value is hardcoded as a default to the localize method in the SeekB
 ## Status of translations
 
 <!-- START langtable -->
+
```
mister-ben: The check-translations task would remove these lines when it updates the table.

gkatsev: yeah, we should probably fix that up since remark adds these lines and we'll get thrashing, but yeah, I should just remove these lines.

PR #5471, opened 2018-10-02
feat: Add breakpoints option to support toggling classes based on player width.

misteroneill:
Closes videojs/video.js#4371 

## Description
This adds a `breakpoints` option. By default, this option is `false` meaning this is an opt-in feature.

When passing `true`, it will use a default set of breakpoints. Or custom breakpoints can be passed if users do not like our breakpoints (or previously-existing style decisions).

## Specific Changes proposed
* Add `breakpoints` option.
* Adds some new (currently unused) classes: `vjs-layout-medium`, `vjs-layout-large`, `vjs-layout-x-large`, and `vjs-layout-huge`.
* Add `updateCurrentBreakpoint` and `currentBreakpoint` methods to the player.
* Update `css/components/_adaptive.scss`
* In cleaning up sandbox a bit, I noticed several appearances of `vjs-default-skin` which was no longer used anywhere. Removed it everywhere.
* Add `sandbox/responsive.html.example`

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [x] If necessary, more likely in a feature request than a bug fix
  - [x] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [x] Unit Tests updated or fixed
  - [x] Docs/guides updated
- [ ] Reviewed by Two Core Contributors

PR #5496, opened 2018-10-11
feat: Add responsive option, which enables breakpoints support.

misteroneill:
Follow-up for #5471 

This makes the `breakpoints` option and `setBreakpoints()` method clearer and introduces the `responsive` option and `setResponsive()` method, which will turn on the breakpoints.

The return value of `currentBreakpoint()` was simplified to only ever return a string (empty if none).

Also, added convenience methods: `isResponsive()`, `getBreakpointClass()`, and `currentBreakpointClass()`.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/docs/guides/layout.md
+++ b/docs/guides/layout.md
@@ -2,10 +2,12 @@
 
 Video.js generally lays out the player to the dimensions that are set as attributes or via CSS, like other DOM elements. However, we provide a few ways to make the player be more fluid.
 
-## Fluid mode
+## Fluid Mode
 
 Video.js has a fluid mode that keeps the player sized to a particular aspect ratio.
+
 By default, fluid mode will use the intrinsic size of the video once loaded but you can change it with classes or with the `aspectRatio` option.
+
 Enabling fluid mode will disable fill mode. If both are enabled, fluid mode takes precedence.
 
 You can enable fluid in a few ways:
@@ -19,10 +21,12 @@ You can enable fluid in a few ways:
 ### Classes
 
 There are three classes associated with fluid mode, `vjs-fluid`, `vjs-16-9`, and `vjs-4-3`.
+
 `vjs-fluid` turns on the general fluid mode which will wait for the video to load to calculate the aspect ratio of the video.
+
 Alternatively, because 16:9 and 4:3 aspect ratios are so common, we provided them as classes by default for you to use if you know that your videos are 16:9 or 4:3.
 
-### Enabling fluid mode
+### Enabling Fluid Mode
 
 You can pass in the `fluid` option to the player or call `player.fluid(true)`. This will enable the generic fluid mode.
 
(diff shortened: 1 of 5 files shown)


src/js/player.js
```diff
@@ -3749,59 +3750,153 @@ class Player extends Component {
...
-    this.breakpoints_ = assign({}, DEFAULT_BREAKPOINTS);
+  /**
+   * Set a flag indicating whether or not this player should adjust its UI
+   * based on its dimensions.
+   *
+   * @param  {boolean} value
+   *         Should be `true` if the player should adjust its UI based on its
+   *         dimensions; otherwise, should be `false`.
+   */
+  setResponsive(value) {
```
gkatsev: why `setResponsive`/`isRespnsive` pair? Most of our other methods are getters/setters:
```js
player.responsive();
// false
player.responsive(true);
// true and responsive mode is now enabled
```

misteroneill: Yeah, that's true. I thought we wanted to move away from that, though. I guess it makes sense to be more internally consistent for now.

docs/guides/layout.md
```diff
@@ -92,10 +96,90 @@ var player = videojs('vid2');
...
 
 You can disable fill mode by removing the associated class or by passing `false` in to the method.
 
 ```js
 player.fill(false);
 ```
+
+## Responsive Mode
+
+Responsive mode will make the player's UI customize itself based on the size of the player. This is useful if you have embeds of varying sizes - or if you want a fluid/fill player to adjust its UI based on its size.
```
BrandonOCasey: Does this take precedence over fluid or fill mode? Could be good to clarify that here.

src/js/player.js
```diff
@@ -3749,59 +3750,134 @@ class Player extends Component {
...
+   *         dimensions; otherwise, should be `false`.
+   *
+   * @return {boolean}
+   *         Will be `true` if this player should adjust its UI based on its
+   *         dimensions; otherwise, will be `false`.
    */
-  getBreakpoints() {
-    return this.breakpoints_ && assign(this.breakpoints_);
+  responsive(value) {
+    if (value !== undefined) {
```
BrandonOCasey: It would be good to make the getter usage right here with `if (value === undefined)`. That way we get rid of some whitespace.

docs/guides/layout.md
misteroneill: The next line says it is independent of fluid/fill. But it may be useful to say that it can be used in conjunction with them.

src/js/player.js
misteroneill: In that case, we also repeat the `return` statement, but you're right - that's closer to our usual pattern.

docs/guides/layout.md
misteroneill: Added a bit more to the next paragraph.

PR #5498, opened 2018-10-11
chore(babel): Upgrade to Babel 7

gkatsev:
Use `npx babel-upgrade --write --install` to initiate the upgrade. Then update grunt and rollup. May been to fiddle a bit with the env preset a bit as well.


gkatsev:
I verified this works on Chrome, Edge, IE11 Win 10, Firefox, and Safari.
However, there's still some test-related issues that need to be ironed out.

gkatsev:
Probably worth checking if IE11 on Win7 still works, but I think it will since if env preset isn't configured with targets, it's as if we added the es2015, es2016, es2017 presets.

Also, with these changes, builds are slightly faster (rollup runs around 1 minutes faster and karma around 2 minutes) and the main minified gzipped file is 1KB smaller!

misteroneill:
IE11 works fine! 👍 `npm start` does not... 😢

BrandonOCasey:
We should probably use `browserslist` for this like we do in [videojs-generate-rollup-config](https://github.com/videojs/videojs-generate-rollup-config/blob/master/index.js#L148)

gkatsev:
It may be nice eventually use browserlist. I think for now, I'd rather keeping it unconfigured as that's closest to what we had before.

gkatsev:
`npm start` was broken because contrib-watch package-lock field accidentally got set to also be jshint, so, it was never installed. 017510f fixed it.

PR #5499, opened 2018-10-12
feat: Clicking live display seeks to live edge, improved live UI

BrandonOCasey:
## Changes
* Add a "fake-live" example to sandbox, which should emulate live well enough for code reviews
* Added a circle to the live display that indicates if you are at the live edge or not. It is red when you are on the live edge, and gray when you are not.
* Clicking the live display brings you to the live edge, but it takes a segment length amount of time before we can do that.

### DIFF
--- a/src/css/components/_live.scss
+++ b/src/css/components/_live.scss
@@ -12,3 +12,18 @@
   width: auto;
   text-align: left;
 }
+
+.vjs-live-circle {
+  border-radius: 50%;
+  width: 1.25em;
+  height: 1.25em;
+  margin-top: auto;
+  margin-bottom: auto;
+  margin-right: 0.3em;
+  margin-left: 0.3em;
+  background: radial-gradient(#808080, #666666);
+}
+
+.video-js .vjs-live-control.vjs-control.vjs-at-live-edge .vjs-live-circle {
+  background: radial-gradient(#ff0000, #990000);
+}
--- a/src/js/control-bar/live-display.js
+++ b/src/js/control-bar/live-display.js
@@ -1,17 +1,16 @@
 /**
  * @file live-display.js
  */
+import ClickableComponent from '../clickable-component';
 import Component from '../component';
 import * as Dom from '../utils/dom.js';
-
-// TODO - Future make it click to snap to live
 
 /**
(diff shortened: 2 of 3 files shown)


BrandonOCasey:
https://deploy-preview-5499--videojs-docs.netlify.com/test-example/fake-live.html

src/css/components/_live.scss
```diff
@@ -3,12 +3,27 @@
 .video-js .vjs-live-control {
   @include display-flex(flex-start);
   @include flex(auto);
-  font-size: 1em;
-  line-height: 3em;
+  font-size: 1.5em;
```
BrandonOCasey: The text was very small in relation to the icons on the control bar, I bumped it up a tiny bit.

gkatsev: Unfortunately, this means that the LIVE display is now inconsistent with the size of other text in the control bar like the various time displays. I would leave this as is and we should open an issue to look into increase the text size.

gkatsev: Created one from the comment #5500

BrandonOCasey: ok

src/js/control-bar/live-display.js
```diff
@@ -1,17 +1,16 @@
...
 
-// TODO - Future make it click to snap to live
-
 /**
  * Displays the live indicator when duration is Infinity.
  *
  * @extends Component
  */
-class LiveDisplay extends Component {
+class LiveDisplay extends ClickableComponent {
```
gkatsev: This probably should be a Button, if possible.

gkatsev:
Is a lot of LiveDisplay stays around in https://github.com/videojs/video.js/pull/5511/?

Also, sorry for realizing this so late in the game but I wonder if these updated to the LiveDisplay should be a new component? I think the changes, particularly to the UI of it, may be big enough that it'll cause issues if anyone relied on the LiveDisplay as it exists currently.

BrandonOCasey:
I am going to close this PR and make future changes in the other live pr (going to rename that pr to liveui). I will work on moving all of the changes to another component something like `SeekToLive`

Issue #888, opened 2013-12-11
It is forbidden for me to drag the progress forward or backward  that i want , Using Flash to play MP4 video with VideoJs, why?


joostshao:
> It is forbidden for me to drag the progress forward or backward  that i want , 
> Using Flash to play MP4 video with VideoJs, why?
> 
> What should i do to implement the Flash allow me to choose what startTime to begin?
> 
> Just Like Youtube or Vimeo ...

PR #5511, opened 2018-10-17
feat: Better UI for live playback via an option

BrandonOCasey:
## Changes
All of the following is behind a new option called `liveui` that defaults to `false`:

### LiveDisplay
* Shown when `vjs-live` is set on the player but hidden when `vjs-live` and `vjs-liveui` are both on the player

### SeekToLive
* A new component that will show when `vjs-live` and `vjs-liveui` classes are on the player
* Used to show when the player is at the live edge, and to indicate that the player is live
* Can be clicked to seek to the live edge when behind.

### LiveTracker
* Handles tracking of the "live edge" and fires events when the live edge and currentTime are not in sync.
* Handles tracking of `seekable end changes` so that we can keep up to date with what the live current time should be

### Seek/Progress Bar
* Show during live playback
* Make all time tooltips show negative numbers indicating the time from live.
* Seek within the live window
* Keep updating all the time when live streaming, since if you pause you will fall behind the live stream.
* Hide and show various components via live css


sandbox/fake-live.html.example
```diff
@@ -23,11 +23,12 @@
     // fake a livestream for easy testing
     var vid = document.getElementById('vid1');
     var player = videojs(vid, {muted: true, loop: true});
-    var segmentLength = 6;
+    var segmentLength = 2;
 
-    window.currentLiveTimeRange = videojs.createTimeRanges(0, 0);
+    window.currentLiveTimeRange = videojs.createTimeRanges(0, segmentLength);
```
BrandonOCasey: This fixes an issue where it would not show up as live for the first few seconds.

sandbox/fake-live.html.example
```diff
@@ -23,11 +23,12 @@
...
 
     player.ready(function() {
+      player.currentTime(segmentLength);
```
BrandonOCasey: This makes it so that there are a few seconds before the "live" point as in a normal live stream

src/css/components/_live.scss
```diff
@@ -1,19 +1,11 @@
-// We are assuming there is no progress bar and using the live display
-// to fill in the middle space. Live+DVR will need to adjust this.
 .video-js .vjs-live-control {
-  @include display-flex(flex-start);
-  @include flex(auto);
-  font-size: 1em;
-  line-height: 3em;
+  cursor: pointer;
```
BrandonOCasey: Makes the live control have a pointer, and makes it take up a similar amount of space to a button. Previously it took up all of the progress control space but the progress control will now show.

src/css/components/_progress.scss
```diff
@@ -13,10 +13,6 @@
   cursor: default;
 }
 
-.vjs-live .vjs-progress-control {
```
BrandonOCasey: we show this now

src/css/components/_progress.scss
```diff
@@ -173,3 +169,7 @@
   color: #fff;
   @include background-color-with-alpha(#000, 0.8);
 }
+
+.vjs-live .vjs-load-progress {
+  display: none;
```
BrandonOCasey: I am not sure if we want to hide this or fix it, basically this can make it so that the progress bar goes past the player.

src/js/control-bar/live-display.js
```diff
@@ -24,7 +24,7 @@ class LiveDisplay extends Button {
   constructor(player, options) {
     super(player, options);
 
-    this.updateShowing();
+    player.ready(() => this.updateShowing());
```
BrandonOCasey: We may not know if the player is live until it is ready.

src/css/components/_live.scss
mister-ben: Should it have a pointer when at the live point, if clicking will do nothing?

mister-ben:
The button width might have to be flexible to account for longer localised strings. Catalan 'EN DIRECTE' we currently have in the translations.

src/css/components/_live.scss
BrandonOCasey: hmm probably not good call.

gkatsev:
Is there a reason this isn't implemented as a middleware?
I think this should be enabled by default but with the ability to turn it off but there's still time to discuss.

misteroneill:
@gkatsev We discussed that a bit in our grooming call, but @BrandonOCasey had some uncertainty. We can discuss further on Monday when you're back. 👍

BrandonOCasey:
Internally we originally though that this could be implemented as middleware, but I found several issues with that in the initial implementation. Here are the issues

1. We would have to change currentTime internally to `0` at the live edge and negative seconds before that. This caused a bunch of issues across the board, since the UI did not know how to deal with currentTime not changing, or going negative. It would also be inconsistent with the video element API.
2. In order to keep the progress bar within the `seekable` range of the video element we would have to change the `duration` from `Infinity`. Which would hide that it was a livestream. furthermore we would have to update the duration every time `seekable` changed which for short segment live streams could be every 2s.
3. Even if we were to do the two things above, there would still be UI specific code for live streams, and it would probably be very similar to what we have now. For instance we would have to add support for negative time tool tips and the play progress would still have to update during pause when live streaming (so that the seek bar can decrease)


After talking about it we think that the issues do warrant this being a UI only change. As having live streams now have a finite duration and negative current times would be too far from what a video element does.

gkatsev:
I haven't done in depth testing but a there's documentation missing.
Also, testing with a stream with a window of 5 segments of 10 seconds each, I'd have expected the progress bar to be let us seek back the full 20 seconds, at least after playing back a segment. Is this something we should talk with the playback team about?

BrandonOCasey:
Yeah I think we would need to talk to them about that. We can only seek back as far as seekable start, as far as I know.

src/js/live-tracker.js
```diff
@@ -0,0 +1,201 @@
...
+      this.pastSeekEnd_ = 0;
+      this.lastSeekEnd_ = newSeekEnd;
+      this.trigger('seek-end-change');
+    }
+
+    this.pastSeekEnd_ = this.pastSeekEnd() + 0.03;
+
+    if (this.isBehind() !== this.behindLiveEdge()) {
+      this.behindLiveEdge_ = this.isBehind();
+      this.trigger('live-edge-change');
```
misteroneill: This event naming seems contradictory to what we have in most places (e.g. no punctuation, all lowercase).

BrandonOCasey: yeah, I didn't think about that there is also `seek-end-change`. If we want to change it to `liveedgechange` and `seekendchange` we can do that.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -213,12 +233,42 @@ class SeekBar extends Slider {
...
+
+      // Compensate for precision differences so that currentTime is not less
+      // than seekable start
+      if (newTime <= seekStart) {
+        newTime = seekStart + 0.1;
+      }
+    }
+
+    if (newTime === Infinity) {
+      return;
```
gkatsev: how can newTime be infinity?

src/js/player.js
```diff
@@ -4074,6 +4080,7 @@ Player.prototype.options_ = {
     'textTrackDisplay',
     'loadingSpinner',
     'bigPlayButton',
+    'liveTracker',
```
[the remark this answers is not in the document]
BrandonOCasey: yeah your probably right.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -66,8 +67,10 @@ class SeekBar extends Slider {
       }, UPDATE_REFRESH_INTERVAL);
     });
 
-    this.on(this.player_, ['ended', 'pause', 'waiting'], () => {
-      this.clearInterval(this.updateInterval);
+    this.on(this.player_, ['ended', 'pause', 'waiting'], (e) => {
+      if (!this.player_.liveTracker.isLive()) {
```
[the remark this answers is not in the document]
BrandonOCasey: Because the progress bar will continue to decrease when paused/waiting. I guess we could check if the event is ended here

src/js/live-tracker.js
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * handle a durationchange event on the player
+   * and start/stop tracking accordingly.
+   */
+  handleDurationchange() {
+    if (this.player_.duration() === Infinity) {
+      this.start();
+    } else {
+      this.stop();
```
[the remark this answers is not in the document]
BrandonOCasey: Since all the controls update on duration change it should, going to make sure liveTracker is inserted above controlBar though, that way it listens to durationchange before components it depends on.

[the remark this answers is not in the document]
BrandonOCasey: actually it already is before controlBar.

sandbox/live.html.example
```diff
@@ -0,0 +1,29 @@
...
+<body>
+  <div style="background-color:#eee; border: 1px solid #777; padding: 10px; margin-bottom: 20px; font-size: .8em; line-height: 1.5em; font-family: Verdana, sans-serif;">
+    <p>You can use /sandbox/ for writing and testing your own code. Nothing in /sandbox/ will get checked into the repo, except files that end in .example (so don't edit or add those files). To get started make a copy of index.html.example and rename it to index.html.</p>
+    <pre>cp sandbox/index.html.example sandbox/index.html</pre>
+    <pre>npm run start</pre>
+    <pre>open http://localhost:9999/sandbox/index.html</pre>
+  </div>
+
+  <video-js id="vid1" controls preload="auto" width="640"  height="264">
+    <source src="https://akamai-axtest.akamaized.net/routes/lapd-v1-acceptance/www_c4/Manifest.m3u8" type="application/x-mpegURL">
```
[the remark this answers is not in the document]
BrandonOCasey: I can't really think of a compelling reason, going to remove it.

src/js/control-bar/progress-control/load-progress-bar.js
```diff
@@ -67,8 +67,10 @@ class LoadProgressBar extends Component {
       return ((percent >= 1 ? 1 : percent) * 100) + '%';
     };
 
-    // update the width of the progress bar
-    this.el_.style.width = percentify(bufferedEnd, duration);
+    // update the width of the progress bar if not live
+    if (!this.player_.liveTracker.isLive()) {
```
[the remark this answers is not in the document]
BrandonOCasey: duration is infinity, and the loadProgress bar goes past the player.

src/css/components/_live.scss
BrandonOCasey: yes

src/css/components/_progress.scss
```diff
@@ -17,6 +17,10 @@
   display: none;
 }
 
+.vjs-live.vjs-liveui .vjs-progress-control {
```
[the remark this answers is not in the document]
BrandonOCasey: we can

src/js/live-tracker.js
```diff
@@ -0,0 +1,201 @@
...
+  constructor(player, options, ready) {
+    super(player, options, ready);
+
+    this.on(this.player_, 'durationchange', this.handleDurationchange);
+  }
+
+  isBehind() {
+    const liveCurrentTime = this.liveCurrentTime();
+    const currentTime = this.player_.currentTime();
+    const segmentLength = this.segmentLength_ || 12;
```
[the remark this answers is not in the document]
BrandonOCasey: segmentLength is probably a bad name. It the number of seconds between seekableEnd changes. So if seekable end changes every 2s it will be 2

src/js/live-tracker.js
```diff
@@ -0,0 +1,201 @@
...
+      if (this.lastSeekEnd_) {
+        this.segmentLength_ = newSeekEnd - this.lastSeekEnd_;
+      }
+
+      this.pastSeekEnd_ = 0;
+      this.lastSeekEnd_ = newSeekEnd;
+      this.trigger('seek-end-change');
+    }
+
+    this.pastSeekEnd_ = this.pastSeekEnd() + 0.03;
```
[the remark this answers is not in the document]
BrandonOCasey: To give some more headroom, This interval will not actually run every 30ms, sometimes it will run faster or slower.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -173,7 +180,20 @@ class SeekBar extends Slider {
    *         The percentage of media played so far (0 to 1).
    */
   getPercent() {
-    const percent = this.getCurrentTime_() / this.player_.duration();
+    const currentTime = this.getCurrentTime_();
+    let percent;
+    const liveTracker = this.player_.liveTracker;
+
+    if (liveTracker.isLive()) {
+      percent = (this.player_.currentTime() - liveTracker.seekStart()) / liveTracker.liveTimeWindow();
```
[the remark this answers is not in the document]
BrandonOCasey: I'm not sure it will have a use outside of the seek-bar though.

src/js/player.js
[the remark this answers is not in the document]
BrandonOCasey: Actually I think we always want `liveTracker` as a component. It would be hard to add it as a child, and have its `durationchange` handler run before the rest of the `controlBar`

src/css/components/_progress.scss
```diff
@@ -17,6 +17,10 @@
   display: none;
 }
 
+.vjs-liveui .vjs-progress-control {
+  @include display-flex(center);
```
ldayananda: shouldn't this be `flex-start`?

src/js/control-bar/control-bar.js
```diff
@@ -58,6 +59,7 @@ ControlBar.prototype.options_ = {
     'durationDisplay',
     'progressControl',
     'liveDisplay',
+    'seekToLive',
```
ldayananda: As a general question, do we expect the live components to work independently of each other? i.e the seekToLive could be used without the liveDisplay

src/js/control-bar/progress-control/load-progress-bar.js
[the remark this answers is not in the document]
ldayananda: should we use the last `seekable.end` instead of `duration` for live?

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -173,7 +182,20 @@ class SeekBar extends Slider {
    *         The percentage of media played so far (0 to 1).
    */
   getPercent() {
-    const percent = this.getCurrentTime_() / this.player_.duration();
+    const currentTime = this.getCurrentTime_();
+    let percent;
+    const liveTracker = this.player_.liveTracker;
+
+    if (liveTracker.isLive()) {
+      percent = (this.player_.currentTime() - liveTracker.seekableStart()) / liveTracker.liveWindow();
```
ldayananda: could you just use `getCurrentTime_ ` here?

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+    const currentTime = this.player_.currentTime();
+    const segmentLength = this.segmentLength_ || 12;
+
+    // the live edge window is the amount of seconds away from live
+    // that a player can be, but still be considered live.
+    // we add 0.07 because the live tracking happens every 30ms
+    // and we want some wiggle room for short segment live playback
+    const liveEdgeWindow = segmentLength + 0.07;
+
+    return liveCurrentTime !== Infinity && (liveCurrentTime - liveEdgeWindow) >= currentTime;
```
ldayananda: why would liveCurrentTime be `Infinity`?

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+  trackLive_() {
+    this.pastSeekEnd_ = this.pastSeekEnd_ || 0;
+    const seekable = this.player_.seekable();
+
+    // skip undefined seekable
+    if (!seekable || !seekable.length) {
+      return;
+    }
+
+    const newSeekEnd = this.seekableEnd(0);
```
ldayananda: even though I haven't really seen it, might be worth supporting multiple seekable ranges from the beginning

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+
+    // we can only tell if we are behing live, when seekable changes
```
ldayananda: typo `behing`

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+    } else {
+      this.stop();
+    }
+  }
+
+  /**
+   * start tracking live playback
+   */
+  start() {
+    if (this.started()) {
```
ldayananda: might want to make the name a little more descriptive so it's not confused with `hasStarted`

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Get the live time window
+   */
+  liveWindow() {
+    return this.seekableEnd() - this.seekableStart();
+  }
+
+  /**
+   * Determines if the player is live, only checks wether this component
```
ldayananda: type `wether`

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+   * is tracking live playback or not
+   */
+  isLive() {
+    return this.started();
```
ldayananda: it's possible then to be live but not be using the LiveTracker?

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+   * Seek to the live edge if we are behind the live edge
+   */
+  seekToLiveEdge() {
+    if (this.atLiveEdge()) {
+      return;
+    }
+
+    this.player().pause();
+    this.player().addClass('vjs-waiting');
+    this.one('seekableendchange', () => {
```
ldayananda: wouldn't we call `seekToLiveEdge` after a `seekableendchange`?

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+  pastSeekEnd() {
+    return this.pastSeekEnd_ || 0;
+  }
+
+  /**
+   * If we are currently behind the live edge, aka currentTime will be
+   * behind on a seekableendchange
+   */
+  behindLiveEdge() {
+    return this.behindLiveEdge_;
```
ldayananda: why have both `isBehind()` and this?

src/css/components/_live.scss
```diff
@@ -12,3 +11,52 @@
...
+  @include flex(none);
+  display: inline-flex;
+  height: 100%;
+  margin-left: 0.3em;
+}
+
+.vjs-no-flex .vjs-seek-to-live-control {
+  display: table-cell;
+  width: auto;
+  text-align: left;
```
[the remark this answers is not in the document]
BrandonOCasey: This was mostly a copy paste from what live display does, but I don't think I see any other buttons doing it. I will test to see if we need this code still.

src/css/components/_progress.scss
BrandonOCasey: this simply brings back the normal [display value seen above in the file](https://github.com/videojs/video.js/blob/master/src/css/components/_progress.scss#L7). Previously this was hidden during live playback with `display: none` this code makes sure that it is show, but only when the `liveui` class is on the player. Which is added through the `liveui` option.

src/js/control-bar/control-bar.js
BrandonOCasey: `seekToLive` replaces `liveDisplay` but we wanted the player to be backwards compatible. Without the new option (`liveui`). `liveDisplay` will be shown since the player will only have the `vjs-live` class and `seekToLive` will be hidden since there is no `vjs-liveui` class. With the new option `seekToLive` will be shown since there will be a `vjs-liveui` class and `liveDisplay` will be hidden.

src/js/control-bar/progress-control/load-progress-bar.js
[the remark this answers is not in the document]
BrandonOCasey: Let me see if this causes, issues. That may make more senes.

src/js/control-bar/progress-control/seek-bar.js
BrandonOCasey: yeah we should

src/js/live-tracker.js
BrandonOCasey: I don't think we would really need a function for that though. You would just wait for `seekableendchange` and then seek to the new `seekable.end` to get to the current live edge. This function just does all that for the user and keeps the UI up to date with what is happening.

src/js/live-tracker.js
BrandonOCasey: Yeah if the `liveui` option is not passed in then liveTracker will never be started/live. This one done to preserve backwards compatibility.

src/js/live-tracker.js
BrandonOCasey: So that we can fire a `liveedgechange` event, and so that we return the current state rather than a possible state. We could make isBehind_ private.

gkatsev:
It's probably a caching issue on the CI side.
Also, should 68d97d6 be in a PR against master as well?

BrandonOCasey:
We are seeing issues with `npm run start` but only on node 8, seems like a race condition

BrandonOCasey:
The actual build fix commit doesn't seem to work.

gkatsev:
clearing the cache didn't help, I'd assume the package-lock needs to be regenerated

gkatsev:
I wonder if some package updated for node 10 support and it broke in node 8 or something

src/js/control-bar/control-bar.js
ldayananda: That sounds like `seekToLive` should only be added to the children if `liveui` option is true. 

Also, seems I meant to ask if `liveDisplay` works with the liveTracker as well

src/js/live-tracker.js
ldayananda: seems to make sense for one of two to not be exposed

src/js/control-bar/seek-to-live.js
```diff
@@ -0,0 +1,92 @@
...
+import Button from '../button';
+import Component from '../component';
+import * as Dom from '../utils/dom.js';
+
+/**
+ * Displays the live indicator when duration is Infinity.
+ *
+ * @extends Component
+ */
+class SeekToLive extends Button {
```
gkatsev: there's probably no need and it will overly complicate things, but should this extend LiveDisplay?

BrandonOCasey:
OK so I think it might make sense for either `LiveDisplay` or `SeekToLive` to be added/removed on `durationchange` depending upon options. From a backwards compatibility standpoint I don't know if we can do that for `LiveDisplay`, but we could do it for SeekToLive.

src/js/control-bar/seek-to-live.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    this.contentEl_ = Dom.createEl('div', {
+      className: 'vjs-seek-to-live',
+      innerHTML: `<span class="vjs-control-text">${this.localize('Stream Type')}\u00a0</span>${this.localize('LIVE')}`
+    }, {
+      'aria-live': 'off'
+    });
+
+    this.circleEl_ = Dom.createEl('span', {
```
gkatsev: @OwenEdwards this span creates a visual indicator that we're at the live edge. For accessibility, we'd probably want some control text here that says something like `at live edge`/`not at live edge`. Thoughts about wording or what type of info should be included?

src/js/control-bar/seek-to-live.js
```diff
@@ -0,0 +1,92 @@
...
+  handleClick() {
+    this.player_.liveTracker.seekToLiveEdge();
+  }
+
+  /**
+   * Dispose of the element and stop tracking
+   */
+  dispose() {
+    this.off(this.player_.liveTracker, 'liveedgechange', this.updateLiveEdgeStatus);
+    this.contentEl_ = null;
```
gkatsev: we should null out `this.circleEl_` as well.

src/js/control-bar/seek-to-live.js
gkatsev: `updateLiveEdgeStatus` should then update the control text for this.

src/js/live-tracker.js
gkatsev: we're unsure how that's supposed to work. Any input would be greatly appreciated.

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /**
+   * stop tracking live playback
+   */
+  stop() {
+    if (!this.started()) {
+      return;
+    }
+    this.pastSeekEnd_ = 0;
```
gkatsev: we should set default values for these properties in the constructor, including the `trackingInterval_`, it gives the engine the expectation that these will get set.

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+  liveWindow() {
+    return this.seekableEnd() - this.seekableStart();
+  }
+
+  /**
+   * Determines if the player is live, only checks if this component
+   * is tracking live playback or not
+   */
+  isLive() {
+    return this.started();
```
gkatsev: if someone calls `liveTracker.stop()` this will start reporting false for `isLive` even if we are live. Maybe we should continue verifying that `duration` from the video element is `Infinity`?

BrandonOCasey:
Re: `width: auto` (couldn't find a way to respond to the comment). We need it because localization may make the live button take up more space.

src/js/control-bar/seek-to-live.js
BrandonOCasey: I don't think that it should, as it isn't a button right now.

src/js/player.js
```diff
@@ -2289,7 +2289,11 @@ class Player extends Component {
 
       if (seconds === Infinity) {
         this.addClass('vjs-live');
-        if (this.options_.liveui) {
+        // on Android the seekable end is often Infinity,
+        // this makes the new liveui useless since we cannot
+        // seek within the live window at all. Android will
+        // revert back to the old liveui.
+        if (this.options_.liveui && !browser.IS_ANDROID) {
```
gkatsev: actually, given that this is off by default, should we just document that Android is broken and you shouldn't turn liveui on on Android unless you are using some MSE-based HLS?

src/js/live-tracker.js
BrandonOCasey: The problem is that since we are not tracking live when it is stopped. A lot of the UI won't work correctly if the liveTracker is stopped during live playback. I think it might be better to revert to liveDisplay if live-tracker is stopped during live playback.

src/js/player.js
BrandonOCasey: Yeah that may be a better idea.

src/js/live-tracker.js
BrandonOCasey: what about calling `stop` in the constructor? That would set all the values for us.

BrandonOCasey: actually nevermind that won't work.

BrandonOCasey: OK I added a `reset_` function that will be shared in the `constructor` and `stop`

src/js/live-tracker.js
```diff
@@ -0,0 +1,204 @@
...
+   * Seek to the live edge if we are behind the live edge
+   */
+  seekToLiveEdge() {
+    if (this.atLiveEdge()) {
+      return;
+    }
+
+    this.player().pause();
+    this.player().addClass('vjs-waiting');
+    this.one('seekableendchange', () => {
```
gkatsev: why wait for `seekableendchange` for `seekToLiveEdge`? Shouldn't just set `currentTime(this.seekableEnd())`?

sandbox/live.html.example
```diff
@@ -22,7 +22,12 @@
   <script>
     // fake a livestream for easy testing
     var vid = document.getElementById('vid1');
-    var player = videojs(vid, {liveui: true});
+    var liveui = true
+
+    if (videojs.browser.IS_ANDROID) {
+      liveui = false;
```
gkatsev: alternatively, we should force VHS on on android

gkatsev:
The at edge marker doesn't look very circular to me, even when I turn off the gradient. Should we just use [circle font-icon](http://videojs.github.io/font/) like the progress bar uses?

src/js/live-tracker.js
BrandonOCasey: So we have to wait for `seekableendchange` because you can only seek within `seekable`. Example:

* You are at 4s 
* seekable end is currently at 8s
* seekable has 4s added to it every 4s (I called this seekableIncrement_)
* live playback is a bit passed seekable end, say 10s


If we seek right away you will appear to be live until seekable end changes, then you will be 2s behind, and catch up after 2s. Then you will be 2s behind again when seekable end changes again.

BrandonOCasey:
didn't even think to do that, but yeah that might work better.

src/css/components/_live.scss
```diff
@@ -12,3 +11,55 @@
...
+// css for the new live ui below
+.video-js .vjs-seek-to-live-control {
+  cursor: pointer;
+  @include flex(none);
+  display: inline-flex;
+  height: 100%;
+  margin-left: 0.7em;
+  font-size: 1em;
+  line-height: 3em;
+  width: auto;
```
gkatsev: we should add a `min-width: 4em;` here, so, it'll be at least as big as all other components.

src/css/components/_live.scss
```diff
@@ -12,3 +11,55 @@
...
+  display: none;
+}
+
+// css for the new live ui below
+.video-js .vjs-seek-to-live-control {
+  cursor: pointer;
+  @include flex(none);
+  display: inline-flex;
+  height: 100%;
+  margin-left: 0.7em;
```
gkatsev: rather than a margin-left, I think adding a 0.5em padding on the left and right would produce a better result that matches other controls.

src/css/components/_live.scss
```diff
@@ -12,3 +11,55 @@
...
+  cursor: auto;
+}
+
+.vjs-seek-to-live-control .vjs-seek-to-live-circle {
+  border-radius: 50%;
+  width: 0.75em;
+  height: 0.75em;
+  margin-top: auto;
+  margin-bottom: auto;
+  margin-right: 0.3em;
```
gkatsev: I think a 0.5em right margin looks better.

src/css/components/_live.scss
```diff
@@ -12,3 +11,55 @@
...
+  background: radial-gradient(#808080, #666666);
```
gkatsev: I think going with a flat color might be better, like `#888` here and `red` below

src/js/live-tracker.js
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * start tracking live playback
+   */
+  start() {
+    if (this.started()) {
+      return;
+    }
+
+    this.trackingInterval_ = this.setInterval(this.trackLive_, 30);
```
gkatsev: we should also call trackLive on pause and play, if we don't do so already.

src/js/live-tracker.js
gkatsev: I'm not sure I follow. Also, what happens if we had just missed the previous seekable end change? Would we have to wait 4 seconds to get to the live edge?

BrandonOCasey: yes it would

gkatsev: As a user I would be very confused and think that something is broken. @videojs/playback any ideas on how to better handle this?

gkatsev: I see it change from at live edge and not at live edge as part of regular playback.

gesinger: We may want to be careful with how we define live. If segments are 4 seconds long, then we probably want live to be considered within the last 8 seconds (2 segment durations), because of the case mentioned (where, if we are unlucky with manifest refreshes, we could end up seeking to a segment we think is at the live edge, but the playlist refreshes just after the request, so we start 4 seconds + request and processing duration back from what we think is seekable end).

The other solution is to seek to a point past where the seekable end is, figuring that we'll wait until the next refresh to ensure we start at the most live point, but this delays startup up to the duration of a segment (plus request time). I think this experience would be worse, but can be controlled through shorter segment durations.

src/js/control-bar/control-bar.js
gkatsev: Thinking about it some more, we should just have it in by default and hidden by CSS. At some point, we should fix it so that just calling `component.show()` from a default-hidden component actually works.

src/js/live-tracker.js
mjneil: Could we just keep track of timestamps for when we get seekableendchange events, and then seek to `this.seekableEnd() + timeSinceLastSeekableendchange`?

src/js/live-tracker.js
BrandonOCasey: if seekable end is Infinity

gkatsev:
I think things we're going to do in another PR:
- update seekToLive
- make the live at edge marker be accessible

Anything else?

BrandonOCasey:
I think thats it

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -100,7 +105,18 @@ class SeekBar extends Slider {
...
   update_(currentTime, percent) {
-    const duration = this.player_.duration();
+    const liveTracker = this.player_.liveTracker;
+    let duration = this.player_.duration();
+
+    if (liveTracker.isLive()) {
+      duration = this.player_.liveTracker.seekableEnd();
+    }
+
+    if (liveTracker.seekableEnd() === Infinity) {
```
ldayananda: I'm not sure if seekable.end is ever `Infinity`

ldayananda: actually this is the android thing again, right? If so, a comment will help a lot!

BrandonOCasey: it is on Android native hls

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -173,7 +189,20 @@ class SeekBar extends Slider {
...
+    const currentTime = this.getCurrentTime_();
+    let percent;
+    const liveTracker = this.player_.liveTracker;
+
+    if (liveTracker.isLive()) {
+      percent = (currentTime - liveTracker.seekableStart()) / liveTracker.liveWindow();
+
+      // prevent the percent from changing at the live edge
+      if (liveTracker.atLiveEdge()) {
+        percent = 1;
```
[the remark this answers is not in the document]
BrandonOCasey: If the liveTracker  does not report that we are at the live edge the button will not show that we are at the live edge either. This code allows us to be a bit behind the live edge, but still show that we are live to prevent small seek bar movements.

src/js/live-tracker.js
BrandonOCasey: added

BrandonOCasey:
tests pass via browsersync edge "locally"

Issue #2470, opened 2015-08-13
Control buttons submit form, missing type="button" (5.0.0-rc.49)


nebrot:
In my web-application i have a player which is integrated in a form with a checkbox to mark uploaded video as deleted. My problem is that all control buttons of the player submit form, because they don't have type="button", default value is "submit" (http://www.w3.org/TR/html-markup/button.html).

Example: http://jsfiddle.net/sLsvLv0q/12/
Windows 8.1: Firefox 40, Chrome 44, IE 11

mmcc:
@nebrot Thanks for the heads up! I opened #2471 to address this.

nebrot:
@mmcc Thanks for your fast (upcoming) fix. Keep up the good work!

PR #5512, opened 2018-10-18
fix: Do not use default button type.

syranez:
The default button type is "submit" which triggers a form submit. Thus
if videojs is embedded in a form element the button triggers a form
submit.

This is the same problem with same solution as discussed in https://github.com/videojs/video.js/issues/2470


gkatsev:
Yeah, our Button component has that by default which is nice https://github.com/videojs/video.js/blob/master/src/js/button.js#L43-L44

PR #5518, opened 2018-10-22
chore: remove rollup filesize plugin to double build speed

BrandonOCasey:
So `rollup-plugin-filesize` is minifying all of our code in video.js, then throwing it out to report a minified file size. I think we should just get rid of this plugin because there isn't a great way to turn it off

initial builds for each build target go from ~12s -> 6s and rebuilds go from 6s -> 1s

Closes #5506


gkatsev:
Things are passing in travis as well but github is kind of broken https://travis-ci.org/videojs/video.js/builds/444768352, anyway, merging.

PR #5519, opened 2018-10-22
chore: fix rollup watch bulid with globals/externals

BrandonOCasey:
Currently rollup watch is broken for most "module" (cjs/es) builds. All it took was some externals/globals to get it to work correctly.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -129,9 +129,7 @@ export default cliargs => [
         'video.js': path.resolve(__dirname, './src/js/video.js'),
         '@videojs/http-streaming': path.resolve(__dirname, './node_modules/@videojs/http-streaming/dist/videojs-http-streaming.es.js')
       }),
-      primedResolve,
       json(),
-      primedCjs,
       primedBabel,
       cliargs.progress !== false ? progress() : {}
     ],
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -73,7 +73,15 @@ const externals = {
     'xhr',
     'tsml',
     'safe-json-parse/tuple',
-    'videojs-vtt.js'
+    'videojs-vtt.js',
+    'url-toolkit',
+    'm3u8-parser',
+    'mpd-parser',
+    'mux.js',
+    'mux.js/lib/mp4',
+    'mux.js/lib/tools/ts-inspector.js',
+    'mux.js/lib/mp4/probe',
+    'aes-decrypter'
   ]),
   test: Object.keys(globals.test).concat([
   ])
@@ -174,9 +182,7 @@ export default cliargs => [
     },
     external: externals.module,
     plugins: [
-      primedResolve,
       json(),
-      primedCjs,
       primedBabel,
       cliargs.progress !== false ? progress() : {}
     ],
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -14,8 +14,19 @@ const compiledLicense = _.template(fs.readFileSync('./build/license-header.txt',
 const bannerData = _.pick(pkg, ['version', 'copyright']);
 const banner = compiledLicense(Object.assign({includesVtt: true}, bannerData));
 
-// to prevent going into a screen during rollup
-process.stderr.isTTY = false;
+const watch = {
+  clearScreen: false
+};
+
+const onwarn = (warning) => {
+  // ignore unknow option for --no-progress
+  if (warning.code === 'UNKNOWN_OPTION' && warning.message.indexOf('progress') !== -1) {
+    return;
+  }
+
+  // eslint-disable-next-line no-console
+  console.warn(warning.message);
+};
 
 const primedIgnore = ignore(['videojs-vtt.js']);
 const primedResolve = resolve({
@@ -90,7 +101,9 @@ export default cliargs => [
       primedCjs,
       primedBabel,
       cliargs.progress !== false ? progress() : {}
-    ]
+    ],
+    onwarn,
+    watch
   },
   // es, cjs
   {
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -14,21 +14,8 @@ const compiledLicense = _.template(fs.readFileSync('./build/license-header.txt',
 const bannerData = _.pick(pkg, ['version', 'copyright']);
 const banner = compiledLicense(Object.assign({includesVtt: true}, bannerData));
 
-const watch = {
-  clearScreen: false
-};
-
-const onwarn = (warning) => {
```
BrandonOCasey: This was hiding several warnings that we should have handled.

rollup.config.js
```diff
@@ -14,21 +14,8 @@ const compiledLicense = _.template(fs.readFileSync('./build/license-header.txt',
 const bannerData = _.pick(pkg, ['version', 'copyright']);
 const banner = compiledLicense(Object.assign({includesVtt: true}, bannerData));
 
-const watch = {
-  clearScreen: false
```
BrandonOCasey: I don't think we want to go into a screen ever, so I added the fix from `videojs-generate-rollup-config`

gkatsev: Why isTTY over clearScreen? I would think it would be better to use the rollup option over modifying the `isTTY` option, especially since clearScreen should be working (I fixed it).

rollup.config.js
gkatsev: What type of thing specifically? Does adding the globals/externals prevent it from printing import errors but still keeps the same behavior?

rollup.config.js
BrandonOCasey: There isn't an option to disable screen usage when doing rollup outside of watch. The last time I checked it was still going into a screen for builds.

gkatsev: Does it matter if we go into screen mode outside of watch? Also, I don't think it actually uses the alternate screen during regular builds, it should only be happening during watches.

BrandonOCasey: I guess it doesn't matter too much, it just messed with scroll back history

BrandonOCasey: I will have to test and see if it still uses a screen during build.

rollup.config.js
```diff
@@ -14,21 +14,8 @@ const compiledLicense = _.template(fs.readFileSync('./build/license-header.txt',
...
-};
-
-const onwarn = (warning) => {
-  if (warning.code === 'UNUSED_EXTERNAL_IMPORT' ||
-      warning.code === 'UNRESOLVED_IMPORT' ||
-      (warning.code === 'UNKNOWN_OPTION' &&
-       warning.message.indexOf('progress') !== -1)) {
```
gkatsev: looks like the "progress" warning is still getting printed out, https://travis-ci.org/videojs/video.js/builds/444848160#L591-L618, since imports have been fixed, we can remove that piece but keep the unknown progress option piece.

rollup.config.js
gkatsev: Just tried it out and rollup doesn't use the second screen during regular builds, it only uses it during watch builds. This is disabled with `clearScreen: false`.

rollup.config.js
BrandonOCasey: ok

rollup.config.js
```diff
@@ -60,8 +88,10 @@ export default cliargs => [
       file: 'dist/video.js',
       name: 'videojs',
       strict: false,
-      banner
+      banner,
+      globals: globals.browser
```
gkatsev: do we want to specify globals/external for the UMD files? Since we want the dependencies to be included for those or does resolve and cjs plugins take care of it?

rollup.config.js
```diff
@@ -83,20 +113,25 @@ export default cliargs => [
...
+        globals: globals.module
       }
     ],
+    external: externals.module,
     plugins: [
       alias({
         'video.js': path.resolve(__dirname, './src/js/video.js'),
         '@videojs/http-streaming': path.resolve(__dirname, './node_modules/@videojs/http-streaming/dist/videojs-http-streaming.es.js')
       }),
+      primedResolve,
```
gkatsev: we specifically don't want to include resolve and cjs plugins for the es/cjs builds because then the dependencies are included in the build.

gkatsev: Same with the core.js file below.

gkatsev: If you run `npm run assets`, the esm and cjs builds should be significantly smaller than the main video.js file compare the output in this PR https://travis-ci.org/videojs/video.js/builds/445200591#L964-L968
```
┌─────────────────────────────┬───────────┬───────────┐
│ video.cjs.js                │   1.55 MB │ 369.78 KB │
├─────────────────────────────┼───────────┼───────────┤
│ video.es.js                 │   1.55 MB │ 369.73 KB │
├─────────────────────────────┼───────────┼───────────┤
│ video.js                    │    1.7 MB │ 392.46 KB │
└─────────────────────────────┴───────────┴───────────┘
```

vs master
```
┌─────────────────────────────┬───────────┬───────────┐
│ video.cjs.js                │   1.23 MB │ 290.51 KB │
├─────────────────────────────┼───────────┼───────────┤
│ video.es.js                 │   1.23 MB │ 290.46 KB │
├─────────────────────────────┼───────────┼───────────┤
│ video.js                    │   1.71 MB │ 394.92 KB │
└─────────────────────────────┴───────────┴───────────┘
```

Eventually, when we figure out how not to bundle VHS, it can be a smaller size difference.

rollup.config.js
BrandonOCasey: the only modules we want to exclude in the umd are `global` since those don't have to be a module and can just be the actual global in the browser

PR #5520, opened 2018-10-22
chore: move scss grunt tasks to npm scripts

BrandonOCasey:
Move all the scss login from grunt into npm scripts, so that one day we can remove grunt


build/grunt.js
```diff
@@ -85,19 +84,6 @@ module.exports = function(grunt) {
     },
     dist: {},
     watch: {
-      dist: {
-        files: [
-          'build/temp/video.js',
-          'build/temp/alt/video.novtt.js',
-          'build/temp/video-js.css',
```
BrandonOCasey: We no longer need this since `rollup` handles the the js files and `node-sass` will handle the css

build/grunt.js
```diff
@@ -128,27 +114,7 @@ module.exports = function(grunt) {
       a11y:  { src: 'sandbox/descriptions.html.example', dest: 'sandbox/descriptions.test-a11y.html' }, // Can only test a file with a .html or .htm extension
       examples: { cwd: 'docs/examples/', src: ['**/**'], dest: 'dist/examples/', expand: true, filter: 'isFile' }
     },
-    cssmin: {
```
BrandonOCasey: add as a npm script under `cssmin`

build/grunt.js
```diff
@@ -200,14 +166,6 @@ module.exports = function(grunt) {
         dest: 'dist/video-js-' + version.full + '.zip'
       }
     },
-    version: {
```
BrandonOCasey: i'm not sure that this actually did anything, I see no difference in master with it removed

package.json
```diff
@@ -20,9 +20,21 @@
     "changelog": "conventional-changelog -p videojs -i CHANGELOG.md -s",
     "build": "grunt dist",
     "minify": "babel-node build/minify.js",
+    "cssmin": "npm-run-all cssmin:*",
+    "cssmin:cdn": "cleancss dist/alt/video-js-cdn.css -o dist/alt/video-js-cdn.min.css",
+    "cssmin:default": "cleancss dist/video-js.css -o dist/video-js.min.css",
+    "sass": "npm-run-all sass:* autoprefixer",
+    "sass:cdn": "node-sass src/css/vjs-cdn.scss dist/alt/video-js-cdn.css",
+    "sass:default": "node-sass src/css/vjs.scss dist/video-js.css",
```
gkatsev: Tests still rely on `build/temp/video-js.css`, we should update that and also make sure that there aren't any other uses of `build/temp/video-js.css` like in sandbox or doc files.

BrandonOCasey: good call

build/grunt.js
```diff
@@ -113,41 +99,11 @@ module.exports = function(grunt) {
       }
     },
     copy: {
-      minor: {
-        files: [
-          {expand: true, cwd: 'build/temp/', src: ['*'], dest: 'dist/'+version.majorMinor+'/', filter: 'isFile'} // includes files in path
-        ]
```
BrandonOCasey: We don't use these

build/grunt.js
```diff
@@ -200,50 +156,7 @@ module.exports = function(grunt) {
...
-    },
     browserify: {
-      build: {
-        options: browserifyGruntOptions(),
-        files: {
-          'build/temp/video.js': ['es5/video.js']
-        }
-      },
-      buildnovtt: {
-        options: browserifyGruntOptions({transform: [
```
BrandonOCasey: We don't use any of these anymore either

build/grunt.js
```diff
@@ -276,11 +189,6 @@ module.exports = function(grunt) {
         src: 'test/coverage/lcov.info'
       }
     },
-    concat: {
-      options: {
```
BrandonOCasey: not used

test/karma.conf.js
```diff
@@ -29,7 +29,6 @@ module.exports = function(config) {
 
     // Using precompiled tests
     // files: [
-    //   '../build/temp/video-js.css',
```
BrandonOCasey: Should we just remove this comment

package.json
```diff
@@ -92,13 +102,10 @@
     "grunt-contrib-concat": "^1.0.1",
```
gkatsev: I think this isn't used anymore and could be removed

test/karma.conf.js
gkatsev: We may as well.

build/grunt.js
```diff
@@ -312,25 +218,19 @@ module.exports = function(grunt) {
         'browserify:watch'
       ]
     },
-    usebanner: {
```
gkatsev: I guess we can remove `grunt-banner` as well, though, this made me realize we still keep the vttjs license in the novtt file banner (though we shouldn't bother with it in this PR)

gkatsev: https://github.com/videojs/video.js/issues/5526

sandbox/vertical-volume.html.example
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
gkatsev: seems useful, though I tend to just manually call `player.controlBar.addChild('volumePanel', {inline: false});`

PR #5522, opened 2018-10-23
fix: Prevent ResizeManager from being clicked on safari, fix playerresize on firefox

BrandonOCasey:
Currently right clicking on the player on safari does not fire a `contextmenu` event, but it does on every other browser. Right clicking on the player in safari fires a `contextmenu` event on the `ResizeManager` iframe instead. The only change that needed to happen for right click to fire on the player again, was to hide `ResizeManager`.

EDIT:
I also found that the `playerresize` event does not work in Firefox. After removing `visibility: hidden;` it fires as we would expect, I added that change to this pull request, as I think it makes sense.

### DIFF
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -50,5 +50,4 @@
   width: 100%;
   height: 100%;
   border: none;
-  visibility: hidden;
 }
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -50,4 +50,5 @@
   width: 100%;
   height: 100%;
   border: none;
+  z-index: -1000;
 }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4069,6 +4069,7 @@ Player.prototype.options_ = {
 
   // Included control sets
   children: [
+    'resizeManager',
     'mediaLoader',
     'posterImage',
     'textTrackDisplay',
@@ -4076,8 +4077,7 @@ Player.prototype.options_ = {
     'bigPlayButton',
     'controlBar',
     'errorDisplay',
-    'textTrackSettings',
-    'resizeManager'
+    'textTrackSettings'
   ],
 
   language: navigator && (navigator.languages && navigator.languages[0] || navigator.userLanguage || navigator.language) || 'en',
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4069,7 +4069,6 @@ Player.prototype.options_ = {
 
   // Included control sets
   children: [
-    'resizeManager',
     'mediaLoader',
     'posterImage',
     'textTrackDisplay',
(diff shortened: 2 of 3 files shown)


BrandonOCasey:
Oh I see what you mean, when the video is actually playing it  gives a menu for the iframe, but before playback it gives a menu for the player. thats why I thought it was fixed. Looking further into this 👀

BrandonOCasey:
Here is what I tried so far
1. `pointer-events: none` on the iframe
  * `contextmenu` is still hijacked
1. Wrap the iframe in a div and add `pointer-events: none` to it
  * `contextmenu` is still hijacked
1. Move the iframe before the video element
 * `contextmenu` is not hijacked but `playerresize` does not work
1. Try various `visiblitiy: none`/`display: none` on the iframe or a wrapper div
 * Causes `playerresize` not to fire but usually fixed the `contextmenu` issue

After all that I opted to re-trigger the `contextmenu` event on the player from the iframe, and everything appears to work as expected.

src/js/player.js
```diff
@@ -4069,15 +4069,15 @@ Player.prototype.options_ = {
 
   // Included control sets
   children: [
+    'resizeManager',
```
gkatsev: maybe revert this change if we need to re-trigger the event?

src/js/resize-manager.js
```diff
@@ -84,6 +86,18 @@ class ResizeManager extends Component {
...
+  /**
+   * Due to an issue on safari, the iframe for this object takes over
+   * right clicks on the player. So we have to re-trigger them from the
+   * iframe to the player.
+   * @see https://github.com/videojs/video.js/pull/5522
+   *
+   * @param {Event} e the contextmenu event that fired
+   */
+  handleContextmenu_(e) {
+    this.player_.trigger(e);
```
gkatsev: Should we preventDefault here? Thoughts @misteroneill?

src/js/player.js
BrandonOCasey: OK, so I was going to remove this change and put resizeManager back at the bottom of the children list, but it breaks the controls for Safari. When I removed the `visibility: hidden;` css for the resize manager it made it so that the iframe was also intercepting other clicks on the video element. So we have to keep resizeManager at the top for controls to work correctly.

src/css/video-js.scss
```diff
@@ -50,5 +50,5 @@
   width: 100%;
   height: 100%;
   border: none;
-  visibility: hidden;
```
BrandonOCasey: Fixes firefox not firing `playerresize`

src/css/video-js.scss
```diff
@@ -50,5 +50,5 @@
...
+  z-index: -1000;
```
BrandonOCasey: Fixes safari and firefox iframe `contextmenu` hijacking

src/js/player.js
gkatsev: If we still wanted to move it, I would put it after mediaLoader.

src/css/video-js.scss
BrandonOCasey: (firefox was started to hijack `contextmenu` once we removed `visibility: none`)

PR #5525, opened 2018-10-24
fix: blob sources being updated in the source cache

### LINKED ISSUE #5504: Blob Url Support broke with 7.2.1
## Description
When using a dynamically generated blob url as a video source, any player interaction will be silenced by video.js

It works flawlessly when triggering `.play()` directly on the HTMLVideoElement though.

Reproduction case with comparison of 7.2.0 and 7.2.1: https://github.com/flipace/video.js-7.2.1-blob-issue. This is related to a change made in https://github.com/videojs/video.js/pull/5371

## Steps to reproduce
1. Generate a blob url for a video
2. Pass the blob url as a src (with type) to videojs
3. Try triggering playback of the video

### Expected
Playback should start when clicking the play button(s).

### Actual
Nothing happens.

### Error output
No errors. video.js is stuck and will always wait for the video to be loaded into its internal source cache (which never happens though)

#### videojs
7.2.1 and above are affected
#### browsers | OSes
since this is a videojs internal issue probably all of them

BrandonOCasey:
Fixes #5504

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1465,13 +1465,10 @@ class Player extends Component {
       const eventSrc = event.src;
 
       // if we have a playerSrc that is not a blob, and a tech src that is a blob
-      // we may skip the source cache update
       if (playerSrc && !(/^blob:/).test(playerSrc) && (/^blob:/).test(eventSrc)) {
 
-        // we skip updating the source cache if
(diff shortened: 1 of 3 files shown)


gkatsev:
I can confirm it fixes #5504 using the example provided there.

PR #5528, opened 2018-10-25
chore: switch to videojs-generate-karma-config

BrandonOCasey:
Heavily simplifies a lot of things and uses the shared logic from `videojs-generate-karma-config`

## Changes
* Removes all browserify stuff outside of karma
* Uses `karma-static-server` middleware instead of connect
* Uses `test/debug.html` rather than `test/index.html` for debugging tests


test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
...
-      base: 'BrowserStack',
-      browser: 'safari',
-      os: 'OS X',
-      os_version: 'High Sierra'
-    },
+  config = generate(config, options);
+
+  config.frameworks.push('browserify');
+  config.plugins.push('karma-browserify');
+  config.browserify = {
```
BrandonOCasey: ideally this would all be in rollup, but the tests are heavily reliant on some things browserify does right now, so that will be a big job.

test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
...
+    debug: true,
+    plugin: ['proxyquireify/plugin'],
+    transform: [
+      ['babelify', {"presets": [["@babel/preset-env", {"loose": true}]]}],
+      'browserify-istanbul'
+    ]
+  };
 
-    safari9_bs: {
```
gkatsev: we should make sure we're testing on the same browsers as before.

test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
-module.exports = function(config) {
-  // build out a name for browserstack
-  // {TRAVIS_BUILD_NUMBER} [{TRAVIS_PULL_REQUEST} {PR_BRANCH}] {TRAVIS_BRANCH}
-  var browserstackName = process.env.TRAVIS_BUILD_NUMBER;
-
-  if (process.env.TRAVIS_PULL_REQUEST !== 'false') {
-    browserstackName += ' ' + process.env.TRAVIS_PULL_REQUEST + ' ' + process.env.TRAVIS_PULL_REQUEST_BRANCH;
```
gkatsev: the karma config package is missing the initial whitespace here, so, the build number and PR value would be smooshed together.

test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
...
-      pollingTimeout: 30000,
-      captureTimeout: 600,
-      timeout: 600
-    },
-    customLaunchers: getCustomLaunchers(),
+const generate = require('videojs-generate-karma-config');
 
-    // The HTML reporter seems to be busted right now, so we're just using text in the meantime
-    // along with the summary after the test run.
-    coverageReporter: {
```
gkatsev: we should figure out how to enable coverage again. Previously we've used coveralls but I have no qualms in switching to codecov

test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
...
+  config = generate(config, options);
+
+  config.frameworks.push('browserify');
+  config.plugins.push('karma-browserify');
+  config.browserify = {
+    debug: true,
+    plugin: ['proxyquireify/plugin'],
+    transform: [
+      ['babelify', {"presets": [["@babel/preset-env", {"loose": true}]]}],
+      'browserify-istanbul'
```
gkatsev: is it possible to only include this it runs with coverage?

test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
...
+    ]
+  };
 
-    safari9_bs: {
-      base: 'BrowserStack',
-      browser: 'safari',
-      os: 'OS X',
-      os_version: 'El Capitan'
-    },
+  config.files = [
```
gkatsev: can files be moved to the top of the configurations?

test/karma.conf.js
```diff
@@ -1,196 +1,41 @@
...
-    // cli runner port
-    runnerPort: 9100,
-    colors: true,
-    logLevel: config.LOG_INFO,
-    captureTimeout: 300000,
-    browserNoActivityTimeout: 300000,
-    browserDisconnectTimeout: 300000,
-    browserDisconnectTolerance: 3,
-
-    browserStack: {
```
gkatsev: the project, name, and build, don't all seem to be configurable or that they'll be set to the correct value.

build/grunt.js
```diff
@@ -1,86 +1,26 @@
-import {gruntCustomizer, gruntOptionsMaker} from './options-customizer.js';
```
gkatsev: I don't think this file is used anywhere, so, it could be removed.

build/grunt.js
```diff
@@ -1,86 +1,26 @@
...
 import isDocsOnly from './docs-only.js';
 
 module.exports = function(grunt) {
   require('time-grunt')(grunt);
 
-  let _ = require('lodash');
```
gkatsev: Other than here and in options-customizer, lodash isn't used anywhere else, I don't think.

test/karma.conf.js
gkatsev: That is, we should test in chrome, firefox, safari on High Sierra (maybe time to mojave) and Safari on El Capitan, MS Edge, and IE11

test/karma.conf.js
BrandonOCasey: https://github.com/videojs/videojs-generate-karma-config/pull/8

test/karma.conf.js
BrandonOCasey: In videojs-generate-karma-config we always generate coverage

build/grunt.js
BrandonOCasey: I think the file in question is already gone.

test/karma.conf.js
BrandonOCasey: Three are different here things here:
1. The project name will be `video.js` instead of `Video.js`, i'm not sure if that is an issue
1. Missed a value above in  videojs-generate-karma. https://github.com/videojs/videojs-generate-karma-config/pull/9
1. The browserStack project name will be lowercase, and the `build` value will be `process.env.TRAVIS_BUILD_NUMBER || process.env.BUILD_NUMBER`, but I think that makes more sense than the previous value

test/karma.conf.js
BrandonOCasey: Going to add Safari EL capitan and make a pr to update high sierra to mojave in videojs-generate-karma-config (probably a major)

build/grunt.js
gkatsev: ah, missed that because it's a collapsed diff

test/karma.conf.js
gkatsev: Renaming it isn't huge but it will mean that previous runs will be considered separate from the new runs. It would be nice if we can keep the name.
Also, I set the name as the build specifically so that there will be more info there https://github.com/videojs/video.js/pull/4903#issuecomment-360634649. The Browserstack UI will show the build value and not the name, so, having the info from the full name, is useful in narrowing down the specific build you want to look at.

Also, since we no longer care about IE8, we should turn off video recording of the browserstack session, it'll speed things up a lot.

test/karma.conf.js
gkatsev: I would prefer if coverage was only generated when `npm test --coverage` was run and not if you just run `npm test`.

gkatsev: The point isn't just the time spent. Coverage reporting instruments the code and can cause issues, so, it's always nice to be able to run without it.

Fixing it up in another PR is reasonable as long as we don't forget to do it.

BrandonOCasey: If we decide to turn of coverage I added an option. https://github.com/videojs/videojs-generate-karma-config/pull/11

test/karma.conf.js
BrandonOCasey: https://github.com/videojs/videojs-generate-karma-config/pull/12

BrandonOCasey:
coverage reporting in `karma-coverage` is throwing an error, I think we should opt to fix it in another pr, but coverage generation appears to be working.

test/karma.conf.js
gkatsev: Looks like project names are not case sensitive, since this build shows up with our old builds.

gkatsev:
Switching to rollup for tests and enabling coverage definitely belong other PRs. Changes look good, I'd like to run this locally.

gkatsev:
Safari and Safari Technology Preview won't open/run tests locally when I do `npm test`.

gkatsev:
This may be broken in general for me and not just in this branch.

gkatsev:
Safari and STP tests were failing for me because of https://github.com/karma-runner/karma-safari-launcher/issues/29

gkatsev:
So, with these changes there's no way to access tests without going through the karma server. The built test file is not really available anywhere I assume? Maybe it's just something I need to get used to. Also, we should potentially add a link to it from the sandbox/index.html file.

BrandonOCasey:
I was thinking that we should have an index.html in the root of the repo that points to sandbox and `/test/debug.html` which is similar to our previous page that does not run tests through karma

PR #5534, opened 2018-10-30
feat: responsive caption settings

BrandonOCasey:
* Makes reset/done buttons take up two columns and centers them
* Makes the width of all selects the same 
* When the caption settings are open on a small player, move to a single column layout. Currently there is a horizontal scroll bar.

### DIFF
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -17,7 +17,7 @@
 }
 
 .vjs-text-track-settings .vjs-track-settings-controls {
-  text-align: right;
+  text-align: center;
   vertical-align: bottom;
 }
 
@@ -41,21 +41,36 @@
   }
 
   .vjs-text-track-settings .vjs-track-settings-controls {
-    grid-column: 2;
+    grid-column: 1 / -1;
     grid-row: 2;
-
   }
+
+
+  // 1 column for small players
+  @media (max-width: 425px) {
+
+    .vjs-text-track-settings .vjs-track-settings-font {
+      grid-column: 1;
+      grid-row: 2;
+    }
+
+    .vjs-text-track-settings .vjs-track-settings-controls {
+      grid-column: 1 / -1;
+      grid-row: 3;
+    }
+  }
+
 }
 
 // Form elements
 .vjs-track-setting > select {
-  margin-right: 5px;
+  width: 15em;
 }
 
 .vjs-text-track-settings fieldset {
-  margin: 5px;
-  padding: 3px;
   border: none;
+  height: 8em;
+  width: 1em;
 }
 
 .vjs-text-track-settings fieldset span {
@@ -64,7 +79,7 @@
 
 .vjs-text-track-settings legend {
   color: $primary-foreground-color;
-  margin: 0 0 5px 0;
+  margin: 0 0 1px 0;
 }
 
 // Hide labels, so they are only for screen reader users
@@ -102,4 +117,5 @@
 
 .vjs-track-settings-controls .vjs-default-button {
   margin-right: 1em;
+  margin-bottom: 1em;
 }
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -27,7 +27,7 @@
     display: grid;
     grid-template-columns: 1fr 1fr;
     grid-template-rows: 1fr;
-    // flex an grid for firefox, ie, edge remove bottom padding/margin in a container as size decreases
+    // Flex and Grid for Firefox, IE, and Edge remove the bottom padding/margin in a container as size decreases
     // so we add bottom padding/margin to the last item in the grid instead of here
     // see https://stackoverflow.com/a/23754080
     padding: 20px 24px 0px 24px;
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -34,7 +34,9 @@
   }
 
   // 1 column for small players
-  .vjs-layout-small .vjs-text-track-settings .vjs-modal-dialog-content {
+  .vjs-layout-small .vjs-text-track-settings .vjs-modal-dialog-content ,
+  .vjs-layout-x-small .vjs-text-track-settings .vjs-modal-dialog-content,
+  .vjs-layout-tiny .vjs-text-track-settings .vjs-modal-dialog-content {
     grid-template-columns: 1fr;
   }
 
@@ -43,7 +45,7 @@
 
 // Form elements
 .vjs-track-setting > select {
-  width: 15em;
+  width: 10em;
 }
 
 .vjs-text-track-settings fieldset {
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -25,24 +25,18 @@
 @supports (display: grid) {
   .vjs-text-track-settings .vjs-modal-dialog-content {
     display: grid;
-    grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
-    // grid-template-rows: repeat(425, minmax(1fr 1fr));
+    grid-template-columns: 1fr 1fr;
+    grid-template-rows: 1fr;
   }
 
-  /*
+  .vjs-text-track-settings .vjs-track-settings-controls {
+    grid-column: 1 / -1;
+  }
+
   // 1 column for small players
-  @media (max-width: 425px) {
-
-    .vjs-text-track-settings .vjs-track-settings-font {
-      grid-column: 1;
-      grid-row: 2;
-    }
-
-    .vjs-text-track-settings .vjs-track-settings-controls {
-      grid-column: 1 / -1;
-      grid-row: 3;
-    }
-  }*/
+  .vjs-layout-small .vjs-text-track-settings .vjs-modal-dialog-content {
+    grid-template-columns: 1fr;
+  }
 
 }
 
@@ -99,6 +93,5 @@
 }
 
 .vjs-track-settings-controls .vjs-default-button {
-  margin-right: 1em;
-  margin-bottom: 1em;
+  margin: 1em;
 }
--- a/src/css/components/_captions-settings.scss
+++ b/src/css/components/_captions-settings.scss
@@ -27,6 +27,7 @@
     display: grid;
     grid-template-columns: 1fr 1fr;
     grid-template-rows: 1fr;
+    min-height: 23em;
   }
 
   .vjs-text-track-settings .vjs-track-settings-controls {
@@ -38,6 +39,7 @@
   .vjs-layout-x-small .vjs-text-track-settings .vjs-modal-dialog-content,
   .vjs-layout-tiny .vjs-text-track-settings .vjs-modal-dialog-content {
     grid-template-columns: 1fr;
+    min-height: 34.2em;
   }
 
 }
@@ -46,7 +48,6 @@
 .vjs-track-setting > select {
   margin-right: 1em;
   margin-bottom: 0.5em;
-  width: 10em;
 }
 
 .vjs-text-track-settings fieldset {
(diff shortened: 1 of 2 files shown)


src/js/tracks/text-track-settings.js
```diff
@@ -375,7 +375,7 @@ class TextTrackSettings extends ModalDialog {
       this.localize('Background'),
       '</legend>',
       this.createElSelect_('backgroundColor', legendId),
-      '<span class="vjs-bg-opacity vjs-opacity">',
+      '<span class="vjs-bg-opacity vjs-opacity vjs-track-setting">',
```
BrandonOCasey: added this so that all the selects below them could be the same width.

src/css/components/_captions-settings.scss
```diff
@@ -102,4 +117,5 @@
 
 .vjs-track-settings-controls .vjs-default-button {
   margin-right: 1em;
+  margin-bottom: 1em;
```
BrandonOCasey: the bottom buttons appear right on the border of the dialog box for small players, this fixes that.

gkatsev:
Hm... the netlify deploy is broken. The CSS isn't loading: https://deploy-preview-5534--videojs-docs.netlify.com/test-example/

gkatsev:
Ah, it failed because of node-sass issues https://app.netlify.com/sites/videojs-docs/deploys/5bd8cefc82d3f12505818432

src/css/components/_captions-settings.scss
```diff
@@ -41,21 +41,36 @@
...
 
   .vjs-text-track-settings .vjs-track-settings-controls {
-    grid-column: 2;
+    grid-column: 1 / -1;
     grid-row: 2;
+  }
+
+
+  // 1 column for small players
+  @media (max-width: 425px) {
```
gkatsev: what about using the responsive classes instead of media queries?
Also, shouldn't grid automatically resize if we set a minimum size?

gkatsev:
There's a lot of extra whitespace with the changes made here. 
screen shot 2018-11-02 at 2 55 21 pm
compared to what it looks like now
screen shot 2018-11-02 at 2 57 18 pm
Though, we can probably make it a bit better by increasing the height of the dialog a bit, so, at this size the buttons aren't cut off.

BrandonOCasey:
Going to play around with this some more to reduce the whitespace.

src/css/components/_captions-settings.scss
BrandonOCasey: If responsive was the default, I would think that would be a good idea. Going to check on the grid with a minimum size.

gkatsev:
I'd expect it to still look about the same as it does currently.

src/css/components/_captions-settings.scss
gkatsev: I think it's reasonable to use that feature, though. The idea of the feature is that components can use it to be able to expect what size the player is without doing extra stuff in javascript themselves. I'd want to have it enabled by default eventually.

But if it can just automatically resize based on grid, that would be even better, probably.

BrandonOCasey: Seems like we cannot auto-resize the grid, at least from all I have tried, I changed it to use the responsive class.

gkatsev:
The buttons should be in the bottom right corner rather than in the middle of the dialog.

src/css/components/_captions-settings.scss
```diff
@@ -17,7 +17,7 @@
 }
 
 .vjs-text-track-settings .vjs-track-settings-controls {
-  text-align: right;
+  text-align: center;
```
gkatsev: this should stay `right`

src/css/components/_captions-settings.scss
```diff
@@ -26,35 +26,29 @@
...
-    grid-column: 2;
-    grid-row: 2;
-
-  }
 }
 
+
 // Form elements
 .vjs-track-setting > select {
-  margin-right: 5px;
```
gkatsev: we should bring this back for some whitespace between the select elements

src/css/components/_captions-settings.scss
```diff
@@ -64,7 +58,7 @@
 
 .vjs-text-track-settings legend {
   color: $primary-foreground-color;
-  margin: 0 0 5px 0;
+  margin: 0 0 1px 0;
```
gkatsev: why make this 1px?

src/css/components/_captions-settings.scss
```diff
@@ -26,35 +26,29 @@
...
 
+
 // Form elements
 .vjs-track-setting > select {
-  margin-right: 5px;
+  width: 10em;
 }
 
 .vjs-text-track-settings fieldset {
-  margin: 5px;
```
gkatsev: why remove this margin and padding?

misteroneill:
Agree with @gkatsev's input above.

src/css/components/_captions-settings.scss
```diff
@@ -101,5 +97,5 @@
 }
 
 .vjs-track-settings-controls .vjs-default-button {
-  margin-right: 1em;
+  margin: 1em;
```
gkatsev: with this change, the controls end up being 20px+1em from the button of the dialog, if we keep it as 1em, they'll only be the 20px from the bottom and look more consistent with the other controls.
The 20px come from the padding of the dialog itself.

BrandonOCasey: So the reason why this was changed, was that when the player is small the reset/done buttons touch the very bottom of the captions-setting modal. Maybe we can just add the margin-bottom at small sizes, and make this margin-right again.'

src/css/components/_captions-settings.scss
```diff
@@ -26,30 +26,26 @@
...
-  .vjs-text-track-settings .vjs-track-settings-controls {
-    grid-column: 2;
-    grid-row: 2;
-
-  }
 }
 
 // Form elements
 .vjs-track-setting > select {
   margin-right: 5px;
```
misteroneill: I think we ought to make this a bit bigger. And `em`s instead - maybe `1em`?

Also, I wonder if we can get some vertical spacing for the `select` controls? One way to do this would be:

- Add `margin-bottom: 1em;` here.
- Remove all `margin` from `.vjs-text-track-settings fieldset`.

That looks a lot better at all sizes, I think.

src/css/components/_captions-settings.scss
gkatsev: Ah, didn't notice that. Looking into it, it seems that when it overflows, the "modal" itself doesn't resize, so, we don't get the padding from it. It looks like what we want to do is set `overflow: scroll` on the modal itself when we know it'll scroll, or probably just always set overflow scroll for this modal.
Tried it out and seems to work.

src/css/components/_captions-settings.scss
gkatsev: Works for me.

src/css/components/_captions-settings.scss
BrandonOCasey: after testing this out, it worked but only on chrome, on firefox it actually made the settings go off the modal, and they could not be scrolled to.

src/css/components/_captions-settings.scss
BrandonOCasey: testing

src/css/components/_captions-settings.scss
```diff
@@ -103,3 +100,30 @@
 .vjs-track-settings-controls .vjs-default-button {
   margin-right: 1em;
 }
+
+
+// halfway through vjs-layout-medium side-by-side selects in
+// track-settings-colors collapse, apply fixes as needed
+@media (max-width: 563px) {
```
BrandonOCasey: When the selects collapse in `vjs-layout-medium` the right fieldset for font settings is no longer aligned with the left. This is the only fix I could think of, if we don't think it's an issue we can remove it.

gkatsev: I'm not sure this really matters. Each column can layout by itself, plus, 4 lines less CSS :)

BrandonOCasey: or removing

gkatsev:
Looks like the caption controls are hitting the bottom of the dialog on scroll in non-responsive mode again. Otherwise, this is looking super good.

src/css/components/_captions-settings.scss
```diff
@@ -62,6 +60,11 @@
   display: inline-block;
 }
 
+// style the second select for text colors
+.vjs-text-track-settings fieldset span > select {
+  max-width: 7.3em;
```
BrandonOCasey: At 7.3em the selects won't ever collapse, so that we can ensure that we always have the right `min-height` so that the `controls` never touch the bottom of the modal.

src/css/components/_captions-settings.scss
```diff
@@ -26,30 +26,28 @@
...
 
-  .vjs-text-track-settings .vjs-track-settings-font {
-    grid-column: 2;
-    grid-row: 1;
+  // 1 column for small players
+  .vjs-layout-small .vjs-text-track-settings .vjs-modal-dialog-content ,
+  .vjs-layout-x-small .vjs-text-track-settings .vjs-modal-dialog-content,
+  .vjs-layout-tiny .vjs-text-track-settings .vjs-modal-dialog-content {
+    grid-template-columns: 1fr;
+    min-height: 34.2em;
```
BrandonOCasey: We have to update the `min-height` here so that controls never touch the bottom of the modal

src/css/components/_captions-settings.scss
```diff
@@ -26,30 +26,28 @@
   .vjs-text-track-settings .vjs-modal-dialog-content {
     display: grid;
     grid-template-columns: 1fr 1fr;
-    grid-template-rows: 1fr auto;
+    grid-template-rows: 1fr;
+    min-height: 23em;
```
BrandonOCasey: We set the min-height here so that the `settings-controls` will never touch the bottom of the modal.

BrandonOCasey:
Should be ready for another review, It took me a long time to get the buttons not to hit the bottom of the modal. It seems like it won't actually be possible to do this for `small` players in  non-responsive mode. I think thats OK though and we should just recommend they use responsive mode.

gkatsev:
> I think thats OK though and we should just recommend they use responsive mode.

This is a regression, we should try not to be introducing regressions even if there are better options for people to use. Let's talk some more next week.

BrandonOCasey:
Currently small players (even the default one in sandbox/index.html) sits on the bottom of the modal border

gkatsev:
Ok, yeah, I can repro. Looks like adding a `margin-bottom` to `vjs-track-settings-controls` seems to help. Would that interfere with the rest of the styles? We could have it only be set when responsive mode isn't enabled.

BrandonOCasey:
Seems like that won't work. To clarify the current issue. Chrome appears to work at all sizes and in responsive/not responsive mode with the current pull request. Firefox works in responsive mode, but in non-responsive mode at sizes we would consider less than medium lose the bottom gap. I am going to see if there is anything we can do about it

gkatsev:
Given that it's not a regression, if we don't have a fix for it by EOD, we can just continue without it. However, we should then open a new issue to track this so that we remember to get back to it at some point.

BrandonOCasey:
OK I think I got it working, due to spec implementation details in every browser other than chrome flexbox (and grid apparently) layouts can lose their bottom padding. The fix was adding a margin to the bottom of one of the buttons in the last grid row, and removing the bottom padding from the container. The other tricky thing here was that it had to be done within the `supports grid` context, so that IE did not loose its padding.

src/css/components/_captions-settings.scss
```diff
@@ -26,30 +26,36 @@
   .vjs-text-track-settings .vjs-modal-dialog-content {
     display: grid;
     grid-template-columns: 1fr 1fr;
-    grid-template-rows: 1fr auto;
+    grid-template-rows: 1fr;
+    // flex an grid for firefox, ie, edge remove bottom padding/margin in a container as size decreases
```
OwenEdwards: This comment isn't clear to me - is there a typo in "flex an grid"? Is it "... for firefox, ie & edge" or "... for firefox. ie & edge ..."?

gkatsev: yeah, should be `flex and grid`

Issue #5092, opened 2018-04-12
es6 import of an language file does not work


axten:
## Description
if I try to import a language file I got `ReferenceError: videojs is not defined`

so any lang file should have an import statement vor videojs to fix this.
besides that, a work around could be to provide the pure json files in videojs build.

## Steps to reproduce
just do 
```
import videojs from 'video.js';
import 'video.js/dist/lang/de';
```

### Expected
should work as expected

### Actual
it breaks 

## Additional Information
videojs: 6.8.0

gkatsev:
Oops, I guess the language files we build expect videojs to be available globally. We the build system for them a long time ago before importing/requiring and bundling was popular. We should revisit it soon.

axten:
so in the meantime it would be great to have the pure json files in the npm package

hojas:
+1

gkatsev:
@axten want to submit a PR that adds the json files to the npm publish?

michelson:
👍

eranshmil:
I think you could close this issue since PR #5549 has been merged.

gkatsev:
I think I'd like to keep this open for a bit longer to see if we can figure out a better solution.
Potentially, compile the json files into a `.mjs` file that imports `video.js` at the top of the file.

PR #5549, opened 2018-11-05
feat(lang): es6 import of a language file

eranshmil:
## Description
A temporary solution to enable es6 import of the language files.
Related issue: #5092

### How to use:
```es6
import es from 'video.js/dist/lang/es.json';

videojs(element, {
  language: 'es',
  languages: {
    es
  }
});
```

If you use typescript, you need to add the following compilerOptions to the tsconfig:
```json
"resolveJsonModule": true,
"esModuleInterop": true,
"allowSyntheticDefaultImports": true
```

## Specific Changes proposed
Add a grunt task to copy all json files to dist/lang directory.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/build/grunt.js
+++ b/build/grunt.js
@@ -31,6 +31,7 @@ module.exports = function(grunt) {
     },
     copy: {
       fonts: { cwd: 'node_modules/videojs-font/fonts/', src: ['*'], dest: 'build/temp/font/', expand: true, filter: 'isFile' },
+      lang:  { cwd: 'lang/', src: ['*'], dest: 'dist/lang/', expand: true, filter: 'isFile' },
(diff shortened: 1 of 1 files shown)


gkatsev:
This may be a reasonable workaround. Is importing JSON something that's actively supported? Or is that why you mention the need to enable it in TS?

eranshmil:
It's supported.
Tested using this boilerplate: https://github.com/metagrover/ES6-boilerplate

eranshmil:
This is the quickest patch, the longer one is to create another directory with module.exports js files.
I guess I can make a patch directly in the grunt-videojs-languages package.
This is also not the best approach, but it's also won't take a lot of time.

What do you think?

gkatsev:
Yeah, I think this is fine.

gkatsev:
Going to merge even though the tests are failing as it's unrelated and may not be an issue in master. We're still investigating why the tests are failing.

PR #5555, opened 2018-11-06
chore: update deps, remove coveralls, fix audit issues

gkatsev:
Also, run `npm audit`. Then remove grunt-coveralls because it isn't used and coverage will be added separately. The last audit issue is waiting on https://github.com/karma-runner/karma-browserstack-launcher/issues/135
replace was updated to 1.0 when ownership changed and security issues were fixed.


package.json
```diff
@@ -89,7 +89,7 @@
     "conventional-changelog-cli": "^2.0.11",
     "conventional-changelog-videojs": "^3.0.0",
     "es5-shim": "^4.5.10",
```
BrandonOCasey: I have been wondering this, can we remove es5-shim at this point?

gkatsev: Maybe, it doesn't hurt to keep it in. It's only used for tests and helps make sure everything is up-to-spec. Since we're loading in es6-shim and es7-shim, it's probably best to keep es5-shim as well.

PR #5556, opened 2018-11-06
fix(tracks): don't select tracks based on user pref if no langauge is set

### LINKED ISSUE #5553: Segments metadata track is shown when changing source
## Description
If you had captions on, and then change the video source, you get some weird text on the screen. The guys on videojs slack thought it was the segments metadata track.

## Steps to reproduce
1. Visit the VHS test page at https://videojs.github.io/http-streaming/
2. Play the video, choosing the very bottom closed caption track, "cc1". (Not sure what language it is for, but the captions show English)
3. Reset the player's source by clicking the "Load" button again.
4. Play the video, and see the weird text/json

### Expected
No captions after changing source.

### Actual
Even though closed captions are set to "off" after switching the source, this appears as a caption over the video:
`{"bandwidth":915905,"resolution":{"width":960,"height":540},"codecs":"mpra.40.2,avc1.4d401f","byteLength":1046972,"uri":"https://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/gear3/main.ts","timeline":0,"playlist":"gear3/prog_index.m3u8","start":9.976644444444446,"end":19.91992222222223}`

#### videojs
I've seen it in 7.2.2 and 7.2.4
#### browsers
Chrome, Firefox, and IE11. 
Edge doesn't even show captions.
Mobile Safari doesn't show that bottom "CC" caption option, so it doesn't seem to have the problem.
#### OSes
Microsoft Windows 10
### plugins
None

gkatsev:
Fixes #5553

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -150,8 +150,11 @@ class TextTrackDisplay extends Component {
     for (let i = 0; i < trackList.length; i++) {
       const track = trackList[i];
 
-      if (userPref && userPref.enabled &&
-        userPref.language === track.language) {
+      if (
+        userPref && userPref.enabled &&
+        userPref.language && userPref.language === track.language &&
+        track.kind in modes
+      ) {
         // Always choose the track that matches both language and kind
         if (track.kind === userPref.kind) {
           preferredTrack = track;
--- a/test/unit/tracks/text-track-display.test.js
+++ b/test/unit/tracks/text-track-display.test.js
@@ -183,6 +183,50 @@ if (!Html5.supportsNativeTextTracks()) {
     player.dispose();
   });
 
+  QUnit.test("don't select user langauge if it is an empty string", function(assert) {
+    const player = TestHelpers.makePlayer();
+    const track1 = {
+      kind: 'captions',
+      label: 'English',
+      language: 'en',
+      src: 'en.vtt'
+    };
+    const track2 = {
+      kind: 'captions',
+      label: 'Spanish',
+      language: 'es',
+      src: 'es.vtt'
+    };
+    const track3 = {
+      kind: 'metadata',
+      label: 'segment-metadata'
+    };
+
+    player.src({type: 'video/mp4', src: 'http://google.com'});
+    // manualCleanUp = true by default
+    const englishTrack = player.addRemoteTextTrack(track1, true).track;
+    const spanishTrack = player.addRemoteTextTrack(track2, true).track;
+    const metadataTrack = player.addRemoteTextTrack(track3, true).track;
+
+    // Force empty string ('') as "user-selected" track
+    player.cache_.selectedLanguage = { enabled: true, language: '', kind: 'captions' };
+    this.clock.tick(1);
+
+    assert.equal(spanishTrack.mode, 'disabled', 'Spanish captions should be disabled');
+    assert.equal(englishTrack.mode, 'disabled', 'English captions should be disabled');
+    assert.notEqual(metadataTrack.mode, 'showing', 'Metadata track should be not be showing');
+
+    // Force es as "user-selected" track
+    player.cache_.selectedLanguage = { enabled: true, language: 'es', kind: 'captions' };
+    player.trigger('loadstart');
+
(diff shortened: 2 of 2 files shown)


src/js/tracks/text-track-display.js
```diff
@@ -150,8 +150,11 @@ class TextTrackDisplay extends Component {
     for (let i = 0; i < trackList.length; i++) {
       const track = trackList[i];
 
-      if (userPref && userPref.enabled &&
-        userPref.language === track.language) {
+      if (
+        userPref && userPref.enabled &&
+        userPref.language && userPref.language === track.language &&
+        track.kind in modes
```
ldayananda: 💯

test/unit/tracks/text-track-display.test.js
```diff
@@ -183,6 +183,50 @@ if (!Html5.supportsNativeTextTracks()) {
...
+    const spanishTrack = player.addRemoteTextTrack(track2, true).track;
+    const metadataTrack = player.addRemoteTextTrack(track3, true).track;
+
+    // Force empty string ('') as "user-selected" track
+    player.cache_.selectedLanguage = { enabled: true, language: '', kind: 'captions' };
+    this.clock.tick(1);
+
+    assert.equal(spanishTrack.mode, 'disabled', 'Spanish captions should be disabled');
+    assert.equal(englishTrack.mode, 'disabled', 'English captions should be disabled');
+    assert.notEqual(metadataTrack.mode, 'showing', 'Metadata track should be not be showing');
```
ldayananda: should not be showing? 😄

test/unit/tracks/text-track-display.test.js
```diff
@@ -183,6 +183,50 @@ if (!Html5.supportsNativeTextTracks()) {
...
+      src: 'es.vtt'
+    };
+    const track3 = {
+      kind: 'metadata',
+      label: 'segment-metadata'
+    };
+
+    player.src({type: 'video/mp4', src: 'http://google.com'});
+    // manualCleanUp = true by default
+    const englishTrack = player.addRemoteTextTrack(track1, true).track;
```
ldayananda: might want to add the metadata track first just to be extra sure we're fixing the issue

test/unit/tracks/text-track-display.test.js
```diff
@@ -183,6 +183,50 @@ if (!Html5.supportsNativeTextTracks()) {
...
+    const spanishTrack = player.addRemoteTextTrack(track2, true).track;
+    const metadataTrack = player.addRemoteTextTrack(track3, true).track;
+
+    // Force empty string ('') as "user-selected" track
+    player.cache_.selectedLanguage = { enabled: true, language: '', kind: 'captions' };
+    this.clock.tick(1);
```
ldayananda: should loadstart have been triggered here too?

gkatsev: this tick is to make the initial loadstart from setting the source run

test/unit/tracks/text-track-display.test.js
gkatsev: it shouldn't matter the order they are added, especially since we're waiting till loadstart (or the tick) to do the work. Plus, we can see it fail and then succeed in the builds.

test/unit/tracks/text-track-display.test.js
```diff
@@ -183,6 +183,50 @@ if (!Html5.supportsNativeTextTracks()) {
...
+
+    assert.equal(spanishTrack.mode, 'disabled', 'Spanish captions should be disabled');
+    assert.equal(englishTrack.mode, 'disabled', 'English captions should be disabled');
+    assert.notEqual(metadataTrack.mode, 'showing', 'Metadata track should not be showing');
```
squarebracket: Is there a reason you're opting for `notEqual`'ing `showing` instead of `equal`'ing `hidden`? Like Safari's native HLS support or something?

gkatsev: Yeah, I figured it'll be safer to just make sure it isn't showing since it doesn't really matter what this is set to for this test

Issue #4048, opened 2017-02-09
Add support for "bubbling" unused keypresses up through control bar components to the player


OwenEdwards:
We've been working hard on keyboard accessibility for the video.js control bar, using Tab/Shift-Tab, Enter, Space, and the arrow keys. Each component knows what key presses it supports, and which ones it ignores. There's a basic structure in place to pass keypresses up through the component hierarchy to the base `Component`, but it would make sense to then pass them to the `parent` Component, and onwards up to the player itself.

That way, we could either implement hotkeys ourselves later on, or provide a single point to attach a plug-in to handle keypresses which aren't used by the controls themselves.

Thoughts?

petercunha:
Has anyone followed up on this issue? I think hotkey support would be a very valuable feature.

I might try to work on this myself and make a pull request if nothing is in the works yet.

OwenEdwards:
@petercunha I've got a project on the back burner to implement this; I'll do a little more work on it and create a PR, so you can at least take a look at it. Give me a day or two.

Issue #3022, opened 2016-01-22
Make or use a module for resolving keycodes


gkatsev:
In a lot of places in our code, we do checks against `event.which` and then having a comment saying that `9` is a tab and `13` is enter.
We should investigate the [keycode](https://github.com/timoxley/keycode) module and decide whether we need a whole module or if making a simple object map in our code is sufficient or worth it.
Afterwards, we should update the code that checks `event.which` to use what we chose above instead of magic numbers with comments.

rizcheldayao:
@gkatsev can I take on this issue? Is it still needed?

gkatsev:
I think it would be nice to do for readability and maintenance, though, the main thing I'd be worried about is that the size of the player increases too much if we use an existing library.

PR #5571, opened 2018-11-09
Extend SeekBar keyboard handling, and support hotkeys "m", "f", "k" and Space

OwenEdwards:
## Description
This PR does two things; it extends the keyboard handling for the seek bar to support a lot more keys (similar to YouTube), and then it passes all unused key presses from UI components to the player, <del>so that a subsequent PR (or plugin) can perform</del> to support player-wide keyboard shortcuts "m" (mute toggle), "f" (full-screen toggle), and "k" (play/pause toggle).

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/docs/guides/options.md
+++ b/docs/guides/options.md
@@ -37,6 +37,9 @@
   * [techOrder](#techorder)
   * [userActions](#useractions)
   * [userActions.hotkeys](#useractions.hotkeys)
+  * [userActions.hotkeys.fullscreenKey](#useractions.hotkeys.fullscreenkey)
+  * [userActions.hotkeys.muteKey](#useractions.hotkeys.mutekey)
+  * [userActions.hotkeys.playPauseKey](#useractions.hotkeys.playpausekey)
   * [vtt.js](#vttjs)
 * [Component Options](#component-options)
   * [children](#children-1)
@@ -367,7 +370,7 @@ This can be useful when multiple techs are used and each has to set their own po
 
 > Type: `Array`, Default: `['html5']`
 
-Defines the order in which Video.js techs are preferred. By default, this means that the `Html5` tech is preferred. Other regisetered techs will be added after this tech in the order in which they are registered.
+Defines the order in which Video.js techs are preferred. By default, this means that the `Html5` tech is preferred. Other registered techs will be added after this tech in the order in which they are registered.
 
 ### `userActions`
 
@@ -375,25 +378,25 @@ Defines the order in which Video.js techs are preferred. By default, this means 
 
 ### `userActions.hotkeys`
 
-> Type: `boolean|function`
+> Type: `boolean|function|object`
 
-Controls how player-wide hotkeys operate. If set to `false`, or `undefined`, hotkeys are disabled. If set to
-`true`, hotkeys are enabled as described below. To override the default hotkey handling, set `userActions.hotkeys`
-to a function which accepts a `keydown` event:
+Controls how player-wide hotkeys operate. If set to `false`, or `undefined`, hotkeys are disabled. If set to `true` or an object (to allow definitions of `fullscreenKey` etc. below), hotkeys are enabled as described below. To override the default hotkey handling, set `userActions.hotkeys` to a function which accepts a `keydown` event:
 
 ```js
-function myHandleHotkeys(event) = {
-  // `this` is the player in this context
-
-  // `p` key = pause
-  if (event.which === 80) {
-    this.pause();
-  }
-};
-
-videojs('my-player', {
+var player = videojs('my-player', {
   userActions: {
-    hotkeys: myHandleHotkeys
+    hotkeys: function(event) {
+      // `this` is the player in this context
+
+      // `x` key = pause
+      if (event.which === 88) {
+        this.pause();
+      }
+      // `y` key = play
+      if (event.which === 89) {
+        this.play();
+      }
+    }
   }
 });
 ```
@@ -409,6 +412,40 @@ Default hotkey handling is:
 
 Note that the `Space` key activates controls such as buttons and menus if that control has keyboard focus. The other hotkeys work regardless of which
 control in the player has focus.
+
+### `userActions.hotkeys.fullscreenKey`
+
+> Type: `function`
+
+Override the fullscreen key definition. If this is set, the function receives the `keydown` event; if the function returns `true`, then the fullscreen toggle action is performed.
+
+```js
+var player = videojs('my-player', {
+  userActions: {
+    hotkeys: {
+      muteKey: function(event) {
+        // disable mute key
+      },
+      fullscreenKey: function(event) {
+        // override fullscreen to trigger when pressing the v key
+        return (event.which === 86);
+      }
+    }
+  }
+});
+```
+
+### `userActions.hotkeys.muteKey`
+
+> Type: `function`
+
+Override the mute key definition. If this is set, the function receives the `keydown` event; if the function returns `true`, then the mute toggle action is performed.
+
+### `userActions.hotkeys.playPauseKey`
+
+> Type: `function`
+
+Override the play/pause key definition. If this is set, the function receives the `keydown` event; if the function returns `true`, then the play/pause toggle action is performed.
 
 ### `vtt.js`
 
--- a/docs/guides/options.md
+++ b/docs/guides/options.md
@@ -35,6 +35,8 @@
   * [sources](#sources)
   * [techCanOverridePoster](#techcanoverrideposter)
   * [techOrder](#techorder)
+  * [userActions](#useractions)
+  * [userActions.hotkeys](#useractions.hotkeys)
   * [vtt.js](#vttjs)
 * [Component Options](#component-options)
   * [children](#children-1)
@@ -366,6 +368,47 @@ This can be useful when multiple techs are used and each has to set their own po
 > Type: `Array`, Default: `['html5']`
 
 Defines the order in which Video.js techs are preferred. By default, this means that the `Html5` tech is preferred. Other regisetered techs will be added after this tech in the order in which they are registered.
+
+### `userActions`
+
+> Type: `Object`
+
+### `userActions.hotkeys`
+
+> Type: `boolean|function`
+
+Controls how player-wide hotkeys operate. If set to `false`, or `undefined`, hotkeys are disabled. If set to
+`true`, hotkeys are enabled as described below. To override the default hotkey handling, set `userActions.hotkeys`
+to a function which accepts a `keydown` event:
+
+```js
+function myHandleHotkeys(event) = {
+  // `this` is the player in this context
+
+  // `p` key = pause
+  if (event.which === 80) {
+    this.pause();
+  }
+};
+
+videojs('my-player', {
+  userActions: {
+    hotkeys: myHandleHotkeys
+  }
+});
+```
+
+Default hotkey handling is:
+
+| Key | Action | Enabled by |
+| :-: | ------ | ---------- |
+| `f` | toggle fullscreen | only enabled if a Fullscreen button is present in the Control Bar
+| `m` | toggle mute | always enabled, even if no Control Bar is present
+| `k` | toggle play/pause | always enabled, even if no Control Bar is present
+| `Space` | toggle play/pause | always enabled, even if no Control Bar is present
+
+Note that the `Space` key activates controls such as buttons and menus if that control has keyboard focus. The other hotkeys work regardless of which
+control in the player has focus.
 
 ### `vtt.js`
 
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@babel/runtime": "^7.2.0",
     "@videojs/http-streaming": "1.8.0",
     "global": "4.3.2",
+    "keyboard-key": "^1.0.2",
     "safe-json-parse": "4.0.0",
     "tsml": "1.0.1",
     "videojs-font": "3.1.0",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "@babel/runtime": "^7.2.0",
     "@videojs/http-streaming": "1.8.0",
     "global": "4.3.2",
-    "keyboard-key": "^1.0.2",
+    "keycode": "^2.2.0",
     "safe-json-parse": "4.0.0",
     "tsml": "1.0.1",
     "videojs-font": "3.1.0",
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1078,6 +1078,19 @@ class Component {
   }
 
   /**
+   * When this Component receives a keydown event which it does not process,
+   *  it passes the event to the Player for handling.
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that caused this function to be called.
+   */
+  handleKeyPress(event) {
+    if (this.player_) {
+      this.player_.handleKeyPress(event);
+    }
+  }
+
+  /**
    * Emit a 'tap' events when touch event support gets detected. This gets used to
    * support toggling the controls through a tap on the video. They get enabled
    * because every sub-component would have extra overhead otherwise.
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -1,4 +1,4 @@
-*/**
+/**
  * @file seek-bar.js
  */
 import Slider from '../../slider/slider.js';
@@ -401,7 +401,7 @@ class SeekBar extends Slider {
     } else if (keyCode === keyboardKey.PageDown) {
       event.preventDefault();
       this.player_.currentTime(this.player_.currentTime() - (STEP_SECONDS * PAGE_KEY_MULTIPLIER));
-    } else if (keyCode ===keyboardKey.PageUp) {
+    } else if (keyCode === keyboardKey.PageUp) {
       event.preventDefault();
       this.player_.currentTime(this.player_.currentTime() + (STEP_SECONDS * PAGE_KEY_MULTIPLIER));
     } else {
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -1,4 +1,4 @@
-/**
+*/**
  * @file seek-bar.js
  */
 import Slider from '../../slider/slider.js';
@@ -8,6 +8,7 @@ import * as Dom from '../../utils/dom.js';
 import * as Fn from '../../utils/fn.js';
 import formatTime from '../../utils/format-time.js';
 import {silencePromise} from '../../utils/promise';
+import keyboardKey from 'keyboard-key';
 
 import './load-progress-bar.js';
 import './play-progress-bar.js';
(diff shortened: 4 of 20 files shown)


gkatsev:
Yeah, I'm not sure what's happening. It seems to be fine in master, mostly. We should investigate the sourceset tests on firefox but it seems to be passing locally which makes it a lot harder to track down, unfortunately.

OwenEdwards:
@gkatsev - also, any idea about these warnings? These ones I do get locally too, but I'm not sure what the issue is:

````
Running "shell:rollupall" (shell) task
> video.js@7.3.0 rollup /home/travis/build/videojs/video.js
> rollup -c "--no-progress"

src/js/index.js → dist/video.js...
created dist/video.js in 10.3s

src/js/index.js → dist/video.es.js, dist/video.cjs.js...
'keyboard-key' is imported by src/js/modal-dialog.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/clickable-component.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/button.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/control-bar/progress-control/seek-bar.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/menu/menu-button.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/menu/menu.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/slider/slider.js, but could not be resolved – treating it as an external dependency
'keyboard-key' is imported by src/js/menu/menu-item.js, but could not be resolved – treating it as an external dependency
created dist/video.es.js, dist/video.cjs.js in 5.3s
````

**Comment by gkatsev** [comment]
That's from the addition of the keyboard-key dep, it should be added to the appropriate place in the rollup config. I'll look it up when reviewing this PR. Should be some time today.

File: src/js/component.js [hunk 1]
```diff
@@ -1066,6 +1066,19 @@ class Component {
...
   }
 
+  /**
+   * When this Component receives a keydown event which it does not process,
+   *  it passes the event to the Player for handling.
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that caused this function to be called.
+   */
+  handleKeyPress(event) {
```
gkatsev: so, basically every component calls `super.handleKeyPress` which eventually gets here, which in turn calls `handleKeyPress` on the player itself?

src/js/menu/menu-item.js
```diff
@@ -68,6 +69,26 @@ class MenuItem extends ClickableComponent {
...
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that caused this function to be called.
+   *
+   * @listens keydown
+   */
+  handleKeyPress(event) {
+    const keyCode = keyboardKey.getCode(event);
+
+    if (keyCode !== keyboardKey.Escape && keyCode !== keyboardKey.Tab &&
```
gkatsev: These are all keycodes that `menu.js` uses? Is there a way to centralize this so that if we modify menu.js we don't break things by forgetting to add stuff in here?

gkatsev: Might better to do something like:
```js
if ([keyboardKey.Escape, keyboardKey.Tab /* ... */].indexOf(keyCode) !== -1) {
  super.handleKeyPress(event);
}
```

src/js/player.js
```diff
@@ -2664,6 +2664,16 @@ class Player extends Component {
...
   }
 
+  /**
+   * Handle keydown events from any Component of this player which is not
+   *  andled by that Component; allows for player-wide keyboard shortcuts
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that caused this function to be called.
+   */
+  handleKeyPress(event) {
```
gkatsev: this is empty for now but eventually we can add other things like `f` going to fullscreen?

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -308,22 +312,45 @@ class SeekBar extends Slider {
   }
 
   /**
-   * Called when this SeekBar has focus and a key gets pressed down. By
-   * default it will call `this.handleAction` when the key is space or enter.
+   * Called when this SeekBar has focus and a key gets pressed down.
+   * Supports the following keys:
+   *
+   *   Space or Enter key fire a click event
```
gkatsev: should these extra functions be on the slider itself? similar to how arrow keys call `stepBack` or `stepForward`? Would this functionality be helpful when focused on the volume slider as well?

gkatsev:
Also, to resolve the rollup warnings, add `keyboard-key` to this array: https://github.com/videojs/video.js/blob/209d9f92d72b75fbc5992964a27fd91fd2e71344/rollup.config.js#L69-L85

src/js/player.js
OwenEdwards: Yes. See https://github.com/videojs/video.js/issues/4048#issuecomment-437499539; I have a separate PR which builds on top of this one to do exactly that, but wasn't sure if you'd want it in the core or in a plugin.

src/js/control-bar/progress-control/seek-bar.js
OwenEdwards: Support for the additional keys is not expected on the volume slider; there's already a mute button to do much of what the additional keys here do, and it's only 10 presses of Up Arrow or Down Arrow to move full-range on the volume control. I *guess* that support for 0..9 might be nice, but it's not available on any general purpose sliders, nor or YouTube's volume control.

src/js/component.js
OwenEdwards: Right.

package.json
```diff
@@ -73,6 +73,7 @@
     "@babel/runtime": "^7.0.0",
     "@videojs/http-streaming": "1.4.2",
     "global": "4.3.2",
+    "keyboard-key": "^1.0.2",
```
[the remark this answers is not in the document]
OwenEdwards: No specific experience with this module. Yeah, https://www.npmjs.com/package/keycode looks like a viable alternative (I'd forgotten that you'd mentioned it in the previous issue), I just went with this one because it made sense.

Do you want me to try that other one?

src/js/control-bar/progress-control/seek-bar.js
gkatsev: Sounds good

package.json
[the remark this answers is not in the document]
gkatsev: Yeah, maybe. Would be good to see if keycode provides everything we need and how the filesize difference affects Video.js bundles.

src/js/player.js
gkatsev: I think the hotkeys plugin is pretty popular, it's probably something we want included. Would be nice to figure out how to make it pluggable so that it can be extended with custom shortcuts as well. Though, that can happen down the line.

OwenEdwards:
@gkatsev take a look at the latest push - it uses the `keycode` package you mentioned in #3022. I've made some of the other changes you suggested, but some are harder than I thought. This seems like a good first step, if you prefer the size of this package.

src/js/menu/menu.js
```diff
@@ -112,7 +113,7 @@ class Menu extends Component {
 
     // Close menu popup when a user clicks outside the menu
     if (!this.children().some((element) => {
-      return element.el() === relatedTarget;
+      return element.el && (element.el() === relatedTarget);
```
OwenEdwards: This is actually a bugfix for a previous change, not specifically related to this PR; it shows up with the Chapters menu where the first child element isn't a video.js component, it's just an HTML `<li>` element which doesn't have a `.el()` method.

src/js/menu/menu-item.js
OwenEdwards: Done in the latest change

OwenEdwards: Done - see the latest change.

OwenEdwards:
@gkatsev can you take another look at this?

OwenEdwards:
@gkatsev the Travis CI build is still failing after I merged the latest master branch changes; the error is the same as before. Can you take a look?

```
Firefox 63.0.0 (Linux 0.0.0) sourceset > source before player > video el data-setup one source FAILED
	Test took longer than 5000ms; test timed out.
	internalStop/config.timeout<@node_modules/qunit/qunit/qunit.js:2001:7
	
	
	no additional sourcesets
	Expected: 1
	Actual: 0
	setupAfterEach/</<@/tmp/test/unit/sourceset.test.js:100:6 <- /tmp/9650fad653464ed42db5b0d2e985fafa.browserify:49572:7
	
	
	no source set on dispose
	Expected: 1
	Actual: 0
	setupAfterEach/</<@/tmp/test/unit/sourceset.test.js:103:6 <- /tmp/9650fad653464ed42db5b0d2e985fafa.browserify:49576:7
```

gkatsev:
Also, I'm really happy this along with the keycode change is happening. Thanks!

src/js/menu/menu-button.js
```diff
@@ -281,24 +282,26 @@ class MenuButton extends Component {
...
+    // Up Arrow or Down Arrow also 'press' the button to open the menu
+    } else if (keycode.isEventKey(event, 'Up') || keycode.isEventKey(event, 'Down')) {
       if (!this.buttonPressed_) {
-        this.pressButton();
         event.preventDefault();
+        this.pressButton();
       }
+    } else {
+      // NOTE: This is a special case where we don't pass unhandled
+      //  keypress events up to the Component handler
```
BrandonOCasey: Would be good to add a reason in this comment, as well and the one below it in `handleSubmenuKeyPress`

src/js/menu/menu-item.js
```diff
@@ -68,6 +70,24 @@ class MenuItem extends ClickableComponent {
...
+   * Ignore keys which are used by the menu, but pass any other ones up. See
+   * {@link ClickableComponent#handleKeyPress} for instances where this is called.
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that caused this function to be called.
+   *
+   * @listens keydown
+   */
+  handleKeyPress(event) {
+    if (!MenuKeys.some((key) => {
```
BrandonOCasey: You should be able to do this all on one line a bit cleaner with something like this: 
```suggestion
    if (!MenuKeys.some((key) => keycode.isEventKey(event, key)) {
```

src/js/menu/menu.js
```diff
@@ -112,7 +113,7 @@ class Menu extends Component {
 
     // Close menu popup when a user clicks outside the menu
     if (!this.children().some((element) => {
```
BrandonOCasey: This could be one line as well
```suggestion
    if (!this.children().some((element) => element.el && element.el() === relatedTarget)) {
```

src/js/menu/menu.js
```diff
@@ -132,14 +133,20 @@ class Menu extends Component {
...
     // Up and Right Arrows
-    } else if (event.which === 38 || event.which === 39) {
+    } else if (keycode.isEventKey(event, 'Right') || keycode.isEventKey(event, 'Up')) {
       event.preventDefault();
       this.stepBack();
+    } else {
+      // NOTE: This is a special case where we don't pass unhandled
+      //  keypress events up to the Component handler, because this
+      //  is just adding a keypress handler on top of the MenuItem's
+      //  existing keypress handler, which handles that passing keypress
```
BrandonOCasey:
```suggestion
      //  existing keypress handler, which already handles passing keypress
```

src/js/modal-dialog.js
```diff
@@ -119,14 +119,18 @@ class ModalDialog extends Component {
...
    * @listens keydown
    */
-  handleKeyPress(e) {
-    if (e.which === ESC && this.closeable()) {
+  handleKeyPress(event) {
+    if (keycode.isEventKey(event, 'Escape') && this.closeable()) {
       this.close();
+    } else {
+
+      // Pass keypress handling up for unsupported keys
```
BrandonOCasey: Do we want to pass keyboard events to the player while the modal is over the player?

src/js/player.js
```diff
@@ -2672,6 +2673,62 @@ class Player extends Component {
...
+  /**
+   * Handle a keydown event from any Component of this player which is not
+   *  handled by that Component to support player-wide hotkeys.
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that was passed from a Component.
+   */
+  handleKeyPress(event) {
+
+    if (this.options_.hotkeys !== false) {
```
BrandonOCasey: We should do this check at the start of the function and return, that way we can eliminate this scope for the whole function.

BrandonOCasey: It would also be good to change this to the option named in the `dblclick` pull request: https://github.com/videojs/video.js/pull/5611#issuecomment-443266964 Although that could be done in a separate pull request if that works out better.

src/js/player.js
```diff
@@ -2672,6 +2673,62 @@ class Player extends Component {
...
+
+        this.options_.hotkeys(event);
+
+      // Trigger the ControlBar equivalent for each key, if possible
+      } else if (keycode.isEventKey(event, 'f')) {
+        // "f" = toggle fullscreen
+
+        event.preventDefault();
+
+        const cb = this.player_.controlBar;
```
BrandonOCasey: Should we also define, and check for this at the top of the function? That would prevent some of the code duplication in each if statement.

src/js/utils/events.js
```diff
@@ -181,6 +181,7 @@ export function fixEvent(event) {
         (doc && doc.clientTop || body && body.clientTop || 0);
     }
 
+    // NOTE: this may no longer be needed since the adoption of the 'keycode' package
```
BrandonOCasey: It could be good to make this a TODO comment to evaluate if we still need it. It's possible we do for backwards compatibility

BrandonOCasey:
Just realized that this should have doc updates as well.

src/js/utils/events.js
gkatsev: Yup, we probably need to keep this for backwards compatibility.

BrandonOCasey:
Just realized that this should have doc updates as well.

src/js/modal-dialog.js
OwenEdwards: Good question! If not, then we need to change the CloseButton as well, because it extends Button which *does* pass keyboard events up to the player.

But yes, now I think about it, the whole point of a "modal" is that it makes the thing behind it (usually the whole window, but in this case just the player, *inert*. I'll change that, and the CloseButton.

src/js/utils/events.js
OwenEdwards: So should I just drop the note altogether, or note that it *may* be unnecessary but is retained for backward compatibility?

src/js/player.js
OwenEdwards: @gkatsev I'm going to leave it as it is for now; having a method per key can be a separate PR.

src/js/utils/events.js
gkatsev: yeah, we can just remove this NOTE

src/js/player.js
gkatsev: Sounds good, though, we should make sure on the direction we want to go with it so we don't ship with it as is if we wanted to have it per key.

OwenEdwards:
@gkatsev I think the only outstanding issue now is "_should each individual key have its own method or have a general hotkeys one?_"

BrandonOCasey:
I think it might make more sense for each key to have its own method/option, but I don't feel strongly about it either way.

OwenEdwards:
@gkatsev & @BrandonOCasey - the latest push adds an option for each defined hotkey, much like the hotkeys plugin, as well as allowing complete override of built-in hotkey handling. Take a look (I included documentation in `options.md`, and an example in `sandbox/hotkeys.html.example`). Let me know if this needs any more changes; I'd like to get it wrapped up!

docs/guides/options.md
```diff
@@ -365,7 +370,82 @@ This can be useful when multiple techs are used and each has to set their own po
...
+```js
+var player = videojs('my-player', {
+  userActions: {
+    hotkeys: {
+      muteKey: function(event) {
+        // disable mute key
+      },
+      fullscreenKey: function(event) {
+        // override fullscreen to trigger when pressing the v key
+        return (event.which === 86);
```
gkatsev: I'm not sure it's better, but here's a thought:
the event method has a `preventDefault` function on it, when called, it sets the `defaultPrevented` flag on the event. We could then check that flag to see whether we do the default action or not.

gkatsev: Also, should the double click PR be similar? Where if the handler returns true we do our default action?

src/js/player.js
```diff
@@ -2679,57 +2679,138 @@ class Player extends Component {
...
-        } else {
-          this.player_.muted(this.player_.muted() ? false : true);
-        }
-      } else if (keycode.isEventKey(event, 'k')) {
-        // "k" = toggle play/pause
+    // Trigger the ControlBar equivalent for each key, if possible
+    const cb = this.controlBar;
 
-        event.preventDefault();
+    if (fullscreenKey.call(this, event)) {
```
gkatsev: if someone overrides the fullscreen key with a method that returns false, this will try to evaluate the other methods, though, maybe not a big deal.

src/js/player.js
```diff
@@ -2679,57 +2679,138 @@ class Player extends Component {
...
-        if (playToggle && playToggle.handleClick) {
-          playToggle.handleClick();
-        }
+      const fullscreenToggle = cb && cb.fullscreenToggle;
+
+      // TODO: make sure that this hotkey doesn't "toggle fullscreen" if
+      //  fullscreen is disabled. For now, ONLY allow fullscreen if the
+      //  fullscreen button is present in the ControlBar.
+
+      if (fullscreenToggle && fullscreenToggle.handleClick) {
```
gkatsev: would be nice to make these not depend on the control bar components as they can be disabled.

gkatsev: We can do it in a separate PR, as long as it's done before we make a release (I might make a PR against this PR for it)

src/js/player.js
```diff
@@ -2679,57 +2679,138 @@ class Player extends Component {
...
+   *
+   *   f          - toggle fullscreen
+   *   m          - toggle mute
+   *   k or Space - toggle play/pause
+   *
+   * @param {EventTarget~Event} event
+   *        The `keydown` event that caused this function to be called.
+   */
+  handleHotkeys(event) {
+    const hotkeys = this.options_.userActions ? this.options_.userActions.hotkeys : null;
```
gkatsev: if you default to `{}`, you could then do something like:
```js
let {fullscreenKey, muteKey, playPauseKey} = hotkeys;
```
saving lines 2752-2756

src/js/player.js
```diff
@@ -2679,57 +2679,138 @@ class Player extends Component {
...
+    if (hotkeys) {
+      fullscreenKey = hotkeys.fullscreenKey;
+      muteKey = hotkeys.muteKey;
+      playPauseKey = hotkeys.playPauseKey;
+    }
 
-        const cb = this.player_.controlBar;
-        const volumePanel = cb && cb.volumePanel;
-        const muteToggle = (cb && cb.muteToggle) || (volumePanel && volumePanel.muteToggle);
+    if (typeof fullscreenKey !== 'function') {
```
gkatsev: should we respect a falsy value here to disable that hotkey?

OwenEdwards: Maybe in the future, but for now the `options.userActions.hotkeys` is one type of override (the same as `userActions.doubleClick`), that can take `true`, `false` or an override `function`, whereas `fullscreenKey` is modeled after the option in https://github.com/ctd1500/videojs-hotkeys, which is *only* a function.

I realize that the `options.userActions` got pretty complicated with essentially two different kinds of overrides; I did my best to document it in `options.md`, but I can cut some of it out if it makes more sense?

gkatsev:
Made a PR with the changes mentioned above: https://github.com/OwenEdwards/video.js/pull/2

PR #5576, opened 2018-11-13
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
## Description

## Specific Changes proposed
`lts/*` uses only the last LTS branch but not the others. Currently Node.js 6, 8 and 10 are LTS. See https://github.com/nodejs/Release/blob/master/README.mdy

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: node_js
 node_js:
-  - lts/*
+  - 'node'
+  - '10'
+  - '8'
+  - '6'
 before_install:
   - npm install -g greenkeeper-lockfile@1
   - export CHROME_BIN=/usr/bin/google-chrome


gkatsev:
Thanks for the PR, however, because we test in browsers, we don't need to run against multiple versions of node.

DanielRuf:
What about the dependencies and contributors and Node.js projects which require/import video.js and bundle files using webpack?

DanielRuf:
See https://www.npmjs.com/package/video.js so there are definitely people who consume video.js using Node.js and we for example require the source files and build them ourself for treeshaking, different (vendor) bundles and so on.

DanielRuf:
Which is not relevant for devs like me who bundle it themself with webpack and babelify and treeshake the code (as we often do with frontend frameworks).

DanielRuf:
Are consumers allowed to use Node.js or should it be set to `node >= 8` in the `engines` field of package.json?

Just asking to also make sure that we can move forward with the required Node.js version for contributors and video.js building itself.

gkatsev:
There's no need to babelify Video.js, fwiw, as we compile our code down to ES5. We also bundle all the library code into a single file via rollup. We have .es.js for esmodules and import statements, .cjs.js for commonjs and require statements. Those two files don't have dependencies included.

If you require Video.js directly in node.js as opposed to shipping it to the web, you're basically going to get a function that does nothing. This can be useful if you're doing server-side rendering and then rehydrate client-side. So you want to import Video.js but have it basically do nothing, which is what it does.

The version of Node is mostly relevant just for building and developing Video.js. Our [nvmrc file](https://github.com/videojs/video.js/blob/master/.nvmrc) (and I guess the travis.yml) serves as documentation of which node version to use.

PR #5589, opened 2018-11-14
chore: Move a11y, lang, browserify, and webpack out of grunt

BrandonOCasey:
Moves language, a11y, browserify, and webpack tasks out of grunt.

After this pull request, I think we can have two more pull requests:
1. That removes grunt and migrates the meta tasks over (`dist`, `test`, `build`, `watchAll`)
2. A final pull request that cleans all of the package.json scripts up, making sure it still does everything we want, but in less scripts/code.


package.json
```diff
@@ -17,11 +17,17 @@
...
+    "build:browserify-test": "browserify test/require/browserify.js -o build/temp/browserify.js",
+    "build:webpack-test": "webpack --hide-modules test/require/webpack.js build/temp/webpack.js",
     "copy-fonts": "shx cp -R node_modules/videojs-font/fonts build/temp/font",
     "copy-dist": "shx cp -R build/temp/* dist/",
-    "copy-a11y": "shx cp sandbox/descriptions.html.example sandbox/descriptions.html",
     "copy-examples": "shx cp -R docs/examples dist/",
-    "copy-lang": "shx cp -R lang/* dist/lang/",
+    "build:lang": "npm-run-all build:lang:*",
+    "build:lang:js": "vjslang --dir dist/lang",
+    "build:lang:copy": "shx --verbose cp -R lang/* dist/lang/",
```
gkatsev: why `--verbose`?

package.json
```diff
@@ -17,11 +17,17 @@
...
+    "pretest:a11y": "shx cp sandbox/descriptions.html.example sandbox/descriptions.test-a11y.html",
```
gkatsev: at some point in the future, this may want to check if the file already exists, this way you could run the a11y test with local changes, if we had any. Definitely nothing to do right now, though.

package.json
```diff
@@ -17,11 +17,17 @@
   "homepage": "https://videojs.com",
   "author": "Steve Heffernan",
   "scripts": {
```
gkatsev: we're starting to get into fiddly territory with the amount of scripts here.

package.json
BrandonOCasey: I ran into issues with it previously and forgot to remove it, I think we can remove it now.

package.json
BrandonOCasey: Yeah, I think we need a pr once we are off of grunt to clean it all up.

Issue #5056, opened 2018-03-28
Loading Spinner should have functional show() and hide() methods


ldayananda:
## Description
The [LoadingSpinner](https://github.com/videojs/video.js/blob/master/src/js/loading-spinner.js) component currently uses the [`vjs-waiting` and `vjs-seeking`](https://github.com/videojs/video.js/blob/master/src/css/components/_loading.scss) classes to display and hide itself. It would be much easier to use if the Component.hide() and Component.show() methods worked to hide and display the component and that is called where the classes are added currently.

### Expected
`LoadingSpinner.hide()` will hide the component and leave it in the DOM, `LoadingSpinner.show()` will show the component.

### Actual
Calling these methods does nothing

#### videojs
6.x

pilishen:
yeah,
`player.loadingSpinner.show()` or any other API would definitely be more clear than just `player.addClass('vjs-waiting')`

pilishen:
since there's already a `show()` and `hide()`, maybe we could also use `appear()` and `disappear()`

PR #5590, opened 2018-11-14
feat: Add an api for vjs-waiting

BrandonOCasey:
Right now a few of the plugins that we manage have been manually adding and removing the `vjs-waiting` class on the player, which is a bit weird. I think it would be better to have an API for this, but if we decide not to go this route that would be fine.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1664,7 +1664,7 @@ class Player extends Component {
     }
     ['progressControl', 'playToggle'].forEach(function(k) {
       if (this.controlBar[k]) {
-        this.controlBar[k].enable();
+        this.controlBar[k].disable();
       }
     });
   }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1602,6 +1602,88 @@ class Player extends Component {
   }
 
   /**
+   * Get or set vjs-waiting class on the player. This shows the loading spinner.
+   *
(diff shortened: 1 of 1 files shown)


gkatsev:
I wonder if we need this as opposed to just making the loading spinner actually work independently of the vjs-waiting class? Then plugins could just hide the spinner or show it whenever they want independent of vjs-waiting, then vjs-waiting/waiting event can also show it

BrandonOCasey:
Yeah we could essentially do that to, but I am not sure we can do it without a major version

src/js/player.js
```diff
@@ -1601,6 +1601,88 @@ class Player extends Component {
...
+  /**
+   * Disable playToggle and progress controls, effectively disabling playback
+   */
+  disableControls() {
+    if (!this.controlBar) {
+      return;
+    }
+    ['progressControl', 'playToggle'].forEach(function(k) {
+      if (this.controlBar[k]) {
+        this.controlBar[k].enable();
```
OwenEdwards: Should this be `this.controlBar[k].disable();`?

gkatsev:
I think I'm going to close this for now. I think we need to think more about what we want out of this and what we can do with the loading spinner currently for #5056 and #5629. We can revisit this PR if we think that's the direction we want to go.

PR #5591, opened 2018-11-14
chore: Lint on pre-commit with lint-staged, use npm-merge-driver

BrandonOCasey:
Not having `lint-staged` or `npm-merge-driver-install` has been bugging me, they make the workflow a lot quicker when committing or merging package-lock changes. It also allows us to only lint the partially staged changes that we are working on during commit/push.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,11 +114,13 @@
     "karma": "^3.0.0",
     "karma-browserify": "^5.3.0",
     "klaw-sync": "^6.0.0",
+    "lint-staged": "^8.0.4",
     "load-grunt-tasks": "^4.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -176,7 +178,13 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "npm run lint -- --errors"
+      "pre-commit": "lint-staged"
     }
+  },
+  "lint-staged": {
+    "*.js": [
+      "vjsstandard --fix --errors",
```
gkatsev: looks like `--errors` can't be combined with `--fix`. Removing it makes it fix things.

PR #5592, opened 2018-11-14
chore: Remove grunt move to npm scripts

BrandonOCasey:
Depends on videojs/video.js#5589 

Remove grunt altogether for npm scripts.


gkatsev:
https://github.com/videojs/video.js/blob/master/Gruntfile.js can be removed as well

gkatsev:
We need to comb through out docs and remove grunt and update to npm commands

package.json
```diff
@@ -131,7 +121,7 @@
     "npm-run-all": "^4.1.2",
     "postcss-cli": "^6.0.0",
     "proxyquireify": "^3.0.0",
-    "qunit": "~2.5.1",
+    "qunit": "^2.8.0",
```
gkatsev: we want to keep this pinned to ~2.5.1 due to the stack trace issue

package.json
```diff
@@ -151,12 +141,11 @@
     "shelljs": "^0.8.1",
     "shx": "^0.3.2",
     "sinon": "^1.16.1",
-    "time-grunt": "^1.1.1",
     "tui-jsdoc-template": "^1.1.0",
     "uglify-js": "^3.3.21",
     "unified": "^7.0.1",
     "videojs-flash": "^2.1.0",
-    "videojs-generate-karma-config": "~5.0.0",
+    "videojs-generate-karma-config": "~4.0.2",
```
gkatsev: This also should be 5.x

package.json
```diff
@@ -17,56 +17,52 @@
...
+    "minify:js": "babel-node build/minify.js",
+    "minify:css": "npm-run-all minify:css:*",
+    "minify:css:cdn": "cleancss dist/alt/video-js-cdn.css -o dist/alt/video-js-cdn.min.css",
+    "minify:css:default": "cleancss dist/video-js.css -o dist/video-js.min.css",
     "watch": "npm-run-all -p watch:*",
     "watch:lang": "chokidar --initial 'lang/**/*.json' -c 'npm run build:lang'",
-    "watch:rollup": "npm run rollup-dev",
-    "watch:sass-default": "npm run sass:default -- --watch 'src/**/**/*.scss'",
-    "watch:sass-cdn": "npm run sass:cdn -- --watch 'src/**/**/*.scss'",
+    "watch:rollup": "rollup -c -w ",
```
gkatsev: this is missing `--no-progress`

package.json
```diff
@@ -17,56 +17,52 @@
...
-    "build:browserify-test": "browserify test/require/browserify.js -o build/temp/browserify.js",
-    "build:webpack-test": "webpack --hide-modules test/require/webpack.js build/temp/webpack.js",
-    "copy-fonts": "shx cp -R node_modules/videojs-font/fonts build/temp/font",
-    "copy-dist": "shx cp -R build/temp/* dist/",
-    "copy-examples": "shx cp -R docs/examples dist/",
+    "clean": "shx rm -rf ./dist ./test/dist ./docs/api",
+    "postclean": "shx mkdir -p ./dist/lang ./test/dist",
+    "changelog": "conventional-changelog -p videojs -i CHANGELOG.md -s",
+    "build": "npm-run-all build-dev minify zip",
+    "build-dev": "npm-run-all clean lint-errors build:js build:css build:lang build:copy build:test",
```
gkatsev: I would move `build:test` just to `npm test`

package.json
```diff
@@ -17,56 +17,52 @@
...
+    "build:test": "npm-run-all build:test:*",
+    "build:test:a11y": "shx cp sandbox/descriptions.html.example sandbox/descriptions.test-a11y.html",
+    "build:test:browserify": "browserify test/require/browserify.js -o test/dist/browserify.js",
+    "build:test:webpack": "webpack --hide-modules test/require/webpack.js test/dist/webpack.js",
+    "build:copy": "npm-run-all build:copy:*",
+    "build:copy:fonts": "shx cp -R node_modules/videojs-font/fonts dist/font",
+    "build:copy:examples": "shx cp -R docs/examples dist/",
+    "build:js": "rollup -c --no-progress",
```
gkatsev: I believe we want progress here, just not in the watch

package.json
```diff
@@ -17,56 +17,52 @@
...
-    "cssmin:default": "cleancss dist/video-js.css -o dist/video-js.min.css",
-    "sass": "npm-run-all sass:* autoprefixer",
-    "sass:cdn": "node-sass src/css/vjs-cdn.scss dist/alt/video-js-cdn.css",
-    "sass:default": "node-sass src/css/vjs.scss dist/video-js.css",
-    "autoprefixer": "npm-run-all -p autoprefixer:*",
-    "autoprefixer:main": "postcss --verbose --config postcss.config.js -d dist/ dist/video-js.css",
-    "autoprefixer:cdn": "postcss --verbose --config postcss.config.js -d dist/alt dist/alt/video-js-cdn.css",
-    "rollup": "rollup -c",
-    "rollup-dev": "rollup -c -w",
+    "build:lang:copy": "shx --verbose cp -R lang/* dist/lang/",
```
gkatsev: this shouldn't have --verbose in it

package.json
```diff
@@ -17,56 +17,52 @@
...
+    "lint-errors": "vjsstandard --errors",
+    "start": "npm-run-all -p watch karma-server",
     "karma-server": "karma start test/karma.conf.js --singleRun=false --auto-watch",
-    "karma-run": "karma start test/karma.conf.js",
-    "test": "grunt test",
-    "docs": "npm run docs:lint && npm run docs:api",
-    "jsdoc": "jsdoc",
-    "predocs:api": "node -e \"var s=require('shelljs'),d=['docs/api'];s.rm('-rf',d);\"",
-    "checktranslations": "grunt check-translations",
+    "test": "npm-run-all build-dev test:*",
```
gkatsev: what about moving `build-dev` to `pretest`, then running `test:*` in parallel?

package.json
```diff
@@ -17,56 +17,52 @@
...
+    "watch:rollup": "rollup -c -w ",
+    "watch:css-default": "npm run build:css:default -- --watch 'src/**/**/*.scss'",
+    "watch:css-cdn": "npm run build:css:cdn -- --watch 'src/**/**/*.scss'",
     "assets": "node build/assets.js",
-    "grunt": "grunt",
     "lint": "vjsstandard",
-    "start": "npm-run-all -p start:grunt watch",
-    "start:grunt": "grunt dev",
+    "lint-errors": "vjsstandard --errors",
+    "start": "npm-run-all -p watch karma-server",
```
gkatsev: can we move `start` to the top of the scripts section?

package.json
```diff
@@ -17,56 +17,52 @@
...
-    "build:webpack-test": "webpack --hide-modules test/require/webpack.js build/temp/webpack.js",
-    "copy-fonts": "shx cp -R node_modules/videojs-font/fonts build/temp/font",
-    "copy-dist": "shx cp -R build/temp/* dist/",
-    "copy-examples": "shx cp -R docs/examples dist/",
+    "start": "npm-run-all -p watch karma-server",
+    "clean": "shx rm -rf ./dist ./test/dist ./docs/api",
+    "postclean": "shx mkdir -p ./dist/lang ./test/dist",
+    "changelog": "conventional-changelog -p videojs -i CHANGELOG.md -s",
+    "build": "npm-run-all build-dev minify zip",
+    "build-dev": "npm-run-all clean lint-errors build:js build:css build:lang build:copy build:test",
```
gkatsev: `build:copy` only needs to run on publish, there's nothing in there that's used for dev

package.json
```diff
@@ -17,56 +17,52 @@
...
+    "build:test": "npm-run-all build:test:*",
+    "build:test:a11y": "shx cp sandbox/descriptions.html.example sandbox/descriptions.test-a11y.html",
+    "build:test:browserify": "browserify test/require/browserify.js -o test/dist/browserify.js",
+    "build:test:webpack": "webpack --hide-modules test/require/webpack.js test/dist/webpack.js",
+    "build:copy": "npm-run-all build:copy:*",
```
gkatsev: I would make these just `copy` and `copy:fonts`, etc

package.json
gkatsev: maybe it's fine to leave it to match current functionality

package.json
BrandonOCasey: makes sense

package.json
BrandonOCasey: While that kind of makes sense, I think these should be bulit here so that they exist for the dev server.

package.json
BrandonOCasey: what if we just changed it to `build:fonts` and `build:examples` instead? To me even though these are not really building anything, they need to run with the rest of the build.

BrandonOCasey: I guess they could run as `copy` with `build` instead of `build-dev`

gkatsev: I don't think either fonts or examples are used anywhere directly except for providing those files in the published tarball

gkatsev:
Also, I realized that there are grunt tasks that were being used but not currently set:
- [`cdn-links`](https://github.com/videojs/video.js/blob/master/build/tasks/cdn-links.js) - used to fix up some URLs in the generated API docs
- [languages](https://github.com/videojs/video.js/blob/master/build/tasks/languages.js) - used to update the translations-needed doc
- [`sandbox`](https://github.com/videojs/video.js/blob/master/build/tasks/sandbox.js) - used to set up the initial sandbox folder and keep it maintained by copying the example files over to regular html files

gkatsev:
also, `build/bin` folder can be removed

gkatsev:
Last grunt tasks are ported via this PR https://github.com/videojs/video.js/pull/5634

PR #5611, opened 2018-11-24
Feat(player): add option to disable or override double-click handling. Fixes #5604

### LINKED ISSUE #5604: Disable full screen on double click
I want to completely disable full screen functionality. I removed fullscreen button like this: 

```
videojs('videoPlayer', {
            controlBar: {
                fullscreenToggle: false
            }
        });
```

But on double click it still goes to full screen. How to disable double click?

OwenEdwards:
## Description
Add a player option `dblclick` to disable or override double-click handling. Setting the option to `false` disables handling, setting it to a function allows double-click handling in that function.

## Specific Changes proposed
Added `dblclick` option to the player, and documented the option in docs/guides/options.md.

## Requirements Checklist
  - [x] Docs/guides updated


src/js/player.js
```diff
@@ -1856,10 +1856,19 @@ class Player extends Component {
...
-      } else {
-        this.requestFullscreen();
+      /*
+       * options.dblclick
+       * Set to false to disable double click handling
+       * Set to a function to substitute an external double click handler
+       */
+      if (this.options_.dblclick !== false) {
+        if (typeof this.options_.dblclick === 'function') {
+          this.options_.dblclick(event);
```
BrandonOCasey: Should we not allow this to be a function? Then if they want to handle it, they should listen for the event themselves and set this to false.

gkatsev:
We spoke about it a bit at grooming, we think that maybe having the option be a sub property of `useractions` or `actions` (either plural or singular) might be better. Then all hotkeys and gestures can be under this property.

src/js/player.js
gkatsev: An argument in favor of this being a method is that it makes it much easier to override the default behavior for a particular hotkey/gesture.

gkatsev: We should probably make this be a `.call(this, event)` so it's more similar to a regular event handler.

```suggestion
          this.options_.dblclick.call(this, event);
```

BrandonOCasey: I think its fine

OwenEdwards:
The `options` handling in this is compatible with #5571, so I think it's ready to be merged (although some manual merging will be needed on whichever one gets merged second). @gkatsev let me know what you want to do with this; I'd like to get it wrapped up soon too!

gkatsev:
Thanks @OwenEdwards. I'll take a look at this and #5571 tomorrow but it's possible it won't go in till next week. However, I'm thinking that we should just start merging things and we can have a patch release off on a separate branch for stuff like the fullscreen fix to stop holding up these PRs (and since I'm thinking that the `reset` changes maybe should just be a minor).

OwenEdwards:
@gkatsev grrr! I accepted your "suggestion[s] to updating the if statements a bit", and it looks like the GitHub UI merged them with trailing spaces which then fails the lint testing in Travis.

gkatsev:
Heh, whoops.

Issue #5616, opened 2018-11-27
All Travis CI builds of new PRs are failing


OwenEdwards:
## Description
All Travis CI builds of new PRs are failing, even ones which don't change the codebase (see #5602 and https://travis-ci.org/videojs/video.js/builds/456148811?utm_source=github_status&utm_medium=notification).

### Error output (From Travis CI)
```
Running "shell:karma-run" (shell) task
> video.js@7.3.0 karma-run /home/travis/build/videojs/video.js
> karma start test/karma.conf.js
16 11 2018 20:43:24.762:INFO [framework.detect-browsers]: Detecting browsers is disabled. The browsers of the browsers array are used.
16 11 2018 20:43:35.409:INFO [framework.browserify]: bundle built
16 11 2018 20:43:35.432:INFO [karma-server]: Karma v3.1.1 server started at http://0.0.0.0:9999/test/
16 11 2018 20:43:35.432:INFO [launcher]: Launching browsers travisFirefox, travisChrome with concurrency unlimited
16 11 2018 20:43:35.437:INFO [launcher]: Starting browser FirefoxHeadless
16 11 2018 20:43:35.454:INFO [launcher]: Starting browser ChromeHeadless
16 11 2018 20:43:36.127:INFO [HeadlessChrome 0.0.0 (Linux 0.0.0)]: Connected on socket Bx06whqXUwgXJnYxAAAA with id 26409926
16 11 2018 20:43:36.153:WARN [middleware:karma]: Invalid file type, defaulting to js. browserify
16 11 2018 20:43:37.213:INFO [Firefox 63.0.0 (Linux 0.0.0)]: Connected on socket UI5h0oVzBDG8gFZIAAAB with id 72587728
16 11 2018 20:43:37.236:WARN [middleware:karma]: Invalid file type, defaulting to js. browserify
................................................................................
................................................................................
....................................16 11 2018 20:43:41.113:WARN [middleware:karma]: Invalid file type, defaulting to js. browserify
............................................
................................................................................
................................................................................
........................................................16 11 2018 20:43:43.915:WARN [web-server]: 404: /test/relative-one.mp4
16 11 2018 20:43:43.927:WARN [web-server]: 404: /test/relative-one.mp4
16 11 2018 20:43:43.941:WARN [web-server]: 404: /test/relative-three.mp4?test=test
........................
............16 11 2018 20:43:44.966:WARN [web-server]: 404: /test/relative-one.mp4
[...]

PR #5619, opened 2018-11-27
WIP: update test timeout

gkatsev:
Investigate #5616


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - lts/*
+  - lts/carbon
```
BrandonOCasey: fyi travis will use .nvmrc, so we don't need to specify it twice :)

PR #5621, opened 2018-11-28
tests: rollup tests instead of browserify

BrandonOCasey:
Move the tests off of rollup completely. Previously we were overriding an import. In es5 that was OK, but in es6 that is not possible, since it is a const. Any tests that override a const where changed to use `rollup-plugin-stub`.

Based on #5601


test/globals-shim.js
```diff
@@ -1,19 +0,0 @@
-/* eslint-env qunit */
```
gkatsev: Is there a replacement for this? It does a few important things besides just introducing globals.

BrandonOCasey: I thought it just added qunit global variables to an ignore list. Since we are already importing qunit in every file where we use it, its doesn't seem useful.

BrandonOCasey: https://eslint.org/docs/user-guide/configuring#specifying-environments

gkatsev: It also sets up the qunit fixture and `window.xhr` before video.js loads.

The importing of es5-shim and es6-shim is incredibly useful as well since some tests depend on a good ES6 environment.

gkatsev: Looks like the qunit fixture is handled by karma-qunit now https://github.com/karma-runner/karma-qunit/blob/master/src/adapter.js#L41-L56

I'm still a bit skeptical of the `xhr` change. Our XHR dep grabs the global XHR when it loads and thus makes it really hard to mock unless we set up the fake XHR before that code loads.

BrandonOCasey: hls and http-streaming just override the `XMLHttpRequest` contained on the `videojs.xhr` object. Internally we just have to override our xhr import the same way (with the sinon xhr). https://github.com/videojs/video.js/pull/5621/files#diff-3b4ce30c38112a433d137fb9bf1bcbc9R19

BrandonOCasey: https://github.com/videojs/http-streaming/blob/master/test/test-helpers.js#L200

gkatsev: Ah, cool. It's been a while, and I guess things got a lot better and we don't need to make sure it's overridden ahead of time.

gkatsev: So, I guess the only thing that we'd need to keep from this is the es-shims (which I see you've added) and the rest should be taken care of elsewhere.

PR #5629, opened 2018-11-29
Refactor/loading spinner show hide

imnotteixeira:
## Description
Solves videojs/video.js#5056 

## Changes
Created show() and hide() methods to LoadingSpinner class, and changed player.js methods to utilize those loading spinner methods instead of directly changing the DOM.
Also created unit tests for this refactor, which are currently passing.

The show() method adds the `vjs-waiting` css class while the hide() method removes it.
The `vjs-seeking` class was removed, as it was not necessary, because it did the exact same as the `vjs-waiting`

## Requirements Checklist
- [x] Refactor implemented
- [x] Change has been verified in an actual browser (Chome)


imnotteixeira:
@ldayananda @gkatsev please review :)

gkatsev:
Unfortunately, the removal of vjs-seeking classes makes this change not backwards compatible.

imnotteixeira:
Ok then, so if I add and remove always the `vjs-seeking` and `vjs-waiting` classes, would that work?

gkatsev:
Maybe, the other problem is the semantic meaning of the change in behavior. Though, it's possible that there isn't a better way of doing it and still staying backwards compatible.

mister-ben:
This should check whether the player has a loading spinner child before calling methods on it. 

I'm concerned that adding both the seeking and the waiting classes in either state would also be a breaking change. I'm sure I've seen customised players with distinct styles for those classes.

imnotteixeira:
Also added the verification of spinner existence as @mister-ben suggested

src/js/player.js
```diff
@@ -1286,6 +1286,11 @@ class Player extends Component {
 
     this.removeClass('vjs-ended');
     this.removeClass('vjs-seeking');
+    const loadingSpinner = this.getChild('loadingSpinner');
```
BrandonOCasey: I think it would be better to make a `showSpinner` and `hideSpinner` function and call that it each of these places.

gkatsev: components are already supposed to have show/hide methods, so, I think this is fine

imnotteixeira: @BrandonOCasey I'm trying not to make the player class feature envious, which is a code smell (https://refactoring.guru/smells/feature-envy), thus the override of existing show and hide methods is appropriate

dcm1984:
Working as intended and the tests should cover all the use cases

imnotteixeira:
@gkatsev Do you think the PR would be eligible to be merged, or would you recommend any more changes?

xRuiAlves:
@gkatsev  Would this PR be ready to be accepted? We (me, @imnotteixeira  and @reeckset) are currently working on this open source project in the context of a university course. 
In this course we are challenged to participate in an open source project, solving two issues and documenting the project's architecture.
It would be very beneficial if this PR could be approved (should not even need to be merged, just have its changes accepted). Thanks in advance!

gkatsev:
You probably need to update your fork with the latest stuff in the master branch.

imnotteixeira:
@gkatsev 

I did that, I also tried just rewriting the package-lock with the one in the master branch, nothing works..

xRuiAlves:
Good afternoon @gkatsev , @mister-ben , @BrandonOCasey ,
As mentioned by my collegue @miguelpduarte in PR #5663, me and an a few more collegues are currently working on videojs in the context of an university course project. The project's due date is tomorrow morning and it would help us a real lot if this PR (aswell as our other PR, #5684) was approved (it needs not even need to be merged; just having it accepted would already be a great help for us!). 
Do you think that that would be possible? Thanks in advance!

src/js/loading-spinner.js
```diff
@@ -34,6 +34,22 @@ class LoadingSpinner extends Component {
 
     return el;
   }
+
+  /**
+   * Show the `LoadingSpinner`s DOM element, by adding the `show` css class to it.
+   */
+  show() {
+    super.show();
+    this.el().classList.add('show');
```
gkatsev: Component has an API for adding and removing classes [`this.addClass`](https://github.com/videojs/video.js/blob/master/src/js/component.js#L721)

src/css/components/_loading.scss
```diff
@@ -20,8 +20,7 @@
   visibility: hidden;
 }
 
-.vjs-seeking .vjs-loading-spinner,
-.vjs-waiting .vjs-loading-spinner {
+.vjs-loading-spinner.show {
```
gkatsev: the `.show` class is too generic. We generally prefix everything with `vjs`

gkatsev:
I'm not sure that we'll be able to approve it in time for your deadline. This change is potentially not backwards compatible and so we'd want to verify that before approving. We won't be able to accept it if it isn't backwards compatible or we're not confidant enough that the change is backwards compatible. I'm unsure whether I'll have enough time to review this PR today and think about its implications in the project.

xRuiAlves:
Allright, we are going to add the requested changes to the PR anyhow. Thank you for the input

gkatsev:
Thanks! We'll definitely spend the time looking at the PR, it just may not happen tonight.

Is there a particular reason why it needs to be approved?

xRuiAlves:
In our course's project, we are "challenged" to take part and collaborate in an open source project of our interest. As I had already worked with video.js quite a lot in my summer internship in the past summer, we ended up trying to collaborate to it! We got bonus points in our final grade assessment if we get our PR accepted by a project's member. Sorry if I was a bit to insistent :sweat_smile:

src/js/loading-spinner.js
imnotteixeira: Changed in latest PR

src/css/components/_loading.scss
imnotteixeira: Changed in latest commit to .vjs-loading-spinner-show

dcm1984:
I think the Loading Spinner show/hide API actually solves another (older) issue regarding the loading spinner. I can't seem to find it tho, maybe it was marked as stale?

gkatsev:
Finally got a chance to take a look at this in detail. I think the changes look good but as they stand this is *not* backwards compatible. For better or worse, the loading spinner needs to come up when `vjs-waiting` class is added to the player. I wonder if keeping the old CSS around with the rest of the changes will satisfy all requirements.

PR #5649, opened 2018-12-04
chore: package json cleanup

brandonocasey:
1. Remove packages that we really don't need:
  * `bluebird` - use native promises instead
  * `klawsync` - use `shelljs`
  * `babel/*` - lots of unused/uneeded babel packages, especially after converting from `import` to `require` for our build scripts
  * `minimist` - Was only used in one place, and we can just manually look for the one argument.
  * `tsml` no longer maintained
2. Turns on linting for almost everything
3. Uses `lint-staged` to update the `translations-needed` doc
4. Uses `lint-staged` to lint and fix docs


package.json
```diff
@@ -18,45 +18,36 @@
...
     "build:test:a11y": "shx cp sandbox/descriptions.html.example sandbox/descriptions.test-a11y.html",
     "build:test:browserify": "browserify test/require/browserify.js -o test/dist/browserify.js",
     "build:test:webpack": "webpack --hide-modules test/require/webpack.js test/dist/webpack.js",
-    "copy": "npm-run-all copy:*",
-    "copy:fonts": "shx cp -R node_modules/videojs-font/fonts dist/font",
-    "copy:examples": "shx cp -R docs/examples dist/",
     "build:js": "rollup -c",
-    "build:css": "npm-run-all build:css:*",
-    "build:css:cdn": "node-sass src/css/vjs-cdn.scss dist/alt/video-js-cdn.css",
-    "postbuild:css:cdn": "postcss --verbose --config postcss.config.js -d dist/alt dist/alt/video-js-cdn.css",
```
BrandonOCasey: this has been moved into `build/build-css.js`

package.json
```diff
@@ -18,45 +18,36 @@
...
-    "build:css:default": "node-sass src/css/vjs.scss dist/video-js.css",
-    "postbuild:css:default": "postcss --verbose --config postcss.config.js -d dist/ dist/video-js.css",
+    "build:css": "node build/build-css.js",
     "build:lang": "npm-run-all build:lang:*",
     "build:lang:js": "vjslang --dir dist/lang",
     "build:lang:copy": "shx cp -R lang/* dist/lang/",
     "minify": "npm-run-all minify:*",
-    "minify:js": "babel-node build/minify.js",
```
BrandonOCasey: After the rollup tests pr is merged, I will work on a pr to use the shared rollup config, then the `minify` script will go away entirely.

gkatsev: Having rollup minify significantly increased build times because it had to be a separate rollup build.

build/assets.js
```diff
@@ -1,65 +1,60 @@
...
 
-      resolve([file, gzStat]);
+      resolve({filepath, gzStat, fileStat});
     })
-    .on('error', reject);
+      .on('error', reject);
   });
-}
-
-function logTable(files) {
```
gkatsev: keeping these as named functions rather than inlining them into the promise chain increases readability of this file and the promise chain.

build/sandbox.js
```diff
@@ -1,28 +1,32 @@
+/* eslint-disable no-console */
+
 const fs = require('fs');
 const path = require('path');
-const klawSync = require('klaw-sync');
+const sh = require('shelljs');
 
-const files = klawSync('sandbox/').filter((file) => path.extname(file.path) === '.example');
+const files = sh.find(path.join(__dirname, '..', 'sandbox', '**', '*.*'))
```
gkatsev: klawSync is more readable, I think

build/sandbox.js
```diff
@@ -1,28 +1,32 @@
...
+
   return {
     file: file.path,
     copy: nonExample
   };
 })
-.filter(function(change) {
-  return !fs.existsSync(change.copy);
-});
+  .filter(function(change) {
```
gkatsev: does the linter fix this?

build/generate-example.js
```diff
@@ -42,11 +41,20 @@ export default function generateExample({skipBuild} = {}) {
     sh.cp(path.join(vjsSwf, 'dist', 'video-js.swf'), swfDest);
   }
 
-  const files = klawSync('sandbox/').filter((file) => path.extname(file.path) === '.example');
+  const filepaths = sh.find(path.join(__dirname, '..', 'sandbox', '**', '*.*'))
```
gkatsev: klawSync is more readable, I think

BrandonOCasey: We could use just `sandbox/` for `find` too, but then the scripts don't work if run from somewhere other than the root dir

build/sandbox.js
BrandonOCasey: I think that it did

build/assets.js
```diff
@@ -1,65 +1,60 @@
...
-const files = klawSync('dist/', {
-  ignore: ['examples', 'lang', 'font', '*.zip', '*.gz'],
-  nodir: true
-});
-
-Promise.all(files.map(gzipAndStat))
-.then(mapFiles)
-.then(function(files) {
-  logTable(files);
+const filepaths = sh.find(path.join(__dirname, '..', 'dist', '**', '*.{js,css}')).filter(function(filepath) {
```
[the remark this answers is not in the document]
BrandonOCasey: So for me `klawSync` wasn't actually working here. I did not see an `ignore` option anymore. Now it seems you have to filter on your own. So `klawSync` would be mostly the same. The `path.join` is just so we can run this script from any directory and not just the root of the project.

BrandonOCasey:
General comment about `klawSync` I think it's doing more than we need in all but one script, ie `find` and `stat` when we just need `find`. I think removing the dependency is worth it in this case.

package.json
```diff
@@ -188,6 +167,14 @@
     "*.js": [
       "vjsstandard --fix",
       "git add"
+    ],
+    "!(CHANGELOG)*.md": [
+      "remark --output --",
+      "git add"
+    ],
+    "lang/*.json": [
```
brandonocasey: automatic translation updates on commit when they are changed!

package.json
brandonocasey: now that terser supports `include/exclude` this won't be an issue.

build/assets.js
```diff
@@ -1,57 +1,41 @@
...
-});
-
-Promise.all(files.map(gzipAndStat))
-.then(mapFiles)
-.then(function(files) {
-  logTable(files);
+// find all js/css files in the dist dir
+// but ignore any files in lang, example, or font directories
+const filepaths = sh.find(path.join(__dirname, '..', 'dist', '**', '*.{js,css}')).filter(function(filepath) {
+  if ((/\/(lang|example|font)\//).test(filepath)) {
```
gkatsev: you could just `return !(/.../.test(filepath);`

build/generate-example.js
```diff
@@ -42,11 +41,20 @@ export default function generateExample({skipBuild} = {}) {
...
-    sh.cp(file.path, path.join(exampleDest, p.name));
+  filepaths.forEach(function(filepath) {
+    const p = path.parse(filepath);
+
+    sh.cp(filepath, path.join(exampleDest, p.name));
   });
-}
+};
+
+generateExample({skipBuild: true});
```
gkatsev: I don't think we need to run this on require.

package.json
```diff
@@ -18,6 +18,7 @@
   "author": "Steve Heffernan",
   "scripts": {
     "sandbox": "node build/sandbox.js",
+    "assert": "node build/assets.js",
```
gkatsev: This already exists as `assets` script.

package.json
```diff
@@ -69,10 +70,11 @@
     "docs:lint": "remark -- './{,!(node_modules)/**/}!(CHANGELOG)*.md'",
     "docs:fix": "remark --output -- './{,!(node_modules)/**/}!(CHANGELOG)*.md'",
     "docs:lang": "node build/translations.js",
-    "netlify": "babel-node ./build/netlify-docs.js",
+    "netlify": "node ./build/netlify-docs.js",
     "prepublishOnly": "run-p build",
     "publish": "node build/gh-release.js",
-    "version": "node build/version.js && git add CHANGELOG.md",
+    "update-changelog": "conventional-changelog -p videojs -i CHANGELOG.md -s",
```
gkatsev: we already have the `changelog` script.

PR #5650, opened 2018-12-04
fix: liveui tweaks

BrandonOCasey:
## Changes
1. Make sure that we seek to live playback on the first timeupdate
1. Do not report that we are not live before playback has started (a timeupdate has been seen)
1. Prevent negative seekable increments
1. We can seek past the `seekable` value in the video element, so we use that to seek to live, rather than waiting for a seekable end change.

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -49,7 +49,7 @@ class LiveTracker extends Component {
     // end against current time, with a fudge value of half a second.
     if (newSeekEnd !== this.lastSeekEnd_) {
       if (this.lastSeekEnd_) {
-        this.seekableIncrement_ = newSeekEnd - this.lastSeekEnd_;
+        this.seekableIncrement_ = Math.abs(newSeekEnd - this.lastSeekEnd_);
       }
 
       this.pastSeekEnd_ = 0;
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -16,6 +16,10 @@ class LiveTracker extends Component {
   }
 
   isBehind_() {
+    // don't report that we are behind until a timeupdate has been seen
+    if (!this.timeupdateSeen_) {
+      return false;
+    }
     const liveCurrentTime = this.liveCurrentTime();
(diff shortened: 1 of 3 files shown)


test/unit/live-tracker.test.js
```diff
@@ -169,15 +154,15 @@ QUnit.module('LiveTracker', () => {
   QUnit.test('single seekable, helpers should be correct', function(assert) {
     // simple
     this.player.seekable = () => createTimeRanges(10, 50);
-    assert.strictEqual(this.liveTracker.liveWindow(), 40, 'liveWindow is 40s');
+    assert.strictEqual(this.liveTracker.liveWindow(), 40.03, 'liveWindow is 40s');
```
misteroneill: Why the .03?

BrandonOCasey: `liveWindow()` now uses `liveCurrentTime` instead of `seekableEnd`. We increment `live CurrentTime` by 0.03 every 30ms, hence `.03`

BrandonOCasey: oh and we start at 0.03 ahead at init, since we are falling behind every ms the player isn't ready.

src/js/live-tracker.js
```diff
@@ -90,6 +94,21 @@ class LiveTracker extends Component {
 
     this.on(this.player_, 'play', this.trackLive_);
     this.on(this.player_, 'pause', this.trackLive_);
+    this.on(this.player_, 'firstplay', this.handleFirstplay);
```
gkatsev: we should not be using `firstplay`. It's non-standard, deprecated, and I'm not sure it's complete.

gkatsev:
This is a great change towards making the button seek to "live" asap. In another PR, we probably need to do @gesinger's suggestion of looseing up what "at edge" means as I can see the inidicator toggling between grey and red during regular playback.

src/js/live-tracker.js
BrandonOCasey: done, I didn't realize that it wasn't something we fired internally.

PR #5651, opened 2018-12-04
fix: make seek-to-live accessible

BrandonOCasey:
Makes seek to live accessible and indicates when the button can be pressed or not to the screen reader.


lang/en.json
```diff
@@ -9,6 +9,8 @@
   "Remaining Time": "Remaining Time",
   "Stream Type": "Stream Type",
   "LIVE": "LIVE",
+  "Seek to LIVE": "Seek to LIVE",
```
gkatsev: we should run `npm run docs:lang` to update the translations-needed doc.

src/js/control-bar/seek-to-live.js
```diff
@@ -39,19 +39,19 @@ class SeekToLive extends Button {
...
+    this.textEl_ = Dom.createEl('span', {
+      className: 'vjs-seek-to-live-text',
+      innerHTML: this.localize('LIVE')
     }, {
-      'aria-live': 'off'
+      'aria-hidden': 'true'
     });
 
     this.circleEl_ = Dom.createEl('span', {
       className: 'vjs-seek-to-live-circle'
```
gkatsev: while we're at it, should we just reuse the vjs-icon-placeholder element for this? Probably fine to leave it as is, though.

src/js/control-bar/seek-to-live.js
```diff
@@ -39,19 +39,19 @@ class SeekToLive extends Button {
       className: 'vjs-seek-to-live-control vjs-control'
     });
 
-    this.contentEl_ = Dom.createEl('div', {
-      className: 'vjs-seek-to-live',
-      innerHTML: `<span class="vjs-control-text">${this.localize('Stream Type')}\u00a0</span>${this.localize('LIVE')}`
+    this.textEl_ = Dom.createEl('span', {
+      className: 'vjs-seek-to-live-text',
+      innerHTML: this.localize('LIVE')
```
gkatsev: so make `LIVE` be purely visual and have the control-text be what is announced to screen readers?

src/js/control-bar/seek-to-live.js
```diff
@@ -61,9 +61,13 @@ class SeekToLive extends Button {
    */
   updateLiveEdgeStatus(e) {
     if (this.player_.liveTracker.behindLiveEdge()) {
+      this.setAttribute('aria-disabled', false);
       this.removeClass('vjs-at-live-edge');
+      this.controlText('Seek To LIVE');
```
gkatsev: I wonder if this should be more verbose like `not at live edge, click to seek to live`.

src/js/control-bar/seek-to-live.js
BrandonOCasey: yes

src/js/control-bar/seek-to-live.js
BrandonOCasey: yeah let me see if that has any unknown side effects

lang/en.json
BrandonOCasey: Will do that now, just so we remember, but I am not sold on what it should actually say yet.

src/js/control-bar/seek-to-live.js
BrandonOCasey: I am going to change it to that, as being verbose is probably better.

lang/en.json
gkatsev: we should remember to do this right before merging

src/js/control-bar/seek-to-live.js
OwenEdwards: Is there a display somewhere in the control bar which shows how far behind live you are? I'm wondering if that ought to also be indicated here, e.g. "Seek to Live, currently 1:23 behind live". I think you should stay away from the term "edge" - that's a technical term, but not one a typical user would recognize.

src/js/control-bar/seek-to-live.js
```diff
@@ -61,9 +61,13 @@ class SeekToLive extends Button {
...
     } else {
       this.addClass('vjs-at-live-edge');
+      this.controlText('At LIVE edge');
```
OwenEdwards: "Playing Live", perhaps? As I mentioned, I don't think "edge" is meaningful outside the technical world.

src/js/control-bar/seek-to-live.js
gkatsev: the current time shows up as how far behind live you are, I think. We can probably use that to update this.

OwenEdwards: Also, try to avoid "click to..." - that's implied with a button, but some users may tap, some click, some use the keyboard. And try to front-load the action, then add more information afterwards - "Seek to Live, currently 1:23 behind live" instead of "not currently live, seek to live".

src/js/control-bar/seek-to-live.js
```diff
@@ -39,19 +39,19 @@ class SeekToLive extends Button {
...
 
-    this.contentEl_ = Dom.createEl('div', {
-      className: 'vjs-seek-to-live',
-      innerHTML: `<span class="vjs-control-text">${this.localize('Stream Type')}\u00a0</span>${this.localize('LIVE')}`
+    this.textEl_ = Dom.createEl('span', {
+      className: 'vjs-seek-to-live-text',
+      innerHTML: this.localize('LIVE')
     }, {
-      'aria-live': 'off'
+      'aria-hidden': 'true'
```
OwenEdwards: When does this button get `aria-hidden`? It should not be displayed when it's not relevant (and "display: none" prevents it from being visible to a screen reader as well as to a sighted user), and if it is disabled it is better to mark it as disabled (or `aria-disabled`) than `aria-hidden`.

src/js/control-bar/seek-to-live.js
OwenEdwards: Or "Seek to Live, currently playing Live", which would then be announced as "button, disabled".

src/js/control-bar/seek-to-live.js
gkatsev: Ah, makes sense for the action to be front loaded, I was thinking that maybe the info is more important but it isn't

src/js/control-bar/seek-to-live.js
gkatsev: I think the idea is that the span that says "LIVE" is only for sighted users and SR/keyboard users get the same information elsewhere (like the control text)

src/js/control-bar/seek-to-live.js
BrandonOCasey: seems better to me, implemented 👍

src/js/control-bar/seek-to-live.js
BrandonOCasey: exactly as @gkatsev said, that is the idea

src/js/control-bar/seek-to-live.js
```diff
@@ -61,9 +56,13 @@ class SeekToLive extends Button {
    */
   updateLiveEdgeStatus(e) {
     if (this.player_.liveTracker.behindLiveEdge()) {
+      this.setAttribute('aria-disabled', false);
       this.removeClass('vjs-at-live-edge');
+      this.controlText('Seek to live, currently behind live');
     } else {
       this.addClass('vjs-at-live-edge');
+      this.controlText('Seek to live, currently playing live');
+      this.setAttribute('aria-disabled', true);
```
gkatsev: do we want to use `this.disable()` and `this.enable()` here instead? https://github.com/videojs/video.js/blob/master/src/js/clickable-component.js#L147-L177 and
https://github.com/videojs/video.js/blob/master/src/js/button.js#L79-L95

BrandonOCasey: I don't know if we want the button to dim visually, which will happen if we do that. I think we just want the button to "dim" for screen readers, normal users will see a cursor change and will already be at live with a red circle.

gkatsev:
Just tried it out with VO, seems to be working well. @OwenEdwards do you have any other input or is this good to go?

PR #5652, opened 2018-12-04
feat: Add loadMedia and getMedia methods

### LINKED ISSUE #4342: Feature Proposal: loadMedia, getMedia
More and more, when videojs is used, more than just a source is needed. There's the poster image, there's captions, and there's also various properties like muted, and loop, and autoplay.
Loading all these things are often implemented in places like [videojs-playlist](https://github.com/brightcove/videojs-playlist/blob/master/src/js/play-item.js#L36-L49) and Google has a new [Media Session API](https://github.com/videojs/videojs-media-session) that expects similar items.
This method would accept an object that has all these properties and when called, it'll make sure to unset all the old things. That means that all old posters, all old text tracks would be removed and unset and then the new things would be set. This would be different than something like `src` which wouldn't necessarily remove old textTracks or posters.

@imbcmdth I wonder if we'd want to revert the change to the addRemoteTextTrack flag for remove on src change in this case? Or maybe, just keep the default the same?

Also, we're thinking of having it be native in 6.x and be a simpler plugin 5.x that doesn't have as deep of integration with the rest of the player (like removing old text tracks).

misteroneill:
## Description
This adds methods for loading media into a player.

## Specific Changes proposed
- `loadMedia` will set source, poster, and text tracks. It will create an `artworks` array for uses with the `MediaSession` API.
- `getMedia` will return a clone of the media object.

Fixes #4342

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2931,7 +2931,9 @@ class Player extends Component {
     if (this.tech_) {
       this.tech_.clearTracks('text');
     }
-    this.cache_.media = null;
+    if (this.cache_) {
+      this.cache_.media = null;
+    }
     this.loadTech_(this.options_.techOrder[0], null);
     this.techCall_('reset');
     if (isEvented(this)) {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4029,19 +4029,38 @@ class Player extends Component {
   }
 
   /**
-   * Get a clone of the current {@link Player~MediaObject} for this player or
-   * `null` if there is none.
+   * Get a clone of the current {@link Player~MediaObject} for this player.
    *
-   * @return {Player~MediaObject|null}
+   * If the `loadMedia` method has not been used, will attempt to return a
+   * {@link Player~MediaObject} based on the current state of the player.
+   *
+   * @return {Player~MediaObject}
    */
   getMedia() {
     if (!this.cache_.media) {
-      return null;
+      const poster = this.poster();
+      const src = this.currentSources();
+      const textTracks = Array.prototype.map.call(this.remoteTextTracks(), (tt) => ({
+        kind: tt.kind,
+        label: tt.label,
+        language: tt.language,
+        src: tt.src
+      }));
+
+      const media = {src, textTracks};
+
+      if (poster) {
+        media.poster = poster;
+        media.artwork = [{
+          src: media.poster,
+          type: getMimetype(media.poster)
+        }];
+      }
+
+      return media;
     }
 
-    const media = mergeOptions(this.cache_.media);
-
-    return media;
+    return mergeOptions(this.cache_.media);
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2934,6 +2934,7 @@ class Player extends Component {
     if (this.cache_) {
       this.cache_.media = null;
     }
+    this.poster('');
     this.loadTech_(this.options_.techOrder[0], null);
     this.techCall_('reset');
     if (isEvented(this)) {
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3980,8 +3980,8 @@ class Player extends Component {
    *           For ease of removal, these will be created as "remote" text
    *           tracks and set to automatically clean up on source changes.
    *
-   *           For more information on what properties are available for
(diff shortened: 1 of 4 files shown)


src/js/player.js
```diff
@@ -3930,6 +3930,113 @@ class Player extends Component {
...
+   */
+  loadMedia(media, ready) {
+    if (!media || typeof media !== 'object') {
+      return;
+    }
+
+    this.reset();
+
+    // Clone the media object so it cannot be mutated from outside.
+    this.media_ = mergeOptions(media);
```
gkatsev: we should add this to the `this.cache_` object and then make sure that `reset()` clears it out as appropriate.

src/js/player.js
```diff
@@ -3930,6 +3934,113 @@ class Player extends Component {
...
+  /**
+   * Populate the player using a {@link Player~MediaObject|MediaObject}.
+   *
+   * @param  {Player~MediaObject} media
+   *         A media object.
+   *
+   * @param  {Function} ready
+   *         A callback to be called when the player is ready.
+   */
+  loadMedia(media, ready) {
```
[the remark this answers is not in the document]
misteroneill: I asked myself the same thing, but I think sticking with a consistent API is useful.

That said, the other question could be: is this argument even necessary when you could just as easily do this:

```js
player.loadMedia({...});
player.ready(() => {});
```
instead of:

```js
player.loadMedia({...}, () => {});
```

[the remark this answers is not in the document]
BrandonOCasey: OK. I think it makes sense to have a ready argument here, instead of forcing users to do it manually since this function won't have completed until after ready.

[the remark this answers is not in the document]
gkatsev: You can also listen to the `ready` or `loadstart` events that will get triggered as part of this.

src/js/player.js
```diff
@@ -3930,6 +3934,113 @@ class Player extends Component {
...
+    if (src) {
+      this.src(src);
+    }
+
+    if (poster) {
+      this.poster(poster);
+    }
+
+    if (Array.isArray(textTracks)) {
+      this.ready(() => {
```
gkatsev: why add the tracks on ready?

src/js/utils/mimetypes.js
```diff
@@ -17,7 +17,17 @@ export const MimetypesKind = {
   mp3: 'audio/mpeg',
   aac: 'audio/aac',
   oga: 'audio/ogg',
-  m3u8: 'application/x-mpegURL'
+  m3u8: 'application/x-mpegURL',
+  bmp: 'image/bmp',
```
gkatsev: do we need the ones like bmp, icon, tif, tiff? I'd vote to keep just jpg, jpeg, png, svg, webp

test/unit/player-loadmedia.test.js
```diff
@@ -0,0 +1,152 @@
...
+    src: {
+      src: 'foo.mp4',
+      type: 'video/mp4'
+    }
+  });
+
+  assert.strictEqual(this.player.currentSrc(), 'foo.mp4', 'currentSrc was correct');
+});
+
+QUnit.test('loadMedia sets source from an array', function(assert) {
```
gkatsev: might be worth adding a test that has several sources and then `currentSources` should return those multiple sources.

src/js/player.js
misteroneill: I was thinking because it delegates to the tech, but if you don't think it's necessary, it can easily be removed.

src/js/utils/mimetypes.js
misteroneill: Sure, that's fine. I don't imagine people would use those, but wanted to cover common image formats.

gkatsev:
Fixes #4342

gkatsev:
What we have now looks and works great, though, I when playing around on the [test page](https://deploy-preview-5652--videojs-docs.netlify.com/test-example/), I ran `player.getMedia()` without first calling `loadMedia()`. I wonder if we should construct a media object based on current sources and tracks and posters if a `cache_.media` isn't available?

misteroneill:
That makes sense to me.

Issue #5168, opened 2018-05-10
Change remaining-time-display.js so that the time value announced by a screen reader is not negative


OwenEdwards:
## Description
As mentioned in #5157, the convention to display the remaining time in a video as a negative number is purely a visual convention; for screen reader users, where the remaining time has a hidden label, the actual amount of remaining time is positive, not negative.

So the `remaining-time-display.js` component should either add the negative sign in a way that is decorative and not announced by screen readers, or the visual (negative) number should be aria-hidden from screen readers and a hidden positive value should be added.

smbea:
Hello, this is my approach for fixing the issue. I made it so the '-' is added on every time remaining update in a purely decorative way. I was wondering if this fits the what you had in mind for the solution or it should be a more efficient one.

You can see the screenshots of the solution working:

![screenshot 2018-12-05 at 21 12 19](https://user-images.githubusercontent.com/25825387/49544421-bb0fde00-f8d2-11e8-92bb-c2f5f277fe90.png)
![screenshot 2018-12-05 at 21 12 29](https://user-images.githubusercontent.com/25825387/49544422-bb0fde00-f8d2-11e8-8c04-e77d831f05e6.png)

And here are the changes made:

![screenshot 2018-12-05 at 21 13 54](https://user-images.githubusercontent.com/25825387/49544423-bb0fde00-f8d2-11e8-9451-290215665077.png)
![screenshot 2018-12-05 at 21 14 00](https://user-images.githubusercontent.com/25825387/49544424-bba87480-f8d2-11e8-8e10-d5b499be7601.png)

PR #5653, opened 2018-12-05
fix: current time and duration display accessibility with VoiceOver

alex-barstow:
## Description
The desired behavior when navigating any of the TimeDisplay components (current time, duration, remaining time) with the arrow keys with a screen reader is for the singular parent div to receive focus, and the reader should announce the text of each child span in order, ex. "Current Time 0:17" and "Duration 1:10". This is how it works when using JAWS. However, with VoiceOver each individual child span element receives focus and the contents are announced separately:

[Screenshot of 1st span - control text](https://user-images.githubusercontent.com/20497452/49480053-3e5a0280-f7f3-11e8-9152-22956dc3953c.png)
[Screenshot of 2nd span - numerical time](https://user-images.githubusercontent.com/20497452/49480054-3e5a0280-f7f3-11e8-84f3-7307256833f7.png)

According to the ARIA spec (see [here](https://www.w3.org/TR/html-aam-1.0/#details-id-124) and [example 8](https://www.w3.org/TR/wai-aria-1.1/#presentation)), there is no implicit role for span elements, so the contents should be exposed but the elements themselves should be invisible to screen readers. In other words, `<span> Sample Content </span>` should be the same as `<span role="presentation"> Sample Content </span>`. But this is not the case with VoiceOver. As far as I can tell, the desired behavior is only achieved with VoiceOver if each child span is explicitly made presentational, either by assigning each of them the `presentation` role, or by assigning the parent div a role that makes its children presentational (although there doesn't seem to be a role whose semantics fit the purpose of that div)

The first ARIA doc link above shows the Mac OS X Accessibility Protocol mapping for span elements is to the `group` role. I don't know enough about accessibility API mapping to confidently draw conclusions about root cause, but this seems like a possible explanation. Pinging @gkatsev and @OwenEdwards for your expertise on this.

## Proposed Additions
- add `role="presentation"` to each of the two span elements inside TimeDisplay divs
- add `aria-hidden="true` to the TimeDivider (tangential improvement)
- fix inaccurate description of the TimeDisplay component (tangential improvement)

### DIFF
--- a/src/js/control-bar/time-controls/time-divider.js
+++ b/src/js/control-bar/time-controls/time-divider.js
@@ -20,10 +20,12 @@ class TimeDivider extends Component {
   createEl() {
     return super.createEl('div', {
       className: 'vjs-time-control vjs-time-divider',
+      innerHTML: '<div><span>/</span></div>'
+    }, {
       // this element and its contents can be hidden from assistive techs since
       // it is made extraneous by the announcement of the control text
       // for the current time and duration displays
-      innerHTML: '<div aria-hidden="true"><span>/</span></div>'
+      'aria-hidden': true
     });
   }
 
--- a/src/js/control-bar/time-controls/time-divider.js
+++ b/src/js/control-bar/time-controls/time-divider.js
@@ -20,7 +20,10 @@ class TimeDivider extends Component {
   createEl() {
     return super.createEl('div', {
       className: 'vjs-time-control vjs-time-divider',
-      innerHTML: '<div><span>/</span></div>'
+      // this element and its contents can be hidden from assistive techs since
+      // it is made extraneous by the announcement of the control text
+      // for the current time and duration displays
+      innerHTML: '<div aria-hidden="true"><span>/</span></div>'
     });
   }
 
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -8,7 +8,7 @@ import {bind, throttle} from '../../utils/fn.js';
 import formatTime from '../../utils/format-time.js';
 
 /**
- * Displays the time left in the video
+ * Displays time information about the video
  *
  * @extends Component
  */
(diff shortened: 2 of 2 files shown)


src/js/control-bar/time-controls/time-divider.js
```diff
@@ -20,7 +20,10 @@ class TimeDivider extends Component {
   createEl() {
     return super.createEl('div', {
       className: 'vjs-time-control vjs-time-divider',
-      innerHTML: '<div><span>/</span></div>'
+      // this element and its contents can be hidden from assistive techs since
+      // it is made extraneous by the announcement of the control text
+      // for the current time and duration displays
+      innerHTML: '<div aria-hidden="true"><span>/</span></div>'
```
gkatsev: rather than it being added at this level, we should make the main time-divider div be aria-hidden, super.createEl may accent a 3rd argument object for properties, if not, you can use Dom.createEl directly and return that.

OwenEdwards:
I've got to say that I don't like this at all.

There is *no* "standard" for how screen readers implement their Virtual Cursor mode (moving through all DOM content with the arrow keys, rather than just moving through active elements with Tab/Shift-Tab), and for a long time VoiceOver has treated `<span>`s as a "breath break" where JAWS has not. Trying to get those two screen readers to behave in the same way is a can of worms that we don't want to open; the screen reader behavior should be consistent with what it does everywhere else on the web, quirks and all, because the users know those quirks. That user _may_ use both JAWS and VoiceOver, but they _may_ not; in either case, they know how each one behaves much better than non-users do.

This PR is also using ARIA in a non-standard way that may get flagged by conformance checking and accessibility checking tools. According to https://www.w3.org/TR/html-aria/, the role of a `<div>` and the role of a `<span>` are identical; it's just that VoiceOver treats them differently in its Virtual Cursor mode, which it's entirely free to do. It's definitely inconvenient that adding spans for all kinds of purposes in HTML causes VoiceOver to break up what was previously a single "breath" into multiple "breaths", but I'm not comfortable with this hack to get around it.

OwenEdwards:
@alex-barstow I try to keep this work separate from my "day job", but I talked it over with some of my smartest colleagues, and I'm going to admit that I was wrong. In *general* there are real issues about applying ARIA roles to HTML elements where the role matches the implicit role of the element, and I thought that this was true in this case. I now understand that it's not - a `<span>` does not have an implicit role, rather than having an implicit role of `presentation`. So you're right - this isn't "non-conforming", and that example (8) does specifically use the construct and have that note.

So I change my review, with my appologies. And yes, I do think that hiding the "/" from screen readers makes sense, in the same way that I recommended in #5168 that the "-" be hidden from screen readers; as you say, the additional information for screen reader users makes these visual elements redundant.

alex-barstow:
Tested with JAWS, VoiceOver, and NVDA -- I'm seeing the correct behavior with all three

OwenEdwards:
We might want to make similar changes elsewhere, including in #5655.

Issue #2999, opened 2016-01-13
Add touch support detected class


Soviut:
This is a bit of a TODO item as I might do a PR for this if it seems like a good idea.

I've created a theme that, on desktop, shows controls when the user becomes inactive and hides them when inactive. This is problematic on touch devices as my styles conflict with the ability to tap the player to toggle controls and they keep hiding when the `inactive` timer kicks in.

To cope with this, I've added a basic touch detection test which adds a `vjs-touch-supported` class to the player div.

``` javascript
if (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {
  player.addClass('vjs-touch-supported');
}
```

This seems to work well as Chrome on desktop, which supports touch events but doesn't trigger them, fails this test while Chrome on Android passes, as expected. It allows me to negate certain styles using

``` css
.vjs-user-active:not(.vjs-touch-supported) {
  ...
}
```

I propose that a test/class like this be added to allow for these cases where it isn't so much mobile that we're trying to detect, but the presence of a touch pointer.

PS: Unfortunately, the pointer media queries aren't widely supported enough, otherwise this whole thing could be solved in pure CSS.

PR #5662, opened 2018-12-08
Added keycode package and wrapper function to keyboard events handling

tiagofragoso:
## Description
Fixes issue videojs/video.js#3022 
Now using keycode package and wrapper function instead of 'event.which' checks for keyboard events handling

## Specific Changes proposed
Added keycode package to improve keyboard handling
Created a wrapper function at src/js/utils/keyboard-handler.js which allows for a simpler syntax when calling a sequence of key comparisons by allowing an array parameter containing key names

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [x] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
- [ ] Reviewed by Two Core Contributors

PR #5663, opened 2018-12-08
Add touch supported class for touch supporting devices

tiagofragoso:
## Description
Fixes issue videojs/video.js#2999 

Programatically adding `vjs-touch-enabled` CSS class to Player when the device has touch support.
This enables custom player styling for touch devices as mentioned in the linked issue

## Specific Changes proposed
Adding `vjs-touch-enabled` CSS class in touch devices (based on the value of `browser.TOUCH_ENABLED`)

Fixed previous unit test for class application (which always passed) and added a unit test for the opposite case (make sure the class is not added when the device does **not** support touch)

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
Huh, I wonder why this is was commented out.

gkatsev:
Huh, I wonder why the netlify is breaking...

gkatsev:
Restarting the netlify build, we'll see if it fixes it. If it doesn't, it's not a big deal, since it's just to make it easier for us to test with.

Yeah, especially since given that we don't use the class currently, just uncommenting it is fine. We should probably add another issue to track the window move or at least add a comment in above adding the class saying that it won't update if touch support is modified after the player is initalized.

gkatsev:
Some of these tests are pretty old, we try and update them to make it better if the change is pretty small, like this one.

miguelpduarte:
@gkatsev Updated the tests as requested.

On the subject of the tracking of the window, I have done some further research and have not found anything relevant to go off of (not even any experimental browser API in MDN), so what I suggest is the creation of the relevant issue, seeing as comments documenting the behaviour can be missed (as ocurred in this case)

gkatsev:
Thanks! Yeah, an issue for this would be great. I think adding it in the comment (perhaps with a link to the issue) would also be good.

miguelpduarte:
@gkatsev Created issue #5683 as requested, and added it to the comment in player.js, before adding the class.

Would this PR be ready to be accepted? We are working on this open source project in the context of a university course, and it would be beneficial if this PR could be approved (should not need to be merged, just have its changes accepted).

gkatsev:
Thanks! This PR LGTM, though, we require two approvals for outside contributors, though, I don't see it not being accepted.

That course sounds interesting, can you tell me more about it?

miguelpduarte:
@gkatsev Thanks a lot!

We are from the Faculty of Engineering of the University of Porto (FEUP), and the course is called ESOF, which is a portuguese acronym for Software Engineering. ([This is a link to the course page, which is in portuguese only, unfortunately](https://sigarra.up.pt/feup/pt/ucurr_geral.ficha_uc_view?pv_ocorrencia_id=420001))

In this course we learn about common and historical software development pitfalls (we work on a research paper about a historical software bug), software development methodologies (such as Scrum and Kanban, on which we also prepared a paper and a class presentation) and in the third part of the course each class chooses an open source project to contribute to.

That is the reason you might have seen some more activity in terms of outside contributors lately - for example #5662 is also from my group, and #5629 is from another group in my class.

miguelpduarte:
@gkatsev We can try and merge the `master` branch then, we'll have to update our fork first and then we'll get to it!

miguelpduarte:
@gkatsev Done! That was easier than expected, I didn't know that updating a fork from an upstream was that simple haha

Let's hope this fixes the netlify deploy then

gkatsev:
Yeah, it's not too bad.
And that fixed netlify and the test example is live https://deploy-preview-5663--videojs-docs.netlify.com/test-example/ and I was able to verify that the class was added on iOS and not in desktop chrome!

gkatsev:
This PR is good to go now when we start merging for the next minor release. We may want to make a couple patch releases first.

gkatsev:
BTW, can I ask why Video.js was chosen?

PR #5671, opened 2018-12-11
Move the "-" in remaining-time-display.js so it is not announced by screen readers. Fixes #5168

### LINKED ISSUE #5168: Change remaining-time-display.js so that the time value announced by a screen reader is not negative

smbea:
## Description
Fixes issue videojs/video.js#5168  @gkatsev 
Remaining time is now a positive value by it's own and the negative sign is added decoratively.

## Specific Changes proposed
- Removed negative sign in when returning the remaining time
- Added the negative ('-') sign to the number's container and make it 'aria-hidden' so screen readers can't read it

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/time-controls/remaining-time-display.js
+++ b/src/js/control-bar/time-controls/remaining-time-display.js
@@ -43,7 +43,7 @@ class RemainingTimeDisplay extends TimeDisplay {
    * @return {Element}
    *         The element that was created.
    */
-  createEl(plainName) {
+  createEl() {
     const el = super.createEl();
 
     el.insertBefore(Dom.createEl('span', {}, {'aria-hidden': true}, '-'), this.contentEl_);
--- a/src/js/control-bar/time-controls/remaining-time-display.js
+++ b/src/js/control-bar/time-controls/remaining-time-display.js
@@ -48,7 +48,7 @@ class RemainingTimeDisplay extends TimeDisplay {
    */
   formatTime_(time) {
     // TODO: The "-" should be decorative, and not announced by a screen reader
-    return '-' + super.formatTime_(time);
+    return super.formatTime_(time);
   }
 
   /**
--- a/src/js/control-bar/time-controls/remaining-time-display.js
+++ b/src/js/control-bar/time-controls/remaining-time-display.js
@@ -3,6 +3,8 @@
  */
 import TimeDisplay from './time-display';
 import Component from '../../component.js';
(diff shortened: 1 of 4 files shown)


smbea:
Fixes issue #5168

src/js/control-bar/time-controls/time-display.js
```diff
@@ -51,6 +51,10 @@ class TimeDisplay extends Component {
 
     this.updateTextNode_();
     el.appendChild(this.contentEl_);
+    if (className === 'vjs-remaining-time') {
```
gkatsev: this might be better to do inside of remaining-time-display itself. We can override `createEl()`, call `super.createEl()` in it, then insert this span there before returning the el. That will keep all the remaining time display stuff together.

smbea: @gkatsev In my last commit i followed your suggestion and created a createEl for the remaining-time-display. I also deleted the function formatTime on that file as it isn't being used anymore

src/js/control-bar/time-controls/remaining-time-display.js
```diff
@@ -36,19 +38,16 @@ class RemainingTimeDisplay extends TimeDisplay {
...
-   *
-   * @private
+   * @return {Element}
+   *         The element that was created.
    */
-  formatTime_(time) {
-    // TODO: The "-" should be decorative, and not announced by a screen reader
-    return '-' + super.formatTime_(time);
+  createEl(plainName) {
+    const el = super.createEl();
```
OwenEdwards: Since this method accepts a `plainName`, it should pass it upwards as `super.createEl(plainName);`.

gkatsev: looking at the super component, it seems like `plainName` isn't used anywhere. We can still make the change if you think it's best to have it in (or remove `plainName` from the TimeDisplay component).

gkatsev: I went ahead and removed `plainName` from the remaining time and time display components.

PR #5676, opened 2018-12-12
fix: make sure sources, volume, and playback rate are reset along with the player

### LINKED ISSUE #5675: Player#reset does not dispose cached sources.
## Description
`Player#reset` does not dispose cached sources.

## Steps to reproduce
1. Create a player and set a source
2. Call `player.reset()`
3. Check the output of `player.currentSources()` or other source-getter methods

### Expected
Should see the `null`-equivalent return value (e.g. `[]` for `currentSources()`)

### Actual
See the cached sources

### Error output
None

misteroneill:
Fixes videojs/video.js#5675 

This PR will fully clear the `player.cache_` when calling `Player#reset`.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2012,11 +2012,19 @@ class Player extends Component {
    */
   resetCache_() {
     this.cache_ = {
+
+      // Right now, the currentTime is not _really_ cached because it is always
+      // retrieved from the tech (see: currentTime). However, for completeness,
+      // we set it to zero here to ensure that if we do start actually caching
+      // it, we reset it along with everything else.
+      currentTime: 0,
+      duration: NaN,
       lastVolume: 1,
       lastPlaybackRate: this.defaultPlaybackRate(),
       src: '',
       source: {},
-      sources: []
+      sources: [],
+      volume: 1
     };
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -401,17 +401,13 @@ class Player extends Component {
       this.languages_ = Player.prototype.options_.languages;
     }
 
-    // Cache for video property values.
-    this.cache_ = {};
+    this.resetCache_();
 
     // Set poster
     this.poster_ = options.poster || '';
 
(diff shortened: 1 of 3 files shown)


test/unit/player.test.js
```diff
@@ -1460,6 +1462,39 @@ QUnit.test('player#reset loads the first item in the techOrder and then techCall
...
+  assert.strictEqual(player.volume(), 0.2, 'volume is correct');
+  assert.strictEqual(player.lastVolume_(), 0.2, 'lastVolume_ is correct');
+
+  player.reset();
+
+  assert.strictEqual(player.currentSrc(), '', 'currentSrc is correct');
+  assert.deepEqual(player.currentSource(), {}, 'currentSource is correct');
+  assert.deepEqual(player.currentSources(), [], 'currentSources is correct');
+  assert.strictEqual(player.playbackRate(), 1, 'playbackRate is correct');
+  assert.strictEqual(player.volume(), 0, 'volume is correct');
```
misteroneill: This is `0` because of `TechFaker`'s default.

src/js/player.js
```diff
@@ -2009,6 +2002,24 @@ class Player extends Component {
...
+   *
+   * Using this function outside the player constructor or reset method may
+   * have unintended side-effects.
+   *
+   * @private
+   */
+  resetCache_() {
+    this.cache_ = {
+      lastVolume: 1,
+      lastPlaybackRate: this.defaultPlaybackRate(),
```
BrandonOCasey: Should we set `currentTime`, `duration`, `selectedLanguage` to null/empty string here too?

misteroneill: I left those out because it appeared they got set in their getter/setters, but it probably makes sense to set a default for everything here.

misteroneill: Turns out setting `selectedLanguage` in here has all kinds of side effects. Probably best left to another PR.

PR #5677, opened 2018-12-12
docs(live): Add a guide for the live ui and live api

BrandonOCasey:
Adds a guide for the new live API and information on switching between the new and old live ui.

### DIFF
--- a/docs/guides/live.md
+++ b/docs/guides/live.md
@@ -1,0 +1,105 @@
+# The live user interface and API in Video.js
+> Note the the "old" live user interface is currently the default, see the section on [the new user interface](#the-new-user-interface) for information on setting that up.
+
+
+## The default live user interface
+The default user interface hides the `ProgressControl` component on the controlbar and shows the `LiveDisplay` component when Video.js detects that the video that it is playing is live (via a `durationchange` event).
+
+> Note: It does this by adding the `vjs-live` class to the player and the showing/hiding of components is all handled in css.
+
+This makes the player have now progress/seek bar and display text indicating that the player is live. This will then be switched back to normal if a non-live video is switched to (via another `durationchange`) event.
+
+To view a sample of this user interface please:
+1. clone the repository, and cd into that directory
+2. run `npm install` or `npm ci` to install all necessary packages
+3. run `npm start` to start the local server
+4. open `localhost:9999/sandbox/live.html`
+
+## The new user interface
+> Note: This user interface is currently disabled on Android due to the live HLS implementation not supporting seekable ranges during live streams.
+
+The new user interface is currently opt-in to prevent breaking backwards compatablity. We feel that the new user interface is much better and it will likely become the new default in the next major version. If you want to use the new user interface you will have to pass `{liveui: true}` during player setup. This can be done in two ways:
+
+During `data-setup`
+```html
+  <video-js data-setup="{liveui: true}">
+  </video-js>
+```
+
+Using the `videojs` function
+
+```js
+var player = videojs('some-player-id', {liveui: true});
+```
+
+The new user interface shows the `ProgressControl` component on the controlbar, hides the `LiveDisplay` component, and shows the new `SeekToLive` component when Video.js detects that the video that it is playing is live (via a `durationchange` event). Along with the `ProgressControl` update we also updated all the time tooltips on the player to indicate a negative number from the live current time, rather than seeking to a specific time.
+
+> Note: It does this by adding the `vjs-live` and `vjs-liveui` class to the player and the showing/hiding of components is all handled in css.
+
+The new live user interface shows the progress/seek bar and lets the user seek backwards/forwards within the live window. Next it adds a button, via the `SeekToLive` component that can be clicked when the user is behind live that will seek to the live current time. That same button indicates if the `currentTime` of the player is live via a grey circle when not live and a red circle when live.
+
+To view a sample of this user interface please:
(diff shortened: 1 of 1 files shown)


docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+1. clone the repository, and cd into that directory
+2. run `npm install` or `npm ci` to install all necessary packages
+3. run `npm start` to start the local server
+4. open `localhost:9999/sandbox/liveui.html`
+
+
+## LiveTracker
+> Note: this component can be turned off by passing `liveTracker: false` to the player during initialization.
+
+Along with the new liveui we implemented an API that can be used ragardless of which user interface is in use. This API is a child of the player and should be on the player at `player.liveTracker`. `LiveTracker` provides several useful helper functions and events for dealing with live playback, all of which are used and tested internally. Internally this component keeps track of the live current time through a function that runs on a 30ms interval.
```
misteroneill: - `ragardless` -> `regardless`

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+4. open `localhost:9999/sandbox/live.html`
+
+## The new user interface
+> Note: This user interface is currently disabled on Android due to the live HLS implementation not supporting seekable ranges during live streams.
+
+The new user interface is currently opt-in to prevent breaking backwards compatablity. We feel that the new user interface is much better and it will likely become the new default in the next major version. If you want to use the new user interface you will have to pass `{liveui: true}` during player setup. This can be done in two ways:
+
+During `data-setup`
+```html
+  <video-js data-setup="{liveui: true}">
```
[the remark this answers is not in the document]
BrandonOCasey: yeah I think it does, good call.

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
+# The live user interface and API in Video.js
+> Note the the "old" live user interface is currently the default, see the section on [the new user interface](#the-new-user-interface) for information on setting that up.
+
+
+## The default live user interface
+The default user interface hides the `ProgressControl` component on the controlbar and shows the `LiveDisplay` component when Video.js detects that the video that it is playing is live (via a `durationchange` event).
```
gkatsev: the check is actually whether the duration is infinite, (or not -1). It just happens during a `durationchange` event.

gkatsev: There's also special handling for Android but we don't need to mention that here.

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+
+> Note: It does this by adding the `vjs-live` class to the player and the showing/hiding of components is all handled in css.
+
+This makes the player have hide the  progress bar, seek bar, and display text indicating that the player is live. All of those will be shown again if a non-live video is switched to (via another `durationchange` event).
+
+To view a sample of this user interface please:
+1. clone the repository, and move into that directory
+2. run `npm install` or `npm ci` to install all necessary packages
+3. run `npm start` to start the local server
+4. open `localhost:9999/sandbox/live.html`
```
gkatsev: would be goo to add `http://` here

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+
+## The new user interface
+> Note: This user interface is currently disabled on Android due to the live HLS implementation not supporting seekable ranges during live streams.
```
gkatsev: We should suggest they use override native with VHS here.

docs/guides/live.md
gkatsev: below as well

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+### startTracking() and stopTracking()
+These functions can be called to arbitrarily start/stop tracking live playback. Normally these are handled by automatically when the player triggers a `durationchange` with a duration of `Infinity`. You won't want to call them unless you are doing something fairly specific.
+
+### seekableEnd()
+This function gets the time in seconds of the furthest seekable end. For instance if we have an array of seekable `TimeRanges` where the first element in the array is the `start()` second and the last is the `end()` second:
+
+```js
+// seekable index 0: 0 is start, 1 is end
+// seekable index 1: 2 is the start, 3 is the end
+const seekableExample [[0, 1], [2, 3]];
```
gkatsev: missing an `=`

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+This function would return `3` as that is the furthest seekable point for the current media.
```
gkatsev: "This function would return `3`" -> `seekableEnd()` would return `3`

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+
+> Note: that if Infinity is anywhere in seekable end, this will return Infinity
+
+### seekableStart()
+This function gets the time in seconds of the earliest seekable start. For instance if we have an array of seekable `TimeRanges` where the first element in the array is the `start()` second and the last is the `end()` second:
+
+```js
+// seekable index 0: 0 is start, 1 is end
+// seekable index 1: 2 is the start, 3 is the end
+const seekableExample [[0, 1], [2, 3]];
```
gkatsev: missing `=`

docs/guides/live.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+This function would return `0` as that is the first seekable point for the current media.
```
gkatsev: `seekableStart()` instead of `this function`

Issue #5557, opened 2018-11-06
Clicking on a menu button causes the first menu item to change background color


gjanblaszczyk:
## Description
The first captions in the language menu change the background color after click on the captions button on the control bar.

![videojsmenuclickfocusissue](https://user-images.githubusercontent.com/5346546/48101526-9e2ab100-e227-11e8-9bdb-a2e004570430.gif)

## Steps to reproduce
1. Go to the focus-visible sandbox page
2. Click the captions button on the control bar.
3. You will see that first element in the menu changes the background color.

### Expected
The first element in the menu doesn't change the background color.
### Actual
The first element in the menu changes the background color.

#### videojs
Latest
#### browsers
all
#### OSes
all

gkatsev:
Closed by #5558

PR #5679, opened 2018-12-12
Fixed Issue-5557

joaomalheiro:
## Description
Provides a better solution to issue #5557. Instead of removing the background on the CSS file the focus is now attributed to menu-item that is currently selected. 
@gkatsev 

## Specific Changes proposed
Added a function that sets the focus of the menu to the item that is currently selected after receiving the click event.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5865,12 +5865,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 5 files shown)


BrandonOCasey:
Seems like this pull request is currently failing due to linting errors. Can you run `npm run lint -- --errors` and fix the errors that are reported?

joaomalheiro:
I fixed the lint errors in the last commit.
@gkatsev The issue had been fixed by hiding the focus with css, we provided a fix that would actually focus only on the selected and move with the keyboard arrows. We know it was already closed but we are working on a class project on which we had to fix an issue and we chose this one before it was closed so we decided to provide a different solution.

gkatsev:
Ah, I see. To me this seems like it hides the problem rather than fixes it. The issue is about the different expectations of mouse and non-mouse users. A mouse user wouldn't expect focus styles to show up where non-mouse users would and need. The current fixes uses the focus-visible selector (or polyfill) to hide the style for mouse users.
By focusing the currently selected option, to me it seems like hiding the issue as the focus styles wouldn't be noticeable compared to the selected styles.

gkatsev:
However, whether we should focus the selected option is actually a good question that is not related to issue #5557. Any thoughts on that @OwenEdwards?

OwenEdwards:
Focusing the currently-selected menu item seems nice, but is not the expected behavior for a menu. It will confuse screen reader users.

gkatsev:
Thanks @OwenEdwards, that's good to know.
Unfortunately, this means we can't accept this contribution as it would violate user expectations. Thanks for making the PR and having a conversation with us, though.

Issue #4683, opened 2017-10-24
player reset() should reset UI state of control bar etc


Force-Wj:
## Description
when i use reset() api,   controlBar UI not reset

## Steps to reproduce
1. playing video
2. call reset function

gkatsev:
Can you elaborate on what you mean by the control bar UI not being reset?

Force-Wj:
when i execute player.reset() , the vjs-control-bar has not be reset. The status of play buttom is playing, the progress-control UI is not right too

gogoku:
Any update on this issue?, i am still facing the same issue
#3816 trying to use this solution to update the video player

gkatsev:
No update yet, we should reset the control when `reset` is called, but its main purpose was to reset the state of the techs, which it does. Any help with this would be appreciated.

gogoku:
i guess calling dispose to remove the player from the dom and adding again is the right way to do it ,
Can you please tell me , What is the case that reset tech is used for?

PR #5684, opened 2018-12-13
Feature/reset player ui #4683

reeckset:
## Description
Solves videojs/video.js#4683 

## Changes
Adds a method in player for resetting its UI (progress bar, volume, playback rate, remaining time display). This is achieved through 3 other new methods to keep a more modular structure. Also created unit tests for this feature, which are passing at the moment.

Besides the mentioned changes, class RemainingTimeDisplay and the player's tech reset unit tests were also modified to fully support the new added changes.

## Requirements Checklist
- [x] Feature implemented
- [x] Tested in multiple browsers
- [x] Test cases passed


src/js/player.js
```diff
@@ -2931,6 +2931,45 @@ class Player extends Component {
...
+    this.resetPlaybackRate();
+    this.resetVolumeBar();
+  }
+
+  /**
+   * Reset tech's progress so progress bar is reset in the UI
+   */
+  resetProgressBar() {
+    this.currentTime(0);
+    if (isNaN(this.duration())) {
```
gkatsev: as part of https://github.com/videojs/video.js/pull/5348, duration can now be NaN, and it's probably more accurate as well.

src/js/player.js
```diff
@@ -2931,6 +2931,45 @@ class Player extends Component {
...
       this.trigger('playerreset');
     }
+    this.resetControlBarUI();
+  }
+
+  /**
+   * Reset Control Bar's UI by calling sub-methods that reset
+   * all of Control Bar's components
+   */
+  resetControlBarUI() {
```
gkatsev: these methods should be suffixed with an `_` to show that they're "private" or shouldn't be used directly outside of Video.js.

gkatsev:
Also, package-lock changes should probably be backed out.

xRuiAlves:
Thank you for the input, we will fix these issues ASAP!

src/js/player.js
```diff
@@ -2931,6 +2931,45 @@ class Player extends Component {
     if (isEvented(this)) {
       this.trigger('playerreset');
     }
+    this.resetControlBarUI();
```
misteroneill: I wonder if this should come before the event being triggered above? It seems like we'd want to make our changes before we notified that we had reset the player.

xRuiAlves: It does makes sense. However, since the resetting of the control bar UI itself triggers certain tech events, couldn't the player reset immediately after cause, in certain conditions, a change of behavior in how the triggered events are handled?

src/js/player.js
reeckset: Thank you for the input. The reason why i added this is that when the duration is NaN, the remaining-time-display does not change or update. Hence setting it to 0. Is there any other way to fix this?

src/js/player.js
reeckset: @misteroneill that makes perfect sense. I have now moved it to before the event is triggered.

reeckset:
@gkatsev I think I solved all the problems with this pull request, with the exception of the NaN one because of the above mentioned behaviour. Any other issues with it please let me know so it can be accepted.

xRuiAlves:
@misteroneill we have implemented the changes you requested. Do you think de PR would eligible to be accepted?

src/js/player.js
```diff
@@ -2928,11 +2928,50 @@ class Player extends Component {
...
+
+  /**
+   * Reset tech's progress so progress bar is reset in the UI
+   */
+  resetProgressBar_() {
+    this.currentTime(0);
+    if (isNaN(this.duration())) {
+      this.duration(0);
+    }
+    this.tech_.trigger({ type: 'durationchange', target: this.tech_, manuallyTriggered: true });
```
gkatsev: According to the spec, we shouldn't be triggering a `durationchange` if the duration is NaN.

src/js/player.js
gkatsev: I think it's more correct to set the duration to `NaN`. We should call duration display's update directly here instead.

src/js/player.js
```diff
@@ -2928,11 +2928,50 @@ class Player extends Component {
...
+    this.tech_.trigger({ type: 'timeupdate', target: this.tech_, manuallyTriggered: true });
```
gkatsev: This PR will likely follow https://github.com/videojs/video.js/pull/5676, I'm not sure that we'd need to trigger `timeupdate` with the changes in #5676.

test/unit/reset-ui.test.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-env qunit */
+import TestHelpers from './test-helpers.js';
```
gkatsev: we should add a `QUnit.module('player reset-ui');` call here.

test/unit/reset-ui.test.js
```diff
@@ -0,0 +1,54 @@
...
+  const player = TestHelpers.makePlayer();
+
+  player.currentTime(20);
+  player.trigger('timeupdate');
+  player.resetProgressBar_();
+  assert.ok(
+    /0/.test(player.controlBar.progressControl.seekBar.playProgressBar.el().offsetWidth),
+    'progress bar is reset to width 0%'
+  );
+  assert.ok(
```
gkatsev: this should be `assert.equal`

test/unit/reset-ui.test.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-env qunit */
+import TestHelpers from './test-helpers.js';
+
+QUnit.test('Calling resetProgressBar player method should place progress bar at 0% width', function(assert) {
+  const player = TestHelpers.makePlayer();
+
+  player.currentTime(20);
+  player.trigger('timeupdate');
+  player.resetProgressBar_();
+  assert.ok(
```
gkatsev: this should be `assert.equal`

test/unit/reset-ui.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+QUnit.test('Calling resetPlaybackRate player method should place play rate at 1x', function(assert) {
+  const player = TestHelpers.makePlayer({techOrder: ['html5']});
+
+  player.playbackRate(2);
+  player.handleTechRateChange_();
+  player.resetPlaybackRate_();
+  const defaultRate = player.defaultPlaybackRate();
+
+  assert.ok(
```
gkatsev: these two should be `assert.equal`

src/js/player.js
reeckset: @gkatsev the thing is if I set it to NaN the remaining time display does not change. Even when calling duration display's update directly.

xRuiAlves: From our analysis, PR #5348 would make setting the duration to NaN update the duration display. Therefore, if this PR follows #5348, setting the duration to NaN will trigger the update in the remaining time display

gkatsev: #5348 is in 7.4.0 and 7.4 is about to be promoted to latest. I can confirm that calling `updateContent` on the duration display after setting the duration to `NaN` will change the duration display to be `-:-`. However, the remaining time display does a null check on the duration. This probably needs to be updated to verify that the duration is a number. So, https://github.com/videojs/video.js/blob/f0ba1f5b2affb227cbe69149b2abcfbad8469d16/src/js/control-bar/time-controls/remaining-time-display.js#L64-L66
probably should be something like:
```js
if (typeof this.player_.duration() !== 'number') {
  return;
}
```

xRuiAlves: Allright, we will also add that change to the PR then. It should be submitted in a couple of hours. Thank you for all the help :)

reeckset: @gkatsev I think it's done. Anything else please let me know!

gkatsev:
I guess the duration automatically gets reset to NaN as part of the reset process right now?

Also, can `resetControlBarUI` be made "private" by appending the `_` suffix? https://github.com/videojs/video.js/pull/5684/files#r241475268

However, this is more or less good to go.

reeckset:
@gkatsev yes it does. I just added the suffix. Ready to go!

xRuiAlves:
Thanks @gkatsev ! Any changes we can help in, let us know

gkatsev:
Looks like there wasn't much conflicts after merging in https://github.com/videojs/video.js/pull/5676. I've resolved the conflict via github.

gkatsev:
Thanks all!

matt-snider:
@reeckset @imnotteixeira @xRuiAlves I just tried this out in [v7.5.0](https://github.com/videojs/video.js/releases/tag/v7.5.0). Looks good, but one thing I noticed is that the loading spinner isn't hidden. So for example, if the content was still loading when `reset()` was called, it will remain there indefinitely. I saw that there is also #5629 which is open so maybe it's meant to be addressed there?

PR #5686, opened 2018-12-14
fix(fs): make sure there's only one fullscreenchange event

### LINKED ISSUE #5685: player fullscreenchange firing twice
In browsers that support the unprefixed fullscreen API (FF 64+ GC71+), the fullscreenchange event is firing twice every time fullscreen is toggled.

https://jsfiddle.net/jbrhq4s1/1/

gkatsev:
Before the fullscreen API was un-prefixed, we listened to triggered a
`fullscreenchange` event on the player manually. This worked fine
because a prefixed `fullscreenchange` event was triggered on the player
element itself. But when it was unprefixed, we ended up with two events
now, the unprefixed event and the one we triggered.

Instead, we should listen the event the browser supports and re-trigger
it as the unprefixed event as necessary.

We also make sure that the handler gets calls when the document level
fullscreenchange handler gets called so that it is cancelled properly.

Fixes #5685.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -21,7 +21,7 @@ import toTitleCase, { titleCaseEquals } from './utils/to-title-case.js';
 import { createTimeRange } from './utils/time-ranges.js';
 import { bufferedPercent } from './utils/buffer.js';
 import * as stylesheet from './utils/stylesheet.js';
-import FullscreenApi from './fullscreen-api.js';
+import FullscreenApi, {prefixedAPI as prefixedFS} from './fullscreen-api.js';
 import MediaError from './media-error.js';
 import safeParseTuple from 'safe-json-parse/tuple';
 import {assign} from './utils/obj';
@@ -1939,14 +1939,14 @@ class Player extends Component {
    * @listens Player#MSFullscreenChange
    * @fires Player#fullscreenchange
    */
-  handleFullscreenChange_(event = {}) {
+  handleFullscreenChange_(event = {}, retriggerEvent = true) {
     if (this.isFullscreen()) {
       this.addClass('vjs-fullscreen');
     } else {
       this.removeClass('vjs-fullscreen');
     }
 
-    if (event.type !== 'fullscreenchange') {
+    if (prefixedFS && retriggerEvent) {
       /**
        * @event Player#fullscreenchange
        * @type {EventTarget~Event}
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -2557,12 +2569,8 @@ class Player extends Component {
         // If cancelling fullscreen, remove event listener.
         if (this.isFullscreen() === false) {
           Events.off(document, fsApi.fullscreenchange, documentFullscreenChange);
+          this.handleFullscreenChange_();
```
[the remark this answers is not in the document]
gkatsev: Yup, because this will get called from the document's handler, there's no event on the element itself.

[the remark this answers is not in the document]
misteroneill: Perfect. I thought that was the case.

gkatsev:
Looks like this works great everywhere except IE11.

gkatsev:
Firefox has problems with closing.

gkatsev:
Playground: https://deploy-preview-5686--videojs-docs.netlify.com/test-example/

src/js/player.js
```diff
@@ -524,7 +524,15 @@ class Player extends Component {
...
 
     this.one('play', this.listenForUserActivity_);
-    this.on('fullscreenchange', this.handleFullscreenChange_);
+
+    if (FullscreenApi.fullscreenchange) {
+      this.on(FullscreenApi.fullscreenchange, this.handleFullscreenChange_);
+
+      if (IE_VERSION) {
+        this.on(document, FullscreenApi.fullscreenchange, this.handleFullscreenChange_);
+      }
```
misteroneill: I assume IE does not trigger the previous listener at all? If it did, this would seem to have potential to call `handleFullscreenChange_` twice.

gkatsev: Yeah, as far as I can tell IE11 only triggers the document level event in my testing.

misteroneill: Cool. And I checked in IE11 and it works the same as other browsers.

nackd:
I tried this on a lot of different browsers and devices, including GC and FF from before and after the move to unprefixed. It worked everywhere, except on Android and desktop Firefox before unprefixed, where the event didn't fire. Note that depending on the Android device, that might be the latest release available.

gkatsev:
@nackd that is firefox on android? Also, thanks for checking!

gkatsev:
Thanks! Yeah, it's likely the same issue. I'll take a look.

gkatsev:
@nackd I just pushed a change (c026af80) that I think fixes older firefox, would you be able to give it a try again?

nackd:
Sure, but only on Thursday, @gkatsev.

gkatsev:
@nackd that's fine. We're more or less all on break for the week anyway :)

gkatsev:
@nackd hope you had a good new year, did you have a chance to take a look at this PR? Thanks!

nackd:
I was finally able to get to this and it seems to work great, be it prefixed, unprefixed, desktop or Android Firefox. Thanks a lot, @gkatsev!

PR #5687, opened 2018-12-14
Fix css animation order

webdeveloperpr:
## Description
This PR fixes the animation shorthand order.

Animation shorthand order can be found here:
https://alligator.io/css/animation-shorthand/
https://css-tricks.com/almanac/properties/a/animation/

name | duration | ease-in-out | delay | iteration-count | backwards

BEFORE

```css
.vjs-seeking .vjs-loading-spinner,
.vjs-waiting .vjs-loading-spinner {
  display: block;
  animation: 0s linear 0.3s forwards vjs-spinner-show ;
}
```

AFTER

```css
.vjs-seeking .vjs-loading-spinner,
.vjs-waiting .vjs-loading-spinner {
  display: block;
  animation: vjs-spinner-show 0s linear 0.3s forwards;
}
```
This will fix css minimization issues when using https://github.com/webpack-contrib/mini-css-extract-plugin#minimizing-for-production.

**incorrect vs correct**  css syntax

bug

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_loading.scss
+++ b/src/css/components/_loading.scss
@@ -24,7 +24,7 @@
 .vjs-waiting .vjs-loading-spinner {
   display: block;
   // add a delay before actual show the spinner
-  animation: 0s linear 0.3s forwards vjs-spinner-show;
+  animation: vjs-spinner-show 0s linear 0.3s forwards;
 }
 
 .vjs-loading-spinner:before,


gkatsev:
Interesting, looking at mdn and the spec, it seems like the name comes last, however, reading a bit more into the spec, it doesn't really matter the order. It'll basically try its best to parse things. So, I think that the tool isn't right here. It should be able to parse it out properly. However, it does seem like we already have the name as the first item in a few of the other animation properties, so, we may as well be more consistent.

Issue #5404, opened 2018-08-28
Subtitles menu should close in more cases automatically


Akxe:
## Description
The whole issue is only apparent if menu was opened via clicking/taping the icon as such it will not close on mouseleave.
The subtitles menu should close on multiple occasion. As of now, they only get closed on an other click on the menu icon or any of it's children. I would suggest that all of these are implemented:

- [x] Click/tap on icon that opened the menu.
- [x] Click/tap on any menu item.
- [ ] Tap/click on video.
  - No other action should be done via closing tap/click
  - Thought I think that click should do the same action as tap in this case, I am not completely settled on it.
- [ ] On user inactive
  - I would even call this current behaviour a bug, as once the control bar disappears, user is still able to click it's content. That can lead to opening subtitle settings, thus confusing user.

Note: This is true for all menus (Menu class)

gkatsev:
Actually, mouseleave does currently close it, but I believe that's actually a bug and not the intended behavior.
Tap/click on the video makes sense. Probably should work when clicking anywhere else if it blurs the menu. This also matches the behavior on youtube.
The fact that the control bar is still clickable during user-inactive is a bug that should be fixed (https://github.com/videojs/video.js/issues/5060). Also, in contrast, youtube doesn't hide the control bar at all when a menu is opened. That may be a reasonable alternative as well.

Akxe:
Yeah, tap on video & disable hiding control bar, when any menu is opened seem like a neat solution. But would live with closing on inactivity too, as if user don't do anything, he probably doesn't care...

CoffeyMachine:
Additionally, it would be ideal to make the click behavior consistent across menus. The playback rate menu, for instance, rotates through the available menu options on click. Why doesn't the subtitles menu do this too?

gkatsev:
I think having the subtitles menu switch between subtitles would be unexpected. In addition, it's an accessibility thing. We want a way to open and pin the menu, especially since the captions settings dialog is inside the menu. If anything were to change, it's likely making the playback rate menu behave like the subtitles menu.

CoffeyMachine:
> If anything were to change, it's likely making the playback rate menu behave like the subtitles menu.

I think this is a good option also.

gkatsev:
We'll consider it but it's unlikely to change in the short term.

PR #5465, opened 2018-10-01
fix: vjs-lock-showing class gets removed from menu when no longer hovering on menu-button.

gjanblaszczyk:
## Description
A proposed fix for videojs/video.js#1690 

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [x] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))
- [ ] Reviewed by Two Core Contributors

PR #5688, opened 2018-12-14
feat: Close subtitles menu when clicking the seek bar

Eduardo-LP-Silva:
## Description
Adds a feature to close the subtitles menu when clicking the seek bar in order to increase consistency with the other control bar components. It correlates to issue videojs/video.js#5404 (although not directly mentioned there). 

## Specific Changes proposed
Adds a call to the subtitles menu button 'unpressButton()' function inside the seek bar 'mousedown' event handler.

## Requirements Checklist
- [x] Feature implemented
- [x] Change has been verified in Firefox and Chrome
- [x] Test cases passed

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -230,6 +230,9 @@ class SeekBar extends Slider {
     this.videoWasPlaying = !this.player_.paused();
     this.player_.pause();
 
+    // To close the subtitles menu when clicking the progress bar
+    this.player().getChild('controlBar').getChild('subsCapsButton').unpressButton();
+
     super.handleMouseDown(event);
   }
 
--- a/test/unit/subtitle-menu.test.js
+++ b/test/unit/subtitle-menu.test.js
@@ -1,0 +1,31 @@
+/* eslint-env qunit */
+import TestHelpers from './test-helpers.js';
+import * as Events from '../../src/js/utils/events.js';
+
+QUnit.module('Subtitles Menu');
+
+QUnit.test('should close when clicking the control bar', function(assert) {
+  const player = TestHelpers.makePlayer();
+  const subsCapsButton = player.getChild('controlBar').getChild('subsCapsButton');
+  const seekBar = player.getChild('controlBar').getChild('progressControl').getChild('seekBar');
+
+  const el = subsCapsButton.menuButton_.el();
+  const el2 = seekBar.el();
+
+  // Open menu
+  Events.trigger(el, 'click');
+
+  // Verify menu is now open
+  assert.ok(subsCapsButton.buttonPressed_, 'Button was pressed');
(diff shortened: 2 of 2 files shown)


gkatsev:
Ah, interesting, this is because in `slider.js` we call `preventDefault()` https://github.com/videojs/video.js/blob/f0ba1f5b2affb227cbe69149b2abcfbad8469d16/src/js/slider/slider.js#L149-L158

So, I spent quite a long time spelunking in the code trying to find out where and when it was added and for what reason. I went through the [Paleogene](https://en.wikipedia.org/wiki/Paleogene) [closure compiler period](https://github.com/videojs/video.js/commit/2138d4f2a), the [Cretaceous](https://en.wikipedia.org/wiki/Cretaceous) [concatenate-js period](https://github.com/videojs/video.js/commit/3a32f44f), the [jurrasic](https://en.wikipedia.org/wiki/Jurassic) [single file period](https://github.com/videojs/video.js/commit/442696cfc), all the way back to the [triassic](https://en.wikipedia.org/wiki/Triassic) Video.js 2.0 period: https://github.com/videojs/video.js/commit/31f6fcdbd6c692957f6555ff44091a531c4f992c. The change that added `preventDefault` was made 8 years, 1 month, and 1 week ago and unfortunately, doesn't provide much information as to why. Testing locally, it seems that what this does is mostly prevent the focus outline from appearing around the seek bar and the time tooltips.
Therefore, I wonder if a better fix would be to hide the outline (to match current behavior, caveat is that we can't break accessibility otherwise) another way and then remove the `preventDefault()` call so that [Menu's `handleBlur` can properly close the menu](https://github.com/videojs/video.js/pull/5465)? Would love your input on this @OwenEdwards.

heff:
Wow...that's incredible. What a journey. :) That's a lot to dig through.

I wish I could tell you specifically why that was there. Just preventing the outline feels like silly reason to include that, but maybe. My best guess otherwise is I was being overly safe, in case the slider ever became an element other than a div. But it could have just as easily been something to fix IE or iPad 1.

I would feel safe enough pulling that out with some testing.

OwenEdwards:
I think we need to do a broad review of where preventDefault is and isn’t used. It definitely shouldn’t be used to prevent focus movement now that we have support for focus-visible (as a browser feature or a polyfill).

gkatsev:
I don't think we really use `preventDefault` that much in the first place but commenting it out fixes the menu closing issue and the only side-effect is that the progress bar gets the focus outline, at least as far as I can tell. Would be good to have a PR for that and then some thorough testing.

OwenEdwards:
@gkatsev if you disable `preventDefault()` and then press Space, does the webpage scroll down? Because that's the browser default for Space, so that'd be my concern if `preventDefault()` was removed.

gkatsev:
Nope, doesn't seem like it. But would still require more thorough testing.

gkatsev:
Thanks for opening the other issue @OwenEdwards. Going to close this in favor of #5774.

Issue #5258, opened 2018-06-16
Videojs controls autohide


mysuf:
Hi guys,

I have one suggestion for possible improvement. It is logic behind controls bar showing/hiding. Now, it has really confusing behaviour that gradually increases user's rage and decreases user's experience. It disapears even if cursor is placed on control bar because of 2s hard timeout. So user is going to click, but it fades out meantime but shows again 1s later, because cursor probably trigger event on underlying player's DOM (this actually loops). I would expect behaviour like:

1. cursor moves over player -> user active
2. cursor stops inside player's bounds(outside control bar) -> sets timeout 2s before going inactive
3. cursor stops inside control bar -> nothing happens or increased timeout about 10s
4. cursor leaves player bounds -> sets timeout 2s before going inactive

Thanks

Version: 7.0.3
Player state: playing

gkatsev:
Thanks for the issue. There's definitely room for improvement here and you've provided some interesting ideas. I'm currently working on a roadmap for Video.js and there's already some stuff related to this there, so, definitely going to reference this issue there.

PR #5692, opened 2018-12-17
fix: Control-bar autohide when cursor placed over it #5258

xjoaoalvesx:
## Description
Solves videojs/video.js#5258 

## Specific Changes proposed
Adds two event handlers for 'mouseenter' and 'mouseleave' events on the control-bar.
When the 'mouseenter' is triggered the inactivityTimeout is set to 0 and on 'mouseleave' is set to the default value (2s).

Expected behavior: 

* cursor stops inside control bar -> sets timeout 0s so nothing happens
* cursor leaves player bounds -> sets timeout 2s before going inactive

## Requirements Checklist
- [x] Change has been verified in an actual browser (Chrome, Firefox)
- [x] Tests cases passed

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3444,17 +3444,22 @@ class Player extends Component {
     this.on('mousemove', handleMouseMove);
     this.on('mouseup', handleMouseUp);
 
-    this.on(this.getChild('controlBar'), 'mouseenter', function(event) {
+    const controlBar = this.getChild('controlBar');
 
-      this.cache_.inactivityTimeout = this.options_.inactivityTimeout;
-      this.options_.inactivityTimeout = 0;
+    if (controlBar) {
 
-    });
+      controlBar.on('mouseenter', function(event) {
 
-    this.on(this.getChild('controlBar'), 'mouseleave', function(event) {
-      this.options_.inactivityTimeout = this.cache_.inactivityTimeout;
+        this.player().cache_.inactivityTimeout = this.player().options_.inactivityTimeout;
+        this.player().options_.inactivityTimeout = 0;
 
-    });
+      });
+
+      controlBar.on('mouseleave', function(event) {
+        this.player().options_.inactivityTimeout = this.player().cache_.inactivityTimeout;
+      });
+
+    }
 
     // Listen for keyboard navigation
     // Shouldn't need to use inProgress interval because of key repeat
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -3444,6 +3444,16 @@ class Player extends Component {
     this.on('mousemove', handleMouseMove);
     this.on('mouseup', handleMouseUp);
 
+    this.on(this.getChild('controlBar'), 'mouseenter', function(event) {
+      this.options_.inactivityTimeout = 0;
+
+    });
+
+    this.on(this.getChild('controlBar'), 'mouseleave', function(event) {
+      this.options_.inactivityTimeout = 2000;
```
gkatsev: if someone set a custom inactivity timeout, this will reset that value.

gkatsev: We could store the inactivityTimeout in `this.cache_` and then reset to that value.

xjoaoalvesx: Good point, I now save the value in the cache object on mouseenter and retrieve it on mouseleave

src/js/player.js
```diff
@@ -3444,6 +3444,18 @@ class Player extends Component {
     this.on('mousemove', handleMouseMove);
     this.on('mouseup', handleMouseUp);
 
+    this.on(this.getChild('controlBar'), 'mouseenter', function(event) {
```
gkatsev: we should store the control bar object in a variable (to also re-use below) and do a null-check on it.

Additionally, it would be better to call `.on` directly on the contrlBar itself as the handler will automatically be removed when the control bar is disposed. I don't think the current event handler gets cleared up automatically, currently.

xjoaoalvesx: I'm sorry, I didn't really understand your second point. Wouldn't it be the same if I associated the handler on the immediate return of the getChild function or a variable with its value? Or do you mean something else?

gkatsev: Basically, `this.on(anotherComponent, 'mouseenter', () => {})` vs `anotherComponent.on('mouseenter', () => {})`. I am unsure that the first one will have the `mouseenter` event removed when the component is disposed where-as I'm certain the second one will be done properly.

Ultimately, the code should end up something like:
```js
const controlBar = this.getChild('controlBar');

if (controlBar) {
  controlBar.on('mouseenter', (event) => /* ... */);
  controlBar.on('mouseleave', (event) => /* ... */);
}
```

xjoaoalvesx: I completely understand what you say so now we can see that on the code.

PR #5693, opened 2018-12-17
Fix: duration display when changing video duration or source

DuarteOliveira8:
## Description
Solves [issue #5347](https://github.com/videojs/video.js/issues/5347).
In this pull request, I propose a different way of solving this issue along with more comments in the code explaining my resolution.

## Specific Changes proposed

For case A (duration change), I removed the if condition to check if the time is different than `NaN`. I don't think that condition is necessary because the time display formatting already handles values that are not numbers and displays `'-:-'` instead.

As for case B (video source change), I don't think the events added are needed and, instead, I added a condition to check if `preload` is `none` when changing source and update the duration to the current duration, which should be `NaN` since `preload='none'`. This duration update is necessary because when the duration updates to `NaN`, the `durationchange` event isn't triggered and we must change the duration manually.

## Requirements Checklist
- [x] Bug fixed
- [x] Change has been verified in an actual browser (Firefox)
- [x] Test cases passed 

### DIFF
--- a/src/js/control-bar/time-controls/duration-display.js
+++ b/src/js/control-bar/time-controls/duration-display.js
@@ -51,6 +51,7 @@ class DurationDisplay extends TimeDisplay {
 
   /**
    * Update duration time display.
+   * In case of the duration passed as argument being NaN, the result of the formatted time will be '-:-'.
    *
    * @param {EventTarget~Event} [event]
    *        The `durationchange`, `timeupdate`, or `loadedmetadata` event that caused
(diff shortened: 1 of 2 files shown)


gkatsev:
Seems like this more or less reverts the changes in https://github.com/videojs/video.js/pull/5348 which fixes https://github.com/videojs/video.js/issues/5347. I think we should be following the spec with regards to how it handles triggering of `durationchange` and that means that we shouldn't trigger it if the value is `NaN`. If this is a follow-up to https://github.com/videojs/video.js/pull/5684, it may be better to update that to call the duration display's update method directly.

gkatsev:
I think I'm going to close this as we have a more spec compliant solution as outlined in #5684.

Issue #5233, opened 2018-06-06
Autoplay (muted or unmuted) doesn't work in Chrome with latest video.js


WillMuijrers:
## Description
Using the CDN of the latest video.js version the `autoplay` doesn't seem do to anything in Chrome. I'm well aware of the limitations of Chrome. Even applying `muted` does nothing. 
When using an old version of video.js (i used 6.4.0) the autplay and muted work as aspected.

## Steps to reproduce
```
// version 7.0.4 -> doesn't work 
// version 6.4.0 -> works fine
<video muted class="video-js" id="player"> 
     <source src=http://techslides.com/demos/sample-videos/small.mp4 type=video/mp4>
</video>

var player = videojs("player");

player.ready(function() {
      player.play();
  });
```
### Error output
When muted is **not** applied, one would expect a console error from Chrome. This also does not seem to happen.

dszewczyk1:
I also have a problem with autoplay, in a bit different code, as it should start playing just by adding "autoplay" attribute, not by running it manually.

` <script src="https://vjs.zencdn.net/7.0.4/video.js"></script> // worksw with 6.7.1 
<video class="video-js" data-setup='{ "controls": true, "preload": "auto", "loop": true, "autoplay": true, "muted": true}'> 
     <source src="http://techslides.com/demos/sample-videos/small.mp4" type=video/mp4>
</video>`

It seems to work in Chrome in 6.7.1 for example, but not with any newer version
http://jsbin.com/ribacaxezu/1/edit?html,output

Sizzahandz:
Someone have a fix for this?

gkatsev:
We just merged #5209 which gives a few more tools to get autoplay working. I am able to get autoplay working with latest Video.js but I can also confirm seeing autoplay not working properly in the example that @dszewczyk1 posted.

richardlnnr:
You can use `oncanplay="this.muted=true"` on your video tag

    <video  muted loop autoplay oncanplay="this.muted=true">
        <source src="your src" type="video/mp4">
        Your browser does not support HTML5 video.
    </video>

mysuf:
I can confirm that too. I can also say that it happend somewhere between v6.7.4(works) and v6.10.0(doesnt work) and probably will be somehow related to cloning video tag and its attributes..

ksperling:
I'm running into the same issue (Chrome 67.0.3396.99).

[...]

PR #5698, opened 2018-12-19
fix: autoplay issue on Chrome #5233

claudiolemos:
## Description
This commit fixes the autoplay issue (#5233) on Chrome.

## Specific Changes proposed
In order for the autoplay to work on Chrome, after the video is loaded we check if autoplay is turned on and if the video is paused, thus forcing the video to be played. On tech/HTML5.js, autoplay is forced as the last command

## Requirements Checklist
- [x] Unit Tests passed

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -112,6 +112,12 @@ class Html5 extends Tech {
     this.proxyWebkitFullscreen_();
 
     this.triggerReady();
+
+    // On load, confirms if autoplay is turned on and if the
+    // player is paused, forcing the video to be played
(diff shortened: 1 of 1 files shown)


gkatsev:
We used to have a similar change which was targeted specifically as chrome (which we reverted because we thought things got fixed) but it seems like things got fixed again. If we feel the need to address this in Video.js, though, all signs point to a browser bug, we should consider unreverting https://github.com/videojs/video.js/commit/8d86afdb75c1061c8b340fa8cec4954b65a8c704. Going to close this for now and we can continue discussion in #5233.

Issue #5524, opened 2018-10-24
Change max-height of menus based on responsive classes


gkatsev:
## Description
Currently, it's possible that if you have a lot of items in a menu, it'll require scrolling the menu. The max-height of the menus is designed so that on all sized (except really small) the menu will never go out of bounds. If you have just one more than the max-height, one of the items will get cut off and if your player is large enough, that's uncessary.
We should use use the responsive breakpoint classes to make the height of the menu variable depending on the size of the player. This way, a larger player can have a taller menu. 

## Steps to reproduce
1. Create a player with 6 caption tracks
2. Open up the responsive.html example
3. Use devtools to keep the captions menu open
4. resize the page to see each player size

### Expected
At smaller sizes, the menu should scroll, if the player is rather large, the menu with 6 tracks should not scroll.
The menu should never be taller than the player and exit the player bounds.

### Actual
The menu is always sized to 15em.

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
7.x, 6.x
#### browsers
All modern browsers
#### OSes
All platforms
### plugins
None

PR #5699, opened 2018-12-19
Issue fixed

claudiolemos:
## Description
This commit fixes the subtitle box issue (#5524), where when the height of the player was very small or the number of subtitles was big, the subtitle box would exceed the players height.

## Specific Changes proposed
We fixed this issue by adding responsive classes to the already established breakpoints that display the subtitles menu icon.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/menu/_menu-popup.scss
+++ b/src/css/components/menu/_menu-popup.scss
@@ -16,7 +16,36 @@
   position: absolute;
   width: 100%;
   bottom: 1.5em; // Same bottom as vjs-menu border-top
-  max-height: 15em;
+}
+
+@media (min-width: 321px) and (max-width: 425px) { // small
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks for the PR! However, we don't want to use media queries as these reference the size of the page itself which may or may not have any relation to the size of the player. In https://github.com/videojs/video.js/pull/5496, we added in the concept of breakpoints in the player, which add a particular class to the player depending on the player size (see https://github.com/videojs/video.js/blob/master/docs/guides/layout.md#class-1). We should use these classes to set the max height.

In addition, we should try and use em units as that's more consistent with the rest of the code and we need to keep the previous default max-height.

gkatsev:
Hi, would you be able to make the changes as outlines above? Let me know if you have questions!

PR #5700, opened 2018-12-19
Update player.js

dioguinix:
## Description
Pull request for issue #5604 

## Specific Changes proposed
Changing handleTechDoubleClick_(event) {} function in order to not trigger fullscreen even when the video is double-clicked. This will only happen when the option fullscreenToggle is set to false, of course.

## Requirements Checklist
Improved functionality of disabling fullscreen mode.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1857,10 +1857,12 @@ class Player extends Component {
     );
 
     if (!inAllowedEls) {
-      if (this.isFullscreen()) {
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks for the PR, unfortunately, we can't accept it. The presence of the fullscreen toggle doesn't mean that we don't want this functionality as a custom fullscreen button could have been implemented. In addition, there's a pending PR with an option to disable double-click to fullscreen that will be going in next time we're working on a minor release: #5611.

PR #5702, opened 2018-12-20
fix: remove child from old parent when moving to new parent via addChild

dustin71728:
## Description
A child component may have been assigned to another
parent before assigning that child component to the
new parent via "addChild" method. In this case, the
original parent should remove the child then it can
be safely added back to the new parent. This commit
will keep the parent's "children_" and its DOM
element's child nodes in the consistent state.

## Specific Changes proposed
Remove the specific child component from the old parent component before adding the child component to the new parent

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -58,6 +58,9 @@ class Component {
       this.player_ = player;
     }
 
+    // Hold the reference to the parent component via `addChild` method
+    this.parentComponent_ = null;
+
     // Make a copy of prototype.options_ to protect against overriding defaults
     this.options_ = mergeOptions({}, this.options_);
 
@@ -141,6 +144,8 @@ class Component {
     this.children_ = null;
     this.childIndex_ = null;
     this.childNameIndex_ = null;
+
+    this.parentComponent_ = null;
 
     if (this.el_) {
       // Remove element from DOM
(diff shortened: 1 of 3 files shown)


gkatsev:
So, the reason for this is so that the internal `children` array in our component system stays up-to-date when components move around in the DOM? Otherwise, the previous parent would have a reference to the child component even if the child component wasn't actually a child anymore?

dustin71728:
> So, the reason for this is so that the internal `children` array in our component system stays up-to-date when components move around in the DOM? Otherwise, the previous parent would have a reference to the child component even if the child component wasn't actually a child anymore?
> Would you be able to add a test for this?

Yes, for example, if I want to add  the new menu item of the menu button dynamically, a class can be implemented as follows:
```
export class MutableMenuButton extends MenuButton {
    super(player, options)

    this.items_ = []
  }

  insertItem(item, index) {
    this.items_.splice(index, 0, item)

    // Trigger the `update` method from `MenuButton` to re-render the layout
    this.update()
  }

  // Override the abstract method
  createItems() {
    return this.items_
  }
}

export default MutableMenuButton
```
But what i found is whenever the "MutableMenuButton".inserItem gets called, the existed menu item is removed from the new menu that belongs to the menu button.
After digging into the code, I found that the old, about to be removed menu instance still hold the child component that is already to move to the new menu component.
Because any discarded component will dispose of its children, so the DOM element of the existed menu item attached to the new menu will be erased from the DOM, this is why the new menu can't see the menu item that already in place.

So that is why I commit this change, if you think that is acceptable, I will add some tests. Thanks in advance.

dustin71728:
It is my pleasure. :smile:

PR #5721, opened 2019-01-02
fix(seekbar): don't disable if live tracker's seekable is infinity

gkatsev:
The seekbar is disabled if the seekable end is Infinity and re-enabled
when it isn't. Instead, we should only disable if seekable end is
Infinity and re-enable only if we disabled for the seekable end.

This auto disabling and re-enabling, without this fix, caused issues
with disabling the progress control since the seekbar got re-enabled.

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -39,6 +39,7 @@ class SeekBar extends Slider {
   constructor(player, options) {
     super(player, options);
     this.setEventHandlers_();
+    this.disabledForLive_ = false;
   }
 
   /**
@@ -116,8 +117,12 @@ class SeekBar extends Slider {
     }
 
     if (liveTracker && liveTracker.seekableEnd() === Infinity) {
-      this.disable();
-    } else {
+      if (this.player_.readyState() === 0) {
+        this.disabledForLive_ = true;
+        this.disable();
+      }
+    } else if (this.disabledForLive_) {
+      this.disabledForLive_ = false;
       this.enable();
     }
 
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -39,7 +39,6 @@ class SeekBar extends Slider {
   constructor(player, options) {
     super(player, options);
     this.setEventHandlers_();
-    this.disabledForLive_ = false;
   }
 
   /**
@@ -114,16 +113,6 @@ class SeekBar extends Slider {
 
     if (liveTracker && liveTracker.isLive()) {
       duration = this.player_.liveTracker.liveCurrentTime();
-    }
-
-    if (liveTracker && liveTracker.seekableEnd() === Infinity) {
-      if (this.player_.readyState() === 0) {
-        this.disabledForLive_ = true;
-        this.disable();
-      }
-    } else if (this.disabledForLive_) {
-      this.disabledForLive_ = false;
-      this.enable();
     }
 
     // machine readable value of progress bar (percentage complete)


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -116,8 +117,12 @@ class SeekBar extends Slider {
     }
 
     if (liveTracker && liveTracker.seekableEnd() === Infinity) {
-      this.disable();
-    } else {
+      if (this.player_.readyState() === 0) {
```
ldayananda: How is the readyState === 0 if we already know the seekableEnd?

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -116,8 +117,12 @@ class SeekBar extends Slider {
     }
```
ldayananda: btw mis-indented `}` while you're here 😄

src/js/control-bar/progress-control/seek-bar.js
ldayananda: I suppose the confusing part is that we would want the seekBar to be disabled only when we know we're live right? Rather than when we don't know?

src/js/control-bar/progress-control/seek-bar.js
gkatsev: Github just makes it look that way, it's properly indented in code (and not PRs)
https://github.com/videojs/video.js/blob/3d093ede98f2ad6204ff44c9631b0aa18bbf0cc5/src/js/control-bar/progress-control/seek-bar.js#L114-L116

src/js/control-bar/progress-control/seek-bar.js
gkatsev: Basically, my change is for us to only re-enable the seek bar if it was disabled due to the seekable end being infinity.

src/js/control-bar/progress-control/seek-bar.js
ldayananda: _flips table_

src/js/control-bar/progress-control/seek-bar.js
ldayananda: right, but what does the readyState check add to that? I believe seekable end can also be reported as Infinity on Android (which this case used to also cover)

gkatsev: readyState of 0 means that no data has loaded, so, any value for seekable end is meaningless but the way that method is implemented, it'll still return Infinity. So, without the changes, it means that we would disable the progress control due to seekable end being infinity and then subsequently re-enable it when the video metadata loads. This means that if you disable the progress control yourself, it'll get re-enabled from under you. (We should also potentially check to see if we're currently disabled not via this codepath and then do nothing)

ldayananda: To be more clear, I think `disabledForLive_` is valuable, as an indication that we disabled the seekBar and we should only re-enable again if it is `true`.

However, in the case where metadata takes a while to load we would actually want to disable the seek bar until we know if we are live, assuming we use the flag to check that is why we disabled it. But we _also_ want to disable the seek bar if we're on Android where DVR is not reliable because seekable end can be reported as Infinity.

I'm not sure exactly what the best course of action would be for the scenario where the liveTracker is disabled by a user and we're on Android. Should the progress bar show without allowing seeking?

gkatsev: I think it's fine if the control bar isn't disabled in this case until the metadata is loaded, especially since it matches the current behavior. I think being permissive is a better side to err on in this case.

ldayananda: > I think it's fine if the control bar isn't disabled in this case until the metadata is loaded

the Android case?

ldayananda: That might work out, except for the initial setup, which we could put in an explicit case for.

Throwing out ideas: could we just explicitly have a flag on the liveTracker which determines if the seekbar should be enabled and use that in the constructor for seekBar? That would still allow `seekBar.disable` I believe

gkatsev: Wouldn't that be the same as what we have now? Whether the check is here or in liveTracker, it'll still disable/enable incorrectly.
Also, I don't think that the livetracker should apply to the seekbar until/if we're in a livestream. So, I guess we could also check the duration and if it's not Infinity/-1 ignore it?

gkatsev:
So, spoke with @ldayananda offline.
The issue is that the livetracker affects the seekbar whether we are playing back a live stream or not. Also, this happened every 30ms or so, meaning that if a user tried to disable the progress control, this would potentially come back in and re-enable it. In particular, live tracker defaults seekable end to be Infinity, a reasonable assumption, however, before we end up loading the video, we think that we're at seekable end of infinity and disable the seekbar, which is why just one flag for `disableForLive_` wasn't enough. So, here are some potential directions:
- keep track of what the previous state was and then go back to that state instead of always re-enabling. This likely still needs to keep track of `disableForLive_` for us to be able to distinguish between a "user-generated" disable or a live tracker disable.
- do what we have now and ignore any live tracker information until video metadata has loaded
- ignore live tracker info if we're not in a live video by checking the duration
- some combination of the above but only on android
- move the disable/enable logic out of `update_` and into a `durationchange` handler. Then specifically target live on android. This means that disable/enable happens *once* (as we'll get another `durationchange` if the duration goes from Infinity back to finite number) and a user could call `progressControl.disable()` and we won't disable/enable it from under them.

src/js/control-bar/progress-control/seek-bar.js
gkatsev: https://github.com/videojs/video.js/pull/5721#issuecomment-451307233

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -116,8 +117,12 @@ class SeekBar extends Slider {
...
 
     if (liveTracker && liveTracker.seekableEnd() === Infinity) {
```
BrandonOCasey: Perhaps we should run this code on a `seekableendchange` triggered by the `liveTracker`? That should be the only time this code needs to run.

src/js/control-bar/progress-control/seek-bar.js
BrandonOCasey: We could also check if `player.seekable().length` is non-zero rather than checking `player.readyState()`. Not sure which is better but `player.seekable().length` seems more straightforward to me. Regardless of what we do this section of code could use a comment.

gkatsev:
So, we've decided to just remove the code in question so it doesn't cause any potential issues. We recommend users on Android overrideNative to use our MSE-based playback engine.

Issue #3612, opened 2016-09-14
Menu's have non-component children


nickygerritsen:
Both the default menu-button (https://github.com/videojs/video.js/blob/master/src/js/menu/menu-button.js) and the chapters button (https://github.com/videojs/video.js/blob/master/src/js/control-bar/text-track-controls/chapters-button.js) (might) add a `<li>` to the `children_` array of a menu, while I think children of `Component`s should all be a `Component`. We can (probably) fix this by creating something like a `MenuItemTitle` component.

misteroneill:
I am on board with the idea that we should be able to expect the contents of `children_` to be components.

Probably wouldn't even need a custom component - unless it's doing something beyond what a generic component does - it could create the el and pass it as an option to a generic component.

Perhaps generic components should be able to take a DOM configuration to make this easier? Something like:

```js
new Component(this, {
  dom: {
    tagName: 'li',
    props: {},
    attrs: {}
  }
});
```

nickygerritsen:
That does sound like a good idea IMHO :).

misteroneill:
Or it could be an array to match the `createEl` arguments: `dom: ['li', {}, {}]`.

nickygerritsen:
Yeah that'd make even more sense I guess, then we can just pass it to that function

PR #5722, opened 2019-01-03
fix(menu-button): make menu button title a component

chrisboustead:
## Description
Currently menu titles are not components, but Dom elements instead.  This is an issue specifically for existing code such as in [src/js/menu/menu.js:114](https://github.com/videojs/video.js/blob/3d093ede98f2ad6204ff44c9631b0aa18bbf0cc5/src/js/menu/menu.js#L114,L115) where each child element is expected to be a videojs component (`element.el()`) though when a title is provided to `MenuButton`, it is created as a Dom element in `src/js/menu/menu-button.js:111`

Relevant issue already exists, though it has only popped up as an issue for me after VJS 7.x updates:
videojs/video.js#3612 

```
 /**
   * Called when a `MenuItem` loses focus.
   *
   * @param {EventTarget~Event} event
   *        The `blur` event that caused this function to be called.
   *
   * @listens blur
   */
  handleBlur(event) {
    const relatedTarget = event.relatedTarget || document.activeElement;

    // Close menu popup when a user clicks outside the menu
    if (!this.children().some((element) => {
      return element.el() === relatedTarget;
    })) {
      const btn = this.menuButton_;

      if (btn && btn.buttonPressed_ && relatedTarget !== btn.el().firstChild) {
        btn.unpressButton();
      }
    }
  }
```

## Specific Changes proposed
This update simply adds the created DOM element as a component, rather than the current dom manipulation and children array manipulation.

## Requirements Checklist

### DIFF
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -108,7 +108,7 @@ class MenuButton extends Component {
 
     // Add a title list item to the top
     if (this.options_.title) {
-      const title = Dom.createEl('li', {
+      const titleEl = Dom.createEl('li', {
         className: 'vjs-menu-title',
         innerHTML: toTitleCase(this.options_.title),
         tabIndex: -1
@@ -116,8 +116,9 @@ class MenuButton extends Component {
 
       this.hideThreshold_ += 1;
 
-      menu.children_.unshift(title);
-      Dom.prependTo(title, menu.contentEl());
+      const titleComponent = new Component(this.player_, {el: titleEl});
+
+      menu.addItem(titleComponent);
     }
 
     this.items = this.createItems();


gkatsev:
Huh, I wonder why travis-ci didn't run for this.

chrisboustead:
Not sure! 🤷‍♂️ 

Tests/linting/etc passed locally but please keep me posted if I need to make updates to get that corrected.

gkatsev:
Can you try pushing an empty commit to see if it triggers it? `git commit --allow-empty -m "trigger travis"`

chrisboustead:
No good :( Also tried a fresh PR, with the same result.  Just the one netlify preview check ran.

gkatsev:
weird, I'll take a look tomorrow.

gkatsev:
I uploaded the branch to upstream and github is smart enough to associate the triggers with the PR :)

OwenEdwards:
This is nice! It means that the one-line fix https://github.com/videojs/video.js/pull/5571/files#diff-b16a7a8734cdadf51bb96ac889633515R115 is no longer necessary. My only concern is how thoroughly it needs testing? The menu title feature isn't used very often (it took me testing with a Chapters menu to notice the bug), so would we notice if this change broke something?

chrisboustead:
@OwenEdwards I thought about fixing this issue that in a similar fashion to the line in #5571 also, but decided it was a bit of a bandaid.  I also tested with the chapters menu, as well as my own [plugin](https://github.com/chrisboustead/videojs-hls-quality-selector) which uses the menu button component (where I came across the issue).  

The playback rate control also uses the menu button control, though it does not use a title.

I'll do some searching around other plugins and see if I come across anything interesting.

OwenEdwards:
Thinking about this a little more, why isn't the menu title just a part of the `Menu` component's HTML / `el`, rather than a `child` of the `Menu` and `MenuButton`?

The structure of the menus as a whole needs revising; the top-level component really ought to be a menu (wrapper `<div>`) which contains the `MenuButton` and the `MenuList` which contains `MenuItem`s. I had to change the structure a while ago because the `MenuList` (which is currently called `Menu` was a child of the `MenuButton`, but HTML & ARIA don't allow actionable elements (e.g. `MenuItem`s) to be children of HTML or ARIA `button`s. So now `MenuButton` is a bit of a mess, and has a child `Button` which is the actual "MenuButton". Fixing that would be a significant piece of work, and would probably have to wait for a major version release (v8?), but I think the non-actionable menu title ought not to be a whole `Component` of it's own.

@gkatsev, what do you think?

misteroneill:
That does sound like a topic for an 8.0 release to me, but I agree that the menu system should be revisited.

gkatsev:
Yeah, major restructring like that is definitely a breaking change.
This might be fine to do as a minor update to minimize potential issues, and if someone encounters a problem we can always back it out.

chrisboustead:
Moving the title to be a child of the menu also becomes a little hairy.  

The way titles are supplied currently are through instantiating a `MenuButton` (or a component inheriting a `MenuButton`) with a `title` prop supplied in the options.  The `Menu` is then created via that menu button.

The parent `MenuButton` is then passed to the `Menu` it created as an option `menuButton`, which means the way the title would have to be accessed would be something like `this.options_.menuButton_.title` from within the menu, which just seems kind of dirty.  Also, the `ChaptersButton`, for example, has it's own methods to create titles based on the tracks/etc, so the title in the `Menu` would potentially need to be updated well after `MenuButton` instantiation.

I don't think the current setup is right- as mentioned think a bigger refactor would be the way to go, but this fixes an existing bug without going down a rabbit hole.

OwenEdwards:
@chrisboustead - yeah, that's a good point. I was trying to be *too* clever, I think! This seems like a good fix for the immediate problem, and I'll start looking at a significant re-structure for v8. <cracks knuckles>

PR #5741, opened 2019-01-09
fix: TextTrackMenuItem components should not disable text tracks of different kind(s).

misteroneill:
## Description
While investigating a non-public issue, I happened to notice that metadata tracks were being disabled when switching captions languages. Upon investigation, I found [this code](https://github.com/videojs/video.js/blob/175f77325346af9ec5fa7f090eba8be2076a54e2/src/js/control-bar/text-track-controls/text-track-menu-item.js#L113-L119), which is the culprit. It will disable any track that's not the selected track.

## Specific Changes proposed
Instead of disabling all tracks when switching, only disable tracks of the appropriate kind(s).

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/text-track-controls/text-track-menu-item.js
+++ b/src/js/control-bar/text-track-controls/text-track-menu-item.js
@@ -102,7 +102,8 @@ class TextTrackMenuItem extends MenuItem {
       return;
     }
 
-    // Determine the relevant kind(s) of tracks for this component.
+    // Determine the relevant kind(s) of tracks for this component and filter
+    // out empty kinds.
     const kinds = (referenceTrack.kinds || [referenceTrack.kind]).filter(Boolean);
 
     for (let i = 0; i < tracks.length; i++) {
--- a/src/js/control-bar/text-track-controls/text-track-menu-item.js
+++ b/src/js/control-bar/text-track-controls/text-track-menu-item.js
@@ -93,13 +93,8 @@ class TextTrackMenuItem extends MenuItem {
    * @listens click
    */
   handleClick(event) {
-    const kind = this.track.kind;
-    let kinds = this.track.kinds;
+    const referenceTrack = this.track;
     const tracks = this.player_.textTracks();
-
-    if (!kinds) {
-      kinds = [kind];
-    }
 
     super.handleClick(event);
 
@@ -107,13 +102,27 @@ class TextTrackMenuItem extends MenuItem {
       return;
     }
 
+    // Determine the relevant kind(s) of tracks for this component.
+    const kinds = (referenceTrack.kinds || [referenceTrack.kind]).filter(Boolean);
+
     for (let i = 0; i < tracks.length; i++) {
       const track = tracks[i];
 
-      if (track === this.track && (kinds.indexOf(track.kind) > -1)) {
+      // If the track from the text tracks list is not of the right kind,
+      // skip it. We do not want to affect tracks of incompatible kind(s).
+      if (kinds.indexOf(track.kind) === -1) {
+        continue;
+      }
+
+      // If this text track is the component's track and it is not showing,
+      // set it to showing.
+      if (track === referenceTrack) {
         if (track.mode !== 'showing') {
           track.mode = 'showing';
         }
+
+      // If this text track is not the component's track and it is not
+      // disabled, set it to disabled.
       } else if (track.mode !== 'disabled') {
         track.mode = 'disabled';
       }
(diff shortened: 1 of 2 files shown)


src/js/control-bar/text-track-controls/text-track-menu-item.js
```diff
@@ -93,27 +93,36 @@ class TextTrackMenuItem extends MenuItem {
...
-    }
-
     super.handleClick(event);
 
     if (!tracks) {
       return;
     }
 
+    // Determine the relevant kind(s) of tracks for this component.
+    const kinds = (referenceTrack.kinds || [referenceTrack.kind]).filter(Boolean);
```
BrandonOCasey: Could be good to note that we also filter out any invalid kinds in a comment.

src/js/control-bar/text-track-controls/text-track-menu-item.js
```diff
@@ -93,27 +93,36 @@ class TextTrackMenuItem extends MenuItem {
...
+      // If the track from the text tracks list is not of the right kind,
+      // skip it. We do not want to affect tracks of incompatible kind(s).
+      if (kinds.indexOf(track.kind) === -1) {
+        continue;
+      }
+
+      // If this text track is the component's track and it is not showing,
+      // set it to showing.
+      if (track === referenceTrack) {
         if (track.mode !== 'showing') {
```
BrandonOCasey: This seems like it belongs in the setter for `mode` in `TextTrack` if its not already there. Not that we have to change it in this pr.

src/js/control-bar/text-track-controls/text-track-menu-item.js
BrandonOCasey: After looking at the tests I was going to suggest that we add a tests for a track with multiple kinds. Then I looked at the HTML5 video spec and our track implementation and it seems like `kinds` isn't actually a thing. Only `kind` should ever exist. Perhaps we can keep it for now but add some code about deprecating this usage?

BrandonOCasey:
also I chuckled at the TDD commit 👍

src/js/control-bar/text-track-controls/text-track-menu-item.js
gkatsev: `kinds` is a thing that was added for the combined subtitles/captions menu, it should be considered an implementation detail and not used outside of Video.js internals.

src/js/control-bar/text-track-controls/text-track-menu-item.js
gkatsev: it's theoretically possible (and maybe eventually we will) to have multiple tracks enabled at the same time, while we disallow it via the UI, we shouldn't limit it if people want to enable multiple items directly via the API.

test/unit/control-bar/text-track-controls/text-track-menu-item.test.js
```diff
@@ -0,0 +1,65 @@
+/* eslint-env qunit */
+import TextTrackMenuItem from '../../../../src/js/control-bar/text-track-controls/text-track-menu-item.js';
+import TestHelpers from '../../test-helpers.js';
+import sinon from 'sinon';
+
+QUnit.module('TextTrackMenuItem', {
+  beforeEach(assert) {
+    this.clock = sinon.useFakeTimers();
```
gkatsev: is clock used anywhere?

test/unit/control-bar/text-track-controls/text-track-menu-item.test.js
```diff
@@ -0,0 +1,65 @@
...
+    this.clock.restore();
+  }
+});
+
+QUnit.test('clicking should enable the selected track', function(assert) {
+  assert.expect(2);
+
+  const foo = this.player.addTextTrack('captions', 'foo', 'en');
+
+  const fooItem = new TextTrackMenuItem(this.player, {
```
gkatsev: this isn't cleaned up/disposed.

gkatsev: `fooItem` and `barItem` below aren't cleaned up either.

misteroneill:
It is a separate commit, but you can trust me that it failed. 😉

src/js/control-bar/text-track-controls/text-track-menu-item.js
misteroneill: Yeah, it might have been good to name it in a way that was more obviously an implementation detail because that confused me at first, too. Oh well.

test/unit/control-bar/text-track-controls/text-track-menu-item.test.js
gkatsev: this object is never added to a menu/the player, so it won't get disposed automatically. It's probably not a huge deal but best to clean up as much as we can to reduce any potential memory leak issues with tests.

gkatsev:
I can confirm that tests pass :)

gkatsev:
Also, tested it locally and works great.

test/unit/control-bar/text-track-controls/text-track-menu-item.test.js
misteroneill: Right, right.

PR #5748, opened 2019-01-13
Remove the event listeners with the menu item

dustin71728:
## Description

    The commit fixes followings error:
    1.  The comment says that `Menu`.`addItem` can accept
    the component name as an argument, but it actually
    raises an error because a string object doesn't have
    the method called `on`.
    
    2. Old menu instance didn't remove the event listeners
    with the menu item, causing the old menu still be
    referenced through the event listeners and can't
    be discarded by the garbage collector.

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/menu/menu.js
+++ b/src/js/menu/menu.js
@@ -35,19 +35,25 @@ class Menu extends Component {
     this.focusedChild_ = -1;
 
     this.on('keydown', this.handleKeyPress);
+
+    // All the menu item instances share the same blur handler provided by the menu container.
+    this.handleBlurBindItself_ = Fn.bind(this, this.handleBlur);
+    this.itemOnMethodArgumentStore_ = new Map();
   }
 
   /**
-   * Add a {@link MenuItem} to the menu.
+   * Add event listeners to the {@link MenuItem}.
    *
-   * @param {Object|string} component
-   *        The name or instance of the `MenuItem` to add.
+   * @param {Object} component
+   *        The instance of the `MenuItem` to add listeners to.
    *
    */
-  addItem(component) {
-    this.addChild(component);
-    component.on('blur', Fn.bind(this, this.handleBlur));
-    component.on(['tap', 'click'], Fn.bind(this, function(event) {
+  addEventListenerForItem(component) {
+    if (!(component instanceof Component)) {
+      return;
+    }
+
+    const handleTapClick = Fn.bind(this, function(event) {
       // Unpress the associated MenuButton, and move focus back to it
       if (this.menuButton_) {
         this.menuButton_.unpressButton();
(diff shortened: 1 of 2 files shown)


gkatsev:
Going to take a look today.

src/js/menu/menu.js
```diff
@@ -122,6 +169,43 @@ class Menu extends Component {
...
+    if (this.menuButton_) {
+      this.menuButton_.unpressButton();
+
+      const childComponents = this.children();
+
+      if (!Array.isArray(childComponents)) {
+        return;
+      }
+
+      const foundComponent = childComponents.find(component => {
```
gkatsev: This is because we no longer have direct access to the component, we need to find the component object that matches the event?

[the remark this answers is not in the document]
gkatsev: Maybe something like 
```js
const foundComponent = childComponent.filter((component) => compononent.el() === event.target())[0];
```

dustin71728: Yes, we lost the component reference after moving "handleTapClick" out of closure because the function will align with "handleBlur" function, and using the same way to free both of these listeners after removing of the menu item from the menu instance.

dustin71728: Also I found the commit that add the check whether the clicked menu item is "CaptionSettingsMenuItem" or not to decide to call menu button's focus method

cb890a965cb0dc0df11f55eabea091b2cbb49d71

gkatsev: Thanks, just wanted to confirm I was reading it right :)

dustin71728:
Why the previous discussion lost? anyway, I found the commit ID related to the menu item being clicked that we need to check whether it is "CaptionSettingsMenuItem" or not.

cb890a965cb0dc0df11f55eabea091b2cbb49d71

dustin71728:
Do you think I should add the unit test to confirm the menu button will not get focused when "CaptionSettingsMenuItem" is clicked?

src/js/menu/menu.js
```diff
@@ -122,6 +169,43 @@ class Menu extends Component {
...
+      this.menuButton_.unpressButton();
+
+      const childComponents = this.children();
+
+      if (!Array.isArray(childComponents)) {
+        return;
+      }
+
+      const foundComponent = childComponents.filter(component => {
+        if (component.el() === event.target) {
```
gkatsev: since filter expects us to return a boolean, we don't need the if statement here, we can just return the value of this comparison.

gkatsev:
If you want! Should be fine otherwise, the bulk of the change is tested.

src/js/menu/menu.js
dustin71728: Oh...I don't think about it. thanks.

gkatsev:
Thanks!

PR #5749, opened 2019-01-13
use https links in documentation

thijstriemstra:
Some links were still using http (even though they redirect to https).


gkatsev:
Thanks.

PR #5754, opened 2019-01-15
[A11y Fix] - Ensure resizeManager is not focusable via tab

evanfarina:
The ResizeManager's iframe element is able to be focused via tabbing, which results in a bad user experience for users that rely on a screen reader to navigate the video and its sibling elements. The fix is to set the tabIndex to "-1", and the aria-hidden property to `true` for good measure. Confirmed the fix by spinning up a local server and attempting to tab to the resizeManager's iframe element. This was possible prior to the fix, and not possible after the fix. In addition, I added to the existing resizeManager unit test.

## Description

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/resize-manager.js
+++ b/src/js/resize-manager.js
@@ -80,7 +80,10 @@ class ResizeManager extends Component {
 
   createEl() {
     return super.createEl('iframe', {
-      className: 'vjs-resize-manager'
+      className: 'vjs-resize-manager',
+      tabIndex: -1
+    }, {
+      'aria-hidden': 'true'
     });
   }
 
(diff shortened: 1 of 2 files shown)


src/js/resize-manager.js
```diff
@@ -80,7 +80,9 @@ class ResizeManager extends Component {
 
   createEl() {
     return super.createEl('iframe', {
-      className: 'vjs-resize-manager'
+      'className': 'vjs-resize-manager',
+      'tabIndex': -1,
+      'aria-hidden': true
```
OwenEdwards: ARIA attributes should go in the third parameter to `createEl`, and take a string value not a boolean. E.g. (https://github.com/videojs/video.js/blob/70a71ae81e5026faf38fc0d675521d66847d8411/src/js/control-bar/progress-control/play-progress-bar.js#L24):

```js
      className: 'vjs-play-progress vjs-slider-bar'
    }, {
      'aria-hidden': 'true'
    });
```

gkatsev:
Thanks @evanfarina!

PR #5755, opened 2019-01-15
chore: fix xvfb in travis config, patch safari karma launchers

BrandonOCasey:
It seems that we are having some trouble using xvfb on travis. I could not find any issues/blog posts related to this, but I noticed that the docs have changed for its usage. This fix has already been used in the generator and appears to have fixed the issue.


package.json
```diff
@@ -67,6 +67,7 @@
     "docs:fix": "remark --output -- './{,!(node_modules)/**/}!(CHANGELOG)*.md'",
     "docs:lang": "node build/translations.js",
     "netlify": "babel-node ./build/netlify-docs.js",
+    "prepare": "patch-package",
```
gkatsev: prepare runs before packing the package when when `npm install` is run without anything https://docs.npmjs.com/misc/scripts

gkatsev:
mojave is a lot stricter about where you can open files from without user interaction and travis got updated to a mojave version with this restriction. We watch to patch the karma-safari-launcher and karma-safaritechpreview-launcher to look for an approved place on mojave when it exists.

Issue #5745, opened 2019-01-12
7.4.2: vjs-fullscreen class not added (regression)


blacktrash:
## Description
Player element in fullscreen is missing the `vjs-fullscreen` class.
Test case: https://blacktrash.org/test/videojs/issue5745.html

## Steps to reproduce
1. Go fullscreen
2. Check player element for `vjs-fullscreen` class - the test case demo logs whether it is present.

### Expected
Player element should have `vjs-fullscreen` class in fullscreen mode.

### Actual
Player element does not have `vjs-fullscreen` class in fullscreen mode.

### Error output
If there are any errors at all, please include them here.

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
7.4.2 - perhaps a regression caused by https://github.com/videojs/video.js/commit/2bc90a1899aec36ef57b7c2f207e7f7b15042f85

gkatsev:
What browser and OS are you using? We tested pretty thoroughly, also, I'm unable to reproduce using your example in Chrome 70 Win 10.

blacktrash:
On Mac OS:
- Firefox 64.0.2
- Safari 12.0.2
- Chrome 71.0.3578.98
- Chrome Canary Version 73.0.3669.0

blacktrash:
@gkatsev - are you sure? This is what I get with Win 10 Chrome 70 in browserstack:
<img width="893" alt="screen shot 2019-01-12 at 5 59 27 pm" src="https://user-images.githubusercontent.com/168338/51076059-0c5e7a80-1694-11e9-8745-a0a741375f9e.png">

gkatsev:
Hm... I think I can reproduce on my mac but on my win machine it seemed to have worked fine. We'll take a look.

gkatsev:
I'll see if I can fix the fix or whether we should revert it in favor of a more complete refactor/rewrite of the fullscreen functionality.

blacktrash:
@gkatsev - thanks!

gkatsev:
Thanks for noticing and helping to test! It's out now in 7.4.3.

PR #5757, opened 2019-01-17
Fullscreen api changes to support custome fullscreen element

KevinBrogan:
### Modify isFullscreen to check the available fullscreen api if it exists.
I believe that the fullscreen api is available in every browser that is supported in vjs 7 since ie10 support has been dropped. This change will use the api if it is available, but the code should maybe be changed to only use the api. I would like feedback from the maintainers about this.

### Update jsdoc to indicate that undefined can be returned by setter.
A minor change. jsdoc pedantism.

### Modify requestFullscreen api call to use an element defined in options if it exists.
Setting options.fullscreenElement will cause the fullscreen routines to use that element as the fullscreen target isntead of the el_ target.

### Update fullscreen handling events to target document instead of bubbled player events.
This was required to get multiple players in a single page to syncronise their fullscreen states. Unfortunately, it has broken the fullscreen control bar unit test. I don't have the neccessary familiarity to rewrite the test in a satisfactory way and would like help from the videojs maintainers. I believe the unit test is incorrect, as true fullscreening of an element has been locked to user gesture events by the browser developers, so this feature can't actually be tested in an automated environment unless Karma can emit isTrusted user gesture events. All browsers that I tested in a live environment work as expected.

### Note
documentation still needs to be updated. I would like to wait till all unit tests pass before completing the documentation. As well, I have noticed that this code change fixes issue videojs/video.js#5745 as a sideeffect, although the change was not intentional.

- [Y ] Feature implemented / Bug fixed
  - [Y ] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/fullscreen-toggle.js
+++ b/src/js/control-bar/fullscreen-toggle.js
@@ -24,7 +24,7 @@ class FullscreenToggle extends Button {
    */
   constructor(player, options) {
     super(player, options);
-    this.on(player, 'fullscreenchange', this.handleFullscreenChange);
+    this.on(document, 'fullscreenchange', this.handleFullscreenChange);
 
     if (document[FullscreenApi.fullscreenEnabled] === false) {
       this.disable();
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -521,11 +521,7 @@ class Player extends Component {
     this.one('play', this.listenForUserActivity_);
 
     if (FullscreenApi.fullscreenchange) {
-      this.on(FullscreenApi.fullscreenchange, this.handleFullscreenChange_);
-
-      if (IE_VERSION || browser.IS_FIREFOX && prefixedFS) {
-        this.on(document, FullscreenApi.fullscreenchange, this.handleFullscreenChange_);
-      }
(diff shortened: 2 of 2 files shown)


gkatsev:
oops

gkatsev:
@KevinBrogan hey would you be able to rebase this PR against master?

KevinBrogan:
I gave it a shot but some internals have been changed and while I was able to fullscreen I wasn't able to unfullscreen. I'm on a bit of a crunch and ran out of time to investigate.

In the meantime, I've since resolved my issue by defining my own FullscreenToggle component and then setting the userActions.doubleClick option to function(){ this.controlBar.fullscreenToggle.handleClick(); }

If you're interested in the fullscreen toggle changes, give me a message on slack or on here and I can give you a copy paste.

gkatsev:
Thanks @KevinBrogan, sorry that we never got this in. We'll take a look into getting it in anyway when we get the chance.

vsx-sieber:
Any updates on this? In your guide you still mention that support for providing a custom `element` property in `fullscreen.options` is planned, but its been a while since it was first considered.

PR #5701, opened 2018-12-19
Using src attribute instead of source element

BVFernandes:
## Description
Possible fix to issue [#4851](https://github.com/videojs/video.js/issues/4851) by using src attribute to generate the source element when the last is undefined.

## Specific Changes proposed
In the _loader.js_, I tried modify the player data structure to add the _source_ object (with _src_ and _type_) using simply the _src_ attribute, when the _video-js_ class is created in the _index.html_ file. But, sadly, it doesn't work, because the playerOptions object is edited and not the player object.

## Requirements Checklist
- [ ] Feature implemented / Bug fixed
- [X] If necessary, more likely in a feature request than a bug fix
  - [X] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
- [ ] Reviewed by Two Core Contributors

PR #5762, opened 2019-01-22
fix: allow src attribute to work

gkatsev:
If we are loading a new tech and sources array is empty and we have a src option from the src attribute, set that to be a sources array of just the string.

This helps a lot with videojs/video.js#4851 and may be the last fix.

Fixes videojs/video.js#5701

### DIFF
--- a/src/js/tech/loader.js
+++ b/src/js/tech/loader.js
@@ -32,6 +32,15 @@ class MediaLoader extends Component {
 
     super(player, options_, ready);
 
+    // if no sources are provides but we have a src attribute, use it. In the source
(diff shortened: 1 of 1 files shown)


gkatsev:
Going to leave this open for now, if someone has time to look at why it's failing, that'll be great.

misteroneill:
Yeah, it appears to be breaking sourceset tests.

rwlaschin:
Has this been fixed?  Looks like no... can this be reopened?

gkatsev:
This is still open. Unfortunately, I haven't had the time to come back to it yet.

PR #5788, opened 2019-02-08
Set cache before 'mouseenter' event for inactivityTimeout

iosamuel:
## Description
We need to define the cache for `inactivityTImeouot` before the `mouseenter` event, because if you do play on the `controlBar` for example and go out, the actual `mouseenter` event of it never gets triggerd, and so the cache is gonna be always undefined. 

It's a really visible bug when you have a custom player with the control-bar all the way bottom (-40px) and you set a `inactivityTimeout: 0`, it gets overriden by the undefined and gets hidden always.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2066,6 +2066,7 @@ class Player extends Component {
       // we set it to zero here to ensure that if we do start actually caching
       // it, we reset it along with everything else.
       currentTime: 0,
+      inactivityTimeout: this.options_.inactivityTimeout,
       duration: NaN,
       lastVolume: 1,
       lastPlaybackRate: this.defaultPlaybackRate(),
@@ -3647,7 +3648,6 @@ class Player extends Component {
     const controlBar = this.getChild('controlBar');
 
     if (controlBar) {
-      this.player().cache_.inactivityTimeout = this.player().options_.inactivityTimeout;
 
       controlBar.on('mouseenter', function(event) {
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3647,6 +3647,7 @@ class Player extends Component {
     const controlBar = this.getChild('controlBar');
 
     if (controlBar) {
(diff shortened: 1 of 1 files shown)


misteroneill:
> if you do play on the `controlBar` for example and go out, the actual `mouseenter` event of it never gets triggerd

This definitely sounds like a bug, but perhaps the bug is that the `mouseenter` event is being obscured by something?

gkatsev:
Can you elaborate on what you mean by 
> because if you do play on the `controlBar` for example and go out

Though, this change seems fine to me regardless, it makes sense to prime it with a default value. Though, it should potentially be done in the [resetCache_ method.](https://github.com/videojs/video.js/blob/master/src/js/player.js#L2053-L2061)

iosamuel:
@misteroneill No, the `mouseenter` is not obscured, it's just that the whole part of user activity is being initialized in the `play` event: https://github.com/videojs/video.js/blob/46d2d494feaf204931721d5eb4dda17e67ac369c/src/js/player.js#L522

iosamuel:
@gkatsev As I said to @misteroneill, since the userActivity events such as `mouseenter` are being listened only when you hit play, if you are over the `controlBar` to hit play (`playButton`), you actually are never gonna fire the `mouseenter` event which sets the cache for `inactivityTimeout` until you go out of the player and come back again with your cursor.

I will definitely check on the resetCache_ method to make it better. Thanks!

gkatsev:
Oh, I see. Thanks.

gkatsev:
This probably wasn't noticed because we don't have the control bar showing by default.

iosamuel:
Yes, exactly, as I made some custom video-players, one of them have a skin that always shows the control bar (or at least it tries haha).

iosamuel:
@gkatsev The `resetCache_` worked like a charm! I updated the PR.

Issue #5624, opened 2018-11-28
In chrome 70 the small play/pause control doesn't work with Polymer 1.x Gestures tap is used on document


mscalora:
## Description
A partner noticed a bug where the the video-js small play/pause control doesn't work when our Polymer 1.x plugin is used on the same page in recent version of chrome (e.g. 70). The same issue appears to affect the latest Polymer 2.x also though we have not migrated to 2.x yet. I narrowed it down to the Polymer Gestures "tap" mechanism is used on the document. This very simple codepen shows the broken behaviour:

Polymer 1.x: https://codepen.io/mscalora/full/ZmqqpR/
Polymer 2.x: https://codepen.io/mscalora/pen/mQzQmp

The codepen contains only a pared down version of the video-js example, the polymer library and the single line:

    Polymer.Gestures.add(document.documentElement, 'tap', null);

My workaround can be exercised by adding 'fix' to the URL like https://codepen.io/mscalora/full/ZmqqpR/?fix  

I'm not sure why the Polymer Gestures code an video-js are incompatible but I have a guess that video-js doesn't call `event.stopPropagation()` on the play/pause clicks and this would prevent the issue. It is not clear to me that this should be necessary or would be problematic.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Load codepen: https://codepen.io/mscalora/full/ZmqqpR/
2. Start video playing (so controls appear on bottom of video element)
3. Click small play/pause button on bottom left

### Expected
small play/pause button should work (play or pause)

### Actual
small play/pause button does not appear to work, I think it actually toggles twice quickly from viewing in the debugger but appears to have no effect.

### Error output
n/a

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
all tested including 7.3
#### browsers
Chrome 70
#### OSes
Mac
### plugins
none

gkatsev:
Reopening the issue because the fix was reverted in #6128. We'll have to think some more about if this can be fixed without introduced other regressions, see #6092.

jheidel:
I ran into this today on a Polymer 3 project. It's a standard Polymer environment based off the [polymer-starter-kit](https://github.com/Polymer/polymer-starter-kit). In case it helps any future readers, I used the following workaround, based off of the workaround mentioned in the original codepen.

```
// Polymer webcomponent lifecycle `ready` method
ready() {
  super.ready();
  this.addEventListener('click', function (e) { 
[...]

PR #5803, opened 2019-02-17
Fix polymer issue by marking click event with stopPropagation

mscalora:
## Description
The play button stops working when recent versions of Google's Polymer
is in use on the page because of the way Polymer synthesizes 'tap'
events on non-touch devices. The Polymer tap code thinks the click
event was ignored unless the DOM event's stopPropagation method is
called. 

Demo of issue: https://codepen.io/mscalora/pen/mQzQmp

Fixes: videojs/video.js#5624 In chrome 70 the small play/pause control doesn't work with Polymer 1.x Gestures tap is used on document

## Specific Changes proposed
The play-pause button's click handler now calls event.stopPropagation() after the click event is handled.

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -61,6 +61,7 @@ class PlayToggle extends Button {
     } else {
       this.player_.pause();
     }
+    event.stopPropagation();
   }
 
   /**


gkatsev:
Looks like the mute toggle also doesn't work.

gkatsev:
However, just tried it and it fixes it the issue with the play toggle.

boris-petrov:
I think this broke the `on("click" ...` event. Check out this reproduction:

https://codepen.io/anon/pen/xoBwPG?editors=1111

Press the big play button and you see a log in the console. But then pressing pause/play (the small button on the bottom left) doesn't log it any more. Changing the version to `7.5.6` fixes the issue and every click produces a log in the console.

Please reopen [the issue](https://github.com/videojs/video.js/issues/5624) or should I open a new one?

gkatsev:
Can you please open a new issue?

PR #5806, opened 2019-02-18
fix: changes max height of menus based on responsive classes.

gjanblaszczyk:
## Description
A proposed fix for #5524 

## Requirements Checklist
  - [ ] Docs/guides updated


src/css/components/menu/_menu-popup.scss
```diff
@@ -19,6 +19,20 @@
   max-height: 15em;
 }
 
+.vjs-layout-small .vjs-menu-button-popup .vjs-menu .vjs-menu-content {
+  max-height: 10em;
+}
+
+.vjs-layout-medium .vjs-menu-button-popup .vjs-menu .vjs-menu-content {
+  max-height: 14em;
```
gkatsev: why not keep this at the default 15em?

src/css/components/menu/_menu-popup.scss
```diff
@@ -19,6 +19,20 @@
   max-height: 15em;
 }
 
+.vjs-layout-small .vjs-menu-button-popup .vjs-menu .vjs-menu-content {
```
gkatsev: should we add `x-small` and `tiny` here? Or maybe just have those together be something like 5em?

gjanblaszczyk: I didn't add it because the player hides captions for x-small and tiny layouts but maybe we should have some default value for all layouts. What do you think?

src/css/components/menu/_menu-popup.scss
gkatsev: Ah, ok, smallest size for medium is smaller than 15em

src/css/components/menu/_menu-popup.scss
gkatsev: Probably worth having it in case someone forces the menu buttons on.

gjanblaszczyk: Ok, I added missing `tiny` and `x-small` classes.

gkatsev: Thanks!

gjanblaszczyk: np

gkatsev:
Thanks @gjanblaszczyk!

PR #5815, opened 2019-02-22
fix vjs-volume-panel visual bug

Kogoruhn:
before / after:
changes

### DIFF
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -19,7 +19,6 @@
 
 .video-js .vjs-volume-control {
   cursor: pointer;
-  margin-right: 1em;
   @include display-flex;
 }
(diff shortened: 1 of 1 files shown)


src/css/components/_volume.scss
```diff
@@ -19,7 +19,6 @@
 
 .video-js .vjs-volume-control {
   cursor: pointer;
-  margin-right: 1em;
```
gkatsev: The volume control component can be added to the control bar separately and this would affects that.
Probably needs to reset the margin-right just in `.video-js .vjs-volume-panel` below (line 38/37)

gkatsev:
@Kogoruhn hey, thanks for the PR, we'd love to get this in but it requires some tweaks. Would you be able to work on it? Thanks!

PR #5820, opened 2019-02-25
[i18n]Update Occitan language

Quenty31:
Update for the Occitan language.
I have to admit I'm not really sure I've got the idea of the 2 lines about the live thing.

### DIFF
--- a/lang/oc.json
+++ b/lang/oc.json
@@ -9,6 +9,8 @@
   "Remaining Time": "Temps restant",
   "Stream Type": "Tipe de difusion",
   "LIVE": "DIRÈCTE",
+  "Seek to live, currently behind live": "Cercar lo dirècte, actualament darrièr lo dirècte",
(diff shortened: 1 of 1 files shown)


gkatsev:
Here's some background and info on the two live items.
We recently added a [new live ui](https://docs.videojs.com/tutorial-live.html#the-new-user-interface). This added a new button that when you click on it, it will take you to the live point. For example, if you were paused for a bit, and want to just skip to what is currently streaming, you'd click on it.
In addition, this button notifies the user whether they are at the live point or are behind it, for example if they were paused. Because this is communicated to the user visually, we need to provide this information to the screen reader.

gkatsev:
Also, thanks for keeping the occitan translations up-to-date!

Quenty31:
New PR opend!
https://github.com/videojs/video.js/pull/5829

PR #5822, opened 2019-02-26
fix: if play is delayed till loadstart, call load

gkatsev:
In Safari, if we call play early and we are going to be waiting till
loadstart, it's possible this will take too long for the user activition
flag to still be available. In this case, we should call load() on the
player to prime the video element so for when loadstart happens.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2234,7 +2234,7 @@ class Player extends Component {
 
       // if we are in Safari, there is a high chance that loadstart will trigger after the gesture timeperiod
       // in that case, we need to prime the video element by calling load so it'll be ready in time
-      if (browser.IS_ANY_SAFARI) {
+      if (browser.IS_ANY_SAFARI || browser.IS_IOS) {
         this.load();
       }
       this.one('loadstart', this.playOnLoadstart_);
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2232,6 +2232,11 @@ class Player extends Component {
         callback(this.play());
       };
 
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -2232,6 +2232,11 @@ class Player extends Component {
         callback(this.play());
       };
 
+      // if we are in Safari, there is a high chance that loadstart will trigger after the gesture timeperiod
+      // in that case, we need to prime the video element by calling load so it'll be ready in time
+      if (browser.IS_ANY_SAFARI) {
```
gkatsev: this is only done in safari to minimize any potential impact from calling load, can be loosed in the future, should we need it.

forbesjo: Is this a concern for iOS Web Views?

gkatsev: If it happens on desktop safari, I can't imagine why it can't happen on iOS as well. We'd need to test it, but theoretically, the extra call to load shouldn't do anything bad.

forbesjo: I mean that should this block also apply to web views, specifically that `IS_ANY_SAFARI` excludes mobile Chrome

gkatsev: That's a good question. Let me see how iOS Chrome handles autoplay restrictions.

gkatsev: Looks like chrome iOS has the same issue. Let me check Android as well.

gkatsev: Not seeing the behavior on Android. So, adding IS_IOS as well.

PR #5829, opened 2019-02-27
[i18n]Update Occitan language

Quenty31:
Translation of new strings + 1 correction.
Thanks @gkatsev for the help with the 2 lines about the live!


gkatsev:
Thanks @Quenty31!

PR #5830, opened 2019-02-27
 Fix Fullscreen Detection when Player is Nested within Document Fullscreen

richardbushell:
As per title, this fixes Fullscreen Detection when Player is Nested within Document


gkatsev:
This is probably fine but I wonder how iOS works with this change. Both iPhones and iPads. iOS 12 on iPads as the fs API now but iPhones don't.

richardbushell:
Identical behaviour on iPhone to current behaviour. As you state, iPhone still does not support fsAPI anyway, so this fix makes absolutely no difference anyway.

Unfortunately I have not got an iPad to test on. I predict it will work in Chrome, probably won't work in Safari (as it won't with the current code either). Even desktop Safari doesn't yet support NESTED fullscreen elements correctly as per the bugs filed. But non-nested video fullscreen will continue to work as expected. If anyone has an iPad with latest iOS release then please confirm that neither the current VJS release nor the fix will work in nested fullscreen mode, using the TWO LINKS BELOW using the instructions given underneath the videos):-

Current VJS release code (7.5.0):-
https://living.video/vjsreducednested.html

Fixed code (tested as fixed for all other browsers):-
https://living.video/vjsreducednestedfix.html

Can anyone with an iPad confirm that neither works in Safari, and whether they work with Chrome please?

richardbushell:
gkatsev - I managed to borrow an iPad running the latest iOS (12.1.4)

Currently they have only enabled fullscreen document elements in Safari, not yet in Chrome with this latest iOS release on iPad.

Tested all combinations. All VJS Video Fullscreen enter/exit combinations continue to work as they do with 7.5.0, so no problem.  iOS (WebKit) still does not support NESTED fullscreen elements at all (although the Fullscreen API spec allows this), so there's no way to make a video fullscreen within an existing document fullscreen. I knew this anyway, and there are already requests with Safari browser to implement this for future as all other browsers already do.

So, I can confirm that the PR fixes NESTED video on: PC Chrome, Edge, IE, Firefox, and Opera.
It also fixes NESTED video on Android.
iOS is unaffected, it still doesn't support nested video (as it didn't previously either).

I'm sure the PR is good to go.

richardbushell:
@gkatsev - is this now OK to merge? I've been testing this in production for the last week across all available device types and browsers and it has fixed the bug, with no problems. Please advise.

gkatsev:
Likely fine, was out last week but should be able to get to it this week.

gkatsev:
Thanks!

PR #5831, opened 2019-02-27
fix warning - element supplied is not included in the DOM

KevinBrogan:
Videojs currently relies on the document object from the window that videojs script is instantiated in, in order to determine if the element it will use for initialization is attached to the dom. The method used to determine dom attachment is incorrect. This fixes that method.

## Description

## Specific Changes proposed
document.contains(el) will only check if el is contained within that one document. This causes problems for elements in iframes.

Instead, use the element's ownerdocument instead of the global document. Additionally, check that the document in question has a default view. If the document is no longer attached to the dom, the defaultView of the document will be null.

## Requirements Checklist
- [X ] Feature implemented / Bug fixed
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -142,6 +142,12 @@ function videojs(id, options, ready) {
     throw new TypeError('The element or ID supplied is not valid. (videojs)');
   }
 
+  // document.contains(el) will only check if el is contained within that one document.
+  // This causes problems for elements in iframes.
+  // Instead, use the element's ownerDocument instead of the global document.
+  // This will make sure that the element is indeed in the dom of that document.
+  // Additionally, check that the document in question has a default view.
+  // If the document is no longer attached to the dom, the defaultView of the document will be null.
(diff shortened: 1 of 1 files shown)


thijstriemstra:
I would include all, or a summary of that info in a code comment

> document.contains(el) will only check if el is contained within that one document. This causes problems for elements in iframes.

> Instead, use the element's ownerdocument instead of the global document. Additionally, check that the document in question has a default view. If the document is no longer attached to the dom, the defaultView of the document will be null.

KevinBrogan:
Not a bad suggestion

thijstriemstra:
> Not a bad suggestion

And ideally, you'd also include a test for this (showing how it behaves in an iframe), but I guess it's better to wait for some confirmation from devs first.

gkatsev:
Thanks!

gkatsev:
FYI, https://github.com/videojs/video.js/pull/5872

PR #5833, opened 2019-02-28
+ jsDelivr CDN to README

tomByrer:
[jsDelivr CDN](https://www.jsdelivr.com/network) is like a mashup of CloudFlare, Fastly, QUANTIL for China & 22+ POPs with backups at every level.  Files are sourced from both [npm & GitHub](https://www.jsdelivr.com/features) for projects not pushed to npm & alphas.  Files can also be [combined into one request](https://www.jsdelivr.com/features#combine).

The new instructions are longer; they could be moved to a webpage, drop cdnjs & unpkg since jsDelivr is technically better (though less popular), or use `<details><summary>` to collapse the instructions, if that is a concern.

Small demo for both CSS & JS: https://codepen.io/tomByrer/pen/vbyNJq

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,25 @@ Alternatively, you can include Video.js by getting it from [npm](https://videojs
 <!-- cdnjs : use a specific version of Video.js (change the version numbers as necessary) -->
 <link href="https://cdnjs.cloudflare.com/ajax/libs/video.js/6.7.3/video-js.min.css" rel="stylesheet">
 <script src="https://cdnjs.cloudflare.com/ajax/libs/video.js/6.7.3/video.min.js"></script>
+
+<!-- jsDelivr : use the latest version of Video.js -->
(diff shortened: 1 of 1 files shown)


gkatsev:
I don't think we need all of the possible jsdelivr permutations. Maybe just the default one with a link to the package page and jsdelivr docs?

tomByrer:
Sure, "default" = `@latest` or specific version?

thijstriemstra:
might as well add unpkg as well (my fav, cdnjs not so much).

tomByrer:
@thijstriemstra Ironically, they both use CloudFlare's CDN.  So does jsDelivr, but adds Fastly's & ~hundred other POPs.

Maybe the verbose directions could go on a document page?

tomByrer:
closing, will attempt later with reduced README & extended docs.

PR #5836, opened 2019-03-01
Fix bug preventing control bar from hiding on mobile

Spartanace1024:
## Description
Bug fix that prevented the control bar from being hidden when progress bar was tapped on on mobile (not seeked on, and only after control bar was already showing).
This was due to a mouseenter event firing (on mobile) that would set the `inactivityTimeout = 0` right before the below bit of code was executed.

```
      const timeout = this.options_.inactivityTimeout;

      if (timeout <= 0) {
        return;
      }

      // In <timeout> milliseconds, if no more activity has occurred the
      // user will be considered inactive
      inactivityTimeout = this.setTimeout(function() {
        // Protect against the case where the inactivityTimeout can trigger just
        // before the next user activity is picked up by the activity check loop
        // causing a flicker
        if (!this.userActivity_) {
          this.userActive(false);
        }
      }, timeout);
```

## Specific Changes proposed
Ensures the mouseenter event never applies to mobile by checking ios & android flags

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3647,15 +3647,13 @@ class Player extends Component {
 
     const controlBar = this.getChild('controlBar');
 
-    if (controlBar) {
+    // Fixes bug on Android & iOS where when tapping progressBar (when control bar is displayed)
+    // controlBar would no longer be hidden by default timeout.
+    if (controlBar && !browser.IS_IOS && !browser.IS_ANDROID) {
 
       controlBar.on('mouseenter', function(event) {
-        // Fixes bug on Android & iOS where when tapping progressBar (when control bar is displayed)
-        // controlBar would no longer be hidden by default timeout.
-        if (!browser.IS_IOS && !browser.IS_ANDROID) {
-          this.player().cache_.inactivityTimeout = this.player().options_.inactivityTimeout;
-          this.player().options_.inactivityTimeout = 0;
-        }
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -3650,10 +3650,12 @@ class Player extends Component {
     if (controlBar) {
 
       controlBar.on('mouseenter', function(event) {
-
-        this.player().cache_.inactivityTimeout = this.player().options_.inactivityTimeout;
-        this.player().options_.inactivityTimeout = 0;
-
+        // Fixes bug on Android & iOS where when tapping progressBar (when control bar is displayed)
+        // controlBar would no longer be hidden by default timeout.
+        if (!browser.IS_IOS && !browser.IS_ANDROID) {
```
gkatsev: what about moving this check into the the if on line 3650? This way, we don't even add these listeners in the first place

Spartanace1024: Yeah. I think that's a better way to go about it actually. Not sure when i'll have time to make this change though. 

I'll try to get it done by next week :)

gkatsev:
@Spartanace1024 hey, did you have a chance to look at it? Would you be able to get to it early next week? Thanks!

Spartanace1024:
Hey @gkatsev I did indeed and added the changes as you have suggested.

PR #5850, opened 2019-03-10
added a baseURL option so media access can be relative

greenpdx:
## Description
This is an enhancement. I added a new option that allows media access for a relative URL.  this means by getting a directory listing, all the media can be played.  I did this so I could automate machine learning using OpenCV.  

## Specific Changes proposed
Added a new option baseUrl and a new method baseUrl([URL]).  
The argument URL is optional for setting the new URL. it always returns the URL. If baseUrl is not set it has no effect 

## Requirements Checklist
- [x ] If necessary, more likely in a feature request than a bug fix
  - [x ] Change has been verified in an actual browser (Chome, Firefox, IE)
     both chrome and Firefox
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "bluebird": "^3.5.3",
     "browserify": "^16.2.3",
     "browserify-istanbul": "^3.0.1",
-    "chokidar-cli": "^1.2.1",
+    "chokidar-cli": "^1.2.2",
     "clean-css-cli": "^4.2.1",
     "cli-table": "^0.3.1",
     "conventional-changelog-cli": "^2.0.11",
@@ -115,7 +115,7 @@
     "humanize-duration": "^3.16.0",
     "husky": "^1.2.0",
     "jsdoc": "^3.5.5",
-    "karma": "^4.0.0",
+    "karma": "^4.0.1",
     "karma-browserify": "^5.3.0",
     "klaw-sync": "^6.0.0",
     "lint-staged": "^8.1.0",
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
It seems like it would make more sense to do this outside of the player before passing your source(s) into `player.src`. It could also be possible that I am not understanding you use case well enough. Do you have more information on why we would want a `baseUrl` option?

gkatsev:
I'd like to hear more about the usecase before we bring we consider bring it in.

greenpdx:
I am using playlist to keep track to the videos.  The use case for this is having a single directory on the server for the videos.  This may a directory it self or a REST function that returns the video.  I am using it for machine learning. I have created a "file selection dialog" for the videos.  Once I select the videos I would have to add the baseUrl to each video while I am entering it into the playlist.  This means creating a new list with the base URL every time I change the selection. By having a baseURL added at load time means that I am able to reuse the selection list with out duplication of the select videos. 
Also it can be used for security. By having a baseUrl separate from the list, security token can be added without modifying the list it self.  The feature would be used in web applications more than just playing random videos in the browser. 
In summary it would simplify video list handling and able to add security token. 
I hope this helps

gkatsev:
Thanks for your patience, I've meant to reply earlier but other things kept getting in the way.

Thanks for providing the usecase, it really helped. Unfortunately, I think unless we get a lot more requests for a baseUrl option, I'd prefer to avoid including it. Instead, you can get the same or similar functionality using middleware. Since middleware allows you to modify source urls, you can make a base-url middleware that will prepend the given baseUrl for all sources that don't start with `http`, or you can give them a special type for the middleware to pick up. I created an example codepen: https://codepen.io/gkatsev/pen/rbrRYo (the javascript is using some new javascript features, but it can be re-written to not use them as well). And docs for middleware are available here: https://docs.videojs.com/tutorial-middleware.html.

Hope this ends up being sufficient for your use case. We appreciate you spending the time opening the PR and providing a detailed usecase.

PR #5852, opened 2019-03-11
fix: Remove `playerEvent` and extra `timeupdate` handler in SeekBar

BrandonOCasey:
## Description
SeekBar ends up adding two extra `timeupdate` handlers for the `update` function.
1. One is added in `Slider` via the `playerEvent` on `SeekBar`s prototype. This one may be worse because it will be added on the original update function before we throttle it.
2. The other is just added in `setEventHandlers_`.

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -81,7 +81,7 @@ class SeekBar extends Slider {
       this.clearInterval(this.updateInterval);
     });
 
-    this.on(this.player_, ['timeupdate', 'ended'], this.update);
+    this.on(this.player_, 'timeupdate', this.update);
   }
 
   /**
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -53,6 +53,7 @@ class SeekBar extends Slider {
   setEventHandlers_() {
     this.update = Fn.throttle(Fn.bind(this, this.update), UPDATE_REFRESH_INTERVAL);
 
+    this.on(this.player_, 'timeupdate', this.update);
     this.on(this.player_, 'ended', this.handleEnded);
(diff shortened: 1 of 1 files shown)


misteroneill:
Approved though it may be good to add a test to make sure we don't encounter this again.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -428,12 +427,5 @@ if (!IS_IOS && !IS_ANDROID) {
   SeekBar.prototype.options_.children.splice(1, 0, 'mouseTimeDisplay');
 }
 
-/**
- * Call the update event for this Slider when this event happens on the player.
- *
- * @type {string}
- */
-SeekBar.prototype.playerEvent = 'timeupdate';
```
gkatsev: Thoughts about keeping this in and removing line 84 (`    this.on(this.player_, ['timeupdate', 'ended'], this.update);
)`? update already gets called from ended via the handleEnded and the slider makes sure that the playerEvent handler is added/removed properly.

BrandonOCasey: This causes the `update` function to be bound to `timeupdate` before it is throttled in `setEventHandlers_()`, which I don't think we want.

gkatsev: Ah, I see.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -53,7 +53,6 @@ class SeekBar extends Slider {
   setEventHandlers_() {
     this.update = Fn.throttle(Fn.bind(this, this.update), UPDATE_REFRESH_INTERVAL);
 
-    this.on(this.player_, 'timeupdate', this.update);
```
gkatsev: can we keep this line and get rid of line 84 then? 😄

BrandonOCasey: Done 👍

PR #5853, opened 2019-03-11
fix: trigger dispose on resize manager

BrandonOCasey:
Right now we leak one event handler in DomData on every player dispose. Funnily enough we are leaking a dispose event handler from the evented mixin: https://github.com/videojs/video.js/blob/master/src/js/mixins/evented.js#L395

### DIFF
--- a/src/js/resize-manager.js
+++ b/src/js/resize-manager.js
@@ -109,7 +109,6 @@ class ResizeManager extends Component {
   }
 
   dispose() {
-    this.trigger('dispose');
     if (this.debouncedHandler_) {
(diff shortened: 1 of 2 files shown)


src/js/resize-manager.js
```diff
@@ -109,6 +109,7 @@ class ResizeManager extends Component {
   }
 
   dispose() {
+    this.trigger('dispose');
```
gkatsev: should we call `super.dispose()` here instead?

BrandonOCasey: yeah that probably makes more sense.

PR #5854, opened 2019-03-11
feat: add several debugging aides

BrandonOCasey:
TODO: Add some tests in this pull request to verify:
1. That domData contains nothing after player is disposed
2. That eventhandlers in DomData do not have the same guid for the same event (IE the same function is listening for the same event)

### DIFF
--- a/src/js/mixins/evented.js
+++ b/src/js/mixins/evented.js
@@ -280,6 +280,7 @@ const EventedMixin = {
       // Use the same function ID as the listener so we can remove it later
       // it using the ID of the original listener.
       wrapper.guid = listener.guid;
+      wrapper.original_ = listener.original_ || listener;
       listen(target, 'one', type, wrapper);
     }
   },
--- a/src/js/utils/fn.js
+++ b/src/js/utils/fn.js
@@ -44,6 +44,9 @@ export const bind = function(context, fn, uid) {
   // currently used in text tracks
   bound.guid = (uid) ? uid + '_' + fn.guid : fn.guid;
 
+  bound.original_ = fn.original_ || fn;
+  bound.context_ = fn.context_ || context;
+
   return bound;
 };
 
@@ -71,6 +74,8 @@ export const throttle = function(fn, wait) {
       last = now;
     }
   };
+
+  throttled.original_ = fn.original_ || fn;
 
   return throttled;
 };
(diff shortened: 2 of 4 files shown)


src/js/video.js
```diff
@@ -553,6 +554,8 @@ videojs.computedStyle = computedStyle;
  */
 videojs.dom = Dom;
 
+videojs.domData = DomData;
```
gkatsev: I am against this being in a production build of Video.js, though, I do see the usefulness of it when debugging.

src/js/mixins/evented.js
```diff
@@ -280,6 +280,7 @@ const EventedMixin = {
       // Use the same function ID as the listener so we can remove it later
       // it using the ID of the original listener.
       wrapper.guid = listener.guid;
+      wrapper.original_ = listener.original_ || listener;
```
gkatsev: we'd want to make sure to null these out so that we're not leaking references.

src/js/utils/fn.js
```diff
@@ -44,6 +44,9 @@ export const bind = function(context, fn, uid) {
   // currently used in text tracks
   bound.guid = (uid) ? uid + '_' + fn.guid : fn.guid;
 
+  bound.original_ = fn.original_ || fn;
```
gkatsev: we should make sure to null these out when they're no longer needed.

src/js/video.js
gkatsev: If we can have this be here during local dev but not when releasing, that's fine.

BrandonOCasey: For local development we can just import DomData (as long as we let it export `elData`). I think that plugins could also benefit from being able to see what `elData` they may be leaking, but I am not quite sure how to do that without exposing it here.

BrandonOCasey: Perhaps we can expose it somewhere else?

gkatsev: I'm not sure we can safely expose it and not deal with consequences down the line or have issues with people fiddling with it and causing issues.

BrandonOCasey:
After thinking about this more, maybe we just want to have another build target. Something like `video.debug.js` and include/exclude code like this accordingly. Most of this code is useful but I could see it causing issues in production.

gkatsev:
A debug build is fine, or having a rollup directive via envify or something that removes the exposing of domData when in production

BrandonOCasey:
OK I am going to open an issue and link back to this pull request, perhaps we can revisit this when we are fully building on rollup via #5649 & #5601 I will work on adding a unit test to prevent leaks via DomData in another pull request.

BrandonOCasey:
The issue in question https://github.com/videojs/video.js/issues/5858

PR #5855, opened 2019-03-12
fix: event target memory leak

BrandonOCasey:
Originally I created this as a part of #5853 but this issue is very complicated and I think it needs its own discussion. 

This fixes the issues I am seeing where `DomData.elData` contains two empty objects after dispose. 

The empty objects are created because `getData` creates an empty object in `DomData.elData` for `event.target` in `Events.trigger` (the `else if` that I changed).
 
This happens in two events:
1. [`document`](https://github.com/videojs/video.js/blob/master/src/js/tracks/track-list.js#L116) which I think is wrong
2. [`HtmlTrackElement`](https://github.com/videojs/video.js/blob/master/src/js/tracks/html-track-element.js#L110) which seems ok

While that does fix the actual issue I am seeing, it seems really strange for the `target` of an `EventTarget` to default to `document` and not `this`.

### DIFF
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -138,10 +138,14 @@ EventTarget.prototype.one = function(type, fn) {
 EventTarget.prototype.trigger = function(event) {
   const type = event.type || event;
 
+  // deprecation
+  // In a future version we should default target to `this`
+  // similar to how we default the target to `elem` in
+  // `Events.trigger`. Right now the default `target` will be
+  // `document` due to the `Event.fixEvent` call.
   if (typeof event === 'string') {
     event = {type};
   }
-
   event = Events.fixEvent(event);
 
   if (this.allowedEvents_[type] && this['on' + type]) {
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -142,7 +142,6 @@ EventTarget.prototype.trigger = function(event) {
     event = {type};
   }
 
-  event = Object.assign({target: this}, event);
   event = Events.fixEvent(event);
 
   if (this.allowedEvents_[type] && this['on' + type]) {
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -141,6 +141,8 @@ EventTarget.prototype.trigger = function(event) {
   if (typeof event === 'string') {
     event = {type};
   }
+
+  event = Object.assign({target: this}, event);
   event = Events.fixEvent(event);
 
   if (this.allowedEvents_[type] && this['on' + type]) {
--- a/src/js/tracks/track-list.js
+++ b/src/js/tracks/track-list.js
@@ -72,7 +72,8 @@ class TrackList extends EventTarget {
        */
       this.trigger({
         track,
-        type: 'addtrack'
+        type: 'addtrack',
+        target: this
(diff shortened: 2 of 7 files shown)


src/js/utils/events.js
```diff
@@ -431,7 +431,7 @@ export function trigger(elem, event, hash) {
     trigger.call(null, parent, event, hash);
 
   // If at the top of the DOM, triggers the default action unless disabled.
-  } else if (!parent && !event.defaultPrevented) {
+  } else if (!parent && !event.defaultPrevented && event.target && event.target[event.type]) {
```
gkatsev: this is the main fix?

gkatsev: Looks like some tests are failing because of this. We probably want to expand the code above that makes sure we have an event object (line 413):

```js
  // If an event name was passed as a string, creates an event out of it
  if (typeof event === 'string') {
    event = {type: event, target: elem};
  } else if (!event.target) {
    event.target = elem;
  }
```

src/js/event-target.js
```diff
@@ -141,6 +141,8 @@ EventTarget.prototype.trigger = function(event) {
   if (typeof event === 'string') {
     event = {type};
   }
+
+  event = Object.assign({target: this}, event);
```
gkatsev: I'm a bit worried about what changing the default target could do

BrandonOCasey: yeah agreed, after testing some more I think that we have to remove this.

src/js/utils/events.js
BrandonOCasey: Yeah this is the main fix

BrandonOCasey: It prevents `getData` from being called on something like `document`.

src/js/tracks/track-list.js
```diff
@@ -115,7 +116,8 @@ class TrackList extends EventTarget {
      */
     this.trigger({
       track,
-      type: 'removetrack'
+      type: 'removetrack',
+      target: this
```
BrandonOCasey: I think the target here is intended to be the tracklist that the event is triggered on and not document.

src/js/tracks/track-list.js
```diff
@@ -72,7 +72,8 @@ class TrackList extends EventTarget {
        */
       this.trigger({
         track,
-        type: 'addtrack'
+        type: 'addtrack',
+        target: this
```
BrandonOCasey: I think the target here is intended to be the tracklist that the event is triggered on and not document.

src/js/event-target.js
```diff
@@ -138,6 +138,11 @@ EventTarget.prototype.one = function(type, fn) {
 EventTarget.prototype.trigger = function(event) {
   const type = event.type || event;
 
+  // deprecation
+  // In a future version we should default target to `this`
+  // similar to how we default the target to `elem` in
```
BrandonOCasey: I am not sure that we can fix `EventTargets` setting `target` to `document` by default in anything but a major version.

src/js/tracks/track-list.js
gkatsev: makes sense

PR #5860, opened 2019-03-13
fix: more memory leaks

BrandonOCasey:
## Description
More memory leaks in DomData.elData uncovered by looking at our unit tests. I separated the code changes here from the test changes which I will open in another pull request.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -425,6 +425,15 @@ class Player extends Component {
       this.autoplay(this.options_.autoplay);
     }
 
+    // check plugins
+    if (options.plugins) {
+      Object.keys(options.plugins).forEach((name) => {
+        if (typeof this[name] !== 'function') {
+          throw new Error(`plugin "${name}" does not exist`);
+        }
+      });
+    }
+
     /*
      * Store the internal state of scrubbing
      *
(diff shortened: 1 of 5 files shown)


src/js/player.js
```diff
@@ -425,6 +425,15 @@ class Player extends Component {
       this.autoplay(this.options_.autoplay);
     }
 
+    // check plugins
+    if (options.plugins) {
```
BrandonOCasey: If we throw an error after making the player `evented` it causes a dispose handler to be created and then lost due to the error. We can either do this, or trigger dispose on the player below before we throw the error but I think this made more sense.

src/js/tracks/text-track.js
```diff
@@ -91,14 +91,21 @@ const loadTrack = function(src, track) {
     // NOTE: this is only used for the alt/video.novtt.js build
     if (typeof window.WebVTT !== 'function') {
       if (track.tech_) {
-        const loadHandler = () => parseCues(responseBody, track);
-
-        track.tech_.on('vttjsloaded', loadHandler);
-        track.tech_.on('vttjserror', () => {
+        // to prevent use before define eslint error, we define loadHandler
+        // as a let here
```
BrandonOCasey: I think that this was very likely a bug waiting to happen here since we never cleaned up the `vttjsloaded` and `vttjserror` handlers properly.

src/js/tech/tech.js
```diff
@@ -1261,7 +1261,7 @@ Tech.withSourceHandlers = function(_Tech) {
     }
 
     this.sourceHandler_ = sh.handleSource(source, this, this.options_);
-    this.on('dispose', this.disposeSourceHandler);
+    this.one('dispose', this.disposeSourceHandler);
```
BrandonOCasey: We want `disposeSourceHandler` itself to be removed on dispose so we need to use `one` instead of `on`. We could also use `off` in `disposeSourceHandler` but I think this is more clear.

gkatsev:
this could potentially be a `perf:` prefix (https://github.com/videojs/conventional-changelog-videojs/blob/master/convention.md#type)

PR #5861, opened 2019-03-13
tests: memory leak fixes in tests

BrandonOCasey:
Using `DomData.elData` contained a ton of improperly cleaned up objects in tests this pull request fixes that.

### DIFF
--- a/test/unit/tech/html5.test.js
+++ b/test/unit/tech/html5.test.js
@@ -444,13 +444,13 @@ if (Html5.supportsNativeTextTracks()) {
 
     el.textTracks = tt;
 
-    /* eslint-disable no-unused-vars */
     const htmlTech = new Html5({el});
-    /* eslint-enable no-unused-vars */
 
     assert.equal(adds[0][0], 'change', 'change event handler added');
     assert.equal(adds[1][0], 'addtrack', 'addtrack event handler added');
     assert.equal(adds[2][0], 'removetrack', 'removetrack event handler added');
+
+    htmlTech.dispose();
   });
 
   QUnit.test('does not add native textTrack listeners when disabled', function(assert) {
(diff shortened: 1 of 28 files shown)


test/unit/tech/html5.test.js
```diff
@@ -444,13 +444,13 @@ if (Html5.supportsNativeTextTracks()) {
 
     el.textTracks = tt;
 
-    /* eslint-disable no-unused-vars */
     const htmlTech = new Html5({el});
```
BrandonOCasey: I guess we should have thought about `no-unused-vars` a bit more here. woops!

test/unit/tracks/text-track-controls.test.js
```diff
@@ -364,6 +364,8 @@ QUnit.test('menu items should polyfill mode change events', function(assert) {
   assert.equal(changes, 2, 'clicks trigger change events');
 
   player.dispose();
+  trackMenuItem.dispose();
+  player.textTracks().off('change');
```
BrandonOCasey: I almost wonder if we should call `off` on track lists when we dispose, instead of doing this.

test/unit/tracks/track.test.js
```diff
@@ -21,8 +20,9 @@ TrackBaseline(Track, {
   kind: 'subtitles',
   mode: 'disabled',
   label: 'English',
-  language: 'en',
-  tech: new TechFaker()
```
BrandonOCasey: Since we need a way to cleanup this tech, we cannot just pass a `new TechFaker` into baseline. Instead I create one in every test inside `TrackBaseline` and pass it as an options.

test/unit/tracks/text-track-controls.test.js
gkatsev: probably, though, I thought we did that?

BrandonOCasey: I think we clear them out of `Tracks`. Let's see...

BrandonOCasey: Doesn't seem like it: https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L367

BrandonOCasey: Maybe we want to call `off` on them on player dispose, rather than tech dispose, since the track lists stick around.

BrandonOCasey: I will remove the test changes related to this if #5867

PR #5862, opened 2019-03-13
tests: for memory leaks and duplicate event listeners

BrandonOCasey:
A final pull request in the memory leak testing that I did. This one only test is there is a memory leak or not. 

This test should fail until the following pull requests are merged:
* #5861 
* #5860 
* #5855
* #5853


test/unit/utils/dom-data.test.js
```diff
@@ -21,3 +23,76 @@ QUnit.test('should get and remove data from an element', function(assert) {
 
   assert.notOk(DomData.hasData(el), 'cached item emptied');
 });
+
+QUnit.done(function(details) {
```
BrandonOCasey: It doesn't seem like tests are supposed to be run in `QUnit.done` and it makes re-running them a bit awkward, but I don't see another way to check for memory leaks other than waiting until done.

test/unit/video.test.js
```diff
@@ -603,3 +606,131 @@ QUnit.test('adds video-js class name with the video-js embed', function(assert)
...
+ * as it runs through a basic player lifecycle for real.
+ */
+QUnit.test('create a real player and play', function(assert) {
+  assert.timeout(30000);
+  const done = assert.async();
+  const fixture = document.getElementById('qunit-fixture');
+  const old = {};
+
+  // TODO: remove this code when we have a videojs debug build
+  // see https://github.com/videojs/video.js/issues/5858
```
BrandonOCasey: Technically this doesn't cover all the cases where we will use wrappers for event listeners, but for now it would have caught the bug that we were seeing.

test/unit/video.test.js
```diff
@@ -603,3 +606,131 @@ QUnit.test('adds video-js class name with the video-js embed', function(assert)
   assert.ok(player.hasClass('video-js'), 'video-js class was added to the first embed');
   assert.ok(player2.hasClass('video-js'), 'video-js class was preserved to the second embed');
 });
+
+QUnit.module('videojs integration');
```
gkatsev: can this be moved into a separate file, actually?

BrandonOCasey: sure

test/unit/video.test.js
```diff
@@ -603,3 +606,131 @@ QUnit.test('adds video-js class name with the video-js embed', function(assert)
...
+          }
+        });
+      });
+    });
+  };
+
+  // play
+  player.play();
+  player.one('timeupdate', () => {
+    // then wait 1s and pause/dispose
```
gkatsev: this comment says 1s but the code says 2000ms :)

BrandonOCasey: fixed

test/unit/videojs-integration.test.js
```diff
@@ -0,0 +1,134 @@
...
+            assert.ok(false, `${elId}/${eventName}/${obj.fn.name} is not unique`);
+          }
+        });
+      });
+    });
+  };
+
+  // play
+  player.play();
+  player.one('timeupdate', () => {
```
gkatsev: should we wait for `playing` before adding this `timeupdate` handler?

gkatsev: Also, probably call done() with an error from `player.one('error')`

test/unit/utils/dom-data.test.js
```diff
@@ -21,3 +22,45 @@ QUnit.test('should get and remove data from an element', function(assert) {
...
+   * ```
+   * to the tests to see what test. Then re-run the tests to see
+   * what leaking and where.
+   *
+   * > Note that the id can be off by 1-2 in either direction
+   *   for larger guids, so you may have to account for that.
+   */
+  QUnit.test('Memory is not leaking', function(assert) {
+    if (Object.keys(DomData.elData).length > 0) {
+      videojs.domData = DomData;
```
gkatsev: This is here for easier debugging for when/if this test fails?

BrandonOCasey: yeah

PR #5876, opened 2019-03-18
Fix #5875 - Don't call video.load in resetMediaElement if the video is trying to play

### LINKED ISSUE #5875: "The play request was interrupted by a new load request"
## Description
Briefly describe the issue.
Reproduction is not trivial, but we are seeing many errors in our company error log due to this issue. The issue occurs when a `play` request is in-motion at the time of teardown. On teardown, `resetMediaElement` is called which proceeds to call `video.load()`.

## Steps to reproduce
Again, difficult to setup a reproducible demo but here is the issue, [simplified in a jsFiddle](https://jsfiddle.net/pe1htmw8/1/). See the console for an error.

### Expected
There should be no error.

### Error output
"The play request was interrupted by a new load request"

## Additional Information
#### videojs
video.js @7.4.1
#### browsers
Chrome
#### OSes
I'm currently on a Mac, but this is an OS-independent issue
### plugins
None.

evanfarina:
Fixes https://github.com/videojs/video.js/issues/5875

In addition to the test that I added, I also manually tested that the <sources> are removed from the video once Player.reset is called. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3151,6 +3151,16 @@ class Player extends Component {
    * and calls `reset` on the `tech`.
    */
   reset() {
+    if (this.paused()) {
+      this.doReset_();
+    } else {
+      const playPromise = this.play();
+
+      playPromise.then(() => this.doReset_());
+    }
+  }
+
+  doReset_() {
     if (this.tech_) {
       this.tech_.clearTracks('text');
     }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3156,7 +3156,7 @@ class Player extends Component {
     } else {
       const playPromise = this.play();
 
-      playPromise.then(() => this.doReset_());
+      silencePromise(playPromise.then(() => this.doReset_()));
     }
   }
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3151,7 +3151,9 @@ class Player extends Component {
    * and calls `reset` on the `tech`.
    */
   reset() {
-    if (this.paused()) {
+    const PromiseClass = this.options_.Promise || window.Promise;
+
+    if (this.paused() || !PromiseClass) {
       this.doReset_();
     } else {
       const playPromise = this.play();
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1245,17 +1245,7 @@ Html5.disposeMediaElement = function(el) {
     // wrapping in an iife so it's not deoptimized (#1060#discussion_r10324473)
     (function() {
       try {
-        if (el.paused) {
-          el.load();
-        } else {
-          const playPromise = el.play();
-
-          if (playPromise) {
-            playPromise.then(() => el.load());
-          } else {
-            el.load();
-          }
-        }
+        el.load();
       } catch (e) {
         // not supported
       }
@@ -1283,19 +1273,9 @@ Html5.resetMediaElement = function(el) {
     // wrapping in an iife so it's not deoptimized (#1060#discussion_r10324473)
     (function() {
       try {
-        if (el.paused) {
-          el.load();
-        } else {
-          const playPromise = el.play();
-
-          if (playPromise) {
-            playPromise.then(() => el.load());
-          } else {
-            el.load();
-          }
-        }
+        el.load();
       } catch (e) {
-        // not supported
+        // satisfy linter
       }
     }());
   }
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1245,7 +1245,17 @@ Html5.disposeMediaElement = function(el) {
     // wrapping in an iife so it's not deoptimized (#1060#discussion_r10324473)
     (function() {
       try {
-        el.load();
+        if (el.paused) {
+          el.load();
+        } else {
+          const playPromise = el.play();
+
+          if (playPromise) {
+            playPromise.then(() => el.load());
+          } else {
+            el.load();
+          }
+        }
       } catch (e) {
         // not supported
       }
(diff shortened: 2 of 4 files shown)


test/unit/tech/html5.test.js
```diff
@@ -874,19 +876,22 @@ QUnit.test('Html5.resetMediaElement should remove sources and call load', functi
...
+        removedChildren,
+        children.reverse(),
+        'we removed the children that were present'
+      );
+      assert.equal(removedAttribute, 'src', 'we removed the src attribute');
+      done();
+    },
+
+    play() {
+      return window.Promise.resolve('foo');
```
gkatsev: looks like this is safe because we add es6-shim

test/unit/tech/html5.test.js
```diff
@@ -925,6 +930,19 @@ QUnit.test('When Android Chrome reports Infinity duration with currentTime 0, re
   tech.el_ = oldEl;
 });
 
+QUnit.test('No error is thrown when `load` is called `immediately` after play', function(assert) {
+  const videoTag = TestHelpers.makeTag();
+  const fixture = document.getElementById('qunit-fixture');
+
+  fixture.appendChild(videoTag);
+  const playerInstance = videojs(videoTag.id, { techOrder: ['html5'] });
```
gkatsev: we should dispose of this player at the end of the test (`playerInstance.dispose()` after the assertion)

test/unit/tech/html5.test.js
```diff
@@ -925,6 +930,19 @@ QUnit.test('When Android Chrome reports Infinity duration with currentTime 0, re
...
+
+  playerInstance.play();
+  // `load` is called as part of `disposeMediaElement`
+  Html5.disposeMediaElement();
```
gkatsev: is the promise error enough to cause this test to fail?

src/js/tech/html5.js
```diff
@@ -1245,7 +1245,17 @@ Html5.disposeMediaElement = function(el) {
...
     (function() {
       try {
-        el.load();
+        if (el.paused) {
+          el.load();
+        } else {
+          const playPromise = el.play();
+
+          if (playPromise) {
+            playPromise.then(() => el.load());
```
gkatsev: we probably want to add an empty function (or use silencePromise https://github.com/videojs/video.js/blob/master/src/js/utils/promise.js#L24) to the second part of then to make sure that it doesn't throw another error

either
```js
playPromise.then(() => el.load(), () => {});
```
or
```
// and add an import for silencePromise
silencePromise(playPromise.then(() => el.load()));
```

gkatsev: and below as well

test/unit/tech/html5.test.js
evanfarina: Done.

test/unit/tech/html5.test.js
evanfarina: Updated the test in the latest commit. It indeed fails every time without my fix.

src/js/tech/html5.js
evanfarina: Added in `reset`

test/unit/player.test.js
```diff
@@ -1428,59 +1428,22 @@ QUnit.test('Player#tech logs a warning when called without a safety argument', f
...
-  assert.equal(loadedSource, null, 'with a null source');
-  assert.equal(techCallMethod, 'reset', 'we then reset the tech');
-});
+  this.clock.restore();
+  const done = assert.async();
+  const player = TestHelpers.makePlayer();
+  const loadTechSpy = sinon.spy(player, 'loadTech_');
+  const techCallSpy = sinon.spy(player, 'techCall_');
 
-QUnit.test('player#reset loads the first item in the techOrder and then techCalls reset', function(assert) {
```
gkatsev: are we losing this test?

test/unit/player.test.js
```diff
@@ -1428,59 +1428,22 @@ QUnit.test('Player#tech logs a warning when called without a safety argument', f
...
-  };
-
-  Player.prototype.reset.call(testPlayer);
-
-  assert.equal(loadedTech, 'html5', 'we loaded the html5 tech');
-  assert.equal(loadedSource, null, 'with a null source');
-  assert.equal(techCallMethod, 'reset', 'we then reset the tech');
-});
+  this.clock.restore();
+  const done = assert.async();
```
gkatsev: we try to avoid making tests async if we can.

test/unit/player.test.js
```diff
@@ -1428,59 +1428,22 @@ QUnit.test('Player#tech logs a warning when called without a safety argument', f
 });
 
 QUnit.test('player#reset loads the Html5 tech and then techCalls reset', function(assert) {
```
gkatsev: I think we should keep this test as is and add a new test to test for the delayed reset functionality.

test/unit/player.test.js
evanfarina: I have to disagree here. The previous test was heavily stubbing the player object as well as relying a particular internal API. What I'm most concerned about, though, is the fact that the test was forcing the internal methods of `reset` to be called, even though this code wouldn't work in a prod environment. In prod, the player needs to be `ready` before we can interact with the tech. The point is, if we are forcing behavior to work in tests, then we can't rely on these tests to make sure that our production code will work as expected.

gkatsev:
So, I think I do need to re-evaluate the new test code by itself as it's kind of hard to read in the diff. I do think that the current tests have value even if they aren't perfect and I stand by preferring non-async tests. This can often be achieved with the fake timers from sinon that we use and it's something that we should evaluate.

However, I do think that the change is good, what about separating this PR into two? One of the change to player.js and one for the tests? This was we can get the change in quickly and then we can have a greater discussion around tests and get those in a bit later.

src/js/player.js
```diff
@@ -3151,6 +3151,16 @@ class Player extends Component {
    * and calls `reset` on the `tech`.
    */
   reset() {
+    if (this.paused()) {
+      this.doReset_();
+    } else {
+      const playPromise = this.play();
```
BrandonOCasey: play won't always return a promise

PR #5880, opened 2019-03-20
perf: Fix memory leaks in safari, edge, and ie

### LINKED ISSUE #5878: Investigate dom data usage in tests on Safari, Edge, and IE
In https://github.com/videojs/video.js/pull/5862, we added a test that verifies that dom data is cleared out when a player is disposed. Due to time constraints, we weren't able to look at why Safari, Edge, and IE fail this test. Instead, we denylisted them for the test.
We should re-enable the tests on these browsers and investigate why they are failing.

BrandonOCasey:
## Description
* These where caused because we were not always able to clean up `resize` on the `ResizeManager`s iframe contentWindow after
* Native tracks on safari do not have an `off` so we have to remove our event listeners manually.

Fixes #5878

### DIFF
--- a/src/js/resize-manager.js
+++ b/src/js/resize-manager.js
@@ -71,7 +71,18 @@ class ResizeManager extends Component {
           return;
         }
 
-        Events.on(this.el_.contentWindow, 'resize', this.debouncedHandler_);
+        const debouncedHandler_ = this.debouncedHandler_;
+        let unloadListener_ = this.unloadListener_ = function() {
+          Events.off(this, 'resize', debouncedHandler_);
+          Events.off(this, 'unload', unloadListener_);
+
+          unloadListener_ = null;
+        };
+
+        // safari and edge can unload the iframe before resizemanager dispose
+        // we have to dispose of event handlers correctly before that happens
+        Events.on(this.el_.contentWindow, 'unload', unloadListener_);
+        Events.on(this.el_.contentWindow, 'resize', debouncedHandler_);
       };
 
       this.one('load', this.loadListener_);
@@ -120,12 +131,8 @@ class ResizeManager extends Component {
       this.resizeObserver_.disconnect();
     }
 
-    if (this.el_ && this.el_.contentWindow) {
-      Events.off(this.el_.contentWindow, 'resize', this.debouncedHandler_);
-    }
-
-    if (this.loadListener_) {
-      this.off('load', this.loadListener_);
+    if (this.el_ && this.el_.contentWindow && this.unloadListener_) {
+      this.unloadListener_.call(this.el_.contentWindow);
     }
 
     this.ResizeObserver = null;
--- a/src/js/resize-manager.js
+++ b/src/js/resize-manager.js
@@ -131,6 +131,10 @@ class ResizeManager extends Component {
       this.resizeObserver_.disconnect();
     }
 
+    if (this.loadListener_) {
+      this.off('load', this.loadListener_);
+    }
+
     if (this.el_ && this.el_.contentWindow && this.unloadListener_) {
       this.unloadListener_.call(this.el_.contentWindow);
     }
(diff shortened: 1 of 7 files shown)


src/js/resize-manager.js
```diff
@@ -71,7 +71,18 @@ class ResizeManager extends Component {
...
+        let unloadListener_ = this.unloadListener_ = function() {
+          Events.off(this, 'resize', debouncedHandler_);
+          Events.off(this, 'unload', unloadListener_);
+
+          unloadListener_ = null;
+        };
+
+        // safari and edge can unload the iframe before resizemanager dispose
+        // we have to dispose of event handlers correctly before that happens
+        Events.on(this.el_.contentWindow, 'unload', unloadListener_);
```
BrandonOCasey: We have to listen to the content window `unload` event so that we always dispose the iframe correctly if we add events to it. We also make sure to use the `context` of the `contentWindow` as `this` since IE11 does not work any other way for some reason.

test/unit/tracks/text-track-list-converter.test.js
```diff
@@ -140,6 +140,9 @@ if (Html5.supportsNativeTextTracks()) {
     c.jsonToTextTracks(cleanup(c.textTracksToJson(tech)), tech);
 
     assert.equal(addRemotes, 2, 'we added two text tracks');
+
+    tt.removeTrack(nativeTrack.track);
```
BrandonOCasey: This test was leaking but only run on safari

test/unit/tracks/text-tracks.test.js
```diff
@@ -272,12 +272,11 @@ if (Html5.supportsNativeTextTracks()) {
       assert.equal(emulatedTt.length, tt.length, 'we have matching tracks length');
       assert.equal(emulatedTt.length, 1, 'we have one text track');
 
-      emulatedTt.off('addtrack', addtrack);
       el.removeChild(track);
     };
 
-    emulatedTt.on('addtrack', addtrack);
-    emulatedTt.on('removetrack', function() {
+    emulatedTt.one('addtrack', addtrack);
```
BrandonOCasey: This test was leaking but only run on safari

test/unit/utils/dom-data.test.js
```diff
@@ -60,6 +56,7 @@ QUnit.done(function(details) {
   QUnit.test('Memory is not leaking', function(assert) {
     if (Object.keys(DomData.elData).length > 0) {
       videojs.domData = DomData;
+      window.videojs = videojs;
```
BrandonOCasey: It seems that videojs isn't always on `window`

src/js/resize-manager.js
```diff
@@ -120,12 +131,8 @@ class ResizeManager extends Component {
       this.resizeObserver_.disconnect();
     }
 
-    if (this.el_ && this.el_.contentWindow) {
-      Events.off(this.el_.contentWindow, 'resize', this.debouncedHandler_);
-    }
-
-    if (this.loadListener_) {
```
gkatsev: we should keep this around, while unlikely, it is still possible that dispose happens before the the loadListener is triggered.

gkatsev: Or does this end up being cleaned up automatically? If so, maybe it doesn't need to be stored on `this` but just as a local variable?

BrandonOCasey: I think anything on `this` is cleaned up automatically, but I did not intend to remove this, let me fix it.

src/js/tracks/audio-track-list.js
```diff
@@ -72,22 +72,34 @@ class AudioTrackList extends TrackList {
       return;
     }
 
+    if (!this.enabledChange_) {
+      this.enabledChange_ = () => {
```
gkatsev: looks like this change causes the first track to always get selected because we aren't changing it when new tracks come in. https://github.com/videojs/video.js/issues/5887

Probably similar issue in video tracks and text tracks.

PR #5881, opened 2019-03-20
feat: respect preload none

brandonocasey:
> I am not sure this pull request is the correct way to accomplish this, but I want to get a discussion started. 

## Description
Currently we attempt to set the source right away, regardless of the preload option. This seems like it goes against `preload` `none` to me. The browser itself does appear to set the source, but it cancels any pending source request if it detects that we have `preload` set to  `none`. On our end we can't really do that as there isn't a way for middleware to postpone source setting until the user wants to play, without doing crazy things like overriding the players `play` function. The way I see it we have two options to handle this.

Below I have two options laid out, personally I think option 1 is better, because it lets us delay more than just middleware source setting.

### Option 1
Do it in Video.js similar to how it is done here

### Option 2
Do something to allow middleware to postpone source setting until a play:
1. Add a new `playcalled` event that will trigger on tech play, or when `Player#play()` is called
2. Add a new `playCalled` getter to `Player` that is set to true when `playcalled` triggers
3. Make sure that we reset `playCalled` to `false` before going through middleware sourceset.
4. probably add a helper function to make this easier. 

This will allow middleware to delay `sourcesetting` until a `playcalled` event and even when it gets delayed, as long as the `playcalled` getter is respected downstream middleware will still set their source. 

Example:
```js
/**
 * preferably we would provide this on videojs.middleware
 * Returns true if we will delay, false otherwise
 */
const delayForPreloadNone = videojs.middleware.delayForPreloadNone = function(player, cb) {
  // respect preload none and delay until play is called
  if (player.playCalled() || player.preload() !== 'none' || player.autoplay()) {
    return false;
  }

  // turn off previous delay function as a new src call happened
  if (this.delayForPreloadNone_) {
    player.off('playcalled', this.delayForPreloadNone_);
  }

  this.delayForPreloadNone_ = () => {
    this.delayForPreloadNone_ = null;
    cb();
  };
  player.one('playcalled', this.delayForPreloadNone_);
  return true;
};

videojs.use('content/vmap', (player) => ({
  setSource(source, next) {

    if (delayForPreloadNone.call(this, () => this.setSource(source, next))) {
      return;
    }

    /**
     * for this demo:
     * 1. assume the the content/vmap is handled
     * 2. we find that the type and source should now be content/m3u8
     */
    next(source, next);
  }
}));

videojs.use('content/m3u8', (player) => ({
  setSource(source, next) {

    /**
     * In this case this middleware is downstream from
     * content/vmap middleware. So we will only ever get here
     * when `player.playCalled()` is true. But if a `content/m3u8`
     * were passed in we would delay just like in `content/vmap`
     */
    if (delayForPreloadNone.call(this, () => this.setSource(source, next))) {
      return;
    }

    next(source, next);
  }
}));
```

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3038,16 +3038,19 @@ class Player extends Component {
     this.cache_.sources = sources;
     this.updateSourceCaches_(sources[0]);
 
+    // clear any previous src call that is waiting for `playcalled`
     if (this.srcAfterPlayCalled_) {
       this.off('playcalled', this.srcAfterPlayCalled_);
     }
 
+    // if preload is none, there is no autoplay, and play has not been called
     if (!this.playCalled() && this.preload() === 'none' && !this.autoplay()) {
       this.srcAfterPlayCalled_ = () => this.src(source);
       this.one('playcalled', this.srcAfterPlayCalled_);
       return;
     }
 
+    // reset play called to false now that we are past the loop
     this.playCalled(false);
 
     // middlewareSource is the source after it has been changed by middleware
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -367,6 +367,9 @@ class Player extends Component {
     // Init state hasStarted_
     this.hasStarted_ = false;
 
+    // Init state for playCalled_
+    this.playCalled_ = false;
+
     // Init state userActive_
     this.userActive_ = false;
 
@@ -1565,6 +1568,20 @@ class Player extends Component {
     }
   }
 
+  playCalled(val) {
+    if (val === undefined) {
+      return this.playCalled_;
+    }
+
+    if (val && !this.playCalled_) {
+      this.playCalled_ = true;
+      this.trigger('playcalled');
+      return;
+    }
+
+    this.playCalled_ = Boolean(val);
+  }
+
   /**
    * Fired whenever the media begins or resumes playback
    *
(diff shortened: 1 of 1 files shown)


gkatsev:
I think that aside from middleware, the way that preload none is handled is fine (well, we probably should honor it for captions) and that techs/source handlers should handle honor preload.
It does make sense that we want something for middleware to not require them to do work before play is requested.
I wonder what will happen if when setSource is called in middleware and we don't call `next`, would we be able to request play? Then, we could theoretically call next() with the source in the play request mw method? Definitely not the ultimate solution, but I wonder if it'll be ok as a workaround.

gkatsev:
[Looking at the spec](https://html.spec.whatwg.org/multipage/media.html#concept-media-load-algorithm), seems like `currentSrc` ends up not being set if we're loading from a source element with preload=none but it does get set if we're using a src attribute.

BrandonOCasey:
You won't be able to call play as `changingSrc_` will be true, plus its kind of a chicken and the egg as the tech won't be loaded to call play on until we know the source.

gkatsev:
Yeah, that's what I figured, but I can hope, can't I? 😜

BrandonOCasey:
That is the main reason I think we need an event of some kind to be triggers when `Player#play()` is called. Since we do not trigger `play` until we actually call `Tech#play()`. Would it make more sense for this logic to be a part of `middleware.setSource`? or should it just a helper for middleware to handle as needed. Personally I think that most middleware that will use `setSource` will actually need something like this, since most of the reason to use a `setSource` middleware is to support something not-native.

PR #5888, opened 2019-03-23
7.4.1  result.firstKeyFrame is null

naiveliang:
## Description
When I was using video.js with version of 7.4.1, I found a bug at the line 37806, like the following:
`37804        if (probe$2.ts.videoPacketContainsKeyFrame(frame)) {`
`37805             result.firstKeyFrame = probe$2.ts.parsePesTime(frame);`
`37806             result.firstKeyFrame.type = 'video';`
`37807        }`

The return value of 'probe$2.ts.parsePesTime(frame)' may be null, so 'result.firstKeyFrame.type' will throw an error, like this: Uncaught TypeError: Cannot set property 'type' of null.

## Specific Changes proposed
To solve this error, I add a code before line 37806, like the following:
`37804        if (probe$2.ts.videoPacketContainsKeyFrame(frame)) {`
`37805             result.firstKeyFrame = probe$2.ts.parsePesTime(frame);`
`37806             if(result.firstKeyFrame==null) result.firstKeyFrame = {};`
`37807             result.firstKeyFrame.type = 'video';`
`37808        }`

This is able to solve my problem.


BrandonOCasey:
Thanks for bringing this issue to our attention. The code that you referenced does not actually live in this repo (even though we ship it with Video.js). It lives here: https://github.com/videojs/mux.js/blob/master/lib/tools/ts-inspector.js#L222 

in the https://github.com/videojs/mux.js repo

That would be were you want to submit your fix, and not here in the `video.js` repo.

gkatsev:
This should've been an issue rather than PR. However, looks like an issue already exists (https://github.com/videojs/mux.js/issues/250) and there's a pending PR for it as well: https://github.com/videojs/mux.js/pull/251

PR #5895, opened 2019-03-26
fix: Correctly resolve play on middleware terminate

BrandonOCasey:
## Description
Currently when a middleware terminates a call to play, we resolve the promise right away, event though that promise is not resolved. We should instead resolve the promise on the next call to play that works.

Doing this turned out to be a bit tricker than first anticipated. I added detailed tests that make sure that the play `retval` is resolved in a slew different scenarios. I think that we will still need another pull request to handle `play-terminate` events in the `manualAutoplay_` function, but this pull request was getting to big already.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -531,8 +531,8 @@ class Player extends Component {
     this.responsive(this.options_.responsive);
 
     this.changingSrc_ = false;
-    this.playWaitingForReady_ = false;
-    this.playOnLoadstart_ = null;
+    this.waitForPlay_ = false;
+    this.playSuccessCallbacks_ = [];
   }
 
   /**
@@ -2209,54 +2209,58 @@ class Player extends Component {
    *        The callback that should be called when the techs play is actually called
    */
   play_(callback = silencePromise) {
-    // If this is called while we have a play queued up on a loadstart, remove
-    // that listener to avoid getting in a potentially bad state.
-    if (this.playOnLoadstart_) {
-      this.off('loadstart', this.playOnLoadstart_);
+    this.playSuccessCallbacks_.push(callback);
+
+    const isSrcReady = Boolean(!this.changingSrc_ && (this.src() || this.currentSrc()));
+
+    if (this.waitToPlay_) {
+      this.off(['ready', 'loadstart'], this.waitToPlay_);
+      this.waitToPlay_ = null;
     }
 
-    // If the player/tech is not ready, queue up another call to `play()` for
-    // when it is. This will loop back into this method for another attempt at
-    // playback when the tech is ready.
-    if (!this.isReady_) {
-
-      // Bail out if we're already waiting for `ready`!
-      if (this.playWaitingForReady_) {
-        return;
-      }
-
-      this.playWaitingForReady_ = true;
-      this.ready(() => {
-        this.playWaitingForReady_ = false;
-        callback(this.play());
-      });
-
-    // If the player/tech is ready and we have a source, we can attempt playback.
-    } else if (!this.changingSrc_ && (this.src() || this.currentSrc())) {
-      callback(this.techGet_('play'));
-      return;
-
-    // If the tech is ready, but we do not have a source, we'll need to wait
-    // for both the `ready` and a `loadstart` when the source is finally
-    // resolved by middleware and set on the player.
-    //
-    // This can happen if `play()` is called while changing sources or before
-    // one has been set on the player.
-    } else {
-
-      this.playOnLoadstart_ = () => {
-        this.playOnLoadstart_ = null;
-        callback(this.play());
+    // if the player/tech is not ready or the src itself is not ready
+    // queue up a call to play on `ready` or `loadstart`
+    if (!this.isReady_ || !isSrcReady) {
+      this.waitToPlay_ = (e) => {
+        this.play_();
       };
+      this.one(['ready', 'loadstart'], this.waitToPlay_);
 
       // if we are in Safari, there is a high chance that loadstart will trigger after the gesture timeperiod
       // in that case, we need to prime the video element by calling load so it'll be ready in time
-      if (browser.IS_ANY_SAFARI || browser.IS_IOS) {
+      if (!isSrcReady && (browser.IS_ANY_SAFARI || browser.IS_IOS)) {
         this.load();
       }
-      this.one('loadstart', this.playOnLoadstart_);
+      return;
     }
 
+    // If the player/tech is ready and we have a source, we can attempt playback.
+    const val = this.techGet_('play');
+
+    if (val === middleware.TERMINATOR) {
+      this.trigger('play-terminated');
+    } else {
+      this.runPlaySuccessCallbacks_(val);
+    }
+  }
+
+  /**
+   * When a callback to play is delayed we have to run these
+   * callbacks when play is actually called on the tech. This function
+   * runs the callbacks that were delayed and accepts the return value
+   * from the tech.
+   *
+   * @param {undefined|Promise} val
+   *        The return value from the tech.
+   */
+  runPlaySuccessCallbacks_(val) {
+    const callbacks = this.playSuccessCallbacks_.slice(0);
+
+    this.playSuccessCallbacks_ = [];
+
+    callbacks.forEach(function(cb) {
+      cb(val);
+    });
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/js/player.js
```diff
@@ -2209,54 +2209,59 @@ class Player extends Component {
...
-    // If the tech is ready, but we do not have a source, we'll need to wait
-    // for both the `ready` and a `loadstart` when the source is finally
-    // resolved by middleware and set on the player.
-    //
-    // This can happen if `play()` is called while changing sources or before
-    // one has been set on the player.
-    } else {
+    // treat calls to play_ somewhat like the `one` event function
+    if (this.waitToPlay_) {
+      this.off(['ready', 'loadstart'], this.waitToPlay_);
```
BrandonOCasey: previously we had play on loadstart and waiting for ready, but they basically did the same thing. (except for a `load` call on Safari/ios)

src/js/player.js
```diff
@@ -2209,54 +2209,59 @@ class Player extends Component {
...
-    // that listener to avoid getting in a potentially bad state.
-    if (this.playOnLoadstart_) {
-      this.off('loadstart', this.playOnLoadstart_);
-    }
-
-    // If the player/tech is not ready, queue up another call to `play()` for
-    // when it is. This will loop back into this method for another attempt at
-    // playback when the tech is ready.
-    if (!this.isReady_) {
+    this.playSuccessCallbacks_.push(callback);
```
BrandonOCasey: We add every callback to a list, that gets run when a `this.techGet_('play')` is resolved to any non-terminated value.

src/js/tech/middleware.js
```diff
@@ -166,7 +166,7 @@ export function mediate(middleware, tech, method, arg = null) {
   const callMethod = 'call' + toTitleCase(method);
   const middlewareValue = middleware.reduce(middlewareIterator(callMethod), arg);
   const terminated = middlewareValue === TERMINATOR;
-  const returnValue = terminated ? null : tech[method](middlewareValue);
+  const returnValue = terminated ? TERMINATOR : tech[method](middlewareValue);
```
BrandonOCasey: I also noticed that it is possible for our `player` values to be set to `null`, and now `TERMINATOR` if a `call<method>` middleware terminates. We may want to think about handling that.

src/js/player.js
gkatsev: I think the name here can be confusing. While we don't handle rejecting the promise ourselves yet (we really should add it at some point), it doesn't mean that play was successful, just that it wasn't terminated, since we call it with the promise from `play()` which may end up rejecting the promise.

src/js/player.js
```diff
@@ -2209,54 +2209,59 @@ class Player extends Component {
...
       }
-      this.one('loadstart', this.playOnLoadstart_);
+      return;
     }
 
+    // If the player/tech is ready and we have a source, we can attempt playback.
+    const val = this.techGet_('play');
+
+    if (val === middleware.TERMINATOR) {
+      this.trigger('play-terminated');
```
gkatsev: I really wish we had a private event bus we can use. I'd probably prefer if we didn't trigger this if we didn't have to.

src/js/tech/middleware.js
BrandonOCasey: Would it be better to just return `TERMINATOR` for the `play` method? At least we can handle that?

src/js/player.js
BrandonOCasey: renamed this variable and the function for it `playCallbacks_`

src/js/player.js
BrandonOCasey: We could have something like a `playTerminatedQueue`, that we clear when we run `playCallbacks` and run/clear when play is terminated.

gkatsev: that might be better, people should preferably rely on the play promise or middleware to know whether play was rejected/not called

src/js/tech/middleware.js
gkatsev: Can't we check for `null` specifically? Regular functions in javascript return `undefined` if don't have an explicit return value.

test/unit/test-helpers.js
```diff
@@ -23,7 +23,9 @@ const TestHelpers = {
 
     const player = new Player(videoTag, playerOptions);
 
-    player.middleware_ = [player.tech_];
+    if (addTechAsMiddleware) {
```
gkatsev: I'm not sure why this is necessary

src/js/tech/middleware.js
```diff
@@ -5,7 +5,7 @@
 import { assign } from '../utils/obj.js';
 import toTitleCase from '../utils/to-title-case.js';
 
-const middlewares = {};
+export const middlewares = {};
```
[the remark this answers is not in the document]
BrandonOCasey: This is mostly for ease of use in cleaning up tests, but we could probably use `getMiddleware`

test/unit/test-helpers.js
BrandonOCasey: When using the HTML5 tech and not TechFaker, the tech shouldn't be added as a middleware. Adding the tech as middleware makes things happen in a different order then the way that they actually happens outside of tests.

test/unit/play.test.js
```diff
@@ -0,0 +1,442 @@
...
+import TestHelpers from './test-helpers.js';
+import sinon from 'sinon';
+import window from 'global/window';
+import * as middleware from '../../src/js/tech/middleware.js';
+import videojs from '../../src/js/video.js';
+
+const middleWareTerminations = ['terminates', 'does not-terminate'];
+const playReturnValues = ['non-promise'];
+
+if (window.Promise) {
```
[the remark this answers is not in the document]
BrandonOCasey: Well if es6-shim mocks this, then we could just run this tests against every browser, going to see how that goes.

src/js/tech/middleware.js
BrandonOCasey: let me see what ie 11 does, but it may be OK to check against null. I do think we may want to add a deprecation here for the future though.

src/js/tech/middleware.js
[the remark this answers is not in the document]
BrandonOCasey: fixed

test/unit/play.test.js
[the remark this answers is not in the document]
BrandonOCasey: test work on all browsers. Seems that the carry over from the old player promise checks for `window.Promise` were not needed.

src/js/player.js
BrandonOCasey: Done

src/js/tech/middleware.js
BrandonOCasey: We return null again now.

BrandonOCasey:
One more thing to talk about from #5909. 

### Background Info

A change was made so that:
1. `manualAutoplay` does nothing if `player.play()` does not return a `Promise` 
2. If we do not have access to a promise `autoplay` can only be set to true/false.

This was done because reverting `muted` for a play termination would be an excessive amount of code if play promises are not supported. 

Does this constitute a breaking change?

### Why I don't think it does
If `player.play does not return a promise:
1. The biggest reason is that `player.play` cannot reject/fail in a way that we can detect on browsers that do not return a promise from `player.play`. This means that we may as well rely on the builtin HTML5 `autoplay` since it should do everything that we could do.
2. if `autoplay` is `play` or `any` we will call `player.play()` and in the case of `any` will be unable to take other steps because a promise was not returned. 
3. if `autoplay` is `muted` we will mute the player and then call `player.play()` If the `play` fails, we won't be able to restore muted. Although we won't be able to tell if it fails because we rely on promises for that. 
4. `autoplay-success` and `autoplay-failure` will never trigger since they rely on a promise being returned

misteroneill:
Sounds reasonable to me. 👍

BrandonOCasey:
Ultimately if we really want to preserve the way things are we can: 
1. Move back to allowing the string values for any browser (even those without promises)
2. In `manualAutoplay` check for promises and if there isn't one call play and return right away (this is what we do right now anyway)

gkatsev:
So, the code looks good but we need to bring back `any`, `muted`, `play` support back regardless of promise support. Just tested on IE11 and it doesn't end up causing the player to autoplay, plus, if `muted` is passed, it should mute the player. This would be a breaking change if support is removed.

PR #5896, opened 2019-03-26
Fix/live tracker ie slowdown

BrandonOCasey:
## Description
~~We recently had a fix for `setInterval` performance in IE 11 but did not realize that `LiveTracker` has a similar `setInterval`. I added a helper to component that should be used for intervals like this in the future.~~ Originally it seemed like `setInterval` alone was the culprit, but after testing it seems that things in `startTracking` can cause the crash so we have to start and stop when the documents visibility changes.

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -1,5 +1,6 @@
 import Component from './component.js';
 import mergeOptions from './utils/merge-options.js';
+import document from 'global/document';
 
 /* track when we are at the live edge, and other helpers for live playback */
 class LiveTracker extends Component {
@@ -13,6 +14,25 @@ class LiveTracker extends Component {
     this.reset_();
 
     this.on(this.player_, 'durationchange', this.handleDurationchange);
+
+    // we don't need to track live playback if the document is hidden,
+    // also, tracking when the document is hidden can
+    // cause the CPU to spike and eventually crash the page on IE11.
+    if ('hidden' in document && 'visibilityState' in document) {
+      this.on(document, 'visibilitychange', this.handleVisibilityChange);
+    }
+  }
+
+  handleVisibilityChange() {
+    if (this.player_.duration() !== Infinity) {
+      return;
+    }
+
+    if (document.hidden) {
+      this.stopTracking();
+    } else {
+      this.startTracking();
+    }
   }
 
   isBehind_() {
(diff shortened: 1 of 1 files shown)


src/js/component.js
```diff
@@ -1384,6 +1385,84 @@ class Component {
...
+   * @param {number} interval
+   *        Execute the specified function every `x` milliseconds unless the document
+   *        is hidden
+   *
+   * @return {number} guid
+   *         The guid of the visible interval that can be used to clear it in
+   *         {@link Component#clearVisibleInterval}
+   */
+  setVisibleInterval(fn, interval) {
+    this.visibleIntervals_ = this.visibleIntervals_ || {};
```
BrandonOCasey: open to suggestions on a better way to do this, but It seemed like the simplest solution from the user perspective.

src/js/live-tracker.js
```diff
@@ -38,7 +39,6 @@ class LiveTracker extends Component {
   // all the functionality for tracking when seek end changes
   // and for tracking how far past seek end we should be
   trackLive_() {
-    this.pastSeekEnd_ = this.pastSeekEnd_;
```
BrandonOCasey: I can put this back but it seems like an error.

src/js/live-tracker.js
```diff
@@ -13,6 +14,25 @@ class LiveTracker extends Component {
     this.reset_();
 
     this.on(this.player_, 'durationchange', this.handleDurationchange);
+
+    // we don't need to track live playback if the document is hidden,
+    // also, tracking when the document is hidden can
+    // cause the CPU to spike and eventually crash the page on IE11.
+    if ('hidden' in document && 'visibilityState' in document) {
```
gkatsev: should we do this only on IE11?

src/js/live-tracker.js
```diff
@@ -13,6 +14,25 @@ class LiveTracker extends Component {
...
+
+  handleVisibilityChange() {
+    if (this.player_.duration() !== Infinity) {
+      return;
+    }
+
+    if (document.hidden) {
+      this.stopTracking();
+    } else {
+      this.startTracking();
```
gkatsev: does startTracking() do a check of when called? I think it does but wanted confirmation.

BrandonOCasey: Yeah it we are already tracking it will return right away.

src/js/live-tracker.js
BrandonOCasey: Yeah lets limit it to ie 11

Issue #5901, opened 2019-03-28
MIddlewares and React


iseekyouu:
## Description
There is a problem with cleaning up the middlewares, I can not remove the middlewares that Im added to videojs by myself.

### Expected
only one middleware on videojs after react component initialization;

### Actual
I am using React 16.6.
On the file page I have import videojs from 'video.js';
On componentDidMount im add middlware on videojs `videojs.use middleware ('*', player => ({...})`
on `componentWillUnmount`, I do `player.dispose()`;
Then, without overloading the page, I go to the login page (or any other, the main thing is that there is no reload), then I return to the page with the file and the following situation is obtained.
I have 2 middlvars hanging on videojs!
the first with the old component of which is already gone, and the second with the current one.

The documentation does not have a method for cleaning middlewares, there is a clearCacheForPlayer method called in `dispose()`, but it cleans `middlewareInstances`, this is useless, since all middlewares are stored in the middlewares object.

How can I remove all the middleware?
#### videojs
7.5.2
#### browsers
chrome latest
#### OSes
linux\windows\mac
### plugins
videojs-quality-selector

gkatsev:
The way that middleware are designed is that the middleware itself is attached globally to Video.js and then each player initialization, the middleware method is called with a new player. The middleware will then be called depending on whether setSource called `next` with an error or not.
You shouldn't be calling videojs.use() on componentDidMount.

iseekyouu:
and what can i do when i dont need middleware in some instance of player?

gkatsev:
The `setSource` method gets called with a `next` method, you can pass in an Error (or I think any thing that isn't `null` as the first argument to not attach the middleware to the current player/source combination. You can decide whether you want it to be used either by checking the source or even by checking a property/option on the player. We have [documentation on middleware](https://docs.videojs.com/tutorial-middleware.html). Here's an example:

```js
videojs.use('*', function(player) {
  return {
    setSource(srcObj, next) {
      if (player.options()['use-middleware']) {
        next(null, srcObj);
      } else {
        next(new Error("don't use middleware"));
      }
  };
});
```

Also, can I inquire into what you're using middleware? Depending on your usecase, we may want to modify middleware to make what you're doing easier, especially if we see more people doing what you're doing.

iseekyouu:
I use middleware to limit the allowed time to watch the video, in the video preview, my code: 
 ```
this.player = videojs(this.videoNode, videoOptions, () => {
  this.onPlayerReady(this.player);
[...]

PR #5902, opened 2019-03-28
middlwares clean in player dispose

iseekyouu:
## Description
cleans middlwares object in clearCacheForPlayer on player.dispose()


BrandonOCasey:
I am not sure that we can do this, as middlewares are currently shared between players.

gkatsev:
This is not how middleware currently work and will be a breaking change. Closing, but let's continue discussion in #5901.

PR #5909, opened 2019-04-02
fix: Restore muted on autoplay middleware terminate

BrandonOCasey:
## Description
Based on #5895 

1. Restores the old value of `muted` on a `play-terminated` event.
2. Adds more autoplay checks, especially around the new `play-terminated` behavior 
3. Simplifies `manualAutoplay_`, Basically we were trying to support, and not support promises at the same time, which did not make much sense because:
  a. `manualAutoplay_` will only work as expected if we have access to promises.
  b. `play()` always returns a promise, even if the tech doesn't so we can rely on a promise being returned.
  c. we should only bother with `muted` autoplay if `muted` is not already `true`.

I also noticed that there is a small chance for `muted` to be set to `true` after `manualAutoplay_` runs and then be set back to `false` once it fails. Do we want to try to handle that?

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1374,7 +1374,7 @@ class Player extends Component {
       promise = this.play();
     }
 
-    promise.then(() => {
+    return promise.then(() => {
       this.trigger({type: 'autoplay-success', autoplay: type});
     }).catch((e) => {
       this.trigger({type: 'autoplay-failure', autoplay: type});
@@ -3266,7 +3266,7 @@ class Player extends Component {
     const PromiseClass = this.options_.Promise || window.Promise;
 
     // if the value is a valid string and we have a promise class
-    // set it to that
+    // treat is a "manualAutopla".
     if (PromiseClass && typeof value === 'string' && (/(any|play|muted)/).test(value)) {
       this.options_.autoplay = value;
       this.manualAutoplay_(value);
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -418,7 +418,7 @@ class Player extends Component {
 
     // the attribute overrides the option
     if (tag.hasAttribute('autoplay')) {
-      this.options_.autoplay = true;
+      this.autoplay(true);
     } else {
       // otherwise use the setter to validate and
       // set the correct value.
@@ -1338,7 +1338,9 @@ class Player extends Component {
    * found on the autoplay getter at Player#autoplay()
    */
   manualAutoplay_(type) {
-    if (!this.tech_ || typeof type !== 'string') {
+    const PromiseClass = this.options_.Promise || window.Promise;
+
+    if (!this.tech_ || typeof type !== 'string' || !PromiseClass) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -3267,9 +3262,12 @@ class Player extends Component {
...
 
     let techAutoplay;
+    // manual autoplay is only supported where promises are
+    const PromiseClass = this.options_.Promise || window.Promise;
 
-    // if the value is a valid string set it to that
-    if (typeof value === 'string' && (/(any|play|muted)/).test(value)) {
+    // if the value is a valid string and we have a promise class
+    // set it to that
+    if (PromiseClass && typeof value === 'string' && (/(any|play|muted)/).test(value)) {
```
gkatsev: basically, only do fancy autoplay handling if Promise is available?

src/js/player.js
```diff
@@ -3267,9 +3262,12 @@ class Player extends Component {
     }
 
     let techAutoplay;
+    // manual autoplay is only supported where promises are
+    const PromiseClass = this.options_.Promise || window.Promise;
```
gkatsev: I would just call this `Promise`

src/js/player.js
```diff
@@ -3267,9 +3262,12 @@ class Player extends Component {
...
 
-    // if the value is a valid string set it to that
-    if (typeof value === 'string' && (/(any|play|muted)/).test(value)) {
+    // if the value is a valid string and we have a promise class
+    // set it to that
```
gkatsev: > set it to that

is now extra confusing with the part about the promise check

src/js/player.js
gkatsev: Does this mean that if you set `autoplay` to be any of these and are running on IE11, it'll stop working?

src/js/player.js
```diff
@@ -1347,39 +1349,32 @@ class Player extends Component {
...
     } else {
       promise = this.play();
     }
 
-    if (!promise || !promise.then || !promise.catch) {
-      return;
-    }
-
-    return promise.then(() => {
+    promise.then(() => {
```
gkatsev: is the return value from `manualAutoplay_` not used anywhere?

src/js/player.js
BrandonOCasey: Autoplay will just be set to a boolean in that case, which should still work on that browser anyway.

src/js/player.js
BrandonOCasey: I just copy pasted what we used in `play` I think there may have been `nodejs` require issues calling it just `Promise`

src/js/player.js
BrandonOCasey: It isn't used anywhere, but we should probably still return it

src/js/player.js
gkatsev: carry on then

src/js/player.js
gkatsev: We should think about whether IE11 is the only one that could be affected by this, because it could be a breaking change

PR #5912, opened 2019-04-04
 Add FullscreenApi pseudo-class for fullscreen then use in isFullscreen test

richardbushell:
Add FullscreenApi pseudo-class for fullscreen then use in isFullscreen test


richardbushell:
Oops! The second document [] wrapper of fsApi pseudo-class should not be there in the isFullscreen test, will update the PR later.

src/js/player.js
```diff
@@ -1979,7 +1979,7 @@ class Player extends Component {
   documentFullscreenChange_(e) {
     const fsApi = FullscreenApi;
 
-    this.isFullscreen(document[fsApi.fullscreenElement] === this.el());
+    this.isFullscreen(document[fsApi.fullscreenElement] === this.el() || this.el().matches(':' + fsApi.fullscreen));
```
gkatsev: I was going to comment about how on IE11 `matches` is `msMatchesSelector` but I think we'll never actually get here on IE11

gkatsev: Looks like I was wrong :(

PR #5914, opened 2019-04-04
fix: Always show the mute button by default in responsive mode.

misteroneill:
## Description
We received a report that muted autoplay players using the new responsive breakpoints could not be unmuted at small sizes. This was due to an oversight in the initial implementation of responsive mode.

## Requirements Checklist

### DIFF
--- a/src/css/components/_adaptive.scss
+++ b/src/css/components/_adaptive.scss
@@ -42,7 +42,8 @@
     .vjs-volume-panel.vjs-volume-panel-horizontal:hover,
     .vjs-volume-panel.vjs-volume-panel-horizontal:active,
     .vjs-volume-panel.vjs-volume-panel-horizontal.vjs-slider-active {
-      width: 4em;
+      width: auto;
+      width: initial;
     }
   }
 
--- a/src/css/components/_adaptive.scss
+++ b/src/css/components/_adaptive.scss
@@ -68,7 +68,3 @@
     }
   }
 }
-
-.video-js.vjs-layout-x-small:not(.vjs-fullscreen) {
-  .vjs-subs-caps-button, .vjs-audio-button { display: none; }
-}
--- a/src/css/components/_adaptive.scss
+++ b/src/css/components/_adaptive.scss
@@ -1,56 +1,74 @@
-// When the player is absurdly tiny, display nothing but:
-// - Play button
-// - Volume Mute button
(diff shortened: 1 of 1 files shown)


src/css/components/_adaptive.scss
```diff
@@ -1,44 +1,70 @@
...
-  .vjs-chapters-button, .vjs-descriptions-button, .vjs-captions-button,
-  .vjs-subtitles-button, .vjs-subs-caps-button, .vjs-audio-button { display: none; }
-}
+    // Reset the size of the volume panel to the default so we don't see a big
+    // empty space to the right of the mute button.
+    .vjs-volume-panel.vjs-volume-panel-horizontal:hover,
+    .vjs-volume-panel.vjs-volume-panel-horizontal:active,
+    .vjs-volume-panel.vjs-volume-panel-horizontal.vjs-slider-active {
+      width: 4em;
+    }
```
misteroneill: This is part of the functional change.

src/css/components/_adaptive.scss
```diff
@@ -1,44 +1,70 @@
...
+    .vjs-captions-button,
+    .vjs-subtitles-button,
+    .vjs-audio-button,
+    .vjs-volume-control {
+      display: none;
+    }
 
-  .vjs-current-time, .vjs-time-divider, .vjs-duration, .vjs-remaining-time,
-  .vjs-playback-rate, .vjs-progress-control,
-  .vjs-mute-control, .vjs-volume-control, .vjs-volume-panel,
```
misteroneill: Removed `.vjs-mute-control` and `.vjs-volume-panel` from the hidden elements lists.

src/css/components/_adaptive.scss
gkatsev: We should test to see if `initial` works in firefox, safari, ie11 here instead. This way, we won't specify the width of the control in yet another place. Seems to work for me in Chrome

misteroneill: I'll do that.

misteroneill: Does not work in IE11. Works everywhere else... :(

misteroneill:
This will need to be rebased.

JiangKunbo:
But I mute it during initialization, but there is still a voice coming out.

PR #5929, opened 2019-04-11
fix: Explicitly remove all document-level listeners on player dispose.

misteroneill:
We have gotten a report (at Brightcove) of an issue where a disposed player could potentially leave behind the document keydown listener for the new-ish hotkeys feature.

This led me to realize there could be other cases where document-level listeners are left behind. For example, if the player is somehow disposed while in fullscreen, it could leave behind those listeners.

Even though we haven't had reports of any other listeners being left, it seems responsible to make sure they're cleaned up!

## Specific Changes proposed
* Remove all document-level listeners on player dispose.

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -353,8 +353,8 @@ class Player extends Component {
 
     // Create bound methods for document listeners.
     this.boundDocumentFullscreenChange_ = Fn.bind(this, this.documentFullscreenChange_);
-    this.boundFullWindowOnEscKey = Fn.bind(this, this.fullWindowOnEscKey);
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -351,6 +351,11 @@ class Player extends Component {
     // Run base component initializing with new options
     super(null, options, ready);
 
+    // Create bound methods for document listeners.
+    this.boundDocumentFullscreenChange_ = Fn.bind(this, this.documentFullscreenChange_);
+    this.boundFullWindowOnEscKey = Fn.bind(this, this.fullWindowOnEscKey);
```
gkatsev: this should be suffixed with `_`

PR #5933, opened 2019-04-11
chore: Add a sandbox page for testing autoplay values.

brandonocasey:
## Description
Currently our `manualAutoplay` functionality fails quite often in IE 11, I added a test page to prove it and a fix.


brandonocasey:
I don't actually see this happening anymore but I think the autoplay test page could still be useful

PR #5952, opened 2019-04-23
Add role='application' to the video element on Chrome on Windows, to work around a JAWS screen reader bug (for v6.x)

OwenEdwards:
Same as #5863, for v6.x.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -32,7 +32,7 @@ import * as middleware from './tech/middleware.js';
 import {ALL as TRACK_TYPES} from './tracks/track-types';
 import filterSource from './utils/filter-source';
 import {findMimetype} from './utils/mimetypes';
-import {IE_VERSION} from './utils/browser';
+import {IE_VERSION, IS_CHROME, IS_WINDOWS} from './utils/browser.js';
 
 // The following imports are used only to ensure that the corresponding modules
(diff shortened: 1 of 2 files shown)


OwenEdwards:
I don't know what the Travis check is failing. @gkatsev?

gkatsev:
The 6.x branch is woefully out of date, there's definitely some fix we made in master that never made it to this branch.

OwenEdwards:
@gkatsev what do we do about that? I thought we were still supporting v6.x? Specifically, I know of an organization using 6.x which has run into this issue, and is hoping for a release that will fix it so that they don't have to fix it themselves. Is our guidance for them to move to 7.x?

gkatsev:
While the upgrade is probably the best idea, it's definitely time to update 6.x with all the features and fixes in 7.x that can/should go in. It's been on the back burner since so much other stuff have been getting in the way. We probably also want to start a sunset period for 6.x.

OwenEdwards:
Bump.

OwenEdwards:
Apparently this issue was no longer present with JAWS 2020 (see https://github.com/FreedomScientific/VFO-standards-support/issues/78#issuecomment-705162364), so any user who encounters it should either update their version of JAWS, or the site developers should update to video.js 7.x.

PR #5960, opened 2019-04-25
docs: emphasize src object and video-js element

mister-ben:
Favours using a source object as the argument to `src()` in the workflows doc, and adds `<video-js>` to the setup doc.

### DIFF
--- a/docs/guides/player-workflows.md
+++ b/docs/guides/player-workflows.md
@@ -100,7 +100,7 @@ Example
 ```js
 var myPlayer = videojs('some-player-id');
 
-myPlayer.src('http://www.example.com/path/to/video.mp4');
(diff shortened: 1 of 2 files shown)


docs/guides/player-workflows.md
```diff
@@ -306,14 +306,16 @@ var myPlayer = videojs('some-player-id');
 myPlayer.src('http://www.example.com/path/to/video.mp4');
 ```
 
+When a string is provided as the source, Video.js will try to infer the video type from the file extension, but this iferrence will not work in all cases. It is recommended that the source is provided as an object including the type, as below.
```
OwenEdwards: but this **inference** will not work

mister-ben:
Thanks, fixed

PR #5963, opened 2019-04-26
test(MenuButton): silence flaky test

gkatsev:
Since updating sinon, but perhaps a bit before as well, this test has been really flaky. Silence it until we can revisit the test and re-write it to not be flaky.


gkatsev:
Closing in favor of #5964

PR #5969, opened 2019-04-29
fix: make sure hotkeys are not triggered outside the player or in form fields within the player

misteroneill:
## Description
We saw some issues with the recently-added hotkeys feature.

1. The use of `document`-level `keydown` listeners caused players to respond to events they should not have when focus was not in the player.
1. Those `keydown` events which do reach the player are considered possible hotkey triggers. However, there are sometimes elements added to the player - such as in modals - that _should not_ trigger hotkeys.

## Specific Changes proposed
- Rename any `handleKeyPress` method to `handleKeyDown` because that's more accurate. Define `handleKeyPress` as a `Component` method that delegates to `handleKeyDown` so that any existing uses of it should not break.
- Do not bind/unbind listeners on `focus`/`blur` and stop listening for `keydown` events on the `document`. It's better to just listen for `keydown` events on the components themselves.
- Add exceptions for certain elements by their `tagName`.
- Call `stopPropagation()` in key places to prevent double `keydown` handling.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -47,8 +47,6 @@ class BigPlayButton extends Button {
     // exit early if clicked via the mouse
     if (this.mouseused_ && event.clientX && event.clientY) {
       silencePromise(playPromise);
-      // call handleFocus manually to get hotkeys working
-      this.player_.handleFocus({});
       return;
     }
 
@@ -69,10 +67,10 @@ class BigPlayButton extends Button {
     }
   }
 
-  handleKeyPress(event) {
+  handleKeyDown(event) {
     this.mouseused_ = false;
 
-    super.handleKeyPress(event);
+    super.handleKeyDown(event);
   }
 
   handleMouseDown(event) {
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -47,6 +47,7 @@ class BigPlayButton extends Button {
     // exit early if clicked via the mouse
     if (this.mouseused_ && event.clientX && event.clientY) {
       silencePromise(playPromise);
+      this.player_.tech(true).focus();
       return;
     }
 
@@ -54,7 +55,7 @@ class BigPlayButton extends Button {
     const playToggle = cb && cb.getChild('playToggle');
 
     if (!playToggle) {
-      this.player_.focus();
+      this.player_.tech(true).focus();
       return;
     }
 
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -3,16 +3,13 @@
  */
 import Component from './component';
 import * as Dom from './utils/dom.js';
-import * as Events from './utils/events.js';
-import * as Fn from './utils/fn.js';
 import log from './utils/log.js';
-import document from 'global/document';
 import {assign} from './utils/obj';
 import keycode from 'keycode';
 
 /**
- * Clickable Component which is clickable or keyboard actionable,
- * but is not a native HTML button.
+ * Component which is clickable or keyboard actionable, but is not a
+ * native HTML button.
  *
  * @extends Component
  */
@@ -36,7 +33,7 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Create the `Component`s DOM element.
+   * Create the `ClickableComponent`s DOM element.
    *
    * @param {string} [tag=div]
    *        The element's node type.
@@ -83,7 +80,7 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Create a control text element on this `Component`
+   * Create a control text element on this `ClickableComponent`
    *
    * @param {Element} [el]
    *        Parent element for the control text.
@@ -109,7 +106,7 @@ class ClickableComponent extends Component {
   }
 
   /**
-   * Get or set the localize text to use for the controls on the `Component`.
+   * Get or set the localize text to use for the controls on the `ClickableComponent`.
    *
    * @param {string} [text]
    *        Control text for element.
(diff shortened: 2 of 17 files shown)


src/js/big-play-button.js
```diff
@@ -47,8 +47,6 @@ class BigPlayButton extends Button {
     // exit early if clicked via the mouse
     if (this.mouseused_ && event.clientX && event.clientY) {
       silencePromise(playPromise);
-      // call handleFocus manually to get hotkeys working
-      this.player_.handleFocus({});
```
gkatsev: ah, this is why it worked initially, I guess

src/js/clickable-component.js
```diff
@@ -202,52 +186,30 @@ class ClickableComponent extends Component {
...
    */
-  handleKeyPress(event) {
-    // Support Space or Enter key operation to fire a click event
+  handleKeyDown(event) {
+
+    // Support Space or Enter key operation to fire a click event. Also,
+    // prevent the event from propagating through the DOM and triggering
+    // Player hotkeys.
     if (keycode.isEventKey(event, 'Space') || keycode.isEventKey(event, 'Enter')) {
-      event.preventDefault();
```
gkatsev: we want to call `preventDefault` as well. It will prevent stuff like going page down when hitting the spacebar.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -407,30 +407,30 @@ class SeekBar extends Slider {
    *
    * @listens keydown
    */
-  handleKeyPress(event) {
+  handleKeyDown(event) {
     if (keycode.isEventKey(event, 'Space') || keycode.isEventKey(event, 'Enter')) {
-      event.preventDefault();
```
gkatsev: we probably want both.

src/js/menu/menu.js
```diff
@@ -211,22 +211,17 @@ class Menu extends Component {
    *
    * @listens keydown
    */
-  handleKeyPress(event) {
+  handleKeyDown(event) {
+
     // Left and Down Arrows
     if (keycode.isEventKey(event, 'Left') || keycode.isEventKey(event, 'Down')) {
-      event.preventDefault();
```
gkatsev: both again

src/js/player.js
```diff
@@ -2873,7 +2860,12 @@ class Player extends Component {
       playPauseKey = keydownEvent => (keycode.isEventKey(keydownEvent, 'k') || keycode.isEventKey(keydownEvent, 'Space'))
     } = hotkeys;
 
-    if (fullscreenKey.call(this, event)) {
```
gkatsev: this changes the context that `fullscreenKey` is called

src/js/player.js
```diff
@@ -2836,19 +2804,38 @@ class Player extends Component {
...
+      'input',
+      'optgroup',
+      'option',
+      'select',
+      'textarea'
+    ];
 
-        this.handleHotkeys(event);
+    // Bail out if the user is focused on an interactive form element.
+    if (activeTag && hotkeysDisallowed.indexOf(activeTag) !== -1) {
```
gkatsev: I think this would prevent hotkeys from working in any of the control bar buttons, no?

src/js/slider/slider.js
```diff
@@ -353,7 +327,7 @@ class Slider extends Component {
    *        Event that caused this object to run
    */
   handleClick(event) {
-    event.stopImmediatePropagation();
```
gkatsev: stopImmediatePropagation calls stopPropagation, though, not sure why immediate is used here.

src/js/clickable-component.js
misteroneill: Good point. Will update.

src/js/player.js
misteroneill: I don't think so, I did try that, but I'll verify.

src/js/player.js
misteroneill: Oh, right, the arrow function is just the default. Will change that back.

src/js/slider/slider.js
misteroneill: Yeah, it seemed pointless since we turn off the capture phase everywhere.

src/js/player.js
misteroneill: You're right. I must have added this filtering before actually verifying it! D'oh!

gkatsev: I guess you can verify if the element is contained in the control bar and then skip this check

misteroneill: 🤔

misteroneill: Well, buttons are generally only activated on Space/Enter, which we pass through to the browser anyway. Other keys (e.g. m, f, k, etc) are unlikely to affect a button. Seems safe to allow buttons and blacklist them if it becomes an issue for whatever reason.

gkatsev: Potentially, though, you can also define buttons as an `input` tag. It may be worth doing the control bar check regardless of which elements we disallow.

misteroneill: True, true. The control bar check seems dirty, though. I'll ponder it.

misteroneill:
I realized that `userActions` does not appear to have tests. I will rectify that in this PR.

misteroneill:
Glad to hear it. It's fairly easy to reproduce if you use a page with a Video.js player embedded. If you hit the big play button then move focus into a text field or something, you can activate hotkeys when typing into the field.

misteroneill:
For some reason the fullscreen tests are failing in Safari. Will have to investigate tomorrow.

gkatsev:
Clicking on the BPB or the poster initially no longer enables hotkeys.

Issue #5962, opened 2019-04-26
evented mixin `one` when not targeting self only triggers one time for multiple events


brandonocasey:
The `evented` `one` function only triggers one time with multiple listener, when the mixin object is not targeting itself. 

Evented Mixin on self
```js
let triggered = 0;
let el = new Component();

const listener = () => triggered++;
el.one(['a', 'b'], listener);

el.trigger('a');
// triggered is 1
el.trigger('b');
// triggered is 2
```
Evented Mixin on other
```js
let triggered = 0;
let el = new Component();
let el2 = new Component();
el.one(el2, ['a', 'b'], listener);

el2.trigger('a');
// triggered is 1
el2.trigger('b');
// triggered is still 1
```

* The issue is is due to a [`this.off` call](https://github.com/videojs/video.js/blob/master/src/js/mixins/evented.js#L276) that removes all of the event listeners rather than just the one that triggered the listener.
* This issue wasn't caught because a [test is incorrect](https://github.com/videojs/video.js/blob/master/test/unit/mixins/evented.test.js#L249)

PR #5977, opened 2019-05-03
feat: any event function

brandonocasey:
There are a lot of times in Video.js code and plugin code where we want a function to trigger a single time for an array of events, then remove itself after the first event. I have implemented such a function via `any`.

Here are some examples:

Single event
```js
const player = videojs('some-player-id');

player.any('a', (e) => console.log(e.type + ' triggered');

player.trigger('a');
// logs 'a triggered'

player.trigger('a');
// logs nothing as the listener has been removed.
```

Multiple Events
```js
const player = videojs('some-player-id');

player.any(['a', 'b', 'c', 'd'], (e) => console.log(e.type + ' triggered');

player.trigger('d');
// logs 'd triggered'

player.trigger('a');
player.trigger('b');
player.trigger('c');
player.trigger('d');
// all triggers above log nothing as the listener is removed after the first 'd' trigger.
```

I also added some comments to the code about videojs/video.js#5962 

EDIT: updated from `race` to `any`

### DIFF
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -119,6 +119,16 @@ EventTarget.prototype.one = function(type, fn) {
   this.addEventListener = ael;
 };
 
+EventTarget.prototype.race = function(type, fn) {
+  // Remove the addEventListener alialing Events.on
+  // so we don't get into an infinite type loop
+  const ael = this.addEventListener;
+
+  this.addEventListener = () => {};
+  Events.race(this, type, fn);
+  this.addEventListener = ael;
+};
+
 /**
  * This function causes an event to happen. This will then cause any `event listeners`
  * that are waiting for that event, to get called. If there are no `event listeners`
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -119,13 +119,13 @@ EventTarget.prototype.one = function(type, fn) {
   this.addEventListener = ael;
 };
 
-EventTarget.prototype.race = function(type, fn) {
+EventTarget.prototype.any = function(type, fn) {
   // Remove the addEventListener alialing Events.on
   // so we don't get into an infinite type loop
   const ael = this.addEventListener;
 
   this.addEventListener = () => {};
-  Events.race(this, type, fn);
+  Events.any(this, type, fn);
   this.addEventListener = ael;
 };
 
--- a/src/js/event-target.js
+++ b/src/js/event-target.js
@@ -110,7 +110,7 @@ EventTarget.prototype.removeEventListener = EventTarget.prototype.off;
  *        The function to be called once for each event name.
  */
 EventTarget.prototype.one = function(type, fn) {
-  // Remove the addEventListener alialing Events.on
+  // Remove the addEventListener aliasing Events.on
   // so we don't get into an infinite type loop
   const ael = this.addEventListener;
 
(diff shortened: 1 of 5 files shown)


brandonocasey:
Also, if we get this code in we could have another pull request where we actually use it in places where it makes sense:

* https://github.com/videojs/video.js/blob/master/src/js/player.js#L1530-L1541
* https://github.com/videojs/video.js/blob/master/src/js/player.js#L2233-L2244
* https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track.js#L96-L108

thijstriemstra:
`race`... sounds like nascar, not sure about that name (or how to translate it into my native language).

mister-ben:
Nice, this will be useful. If the name's not clear, how about `oneOf()` since we've `on` and `one`?

thijstriemstra:
i wouldn't know the difference between `one` and `oneOf` unless i read the docs..

gkatsev:
`any` may work and already available in other libs like bluebird http://bluebirdjs.com/docs/api/promise.any.html

gkatsev:
Looks like Promise.any is coming to the language and it seems to fit our needs, so, I think we should use that term: https://github.com/tc39/proposal-promise-any

mister-ben:
Works for me, although would people expect `any` to return a promise? In any case, works great and I'll use it whatever we call it.

brandonocasey:
will rename to `any`

src/js/event-target.js
```diff
@@ -119,6 +119,16 @@ EventTarget.prototype.one = function(type, fn) {
   this.addEventListener = ael;
 };
 
+EventTarget.prototype.any = function(type, fn) {
+  // Remove the addEventListener alialing Events.on
```
OwenEdwards: "alialing"? typo?

gkatsev: Yup, apparently it's everywhere else in the file as well, 😆

src/js/mixins/evented.js
```diff
@@ -284,6 +288,51 @@ const EventedMixin = {
...
+    // Targeting another evented object.
+    } else {
+      const wrapper = (...largs) => {
+        this.off(target, type, wrapper);
+        listener.apply(null, largs);
+      };
+
+      // Use the same function ID as the listener so we can remove it later
+      // it using the ID of the original listener.
+      wrapper.guid = listener.guid;
```
gkatsev: should we have a test that verifies that the wrapper is removed? In particular if someone calls `off` before any of the events trigger?

gkatsev:
Tested and works well.

src/js/mixins/evented.js
gkatsev: I'll take a look at writing this test.

test/unit/events.test.js
```diff
@@ -347,3 +347,39 @@ QUnit.test('retrigger with an object should use the old element as target', func
   Events.off(el1, 'click');
   Events.off(el2, 'click');
 });
+
+QUnit.test('should listen only once for any', function(assert) {
```
gkatsev: huh, this test is failing locally in chrome for me.

gkatsev: oh, the `++` on line 354 got deleted somehow.

src/js/mixins/evented.js
gkatsev: Looking into this, I guess our check dom-data test covers this and a specific test is not necessary.

PR #5982, opened 2019-05-07
feat: add 'audio/mp4' mimetype for .m4a files

gjanblaszczyk:
## Description
I added missing mimetype for '.m4a' files.

## Specific Changes proposed
Extends MimetypesKind object by adding 'm4a' key with 'audio/mp4' value.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
Is m4a a common a file extension?

gjanblaszczyk:
Hi Gary, 
Well, I am not sure how common the file extension is but I'm using it.
Currently, the 'MimetypesKind' object contains other not very common file extensions for example .m4v. Maybe if we don't want to add m4a then we should also remove m4v. What do you think?

gkatsev:
Yeah, I guess since we have m4v, we may as well have m4a. No major objections to it, but we do want to keep that list reasonably lean.

PR #5992, opened 2019-05-13
Fix: changes 'mousedown' to the 'mouseup' event in the player

gjanblaszczyk:
## Description
All buttons in the player use 'mouseup' instead of 'mousedown' event.
This player behavior is very useful for accessibility because the user can cancel the action by clicking outside the button area. 
However, in the player, there is one place that the player uses a 'mousedown' instead of 'mouseup' event. I think that is just a mistake. 
You can check it by clicking on the video area during the playback. You will see that the playback pauses/resumes after 'mousedown' event but if you click on the play/pause button over the controlbar it uses 'mouseup' event. 
The WCAG 2.1 "2.5.2 Pointer Cancellation" spec. also suggests using "mouseup" instead of "mousedown" event. 

## Specific Changes proposed
I changed "mousedown' to 'mouseup" event inside the player.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/utils/dom.js
```diff
@@ -786,6 +786,10 @@ export function isSingleLeftClick(event) {
     return true;
   }
 
+  if (event.button === 0 && event.buttons === 0) {
```
gkatsev: What does this condition check for?

gjanblaszczyk: It checks if the left mouse button was pressed. According to the Mozzilla dev doc: 
```
event.button = 0 -> Main button pressed, usually the left button or the un-initialized state,
event.buttons  = 0 -> No button or un-initialized,
```
The problem: for 'mouseup' event we always get event.buttons equal to 0. (You can check here:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons)
which means that we foll into the next if statement 
`if (event.button !== 0 || event.buttons !== 1) `
which always return false for the left mouse up click. So I added one extra 'if statement' to fix it but maybe there is a better way...  Do you have any thoughts about it?

gkatsev: Seems reasonable, can you add a comment explaining we're account for mouseup and buttons being 0 in that case?

gjanblaszczyk: Yes, sure.

OwenEdwards:
Looks like you also need to change the comment/documentation at: https://github.com/videojs/video.js/blob/57dbf0e5d211e8301c7e44f2b0abd2d8d6143cc2/src/js/player.js#L1848

> @listens Tech#mouseup

gkatsev:
Thanks for helping improve the accessibility of Video.js @gjanblaszczyk!

gjanblaszczyk:
Thanks @gkatsev

Issue #5824, opened 2019-02-26
Support for picture-in-picture API


thijstriemstra:
## Description
Chrome 70+ introduces the picture-in-picture API. Would be cool to have an extra icon in the player for this, allowing you to switch to PIP.

Demo: https://googlechrome.github.io/samples/picture-in-picture/

API: https://developers.google.com/web/updates/2018/10/watch-video-using-picture-in-picture

thijstriemstra:
Simple button subclass that works:

```javascript
class PictureInPictureToggle extends Button {
    /**
     * This gets called when the button is clicked.
     *
     * @param {EventTarget~Event} event
     *        The `tap` or `click` event that caused this function to be
     *        called.
     *
     * @listens tap
     * @listens click
     */
    async handleClick(event) {
        let mediaElement = this.player_.el().firstChild();

        // disable button during picture-in-picture switch
        this.disable();

        // switch picture-in-picture mode
        try {
            if (mediaElement !== document.pictureInPictureElement) {
                // request picture-in-picture
                await mediaElement.requestPictureInPicture();
            } else {
                // exit picture-in-picture
                await document.exitPictureInPicture();
            }
        } catch (error) {
            // notify listeners
            this.player_.trigger('error', error);
        } finally {
            // switch completed
            this.enable();
        }
    }
}
```

thijstriemstra:
It's now supported in the videojs-record plugin, e.g.

![picture-in-picture](https://user-images.githubusercontent.com/305679/53538467-07536e80-3b0e-11e9-84ca-412e052b6d64.PNG)

And it's really cool to see your ` video` element flip to an external window like that.

I think PiP should be part of video.js.

Abdulmanan0315:
Yes this should be a good feature .

gkatsev:
Yeah, this is definitely something we'll look into.

samueleastdev:
Put together a really rough version to test just using the airplay icon works in Safari and Chrome not Firefox.
~~~
(function(vjs) {
    var pipPlugin = function(options) {

        var player = this;

        // Needed if the player is loaded multiple times before being disposed
        if (!player.el()) {
            return;
        }

        // VIDEO EL::
        var video = player.player().el().getElementsByTagName('video')[0];

        if (video.webkitSupportsPresentationMode && typeof video.webkitSetPresentationMode === "function" || ('pictureInPictureEnabled' in document)) {

            var pipToggle = videojs.getComponent("Button");
[...]

PR #6001, opened 2019-05-21
Add Picture-in-Picture API methods

beaufortfrancois:
## Description
Following https://github.com/videojs/video.js/issues/5824#issuecomment-494143138, this PR adds support for some Picture-in-Picture methods described in the [spec](https://wicg.github.io/picture-in-picture) and [article](https://developers.google.com/web/updates/2018/10/watch-video-using-picture-in-picture)

## Specific Changes proposed
This PR is about entering and leaving Picture-in-Picture in the player. It also makes sure that we can listen to the `enterpictureinpicture` and `leavepictureinpicture` events on the player.

## Requirements Checklist
- [x] Feature implemented / Bug fixed [#5824]
  - [ ] Docs/guides updated


gkatsev:
Can an "abstract"/empty method be added to `tech.js`, for example https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L767-L772?

beaufortfrancois:
> Can an "abstract"/empty method be added to `tech.js`, for example https://github.com/videojs/video.js/blob/master/src/js/tech/tech.js#L767-L772? It's needed because if another tech that doesn't implement this method yet ends up being used (for example videojs-flash), we don't want it to throw an error.

Done in https://github.com/videojs/video.js/pull/6001/commits/44d09fc7af88b4ba7cd7317a8a6bf16e362055e8

gkatsev:
We try to have two core contributors review PRs before they're merged but this is definitely on the way! Thanks again@

beaufortfrancois:
@gkatsev Hopefully it's gonna be OK then. Thanks for reviewing ;)

src/js/tech/tech.js
```diff
@@ -764,6 +764,22 @@ class Tech extends Component {
...
+   *
+   * @see [Spec]{@link https://wicg.github.io/picture-in-picture}
+   *
+   * @return {Promise}
+   *         A promise with a Picture-in-Picture window.
+   *
+   * @abstract
+   */
+  requestPictureInPicture() {
+    return Promise.reject();
```
gkatsev: I just realized for this, we should make sure that `Promise` is available or use `this.options_.Promise`
Like in the `play()` method: https://github.com/videojs/video.js/blob/ef49a9be7e413915961bb9ab6b20d9f5c92a907c/src/js/player.js#L2206-L2216
We support browsers that may not have Promises and we want to use the Promise polyfill that was passed to us if one is given but we also shouldn't blow up if promises aren't available. Probably should just return null in that case.

beaufortfrancois: @gkatsev `this.options_.Promise` doesn't exist in `tech.js`, it exists in `player.js`. Is that expected?

gkatsev: Oh, yeah, apparently it doesn't get passed in. I guess one way of fixing it is to return null and inside of the player, see whether we got back a promise or or not and if not, return a rejected promise.
But probably easier and makes more sense for longer term to pass the promise into the tech. It can be added to the techOptions inside of `loadTech_` in player.js: https://github.com/videojs/video.js/blob/ef49a9be7e413915961bb9ab6b20d9f5c92a907c/src/js/player.js#L1067-L1083 and then you'll be able to use it like outline above.

beaufortfrancois: Fixed in https://github.com/videojs/video.js/pull/6001/commits/efe840d93c384315d975c74043c9e70b14cac7c1
Thanks!

beaufortfrancois:
@gkatsev I think we're good now. WDYT?

src/js/tech/tech.js
```diff
@@ -765,19 +765,27 @@ class Tech extends Component {
...
+   *         otherwise.
    *
    * @abstract
    */
   requestPictureInPicture() {
-    return Promise.reject();
+    const PromiseClass = this.options_.Promise || window.Promise;
+
+    if (PromiseClass) {
+      return new PromiseClass((resolve, reject) => {
```
gkatsev: I think this can just be `return PromiseClass.reject()`, the polyfill is expected to have the static methods too

beaufortfrancois: Fixed in https://github.com/videojs/video.js/pull/6001/commits/4cbe4139d2a54c5182ea653b0e25185720b15cae

src/js/player.js
```diff
@@ -2801,6 +2859,89 @@ class Player extends Component {
...
+   * This gets called when a `Player` gains focus via a `focus` event.
+   * Turns on listening for `keydown` events. When they happen it
+   * calls `this.handleKeyPress`.
+   *
+   * @param {EventTarget~Event} event
+   *        The `focus` event that caused this function to be called.
+   *
+   * @listens focus
+   */
+  handleFocus(event) {
```
gjanblaszczyk: @gkatsev Why we need `handleFocus` here? I thought it should be removed... and `this.boundHandleKeyPress_` doesn't exist inside the player class.

beaufortfrancois: FYI. This change was not part of my proposed changes.

gkatsev: Hm... must've snuck in when I was fixing merge conflicts. I'll make sure that it's fixed in master. Nice find @gjanblaszczyk

gkatsev: PR to re-remove them: https://github.com/videojs/video.js/pull/6053

GeoTimber:
Jsut a quick question, how do I disable the picture in picture button from the player init options?

gkatsev:
@GeoTimber see the option here https://github.com/videojs/video.js/pull/6002#issuecomment-510620446

PR #6009, opened 2019-05-24
fix(fs): fix isFullscreen check for spec-api

### LINKED ISSUE #5814: Video EXIT fullscreen with ESC key giving wrong isFullscreen info
Hi

When exiting Video Fullscreen using the ESC key, I get an incorrect value back from the fullscreenchange listener. BUT it's fine when exiting with the Video Fullscreen Exit Button instead.

Here's my self-explanatory function used to test, with notes inline.

```
myPlayer.on('fullscreenchange', function () {
	if (myPlayer.isFullscreen()) {
		console.log("Video fullscreenchange: Video entered fullscreen");
		// WEIRD: ESC to EXIT when VIDEO FULLSCREEN triggers this "Entered" message again!
		// wrap this in a setTimeout function and it's OK
		// p.s. but it works fine anyway when using video fullscreen button exit instead of ESC key
	} else {
		console.log("Video fullscreenchange: Video exited fullscreen");
	}
});
```

Is this a known issue? Is there already a best-practice or solution to this as we were getting entirely inconsistent behaviour.

Please advise, thanks.

gkatsev:
For non-prefixed APIs, check directly against the fullscreen element
rather than using the cached value.

Fixes #5814.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2661,6 +2661,21 @@ class Player extends Component {
       this.toggleFullscreenClass_();
       return;
     }
+
+    if (prefixedFS) {
+      const fsApi = FullscreenApi;
+      const el = this.el();
+      let isFs = document[fsApi.fullscreenElement] === el;
+
+      if (!isFs && el.matches) {
+        isFs = el.matches(':' + fsApi.fullscreen);
+      } else if (!isFs && el.msMatchesSelector) {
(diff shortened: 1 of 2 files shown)


gkatsev:
Testable here https://deploy-preview-6009--videojs-docs.netlify.com/test-example/

Issue #5252, opened 2018-06-13
Subtitle Download Behavior


jaruba:
## Description
When adding subtitles with `addRemoteTextTrack({}, false)` it always downloads the subtitles immediately. I understand that this is because `videojs` wants to know if the subtitle type is supported and can be correctly parsed, but as I'm adding a lot of subtitles at once, this effectively freezes the page for some time (especially on phones) and puts an unnecessary strain on the server.

I would rather prefer the subtitles being downloaded when the user actually selects one, even if it leads to an error. Is there a plugin or a known way to accomplish this?

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Let a video start playback
2. Add a lot of subtitles with `addRemoteTextTrack({}, false)`
3. Check Network tab in the console

### Expected
I would expect it to not download a gazillion subtitles at once. None of them actually, as their not even needed yet and may never be needed, the user most likely needs 1 specific subtitle or even none.

### Actual
It downloads a gazillion subtitles at once.

### Error output
None

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
v7.0.3
#### browsers
All
#### OSes
All
### plugins
-

OwenEdwards:
This is really an issue with the WebVTT processing component (`vtt.js`), and was flagged by @gkatsev  in https://github.com/mozilla/vtt.js/issues/340. The downside of not loading all of the text tracks is that there would be a delay if the user chooses to switch captions or subtitles track during playback.

We're definitely looking at improvements to this, but it'll depend on changes in `videojs-vtt.js`, our fork of Mozilla's vtt.js.

gkatsev:
First reported in #1913.

Yeah, is waiting for captions to load upon selection OK? A hard question to answer.

OwenEdwards:
Ideally captions would be loaded and processed in the background, in a way that wouldn't block playback but would make sure that each caption cue is loaded and processed in time for them to be displayed. The best way to implement that is to mux them into an HLS, I would think, but I'm not entirely sure how video.js' VHS handles captions and subtitles in the HLS mux.

jaruba:
I rapidly went through the history of this, i'll read it all more carefully later on as this is not the proper time for it. I prefer waiting on load, yes, as compared to downloading 15-30 subtitles when you might not want a subtitle at all.

How is `vtt.js` blocking? Can't we just push that entire script in a web worker?

jaruba:
[...]

PR #6043, opened 2019-06-11
Load text tracks on demand

brion:
## Description

Reimplementation of https://github.com/videojs/video.js/pull/2192
on current code. Seems to work but has not been carefully tested,
especially on conditions such as slow networks or default tracks.
Would appreciate testing and any advice on ways this may go
wrong and need additional work.

For https://github.com/videojs/video.js/issues/5252

This delays loading of the VTT cue files until they are selected.
For sites like Wikipedia that tend to have large numbers of
crowdsourced subtitles and can show many files together on one
page, this saves a lot of unnecessary network transfer and API
hits.

## Specific Changes proposed

* does not load text tracks immediately unless they are default or non-subtitles
* does load them on demand

## Requirements Checklist
- [X] Feature implemented / Bug fixed (seems to work!)
  - [X] Change has been verified in an actual browser (Chome, Firefox, IE) (tested in Firefox)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -141,6 +141,8 @@ class Tech extends Component {
       this.emulateTextTracks();
     }
 
+    this.preloadTextTracks = options.preloadTextTracks !== false;
+
     this.autoRemoteTextTracks_ = new TRACK_TYPES.ALL.text.ListClass();
 
     this.initTrackListeners();
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -172,6 +172,8 @@ class TextTrack extends Track {
     this.cues_ = [];
     this.activeCues_ = [];
 
+    this.preload_ = this.tech_.preloadTextTracks !== false;
+
     const cues = new TextTrackCueList(this.cues_);
     const activeCues = new TextTrackCueList(this.activeCues_);
     let changed = false;
@@ -229,6 +231,10 @@ class TextTrack extends Track {
             return;
           }
           mode = newMode;
+          if (!this.preload_ && mode !== 'disabled' && this.cues.length === 0) {
+            // On-demand load.
+            loadTrack(this.src, this);
+          }
           if (mode !== 'disabled') {
             this.tech_.ready(() => {
               this.tech_.on('timeupdate', timeupdateHandler);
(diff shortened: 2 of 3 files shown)


brion:
I'll fix this up shortly...

brion:
New version of this patch retains previous default behavior, adding a `preloadTextTracks` option on techs set to true by default. It can now be disabled, leading to on-demand loading for the non-default tracks.

Should pass tests, but I'm not 100% sure this is the right way to set up such a config option. Comments & recommendations welcome. :)

brion:
Not sure why the tests are failing, but there's a lot of 404s reported? Not sure if something's wrong with the test runner or if the code's broken in a way I don't understand.

gkatsev:
Yeah, not sure what's up with the test. Restarted but I'll check it out tomorrow.

src/js/tech/tech.js
```diff
@@ -141,6 +141,8 @@ class Tech extends Component {
       this.emulateTextTracks();
     }
 
+    this.preloadTextTracks = (options.preloadTextTracks !== false);
```
gkatsev: This may not be necessary, we can just get the option directly in the text-track file.

src/js/tracks/text-track.js
```diff
@@ -172,6 +172,8 @@ class TextTrack extends Track {
     this.cues_ = [];
     this.activeCues_ = [];
 
+    this.preload_ = !!this.tech_.preloadTextTracks;
```
gkatsev: Looks like the issue is that this ends up defaulting to `false`, at least during the tests since `preloadTextTracks` ends up not being set for tests.
It might be better to just move the line from above here to make sure that we default to true.
```
this.preload_ = this.tech_.options_.preloadTextTracks !== false;
```
However, it would mean updating some tests to make sure that `options_` is present in the tests. Specifically, the tests towards the bottom of this file (https://github.com/videojs/video.js/blob/master/test/unit/tracks/text-track.test.js#L534) which use the `testTech` variable. Since it's created via `EventTarget`, it needs to just get a `testTech.options_ = {}` before passing it into the `TextTrack` constructor.

gkatsev: I guess leaving the line to set `this.preloadTextTracks` in tech.js is fine, we'll just need to set it with `!== false` here as well. This would mean the least amount of changes overall.

brion: @brion pinging myself a reminder to clean this up tonight/tomorrow.

src/js/tech/tech.js
brion: applying...

src/js/tracks/text-track.js
brion: Ok, applying the second recommendation (using a !== false in this bit and keeping the option on the tech)

brion:
> What should the behavior be when we have `default` attribute set? It looks like it'll preload all the tracks if any have `default` enabled. Also, it looks like it'll always preload the descriptions track, it probably shouldn't be preloaded as well.

It should only load the individual track labeled with `default` if I'm reading it correctly. The check for kinds was inherited from the previous patch I based this on, so might not be 100% correct. I think the idea is that metadata tracks should be passed through and loaded? Can adjust that bit if necessary.

PR #6054, opened 2019-06-17
refactor: switch to fullscreen.options

gkatsev:
This allows us to then extend the `fullscreen` option with `element` and `handler` as necessary.


gkatsev:
Updated to fix the memory leak. This actually would only be an issue with tests.

I kind of like the simplicity of `fullscreen.options`.

gkatsev:
Yeah, it's just how the test was created. Though, a bit surprised that we didn't notice it when the PR was made? Maybe the test was added after?

PR #6055, opened 2019-06-17
chore: switch to dart-sass

### LINKED ISSUE #5826: Replace node-sass
Node-sass is a pain to install for many.  Took me an half an hour of Google, downloading a bunch of MS Visual Studio/C++ stuff that I'll likely only use for node-sass, then re-reinstalling node-sass.  It was great tech 4 years ago, but now feels to be too much when other alternatives are available.

 #### [css-in-js ](https://css-tricks.com/bridging-the-gap-between-css-and-javascript-css-in-js/)
Worth looking into for new projects with some unicorn devs, [many options available](https://blog.bitsrc.io/9-css-in-js-libraries-you-should-know-in-2018-25afb4025b9b), but perhaps too much rewriting for a legacy project like this.

#### [Dart-Sass](https://sass-lang.com/dart-sass)
The reference SASS (RIP Ruby).  Can be used from CLI as binary, or [complied as JS](https://www.npmjs.com/package/dart-sass), which could be a bit slower than binary, but might be fast enough.

#### PostCSS
Seems switching to PostCSS [requires little extra coding](https://medium.com/@im.simonecorsi/moving-from-sass-to-postcss-why-what-and-how-f68b1bc760dc), considering you'll have to sweep though to remove baked-in prefixes & let [Autoprefixer](https://github.com/postcss/autoprefixer) handle that chore. [PostCSS SASS](https://github.com/jonathantneal/postcss-sass) will load Dart SASS if you want less friction, but Dart SASS might have a few differences with libsass.  Perhaps adding some [new CSS tech](https://css-tricks.com/bridging-the-gap-between-css-and-javascript-css-modules-postcss-and-the-future-of-css/#article-header-id-1) will benefit Video.JS also.  Popular enough & used in many large projects.

Most of the projects listed are recently updated.  I vote PostCSS, since it is fast, JS-only to avoid installing extra headaches, reduces the transition costs.

Thoughts?

gkatsev:
This is a rebased and updated PR of #5841.

We wanted to use the `sass` package as that's what the docs recommend. We also wanted to disable source maps that CDN-linked code won't try to download it.

Fixes #5841, fixes #5826.


gkatsev:
Yup, the memory leak is part of the fullscreenOptions PR. I have a fix incoming as part of #6054.

gkatsev:
Just pushed the mem-leak fix in #6054. I'll just rebase this PR once that's merged in.

PR #6057, opened 2019-06-18
feat: Add option to suppress initial error for non-playable sources

mister-ben:
## Description
Video.js checks whether sources are playable and both displays a message to the user and logs an error to the console. In Google's [mobile friendly test](https://search.google.com/test/mobile-friendly) and related tools, this message and error is triggered because their test browser's video element does not support any video formats. Some Video.js users are concerned about the æsthetics of the rendered preview within the tools and whether this might have an SEO impact.

## Specific Changes proposed
Adds a `suppressNotSupportedError` option, defaulting to `false`. If set to `true`, if no sources are playable the error is deferred to the first human interaction (click or touchstart) but cleared if a loadstart occurs.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/docs/guides/options.md
+++ b/docs/guides/options.md
@@ -35,6 +35,7 @@
   * [plugins](#plugins)
   * [responsive](#responsive)
   * [sources](#sources)
+  * [suppressNotSupportedMessage](#suppressNotSupportedMessage)
   * [techCanOverridePoster](#techcanoverrideposter)
   * [techOrder](#techorder)
   * [userActions](#useractions)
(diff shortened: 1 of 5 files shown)


docs/guides/options.md
```diff
@@ -374,6 +375,12 @@ Using `<source>` elements will have the same effect:
 </video>
 ```
 
+### `suppressNotSupportedError`
+
+> Type: `boolean`
+
+If set to true, then the no compatible source error will not be triggered immediately and instead will occur on the first user interaction. This is useful for Google's "mobile friendly" test tool, which can't play video but where you might not want to see an error displayed.
```
gkatsev: I wonder if there's a way to detect google bot and auto set this

mister-ben: They do document the user agents they use but I'm wary about doing that, in case adding special case handling for Googlebot user agents is detected and perceived as an attempt to fiddle SEO.

gkatsev: yeah, that make sense.

PR #6064, opened 2019-06-20
refactor(pip): rely only on WICG spec events

gkatsev:
The WICG spec calls out only two events, enterpictureinpicture and
leavepictureinpicture. We should try and only use those.
If pictureinpicturechange is still necessary, it can be re-added at a
later date.


beaufortfrancois:
LGTM

PR #6070, opened 2019-06-24
Silenced `play()` promise

Wicker25:
## Description
It seems like the method `Player.play()` tries to silence the tech promises only when the promises are not supported.

i.e. If we run the following code in Chrome 75:
```js
var player = window.player = videojs('video-element');
player.ready(function () {
  player.play();
  player.pause();
});
```
the browser will show the error message `Uncaught (in promise) DOMException: The play() request was interrupted`.

I don't think this is the expected behaviour because by default the private method `play_()` wraps the promises returned by the tech method with `silencePromise()`. (https://github.com/videojs/video.js/blob/master/src/js/player.js#L2274)

The method `Player.play()`, however, overrides that argument with the `resolve` of the new promise.

## Related issues
- https://github.com/videojs/video.js/issues/3474
- https://github.com/videojs/video.js/pull/3518

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2254,8 +2254,11 @@ class Player extends Component {
     const PromiseClass = this.options_.Promise || window.Promise;
 
     if (PromiseClass) {
-      return new PromiseClass((resolve) => {
-        this.play_(resolve);
+      return new PromiseClass(resolve => {
+        this.play_(promise => {
+          silencePromise(promise);
+          resolve();
+        });
       });
     }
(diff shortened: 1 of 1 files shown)


gkatsev:
We specifically do not want to silence the promise in all cases because then a Video.js user will not be able to know when and if the Promise was rejected since for them it'll always be resolved.
We should be only silencing the promise whenever it isn't being returned to the user as they can't handle it themselves at that point.

PR #6073, opened 2019-06-24
fix: add use of the safe version of getComputedStyle

TVS-Bruno:
## Description

Add use of the safe version of getComputedStyle in the currentDimension method of Component's class.
This will prevent a null exception when a video.js is implemented in a 'display:none' iframe on Firefox (<62).

This is a fix in continuation of the MR #3664 regarding a bug in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=548397

## Specific Changes proposed

- use the safe version of getComputedStyle already present in video.js
- small improvement of the safe version of the getComputedStyle

## Requirements Checklist
  - [ ] Docs/guides updated
  - [ ] Example created

### DIFF
--- a/src/js/utils/computed-style.js
+++ b/src/js/utils/computed-style.js
@@ -26,9 +26,9 @@ function computedStyle(el, prop) {
   }
 
   if (typeof window.getComputedStyle === 'function') {
-    const cs = window.getComputedStyle(el);
+    const computedStyleValue = window.getComputedStyle(el);
 
-    return cs ? cs[prop] : '';
+    return computedStyleValue ? computedStyleValue.getPropertyValue(prop) || computedStyleValue[prop] : '';
   }
 
   return '';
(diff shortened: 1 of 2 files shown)


src/js/utils/computed-style.js
```diff
@@ -26,9 +26,9 @@ function computedStyle(el, prop) {
   }
 
   if (typeof window.getComputedStyle === 'function') {
-    const cs = window.getComputedStyle(el);
+    const computedStyleValue = window.getComputedStyle(el);
```
brandonocasey: maybe this should just be called `style`? That is what they use on `mdn`

TVS-Bruno: Are you referring to this page?
https://developer.mozilla.org/fr/docs/Web/API/Window/getComputedStyle

They also use `cs` in the examples, but I'm not a fan of very short abbreviations or too much generic term where you have to go back to the declaration to understand exactly the variable.

Do you want me to change my commit to use the name `style`?

gkatsev: I think this is fine.

PR #6080, opened 2019-06-27
chore: use the new `any` event function

brandonocasey:
## Description
Use the new `any` event function where it makes sense.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1528,20 +1528,19 @@ class Player extends Component {
       // if the `sourceset` `src` was an empty string
       // wait for a `loadstart` to update the cache to `currentSrc`.
       // If a sourceset happens before a `loadstart`, we reset the state
-      // as this function will be called again.
       if (!event.src) {
-        const updateCache = (e) => {
-          if (e.type !== 'sourceset') {
-            const techSrc = this.techGet('currentSrc');
-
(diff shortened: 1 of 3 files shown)


src/js/player.js
```diff
@@ -1528,20 +1528,19 @@ class Player extends Component {
...
-        const updateCache = (e) => {
-          if (e.type !== 'sourceset') {
-            const techSrc = this.techGet('currentSrc');
-
-            this.lastSource_.tech = techSrc;
-            this.updateSourceCaches_(techSrc);
+        this.tech_.any(['sourceset', 'loadstart'], (e) => {
+          // if a sourceset happens before a `loadstart` there
+          // is nothing to do as this function will be called again.
+          if (e.type === 'sourceset') {
```
gkatsev: Using `.any`, I'm not sure that this function will get called again anymore.

brandonocasey: I think the comment needs to be updated to say that `handleTechSourceset` will be called again and reset the cache, rather than "this function"

gkatsev: Oh, I see, we called `off` unconditionally previously as well. It looked like we didn't.

brandonocasey: I updated the comment now.

gkatsev: Thanks!

PR #6098, opened 2019-07-15
lang: clone zh-CN to zh-Hans and zh-TW to zh-Hant

misteroneill:
## Description
Currently, `zh-CN` and `zh-TW` are used as shorthands for Simplified and Traditional Chinese character sets, respectively. However, they carry with them a built-in assumption about the geographic locale which may not be appropriate in all cases.

The codes `zh-Hans` and `zh-Hant` are more correct in that they refer only to the character set and not the geographic locale. Further specificity is available via codes like `zh-Hans-CN`.

## Specific Changes proposed
* Clone `zh-CN` to `zh-Hans` and `zh-TW` to `zh-Hant` during the build process.
* For the time being, do not check these copied files into source control. In v8.0, Video.js can switch to the new codes.

## Requirements Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ es5/*
 .idea/
 
 core.js
+
+# Ignore Chinese clones for now.
+lang/zh-Han*.json
--- a/package.json
+++ b/package.json
@@ -39,6 +39,8 @@
     "build:css:default": "sass --no-source-map src/css/vjs.scss dist/video-js.css",
     "postbuild:css:default": "postcss --verbose --config postcss.config.js -d dist/ dist/video-js.css",
     "build:lang": "npm-run-all build:lang:*",
+    "build:lang:chinese-s": "shx cp lang/zh-CN.json lang/zh-Hans.json",
+    "build:lang:chinese-t": "shx cp lang/zh-TW.json lang/zh-Hant.json",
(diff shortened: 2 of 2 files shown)


brandonocasey:
What about just symlinking `zh-CN.json` -> `zh-Hans.json` and `zh-TW.json` to `zh-Hant.json` instead of doing a copy?

misteroneill:
Not a bad idea. It should work. Only concern I could see is it might make building on Windows not work?

gkatsev:
Would publishing and copying the languages into dist/ still contain the files if they're symlinked?

brandonocasey:
After some testing symlinking appears to work.

misteroneill:
@gkatsev It copies the relative symlink into `dist/lang`

misteroneill:
`npm pack` does not work with the symlinks.

mister-ben:
> In v8.0, Video.js can switch to the new codes.

We might need to keep them. For the automatic langauge detection it's the codes browsers and sites actually use that will matter, assuming that's working well for Chinese now.

gkatsev:
@mister-ben that's true. We'll need to evaluate it at the time but would be nice if they can be removed.

brandonocasey:
I also wonder if we should make `videojs-languages` do this for us when it builts

PR #6099, opened 2019-07-15
feat: add isDisposed method to components

misteroneill:
## Description
Usually, when we need to check if a component (or player) has been disposed, we check if its element exists:

```js
if (player.el()) {
  // Do something...
}
```

It would be clearer if we had a method for that:

```js
if (!player.isDisposed()) {
  // Do something...
}
```

## Specific Changes proposed
* Add `isDisposed` method to the base component.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -131,10 +131,12 @@ class Component {
      * @type {EventTarget~Event}
      *
      * @property {boolean} [bubbles=false]
-     *           set to false so that the close event does not
+     *           set to false so that the dispose event does not
      *           bubble up
      */
     this.trigger({type: 'dispose', bubbles: false});
+
+    this.isDisposed_ = true;
 
     // Dispose all children.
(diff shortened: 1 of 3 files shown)


src/js/component.js
```diff
@@ -125,11 +125,13 @@ class Component {
      * @type {EventTarget~Event}
      *
      * @property {boolean} [bubbles=false]
-     *           set to false so that the close event does not
+     *           set to false so that the dispose event does not
      *           bubble up
      */
     this.trigger({type: 'dispose', bubbles: false});
 
+    this.isDisposed_ = true;
```
brandonocasey: should we check at the top of this function and return if `isDisposed_` is true`

gkatsev: That seems reasonable, triggering a dispose event if we're already disposed is going to cause issues.

PR #6103, opened 2019-07-18
perf: Use weakmap for dom data

brandonocasey:
## Description
Use a weakmap so that we don't have to do the listener removal dance anymore on dispose.


gkatsev:
Huh, apparently, WeakMap is supported in IE11. Though, it's only basic support, so, we'd want to double check that the operations we are using are supported in IE11.

src/js/component.js
```diff
@@ -152,7 +151,6 @@ class Component {
         this.el_.parentNode.removeChild(this.el_);
       }
 
-      DomData.removeData(this.el_);
```
brandonocasey: We don't have to do this, once `this.el_` becomes null the weakmap will remove it on its own.

gkatsev: I don't think we should be relying on the behavior of the garbage collector, when we're done using it we should empty it out.

src/js/utils/events.js
```diff
@@ -250,7 +253,11 @@ export function on(elem, type, fn) {
     return _handleMultipleEvents(on, elem, type, fn);
   }
 
-  const data = DomData.getData(elem);
+  if (!DomData.has(elem)) {
```
gkatsev: I see a couple usages of if !has, set, then get. Do you think it's worth a helper?

brandonocasey: I think that it might now make sense in this case since it is only used in two places.

gkatsev:
This is good to go but I think I'm going to hold off on this to be a minor change.

src/js/utils/events.js
gkatsev: Yeah, you're right. If we end up with this a few more times, then it'll make sense.

wind-stone:
Rollup did't compiler the WeapMap, as a result, there is an error "WeapMap is undefined" in Android 4.x

gkatsev:
I have a PR to fix this pending here https://github.com/videojs/video.js/pull/6289

gkatsev:
We merged #6289 and will be releasing it soon.

PR #6128, opened 2019-07-24
revert: "fix(play-toggle): call event.stopPropagation in the click ha…

### LINKED ISSUE #6092: [7.6.0] click event is not triggered
## Description
`click` event is not triggered for the small play button.
[Reduced test case](https://codepen.io/anon/pen/xoBwPG?editors=1111).

## Steps to reproduce
Press the big play button and you see a log in the console. But then pressing pause/play (the small button on the bottom left) doesn't log it any more. Changing the version to `7.5.6` fixes the issue and every click produces a log in the console.

### Expected
Expected multiple logs.

### Actual
A single log appears.

## Additional Information
[Here is more information](https://github.com/videojs/video.js/pull/5803#issuecomment-510389720) about the issue and the potential PR that broke it.

gkatsev:
…ndler (#5803)"

PR #5083 introduces a fix to #5624, an issue with click events when
Polymer's tap gesture is being used. However, this causes an issue where
`player.on('click')` no longer triggers from the play toggle. Thus, we
revert the change. In addition, looking at Polymer 2 and 3, they
recommend against using the tap gesture.

Fixes #6092

### DIFF
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -61,7 +61,6 @@ class PlayToggle extends Button {
     } else {
       this.player_.pause();
     }
-    event.stopPropagation();
   }
 
   /**


gkatsev:
This will re-open #5624 when merged. Potentially, that issue will be closed as a won't/can't fix as it creates other worse issues (#6092).

PR #6131, opened 2019-07-25
fix(pip): hide PiP button in browsers not support the WICG spec

gkatsev:
This came up in #5824. We do currently disable the button if PiP isn't available or isn't allowed, but we should just not show it if the APIs aren't available in the browser to keep the control bar cleaner.
When Safari support is added, this should be augmented to check for Safari PiP support.

### DIFF
--- a/src/js/control-bar/control-bar.js
+++ b/src/js/control-bar/control-bar.js
@@ -2,6 +2,7 @@
  * @file control-bar.js
  */
 import Component from '../component.js';
+import document from 'global/document';
 
 // Required children
 import './play-toggle.js';
@@ -68,10 +69,17 @@ ControlBar.prototype.options_ = {
     'descriptionsButton',
     'subsCapsButton',
     'audioTrackButton',
-    'pictureInPictureToggle',
     'fullscreenToggle'
   ]
 };
+
+if ('exitPictureInPicture' in document) {
+  ControlBar.prototype.options_.children.splice(
+    ControlBar.prototype.options_.children.length - 1,
+    0,
+    'pictureInPictureToggle'
+  );
+}
 
 Component.registerComponent('ControlBar', ControlBar);
 export default ControlBar;


src/js/control-bar/control-bar.js
```diff
@@ -68,10 +69,17 @@ ControlBar.prototype.options_ = {
...
     'subsCapsButton',
     'audioTrackButton',
-    'pictureInPictureToggle',
     'fullscreenToggle'
   ]
 };
 
+if ('exitPictureInPicture' in document) {
+  ControlBar.prototype.options_.children.splice(
+    ControlBar.prototype.options_.children.length - 1,
```
brandonocasey: maybe better to use indexOf here in case the order changes at some point, if you don't agree, then code lgtm

gkatsev: This ordering happens so rarely and this is easily noticeable/discoverable.

bariartz:
In line 16012 to 16109 replace with this code :
  // var PictureInPictureToggle =
  // /*#__PURE__*/
  // function (_Button) {
  //   inheritsLoose(PictureInPictureToggle, _Button);
  // 
  //   /**
  //    * Creates an instance of this class.
  //    *
  //    * @param {Player} player
  //    *        The `Player` that this class should be attached to.
  //    *
  //    * @param {Object} [options]
  //    *        The key/value store of player options.
  //    *
  //    * @listens Player#enterpictureinpicture
  //    * @listens Player#leavepictureinpicture
  //    */
  //   function PictureInPictureToggle(player, options) {
  //     var _this;
  // 
  //     _this = _Button.call(this, player, options) || this;
  // 
  //     _this.on(player, ['enterpictureinpicture', 'leavepictureinpicture'], _this.handlePictureInPictureChange); // TODO: Activate button on player loadedmetadata event.
  //     // TODO: Deactivate button on player emptied event.
  //     // TODO: Deactivate button if disablepictureinpicture attribute is present.
  // 
  // 
  //     if (!document.pictureInPictureEnabled) {
  //       _this.disable();
  //     }
  // 
  //     return _this;
  //   }
  //   /**
  //    * Builds the default DOM `className`.
  //    *
  //    * @return {string}
  //    *         The DOM `className` for this object.
  //    */
  // 
  // 
  //   var _proto = PictureInPictureToggle.prototype;
  // 
  //   _proto.buildCSSClass = function buildCSSClass() {
  //     return "vjs-picture-in-picture-control " + _Button.prototype.buildCSSClass.call(this);
  //   }
  //   /**
  //    * Handles enterpictureinpicture and leavepictureinpicture on the player and change control text accordingly.
  //    *
  //    * @param {EventTarget~Event} [event]
  //    *        The {@link Player#enterpictureinpicture} or {@link Player#leavepictureinpicture} event that caused this function to be
  //    *        called.
  //    *
  //    * @listens Player#enterpictureinpicture
  //    * @listens Player#leavepictureinpicture
  //    */
  //   ;
  // 
  //   _proto.handlePictureInPictureChange = function handlePictureInPictureChange(event) {
  //     if (this.player_.isInPictureInPicture()) {
  //       this.controlText('Exit Picture-in-Picture');
  //     } else {
  //       this.controlText('Picture-in-Picture');
  //     }
  //   }
  //   /**
  //    * This gets called when an `PictureInPictureToggle` is "clicked". See
  //    * {@link ClickableComponent} for more detailed information on what a click can be.
  //    *
  //    * @param {EventTarget~Event} [event]
  //    *        The `keydown`, `tap`, or `click` event that caused this function to be
  //    *        called.
  //    *
  //    * @listens tap
  //    * @listens click
  //    */
  //   ;
  // 
  //   _proto.handleClick = function handleClick(event) {
  //     if (!this.player_.isInPictureInPicture()) {
  //       this.player_.requestPictureInPicture();
  //     } else {
  //       this.player_.exitPictureInPicture();
  //     }
  //   };
  // 
  //   return PictureInPictureToggle;
  // }(Button);
  // /**
  //  * The text that should display over the `PictureInPictureToggle`s controls. Added for localization.
  //  *
  //  * @type {string}
  //  * @private
  //  */
  // 
  // 
  // PictureInPictureToggle.prototype.controlText_ = 'Picture-in-Picture';
  // Component.registerComponent('PictureInPictureToggle', PictureInPictureToggle);

ChanMenglin:
This is my code.
```javascript
controlBar: {
  children: { 
    playToggle: true,
    volumePanel: {
       inline: false,
    },
    volumeMenuButton: true,
    currentTimeDisplay: true,
    timeDivider: true,
    durationDisplay: true,
    remainingTimeDisplay: false,
    progressControl: true,
    subtitlesButton: true,
    pictureInPictureToggle: true,
    fullscreenToggle: true,
  },
},
```

I have some question could you help me?
1. pictureInPictureToggle show on Safari and Chrome on mac, and show on android chrome too, I just tested this, and it disabled on Safari and android chrome, I think pictureInPictureToggle should be hidden , What is wrong?

2. about `currentTimeDisplay`, `timeDivider`, `durationDisplay`, `remainingTimeDisplay`, I set these  like the code but I have to add some style(scss) to show or hide this controls, can I use javascript only？

```scss
.video-js /deep/ .vjs-time-control {
  display: block;
}
.video-js /deep/ .vjs-remaining-time {
  display: none;
}
```

PR #6140, opened 2019-07-29
feat: Allow click handler to be specified in clickable component's options

mister-ben:
## Description
Allow a clickable component's click handler to be passed as a constructor option as a simple alternative to creating a custom component.
Makes for easier answers to "how do I create a button questions".

## Specific Changes proposed
Add `clickHandler` as an option for ClickableComponent
All player controls override the `handleClick` method so are unaffected.

## Requirements Checklist
  - [x] Tests updated or fixed
  - [X] Docs/guides updated

### DIFF
--- a/test/unit/clickable-component.test.js
+++ b/test/unit/clickable-component.test.js
@@ -93,3 +93,21 @@ QUnit.test('handleClick should not be triggered more than once when enabled', fu
   testClickableComponent.dispose();
   player.dispose();
 });
+
+QUnit.test('handleClick should use handler from options', function(assert) {
+  let clicks = 0;
+
+  const player = TestHelpers.makePlayer({});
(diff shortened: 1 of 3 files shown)


test/unit/clickable-component.test.js
```diff
@@ -93,3 +93,21 @@ QUnit.test('handleClick should not be triggered more than once when enabled', fu
...
+
+  const player = TestHelpers.makePlayer({});
+  const testClickableComponent = new ClickableComponent(player, {
+    clickHandler() {
+      clicks++;
+    }
+  });
+  const el = testClickableComponent.el();
+
+  Events.trigger(el, 'click');
```
gkatsev: can we add a test that if we trigger a `space` or `enter` key on it, the click handler gets called?

PR #6141, opened 2019-07-29
perf: save ~10ms on `player.src` call

brandonocasey:
## Description
The regex that we currently use takes about 3ms per call, which happens for each source passed to `player.src`. Meaning 10 sources is a 30ms slowdown on player startup. Rather than trying to optimize it we should use `path.extname` the nodejs builtin in the browser.

This makes the operation so fast it does not even appear on the performance graph in chrome.

Filesize actually goes down 15 bytes gzipped too!

### DIFF
--- a/src/js/utils/filter-source.js
+++ b/src/js/utils/filter-source.js
@@ -56,10 +56,12 @@ const filterSource = function(src) {
  *        src Object with known type
  */
 function fixSource(src) {
-  const mimetype = getMimetype(src.src);
+  if (!src.type) {
+    const mimetype = getMimetype(src.src);
 
-  if (!src.type && mimetype) {
-    src.type = mimetype;
+    if (mimetype) {
+      src.type = mimetype;
+    }
   }
 
   return src;
--- a/src/js/utils/url.js
+++ b/src/js/utils/url.js
@@ -4,7 +4,6 @@
  */
 import document from 'global/document';
 import window from 'global/window';
-import {extname} from 'path';
 
 /**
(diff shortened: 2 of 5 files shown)


src/js/utils/filter-source.js
```diff
@@ -56,10 +56,12 @@ const filterSource = function(src) {
  *        src Object with known type
  */
 function fixSource(src) {
-  const mimetype = getMimetype(src.src);
+  if (!src.type) {
+    const mimetype = getMimetype(src.src);
```
brandonocasey: Even sources with a `type` were effected because we always got the mimetype for every source, even if we did not use it. This change makes it so we only get a type when we don't have one.

src/js/utils/url.js
```diff
@@ -131,16 +132,13 @@ export const getAbsoluteURL = function(url) {
...
-
-    if (pathParts) {
-      return pathParts.pop().toLowerCase();
-    }
+  if (!path || typeof path !== 'string') {
+    return '';
   }
 
-  return '';
+  return extname(parseUrl(path).pathname)
```
brandonocasey: I am unsure if parseUrl is necessary here, but I think that this mostly deals with urls being in the url utils.

brandonocasey:
> Filesize actually goes down 15 bytes gzipped too!

gkatsev:
Does this include only the extname piece? Also, nice that it doesn't change much.

brandonocasey:
Yeah it only includes the `extname` part of `path`

PR #6142, opened 2019-07-29
perf: throttle more timers and use native bind

brandonocasey:
## Description
This pull request contains various performance improvement broken up into separate commits.

### DIFF
--- a/src/js/control-bar/progress-control/play-progress-bar.js
+++ b/src/js/control-bar/progress-control/play-progress-bar.js
@@ -3,6 +3,7 @@
  */
 import Component from '../../component.js';
 import {IS_IOS, IS_ANDROID} from '../../utils/browser.js';
+import * as Fn from '../../utils/fn.js';
 
 import './time-tooltip';
 
@@ -13,6 +14,20 @@ import './time-tooltip';
  * @extends Component
  */
 class PlayProgressBar extends Component {
+
+  /**
+   * Creates an instance of this class.
+   *
+   * @param {Player} player
+   *        The {@link Player} that this class should be attached to.
+   *
+   * @param {Object} [options]
+   *        The key/value store of player options.
+   */
+  constructor(player, options) {
+    super(player, options);
+    this.update = Fn.throttle(Fn.bind(this, this.update), 25);
+  }
 
   /**
    * Create the the DOM element for this class.
--- a/src/js/control-bar/progress-control/play-progress-bar.js
+++ b/src/js/control-bar/progress-control/play-progress-bar.js
@@ -26,7 +26,7 @@ class PlayProgressBar extends Component {
    */
   constructor(player, options) {
     super(player, options);
-    this.update = Fn.throttle(Fn.bind(this, this.update), 25);
+    this.update = Fn.throttle(Fn.bind(this, this.update), Fn.UPDATE_REFRESH_INTERVAL);
   }
 
   /**
--- a/src/js/control-bar/progress-control/time-tooltip.js
+++ b/src/js/control-bar/progress-control/time-tooltip.js
@@ -24,7 +24,7 @@ class TimeTooltip extends Component {
    */
   constructor(player, options) {
     super(player, options);
-    this.update = Fn.throttle(Fn.bind(this, this.update), 25);
+    this.update = Fn.throttle(Fn.bind(this, this.update), Fn.UPDATE_REFRESH_INTERVAL);
   }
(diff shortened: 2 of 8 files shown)


src/js/control-bar/progress-control/play-progress-bar.js
```diff
@@ -14,6 +15,20 @@ import './time-tooltip';
...
+   *
+   * @param {Player} player
+   *        The {@link Player} that this class should be attached to.
+   *
+   * @param {Object} [options]
+   *        The key/value store of player options.
+   */
+  constructor(player, options) {
+    super(player, options);
+    this.update = Fn.throttle(Fn.bind(this, this.update), 25);
```
brandonocasey: This update function was being called much more often then needed

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -169,12 +198,6 @@ class SeekBar extends Slider {
    *          The current percent at a number from 0-1
    */
   update(event) {
-    // if the offsetParent is null, then this element is hidden, in which case
-    // we don't need to update it.
-    if (this.el().offsetParent === null) {
```
brandonocasey: This code never actually runs, since the control bar is always visible for screen readers.

src/js/control-bar/progress-control/time-tooltip.js
```diff
@@ -12,6 +13,20 @@ import formatTime from '../../utils/format-time.js';
...
+   *
+   * @param {Player} player
+   *        The {@link Player} that this class should be attached to.
+   *
+   * @param {Object} [options]
+   *        The key/value store of player options.
+   */
+  constructor(player, options) {
+    super(player, options);
+    this.update = Fn.throttle(Fn.bind(this, this.update), 25);
```
brandonocasey: This update function was being called much more often then needed

src/js/control-bar/progress-control/time-tooltip.js
```diff
@@ -88,7 +103,7 @@ class TimeTooltip extends Component {
   /**
    * Write the time to the tooltip DOM element.
    *
-   * @param {String} content
+   * @param {string} content
```
brandonocasey: I guess the linter fixed this one.

src/js/slider/slider.js
```diff
@@ -262,10 +263,10 @@ class Slider extends Component {
     const style = bar.el().style;
 
     // Set the new bar width or height
-    if (this.vertical()) {
-      style.height = percentage;
-    } else {
-      style.width = percentage;
+    const sizeKey = this.vertical() ? 'height' : 'width';
+
+    if (style[sizeKey] !== percentage) {
```
brandonocasey: Only update on change, prevents a style recalculate

src/js/utils/fn.js
```diff
@@ -32,9 +32,7 @@ export const bind = function(context, fn, uid) {
   }
 
   // Create the new function that changes the context
-  const bound = function() {
-    return fn.apply(context, arguments);
-  };
+  const bound = fn.bind(context);
```
brandonocasey: Use the native bind function

gkatsev:
Unfortunately, I don't think we can accept a large chunk of these changes as is. While a user is inactive, they could have still used their screen reader to focus on the seekbar or the time tooltip and would expect it to continue updating during playback. We probably want to do this for document visibility instead of user activity in the player.

src/js/utils/fn.js
gkatsev: ooh, this likely will be super helpful in debugging.

src/js/control-bar/progress-control/play-progress-bar.js
gkatsev: in the seek bar, we throttle to 30ms: https://github.com/videojs/video.js/blob/master/src/js/control-bar/progress-control/seek-bar.js#L25,  probably worth using the same thing everywhere. Maybe even have this value as a property exported from Fn utils?

src/js/utils/fn.js
gkatsev: Yup, it removes an extra stack frame. Goes from data dispatcher directly to the method instead of through this bound method.

src/js/control-bar/progress-control/play-progress-bar.js
brandonocasey: Yeah let's do that.

PR #6144, opened 2019-07-30
perf: Do not add/remove listeners for each timer

brandonocasey:
## Description
Saves about 1ms per `setTimeout`/`requestAnimationFrame`. Since we have several long running intervals that use these this is a fairly significant performance improvement.

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -98,6 +98,10 @@ class Component {
     this.childIndex_ = {};
     this.childNameIndex_ = {};
 
+    this.setTimeoutIds_ = [];
+    this.setIntervalIds_ = [];
+    this.rafIds_ = [];
+
     // Add any child components in options
     if (options.initChildren !== false) {
       this.initChildren();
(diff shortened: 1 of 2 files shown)


src/js/component.js
```diff
@@ -98,6 +98,10 @@ class Component {
     this.childIndex_ = {};
     this.childNameIndex_ = {};
 
+    this.setTimeoutIds_ = [];
```
gkatsev: Sounds like these would be better served as `Set`s?

brandonocasey:
Better description/example:
Every time we run `setTimeout` we lose 0.5ms of time because we have to set a dispose handler for that timer. When that same timer is run, or cleared we remove that listener losing another 0.5ms of time. Mostly this is bad for `requestAnimationFrame` which we call every 30ms on seek bar or every `timeupdate` on other components. This means seekBars `requestAnimationFrame` takes 1ms extra time to run, while the actual drawing of the update takes less time.

src/js/component.js
brandonocasey: Yeah that would make sense.

gkatsev:
So, the idea is that instead of setting up dispose timers each time, we'd just batch the disposals? Should clearTimers be a private method? Also, should we call it on dispose?

Do we have any tests to make sure that we don't trigger the timeouts/intervals/rAFs if we dispose the component between the time we register it and before it triggers?

brandonocasey:
I have to double check tests (but I think we have them) and it should probably be private.

brandonocasey:
also we shouldn't call it **in dispose**, since if a timer is ever set it will be called on `trigger('dispose')`

PR #6145, opened 2019-07-30
perf: Another 5ms of startup time improvements 

brandonocasey:
See code comments

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
     "@babel/node": "^7.4.5",
+    "@babel/plugin-transform-object-assign": "^7.2.0",
     "@babel/plugin-transform-runtime": "^7.4.4",
     "@babel/preset-env": "^7.4.5",
     "@babel/register": "^7.4.4",
--- a/src/js/error-display.js
+++ b/src/js/error-display.js
@@ -3,7 +3,6 @@
  */
 import Component from './component';
 import ModalDialog from './modal-dialog';
-import mergeOptions from './utils/merge-options';
 
 /**
  * A display that indicates an error has occurred. This means that the video
@@ -57,7 +56,7 @@ class ErrorDisplay extends ModalDialog {
  *
  * @private
  */
-ErrorDisplay.prototype.options_ = mergeOptions(ModalDialog.prototype.options_, {
+ErrorDisplay.prototype.options_ = Object.assign({}, ModalDialog.prototype.options_, {
   pauseOnOpen: false,
   fillAlways: true,
   temporary: false,
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -899,15 +899,15 @@ defineLazyProperty(Html5, 'TEST_VID', function() {
   if (!Dom.isReal()) {
     return;
   }
-  const el = document.createElement('video');
+  const video = document.createElement('video');
   const track = document.createElement('track');
 
   track.kind = 'captions';
   track.srclang = 'en';
   track.label = 'English';
-  el.appendChild(track);
+  video.appendChild(track);
 
-  return el;
+  return video;
 });
 
 /**
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -13,6 +13,7 @@ import mergeOptions from '../utils/merge-options.js';
 import {toTitleCase} from '../utils/string-cases.js';
 import {NORMAL as TRACK_TYPES} from '../tracks/track-types';
 import setupSourceset from './setup-sourceset';
+import defineLazyProperty from '../utils/define-lazy-property.js';
 
 /**
  * HTML5 Media Controller - Wrapper for HTML5 Media API
@@ -887,23 +888,27 @@ class Html5 extends Tech {
 
 /* HTML5 Support Testing ---------------------------------------------------- */
 
-if (Dom.isReal()) {
-
-  /**
-   * Element for testing browser HTML5 media capabilities
-   *
-   * @type {Element}
-   * @constant
-   * @private
-   */
-  Html5.TEST_VID = document.createElement('video');
+/**
+ * Element for testing browser HTML5 media capabilities
+ *
+ * @type {Element}
+ * @constant
+ * @private
+ */
+defineLazyProperty(Html5, 'TEST_VID', function() {
+  if (!Dom.isReal()) {
+    return;
+  }
+  const el = document.createElement('video');
   const track = document.createElement('track');
 
   track.kind = 'captions';
   track.srclang = 'en';
   track.label = 'English';
-  Html5.TEST_VID.appendChild(track);
-}
+  el.appendChild(track);
+
+  return el;
+});
 
 /**
  * Check if HTML5 media is supported by this browser/device.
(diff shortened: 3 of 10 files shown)


package.json
```diff
@@ -94,6 +94,7 @@
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
     "@babel/node": "^7.4.5",
+    "@babel/plugin-transform-object-assign": "^7.2.0",
```
brandonocasey: Babel already provides `_extends` as a helper, we may as well use it.

src/js/component.js
```diff
@@ -1562,13 +1562,17 @@ class Component {
...
-    if (!name) {
+    if (!name || !Component.components_) {
       return;
     }
 
+    if (Component.components_[name]) {
+      return Component.components_[name];
+    }
+
     name = toTitleCase(name);
```
brandonocasey: toTitleCase is one of the biggest offenders for startup time, since we `register` and `get` so many things. We should do anything that we can to prevent needing to call it.

src/js/error-display.js
```diff
@@ -57,7 +56,7 @@ class ErrorDisplay extends ModalDialog {
  *
  * @private
  */
-ErrorDisplay.prototype.options_ = mergeOptions(ModalDialog.prototype.options_, {
```
brandonocasey: Object.assign is much faster.

src/js/tracks/track-types.js
```diff
@@ -55,7 +53,7 @@ const REMOTE = {
   }
 };
 
-const ALL = mergeOptions(NORMAL, REMOTE);
+const ALL = Object.assign({}, NORMAL, REMOTE);
```
brandonocasey: Object.assign is much faster.

src/js/utils/dom.js
```diff
@@ -21,7 +21,7 @@ import computedStyle from './computed-style';
  *
  */
 function isNonBlankString(str) {
-  return typeof str === 'string' && (/\S/).test(str);
+  return typeof str === 'string' && str.trim();
```
brandonocasey: not using regex is faster.

src/js/utils/dom.js
```diff
@@ -36,7 +36,7 @@ function isNonBlankString(str) {
  *         Throws an error if there is whitespace in the string.
  */
 function throwIfWhitespace(str) {
-  if ((/\s/).test(str)) {
+  if (str.indexOf(' ') >= 0) {
```
brandonocasey: not using regex is faster.

src/js/utils/dom.js
```diff
@@ -160,7 +160,7 @@ export function createEl(tagName = 'div', properties = {}, attributes = {}, cont
     // method for it.
     } else if (propName === 'textContent') {
       textContent(el, val);
-    } else {
+    } else if (el[propName] !== val) {
```
brandonocasey: This function is also another big startup time user. We should make sure that the prop is not already equal to what we are going to set it to before doing it.

src/js/utils/events.js
```diff
@@ -84,6 +84,9 @@ function _handleMultipleEvents(fn, elem, types, callback) {
  *         Fixed event object.
  */
 export function fixEvent(event) {
+  if (event.fixed_) {
```
brandonocasey: This costs us 0.1 ms per event, but some event have it done to them 5+ times. Might as well only do it once.

src/js/utils/events.js
```diff
@@ -198,29 +201,35 @@ export function fixEvent(event) {
...
-(function() {
-  try {
-    const opts = Object.defineProperty({}, 'passive', {
-      get() {
-        _supportsPassive = true;
-      }
-    });
+let _supportsPassive;
+
+const supportsPassive = function() {
```
brandonocasey: For some reason this takes about 2ms to complete, making it lazy makes it take much less time later when an event that would need to be passive is actual used.

package.json
gkatsev: Seems reasonable, we should be careful with swapping mergeObject to Object.assign as Object.assign only does a shallow copy where-as mergeObject does a deep merge.

brandonocasey: Agreed I only copied default value usages that we had during startup, but we should see if there are more at some point.

src/js/tech/html5.js
```diff
@@ -1135,15 +1132,58 @@ Html5.prototype.featuresMuteControl = Html5.canMuteVolume();
...
+ */
+[
+  ['featuresVolumeControl', 'canControlVolume'],
+  ['featuresMuteControl', 'canMuteVolume'],
+  ['featuresPlaybackRate', 'canControlPlaybackRate'],
+  ['featuresSourceset', 'canOverrideAttributes'],
+  ['featuresNativeTextTracks', 'supportsNativeTextTracks'],
+  ['featuresNativeVideoTracks', 'supportsNativeVideoTracks'],
+  ['featuresNativeAudioTracks', 'supportsNativeAudioTracks']
+].forEach(function(array) {
```
brandonocasey: altogether these take about 1ms at startup.

src/js/component.js
```diff
@@ -1562,13 +1562,17 @@ class Component {
    *             return that if it exists.
    */
   static getComponent(name) {
-    if (!name) {
+    if (!name || !Component.components_) {
```
gkatsev: What cases happen to have `Component.components_` as null?

src/js/utils/dom.js
gkatsev: Can a comment be added here explaining usage of trim()?

gkatsev: Also, should we wrap `str.trim()` in a call to `Boolean`?

src/js/utils/dom.js
gkatsev: A comment explaining usage would helpful here as well.

src/js/utils/events.js
gkatsev: Do we always go into the line 101/104 block?

src/js/video.js
```diff
@@ -566,5 +567,7 @@ videojs.dom = Dom;
  */
 videojs.url = Url;
 
+videojs.defineLazyProperty = defineLazyProperty;
```
gkatsev: any reason to expose it currently?

src/js/tech/html5.js
```diff
@@ -888,23 +889,27 @@ class Html5 extends Tech {
...
+ *
+ * @type {Element}
+ * @constant
+ * @private
+ */
+defineLazyProperty(Html5, 'TEST_VID', function() {
+  if (!Dom.isReal()) {
+    return;
+  }
+  const el = document.createElement('video');
```
gkatsev: maybe call it `video` instead of `el`?

src/js/utils/dom.js
brandonocasey: yes

src/js/utils/events.js
brandonocasey: It seems like we almost always do yeah.

src/js/component.js
brandonocasey: `Component.components_` is only set after the first registration. So if we ever `getComponent` before `registerComponent` it will be null. This will always happen on the first `registerComponent` because `registerComponent` calls `const Tech = Component.getComponent('Tech');`

src/js/video.js
brandonocasey: I think we will need it in a few other places, vhs for sure.

PR #6147, opened 2019-07-31
chore: fix lang watch loop caused by zh-* copy

brandonocasey:
## Description
Currently running `npm run start` or `npm run watch` causes a loop since chokidar watches for all `.json` changes and during `build:lang` we copy to `.json` files. This causes chokidar to `build:lang` forever.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "minify:css:cdn": "cleancss dist/alt/video-js-cdn.css -o dist/alt/video-js-cdn.min.css",
     "minify:css:default": "cleancss dist/video-js.css -o dist/video-js.min.css",
     "watch": "npm-run-all -p watch:*",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "minify:css:cdn": "cleancss dist/alt/video-js-cdn.css -o dist/alt/video-js-cdn.min.css",
     "minify:css:default": "cleancss dist/video-js.css -o dist/video-js.min.css",
     "watch": "npm-run-all -p watch:*",
-    "watch:lang": "chokidar --initial 'lang/**/*.json' -c 'npm run build:lang'",
+    "watch:lang": "chokidar --initial 'lang/**/!(zh-Hans|zh-Hant)*.json' -c 'npm run build:lang'",
```
brandonocasey: Watch for all json changes except `zh-Hans` and `zh-Hant` which are just copies

PR #6148, opened 2019-07-31
perf: Improve performance of toTitleCase, register with lower and tit…

brandonocasey:
ToTileCase is one of our biggest startup functions, taking about 12ms to run. This pull request speeds that up by:

1. Making the `toTitleCase` function faster
2. Adding a `toLowerCase` function so that we can register title and lower case tech names
3. Since we register both title and lower case tech names, we can omit `toTitleCase` in most common scenarios

with all this done `toTitleCase` + `toLowerCase` take about `0.6ms` combined!


src/js/utils/to-lower-case.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+/**
+ * Lowercase the first letter of a string.
+ *
+ * @param {string} string
+ *        String to be uppercased
+ *
+ * @return {string}
+ *         The string with an uppercased first letter
```
gkatsev: this doesn't match the behavior of this method.

src/js/utils/to-lower-case.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @file to-lower-case.js
```
gkatsev: This could potentially be combined with the to-title-case file into a string-cases.js file or something.
We probably want a test for this as well, a la https://github.com/videojs/video.js/blob/master/test/unit/utils/to-title-case.test.js

src/js/component.js
```diff
@@ -358,8 +359,6 @@ class Component {
       return;
     }
 
-    name = toTitleCase(name);
```
gkatsev: so, basically, this is the big gain here? Because we register the component both as `component` and as `Component`, we can just assume it'll be available and we don't need to toTitleCase the name?

src/js/component.js
```diff
@@ -432,7 +431,8 @@ class Component {
     componentName = componentName || (component.name && toTitleCase(component.name()));
 
     if (componentName) {
-      this.childNameIndex_[componentName] = component;
+      this.childNameIndex_[toTitleCase(componentName)] = component;
```
gkatsev: Is this needed? The behavior related to this isn't changing. Same with line 484.

src/js/component.js
brandonocasey: exactly!

src/js/component.js
brandonocasey: We need this because, `getChild` used to `toTitleCase` the `name` that it was given. Now we won't do that because we register both possible names during `addChild`.

gkatsev: oh, didn't realize that  getChild used this property. However, it sounds like we don't need to call `toTitleCase` again because it sounds like `componentName` is always `toTitleCase`d per lines 430 and 388. So, we would only need to register the toLowerCase version.

brandonocasey: 👍

PR #6149, opened 2019-07-31
fix(lang): Add missing strings for Chinese (Simplified) and Chinese (Traditional)

misteroneill:
## Description
Add four missing strings for both Chinese variants we offer.

There are also script updates.

## Requirements Checklist


package.json
```diff
@@ -38,9 +38,11 @@
     "postbuild:css:cdn": "postcss --verbose --config postcss.config.js -d dist/alt dist/alt/video-js-cdn.css",
     "build:css:default": "sass --no-source-map src/css/vjs.scss dist/video-js.css",
     "postbuild:css:default": "postcss --verbose --config postcss.config.js -d dist/ dist/video-js.css",
+    "prebuild:lang": "npm-run-all -s prebuild:lang:*",
+    "prebuild:lang:zh-rm": "shx rm -f lang/zh-Han*.json",
```
brandonocasey: this should probably go in `clean`

package.json
```diff
@@ -38,9 +38,11 @@
...
+    "prebuild:lang:zh-s": "shx cp lang/zh-CN.json lang/zh-Hans.json",
```
brandonocasey: I think this can stay as `build:lang:chinese-s` and `build:lang:chinese-t`

package.json
```diff
@@ -181,7 +183,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "npm run prebuild:lang && git add lang/ && lint-staged"
```
brandonocasey: We should probably change `lint-staged` here rather than the husky hook. Should be a copy and past from https://github.com/videojs/video.js/pull/5649/files#r304555387

gkatsev:
FYI, we don't have a `lang:` prefix: https://github.com/videojs/conventional-changelog-videojs/blob/master/convention.md#type

package.json
gkatsev: Why is this necessary?

misteroneill: I tried but it didn't want to work. Will try again.

misteroneill: I guess it's not! I thought I was adding the clones to the repo.

misteroneill:
Which makes more sense:

1. Continue ignoring `zh-Hans.json` and `zh-Hant.json` and only using them as temporary build files.
2. Add them to the repository, requiring more scripting to make sure they stay in sync and are added on commit.

My vote is 1.

brandonocasey:
I prefer #1 as well

PR #6150, opened 2019-07-31
perf: Throttle ProgressControl ui updates, add document.hidden check to time-displays

brandonocasey:
So the activeElement mixin does the following:

1. Manages starting and stoping an update function when:
  a. The player is/is not paused
  b. The component gains or loses Keyboard/mouse focus or `player.userActive` changes
  c.  The tabs visibility (wether it is in the background or not) changes.
2. Does not start anything until the player has started playback.

### DIFF
--- a/src/js/control-bar/progress-control/load-progress-bar.js
+++ b/src/js/control-bar/progress-control/load-progress-bar.js
@@ -38,7 +38,6 @@ class LoadProgressBar extends Component {
     this.partEls_ = [];
     this.update = bind(this, this.update);
     activeElement(this, {
-      update: this.update,
       startUpdate: () => {
         this.on(player, 'progress', this.update);
         this.update();
--- a/src/js/control-bar/progress-control/load-progress-bar.js
+++ b/src/js/control-bar/progress-control/load-progress-bar.js
@@ -13,7 +13,8 @@ const percentify = function(time, end) {
 
   percent = (percent >= 1 ? 1 : percent) * 100;
 
-  return percent;
+  // round to two digits
+  return percent.toFixed(2) + '%';
 };
 
 /**
@@ -88,11 +89,11 @@ class LoadProgressBar extends Component {
 
       if (this.el_.style.width !== percent) {
         // update the width of the progress bar
-        this.el_.style.width = percent + '%';
+        this.el_.style.width = percent;
       }
 
       // update the control-text
-      Dom.textContent(controlTextPercentage, percent.toFixed(2) + '%');
+      Dom.textContent(controlTextPercentage, percent);
 
       // add child elements to represent the individual buffered time ranges
       for (let i = 0; i < buffered.length; i++) {
@@ -106,8 +107,8 @@ class LoadProgressBar extends Component {
         }
 
         // set the percent based on the width of the progress bar (bufferedEnd)
-        const leftPercent = percentify(start, bufferedEnd) + '%';
-        const widthPercent = percentify(end - start, bufferedEnd) + '%';
+        const leftPercent = percentify(start, bufferedEnd);
+        const widthPercent = percentify(end - start, bufferedEnd);
 
         if (part.style.left !== leftPercent) {
           part.style.left = leftPercent;
--- a/src/js/control-bar/progress-control/load-progress-bar.js
+++ b/src/js/control-bar/progress-control/load-progress-bar.js
@@ -2,7 +2,19 @@
  * @file load-progress-bar.js
  */
 import Component from '../../component.js';
+import {bind} from '../../utils/fn.js';
 import * as Dom from '../../utils/dom.js';
+import activeElement from '../../mixins/active-element.js';
+
+// get the percent width of a time compared to the total end
+const percentify = function(time, end) {
+  // no NaN
+  let percent = (time / end) || 0;
+
+  percent = (percent >= 1 ? 1 : percent) * 100;
+
+  return percent;
+};
 
 /**
  * Shows loading progress
@@ -23,7 +35,18 @@ class LoadProgressBar extends Component {
   constructor(player, options) {
     super(player, options);
     this.partEls_ = [];
-    this.on(player, 'progress', this.update);
+    this.update = bind(this, this.update);
+    activeElement(this, {
+      liveUpdates: false,
+      update: this.update,
+      startUpdate: () => {
+        this.on(player, 'progress', this.update);
+        this.update();
+      },
+      stopUpdate: () => {
+        this.off(player, 'progress', this.update);
+      }
+    });
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/js/player.js
```diff
@@ -536,6 +536,9 @@ class Player extends Component {
     this.one('play', this.listenForUserActivity_);
     this.on('stageclick', this.handleStageClick_);
     this.on('keydown', this.handleKeyDown);
+    if (this.options_.liveui && this.player_.liveTracker) {
```
brandonocasey: This was added so that its much easier to listen for `liveedgechanges`

src/js/mixins/visiblity.js
```diff
@@ -0,0 +1,28 @@
+import document from 'global/document';
+
+const visibilityMixin = function(object, options) {
```
brandonocasey: It's possible that this should have a better name, but I just wanted to get the idea out here. Perhaps `backgroundTabMixin` and `hide`/`show` as options.

src/js/control-bar/time-controls/time-display.js
```diff
@@ -77,12 +88,19 @@ class TimeDisplay extends Component {
...
 
     this.textNode_ = document.createTextNode(this.formattedTime_ || this.formatTime_(0));
-    this.contentEl_.appendChild(this.textNode_);
+
+    if (!this.textNode_) {
+      return;
+    }
+
+    if (oldNode) {
+      this.contentEl_.replaceChild(this.textNode_, oldNode);
```
brandonocasey: replaceChild appears to be about 10% faster than appendChild, and we call this function a lot

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -130,31 +68,33 @@ class SeekBar extends Slider {
...
+        this.localize(
+          'progress bar timing: currentTime={1} duration={2}',
+          [formatTime(currentTime, duration),
+            formatTime(duration, duration)],
+          '{1} of {2}'
+        )
+      );
+
+      // Update the `PlayProgressBar` when we are in mouse focus
+      if (this.bar && (!this.parentComponent_ || this.parentComponent_.mouseFocus_)) {
```
brandonocasey: This is only shown when the mouse is over `progressControl`

src/js/control-bar/progress-control/progress-control.js
```diff
@@ -4,6 +4,8 @@
 import Component from '../../component.js';
```
brandonocasey: We need to do an update in `progressControl` rather then `seekBar` because `progressControl` is the component that is tabbed to and moused over.

src/js/control-bar/progress-control/load-progress-bar.js
```diff
@@ -2,7 +2,19 @@
  * @file load-progress-bar.js
  */
 import Component from '../../component.js';
+import {bind} from '../../utils/fn.js';
 import * as Dom from '../../utils/dom.js';
+import activeElement from '../../mixins/active-element.js';
+
+// get the percent width of a time compared to the total end
+const percentify = function(time, end) {
```
brandonocasey: There isn't a good reason for this to define this every time the update function is run.

src/js/control-bar/progress-control/load-progress-bar.js
```diff
@@ -54,54 +77,53 @@ class LoadProgressBar extends Component {
...
-    // get the percent width of a time compared to the total end
-    const percentify = function(time, end, rounded) {
-      // no NaN
-      let percent = (time / end) || 0;
-
-      percent = (percent >= 1 ? 1 : percent) * 100;
-
-      if (rounded) {
-        percent = percent.toFixed(2);
+    this.requestAnimationFrame(() => {
```
brandonocasey: Without `requestAnimationFrame` this ran on the main thread, and it took a good chunk of time too.

src/js/control-bar/progress-control/load-progress-bar.js
brandonocasey: Note that we have to manually add `%` sign and round now, but this is much more generic.

src/js/control-bar/progress-control/load-progress-bar.js
brandonocasey: hiding whitespace changes makes this a lot easier to diff.

src/js/control-bar/progress-control/load-progress-bar.js
```diff
@@ -54,54 +77,53 @@ class LoadProgressBar extends Component {
...
+      const liveTracker = this.player_.liveTracker;
+      const buffered = this.player_.buffered();
+      const duration = (liveTracker && liveTracker.isLive()) ? liveTracker.seekableEnd() : this.player_.duration();
+      const bufferedEnd = this.player_.bufferedEnd();
+      const children = this.partEls_;
+      const controlTextPercentage = this.$('.vjs-control-text-loaded-percentage');
+      const percent = percentify(bufferedEnd, duration);
+
+      if (this.el_.style.width !== percent) {
```
brandonocasey: Only update the style if it changed

src/js/control-bar/progress-control/load-progress-bar.js
```diff
@@ -54,54 +78,53 @@ class LoadProgressBar extends Component {
...
-      let part = children[i];
+        // set the percent based on the width of the progress bar (bufferedEnd)
+        const leftPercent = percentify(start, bufferedEnd);
+        const widthPercent = percentify(end - start, bufferedEnd);
 
-      if (!part) {
-        part = this.el_.appendChild(Dom.createEl());
-        children[i] = part;
-      }
+        if (part.style.left !== leftPercent) {
```
brandonocasey: only update when changed.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -169,12 +109,6 @@ class SeekBar extends Slider {
    *          The current percent at a number from 0-1
    */
   update(event) {
-    // if the offsetParent is null, then this element is hidden, in which case
-    // we don't need to update it.
-    if (this.el().offsetParent === null) {
```
brandonocasey: This is not needed as this component is always visible for screen readers.

src/js/control-bar/time-controls/time-display.js
```diff
@@ -73,16 +85,25 @@ class TimeDisplay extends Component {
...
+      this.textNode_ = document.createTextNode(this.formattedTime_ || this.formatTime_(0));
 
-    this.textNode_ = document.createTextNode(this.formattedTime_ || this.formatTime_(0));
-    this.contentEl_.appendChild(this.textNode_);
+      if (!this.textNode_) {
+        return;
+      }
+
+      if (oldNode) {
+        this.contentEl_.replaceChild(this.textNode_, oldNode);
```
brandonocasey: replaceChild is about 10% faster than appendChild.

brandonocasey:
This just became too complicated, I broke out most of the logic into #6155. I will close this one and open another with the changes here.

PR #6153, opened 2019-08-01
perf: Cloning native events is no longer needed

brandonocasey:
As the title says, it appears that we no longer need to clone the event objects from the browser, we can just use the browser object. seems to work on all the browsers I tested with: IE 11, Safari 12/11, Android Chrome, Chrome, Firefox, Edge, iOS

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -74,6 +74,14 @@ function _handleMultipleEvents(fn, elem, types, callback) {
   });
 }
 
+function returnTrue() {
+  return true;
+}
+
+function returnFalse() {
+  return false;
+}
+
 /**
  * Fix a native event to have standard property values
  *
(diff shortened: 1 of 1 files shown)


gkatsev:
There are some event object properties that are read-only (like defaultPrevented) that we set. It's possible that we may not need to fix events anymore but more investigating would be required into what other changes should happen.

brandonocasey:
I agree, this change would need a lot of testing before we could bring it in

gkatsev:
Since we're making sure we're not going to fix events multiple times as part of https://github.com/videojs/video.js/pull/6145, should we close this out until we can investigate this more?

brandonocasey:
We probably still want to look into doing this as fixEvent still takes significant time for each event.

brandonocasey:
I am opening an issue for this #6175 as I don't think we will have the time to get to this, as its a bigger change that will need a lot of testing.

PR #6164, opened 2019-08-07
perf: Save 3740 bytes gizpped by getting rid of xhr deps

brandonocasey:
es-abstract is huge, and there isn't really a reason for us to include it. See the changes here:
EDIT: made an actual pull request. https://github.com/brandonocasey/xhr/pull/1

└─┬ xhr@2.5.0
  └─┬ parse-headers@2.0.2
    └─┬ string.prototype.trim@1.2.0
      └── es-abstract@1.13.0  deduped

We also are not using external helpers correctly, this pr fixes that as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,21 +80,22 @@
     "url": "https://github.com/videojs/video.js.git"
   },
   "dependencies": {
-    "@babel/runtime": "^7.4.5",
     "@videojs/http-streaming": "1.10.6",
+    "@babel/runtime": "^7.5.5",
     "global": "4.3.2",
     "keycode": "^2.2.0",
     "safe-json-parse": "4.0.0",
     "videojs-font": "3.2.0",
     "videojs-vtt.js": "^0.14.1",
-    "xhr": "2.4.0"
+    "xhr": "github:brandonocasey/xhr#perf/less-external"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
     "@babel/node": "^7.4.5",
     "@babel/plugin-transform-object-assign": "^7.2.0",
-    "@babel/plugin-transform-runtime": "^7.4.4",
+    "@babel/plugin-transform-runtime": "^7.5.5",
+    "@babel/plugin-external-helpers": "^7.2.0",
     "@babel/preset-env": "^7.4.5",
     "@babel/register": "^7.4.4",
     "access-sniff": "^3.2.0",
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "safe-json-parse": "4.0.0",
     "videojs-font": "3.2.0",
     "videojs-vtt.js": "^0.14.1",
-    "xhr": "github:brandonocasey/xhr#perf/less-external"
+    "@videojs/xhr": "2.5.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
@@ -95,7 +95,6 @@
     "@babel/node": "^7.4.5",
     "@babel/plugin-transform-object-assign": "^7.2.0",
     "@babel/plugin-transform-runtime": "^7.5.5",
-    "@babel/plugin-external-helpers": "^7.2.0",
     "@babel/preset-env": "^7.4.5",
     "@babel/register": "^7.4.4",
     "access-sniff": "^3.2.0",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -37,6 +37,8 @@ const primedCjs = commonjs({
   sourceMap: false
 });
 const primedBabel = babel({
+  externalHelpers: true,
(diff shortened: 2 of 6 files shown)


src/js/extend.js
```diff
@@ -3,38 +3,7 @@
...
-      configurable: true
-    }
-  });
-
-  if (superClass) {
-    // node
-    subClass.super_ = superClass;
-  }
-};
+import _inherits from '@babel/runtime/helpers/inherits';
```
brandonocasey: _inherits from babel is exactly the same as our code.

gkatsev:
I guess trim is used for IE8 support. I wonder if they have a timeline for xhr@3, which should drop obsolete browser support. Made some comments on your xhr commit.
We'd want it published on npm before we merge it in.

Also, we should make sure that regenerator and its runtime aren't included as part of the runtime and external helpers stuff.

package.json
```diff
@@ -80,21 +80,22 @@
...
     "videojs-vtt.js": "^0.14.1",
-    "xhr": "2.4.0"
+    "xhr": "github:brandonocasey/xhr#perf/less-external"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
     "@babel/node": "^7.4.5",
-    "@babel/plugin-transform-runtime": "^7.4.4",
+    "@babel/plugin-external-helpers": "^7.2.0",
```
gkatsev: Last time I played with it, using the external helpers actually made the file size much bigger, so, we should be careful with it.

src/js/extend.js
brandonocasey: Since babel includes the  `_inherits` from `@babel/runtime/helpers/inherits` in our code regardless, we may as well not include it twice

package.json
brandonocasey: I will do some file size tests without it, and just the xhr change.

rollup.config.js
```diff
@@ -71,28 +77,28 @@ const globals = {
...
-    'mpd-parser',
-    'mux.js',
-    'mux.js/lib/mp4',
-    'mux.js/lib/mp4/probe',
-    'mux.js/lib/tools/ts-inspector.js',
-    'mux.js/lib/tools/mp4-inspector',
-    'aes-decrypter',
-    'keycode'
-  ]),
+  module(id) {
```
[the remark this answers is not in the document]
brandonocasey: https://rollupjs.org/guide/en/#external (ie it can be a function or a list)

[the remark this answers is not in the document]
gkatsev: Ah, I was specifically looking for `module` rather than `external`.

rollup.config.js
```diff
@@ -71,28 +77,28 @@ const globals = {
...
+    const result = moduleExternals.some((ext) => id.indexOf(ext) !== -1);
```
gkatsev: would `id.startsWith(ext)` make sense here instead?

brandonocasey: yeah

brandonocasey: fyi I changed this because `@babel/runtime` includes any number of helpers and we don't want to have to list them all :)

brandonocasey:
--------------------------------------------------------------------
| File                      | w/o helpers | with helpers | master  |
|---------------------------|-------------|--------------|---------|
| `video.min.js`            | `140241`.   | `140241`     | `144533` |
| `video.novtt.min.js`.     | `133869`.   | `133877`     | `138317`|
| `video.core.min.js`.      | `57709`.    | `57736`      | `61914` |
| `video.core.novtt.min.js` | `51329`.    | `51356`       | `55786` |


It seems like with helpers is fine especially when we take into account that `@videojs/http-streaming`, `mpd-parser`, `m3u8-parser`, `pkc7s`, `mux.js`, and `aes-decryptr` all include there own helpers right now because the `es` module that they export does not use `@babel/runtime` since they don't use `external/runtime helpers`. Once we have them doing that with helpers should be much smaller.

gkatsev:
Cool, those differences are much smaller than last time I tried them out, and we still end up with a smaller output. Yeah, updating those to have better es builds would give us even more savings.

brandonocasey:
Just another note here now that I have done the leg work getting vhs to use `@babel/runtime` in https://github.com/videojs/http-streaming/pull/637 and https://github.com/videojs/http-streaming/pull/626
With just 626 we go down to `134197` bytes in `video.min.js` and with 637 added on to that we got down to `129771` 😮

brandonocasey:
Final size versus master (which just saw a size decrease from a video.js change): `131924` to `128184` for `video.min.js` or `3740` gzipped bytes and it should be about the same for every dist file.

src/js/extend.js
gkatsev: This change broke a plugin https://github.com/videojs/video.js/issues/6328.

PR #6165, opened 2019-08-07
test: silence test logs

brandonocasey:
Various things are logging "errors" which are just errors that we expect from tests, silence them.

### DIFF
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -337,6 +337,7 @@ QUnit.test('should asynchronously fire error events during source selection', fu
 });
 
 QUnit.test('should suppress source error messages', function(assert) {
+  sinon.stub(log, 'error');
   const clock = sinon.useFakeTimers();
 
   const player = TestHelpers.makePlayer({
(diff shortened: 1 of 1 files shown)


test/unit/player.test.js
```diff
@@ -363,9 +364,13 @@ QUnit.test('should suppress source error messages', function(assert) {
   assert.strictEqual(errors, 1, 'error after click');
 
   player.dispose();
+
+  assert.strictEqual(log.error.callCount, 2, 'two stubbed errors');
```
brandonocasey: `foo` is not a tech so of course we throw errors. That is what the test is testing anyway.

test/unit/player.test.js
```diff
@@ -363,9 +364,13 @@ QUnit.test('should suppress source error messages', function(assert) {
...
+  log.error.restore();
 });
 
 QUnit.test('should cancel a suppressed error message on loadstart', function(assert) {
+  sinon.stub(log, 'error');
```
brandonocasey: `foo` is not a tech so of course we throw errors. That is what the test is testing anyway.

test/unit/player.test.js
```diff
@@ -1572,6 +1580,8 @@ QUnit.test('player#reset removes remote text tracks', function(assert) {
   assert.strictEqual(player.remoteTextTracks().length, 1, 'there is one RTT');
   player.reset();
   assert.strictEqual(player.remoteTextTracks().length, 0, 'there are zero RTTs');
+  assert.strictEqual(log.warn.callCount, 1, 'one warning about for manualCleanup');
```
brandonocasey: saw this manualCleanup warning, might as well silence it too.

Issue #6166, opened 2019-08-08
Large bundle size (900KB)


ddrjm:
Hello, I'm using videojs via NPM and webpack which is incorporating the CSS files needed for the player and a few kb (3kb) of custom CSS that tweaks color here and there.

When building for production, I notice that the bundle has about 900kb. How can I reduce this size?
Is there a way to only import the bare minimum from videojs to have it working? (both modules and css)

Sorry if this isn't the right place to ask about this subject

sampathkumar7:
const CompressionPlugin = require('compression-webpack-plugin');

plugins:[
	new CompressionPlugin(),	
  ],

Use above setup it will create gzip file something like bundle.js.gz i used it to reduce 600KB file to 60KB
Also use can use google brotli compression plugin which is much better than Gzip

ddrjm:
> const CompressionPlugin = require('compression-webpack-plugin');
> 
> plugins:[
> new CompressionPlugin(),
> ],
> 
> Use above setup it will create gzip file something like bundle.js.gz i used it to reduce 600KB file to 60KB
> Also use can use google brotli compression plugin which is much better than Gzip

Sorry for the dumb question, but if I serve the .js.gz files, won't they be downloaded by the browser instead of being served? 

A 10x reduction sounds interesting but I'm confused about the .js.gz script files

sampathkumar7:
You need to enable in server to use gzip files.

brandonocasey:
You can also use `dist/alt/videojs.novtt.js` if you don't need vtt support baked in, `dist/alt/video.core.js` if you don't need `HLS/dash` support, or `dist/alt/videojs.core.novtt.js` if you don't need either of those.

ddrjm:
@brandonocasey in this case how would I include it via npm? 
At the moment I'm just doing `import videojs from video.js`. Should I change it to `import videojs.core.novtt.js` ?

brandonocasey:
### Solution for now is:
It should be `import videojs from 'video.js/dist/alt/videojs.core.novtt.js;` or if you don't do minification to your own bundle and you want the smallest video.js: `import videojs from 'video.js/dist/alt/video.core.novtt.js';` 

### Alternative solution (not working now due to a build issue on our end #6176) 
[...]

PR #6176, opened 2019-08-12
fix: make vhs external rather than its dependencies

brandonocasey:
## Description
Noticed this in #6166, we don't have a good way for users to exclude vhs from there own builds because we embed it in our `cjs` and `es` files, and only external its dependencies. We should instead make vhs external so that it can be done.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -82,11 +82,7 @@ const moduleExternals = [
   'tsml',
   'safe-json-parse',
   'videojs-vtt.js',
-  'url-toolkit',
-  'm3u8-parser',
-  'mpd-parser',
-  'mux.js',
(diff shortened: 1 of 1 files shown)


gkatsev:
I'll have to spend some time trying to remember why we inlined it for the es/cjs files.

forbesjo:
We should check to make sure video.js isn't doubled. I believe by including VHS the video.js alias is applied. If it's external it will probably use it's own version of video.js and therefore not be registered to the user's "main" video.js.

ckybonist:
This PR is LGTM for smaller bundle size unless 1 approval, Hope we can embrace it on next stable release : )

asahasrabuddhe:
Can we merge this?

PR #6192, opened 2019-08-22
fix: cap log history at 1000 items

brandonocasey:
## Description
Cap log history at 1000 items to prevent running out of memory on long videos.

### DIFF
--- a/src/js/utils/create-logger.js
+++ b/src/js/utils/create-logger.js
@@ -35,7 +35,9 @@ const LogByTypeFactory = (name, log) => (type, level, args) => {
     history.push([].concat(args));
 
     // only store 1000 history entries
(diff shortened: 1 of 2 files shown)


src/js/utils/create-logger.js
```diff
@@ -33,6 +33,11 @@ const LogByTypeFactory = (name, log) => (type, level, args) => {
   // Add a clone of the args at this point to history.
   if (history) {
     history.push([].concat(args));
+
+    // remove old history
+    while (history.length > 1000) {
```
gkatsev: would something like `history = history.slice(-1000)` work? Theoretically, should be faster as well, since we won't be shifting the entire list n times.

brandonocasey: makes sense to me

PR #6199, opened 2019-08-26
fix: validation for stalled and suspended event in iOS when headphones disconnect

marcodeltorob:
Description of the problem: 
When playback is ongoing in iOS + Safari and the user disconnect headphones (wired or Bluetooth) the player will throw 'suspend' or 'stalled' after a couple of seconds and 'progress' events. 

Solution: Validation for stalled and suspended event for iOS browser when headphones are disconnected

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -215,6 +215,7 @@ class Html5 extends Tech {
   handleIOSHeadphonesDisconnection_() {
     // Fudge factor to account for TimeRanges rounding
     const TIME_FUDGE_FACTOR = 1 / 30;
+
     // Comparisons between time values such as current time and the end of the buffered range
     // can be misleading because of precision differences or when the current media has poorly
     // aligned audio and video, which can cause values to be slightly off from what you would
@@ -225,9 +226,16 @@ class Html5 extends Tech {
     // If iOS check if we have a real stalled or supend event or
     // we got stalled/suspend due headphones where disconnected during playback
     this.on(['stalled', 'suspend'], (e) => {
-      if ((!this.el_.paused && window.navigator.onLine) &&
-          (this.el_.buffered.end(this.el_.buffered.length - 1) + SAFE_TIME_DELTA >= this.el_.currentTime)) {
-        this.el_.pause();
+      const buffered = this.buffered();
+
+      if (buffered.length > 0) {
+        const end = buffered.end(buffered.length - 1);
+
+        // if tech is not paused, browser has internet connection & end + safe delta >= current time
+        if ((!this.paused() && window.navigator.onLine) &&
+            (end + SAFE_TIME_DELTA >= this.currentTime())) {
+          this.pause();
+        }
       }
     });
   }
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -233,8 +233,9 @@ class Html5 extends Tech {
 
         // Establish if we have an extra buffer in the current time range playing.
         for (let i = 0; i < buffered.length; i++) {
-          if (buffered.start(i) <= this.currentTime() <= buffered.end(i) + SAFE_TIME_DELTA) {
+          if (buffered.start(i) <= this.currentTime() < buffered.end(i) + SAFE_TIME_DELTA) {
             extraBuffer = true;
+            break;
           }
         }
 
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -230,17 +230,19 @@ class Html5 extends Tech {
 
       if (buffered.length > 0) {
         let extraBuffer = false;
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -203,6 +207,31 @@ class Html5 extends Tech {
...
+  */
+  handleIOSHeadphonesDisconnection_() {
+
+    const TIME_FUDGE_FACTOR = 1 / 30;
+
+    const SAFE_TIME_DELTA = TIME_FUDGE_FACTOR * 3;
+
+    const disconnectionIOSListener = () => {
+      if ((!this.el_.paused && window.navigator.onLine) &&
+          (this.el_.buffered.end(this.el_.buffered.length - 1) + SAFE_TIME_DELTA >= this.el_.currentTime)) {
```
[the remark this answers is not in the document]
gkatsev: Also, it's possible that there are multiple buffered regions, and we're current in the first or a middle buffered region instead of the last one.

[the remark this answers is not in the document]
misteroneill: All good points!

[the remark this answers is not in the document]
marcodeltorob: @gkatsev Can you explain a little more, since we try to get the end time buffered. (length - 1) and if we have multiple regions it does not matter that means that we still have buffered time and we are not experiencing a `Stalled` or `suspend` due another problem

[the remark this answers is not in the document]
gkatsev: @marcodeltorob basically, we could be at the end of a buffered region which isn't the last buffered region available, so, currentTime will end up being less than the buffered region.
For example, we're at 10 seconds in (marked by `|`, and our buffer looks something like:
```
[0-10|----20-30----60-70----]
```

Reading the last buffered region, would give us 70 but our currentTime is 10. However, we still can't play because we've run out of buffer.
But, if the currentTime is say 5, and our buffered region has extra buffer, then we should be fine.
```
[0-5|5-10----20-30----60-70----]
```

src/js/tech/html5.js
```diff
@@ -203,6 +207,45 @@ class Html5 extends Tech {
...
+    // we got stalled/suspend due headphones where disconnected during playback
+    this.on(['stalled', 'suspend'], (e) => {
+      const buffered = this.buffered();
+
+      if (buffered.length > 0) {
+        let extraBuffer = false;
+
+        // Establish if we have an extra buffer in the current time range playing.
+        for (let i = 0; i < buffered.length; i++) {
+          if (buffered.start(i) <= this.currentTime() <= buffered.end(i) + SAFE_TIME_DELTA) {
```
misteroneill: 1. I think what we want is to check if there is anything in the forward buffer. It seems that if the `currentTime` is equal to the `buffered.end`, then we don't have anything in the forward buffer. Maybe just `buffered.start <= currentTime < buffered.end`?
1. Should we adjust the `buffered.start` using `SAFE_TIME_DELTA` as well? Maybe not.
1. We might as well `break` out of the loop when we set `extraBuffer = true`.

src/js/tech/html5.js
```diff
@@ -203,6 +207,46 @@ class Html5 extends Tech {
...
+        // Establish if we have an extra buffer in the current time range playing.
+        for (let i = 0; i < buffered.length; i++) {
+          if (buffered.start(i) <= this.currentTime() < buffered.end(i) + SAFE_TIME_DELTA) {
+            extraBuffer = true;
+            break;
+          }
+        }
+
+        // if tech is not paused, browser has internet connection & player has extraBuffer inside the timeRange
+        if ((!this.paused() && window.navigator.onLine) && extraBuffer) {
```
gkatsev: I don't think the extra parens are necessary as the expression will just evaluate each item in turn.
```suggestion
        if (!this.paused() && window.navigator.onLine && extraBuffer) {
```

gkatsev: Also, should we check extraBuffer first?

marcodeltorob: Make sense and we can decide quicker the state of the playback.

src/js/tech/html5.js
```diff
@@ -203,6 +207,48 @@ class Html5 extends Tech {
...
+    // expect. This value is what we consider to be safe to use in such comparisons to account
+    // for these scenarios.
+    const SAFE_TIME_DELTA = TIME_FUDGE_FACTOR * 3;
+
+    // If iOS check if we have a real stalled or supend event or
+    // we got stalled/suspend due headphones where disconnected during playback
+    this.on(['stalled', 'suspend'], (e) => {
+      const buffered = this.buffered();
+
+      if (buffered.length > 0) {
```
brandonocasey: we should do 
```
if (!bufferred.length) {
  return;
}
```

to eliminate the scope here.

PR #6216, opened 2019-09-05
fix: ensure the default ID of the first player is 'vjs_video_3' as some people have relied on this

misteroneill:
## Description
When a player is created without an `id` on the embed code, Video.js automatically assigns it one based on an auto-incrementing number (a.k.a. a GUID). For the longest time, this has happened to result in the default `id` of the first player being `vjs_video_3`.

It was never intended for users to rely on this value being consistent, but users do strange and inadvisable things.

PR #6103 had an unintended side effect in that it changed the default `id` to `vjs_video_2`, which we worry could affect some users of Video.js.

## Specific Changes proposed
* Increment the internal starting point of the GUID from 1 to 3.
* Avoid doing anything that increments the GUID simply by including Video.js on the page.
* Add a test that simulates the initial player creation by resetting the GUID to its initial value to verify that the first player created has its `id` assigned as `vjs_video_3`.

## Requirements Checklist
  - [ ] ~Docs/guides updated~
  - [ ] ~Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))~

### DIFF
--- a/src/js/setup.js
+++ b/src/js/setup.js
@@ -5,7 +5,6 @@
  * @module setup
  */
 import * as Dom from './utils/dom';
-import * as Events from './utils/events.js';
 import document from 'global/document';
 import window from 'global/window';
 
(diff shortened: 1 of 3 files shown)


misteroneill:
This should probably be merged to a branch off of 7.6.4 as that's what I based this branch on.

misteroneill:
Maybe, but `data-setup` wasn't involved here. The changes to `setup.js` are there to prevent the incrementing of the GUID before any players are created.

brandonocasey:
data-setup is just an easy way to force videojs to use a `guid` for a element with no `id` attribute.

PR #6220, opened 2019-09-08
fix(lang): Update Norwegian translations

danmichaelo:
Add missing translations for Norwegian Bokmål (nb) and Nynorsk (nn).

Note: I translated "Seek to live, currently playing live" as something like "Currently playing live" since I found the original message very odd-sounding. If you are already live, why would you want jump to where you currently are? My guess is that this message was written from a very technical point of view. Would it perhaps be an idea to change the English original too? I can create a separate issue for that if you think it might be a good idea.


gkatsev:
The reason for the message is that it is informative to screen reader users. The first piece informs that user that it is a button that when pressed will take them to the live point, the second part informs them whether they are currently live or not. @OwenEdwards may be able to provide a bit more insight into the decision, if there's anything else to add.

gkatsev:
Also, thanks for updating the translations, it's very much appreciated!

danmichaelo:
Thanks for clarifying. I feel like the best would be if the button was not active when you're already playing live since there's no reason to jump to live if you're already live. That goes for users of screen readers and other users alike.

But I'm digressing from the translation :) It's probably better to kep the translation close to the original even when the original is a bit odd so I amended my translation so that it provides the same message as the English original.

gkatsev:
Thanks! If you think that the text/behavior should be changed, please open another issue.

Issue #6232, opened 2019-09-16
[Bug report] ProgressBar does not update as expected


ustbhuangyi:
## Description

When pausing the video and click the `ProgressBar`, the video update but the `ProgressBar`  does not update as expected.

Test it with 'https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0'.

## Steps to reproduce
1. click `Play Button`
2. click `Pause Button`
3. click `ProgressBar`

### Expected
Video update and `ProgressBar` update to the right position.

### Actual
Video update but `ProgressBar` does not update.

#### videojs
after v7.5.1, v7.5.0 is fine.
#### browsers
Chrome、Firefox etc.
#### OSes
Mac OS

PR #6155, opened 2019-08-02
perf: only update ui on change, wrap things in requestAnimationFrame

brandonocasey:
## Description
Only update ui components when something has actually changed and wrap ui changes in requestAnimationFrame. This nets us most of the benefit from #6150

PR #6234, opened 2019-09-17
fix: update the progress-bar correctly when pausing the video and cli…

ustbhuangyi:
## Description

First, when handle `mouse down` event of the `seek-bar`, we set player's `scrubbing_` to true, and at the same tick, the `update` function called and we call `super.update()`. However, it returns `undefined` because the progress equals to the previous progress, the root cause is that we get cached `currentTime`. So we call `super.update()` in the next tick to make sure we can get the correct progress.

Second, when toggle visibility of the document, we should not call `this.enableInterval_()` when the player is paused or waiting.

Fixed https://github.com/videojs/video.js/issues/6232

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -78,7 +78,9 @@ class SeekBar extends Slider {
     if (document.hidden) {
       this.disableInterval_(e);
     } else {
-      this.enableInterval_();
+      if (!this.player_.paused() && !this.player_.hasClass('vjs-waiting')) {
+        this.enableInterval_();
+      }
 
       // we just switched back to the page and someone may be looking, so, update ASAP
       this.update();
@@ -133,9 +135,8 @@ class SeekBar extends Slider {
    *          The current percent at a number from 0-1
    */
   update(event) {
-    const percent = super.update();
-
     this.requestAnimationFrame(() => {
+      const percent = super.update();
       const currentTime = this.player_.ended() ?
         this.player_.duration() : this.getCurrentTime_();
       const liveTracker = this.player_.liveTracker;
@@ -167,8 +168,6 @@ class SeekBar extends Slider {
         this.duration_ = duration;
       }
     });
-
-    return percent;
   }
 
   /**


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -167,8 +168,6 @@ class SeekBar extends Slider {
         this.duration_ = duration;
       }
     });
-
-    return percent;
```
gkatsev: we should still be returning this value as it may be a breaking change.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -78,7 +78,9 @@ class SeekBar extends Slider {
     if (document.hidden) {
       this.disableInterval_(e);
     } else {
-      this.enableInterval_();
+      if (!this.player_.paused() && !this.player_.hasClass('vjs-waiting')) {
```
gkatsev: I'm not sure that we should end up paused with vjs-waiting.

gkatsev: Also, seems like this is unrelated to the actual issue, just a nice improvement?

ustbhuangyi: Yes, it's a improvement.It's no need to call update if the player is paused or waiting.

src/js/control-bar/progress-control/seek-bar.js
ustbhuangyi: Because the `update` function is throttled, so even you trigger a  `timeupdate` event manually when finish the seek on `mouseup`, the `update` function only called once at the same tick.

ustbhuangyi: The return value of `update` function seems not used.

gkatsev: Ah, yeah, the throttling is a good point.

While we may not be using it, if someone who uses Video.js is extending the SeekBar themselve and expecting `update` to return a value, they will break. We should avoid changes that have a potential breaking change in them.

ustbhuangyi: Before this follow commit: https://github.com/videojs/video.js/pull/6155
The `update` function does not has a return value, so I think it's not breaking change.

ustbhuangyi: Since the `percent` depends on the Slider's `update` function. The return of undefined does not make sense.
https://github.com/videojs/video.js/blob/master/src/js/slider/slider.js#L238-L248

ustbhuangyi: I tried to remove the throttling of the `update` function, it works, but may cause a little perf issue. So there's a trade-off.

ustbhuangyi: @gkatsev, If you think removing the throttling of the `update` function is more reasonable, I will send another pr.

gkatsev: No, I don't think we should unthrottle it. I'll have to think about it a bit.

ustbhuangyi: Ok, hope for a better solution.

PR #6243, opened 2019-09-24
fix: IE11 and Edge focus on fullscreen toggle

### LINKED ISSUE #6242: IE11 + Edge FullscreenToggle button focus (screen-readers / keyboard users)
## Description
When first activating the fullscreen button in IE11 and Edge, focus is lost and placed on the window. This then requires keyboard or screen-reader users to manually navigate through each control to get back to the fullscreen button.

## Steps to reproduce
1. Navigate to https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0 using IE11 or Edge.
2. Play the video, and click on the fullscreen toggle.
3. Check document.activeElement to see if the fullscreen button is still focused on, or see if tapping enter/spacebar closed fullscreen mode.

### Expected
When activating the fullscreen toggle button, focus should remain on the the button.

### Actual
In IE11 and Edge, focus is moved to the window after activating the fullscreen button.

#### browsers
IE11, Edge
#### OSes
Windows

shaneafsar:
On IE11 and Edge, focus is moved to the window when fullscreen is toggled. This ensures screen readers keep focus on the FullscreenToggle element after activating the button.

Fixes #6242

## Description
See #6242

## Specific Changes proposed
-Added this.el().focus() to the FullscreenToggle component whenever "handleClick" is invoked.

## Requirements Checklist
  - [X] Change has been verified in an actual browser (IE, Edge)

### DIFF
--- a/src/js/control-bar/fullscreen-toggle.js
+++ b/src/js/control-bar/fullscreen-toggle.js
@@ -54,6 +54,15 @@ class FullscreenToggle extends Button {
     } else {
       this.controlText('Fullscreen');
     }
+    /**
+     * On IE11 and Edge, focus is moved to the window when fullscreen is
+     * enabled. This ensures screen readers keep focus on the FullscreenToggle
+     * element after activating the button.
+     */
+    if (this._clickElement === this.el()) {
+      this.el().focus();
+      this._clickElement = null;
+    }
   }
 
   /**
@@ -68,6 +77,7 @@ class FullscreenToggle extends Button {
    * @listens click
    */
   handleClick(event) {
+    this._clickElement = event.currentTarget;
     if (!this.player_.isFullscreen()) {
       this.player_.requestFullscreen();
     } else {
--- a/src/js/control-bar/fullscreen-toggle.js
+++ b/src/js/control-bar/fullscreen-toggle.js
@@ -11,7 +11,6 @@ import document from 'global/document';
  * @extends Button
  */
 class FullscreenToggle extends Button {
-
   /**
    * Creates an instance of this class.
    *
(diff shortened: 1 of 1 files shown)


gjanblaszczyk:
Hi ima007, I have one question about testing... Did you test focus-visible.html example page from sandbox directory after your changes on Chrome/Firefox? Does it work the same if you click via the mouse on the fullscreen button? I am asking because I haven't checked your changes yet and I wonder if some behavior changed after your PR.

src/js/control-bar/fullscreen-toggle.js
```diff
@@ -74,8 +73,13 @@ class FullscreenToggle extends Button {
     } else {
       this.player_.exitFullscreen();
     }
+    /**
+     * On IE11 and Edge, focus is moved to the window when fullscreen is
+     * enabled. This ensures screen readers keep focus on the FullscreenToggle
+     * element after activating the button.
+     */
+    this.el().focus();
```
OwenEdwards: I'm wondering if this is going to work consistently; isn't the switch to Full Screen a *request* that may happen asynchronously? In that case, this focus() may happen before the request has completed.

ima007:
You’re correct; I believe the fullscreenchange event is what gets fired
when the promise for requesting fullscreen is resolved.

This change should likely be moved to handleFullscreenChange and detect if
it was from activating the button.

ima007:
@OwenEdwards -- Since a promise isn't returned by requestFullScreen isn't available in all browsers, I took a quick approach with storing the current click target, and checking that in the fullscreenchange event handler for the FullscreenToggle, since by that point, the async should be completed.

o-hreshchuk:
are these changes mean the controls will be visible all the time in the fullscreen if the user doesn't change the focus on another element or click play/pause e.g?

gkatsev:
Seems like this change breaks the focus visible stuff in Firefox. Seems to be fine in Chrome, though.

@o-hreshchuk I don't think it'll keep the controls up.

PR #6249, opened 2019-10-01
feat: resets pastSeekEnd_ variable.

gjanblaszczyk:
## Description
It resets pastSeekEnd_ value to 0 when its value is higher than seeking increment value.
In rare cases, for example when we stop ingesting the live stream. The playlist duration doesn't change at all but we are constantly increasing the value of the pastSeekEnd_ variable which forces the DVR progress forward and increases the size of the gap on the progress bar.

## Specific Changes proposed
If the pastSeekEnd_ value is higher than 1.5 x seeking increment value then set pastSeekEnd_ variable value to 0.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -88,7 +88,13 @@ class LiveTracker extends Component {
       this.trigger('seekableendchange');
     }
 
-    this.pastSeekEnd_ = this.pastSeekEnd() + 0.03;
+    // we should reset pastSeekEnd when the value
+    // is much higher than seeking increment.
+    if (this.pastSeekEnd() > this.seekableIncrement_ * 1.5) {
+      this.pastSeekEnd_ = 0;
+    } else {
+      this.pastSeekEnd_ = this.pastSeekEnd() + 0.03;
+    }
 
     if (this.isBehind_() !== this.behindLiveEdge()) {
       this.behindLiveEdge_ = this.isBehind_();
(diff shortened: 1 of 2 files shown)


gkatsev:
Isn't it possible that we've just fallen behind so much?

gjanblaszczyk:
Hi Gary, Yes, I think so,
you can see the problem here: https://codepen.io/grbla/pen/zYYOqqM
Please play the stream and wait around 40 sec. You will see that the progress bar is increasing the duration value forever but in fact, it should stay inside the live edge bounds and duration should never fallback behind the liveEdgeWindow value.

gkatsev:
It definitely makes sense for that example but is it possible that we've actually fallen behind enough to be more than 1.5x the seekable increment?

gjanblaszczyk:
Hi Gary, I think it may be possible when somebody has a segment duration set much higher than 12. The video.js has a seeking increment by default set to 12. So I guess if somebody will increase segment duration value in the manifest to be a little higher than 1.5 x 12 = 18 then it is possible but only when somebody stops ingestion and reload the player. However, I don't think that someone will really set a segment duration higher than 12 in the manifest... Maybe this segment increment default value should be configurable somehow.

gkatsev:
Looks like tests are failing: https://travis-ci.org/videojs/video.js/builds/592287905#L777-L818. Can you take a look?

gjanblaszczyk:
Yes, of course. I have just updated the tests... should work now.

PR #6254, opened 2019-10-04
Minor update to components.md

OwenEdwards:
Wondering if the recent update to components.md shouldn't also have the note "_(hidden, unless there are relevant tracks)_" next to the `SubsCapsButton` (like all other text track buttons)? Sure, in this case "_relevant_" is two different `kind`s of text tracks, but it still holds, doesn't it?

If I'm wrong about this, feel free to dismiss this PR! I just wanted to add it as I didn't see/review the #6253 before it was merged.


gkatsev:
Makes sense to me.

Issue #6207, opened 2019-08-30
Subtitles gets hidden behind the control bar in Chrome, Firefox & Edge


martinstender:
## Description
As far as I understand, subtitles should move up and out of the way of the control bar, when the control bar becomes visible (on hover, usually).
But in Chrome and Firefox (both macOs and Win10) and Edge in Win10, only the subsequent subtitles are moved up - so the subtitles that are already visible when the control bar becomes active, get covered up by the control bar.

## Steps to reproduce
In Chrome, Firefox or Edge, go to this codepen, press play and move the cursor away from the player, to make the control bar go away.
[https://codepen.io/martinstender/pen/mdbwwGo?editors=1010](https://codepen.io/martinstender/pen/mdbwwGo?editors=1010)
After approximately 9 seconds (when Elton sits down at the piano at home 🙂 ), hover over the player again and you should see the subtitles get covered up by the control bar.

### Expected
Again, as far as I understand, subtitles should move out of the way as the control bar becomes visible, but that only happens in Safari.

### Actual
Subtitles get hidden behind the control bar.

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
v7.0
#### browsers
Chrome, FireFox, but not Safari
#### OSes
macOs and Windows 10
### plugins
None

gkatsev:
Yeah, I can confirm it doesn't work properly. The captions container (text-track-display) does continue getting the `bottom: 3em;` css applied properly, so, something must've been done recently that made that not matter.

(Also, if you update to 7.6.4, you wouldn't need that weird `player.vhs = null` workaround, as we fixed the stuttering issue).

gkatsev:
Trying it out, looks like this never worked. I went all the way back to 6.0 (which was released april 2017). https://codepen.io/gkatsev/pen/xxKLELa?editors=1010
Definitely should be fixed but good to know that it isn't actually a regression.

martinstender:
Thanks for looking into it, @gkatsev.
And I'll upgrade to 7.6.4 right now :-)

tsi:
We also noticed this issue, created https://github.com/videojs/video.js/pull/9023 in an attempt to fix.

PR #6265, opened 2019-10-12
fix: Subtitles gets hidden behind the control bar.

gjanblaszczyk:
## Description
This PR fixes subtitles issue described in PR #6207 .
The Demo with this fix is available here: https://codepen.io/grbla/pen/xxxZPQm?editors=1010
## Specific Changes proposed
Add resize observer for the "vjs-text-track-display" div element.
The fix should work on the latest Firefox, Chrome, Safari and Opera.
However, you will need to use resizeObserver polyfill for IE11 or Edge browser to make things work.
 
## Requirements Checklist
  - [ ] Docs/guides updated
  - [X] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -116,10 +116,26 @@ class TextTrackDisplay extends Component {
       }
 
       player.on('fullscreenchange', updateDisplayHandler);
-      player.on('playerresize', updateDisplayHandler);
+
+      if (this.player_.resizeManager) {
+        this.ResizeObserver = this.player_.resizeManager.ResizeObserver || window.ResizeObserver;
+      }
+      if (this.ResizeObserver) {
+        this.resizeObserver_ = new this.ResizeObserver(Fn.debounce(() => {
+          updateDisplayHandler();
+        }, 100, false, this));
+        this.resizeObserver_.observe(this.el_);
+      } else {
+        player.on('playerresize', updateDisplayHandler);
+      }
 
       window.addEventListener('orientationchange', updateDisplayHandler);
-      player.on('dispose', () => window.removeEventListener('orientationchange', updateDisplayHandler));
+      player.on('dispose', () => {
+        window.removeEventListener('orientationchange', updateDisplayHandler);
+        if (this.resizeObserver_) {
+          this.resizeObserver_.disconnect();
+        }
+      });
 
       const tracks = this.options_.playerOptions.tracks || [];
 
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -117,11 +117,12 @@ class TextTrackDisplay extends Component {
 
       player.on('fullscreenchange', updateDisplayHandler);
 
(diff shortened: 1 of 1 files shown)


gkatsev:
I don't understand this change. `playerresize` is triggered whenever the player resizes so we know we should be updating the display. During normal playback operation, the text track display should be moving up by 3em and theoretically there isn't a need for anything else.

gkatsev:
Also, `playerresize` comes with a built-in polyfill for the ResizeObserver.

gjanblaszczyk:
Hi Gary, you have right with this `playerresize` event and the logic with updating the player. However, I think there is one gap in the player. I think we should also update the text display itself when the visibility of controlbar changes. If you rollover the player area via mouse then the controlbar will be visible and you should update the text display container to make sure that the already visible text is not behind the controlbar. Is that make sense? or maybe we should always set the text display container to have a static 3em bottom margin that doesn't change when the controlbar is visible or hidden. Any additional thoughts?

gkatsev:
the `bottom: 3em;` should be moving the entire text track display when the control bar is visible to prevent the control bar from overlaying captions. It would be good to track down why that stopped working.

gjanblaszczyk:
Ah, right. I think I know why... :) I will try to update the PR with a new simpler solution...

gjanblaszczyk:
I think the problem is related to the cue display style top value. For example, if you open the captions settings and change the font size value from 100% to 125% then you will see that the issue never happens that because 'overrides.fontPercent' rule overrides the top style variable to 'auto' (https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-display.js#L370). However, I am not sure if setting top: auto by default to all subtitles is fine... @gkatsev any thoughts?

gjanblaszczyk:
> ... It would be good to track down why that stopped working.
btw. @gkatsev  Are you sure it's ever worked properly? I have tested video.js version 4, 5, 6 and 7 but I didn't find working video.js version without this issue. It is also possible that the issue is inside vtt.js lib itself but I can't prove it right now...

gkatsev:
Yeah, definitely worked before. I wonder if it's a soft breakage in browsers where something changed in CSS/browser renderers that made it not work anymore.

If this is because top isn't set properly, I wonder if it should be fixed in vtt.js, or at least verify that top isn't already set, as it could override the value set up by vtt.js.

gjanblaszczyk:
I think vtt.js works correctly because it sets a correct value for the top and bottom text position, the problem is that the browser always uses top position and ignore bottom position value.
Maybe we should check if the bottom value is set by vtt.js and if does then we should set `top` position to `auto` to force the browser to use bottom position instead of the top position. Of course, I assume that both the top and bottom values are correctly calculated by vtt.js. Does it make sense for you?

DamirTubak:
I was doing research on my project in order to fix this behaviour, and have failed, but I managed to pinpoint the videoJS version where this feature actually worked, so maybe you can cross-check!
* Video.js version 7.6.4

PR #6287, opened 2019-10-28
Chore: introduces core ES module.

gjanblaszczyk:
## Description
The PR introduces an alternative ES module file for videojs core build.
I created this PR because I would like to use videojs core directly from the browser like this:

`<script type="module">
import videojs from 'video.js/dist/alt/video.core.es.js';
</script>`

For now, main videojs build supports umd, cjs and es module. Not sure why core build supports only cjs and umd.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -192,7 +192,7 @@ export default cliargs => [
     output: [
       {
         format: 'es',
-        file: 'dist/alt/video.core.es.js',
+        file: 'core.es.js',
         strict: false,
         banner,
         globals: globals.module
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -186,16 +186,24 @@ export default cliargs => [
     onwarn,
     watch
   },
-  // core
+  // core cjs, es
   {
     input: 'src/js/video.js',
-    output: {
-      format: 'cjs',
-      file: 'core.js',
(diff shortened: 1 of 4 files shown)


rollup.config.js
```diff
@@ -186,16 +186,24 @@ export default cliargs => [
...
+    output: [
+      {
+        format: 'es',
+        file: 'dist/alt/video.core.es.js',
+        strict: false,
+        banner,
+        globals: globals.module
+      }, {
+        format: 'cjs',
+        file: 'core.js',
```
gkatsev: the reason we kept this top level is to make it easier for consumers to be able to do `require('video.js/core')`. We can probably expose the ES module file there as well, maybe? core.es.js? Then you could do `import videojs from 'video.js/core.es';`. Thoughts?

gkatsev: @misteroneill thoughts?

misteroneill: Yeah, I agree that having it at the same level as `core.js` makes sense.

gjanblaszczyk: Good idea, I have just updated the output file name from `dist/alt/video.core.es.js` to `core.es.js`

PR #6289, opened 2019-10-28
fix: bring back Android 4.x support

gkatsev:
If we are running on a browser that doesn't support WeakMap, use a
simple WeakMap sham as a fallback.
Also, add a tiny Set sham.

We received a request (in https://github.com/videojs/video.js/pull/6103) that we broke Android 4.x due to lack to WeakMap. This creates a WeakMap sham to use as a fallback when WeakMap isn't available.

To test locally, I change the file to always create the FakeWeakMap and use that as the default export.

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -98,9 +98,33 @@ class Component {
     this.childIndex_ = {};
     this.childNameIndex_ = {};
 
-    this.setTimeoutIds_ = new Set();
-    this.setIntervalIds_ = new Set();
-    this.rafIds_ = new Set();
+    let SetSham;
+
+    if (!window.Set) {
+      SetSham = class {
+        constructor() {
+          this.set_ = {};
+        }
+        has(key) {
+          return key in this.set_;
+        }
+        delete(key) {
+          const has = this.has(key);
+
+          delete this.set_[key];
+
+          return has;
+        }
+        add(key) {
+          this.set_[key] = 1;
+          return this;
+        }
+      };
+    }
+
+    this.setTimeoutIds_ = window.Set ? new Set() : new SetSham();
+    this.setIntervalIds_ = window.Set ? new Set() : new SetSham();
+    this.rafIds_ = window.Set ? new Set() : new SetSham();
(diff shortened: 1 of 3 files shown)


src/js/utils/dom-data.js
```diff
@@ -3,6 +3,62 @@
...
+
+if (!WeakMap) {
+  FakeWeakMap = class {
+    constructor() {
+      this.vdata = 'vdata' + Math.floor(window.performance && window.performance.now() || Date.now());
+      this.data = {};
+    }
+
+    set(key, value) {
+      const access = key[this.vdata] || Guid.newGUID();
```
misteroneill: We have a test to make sure the default player ID is still `vjs_video_3`, right? Would hate for that to "regress" again.

gkatsev: Ah, yeah, we'd want to double check that on devices without WeakMap. Might require a few more changes.

gkatsev: It seems to default to vjs_video_3 in my test just now and the test for default player id is passing on the device.

src/js/utils/dom-data.js
```diff
@@ -3,6 +3,62 @@
  * @module dom-data
  */
 
+import log from './log.js';
+import * as Guid from './guid.js';
+import window from 'global/window';
+
+let FakeWeakMap;
+
+if (!WeakMap) {
```
yanyunlong: if (!WeakMap) {   //'WeakMap' is undefined  Should be changed to ===> typeof (WeakMap) == "undefined",
And browser also doesn't support 'new Set()'

src/js/component.js
```diff
@@ -98,9 +98,33 @@ class Component {
...
+          const has = this.has(key);
+
+          delete this.set_[key];
+
+          return has;
+        }
+        add(key) {
+          this.set_[key] = 1;
+          return this;
+        }
```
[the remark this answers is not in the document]
gkatsev: Added a forEach method. Good catch.

PR #6297, opened 2019-11-03
fix: ensure components added with an index are added in the correct location

mister-ben:
## Description
When an index is specified with `Component.addChild()` then the new component's element is added at the given index. This isn't necessarily correct, as not all of a component's children have DOM elements. Adding a component at the index of the control bar in the player's children will insert its el after the control bar el, because of the MediaLoader and LiveTracker.

## Specific Changes proposed
When inserting the elelment, refer to the `el_` of the sibling component this is to be inserted before.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -463,8 +463,12 @@ class Component {
     // Add the UI object's element to the container div (box)
     // Having an element is not required
     if (typeof component.el === 'function' && component.el()) {
-      const childNodes = this.contentEl().children;
-      const refNode = childNodes[index] || null;
+      // If inserting before a compoennt, insert before that component's element
+      let refNode = null;
+
+      if (this.children_[index + 1] && this.children_[index + 1].el_) {
+        refNode = this.children_[index + 1].el_;
+      }
 
       this.contentEl().insertBefore(component.el(), refNode);
     }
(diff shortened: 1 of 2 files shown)


gkatsev:
Took me a bit to figure out what was wrong here because the issue is so subtle. Is it possible that fixing this will cause issues from people inserting components at particular indexes?

mister-ben:
I only noticed this because the LiveTracker changed the position of an element added by a plugin. I was wondering if this could break anything, but I can't think of a reason to give an index at all other than to set the position in DOM.

src/js/component.js
```diff
@@ -463,8 +463,12 @@ class Component {
     // Add the UI object's element to the container div (box)
     // Having an element is not required
     if (typeof component.el === 'function' && component.el()) {
-      const childNodes = this.contentEl().children;
-      const refNode = childNodes[index] || null;
+      // If inserting before a compoennt, insert before that component's element
```
misteroneill: Typo `compoennt`

misteroneill: Otherwise, LGTM

misteroneill:
Conceptually, I think this is a good change to make. There is _some_ risk of causing problems for advanced use-cases that were relying on the broken behavior, but I think a strong argument can be made that, in this case, that wouldn't constitute a breaking change.

gkatsev:
Yeah, makes sense. We can always revert it if we find this causes a major issue.

PR #6318, opened 2019-11-15
fix: encrypted content goes black in IE/Edge when video element focused

### LINKED ISSUE #6270: DRM content playback issue in IE11 and Edge
## Description
Starting with video.js version 7.5.5 on IE11 and edge, if you play DRM content using big-play-button, only sound comes out and the screen is black.
This problem is caused by the addition of a function to focus on the screen when big-play-button is clicked in 7.5.5.

7.5.5
```
if (this.mouseused_ && event.clientX && event.clientY) {
      silencePromise(playPromise);
      this.player_.tech(true).focus(); // Problem
      return;
    }
```

7.5.4
```
if (this.mouseused_ && event.clientX && event.clientY) {
      silencePromise(playPromise);
      // call handleFocus manually to get hotkeys working
      this.player_.handleFocus({}); // OK
      return;
    }
```

## Steps to reproduce
1. Link to video js version 7.6.5.
2. Set DRM content to video js. ex) playready drm content
3. Click the Play Video button in the top left corner.

### Expected
DRM content play.

### Actual
Only sound comes out and the screen is black

### Error output
None

## Additional Information
#### videojs
7.5.5 or later versions
#### browsers
IE11 and Edge
#### OSes
windows 10

alex-barstow:
## Description
Fixes #6270. This appears to be caused by an unreported browser bug. It also reproduces with shaka player if you call `document.querySelector('video').focus()` while playing back DRM content. For now, the only solution is simply to avoid calling `videoEl.focus()` in IE/Edge for DRM. This unfortunately may prevent hotkeys from working correctly in IE/Edge.

## Specific Changes proposed
Modify the conditions under which we call `this.player_.tech(true).focus()` when the big play button is clicked.

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,6 +153,7 @@
   },
   "vjsstandard": {
     "ignore": [
+      "build",
       "dist",
       "docs",
       "sandbox",
--- a/package.json
+++ b/package.json
@@ -153,7 +153,6 @@
   },
   "vjsstandard": {
     "ignore": [
-      "build",
       "dist",
       "docs",
       "sandbox",
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -48,8 +48,8 @@ class BigPlayButton extends Button {
     // exit early if clicked via the mouse
     if (this.mouseused_ && event.clientX && event.clientY) {
       const sourceIsEncrypted = this.player_.usingPlugin('eme') &&
-                          this.player_.eme.sessions &&
-                          this.player_.eme.sessions.length > 0;
+                                this.player_.eme.sessions &&
+                                this.player_.eme.sessions.length > 0;
 
       silencePromise(playPromise);
       if (this.player_.tech(true) &&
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -153,6 +153,7 @@
   },
   "vjsstandard": {
     "ignore": [
+      "build",
```
gkatsev: What type of errors were you seeing? I don't have any issues running the linter locally.

src/js/big-play-button.js
```diff
@@ -46,8 +47,16 @@ class BigPlayButton extends Button {
 
     // exit early if clicked via the mouse
     if (this.mouseused_ && event.clientX && event.clientY) {
+      const sourceIsEncrypted = this.player_.usingPlugin('eme') &&
+                          this.player_.eme.sessions &&
```
gkatsev: should probably either line up or be indented once

alex-barstow: Yeah it was lined up in my editor but the indentation didn't carry over correctly for some reason. I'll figure it out.

package.json
alex-barstow: My local `build/` doirectory was somehow outdated. This is not needed so I removed it.

gkatsev:
Looks like *any* programmatic call to focus causes the video element to turn black.

PR #6319, opened 2019-11-15
Fix: Add detection to for iPadOS in IS_IPAD const

marcodeltorob:
## Description
New iPadOS User-agent does not contain "iPad" string in order to detect the device 

## Specific Changes proposed
Detect  IS_SAFARI && TOUCH_ENABLED we can conclude is and iPad Device.

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -181,8 +181,7 @@ export const TOUCH_ENABLED = Dom.isReal() && (
  * @const
  * @type {Boolean}
  */
-export const IS_IPAD = (/iPad/i).test(USER_AGENT) ||
-(IS_SAFARI && TOUCH_ENABLED);
+export const IS_IPAD = (/iPad/i).test(USER_AGENT) || (IS_SAFARI && TOUCH_ENABLED);
 
 /**
  * Whether or not this device is an iPhone.
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -10,27 +10,6 @@ const webkitVersionMap = (/AppleWebKit\/([\d.]+)/i).exec(USER_AGENT);
 const appleWebkitVersion = webkitVersionMap ? parseFloat(webkitVersionMap.pop()) : null;
 
 /**
- * Whether or not this device is an iPad.
- *
- * @static
- * @const
- * @type {Boolean}
- */
-export const IS_IPAD = (/iPad/i).test(USER_AGENT);
-
-/**
- * Whether or not this device is an iPhone.
- *
- * @static
- * @const
- * @type {Boolean}
- */
-// The Facebook app's UIWebView identifies as both an iPhone and iPad, so
-// to identify iPhones, we need to exclude iPads.
-// http://artsy.github.io/blog/2012/10/18/the-perils-of-ios-user-agent-sniffing/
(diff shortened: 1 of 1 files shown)


src/js/utils/browser.js
```diff
@@ -212,3 +173,43 @@ export const TOUCH_ENABLED = Dom.isReal() && (
...
+
+/**
+ * Whether or not this device is an iPad.
+ *
+ * @static
+ * @const
+ * @type {Boolean}
+ */
+export const IS_IPAD = (/iPad/i).test(USER_AGENT) ||
+(IS_SAFARI && TOUCH_ENABLED);
```
gkatsev: This can just be a single line.

blacktrash: what about Chrome on iPadOS?

gkatsev: chrome is unaffected.

src/js/utils/browser.js
```diff
@@ -212,3 +173,42 @@ export const TOUCH_ENABLED = Dom.isReal() && (
...
   window.DocumentTouch && window.document instanceof window.DocumentTouch);
+
+/**
+ * Whether or not this device is an iPad.
+ *
+ * @static
+ * @const
+ * @type {Boolean}
+ */
+export const IS_IPAD = (/iPad/i).test(USER_AGENT) || (IS_SAFARI && TOUCH_ENABLED);
```
Viktor286hearts: Could it be that `IS_SAFARI && TOUCH_ENABLED` also will include iPhone as well?

gkatsev: That is a good question.

gkatsev: Yup, looks like we should add a `!! IS_IPHONE` or something.

Viktor286hearts: It could be out of scope, sorry for mixing two things. But, is it necessary to use regexp here to detect substring? Could it better for performance to use `includes` or `indexOf`? If you agree, i could make a PR with refactor of this utlis `browser.js` module.

gkatsev: PR here https://github.com/videojs/video.js/pull/6371

gkatsev: Unfortunately, `includes` isn't available on all platforms that we support and I'm not sure that the extra code that would arise from using `indexOf` is worth it, especially since it may end up being slower. Is there a specific benefit to using `include` or `indexOf`?

Viktor286hearts: The idea was to avoid regex because it much more performance demanding while `indexOf` or `include` could be more optimized (now or in later js engines). Although, it's not a huge optimization anyway.

gkatsev: While I don't think benchmarks are that useful, this one does show that the behavior is different depending on the browser https://jsperf.com/startswith-vs-indexof-vs-regex/4. For example in Chrome `indexOf` is much faster while on Firefox `startsWith` is faster with the `Regexp#test` being on par. Also, given that this runs once, I'm not sure it really matters what is being done and shouldn't ever be in the critical path, so I'd be inclined to just leave it as is.

Viktor286hearts: Sounds reasonable! Thanks for the shared info. For me it also performed differently for mobile and for desktop with that benchmark http://jsben.ch/d4o38

PR #6334, opened 2019-11-26
fix: turn on strict mode again

gkatsev:
We had to turn off strict mode (https://github.com/videojs/video.js/pull/4551) in Video.js due to a change in vtt.js. That has now been fixed in videojs/vtt.js#40 and released as part of 0.15.2 which will be available via https://github.com/videojs/video.js/pull/6333.

Fixes videojs/vtt.js#15

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1218,7 +1218,7 @@ Html5.Events = [
   ['featuresNativeVideoTracks', 'supportsNativeVideoTracks'],
   ['featuresNativeAudioTracks', 'supportsNativeAudioTracks']
 ].forEach(function([key, fn]) {
-  defineLazyProperty(Html5.prototype, key, () => Html5[fn](), false);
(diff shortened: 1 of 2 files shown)


src/js/tech/html5.js
```diff
@@ -1218,7 +1218,7 @@ Html5.Events = [
   ['featuresNativeVideoTracks', 'supportsNativeVideoTracks'],
   ['featuresNativeAudioTracks', 'supportsNativeAudioTracks']
 ].forEach(function([key, fn]) {
-  defineLazyProperty(Html5.prototype, key, () => Html5[fn](), false);
+  defineLazyProperty(Html5.prototype, key, () => Html5[fn](), true);
```
gkatsev: The last property here was used to see whether we should create a setter for the property or not. Given that these values do change, particularly with text tracks, we do want these to be settable but because we weren't using strict mode, it was being set without a problem. Instead, we should just make sure these are settable.

Issue #6330, opened 2019-11-23
Video playback constantly stops/pauses on iOS devices.


Robbcode:
## Description
Video playback of MP4 x264/aac file will stop/pause playback shortly after started, if manually unpaused playback will stop again shortly afterwards.

Stop/Pause can happen as quickly as 14 seconds in to playback. 

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Load page on iOS with videojs and mp4 x264/aac video.
2. Press to start playback on iOS device.

### Expected
Video playback of MP4 file should be uninterrupted. 

### Actual
Video playback of MP4 stops/pauses repeatedly. 

### Error output
Browser console shows no errors.

## Additional Information

I have reproduced the problem on Firefox desktop v70 and setting user agent string to iOS.

what version of videojs does this occur with?

videojs v7.7.0 - v7.7.1

The problem does not seem to happen on version before theses.

browsers / OSes

iPhone/iPad.
iPhone; CPU iPhone OS 12_4_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.1.2 Mobile/15E148 Safari/604.1

Firefox desktop v70.0 with user agent changed to same as above seem to reproduce the problem as well.

### plugins
No plugins

gkatsev:
Do you have the file that you're having issues with? Locally, I do not have any issues with mp4 files with h264 and aac codecs in them. It's likely that you need to re-transcode your file.

woniuzfb:
+1 same issue here using it for HLS no such issue before

woniuzfb:
> Do you have the file that you're having issues with? Locally, I do not have any issues with mp4 files with h264 and aac codecs in them. It's likely that you need to re-transcode your file.

[HLS constantly pauses](http://hbo.epub.fun)

DispatchCommit:
What tech are you using for your player? Are you overriding the native player for any (all?) browsers? Does changing the overrideNative value fix the issue?

```javascript
// Create video.js player
this.player = videojs('streamplayer', {
  // ...
  html5: {
    hls: {
      overrideNative: !videojs.browser.IS_SAFARI,
      // ...
    },
  },
});
```

woniuzfb:
> Are you overriding the native player for any (all?) browsers? Does changing the overrideNative value fix the issue?

overrideNative: true
No.It didn't fix the issue.
Chrome/Safari/all browsers on ios happens

[...]

PR #6354, opened 2019-12-12
Apply the IOS headphone disconnect patch only after successful playback

klipstein:
## Description

Since using the version which includes https://github.com/videojs/video.js/pull/6199 we saw issues when using videojs in combination with videojs+ima. Because Google IMA uses the same video-tag on iOS to display video ads it happens that a `suspend` event is occurring when video sources get switched (in my case a preroll in a muted autoplay scenario). And calling `pause()` in this switching phase could lead to a state within IMA that the player gets stalled.

A fix could either be to enable / disable this headphone-disconnect patch or this pull request, which only does stalled/suspend checks after a video-source was played for at least 1s.

## Requirements Checklist
- [X] Bug fixed

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -225,16 +225,20 @@ class Html5 extends Tech {
     const SAFE_TIME_DELTA = TIME_FUDGE_FACTOR * 3;
 
     // try to determine whether a successful playback occurred
-    const SUCCESSFUL_PLAYBACK_TIME = 1; // second
-    const timeupdateAfterOneSecond = false;
+    const SUCCESSFUL_PLAYBACK_TIME = 1;
+
+    let timeupdateAfterOneSecond = false;
+
+    const timeUpdateListener = function() {
+      if (this.currentTime() > SUCCESSFUL_PLAYBACK_TIME) {
+        this.off('timeupdate', timeUpdateListener);
+        timeupdateAfterOneSecond = true;
+      }
+    };
+
     this.on('loadedmetadata', function() {
       timeupdateAfterOneSecond = false;
-      const timeUpdateListener = function() {
-        if (this.currentTime() > SUCCESSFUL_PLAYBACK_TIME) {
-          this.off('timeupdate', timeUpdateListener);
-          timeupdateAfterOneSecond = true;
-        }
-      };
+
       // disconnect previous listener
       this.off('timeupdate', timeUpdateListener);
       this.on('timeupdate', timeUpdateListener);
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -224,9 +224,32 @@ class Html5 extends Tech {
...
 
+    // try to determine whether a successful playback occurred
+    const SUCCESSFUL_PLAYBACK_TIME = 1;
+
+    let successfulPlayback = false;
+
+    const timeUpdateListener = function() {
+      if (this.currentTime() > SUCCESSFUL_PLAYBACK_TIME) {
+        this.off('timeupdate', timeUpdateListener);
+        successfulPlayback = true;
```
gkatsev: Might be better to move the ios check into a helper method, with a way to disable it. Then enable it here on successful playback and disable it below on new video.

ewojjowe: I would agree with that.

klipstein: Thanks for the feedback. Will adjust this pull request.

klipstein: I've adjusted the pull request and tested it locally in combination with videojs-ima on iOS.

src/js/tech/html5.js
```diff
@@ -209,11 +209,40 @@ class Html5 extends Tech {
...
+    const SUCCESSFUL_PLAYBACK_TIME_SECONDS = 1;
+    const boundStalledAndSuspendHandler = this.handleIOSHeadphoneStalledAndSuspendEvent_.bind(this);
+
+    const timeUpdateListener = function() {
+      if (this.currentTime() > SUCCESSFUL_PLAYBACK_TIME_SECONDS) {
+        this.off('timeupdate', timeUpdateListener);
+
+        // If iOS check if we have a real stalled or supend event or
+        // we got stalled/suspend due headphones where disconnected during playback
+        this.on('stalled', boundStalledAndSuspendHandler);
```
gkatsev: this can be combined into one line by using an array of values `['stalled', 'suspend']`, can also be used for the `.off` below.

gkatsev:
also, I wonder if the issue is also just #6330, though, the changes here are good.

src/js/tech/html5.js
klipstein: Adjusted the pull here https://github.com/videojs/video.js/pull/6354/commits/658dcf53931827f1a21254c61541c5a74bb58cb1

gkatsev:
Yeah, I'm looking into that one. It's definitely possible that we may want to revert that change. Though, I'm thinking that maybe the change we may want to pause if we don't continue playback. Unfortunately, locally, when unplugging headphones I'm not getting any events but I see that Video.js is in a playing state.

klipstein:
@gkatsev reverting would also fix my initial problem. Feel free to close my pull in case you revert the previous change.

gkatsev:
yeah, I would recommend sticking with Video.js 7.6.6 for now. If we end up keeping the change I'm definitely going to merge this in because it's a nice improvement but it looks more and more like we may want to revert that change.

gkatsev:
I think we're going to revert the change and be done with it. Really appreciate the PR, though. Thanks!

gkatsev:
7.7.4 is released with the reverted change.

PR #6371, opened 2019-12-23
fix: IS_IPAD should be false on iPhone

gkatsev:
Safari user agent. However, the new check also returns true for iPhones.
Therefore, we should exclude iPhones from the check.

Unfortunately, this means that the iPhone user agent detection on
iPhones inside the Facebook app may break again as previously
(https://artsy.github.io/blog/2012/10/18/the-perils-of-ios-user-agent-sniffing/) we
excluded iPads from the iPhone check but that won't work anymore.

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -181,7 +181,8 @@ export const TOUCH_ENABLED = Dom.isReal() && (
  * @const
  * @type {Boolean}
  */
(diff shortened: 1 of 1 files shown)


gkatsev:
Just verified with latest Facebook app on iPhone and they report an iPhone user agent now so there isn't any issue with doing this unless you're using a super old version of facebook.

Issue #6370, opened 2019-12-23
Progress control UI is not updating correctly on seek of paused video


onerzafer:
## Description
Progress control is not updating as expected when the video is paused and the user would like to seek a point in time by clicking the progress bar.
[reduced test case](https://codesandbox.io/s/gracious-spence-xv0si?fontsize=14&hidenavigation=1&theme=dark)

## Steps to reproduce
1. Open the reduced test case
2. play the video
3. pause the video
4. wait until video fully loaded
5. click to the progress bar to seek the video for a different time
6. observe the progress bar and progress bar handle is not moving to the clicked point.
7. play the video and observe the UI update correctly and video actually play from where the user clicks

### Expected
Regardless of the video is paused or not the user should be able to see an update from the UI.

### Actual
On a paused video clicking the progress bar correctly updates the vide but not the progress bar.

### Error output
There is no js error only UI is behaving oddly.

#### videojs
video.js 7.7.1
#### browsers
Tested on Chrome
#### OSes
Tested on Mac Catalina 10.15.1
### plugins
no plugins used

gkatsev:
Thanks, it's something that we got an issue for previously but accidentally got closed #6232. They also opened a PR but the fix is a bit trickier than expected #6234.

gkatsev:
Going to close this as a duplicate of those two. Thanks for writing up a detailed report.

PR #6372, opened 2019-12-23
fix: updates seekbar position after mouse up event is triggered.

gjanblaszczyk:
## Description
A proposed fix for #6370
Progress control is not updating as expected when the video is paused and the user would like to seek a point in time by clicking the progress bar.

## Specific Changes proposed
Improves handle for `mouseup` event inside the seekbar component. When the player is in pause mode and the `mouseup` event is triggered we want to call the update method without `throttling` to be sure that the progress bar position is up to date.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -326,6 +326,14 @@ class SeekBar extends Slider {
     }
     this.player_.scrubbing(false);
 
+    /**
+     * Trigger timeupdate because we're done seeking and the time has changed.
+     * This is particularly useful for if the player is paused to time the time displays.
+     *
+     * @event Tech#timeupdate
+     * @type {EventTarget~Event}
+     */
+    this.player_.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
     if (this.videoWasPlaying) {
       silencePromise(this.player_.play());
     } else {
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -52,7 +52,8 @@ class SeekBar extends Slider {
    * @private
    */
   setEventHandlers_() {
-    this.update = Fn.throttle(Fn.bind(this, this.update), UPDATE_REFRESH_INTERVAL);
+    this.update_ = Fn.bind(this, this.update);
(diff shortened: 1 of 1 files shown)


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -325,16 +326,12 @@ class SeekBar extends Slider {
...
     this.player_.scrubbing(false);
 
-    /**
-     * Trigger timeupdate because we're done seeking and the time has changed.
-     * This is particularly useful for if the player is paused to time the time displays.
-     *
-     * @event Tech#timeupdate
-     * @type {EventTarget~Event}
-     */
-    this.player_.trigger({ type: 'timeupdate', target: this, manuallyTriggered: true });
```
gkatsev: why remove this?

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -325,16 +326,12 @@ class SeekBar extends Slider {
...
     if (this.videoWasPlaying) {
       silencePromise(this.player_.play());
+    } else {
```
gkatsev: Makes sense. That's what I was thinking about but never gotten around to investigating yet. Thanks.

src/js/control-bar/progress-control/seek-bar.js
gjanblaszczyk: During my tests, it looks like this is not needed anymore but maybe it is used somewhere... I didn't check Flash but not sure if Flash is still in use... However, if you think that I shouldn't remove it I can add it again. any additional thoughts?

gkatsev: Yeah, I'd be wary to remove it without dropping older browser versions and what not, which we won't do except in a major version.

gkatsev: Yeah, I'd be wary to remove it without dropping older browser versions and what not, which we won't do except in a major version.

gjanblaszczyk: Yeah, It sounds sane. I have just re-added the trigger for timeupdate event again.

gkatsev: Thanks!

gkatsev:
This fixes #6232 and #6234.

Issue #6391, opened 2020-01-08
Control bar does not automatically hide following right-click on player in Safari and Chrome on macOS


EarlThorfinn:
## Description

In VideoJS, by default, if you move your mouse cursor over a player which is playing a video stream the control bar is shown at the bottom of the player.  If you then move the cursor away from the player the control bar fades to be hidden after approx 3 seconds.

There appears to be an issue however in VideoJS 7.6.6 (and the 7.7.4 pre-release) in Safari and Chrome on macOS where the control bar does not disappear if you first right-click on the player and then move mouse away.  (Users may right-click to access options such as the Picture-in-picture feature).  Even if you then left-click on a blank space on the page, which hides the right-click context menu, the control bar does not automatically hide.  

I've replicated the issue in Safari 13.0.4 and Chrome 79.0.3945.117 on macOS 10.15.2.  

To hide the control bar the user either needs to refresh the page, or pause and restart the player.

I've only replicated the issue with Chrome and Safari on macOS.  Firefox does not appear to be affected.  Browsers on Windows are also unaffected.

For testing, I've setup a 7.6.6 player at https://codepen.io/EarlThorfinn/pen/ZEYrjdx

The issue is also present for me when testing the demo player currently on the VideoJS homepage at: https://videojs.com/city

## Steps to reproduce

1. Go to https://codepen.io/EarlThorfinn/pen/ZEYrjdx using Safari or Chrome on macOS
2. Play the video
3. Right-click on the video
4. Move the mouse cursor away from the player

### Expected

The control bar should disappear approx 3 seconds after moving the mouse away from the player.

### Actual

The control bar remains shown.

gkatsev:
Can reproduce, likely what's happening is that we never get the mouseout event and thus don't expect to hide the player. Though, the user activity timeout should handle it.

dani3l3:
Hi there, I have a report from a customer of our site (which uses videojs) who is now seeing the opposite issue: when entering fullscreen mode, the control bar disappears and does not ever appear again. It's just NOT visible when in fullscreen. Going out of full screen (pressing ESC or double-clicking on the video) restores the normal behaviour and, when smaller, you can operate the control bar just fine. It just goes away when in full screen. You can still stop and start clicking on the video so for me it's not a big deal.... but I thought I'll add this information here so you can take a look. 
I was able to test/reproduce on a MacOS Catalina with Safari 13.1 (thru Browserstack). 
We are using videojs 7.8.4 (latest stable release).

PR #6400, opened 2020-01-13
fix: improves control bar hiding functionality

gjanblaszczyk:
## Description
A proposed fix for #6391 

## Specific Changes proposed
Adds support for `mouseout` event to catch the user click outside the player.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3860,7 +3860,7 @@ class Player extends Component {
       mouseInProgress = this.setInterval(handleActivity, 250);
     };
 
-    const handleMouseUpAndMouseOut = function(event) {
+    const handleMouseUpAndMouseLeave = function(event) {
       handleActivity();
       // Stop the interval that maintains activity if the mouse/touch is down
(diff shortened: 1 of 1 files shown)


src/js/player.js
```diff
@@ -3869,7 +3869,8 @@ class Player extends Component {
     // Any mouse movement will be considered user activity
     this.on('mousedown', handleMouseDown);
     this.on('mousemove', handleMouseMove);
-    this.on('mouseup', handleMouseUp);
+    this.on('mouseup', handleMouseUpAndMouseOut);
+    this.on('mouseout', handleMouseUpAndMouseOut);
```
gkatsev: I think it would be better to use `mouseleave` so that we don't reset this unnecessarily if we go through other elements in the player (like mousing around in the control bar).
I see it being triggered when leaving the player even with a right click.

gjanblaszczyk: Yeah, It is a great idea. I have just pushed the changes.

Issue #5984, opened 2019-05-07
Invalid ARIA attribute value: aria-valuenow="NaN"


misteroneill:
## Description
There are times when the progress control has an `aria-valuenow="NaN"` attribute. This likely needs to default to `0` instead.

I will flesh out this issue later.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1.
2.
3.

### Expected
Please describe what you expected to see.

### Actual
Please describe what actually happened.

### Error output
If there are any errors at all, please include them here.

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
what version of videojs does this occur with?
#### browsers
what browser are affected?
#### OSes
what platforms (operating systems and devices) are affected?
### plugins
are any videojs plugins being used on the page? If so, please list them below.

PR #6404, opened 2020-01-14
fix: slider screenreader value returning as NaN

brandonocasey:
## Description
Without these changes `area-valuenow` for the seek bar often reports `NaN`

### DIFF
--- a/src/js/slider/slider.js
+++ b/src/js/slider/slider.js
@@ -244,7 +244,7 @@ class Slider extends Component {
     const progress = this.getProgress();
 
     if (progress === this.progress_) {
-      return;
+      return progress;
     }
 
     this.progress_ = progress;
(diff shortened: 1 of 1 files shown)


src/js/slider/slider.js
```diff
@@ -268,7 +268,7 @@ class Slider extends Component {
    *         percentage filled that the slider is
    */
   getProgress() {
-    return clamp(this.getPercent(), 0, 1).toFixed(4);
+    return Number(clamp(this.getPercent(), 0, 1).toFixed(4));
```
brandonocasey: `toFixed` turns the number that we get back from clamp into a `string` so 0 becomes `"0.0000"` and `0.8777657` becomes `"0.8777"` when it should be a number still.

src/js/slider/slider.js
```diff
@@ -244,7 +244,7 @@ class Slider extends Component {
     const progress = this.getProgress();
 
     if (progress === this.progress_) {
-      return;
+      return progress;
```
brandonocasey: This function has a return value of `progress` so early returns should return that.

gkatsev:
This likely fixes https://github.com/videojs/video.js/issues/5984. Wonder if @misteroneill remembers any specific cases we should look out for.

PR #6405, opened 2020-01-14
fix: liveui tweaks to prevent jitter

brandonocasey:
## Description
Currently the liveui has a few issues 
1. `seek-to-live` button to toggles off and on while playing and the seek bar moves around.
2. When playing behind the seek bar jumps around a lot
3. Seekable interval is not consistent between videos

## Changes
1. `liveCurrentTime` can run much faster than once every 30ms depending on how fast the interval is going. This can put `liveCurrentTime` ahead of `currentTime` until a `seekableEnd` changes. To Solve this we are going to use `deltaTime` rather than expecting a 30ms interval to run every 30ms.
2. `pastSeekEnd` is not being reset at the same time as seekable end is changed, which can cause `liveCurrentTime` to report a value that is a segment length or more greater than what it should be. To solve this `pastSeekEnd` needs to be reset to 0 in its getter and all of the seekable end tracking needs to be done there.
3. `isBehind_()` is called two separate times, this can cause different values to be returned.
4. A culmination of these issues caused the `seekBar` to jump every time `seekableEnd` changed because `liveCurrentTime` was so off from what it should have been.
5. Changed seekable interval to `liveTolerance` and set it to 15s. This means that all videos will be live within 15s of live current time
6. `seekableendchange` was removed, it really wasn't needed.
7. We are always behind when seeking backwards or paused

## Questions/Other Problems
* Do we want the `seekBar` to stay at 100% even though the user has seeked 15s into the past? Is it just good enough to keep the seek bar at it's actual percent but keep the `seek-to-live` button reporting that the video is live? Currently creates weird behavior were seeking back 14s does go back in the video, but the progress bar still shows 100% and that we are live.

## Other Pull requests
* Disable the new ui for short live windows < 30s and when liveCurrentTime is Infinity (like on Android native HLS)
* Only update the seek bar when paused while live, since we will only be changing our time when paused.


test/unit/seek-to-live.test.js
```diff
@@ -2,96 +2,88 @@
 import TestHelpers from './test-helpers.js';
 import sinon from 'sinon';
 import computedStyle from '../../src/js/utils/computed-style.js';
+import { createTimeRange } from '../../src/js/utils/time-ranges.js';
+
```
brandonocasey: I re-wrote most of these tests, they still test the same things but there was a ton of code duplicated between the modules and most of the tests were a variation of the same thing.

gkatsev:
> Do we want the `seekBar` to stay at 100% even though the user has seeked 15s into the past? Is it just good enough to keep the seek bar at it's actual percent but keep the `seek-to-live` button reporting that the video is live? Currently creates weird behavior were seeking back 14s does go back in the video, but the progress bar still shows 100% and that we are live.

That doesn't seem like good experience. Maybe we treat it like that until a seek? If I seek back 14 seconds, it should show me as 14 seconds behind live and it shouldn't show us as live.

src/js/live-tracker.js
```diff
@@ -185,6 +155,7 @@ class LiveTracker extends Component {
       return;
     }
     this.reset_();
+    this.trigger('liveedgechange');
```
brandonocasey: We not fire `liveedgechange` on stopTracking so components can update appropriately.

gkatsev:
I did see a weird thing happen with the venture bros stream where when I seeked to live and the current time displayed switched between 1 and 2 seconds behind live for like 10 seconds, which the switch occurring a couple of times a second.

There's still funkiness happening if you are in the 15 second zone and want to seek to live but aren't showing as live or are showing as live as want to seek back a bit.

brandonocasey:
Just pushed an update. The issue you were seeing was because I wasn't tracking seeking correctly. It took awhile to get correct but I think it is working as expected now.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -261,6 +261,11 @@ class SeekBar extends Slider {
         newTime = newTime - 0.1;
       }
     } else {
+
+      if (distance >= 0.99) {
```
gkatsev: so, if we're really close to the edge, just assume we wanted to seek to live? 👍

src/js/live-tracker.js
```diff
@@ -1,19 +1,43 @@
 import Component from './component.js';
-import median from './utils/median.js';
```
gkatsev: I think if we remove it here, we can't just remove `media.js` as it isn't used anywhere else, I don't think?

src/js/live-tracker.js
```diff
@@ -44,67 +71,45 @@ class LiveTracker extends Component {
...
-      if (this.lastSeekEnd_) {
-        // we try to get the best fit value for the seeking increment
-        // variable from the last 12 values.
-        this.seekableIncrementList_ = this.seekableIncrementList_.slice(-11);
-        this.seekableIncrementList_.push(Math.abs(newSeekEnd - this.lastSeekEnd_));
-        if (this.seekableIncrementList_.length > 3) {
-          this.seekableIncrement_ = median(this.seekableIncrementList_);
-        }
-      }
+    const newTime = window.performance.now().toFixed(4);
```
gkatsev: `toFixed` returns a string.
```suggestion
    const newTime = Number(window.performance.now().toFixed(4));
```

src/js/live-tracker.js
```diff
@@ -44,67 +71,45 @@ class LiveTracker extends Component {
...
-    } else {
-      this.pastSeekEnd_ = this.pastSeekEnd() + 0.03;
+    this.pastSeekEnd_ = this.pastSeekEnd() + deltaTime;
+
+    const liveCurrentTime = this.liveCurrentTime();
+    const currentTime = this.player_.currentTime();
+
+    // we are behind live if any are true
+    // 1. the player is paused
+    // 2. the user seeked to a location 2 seconds away from live
```
gkatsev: This is something that may be worth reading in from VHS when available and assume we aren't live if we are a segment/target duration behind.

gkatsev: Definitely nothing to do right now.

src/js/live-tracker.js
gkatsev: makes sense

src/js/live-tracker.js
```diff
@@ -292,18 +353,21 @@ class LiveTracker extends Component {
...
-    if (this.player_.paused()) {
-      this.player_.play();
-    }
   }
 
+  /**
+   * Dispose of liveTracker
+   */
   dispose() {
+    this.off(document, 'visibilitychange', this.handleVisibilityChange);
```
gkatsev: 👍

src/js/live-tracker.js
```diff
@@ -167,24 +192,22 @@ class LiveTracker extends Component {
    * their initial value.
    */
   reset_() {
+    this.lastTime_ = null;
     this.pastSeekEnd_ = 0;
     this.lastSeekEnd_ = null;
```
gkatsev: should we set this to `-1` or something instead of null to designate a turned off value? I would presume that -1 isn't a valid value for `lastSeekEnd_`. Then, below we can just check if it's -1 instead of whether it's a number.
```suggestion
    this.lastSeekEnd_ = -1;
```

src/js/live-tracker.js
```diff
@@ -257,33 +296,55 @@ class LiveTracker extends Component {
...
+   * The number of seconds that have occured after seekable end
+   * changed. This will be reset to 0 once seekable end changes.
+   *
+   * @return {number}
+   *         Seconds past the current seekable end
    */
   pastSeekEnd() {
+    const seekableEnd = this.seekableEnd();
+
+    if (typeof this.lastSeekEnd_ === 'number' && seekableEnd !== this.lastSeekEnd_) {
```
gkatsev: continuing from above, should lastSeekEnd_ be set to -1 on reset rather than `null`?
```suggestion
    if (this.lastSeekEnd_ !== -1 && seekableEnd !== this.lastSeekEnd_) {
```

gkatsev:
Looks to be working well now. Thanks!

src/js/live-tracker.js
brandonocasey: Yeah it's probably best to keep it a consistent type too.

PR #6409, opened 2020-01-15
fix: add a threshold of 30s for the liveui to show

brandonocasey:
## Specific Changes proposed
Use the old liveui for live videos where `seekableEnd - seekableStart` is less than `30s`. This is configurable with an option.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -113,7 +113,7 @@ class LiveTracker extends Component {
    * and start/stop tracking accordingly.
    */
   handleDurationchange() {
-    if (this.player_.duration() === Infinity && this.liveWindow() > this.options_.trackingThreshold) {
+    if (this.player_.duration() === Infinity && this.liveWindow() >= this.options_.trackingThreshold) {
       if (this.player_.options_.liveui) {
         this.player_.addClass('vjs-liveui');
       }
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -4,12 +4,18 @@ import mergeOptions from './utils/merge-options.js';
 import document from 'global/document';
 import * as browser from './utils/browser.js';
 
+const defaults = {
+  // Number of seconds of live window (seekableEnd - seekableStart) that
+  // a video needs to have before the liveui will be shown.
+  trackingThreshold: 30
+};
+
 /* track when we are at the live edge, and other helpers for live playback */
 class LiveTracker extends Component {
 
   constructor(player, options) {
     // LiveTracker does not need an element
-    const options_ = mergeOptions({createEl: false}, options);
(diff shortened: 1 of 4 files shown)


src/js/live-tracker.js
```diff
@@ -4,12 +4,18 @@ import mergeOptions from './utils/merge-options.js';
...
+  trackingThreshold: 30
+};
+
 /* track when we are at the live edge, and other helpers for live playback */
 class LiveTracker extends Component {
 
   constructor(player, options) {
     // LiveTracker does not need an element
-    const options_ = mergeOptions({createEl: false}, options);
+    const options_ = mergeOptions(defaults, options, {createEl: false});
```
brandonocasey: Added defaults into options.

src/js/live-tracker.js
```diff
@@ -107,9 +113,13 @@ class LiveTracker extends Component {
    * and start/stop tracking accordingly.
    */
   handleDurationchange() {
-    if (this.player_.duration() === Infinity) {
+    if (this.player_.duration() === Infinity && this.liveWindow() > this.options_.trackingThreshold) {
```
brandonocasey: The `liveui` is now activated in the `LiveTracker` instead of `Player` so that we can add in things like `trackingThreshold`.

gesinger: Really minor, but we may want to change the wording in the defaults comment or change this to >=.

brandonocasey: changed to `>=`, good catch.

brandonocasey:
added tests for the `liveui` and `trackingThreshold` options

src/js/live-tracker.js
```diff
@@ -107,9 +113,13 @@ class LiveTracker extends Component {
    * and start/stop tracking accordingly.
    */
   handleDurationchange() {
-    if (this.player_.duration() === Infinity) {
+    if (this.player_.duration() === Infinity && this.liveWindow() >= this.options_.trackingThreshold) {
```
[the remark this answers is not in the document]
gkatsev: Answer is yes. Just verified it with a live stream that has ~5 minutes of a live window.

test/unit/live-tracker.test.js
```diff
@@ -4,12 +4,62 @@ import {createTimeRanges} from '../../src/js/utils/time-ranges.js';
...
+      this.player.dispose();
+      this.clock.restore();
+    }
+  });
+
+  QUnit.test('liveui true, trackingThreshold is met', function(assert) {
+    this.player = TestHelpers.makePlayer({liveui: true});
+    this.player.seekable = () => createTimeRanges(0, 30);
+
+    this.player.duration('Infinity');
```
gesinger: Should these be `Infinity` or does the string work?

PR #6410, opened 2020-01-15
fix: trigger change events on remoteTextTrack when nativeTextTrack is set to true

gkatsev:
It seems we have never triggered `change` events on `remoteTextTrack` when we were using native text tracks. This was a problem for VHS, which exclusively uses text tracks.
This makes it so we do trigger the event. Main issue with this change is that it creates a potential for a false positive where a change event was triggered from a non-remote text track but the remoteTextTrack list still received a change event. This issue is mitigated by best practices of looping through the list looking for the modes that you care about.

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -11,7 +11,7 @@ import window from 'global/window';
 import {assign} from '../utils/obj';
 import mergeOptions from '../utils/merge-options.js';
 import {toTitleCase} from '../utils/string-cases.js';
-import {NORMAL as TRACK_TYPES} from '../tracks/track-types';
+import {NORMAL as TRACK_TYPES, REMOTE} from '../tracks/track-types';
 import setupSourceset from './setup-sourceset';
 import defineLazyProperty from '../utils/define-lazy-property.js';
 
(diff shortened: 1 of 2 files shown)


src/js/tech/html5.js
```diff
@@ -273,13 +273,26 @@ class Html5 extends Tech {
       return;
     }
     const listeners = {
-      change(e) {
-        techTracks.trigger({
+      change: (e) => {
```
gkatsev: this was changed to an arrow function so that line 294 can be used as is without requiring to get the remoteText getter in the higher scope as we only need it in one usecase.

src/js/tech/html5.js
```diff
@@ -273,13 +273,26 @@ class Html5 extends Tech {
...
           type: 'change',
           target: techTracks,
           currentTarget: techTracks,
           srcElement: techTracks
-        });
+        };
+
+        techTracks.trigger(event);
+
+        // if we are a text track change event, we should also notify the
```
gesinger: May be worth adding a note about the alternative approach in case there's ever time to implement it, if it's still seen as a benefit.

gkatsev: The comment is really long. Maybe I'll short it to with a link here and elaborate a bit more here.

gkatsev:
Makes sense to add a test.

mbetz08:
Friendly ping- what's the status of this PR?  Is it awaiting the addition of tests?

third774:
A test has been added for this!

third774:
Looks like the tests are failing now in CI (firefox & IE)

gkatsev:
Looks like firefox doesn't like the way that we're loading the track. I even changed it to explicitly set the mode to showing on `addtrack` and we're still failing in firefox.

gkatsev:
Firefox is refusing to add the track to the player in the test.

gkatsev:
running the same code outside of the test works but fails in the test 🤔

gkatsev:
It seems like everything should be working with changes I have locally but for whatever reason even though the video element has the track element added to it, it seems like things aren't linked properly when in the test.

gkatsev:
I got the test passing once and got excited so when to simply what I had written and now can't get it working again.

gkatsev:
@third774 if you have some time to take a look at why firefox isn't working, that would be appreciated, but I don't think it should be a blocker anymore. If you do figure it out after this has been merged, we can always have a separate PR.

third774:
@gkatsev if you say so! Thanks for all the help! I'll poke at it a bit and let you know if I make any progress.

gkatsev:
https://github.com/videojs/http-streaming/pull/759 depends on this PR, though, they can be merged separately.

PR #6413, opened 2020-01-16
fix: Only update seekBar during pause when live

brandonocasey:
## Description
During live playback the seek bar only has to be updated when the player is paused, because we will always be the same period away from live while playing. 

## Specific Changes proposed
* Start updating the seekbar for the first time on duration change.
* If the duration is Infinity only update when paused otherwise update how we do now.
* Make sure visibility change only enables the interval if it was enabled before losing focus.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -10,6 +10,7 @@ import formatTime from '../../utils/format-time.js';
 import {silencePromise} from '../../utils/promise';
 import keycode from 'keycode';
 import document from 'global/document';
+import clamp from '../../utils/clamp.js';
 
 import './load-progress-bar.js';
 import './play-progress-bar.js';
@@ -156,7 +157,7 @@ class SeekBar extends Slider {
    * attributes to whatever is passed in.
    *
    * @param {EventTarget~Event} [event]
-   *        The `timeupdate` or `ended` event that caused this to run.
+   *        The event that caused this to run.
    *
    * @listens Player#timeupdate
    *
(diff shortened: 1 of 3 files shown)


src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -40,62 +40,92 @@ class SeekBar extends Slider {
    */
   constructor(player, options) {
     super(player, options);
-    this.setEventHandlers_();
+    this.update_ = Fn.bind(this, this.update);
```
brandonocasey: Was kind of silly to rebind these in a helper, and we need them in both `doEvent` functions.

gkatsev:
This change makes sense over-all but it sort of introduces the same bug that https://github.com/videojs/video.js/pull/6445 fixed where the current time tooltip doesn't update but it does if you move the mouse which looks a bit weird since when you just hover nothing changes but then if you move the mouse a bit the current time tooltip could update.
Tested via https://deploy-preview-6413--videojs-docs.netlify.com/test-example/liveui.html.
Maybe we should always update the tooltips but we can stop everything else? Maybe that change is fine? It's just a second or two. Maybe this is something that should be fixed/handled by VHS?

gkatsev:
Looks like maybe the issue is further reduced in #6405, so, may be less of a problem with both changes in.

brandonocasey:
We can revisit his one once the other pull request is merged I would think.

brandonocasey:
Going to close this, I don't think that it's worth doing.

Issue #6273, opened 2019-10-16
Version 7.6.1+ breaks seeking on Smart TVs


sheriffmarley:
## Description
Till today we used version 7.6.0 of videojs and upgraded today to 7.6.5. Now a issue occured, that seeking won't work anymore on Smart TVs, e.g. Samsung. 
I tracked it down to this commit https://github.com/videojs/video.js/commit/f2aedb7 in version 7.6.1 where the if logic has changed. I reverted the changes and the seeking is working on my Smart TV. 

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Open https://videojs.com/advanced/#disneys-oceans on a Smart TV
2. Start playback and try seeking

### Expected
I expect the seeking to work

### Actual
Seeking isn't working

### Error output
-

## Additional Information
Please include any additional information necessary here. Including the following:
### versions
7.6.0
7.6.1+
#### videojs
7.6.1 and up
#### browsers
Samsung Smart TV Browser
#### OSes
Tizen
### plugins
-

osamaalnuimi:
also not working on PS4 browser

gjanblaszczyk:
You are in luck, I have a Samsung smart tv at home and I also can reproduce the problem. I will try to fix this problem soon.

sheriffmarley:
Any update in this´?

PR #6414, opened 2020-01-17
Fix: seeking is broken on smart TV.

gjanblaszczyk:
## Description
A proposed fix for #6273 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -791,10 +791,11 @@ export function isSingleLeftClick(event) {
     return true;
   }
 
-  // `mouseup` event on a single left click has
-  // `button` and `buttons` equal to 0
-  if (event.type === 'mouseup' && event.button === 0 &&
-      event.buttons === 0) {
+  // `mouseup` event on a single left click and
+  // `mousedown` event on pressing a button on the remote controller (Samsung TV)
+  //  have `button` and `buttons` equal to 0
+  if ((event.type === 'mouseup' || event.type === 'mousedown') &&
+      event.button === 0 && event.buttons === 0) {
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


src/js/utils/dom.js
```diff
@@ -791,10 +791,11 @@ export function isSingleLeftClick(event) {
...
   }
 
-  // `mouseup` event on a single left click has
-  // `button` and `buttons` equal to 0
-  if (event.type === 'mouseup' && event.button === 0 &&
-      event.buttons === 0) {
+  // `mouseup` event on a single left click and
+  // `mousedown` event on pressing a button on the remote controller (Samsung TV)
+  //  have `button` and `buttons` equal to 0
+  if ((event.type === 'mouseup' || event.type === 'mousedown') &&
```
gkatsev: Wonder if we should be adding a `click` type here too.
Also, whether this affects regular browsers and whether they will now preferentially treat `mousedown` as the click event rather than `mouseup`. I guess it depends on how `isSingleLeftClick` is used specifically. But it could have potential implications on accessibility.

gjanblaszczyk: Yes, this is a 'dirty' hack that works but complicates the code and can introduce side effects on some exotic devices. I think it will be better to revert this code changes inside the 'isSingleLeftClick' function and do the change inside the seekbar component itself. Does it sound better for you?

gkatsev: yeah, that does sound like a better approach. Keep hacks as contained as possible 🙂

gjanblaszczyk: I have just updated the PR. Does it look better now?

PR #6434, opened 2020-02-06
test: Add test for remote text tracks firing change events when using native text tracks

third774:
This PR adds a test for https://github.com/videojs/video.js/pull/6410

The test does *not* pass without the change in https://github.com/videojs/video.js/pull/6410

### DIFF
--- a/test/unit/tracks/text-tracks.test.js
+++ b/test/unit/tracks/text-tracks.test.js
@@ -476,6 +476,31 @@ QUnit.test('should uniformly create html track element when adding text track', 
   player.dispose();
 });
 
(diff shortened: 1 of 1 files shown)


gkatsev:
Thanks for adding a test! Change seems reasonable, though, we try to avoid asynchronous tests. Going to see if there's an easy way to do so and I'll make a PR against your page.

gkatsev:
So looked into this. Turns out I forgot we were dealing with the native text tracks here. I'm not sure there's a reasonable way of making it work beyond making the test dependent on the implementation. Given that the native text tracks are asynchronous, it makes sense to leave it asynchronous in that case.

gkatsev:
Thanks @third774!

PR #6435, opened 2020-02-08
fix(lang): Add missing Arabic translations

misteroneill:
Updates Arabic translations. Some appear to be incomplete, but this how they were returned from our translation service.


lang/ar.json
```diff
@@ -30,5 +30,58 @@
...
+  "Raised": "بارز",
+  "Depressed": "منخفض",
+  "Uniform": "منتظم",
+  "Dropshadow": "ظل خلفي",
+  "Font Family": "عائلة الخطوط",
+  "Proportional Sans-Serif": "Proportional Sans-Serif",
+  "Monospace Sans-Serif": "Monospace Sans-Serif",
+  "Proportional Serif": "Proportional Serif",
+  "Monospace Serif": "Monospace Serif",
+  "Casual": "Casual",
```
mister-ben: Are these left untranslated because they don't relate to Arabic typography?

lang/ar.json
```diff
@@ -30,5 +30,58 @@
...
   ", opens subtitles settings dialog": ", تفتح نافذة  خيارات الترجمة",
-  ", selected": ", مختار"
+  ", selected": ", مختار",
+  "Audio Player": "مشغل الصوت",
+  "Video Player": "مشغل الفيديو",
+  "Replay": "إعادة التشغيل",
+  "Seek to live, currently behind live": "ذهاب إلى نقطة البث المباشر، متأخر عن البث المباشر حاليًا",
+  "Seek to live, currently playing live": "ذهاب إلى نقطة البث المباشر، البث المباشر قيد التشغيل حاليًا",
+  "Progress Bar": "شريط التقدم",
+  "progress bar timing: currentTime={1} duration={2}": "توقيت شريط التقدم: currentTime={1} duration={2}",
```
gkatsev: This line is incorrectly translated. The translation should read something like `{1} of {2}` or something.

lang/ar.json
misteroneill: That's how they came back from the translation service.

lang/ar.json
misteroneill: Oh, right. 🤔 

I wonder if we should just do `{1} / {2}` for now while we work on getting an updated translation?

gkatsev: Might be better to just leave it off for now.

PR #6445, opened 2020-02-13
fix: current time tooltip does not update

marcosmx:
## Description
The progress bar tooltip get stuck and don't update the time as the video plays. It only respond to mouse movements and reflects the time in the current time tool tip.

## Specific Changes proposed
Add a reference for the playProgressBar component at seekBar to trigger the update action on every playing and timeupdate event


marcosmx:
@misteroneill 
I didn't notice about the presence of property class `bar`, that's was the reason for not used but I changed the code to use it and validate the null state of it.

Issue #6436, opened 2020-02-10
Focus does not loop inside the player at the full screen mode


nightpig:
## Description
Focus does not loop inside the player at the full screen mode.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Go to https://videojs.com/advanced?video=disneys-oceans
2. Activate the full screen mode at the video on the page (via Firefox or Safari) 
3. Move the focus by tab key

### Expected
The focus should be trapped inside the player.

### Actual
The focus leaves the player and move to other elements at the page.

## Additional Information
Chrome doesn't have this issue.

gkatsev:
From https://github.com/videojs/video.js/pull/6494#issuecomment-604111796:

We probably want to do something like what we do in the modal dialog where we get a list of focusable elements and then if we're on the last element and tab is pressed, we set focus on the first element and if we're focused on the first element and shift-tab is pressed we focus on the last focusable element. https://github.com/videojs/video.js/blob/2997b808d03e1e2b27eac017e8b98baefde054f1/src/js/modal-dialog.js#L442-L509

This would need to be done in the player to account for any focusable elements that are inside the player div, which would include the control bar but anything else that plugins include.

Also, I wonder if we'd want to include some kind of element that becomes available to SRs in fullscreen that says that the player is in fullscreen and can be closed via Esc key or fullscreen toggle like we do in the modal dialog https://github.com/videojs/video.js/blob/2997b808d03e1e2b27eac017e8b98baefde054f1/src/js/modal-dialog.js#L138-L144.

gkatsev:
Also, it sounds like maybe it's time to add a feature to Video.js to be able to announce things to fullscreen users? Is going fullscreen something that should be specifically announced to screen reader users?
/cc @OwenEdwards

mjrox:
I can see the same issue on Safari, by following these steps:

- Launch player
- Press tab multiple times
- The focus goes from volume slider to scrubber - bypassing the play button, mute button, picture-in-picture button, and full-screen button.

Are there any plans to resolve this? I would like my player to accessible to a wide an audience as possible.

PR #6494, opened 2020-02-24
FIX | 6436 | focus trap in fullscreen mode

alankyshum:
## Description
This is an attempt to fix the focus trap issue (https://github.com/videojs/video.js/issues/6436)
When the fullscreen toggle is in focus in fullscreen mode, pressing `Tab` key on the keyboard will focus element outside the element, i.e. there was no `focus-trap`.

## Specific Changes proposed
* When the player is in fullscreen mode and focus is currently on the fullscreen toggle and the user presses the `Tab` key
  * Programmatically focus on the PlayToggle in the ControlBar
* When the player is in fullscreen mode and focus is currently on the play toggle and the user presses the `Shift+Tab` key
  * Programmatically focus on the FullscreenToggle in the ControlBar

## Requirements Checklist
- [x] Bug fixed
  - [ ] Docs/guides updated

| Mode | Demo |
| --- | --- |
| Normal mode |  <br />`Tab` on the fullscreen toggle moves focus to the next element |
| Fullscreen mode | fullscreenmode  <br />`Tab` on the fullscreen toggle moves focus to the PlayToggle|

### DIFF
--- a/src/js/control-bar/fullscreen-toggle.js
+++ b/src/js/control-bar/fullscreen-toggle.js
@@ -68,12 +68,15 @@ class FullscreenToggle extends Button {
    * @listens keydown
    */
   handleKeyDown(event) {
-    // If not Fullscreen mode and Tab press the 'button'
-    if (!(this.player_.isFullscreen() && keycode.isEventKey(event, 'Tab'))) {
+    const isTabKey = !event.shiftKey && keycode.isEventKey(event, 'Tab');
+
+    if (!(this.player_.isFullscreen() && isTabKey)) {
       return;
     }
+
     event.preventDefault();
 
+    // If not Fullscreen mode and Tab press the 'button'
     // Trap the keyboard focus by rewinding back to the play toggle button in the player
     const cb = this.player_.getChild('controlBar');
     const playToggle = cb && cb.getChild('playToggle');
--- a/src/js/control-bar/fullscreen-toggle.js
+++ b/src/js/control-bar/fullscreen-toggle.js
@@ -4,6 +4,7 @@
 import Button from '../button.js';
 import Component from '../component.js';
 import document from 'global/document';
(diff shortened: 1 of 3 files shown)


gkatsev:
I'll have to think about it some more but I'm not sure this is the way we'd want to solve this, assuming we want to trap focus. While the fullscreen control and play toggle are the two default controls that are on the edges of the control bar, they aren't necessarily going to be available and there may be other elements. In addition, there are plugins that introduce other focusable elements inside the player element that we'd want to be able to focus on.

alankyshum:
What do you think if cycle back to the first element in the control bar instead of the player toggle, and ends with the last element instead of the fullscreen toggle? That might make the focus behaviour more generic and allow other plugins to add more elements to the control bar.

gkatsev:
We probably want to do something like what we do in the modal dialog where we get a list of focusable elements and then if we're on the last element and tab is pressed, we set focus on the first element and if we're focused on the first element and shift-tab is pressed we focus on the last focusable element. https://github.com/videojs/video.js/blob/2997b808d03e1e2b27eac017e8b98baefde054f1/src/js/modal-dialog.js#L442-L509

This would need to be done in the player to account for any focusable elements that are inside the player div, which would include the control bar but anything else that plugins include.

Also, I wonder if we'd want to include some kind of element that becomes available to SRs in fullscreen that says that the player is in fullscreen and can be closed via Esc key or fullscreen toggle like we do in the modal dialog https://github.com/videojs/video.js/blob/2997b808d03e1e2b27eac017e8b98baefde054f1/src/js/modal-dialog.js#L138-L144.

Going to close this because as is, we can't accept the PR but going to also repost this comment in https://github.com/videojs/video.js/issues/6436 so it doesn't get lost.

I also want to add that we really appreciate you bringing this up and getting started on this. Thanks!

PR #6507, opened 2020-03-05
feat: Improve currentTime to allow be used before player is ready to play

marcosmx:
## Description
This change allow the player to seek into a specific time before the player starts to play on all the browsers.

## Specific Changes proposed
Add some checks to detect whenever the player is changing the source or is not ready yet to process the seek attempt and stores in cache the given value in seconds to be applied once the player is able to perform a seek

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2417,13 +2417,14 @@ class Player extends Component {
       if (seconds < 0) {
         seconds = 0;
       }
-      if (!this.isReady_ || this.changingSrc_ || !this.tech_ || !this.tech_.isReady_ && seconds > 0) {
+      if (!this.isReady_ || this.changingSrc_ || !this.tech_ || !this.tech_.isReady_) {
         this.cache_.initTime = seconds;
         this.off('canplay', this.applyInitTime_);
         this.one('canplay', this.applyInitTime_);
         return;
       }
       this.techCall_('setCurrentTime', seconds);
+      this.cache_.initTime = 0;
       return;
     }
 
@@ -2443,7 +2444,7 @@ class Player extends Component {
    * @private
    */
   applyInitTime_() {
-    this.techCall_('setCurrentTime', this.cache_.initTime);
+    this.currentTime(this.cache_.initTime);
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2145,6 +2145,7 @@ class Player extends Component {
       // we set it to zero here to ensure that if we do start actually caching
       // it, we reset it along with everything else.
       currentTime: 0,
+      initTime: 0,
       inactivityTimeout: this.options_.inactivityTimeout,
       duration: NaN,
       lastVolume: 1,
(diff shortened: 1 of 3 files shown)


marcosmx:
I'll add tests later..feel free to let all your comments about the proposed changes

src/js/player.js
```diff
@@ -2430,6 +2437,15 @@ class Player extends Component {
     return this.cache_.currentTime;
   }
 
+  /**
+   * Apply the value of initTime stored in cache as currentTime.
+   *
+   * @private
+   */
+  applyInitTime_() {
+    this.techCall_('setCurrentTime', this.cache_.initTime);
```
marcosmx: Maybe I should restore the value to zero after set the currentTime

src/js/player.js
```diff
@@ -2416,6 +2417,12 @@ class Player extends Component {
       if (seconds < 0) {
         seconds = 0;
       }
+      if (!this.isReady_ || this.changingSrc_ || !this.tech_ || !this.tech_.isReady_ && seconds > 0) {
```
gkatsev: We may want to not do a check for seconds because in a live stream, if you were to seek to zero, we'd want to then start at the beginning of the DVR window rather than at the live window.

src/js/player.js
```diff
@@ -2416,6 +2417,12 @@ class Player extends Component {
...
+        this.cache_.initTime = seconds;
+        this.off('canplay', this.applyInitTime_);
+        this.one('canplay', this.applyInitTime_);
```
gkatsev: My one concern with canplay event is its reliability, since we want to seek to the given time as early as possible to ensure that no playback happens before we can seek. It's possible that this isn't really an issue anymore in modern browsers but I think it used to be an issue, particularly on iOS, where canplay only happens on play, so, we may end up seeing some content from the beginning of the video before we end up seeking. I'll have to test it out and see how it is.

src/js/player.js
gkatsev: One alternative is to call `this.currentTime(this.cache_.initTime)`, then, there if get to the techCall_ line, we can unset (set back to 0) `initTime`, that way all the seeking related logic is always in `currentTime()`. Thoughts?

marcosmx: agree, that way the for the `applyInitTime_`  method his only concern will be to get the time in the cache to call again the original logic at `currentTime`

src/js/player.js
marcosmx: got it, I missed out that scenario...thanks for notice that!

src/js/player.js
marcosmx: In my tests the `canplay` event was fired even when the play action was not triggered, this happens on all the browsers, but I forgot to test it on mobile especially on iOS

gkatsev: yeah, it's probably much improved than the last time I was involved this type of seeking: https://github.com/brightcove/videojs-urlparams/blob/master/videojs.urlparams.js#L80-L126

test/unit/player-play.test.js
```diff
@@ -98,3 +104,26 @@ QUnit.test('tech ready + has source + changing source = wait for loadstart', fun
   this.player.trigger('loadstart');
   assert.strictEqual(this.techPlayCallCount, 1, 'tech_.play was called');
 });
+
+QUnit.test('tech not ready +  currentTime =  wait until canplay event to apply currentTime value', function(assert) {
```
gkatsev: I don't think that `player-play.test.js` is the right location for these tests. This file is around various tests cases around calling `play()`. Probably best to move it to `player.test.js`

marcosmx: Done.

PR #6519, opened 2020-03-11
feat: Add a function for getting descendants from components

brandonocasey:
## Description
I think I find myself having to be defensive about Component children of children a lot. Specifically for control bar components. For instance, to get the `mouseTimeDisplay` we have to do something like this:

```js
const controlBar = player.getChild('ControlBar');

if (!controlBar) {
  return;
}
const progressControl = controlBar.getChild('ProgressControl');

if (!progressControl) {
  return;
}

const seekBar = progressControl.getChild('SeekBar');

if (!seekBar) {
  return;
}
const mouseTimeDisplay = seekBar.getChild('MouseTimeDisplay');

if (!mouseTimeDisplay) {
  return;
}
```

I think it would be much better if we had a built in way to be defensive about getting descendants instead of having all this code. This is the API that I propose:
```js
const mouseTimeDisplay = player.getDescendant(
  'ControlBar',
  'ProgressControl',
  'SeekBar', 
  'MouseTimeDisplay'
);

if (!mouseTimeDisplay) {
  return;
}
```

Or alternatively with an array argument.

```js 
const mouseTimeDisplay = player.getDescendant([
  'ControlBar',
  'ProgressControl',
  'SeekBar', 
  'MouseTimeDisplay'
]);

if (!mouseTimeDisplay) {
  return;
}
```

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -417,6 +417,37 @@ class Component {
   }
 
   /**
+   * Returns the descendant `Component` following the givent
+   * descendant `names`. For instance ['foo', 'bar', 'baz'] would
+   * try to get 'foo' on the current component, 'bar' on the 'foo'
+   * component and 'baz' on the 'bar' component and return undefined
+   * if any of those don't exist.
+   *
+   * @param {...string[]|...string} names
+   *        The name of the child `Component` to get.
+   *
+   * @return {Component|undefined}
+   *         The descendant `Component` following the given descendant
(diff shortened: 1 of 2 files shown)


gkatsev:
Nice, yeah, we definitely needed to be defensive about this everywhere.

What are your thoughts about overloading `getChild`? If a single string is given it'll give direct child, if an array is given, it'll do what getDescendant does?

PR #6525, opened 2020-03-13
feat: support fastSeek during scrubbing if available

gkatsev:
This can be tested on Safari with a HLS playlist containing iframes, for example https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_adv_example_hevc/master.m3u8

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2394,6 +2394,7 @@ class Player extends Component {
       return this.scrubbing_;
     }
     this.scrubbing_ = !!isScrubbing;
+    this.techCall_('setScrubbing', this.scrubbing_);
 
     if (isScrubbing) {
       this.addClass('vjs-scrubbing');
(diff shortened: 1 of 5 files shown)


gkatsev:
Should add some doccomments and maybe a test or two as well.

alex-barstow:
What exactly is the behavior I should be observing in Safari? I'm not able to discern any difference between using `fastSeek()` and `currentTime()`.

gkatsev:
You need to use an HLS playlist with iframe playlists. When it's available, using fastSeek will switch to using iframes during scrubbing, i.e., when you click and drag on the progress bar. This should allow for thumbnail-like behavior during seeking and should help load from the final point faster without downloading as much data as during regular scrubbing.
This will also open us up to be able to implement this feature in VHS.

alex-barstow:
I see an iframe playlist loaded, but the scrubbing doesn't appear any smoother or faster than non-iframe scrubbing.

gkatsev:
Yeah, I'm not really sure how well it works but it would still download less data during scrubbing, so, as long as it's not worse, should be good.

PR #6376, opened 2019-12-26
feat(cors): add cors options to enable cookies on requests

citosid:
fixes #5438, #4851

## Description

Adds ability to send credentials in the XHR requests.

If the video the user needs to play is in a different domain and it is protected with a session, we need to configure the player so it will forward cookies along with the request, otherwise, the user will receive a 403.

This happens, for example, when playing videos from Akamai:

https://community.akamai.com/customers/s/question/0D50f00005Rtq3sCAB/media-services-on-demand-hls-manifests-are-different-for-firefox?language=en_US

## Specific Changes proposed

- Add crossorigin to the options object

## Requirements Checklist

- [x] Feature implemented / Bug fixed
- [x] If necessary, more likely in a feature request than a bug fix
  - [x] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
- [ ] Reviewed by Two Core Contributors

PR #6533, opened 2020-03-19
 feat(cors): add cors options to enable cookies on requests

sodabrew:
## Description

This is a rebase of #6376 with whitespace changes removed. All credit due to @citosid, whose original committer information is retained.

## Specific Changes proposed

Expose getter/setter for the player's `crossOrigin` property, and add `crossorigin` to the list of internal properties that are copied into the HTML5 video elements. This enables resources like subtitles to include cookies in their XHR requests for sites that require authorization to fetch.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -962,7 +962,7 @@ class Player extends Component {
    * A getter/setter for the `Player`'s aspect ratio.
    *
    * @param {string} [ratio]
-   *        The value to set the `Player's aspect ratio to.
+   *        The value to set the `Player`'s aspect ratio to.
    *
    * @return {string|undefined}
    *         - The current aspect ratio of the `Player` when getting.
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -784,10 +784,16 @@ class Player extends Component {
   }
 
   /**
-   * A getter/setter for the `Player`'s crossOrigin. Returns the player's configured value.
+   * Get or set the `Player`'s crossOrigin option. For the HTML5 player, this
+   * sets the `crossorigin` attribute on the `<video>` tag to control the CORS
+   * behavior.
+   *
+   * @see [Video Element Attributes]{@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-crossorigin}
    *
    * @param {string} [value]
-   *        The value to set the `Player`'s crossorigin to.
+   *        The value to set the `Player`'s crossorigin to. If an argument is
+   *        given, must be one of `anonymous` or `use-credentials`.
+   *
    *
    * @return {string}
    *         The current crossorigin value of the `Player`.
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4,35 +4,35 @@
 // Subclasses Component
 import Component from './component.js';
 
-import {version} from '../../package.json';
+import { version } from '../../package.json';
 import document from 'global/document';
 import window from 'global/window';
 import evented from './mixins/evented';
-import {isEvented, addEventedCallback} from './mixins/evented';
+import { isEvented, addEventedCallback } from './mixins/evented';
 import * as Events from './utils/events.js';
 import * as Dom from './utils/dom.js';
 import * as Fn from './utils/fn.js';
 import * as Guid from './utils/guid.js';
 import * as browser from './utils/browser.js';
-import {IE_VERSION, IS_CHROME, IS_WINDOWS} from './utils/browser.js';
+import { IE_VERSION, IS_CHROME, IS_WINDOWS } from './utils/browser.js';
 import log, { createLogger } from './utils/log.js';
-import {toTitleCase, titleCaseEquals} from './utils/string-cases.js';
+import { toTitleCase, titleCaseEquals } from './utils/string-cases.js';
 import { createTimeRange } from './utils/time-ranges.js';
 import { bufferedPercent } from './utils/buffer.js';
 import * as stylesheet from './utils/stylesheet.js';
 import FullscreenApi from './fullscreen-api.js';
 import MediaError from './media-error.js';
 import safeParseTuple from 'safe-json-parse/tuple';
-import {assign} from './utils/obj';
+import { assign } from './utils/obj';
 import mergeOptions from './utils/merge-options.js';
-import {silencePromise, isPromise} from './utils/promise';
+import { silencePromise, isPromise } from './utils/promise';
 import textTrackConverter from './tracks/text-track-list-converter.js';
 import ModalDialog from './modal-dialog';
 import Tech from './tech/tech.js';
 import * as middleware from './tech/middleware.js';
-import {ALL as TRACK_TYPES} from './tracks/track-types';
+import { ALL as TRACK_TYPES } from './tracks/track-types';
 import filterSource from './utils/filter-source';
-import {getMimetype, findMimetype} from './utils/mimetypes';
+import { getMimetype, findMimetype } from './utils/mimetypes';
 import keycode from 'keycode';
 
 // The following imports are used only to ensure that the corresponding modules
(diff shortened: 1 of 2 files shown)


khavishbhundoo:
I really hope that this PR get merged soon @gkatsev as cross origin is necessary when videos are hosted on a subdomain

citosid:
@gkatsev who could review this? @sodabrew did the fixes you asked.

gkatsev:
Going to be working through PRs this week.

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
+   */
+  crossOrigin(value) {
+    if (!value) {
+      return this.techGet_('crossorigin');
+    }
+
+    const validCorsValues = ['anonymous', 'use-credentials'];
+
+    if (validCorsValues.indexOf(value) < 0) {
+      log.error(`Improper value "${value}" supplied for crossorigin`);
```
gkatsev: I think it would be nicer to inform that user of what values are valid.
```suggestion
      log.error(`crossorigin must be "anonymous" or "use-credentials", given "${value}"`);
```

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
+      return;
+    }
+
+    this.techCall_('setCrossorigin', value);
+
+    return this.techGet_('crossorigin');
```
gkatsev: our other setters don't return a value when set, only when get.
```suggestion
    return;
```

src/js/tech/html5.js
```diff
@@ -1850,7 +1864,21 @@ Html5.resetMediaElement = function(el) {
...
+   * different assets/playlists
+   *
+   * @method Html5#setCrossorigin
+   * @param {string} crossorigin
+   *         - anonymous indicates that the media should not sent cookies.
+   *         - use-credentials indicates that the media should sent cookies along the requests.
+   *
+   * @see [Spec]{@link https://html.spec.whatwg.org/#attr-media-crossorigin}
+   */
+  'crossorigin'
```
gkatsev: can `setCrossorigin` be added to the list on line 1777? It's helpful for grepping the source.

src/js/tech/html5.js
```diff
@@ -1750,7 +1750,21 @@ Html5.resetMediaElement = function(el) {
...
```
gkatsev: can `crossorigin` be added to the list on line 1508? It's helpful for grepping the source.

src/js/tech/html5.js
gkatsev:
```suggestion
  'crossOrigin'
```

src/js/tech/html5.js
gkatsev:
```suggestion
  'crossOrigin'
```

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
+    }
+
+    const validCorsValues = ['anonymous', 'use-credentials'];
+
+    if (validCorsValues.indexOf(value) < 0) {
+      log.error(`Improper value "${value}" supplied for crossorigin`);
+      return;
+    }
+
+    this.techCall_('setCrossorigin', value);
```
gkatsev:
```suggestion
    this.techCall_('setCrossOrigin', value);
```

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
+   *        The value to set the `Player`'s crossorigin to. If an argument is
+   *        given, must be one of `anonymous` or `use-credentials`.
+   *
+   *
+   * @return {string}
+   *         The current crossorigin value of the `Player`.
+   */
+  crossOrigin(value) {
+    if (!value) {
+      return this.techGet_('crossorigin');
```
gkatsev:
```suggestion
      return this.techGet_('crossOrigin');
```

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
+    }
+
+    const validCorsValues = ['anonymous', 'use-credentials'];
+
+    if (validCorsValues.indexOf(value) < 0) {
```
gkatsev: probably can just check the two values directly rather than using an array lookup, but either way is fine because it's unlikely to make a huge performance difference. I hope people won't be calling player.crossOrigin 100s of times a second.

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
 
+  /**
+   * Get or set the `Player`'s crossOrigin option. For the HTML5 player, this
+   * sets the `crossorigin` attribute on the `<video>` tag to control the CORS
+   * behavior.
+   *
+   * @see [Video Element Attributes]{@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-crossorigin}
+   *
+   * @param {string} [value]
+   *        The value to set the `Player`'s crossorigin to. If an argument is
```
gkatsev:
```suggestion
   *        The value to set the `Player`'s crossOrigin to. If an argument is
```

src/js/player.js
```diff
@@ -782,6 +783,38 @@ class Player extends Component {
...
+   *        given, must be one of `anonymous` or `use-credentials`.
+   *
+   *
+   * @return {string}
+   *         The current crossorigin value of the `Player`.
```
gkatsev:
```suggestion
   * @return {string|undefined}
   *         - The current crossOrigin value of the `Player` when getting.
   *         - undefined when setting
```

gkatsev:
@sodabrew if you can make the changes in the next day or two, we can get it into the next minor pre-release of Video.js which I hope to be able to release this week. If you can't, since the changes are relatively minor, mostly `crossorigin` to `crossOrigin`, I am happy to make those changes myself to get the PR merged.

sodabrew:
> Some more usage of `crossorigin -> crossOrigin`

Because the attribute is named `crossorigin`, it's not clear to me that camel case is warranted. I did apply the changes directly from the PR review notes (neat feature!) but they're easy enough to back out if we discuss and decide otherwise.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

Are there examples of other HTML attributes that are compound English words being separately camel-cased here?

sodabrew:
`crossoriginTracks` is an existing variable in the codebase and is not camel cased:
 https://github.com/videojs/video.js/blob/da37fab1308117eba062693d6b5c400a10be54d3/src/js/tech/html5.js#L79-L80

sodabrew:
Oh, uh so https://html.spec.whatwg.org/multipage/media.html#location-of-the-media-resource

> The `crossOrigin` IDL attribute must reflect the `crossorigin` content attribute, limited to only known values.

🤦‍♂

gkatsev:
Yeah, the problem is what the attribute in the HTML and the property on the DOM element. In the HTML, it's the `<video crossorigin>` attribute but in the DOM element it's the `videoEl.crossOrigin`. In Video.js we use the property accessor, so, we need to have it as `crossOrigin`.

gkatsev:
Thinking about it, we only need it to be `crossOrigin` inside the tech, the player can have it as `crossorigin`. Do you think it would be preferable? So, users won't need to remember that HTML is weird.

src/js/player.js
```diff
@@ -782,6 +783,39 @@ class Player extends Component {
...
+   */
+  crossOrigin(value) {
+    if (!value) {
+      return this.techGet_('crossOrigin');
+    }
+
+    const validCorsValues = ['anonymous', 'use-credentials'];
+
+    if (validCorsValues.indexOf(value) < 0) {
+      log.error(`crossorigin must be "anonymous" or "use-credentials", given "${value}"`);
```
gkatsev: I don't think this warrants an error but only a warning as we reject bad values and continue without problems.
```suggestion
      log.warn(`crossorigin must be "anonymous" or "use-credentials", given "${value}"`);
```

sodabrew:
> Thoughts on having it be player.crossorigin(...) and crossorigin as an option? Unfortunately, the tech piece will have to continue being crossOrigin.

I completed switching everything other to camel case. I haven't tried comparing the usability of `player.crossorigin` vs. `player.crossOrigin`yet, will try tomorrow.

Please let me know if I missed anything else in the review, however? I'd like to get down to just the question of naming with the rest of the tech items ready to go.

gkatsev:
I think this is more or less good to go.
One idea we had is to just have both names on the player.

gkatsev:
Thanks @sodabrew @citosid!

PR #6546, opened 2020-03-26
localized translation

kslr:
use smoother text


lang/zh-CN.json
```diff
@@ -77,11 +77,11 @@
   "Reset": "重置",
   "restore all settings to the default values": "恢复全部设定至预设值",
   "Done": "完成",
-  "Caption Settings Dialog": "字幕设定视窗",
-  "Beginning of dialog window. Escape will cancel and close the window.": "开始对话视窗。离开会取消及关闭视窗",
-  "End of dialog window.": "结束对话视窗",
-  "Seek to live, currently behind live": "试图直播，当前延时播放",
-  "Seek to live, currently playing live": "试图直播，当前实时播放",
+  "Caption Settings Dialog": "字幕设定窗口",
+  "Beginning of dialog window. Escape will cancel and close the window.": "打开对话窗口。Escape键将取消并关闭对话窗口",
```
gkatsev: Should Escape be translated as `退出键` instead of `Escape键`?

kslr: "退出键" hardly anyone used, commonly use "ESC" "Escape".

So, should I choose a guide vocabulary or a more commonly used word?

gkatsev: I really don't know, just heard elsewhere that the key was also called 退出键 but I definitely don't speak Chinese so will defer to you.

PR #6571, opened 2020-04-06
feat(cors): allow both crossOrigin and crossorigin method and options

gkatsev:
Fixes a bug with the setter and aliases `crossorigin` to `crossOrigin`.

This is a followup from #6533, /cc @sodabrew @citosid


sodabrew:
Thank you!!

src/js/player.js
```diff
@@ -803,7 +804,7 @@ class Player extends Component {
       return this.techGet_('crossOrigin');
     }
 
-    if (value !== 'anonymous' || value !== 'use-credentials') {
```
sodabrew, line 806: 🤦‍♂ thank you.

Issue #3097, opened 2016-02-10
Multiple component instances cannot bind and unbind prototype methods to player events without clobbering each other.


misteroneill:
Two component instances cannot bind the same prototype method(s) to player events without clobbering each other. I discovered this while refactoring a plugin. [Here's a JSBin to demonstrate the issue.](http://jsbin.com/fabopi/edit?html,js,console,output)

There are two ways to work around this:

1. Use `Function.prototype.bind` to create an instance-bound copy of the listener function:

  ```js
  this.listener = this.listener.bind(this);
  this.on(this.player(), 'eventname', this.listener);
  ```
1. Use an anonymous function that calls the prototype method (assuming ES6):
  
  ```js
  this.listener = (e) => Foo.prototype.listener.call(this, e);
  this.on(this.player(), 'eventname', this.listener);
  ```

I tried `videojs.bind` as well, but the GUID magic is part of the problem - it is considered the same function if it has the same GUID; so, as soon as the first instance passes it to `videojs.bind` it is always and forever "the same function" for the purposes of video.js' event-binding behavior. The other part of the problem is that `on` uses the target element for tracking purposes rather than the object doing the binding.

misteroneill:
For what it's worth, I went with option 2 because we've seen a number of cases where our customers were using things like older versions Prototype.js and MooTools that clobbered `Function.prototype.bind`.

I'm not sure this can be solved in a backward-compatible way, but it's something to consider for future releases. Also, it's probably worth documenting this behavior a bit more in-depth because it could be confusing (obviously). :smile:

gkatsev:
One workaround for this is to change the way that the event system is setting the guids on the function and toStringing the functions instead, perhaps with the addition of a timestamp or something. Thus it'll disambiguate between prototype methods. I blame @ljharb for this suggestion.

brandonocasey:
We might consider a major version where we remove `target` event tracking. IE `this.on(<some target>, 'eventName', fn)` on the current object. I don't think we are going to come up with a great solution for this and I think that this aspect of our event system is constantly causing subtle issues.

PR #6576, opened 2020-04-08
fix: update being called on seekbar during dispose

brandonocasey:
Right now `seekBar.update()` is getting called on after it is disposed by `liveedgechange` causing  and error to be thrown about `player` not existing and us trying to call `ended` on it [here](https://github.com/videojs/video.js/blob/master/src/js/control-bar/progress-control/seek-bar.js#L138)

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -92,7 +92,7 @@ class SeekBar extends Slider {
   }
 
   disableInterval_(e) {
-    if (this.player_.liveTracker && this.player_.liveTracker.isLive() && e.type !== 'ended') {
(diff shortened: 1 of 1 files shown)


gkatsev:
Should we also remove the event listeners for other things we are lisetning to like https://github.com/videojs/video.js/blob/5134db8a15ca00343cc4201623a891068a7a1bba/src/js/control-bar/progress-control/seek-bar.js#L55?

brandonocasey:
Yeah let's remove all the event listeners. I am going to open an issue linking back to this pull request. I think we need to investigate why this is happening, and narrow down if it is truly an issue with events or not

gkatsev:
Is #3097/#6583 the reason all the tests are failing right now?

brandonocasey:
Technically, this pr is a work-around to get it working for now. It appears we need to null check event or pass in an empty object for `disableInterval_`. I will do that now and the tests should pass.

PR #6583, opened 2020-04-09
WIP: Expose a flaw in our current evented dispose handling

brandonocasey:
## Description
When one target is disposed that an object is listening to, all target dispose handlers will be removed. See the unit test.

### DIFF
--- a/test/unit/mixins/evented.test.js
+++ b/test/unit/mixins/evented.test.js
@@ -504,3 +504,40 @@ QUnit.test('off() can remove a listener from an array of events on a different t
     target: b.eventBusEl_
   });
 });
+
+QUnit.test('on() with the same function on two different targets should be disposed correctly', function(assert) {
+  const a = this.targets.a = evented({});
+  const b = this.targets.b = evented({});
+  const c = this.targets.c = evented({});
+  let i = 0;
+  const update = () => {
+    i++;
+  };
+
+  a.on(b, 'foo', update);
+  a.on(c, 'bar', update);
+
+  // sanity test with b trigger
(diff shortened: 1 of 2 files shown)


test/unit/mixins/evented.test.js
```diff
@@ -504,3 +504,40 @@ QUnit.test('off() can remove a listener from an array of events on a different t
...
+
+  // sanity test with b trigger
+  b.trigger('foo');
+  assert.strictEqual(i, 1, 'b trigger works');
+
+  // sanity test with c trigger
+  c.trigger('bar');
+  assert.strictEqual(i, 2, 'c trigger works');
+
+  b.trigger('dispose');
```
brandonocasey, line 528: This also causes `removeListenerOnDispose` for `c`'s to be removed for the `update` function. So when `a` disposes we won't run that function and the `foo` listener  will still exist.

brandonocasey: This happens because we set `removeListenerOnDispose`'s `guid` to the `guid` of the initial listener. so all `removeListenerOnDispose`s for the `update` function will have the same `guid` and they will all be removed.

gkatsev:
https://github.com/videojs/video.js/issues/3097

brandonocasey:
I added a test that fails for the code outlined in the issue you linked @gkatsev . I don't know if we can solve this outside of a major version. We have two things we could do with this code.

1. Close this pull request, but we will have something to look back on when we go to solve #3097 
2. Make these tests `skipped` for now and turn them back on when we go to work on #3097

PR #6588, opened 2020-04-10
fix(text-tracks): set withCredentials on XHR if crossOrigin='use-credentials'

### LINKED ISSUE #6587: cors to enable cookies not working for captions
I already have `crossorigin="use-credentials"` in my `<video>` tag and i can stream videos from `test.example.com` just fine with cookies. However within tracks cookies are not being set.

Here is how i set the tracks inside `<video>` 

`<track kind="captions"  src="https://test.example.com/..."   srclang="es-MX" label="Spanish (Mexico)" />`

gkatsev:
Fixes #6587

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -80,6 +80,12 @@ const loadTrack = function(src, track) {
     opts.cors = crossOrigin;
   }
 
+  const withCredentials = track.tech_.crossOrigin() === 'use-credentials';
+
+  if (withCredentials) {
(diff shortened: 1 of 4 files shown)


gkatsev:
Yeah, I guess I should look into adding a test.

skroth:
@gkatsev do you know about when we can expect to see this go out in a release?

gkatsev:
Pre-release will probably be either today or next week with a full release happening about a week or more after.

PR #6603, opened 2020-04-18
Update description of video.js in the package.json file, and add 'hls' keyword

OwenEdwards:
## Description
Just a suggestion; update the description of video.js to better reflect what it is now, rather than what it was. It seems like this change should have happened when v7 was first released, but maybe it makes sense to do it now?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "video.js",
-  "description": "An HTML5 and Flash video player with a common API and skin for both.",
+  "description": "An HTML5 and HLS video player (with optional Flash support) with a common API and skin.",
   "version": "7.8.1",
   "main": "./dist/video.cjs.js",
   "module": "./dist/video.es.js",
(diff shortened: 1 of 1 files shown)


gkatsev:
We also support most of DASH. Maybe we should add HLS and DASH to tags but not specifically call out HLS in the description?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "video.js",
-  "description": "An HTML5 and Flash video player with a common API and skin for both.",
+  "description": "An HTML5 and HLS video player (with optional Flash support) with a common API and skin.",
```
heff, line 3: How about this?

```suggestion
  "description": "An HTML5 video player that supports HLS, DASH, and Flash, with a common API and skin.",
```

OwenEdwards:
@gkatsev and @heff - since HLS (and DASH) support is now in the core of video.js, and Flash support has been pushed out into a plug-in, shouldn't the description make that distinction clear?

gkatsev:
I'm OK just striking Flash from the record given that in a couple of months it won't be a thing.

mister-ben:
And even mentioning Flash just encourages "how do I play RTMP?"

PR #6607, opened 2020-04-21
fix: checks if body has a 'zoom' set and uses it to calculate pointer…

nezriffic:
… position correctly

## Description
If document `body` has set a `zoom` style set, src/utils/dom.js `getPointerPosition` is incorrectly calculating the position, causing improper behaviour of eg. volume control slider.

## Specific Changes proposed
dom.js `getPointerPosition` gets now zoom value and uses it for calculation.
As the `getComputedStyle` used to read the `zoom` style "costs" some cycles - it stores the zoom value for future use.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -592,7 +592,7 @@ export function findPosition(el) {
  */
 export function getElementZoomLevel(el) {
   try {
-    return window.getComputedStyle(el).getPropertyValue('zoom') * 1.0 || 1;
+    return parseFloat(computedStyle(el, 'zoom')) || 1;
   } catch (ignore) {
     log.warn('Can not detect zoom level', el);
     return 1;
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -583,6 +583,23 @@ export function findPosition(el) {
 }
 
 /**
+ * Gets element zoom level (reads it from computed style)
+ *
+ * @param  {Element} el
+ *         Element we get the zoom value from
+ *
+ * @return {number} zoom value
+ */
+export function getElementZoomLevel(el) {
+  try {
+    return window.getComputedStyle(el).getPropertyValue('zoom') * 1.0 || 1;
+  } catch (ignore) {
+    log.warn('Can not detect zoom level', el);
+    return 1;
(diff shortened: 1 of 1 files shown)


src/js/utils/dom.js
```diff
@@ -582,6 +582,23 @@ export function findPosition(el) {
...
+ * Gets element zoom level (reads it from computed style)
+ *
+ * @param  {Element} el
+ *         Element we get the zoom value from
+ *
+ * @return {number} zoom value
+ */
+export function getElementZoomLevel(el) {
+  try {
+    return window.getComputedStyle(el).getPropertyValue('zoom') * 1.0 || 1;
```
gkatsev, line 595: We should use our `computedStyle` helper here:
```suggestion
    return parseFloat(computedStyle(el, 'zoom')) || 1;
```

src/js/utils/dom.js
```diff
@@ -582,6 +582,23 @@ export function findPosition(el) {
...
+  } catch (ignore) {
+    log.warn('Can not detect zoom level', el);
```
gkatsev, line 597: I don't think that we need to log a warning here especially since getPointerPosition is used so often and Firefox doesn't support `zoom`.

src/js/utils/dom.js
```diff
@@ -609,7 +626,13 @@ export function findPosition(el) {
  *         A coordinates object corresponding to the mouse position.
  *
  */
+let zoomLevel;
+
 export function getPointerPosition(el, event) {
+  if (!zoomLevel) {
+    zoomLevel = getElementZoomLevel(document.body);
```
gkatsev, line 633: Shouldn't this get the zoom level of the `el`?

src/js/utils/dom.js
```diff
@@ -609,7 +626,13 @@ export function findPosition(el) {
  *         A coordinates object corresponding to the mouse position.
  *
  */
+let zoomLevel;
+
 export function getPointerPosition(el, event) {
+  if (!zoomLevel) {
```
gkatsev, line 632: isn't it possible for the zoom property to change?

Issue #5937, opened 2019-04-15
High CPU usage after the player stays in background for a while


demetrio812:
## Description
I've implemented an Audio player with VideoJS and we noticed a very high CPU usage after sustained usage in background.

We did some tests and it happens also with the standard demo: https://videojs.com/advanced/

## Steps to reproduce
1. Open Chrome browser and go to the demo page: https://videojs.com/advanced/
2. Open the DevTools
3. Toggle DeviceToolbar in the window and throttle the CPU to Low-end mobile (this step will make it easier to highlight the problem, but the issue is visible also at full CPU)
4. Start playing and put the window in background for 30 minutes
5. Open the window and try to scroll, the page will not respond for some seconds, the more you keep the player in background the more the seconds, eventually will completely block the page

I was able to profile it on my mac with the following procedure:
1-2-3: as above
4. Detach the DevTools from the Window and put the window with the player in background by clicking the yellow round button in the window top bar to minimise it to the dock
5. You should now only have the DevTools window opened, press Cmd+H to hide the browser completely and wait
6. Show Chrome again but clicking on the Chrome icon, you should now see the DevTools window but not the player window that it's still minimised
7. In the devtools window, go to the performances tab and start recording
8. Maximize the player window and wait until the window start reacting again
9. Stop the profiler in the DevTools window and see the results

You can notice a 100% cpu usage for some seconds until it goes back to normal:

<img width="1588" alt="Screenshot 2019-04-15 19 25 12" src="https://user-images.githubusercontent.com/385644/56156972-4a7e7980-5fb6-11e9-9986-317419fe5897.png">

<img width="1592" alt="Screenshot 2019-04-15 19 25 51" src="https://user-images.githubusercontent.com/385644/56156999-58cc9580-5fb6-11e9-9be7-39acf2247b1d.png">

I've tested on multiple Macs (even very fast ones).

Any suggestions?

Thanks,
Dem

gkatsev:
Interesting, I'm a bit surprised that it's happening in more modern browsers as well but we recently fixed a similar bug for IE11 where the tab would just crash due to CPU usage. The fix we implemented isn't IE11 specific, so, it may improve things. It's available in the latest release now.

demetrio812:
[...]

PR #6621, opened 2020-04-28
fix regression: prevent requestAnimationFrame calls from repeatedly being callled when tab is in background

DispatchCommit:
## Description
Fixes: https://github.com/videojs/video.js/issues/5937
Issue originally found and explained by @meikidd, see: https://github.com/videojs/video.js/issues/5937#issuecomment-539442030

This PR fixes a regression introduced in 7.7.0 by this PR: https://github.com/videojs/video.js/pull/6155
This issue had been previously fixed and working in v7.6.6, but was reintroduced when wrapping the component's update call in a `requestAnimationFrame`.

## Specific Changes proposed
Adds requestAnimationFrame (RAF) flag to prevent RAF calls from stacking in background
This component's `update` method is called via a `setInterval` which runs when a tab is in the background.
Since `update` is continuously called, RAF calls are made each tick, but RAF does not fire in the background.
RAF calls stack while tab is in the background and will fire immediately when tab regains focus.
This causes a potentially huge number of RAF calls to attempt to immediately execute when tab regains focus, which can potentially lock up a tab entirely until all queued RAF's finish firing.
To prevent this issue, we can use a variable to track when an RAF call is made, and prevent additional calls from being made until the last queued RAF function executes.

## Requirements Checklist
  - [ ] ~~Unit Tests updated or fixed~~
  - [ ] ~~Docs/guides updated~~
  - [ ] ~~Example created~~

All tests passing.
I have also pulled this code into my project and tested against v7.6.6, v7.7.0, and v7.8.1 and confirmed:
1) bug does *not* exist in v.7.6.6
2) bug *does* exist in v7.7.0 *and* v7.8.1
3) bug does *not* exist in build produced by this PR
4) behavior of PR is nearly identical to performance of v7.6.6 when returning to a background tab.

### DIFF
--- a/src/js/control-bar/progress-control/load-progress-bar.js
+++ b/src/js/control-bar/progress-control/load-progress-bar.js
@@ -28,6 +28,7 @@ class LoadProgressBar extends Component {
   constructor(player, options) {
     super(player, options);
     this.partEls_ = [];
+    this.needsRequestAnimationFrame = true;
     this.on(player, 'progress', this.update);
   }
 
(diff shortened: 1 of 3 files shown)


DispatchCommit:
Upon further performance investigation, I found that `seek-bar.js` also follows this same pattern and is causing issues as well.

You can see here that multiple animation frames are firing in succession originating from the seek-bar now. I'll apply the same fix to the seek-bar component.

PR #6627, opened 2020-05-01
Feat: Add named requestAnimationFrame to prevent performance issues

brandonocasey:
Not sure if we want to fix it like this, but I think this is a problem that we are going to keep having.
Related #6621 

Fixes: #5937
Issue originally found and explained by @meikidd, see: #5937 (comment)

This PR fixes a regression introduced in 7.7.0 by this PR: #6155
This PR is based on the fix by @DispatchCommit in #6621 but with a bit more "magic" 

Closes: #6621

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1523,7 +1523,7 @@ class Component {
       if (this.namedRafs_.has(name)) {
         this.namedRafs_.delete(name);
       }
-    }, false);
+    });
 
     this.namedRafs_.set(name, id);
   }
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1599,6 +1599,9 @@ class Component {
         ['setTimeoutIds_', 'clearTimeout'],
         ['setIntervalIds_', 'clearInterval']
       ].forEach(([idName, cancelName]) => {
+        // for a `Set` key will actually be the value again
+        // so forEach((val, val) =>` but for maps we want to use
+        // the key.
         this[idName].forEach((val, key) => this[cancelName](key));
       });
 
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -13,8 +13,8 @@ import * as Guid from './utils/guid.js';
 import {toTitleCase, toLowerCase} from './utils/string-cases.js';
 import mergeOptions from './utils/merge-options.js';
 import computedStyle from './utils/computed-style';
-import MapSham from './utils/map-sham.js';
-import SetSham from './utils/set-sham.js';
+import Map from './utils/map.js';
+import Set from './utils/set.js';
 
 /**
  * Base class for all UI Components.
@@ -102,10 +102,10 @@ class Component {
     this.childIndex_ = {};
     this.childNameIndex_ = {};
 
-    this.setTimeoutIds_ = window.Set ? new Set() : new SetSham();
-    this.setIntervalIds_ = window.Set ? new Set() : new SetSham();
-    this.rafIds_ = window.Set ? new Set() : new SetSham();
-    this.namedRafs_ = window.Map ? new Map() : new MapSham();
+    this.setTimeoutIds_ = new Set();
+    this.setIntervalIds_ = new Set();
+    this.rafIds_ = new Set();
+    this.namedRafs_ = new Map();
     this.clearingTimersOnDispose_ = false;
 
     // Add any child components in options
@@ -1514,8 +1514,8 @@ class Component {
     if (this.namedRafs_.has(name)) {
       return;
     }
-
     this.clearTimersOnDispose_();
+
     fn = Fn.bind(this, fn);
 
     const id = this.requestAnimationFrame(() => {
@@ -1526,6 +1526,8 @@ class Component {
     });
 
     this.namedRafs_.set(name, id);
+
+    return name;
   }
 
   /**
@@ -1597,7 +1599,7 @@ class Component {
         ['setTimeoutIds_', 'clearTimeout'],
         ['setIntervalIds_', 'clearInterval']
       ].forEach(([idName, cancelName]) => {
-        this[idName].forEach(this[cancelName], this);
+        this[idName].forEach((val, key) => this[cancelName](key));
       });
 
       this.clearingTimersOnDispose_ = false;
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -13,6 +13,8 @@ import * as Guid from './utils/guid.js';
 import {toTitleCase, toLowerCase} from './utils/string-cases.js';
 import mergeOptions from './utils/merge-options.js';
 import computedStyle from './utils/computed-style';
+import MapSham from './utils/map-sham.js';
+import SetSham from './utils/set-sham.js';
 
 /**
  * Base class for all UI Components.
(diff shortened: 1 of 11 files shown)


gkatsev:
Definitely an interesting idea. @DispatchCommit any thoughts on this?

src/js/component.js
```diff
@@ -100,38 +102,10 @@ class Component {
     this.childIndex_ = {};
     this.childNameIndex_ = {};
 
-    let SetSham;
-
-    if (!window.Set) {
```
brandonocasey, line 105: Creating a class on every component creation probably isn't idea, even if we don't go with this pull request we should really consider moving this to it's own file.

DispatchCommit:
I agree that every component does not *need* this, but in the short term this may be a better approach since there are potentially additional components that call RAF infrequently that may trigger this bug. 

E.g. the live indicator may be calling RAF on `liveedgechange` prior to changing it's state.

While those instances may be negligible, it's still preferred to avoid queuing RAF's in all instance. By adding this at the component level it allows for a quick fix (and smaller code change) when found. Personally I disliked how much duplicated code I had when trying to fix each component 1 at a time. Another benefit is that this fix can be easily applied to 3rd party plugins as well.

The only other thought I have is that a new `AnimatedComponent` class could be created that adds this fix, and other components that need this fix can inherit from it, but that may be adding needless complexity.

src/js/component.js
```diff
@@ -1524,6 +1498,51 @@ class Component {
...
+
+    this.clearTimersOnDispose_();
+    fn = Fn.bind(this, fn);
+
+    const id = this.requestAnimationFrame(() => {
+      fn();
+      if (this.namedRafs_.has(name)) {
+        this.namedRafs_.delete(name);
+      }
+    }, false);
```
DispatchCommit, line 1526: Why is `false` being passing to RAF? I'm not aware of a second parameter for RAF.

brandonocasey: good catch, I think I originally had a second parameter being passed to `this.requestAnimationFrame` but refactored it out. I guess I left it's usage here though.

brandonocasey:
I like the idea of `AnimatedComponent` but I think we would need/want to do that in a major version update and remove `requestAnimationFrame` from component entirely.

brandonocasey:
I will work on adding tests for this pull request as I think this may be a good way to go for now.

test/unit/component.test.js
```diff
@@ -932,7 +932,7 @@ QUnit.test('should provide interval methods that automatically get cleared on co
   assert.ok(intervalsFired === 5, 'Interval was cleared when component was disposed');
 });
 
-QUnit.test('should provide *AnimationFrame methods that automatically get cleared on component disposal', function(assert) {
+QUnit.test('should provide a requestAnimationFrame method that is cleared on disposal', function(assert) {
```
brandonocasey, line 935: Since we have a new animation frame method this test was confusing named `*AnimationFrame`.

test/unit/component.test.js
```diff
@@ -1149,6 +1187,115 @@ QUnit.test('setInterval should be canceled on dispose', function(assert) {
...
+  assert.equal(comp.namedRafs_.size, 0, 'we removed a named raf');
+  assert.equal(comp.rafIds_.size, 0, 'we removed a raf id');
+  assert.equal(clearName, setName, 'cancelNamedAnimationFrame was called');
+
+  this.clock.tick(1);
+
+  assert.equal(called, false, 'requestNamedAnimationFrame was never called');
+});
+
+QUnit.test('requestNamedAnimationFrame should only allow one raf of a specific name at a time', function(assert) {
```
brandonocasey, line 1219: A test for the unique functionality that `requestNamedAnimationFrame` provides.

src/js/component.js
```diff
@@ -100,38 +102,10 @@ class Component {
...
-      };
-    }
-
-    this.setTimeoutIds_ = window.Set ? new Set() : new SetSham();
-    this.setIntervalIds_ = window.Set ? new Set() : new SetSham();
-    this.rafIds_ = window.Set ? new Set() : new SetSham();
+    this.setTimeoutIds_ = new Set();
+    this.setIntervalIds_ = new Set();
+    this.rafIds_ = new Set();
+    this.namedRafs_ = new Map();
```
gkatsev, line 108: At first I was wondering where the checks for the built-ins went but I see that we export the native method or our Sham from the `set.js` and `map.js` files which definitely makes it easier to use them, 👍

src/js/component.js
```diff
@@ -1524,6 +1498,53 @@ class Component {
...
+    }
+    this.clearTimersOnDispose_();
+
+    fn = Fn.bind(this, fn);
+
+    const id = this.requestAnimationFrame(() => {
+      fn();
+      if (this.namedRafs_.has(name)) {
+        this.namedRafs_.delete(name);
+      }
```
gkatsev, line 1525: I noticed that rAF has a similar implementation on lines 1491-1494 but it first deletes the id from the rafIds before calling the method. Not sure if the ordering matters much. Calling the function first sounds ok for me.

src/js/component.js
```diff
@@ -1573,11 +1594,12 @@ class Component {
     this.clearingTimersOnDispose_ = true;
     this.one('dispose', () => {
       [
+        ['namedRafs_', 'cancelNamedAnimationFrame'],
         ['rafIds_', 'cancelAnimationFrame'],
         ['setTimeoutIds_', 'clearTimeout'],
         ['setIntervalIds_', 'clearInterval']
       ].forEach(([idName, cancelName]) => {
-        this[idName].forEach(this[cancelName], this);
+        this[idName].forEach((val, key) => this[cancelName](key));
```
gkatsev, line 1602: this is to get it to work with both Map and Sets, right?

test/unit/component.test.js
gkatsev: Do we have a test that verifies `requestAnimationFrame` is cleared on dispose? Since this test seems to have been updated to use `requestNamedAimtionFrame`. Maybe this test wasn't supposed to be updated?

src/js/component.js
brandonocasey: I think we delete it afterwards because we want to make sure that the entire update runs before we try to queue another one. If an update takes awhile we could delete the animation frame and then queue another one before that update finishes.

src/js/component.js
brandonocasey: Yes

test/unit/component.test.js
brandonocasey: I should have mentioned this but this test was duplicated for `requestAnimationFrame` https://github.com/videojs/video.js/blob/master/test/unit/component.test.js#L1013-L1069 so I changed the duplicate into a test for `requestNamedAnimationFrame`

src/js/component.js
brandonocasey: I added assertions to the unit test for this function so that we test for this.

gkatsev: ok, makes sense.

test/unit/component.test.js
gkatsev: Ok, cool, the tests do exist but was kind of confusing to find them.

PR #6644, opened 2020-05-12
fix: addChild with index should allow for children that are elements

mister-ben:
## Description
The fix in #6297 doesn't work where the child to insert before is an element rather than a component, e.g. the video element.

## Specific Changes proposed
Check if the child to insert before is an element, as well as checking if it has an `el_`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -526,8 +526,13 @@ class Component {
       // If inserting before a component, insert before that component's element
       let refNode = null;
 
-      if (this.children_[index + 1] && this.children_[index + 1].el_) {
-        refNode = this.children_[index + 1].el_;
+      if (this.children_[index + 1]) {
+        // Most children are components, but the video tech is an HTML element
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks @mister-ben. I think we'd probably want to release it as a patch against 7.7, 7.8 and master. Thoughts?

mister-ben:
@gkatsev makes sense, i've branches off those tags ready once this is approved.

@evanfarina thanks for testing

evanfarina:
@mister-ben @gkatsev  Any update as far as getting this in as a patch goes?

gkatsev:
@evanfarina sorry, got distracted with other things. Merging and releasing now.

gkatsev:
@evanfarina 7.7.7! is now out with this change. It's also in 7.8.2 if you felt like updating to that release line. It's also in master as of this change.

Thanks for the changes @mister-ben!

PR #6653, opened 2020-05-14
docs: add fluid, vertical video example

RyKilleen:
## Description
Creates an example in the sandbox for fluid videos, specifically a vertical video example.

## Specific Changes proposed

## Requirements Checklist
- [X ] If necessary, more likely in a feature request than a bug fix
  - [ ] Docs/guides updated
  - [X] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))

### DIFF
--- a/sandbox/vertical-video.html.example
+++ b/sandbox/vertical-video.html.example
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>Video.js Sandbox</title>
+  <link href="../dist/video-js.css" rel="stylesheet" type="text/css">
+  <script src="../dist/video.js"></script>
+</head>
+<body>
+  <div style="background-color:#eee; border: 1px solid #777; padding: 10px; margin-bottom: 20px; font-size: .8em; line-height: 1.5em; font-family: Verdana, sans-serif;">
+    <p>You can use /sandbox/ for writing and testing your own code. Nothing in /sandbox/ will get checked into the repo, except files that end in .example (so don't edit or add those files). To get started run `npm start` and open the vertical-volume.html</p>
+    <pre>npm start</pre>
+    <pre>open http://localhost:9999/sandbox/vertical-video.html</pre>
+  </div>
+
+  <!--  To make this truly responsive, we have to constrain the container's width.        -->
+  <!--  Because vertical videos are concerned more with height, we base it off vh units.  -->
+  <!-- calc(100vh * 0.5625) is equivalant to "screen height * (9:16)"                     -->
+ <div style="
+        width: calc(100vh * 0.5625); 
+        max-width: 100%;
+        margin: 0 auto;">
+  <video-js
+    id="vid1"
+    controls
+    preload="auto"
+    class="vjs-fluid"
+    >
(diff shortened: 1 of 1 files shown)


sandbox/vertical-video.html.example
```diff
@@ -0,0 +1,39 @@
...
+        width: calc(100vh * 0.5625); 
+        max-width: 100%;
+        margin: 0 auto;">
+  <video-js
+    id="vid1"
+    controls
+    preload="auto"
+    class="vjs-fluid"
+    >
+    <source src="http://www.exit109.com/~dnn/clips/RW20seconds_1.mp4" type="video/mp4">
```
gkatsev, line 29: is this a source we can rely on?

RyKilleen: That's a valid question, I'm open to any other public test videos! I included based solely on this page here:

http://www.exit109.com/~dnn/clips/VerticalVideo.html

gkatsev: Thanks, it sounds like it's up for usage, though, we may want to rehost the file.
When I look the other day, that page wasn't loading for me.

sandbox/vertical-video.html.example
```diff
@@ -0,0 +1,39 @@
...
+    <pre>open http://localhost:9999/sandbox/vertical-video.html</pre>
+  </div>
+
+  <!--  To make this truly responsive, we have to constrain the container's width.        -->
+  <!--  Because vertical videos are concerned more with height, we base it off vh units.  -->
+  <!-- calc(100vh * 0.5625) is equivalant to "screen height * (9:16)"                     -->
+ <div style="
+        width: calc(100vh * 0.5625); 
+        max-width: 100%;
+        margin: 0 auto;">
```
gkatsev, line 22: I think this outer div can be removed and instead the aspect ratio option be passed to the player.

sandbox/vertical-video.html.example
```diff
@@ -0,0 +1,39 @@
...
+    preload="auto"
+    class="vjs-fluid"
+    >
+    <source src="http://www.exit109.com/~dnn/clips/RW20seconds_1.mp4" type="video/mp4">
+    <p class="vjs-no-js vjs-">To view this video please enable JavaScript, and consider upgrading to a web browser that <a href="http://videojs.com/html5-video-support/" target="_blank">supports HTML5 video</a></p>
+  </video-js>
+</div>
+  <script>
+    var vid = document.getElementById('vid1');
+    var player = videojs(vid, {controlBar: {volumePanel: {inline: false}}});
```
gkatsev, line 35:
```suggestion
    var player = videojs(vid, {
      aspectRatio: '9:16',
      controlBar: 
        {volumePanel: {inline: false}
      }
    });
```

sandbox/vertical-video.html.example
RyKilleen: Does the fluid nature of VideoJS take the screen height into account? That was not my experience with it.

gkatsev: It's possible that a `min-height` or something like that should be added to the player div. But with the aspect ratio it should mostly do the right thing.

PR #6657, opened 2020-05-17
Add support for CAF, FLAC and WAV format

yuyichao:
## Description

Following https://en.wikipedia.org/wiki/HTML5_audio#Supported_audio_coding_formats
and https://hpr.dogphilosophy.net/test/index.php . These are formats that are supported
on some (most?) browsers.

Tested with a patched version of seafile. The flac and wav files correctly plays on
chromium on my Linux laptop and the caf file plays on my iPad.
(Chromium appears to play the caf file just fine but claim to not support it.)

Of course not all browsers support these but according to #5982 and my test result
I assume the browser detection is implemented somewhere else already.

## Specific Changes proposed
See commit message, new extention to mimetype mapping added.

## Requirements Checklist
- [x] Feature implemented
  - [x] Change has been verified in an actual browser (Chomium on Linux and iPad)
  - [ ] Docs/guides updated


yuyichao:
As a side note, the comment about seems to be out-dated. http://hul.harvard.edu/ois/////systems/wax/wax-public-help/mimetypes.htm is not accessable AFAICT. (Redirected back to harvard library home pape...)

yuyichao:
Gentle ping. Seems that there were activities on the repo again (after a quiet week or two).

yuyichao:
Is the `needs: LGTM` there for

> Reviewed by Two Core Contributors

?

PR #6685, opened 2020-06-03
Fix on demand load of default text track

Chocobozzz:
Hi, and thanks for the *load on demand* text track feature.

Just noticed a bug: when you use `addRemoteTextTrack` and set `default` to `true` before/while the player is being initialized, then videojs displays duplicate captions.

That's because https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track.js#L73 is called twice at the same time:
 * In https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track.js#L345 when we use `addRemoteTextTrack`
 * In https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track.js#L242 (called by https://github.com/videojs/video.js/blob/master/src/js/tracks/text-track-display.js#L143 on player initialization)

This PR adds a guard in `loadTrack` to avoid parallel loading of the same track

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -71,6 +71,12 @@ const parseCues = function(srcContent, track) {
  * @private
  */
 const loadTrack = function(src, track) {
+  if (track._loading === true) {
+    return;
+  }
+
+  track._loading = true;
+
   const opts = {
     uri: src
   };
@@ -87,6 +93,8 @@ const loadTrack = function(src, track) {
   }
 
   XHR(opts, Fn.bind(this, function(err, response, responseBody) {
+    track._loading = false;
+
     if (err) {
       return log.error(err, response);
     }


Chocobozzz:
Is there anything wrong with this PR?

Chocobozzz:
Please don't close, this is a bug preventing us to use the *on demand* text track feature

gkatsev:
@Chocobozzz hey, thanks for your patience. Do you have a live sample or a code snippet I can try out to test this?
Locally, I'm trying:
```js
var player = videojs(document.querySelector('video'), {
  html5: {
    nativeAudioTracks: false,
    preloadTextTracks: false, // I've also tried true which should be the default
  }
});

player.addRemoteTextTrack({
  kind: 'captions',
  src: '../docs/examples/shared/example-captions.vtt?foo',
  label: 'English',
  srclang: 'en-us',
  default: true
})
```
I've also wrapped `player.addRemoteTextTrack` in a `player.ready()`. Without it, I actually don't get a track added and with it, I only get 1.

Chocobozzz:
Hi @gkatsev 

I can reproduce using this HTML code:

```
<html>

<head>
  <link href="https://unpkg.com/video.js/dist/video-js.css" rel="stylesheet">
  <meta charset="utf-8">
  <title>Video.js Starter Template</title>
</head>

<body>
  <video id="my_video" class="video-js" controls width="640" height="268" poster="https://vjs.zencdn.net/v/oceans.png">
  </video>
  <script src="https://unpkg.com/video.js/dist/video.js"></script>

  <script>
    var player = videojs(document.querySelector('video'), {
      html5: {
        preloadTextTracks: false
      }
    });

    player.ready(() => {

      player.src({
        type: 'video/mp4',
        src: 'https://vjs.zencdn.net/v/oceans.mp4'
      })

      player.addRemoteTextTrack({
        kind: 'captions',
        id: 'toto',
        src: 'https://raw.githubusercontent.com/Chocobozzz/PeerTube/develop/server/tests/fixtures/subtitle-good2.vtt',
        label: 'English',
        default: true
      }, false)
    })
  </script>
</body>

</html>
```

Since this is a *timing* issue, sometimes it correctly displays 1 subtitle.

brandonocasey (approved):
I was able to reproduce this issue and the fix works.

jdmunro:
I can also confirm the fix works.

Chocobozzz:
https://github.com/videojs/video.js/pull/7021 seems to fix the source of the bug. I think we can close this PR.

gkatsev:
Thanks, I'll take a look at both PRs soon. I haven't forgotten, just busy...

gkatsev:
Yeah, can confirm that both fix the issue. https://github.com/videojs/video.js/pull/7021 does seem like a much simpler fix.
Thank you for taking the time to check this PR and also reviewing the other PR!

PR #6694, opened 2020-06-06
feat: support vtt regions

gkatsev:
This depends on PR https://github.com/videojs/vtt.js/pull/50 and would need a new vttjs release.
The change is fairly simple and stays within the current API constraints of vttjs.
The only change necessary for regions is to not clear the display ourselves but let vttjs handle it now.


src/js/tracks/text-track-display.js
```diff
@@ -243,7 +243,7 @@ class TextTrackDisplay extends Component {
     const tracks = this.player_.textTracks();
     const allowMultipleShowingTracks = this.options_.allowMultipleShowingTracks;
 
-    this.clearDisplay();
+    // this.clearDisplay();
```
gkatsev, line 246: probably should do this in a backwards compatible way, so, if someone is using an older vttjs, captions don't just accumulate.

gkatsev: Looks like vtt.js already removed old things, so, this was never necessary, really. The addition of the else below with clearDisplay is necessary though for when turning all the captions off.
Thus, this line can be removed and we will be backward and forward compatible.

src/js/tracks/text-track-display.js
```diff
@@ -290,6 +288,8 @@ class TextTrackDisplay extends Component {
         this.setAttribute('aria-live', 'assertive');
       }
       this.updateForTrack(descriptionsTrack);
+    } else {
+      this.clearDisplay();
```
gkatsev, line 292: This is necessary to clear the cues whenever we turn off all the captions. This makes it so that this PR is backward and forwards compatible.

Issue #5706, opened 2018-12-23
dispose() gives error: Cannot read property 'classList' of null


DimaM-BeStreet:
## Description
Hey guys, I've been going crazy for a full day over this bug, thinking i must be missing something (i still might be)
basically I get an error every time I use dispose, including this super simple reduced case:
https://codepen.io/dimam-bestreet/pen/GPmoxX

## Steps to reproduce
1. add dynamically a video
2. run .dispose on it
3. look in the chrome (Chromium 65) dev-tools for this error:
VM76237:214 Uncaught TypeError: Cannot read property 'classList' of null
    at VideojsDelegate (<anonymous>:214:46)
    at add_element (<anonymous>:395:17)
    at MutationObserver.<anonymous> (<anonymous>:419:17)

### Expected
no error..

### Actual
error... ^^

### Error output
If there are any errors at all, please include them here.
VM76237:214 Uncaught TypeError: Cannot read property 'classList' of null
    at VideojsDelegate (<anonymous>:214:46)
    at add_element (<anonymous>:395:17)
    at MutationObserver.<anonymous> (<anonymous>:419:17)

## Additional Information
I've checked in firefox - no error!
I am using crome (Chromium 65) on linux (if that has anything to do with anything)

#### videojs
7.3.0
#### browsers
Chromium 65, firefox seems ok
#### OSes
I've checked only mine - linux mint 17.3
### plugins
nope

Is this is only related to the fact im using Chromium 65??

gkatsev:
Unfortunately, I'm unable to reproduce locally or in browserstack on chrome 65 on win10 or macos.

DimaM-BeStreet:
I'll test it on my phone this week, if it works - no problem really...
Thank you!
Dima

thijstriemstra:
I think i've seen this in a unit test before (so don't wait for `ready` event but `dispose` right away or something like that).

armaha:
Hi,
I also notice the same issue, with almost the exact same stack trace for VideoJs 6.6. It also only occurs in Chrome. We block our instance from calling dispose if not ready like this: 
```
if(this.player) {
    this.player.dispose();
}

```
We instantiate the player like this:
```
const nonReadyPlayer = videojs(
                    this.videoId,
                    this.videoConfig,
                    () => {
                        this.player = nonReadyPlayer;
                    }
);
```

[...]

PR #6701, opened 2020-06-09
fix: check element for null in case of a disposed player

travisbader:
## Description
https://github.com/videojs/video.js/issues/5706

## Specific Changes proposed
Fail fast

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -280,6 +280,10 @@ export function addClass(element, classToAdd) {
  *         The DOM element with class name removed.
  */
 export function removeClass(element, classToRemove) {
+  // Protect in case the player gets disposed
+  if (!element) {
+    return null;
+  }
   if (element.classList) {
     element.classList.remove(classToRemove);
   } else {
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -282,6 +282,7 @@ export function addClass(element, classToAdd) {
 export function removeClass(element, classToRemove) {
   // Protect in case the player gets disposed
   if (!element) {
+    log.warn("removeClass was called with an element that doesn't exist");
     return null;
   }
   if (element.classList) {


brandonocasey (commented):
Should we do this in most of the functions in this file?

gkatsev:
It's possible but generally they expose worse errors which are otherwise hidden by this null check.

brandonocasey:
I wonder if we should log a warning in this case then @gkatsev

src/js/utils/dom.js
```diff
@@ -280,6 +280,10 @@ export function addClass(element, classToAdd) {
  *         The DOM element with class name removed.
  */
 export function removeClass(element, classToRemove) {
+  // Protect in case the player gets disposed
+  if (!element) {
```
gkatsev, line 284: Can we add a log here something like
```
log.warn("removeClass was called with an element that doesn't exist");
```

gkatsev (commented):
We spoke about it and while we still have reservations about it, if we add a log for this and do the null check we can merge it in.

DaveStein:
@brandonocasey @gkatsev does this address my errors like this?

```
Uncaught TypeError: Cannot read property 'parentNode' of null
Uncaught TypeError: Cannot read property 'language' of null
Uncaught TypeError: Cannot read property 'classList' of null
Uncaught TypeError: Cannot read property 'classList' of null
Uncaught TypeError: Cannot read property 'classList' of null
```

in minified code I had seen it as coming from this line, when `on` is fired for `ready` event

```
this.readyQueue_=[],e&&e.length>0&&e.forEach((function(e){e.call(this)}),this),this.trigger("ready")}),1)},t.$=function(e,t){return jt(e,t||this.contentEl())},t.$$=function(e,t){return Ft(e,t||this.contentEl())},
```

gkatsev:
@DaveStein could be. Do you happen to have a live test page for this that you can share?

DaveStein:
@gkatsev unfortunately i haven't reproduced this yet myself but just see it in error logs.

gkatsev:
@DaveStein if you try latest `main`, we just merged a fix related to the readyQueue not being cleared out properly. We don't have a release out yet, but hopefully soon.

DaveStein:
Thanks for the update @gkatsev. Since I haven't reproduced this locally, I'll need to wait for the release to be able to update my repo and deploy it. I imagine the worst case scenario for a release update is that my issue won't be fixed. If it does, I'll comment in this thread as well so future people searching Google for the error will see this PR fixes both issues.

gkatsev:
Yes, once we get a release out, please circle back whether it's fixed or not.

Szefczuk:
Hi, when are you planning to merge this pr?

damjan25:
Any plans to merge this in next release ?

src/js/utils/dom.js
gkatsev:
```suggestion
  if (!element) {
    log.warn("removeClass was called with an element that doesn't exist");
```

gkatsev (approved):

brandonocasey (approved):

PR #6719, opened 2020-06-19
test: skip requestPictureInPicture test if API isn't available

gkatsev:
Followup from #6678.


gkatsev:
Merging because it's a small change in tests and will unbreak tests in `main`.

PR #6735, opened 2020-07-01
fix: Fullscreen broken in iOS

### LINKED ISSUE #6707: pause/play button doesn't toggle after leaving fullscreen with playsinline attribute
## Description

When `playsinline` attribute is used video pause/play button stops toggling after manually entering and leaving the fullscreen.

[screen recording](https://user-images.githubusercontent.com/24650518/84501744-b2fede80-acbf-11ea-8572-f1372e0e125f.gif)

In the same time video from [videojs homepage](https://videojs.com/) also has `playsinline` attribute but doesn't have same issue. 

**UPDATE:**
 From [video.js github package.json file](https://github.com/videojs/videojs.com/blob/master/package.json) we can see that they are using older version (7.6.0) for videojs. Experimented with different versions and found that **this bug is introduced with 7.8.0 version**

I'm not particularly strong JS developer so can't find exact problem thus providing the PR for fix but my first guess would be to look into [feat(fs): return a promise from requestFullscreen and exitFullscreen](https://github.com/videojs/video.js/commit/da37fab) commit which is part of that 7.8.0 tag.

## Steps to reproduce
1. Open [starter template](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0)
2. Add `playsinline` attribute.
3. Open on mobile device
4. Press to open fullscreen
5. Leave fullscreen
6. toggle play/pause button. It does play and pause video but button is never updated, it's always the "pause" button. 
7. change videojs version to `@7.7.7` (last version without bug) and repeat 3-5 to see that the bug is not present there.

### Expected
When video is playing "pause" button can be seen. WHen video paused "play" button can be seen. When pressed, it toggles.

### Actual
What ever the state of video, after leaving fullscreen button doesn't toggle.

### Error output
No errors

## Additional Information

This is happening on any IOS mobile device. Haven't tested with android. 

I have managed to also get situations when (after similar open/close fullscreen) duration bar isn't updated as well.

alex-barstow:
## Description
Fixes #6707. The Fullscreen API is unsupported in iOS, so `self.fsApi_.fullscreenerror` and `self.fsApi_.fullscreenchange` are undefined, which was breaking the player after entering fullscreen by [removing all bound player events](https://github.com/videojs/video.js/blob/main/src/js/utils/events.js#L369-L377).

## Specific Additions
Both `requestFullscreen()` and `exitFullscreen()` add handlers for `'fullscreenchange'` and `'fullscreenerror'` events, but remove those handlers for `self.fsApi_.fullscreenchange` and `self.fsApi_.fullscreenerror` events. There may be a reason for this inconsistency, but it seems to me that if `.one()` uses the event string, `.off()` should as well (especially considering the latter two are not always defined).

Another option might be to check for Fullscreen API support as well as Promise support in `requestFullscreen()` and `exitFullscreen()`

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2872,8 +2872,8 @@ class Player extends Component {
 
       return new PromiseClass((resolve, reject) => {
         function offHandler() {
-          self.off(self.fsApi_.fullscreenerror, errorHandler);
-          self.off(self.fsApi_.fullscreenchange, changeHandler);
+          self.off('fullscreenerror', errorHandler);
+          self.off('fullscreenchange', changeHandler);
         }
         function changeHandler() {
           offHandler();
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2794,8 +2794,8 @@ class Player extends Component {
 
       return new PromiseClass((resolve, reject) => {
         function offHandler() {
-          self.off(self.fsApi_.fullscreenerror, errorHandler);
-          self.off(self.fsApi_.fullscreenchange, changeHandler);
+          self.off('fullscreenerror', errorHandler);
+          self.off('fullscreenchange', changeHandler);
         }
         function changeHandler() {
           offHandler();


marguinbc:
Do we still need `self.fsApi_` for non iOS implementations?

mikeswart:
Confirmed working. 👍

PR #6759, opened 2020-07-18
[Fix] common misspelling errors

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines


thijstriemstra (approved):
awesome, thanks!

PR #6765, opened 2020-07-20
fix(tech): add abstract crossOrigin method on Tech

gkatsev:
As part of #6588, we started using the crossOrigin method. However, it's
possible that a tech doesn't support this. Notably, the Flash tech. We
should instead have an abstract method on Tech that returns nothing so
we don't fail on those browsers.

### DIFF
--- a/src/js/tech/tech.js
+++ b/src/js/tech/tech.js
@@ -431,6 +431,25 @@ class Tech extends Component {
   reset() {}
 
   /**
+   * Get the value of `crossOrigin` from the tech.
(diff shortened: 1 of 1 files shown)


gkatsev:
Might be worth looking at and making sure that prefixed/suffixed things are ignored. Though, I think that should be in a separate PR because I want to get this fix out immediately.

gkatsev:
Created https://github.com/videojs/video.js/issues/6766 as a followup

PR #6824, opened 2020-08-31
Add volume tooltip

gjanblaszczyk:
## Description
Add tooltip for volume level so the user knows what they will change too. 
Maybe in the scope of WCAG "1.1.1 Non-text Content".

volume-tooltip

## Specific Changes proposed
Add volume level tooltip.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -138,6 +138,7 @@
   &:before {
     position: absolute;
     font-size: 0.9em; // Doing this to match the handle on play progress.
+    z-index: 1;
   }
 }
 
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -148,6 +148,7 @@
   &:before {
     top: -0.5em;
     left: -0.3em;
+    z-index: 1;
   }
 }
 .vjs-slider-horizontal .vjs-volume-level {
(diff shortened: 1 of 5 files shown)


gkatsev:
Hey, thanks for the PR. Totally forgot to take a look at it.

Two main things that is that we should document and perhaps make it easier to hide the volume tooltip. Also, if the inline:false volume panel is added as the last thing in the control bar, we should potentially have the tooltip show up on the left so that it doesn't go outside the player boundary.
Screen Shot 2020-11-17 at 18 29 15


@OwenEdwards thoughts on this?

brandonocasey:
Having a tooltip is awesome, never realized that we were missing this!

gjanblaszczyk:
Thanks for the review and sorry for the late response. I am planning to update this PR this week.

gkatsev:
No worries @gjanblaszczyk, we appreciate the help. I was actually thinking of maybe merging this in and then having the direction get changed as a separate PR. However, if you think you can get this in soon, we may as well wait.

gjanblaszczyk:
Hi @gkatsev, I am pretty busy right now, so I think it is a good idea to merge this and then to have a separate PR for the improvements. I am planning to back to it as soon as I have some free time.

thijstriemstra:
> I was actually thinking of maybe merging this in and then having the direction get changed as a separate PR.

+1

gkatsev (approved):

gkatsev:
I opened up #7145 to not forget fixing the issue of the volume control being the last item in the control bar.

PR #6828, opened 2020-09-01
[#6436] - Create a helper to get focusable elements and trap focus in fullscreen mode to fix an a11y bug

evanfarina:
## Description
Addresses issue #6436 

## Specific Changes proposed
- Get focusable controlBar elements on first `play`
- When a user tabs on the last focusable controlBar el, set the focus back to the first focusable El in the control bar
- When a user presses Shift + tab on the first focusable controlBar el, set the focus back to the last focusable El in the control bar

## Requirements Checklist
  - [x ] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ x] Unit Tests updated or fixed

### DIFF
--- a/src/js/modal-dialog.js
+++ b/src/js/modal-dialog.js
@@ -3,7 +3,6 @@
  */
 import * as Dom from './utils/dom';
 import Component from './component';
-import window from 'global/window';
 import document from 'global/document';
 import keycode from 'keycode';
 
@@ -493,17 +492,7 @@ class ModalDialog extends Component {
     const allChildren = this.el_.querySelectorAll('*');
 
     return Array.prototype.filter.call(allChildren, (child) => {
-      return ((child instanceof window.HTMLAnchorElement ||
-               child instanceof window.HTMLAreaElement) && child.hasAttribute('href')) ||
-             ((child instanceof window.HTMLInputElement ||
-               child instanceof window.HTMLSelectElement ||
-               child instanceof window.HTMLTextAreaElement ||
-               child instanceof window.HTMLButtonElement) && !child.hasAttribute('disabled')) ||
-             (child instanceof window.HTMLIFrameElement ||
-               child instanceof window.HTMLObjectElement ||
-               child instanceof window.HTMLEmbedElement) ||
-             (child.hasAttribute('tabindex') && child.getAttribute('tabindex') !== -1) ||
-             (child.hasAttribute('contenteditable'));
+      return Dom.isFocusable(child);
     });
   }
 }
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2229,6 +2229,10 @@ class Player extends Component {
   }
 
   trapFullscreenTab_(event) {
+    if (!Array.isArray(this._focusableControls) || !this._focusableControls.length) {
+      return;
+    }
+
     // We'll use this to figure out if the user just tabbed on the first or last focusable control
     const focusIndex = this._focusableControls.indexOf(event.target);
     let indexToFocus;
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -545,10 +545,6 @@ class Player extends Component {
     this.reportUserActivity();
 
     this.one('play', this.listenForUserActivity_);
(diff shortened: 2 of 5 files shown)


mister-ben:
There are a bunch of edge cases which will make this more difficult. This wouldn't account for focussable controls which are not children of the controlBar (not present in the vanilla player, but can be added by plugins), or control bar controls added/removed/enabled/disabled since the first play.

evanfarina:
> There are a bunch of edge cases which will make this more difficult. This wouldn't account for focussable controls which are not children of the controlBar (not present in the vanilla player, but can be added by plugins), or control bar controls added/removed/enabled/disabled since the first play.

Thanks @mister-ben . I modified my solution and now account for both of the edge cases you mentioned. A high-level overview of the changes:

- Only calculate focusable controls when fullscreen is requested
- Only calculate first and last focusable control. Previous solution was calculating _all_ focusable controls within the control bar
- Account for focusable controls outside of the control bar (anything in the player)

src/js/player.js
```diff
@@ -2199,6 +2199,91 @@ class Player extends Component {
...
+        elToFocus = this._lastFocusableControl;
+      }
+    } else if (focusedEl === this._lastFocusableControl) {
+      elToFocus = this._firstFocusableControl;
+    }
+    if (elToFocus && typeof elToFocus.focus === 'function') {
+      elToFocus.focus();
+      event.preventDefault();
+    }
+  }
```
gkatsev, line 2285: Looks like there's a lot of overlap here with the `handleKeyDown` method in `modal-dialog`. I wonder if the code can be shared.

src/js/player.js
```diff
@@ -2199,6 +2199,91 @@ class Player extends Component {
...
+    return focusableEl;
+  }
+
+  /**
+   * Given a parent control, finds the last focusable child control
+   *
+   * @param {HTMLElement} parent - The control in which we'll search
+   * @return {HTMLElement | null} - Returns the last focusable control in the node list or null if not focusable control is found
+   */
+  getLastFocusableControl_(parent) {
```
gkatsev, line 2240: The reason for this method and the one above it is if someone added an element to the player DOM but their top level element isn't focusable but some child is, we should return that child element?

src/js/modal-dialog.js
```diff
@@ -493,17 +492,7 @@ class ModalDialog extends Component {
...
-             ((child instanceof window.HTMLInputElement ||
-               child instanceof window.HTMLSelectElement ||
-               child instanceof window.HTMLTextAreaElement ||
-               child instanceof window.HTMLButtonElement) && !child.hasAttribute('disabled')) ||
-             (child instanceof window.HTMLIFrameElement ||
-               child instanceof window.HTMLObjectElement ||
-               child instanceof window.HTMLEmbedElement) ||
-             (child.hasAttribute('tabindex') && child.getAttribute('tabindex') !== -1) ||
-             (child.hasAttribute('contenteditable'));
+      return Dom.isFocusable(child);
```
gkatsev, line 495: I think `Dom.isFocusable` can just be passed directly without wrapping it in the arrow function.

src/js/utils/dom.js
```diff
@@ -181,6 +181,48 @@ export function createEl(tagName = 'div', properties = {}, attributes = {}, cont
...
+ * @param {HTMLElement} control - The control whose visibility is being checked
+ * @return {boolean} - True if the control is visible
+ */
+export function isVisible(control) {
+  if (!control) {
+    return false;
+  }
+
+  // Using `includes` below to catch cases where `!important` is used in the style
+  return !computedStyle(control, 'display').includes('none') && !computedStyle(control, 'visibility').includes('hidden');
```
gkatsev, line 196: Unfortunately, we can't use `.includes` as we still support IE11 and do not require a shim to be included (pun intended).
This can probably be something like
```suggestion
  // Account for usage of `!important
  return !(/^none/).test(computedStyle(control, 'display')) && !(/^hidden/).test(computedStyle(control, 'visibility'));
```

PR #6842, opened 2020-09-11
(feat) language change handler

marcodeltorob:
## Description
This change will trigger `languagechange` event each time `player.language()` is used as a setter. Also, `Component` will call `handleLanguageChange` method to manage the localization behavior.


src/js/player.js
```diff
@@ -4133,6 +4133,7 @@ class Player extends Component {
     }
 
     this.language_ = String(code).toLowerCase();
+    this.trigger('languagechange');
```
gkatsev, line 4136: If the same language was set, we aren't actually changing the language and we shouldn't trigger the event.

gkatsev, line 4136: Also, we want to add an event jsdoc here, see https://github.com/videojs/video.js/blob/cb70ea84a1b3bb02b81af16628fbd617700ca144/src/js/player.js#L572-L578 for an example
and then add a `@fires Player#langaugechange` in the jsdoc comment for the method.

gkatsev (changes requested):
This is a great start!
Looks like a bunch of tests are failing because probably because we started using `.on` in the Component constructor and the tests aren't expecting that.
Also, we should have additional tests to verify that we trigger the event on a language change and that the control text is properly re-localized when the language changes.

src/js/component.js
```diff
@@ -115,6 +115,8 @@ class Component {
     if (options.reportTouchActivity !== false) {
       this.enableTouchActivity();
     }
+
+    this.handleLanguageChange = this.handleLanguageChange.bind(this);
```
brandonocasey, line 119: This is around where the change I mentioned in clickable component should live. I don't think we will need the bind anymore either.

src/js/clickable-component.js
```diff
@@ -132,6 +132,18 @@ class ClickableComponent extends Component {
     }
   }
 
+  /**
+   * Handles language change in ClickableComponent for the player in components
+   *
+   *
+   */
+  handleLanguageChange() {
```
brandonocasey, line 140: I think this should use a lowercase `c` ie: `handleLanguagechange`, as I think that we try to name things `handle<uppercase first letter of event>`

test/unit/clickable-component.test.js
```diff
@@ -93,3 +93,31 @@ QUnit.test('handleClick should not be triggered more than once when enabled', fu
...
+  });
+
+  const testClickableComponent = new ClickableComponent(player);
+
+  testClickableComponent.controlText_ = 'Play';
+  player.language('en');
+
+  const el = testClickableComponent.createEl();
+
+  assert.equal(el.getAttribute('title'), 'Play');
```
brandonocasey, line 119: It would be good to also verify changing the language to `es` here causes the change this to change to `Juego`

test/unit/player.test.js
```diff
@@ -1355,6 +1355,18 @@ QUnit.test('sets lang attribute on player el', function(assert) {
...
 
+QUnit.test('language changed should trigger languagechange event', function(assert) {
+  const player = TestHelpers.makePlayer({});
+
+  assert.expect(1);
+
+  player.on('languagechange', function() {
+    assert.ok(true, 'languagechange event triggered');
+  });
+  player.language('es-MX');
```
brandonocasey, line 1366: Should also add a test to verify that we do not trigger `languagechange` when changing to the same language.

src/js/clickable-component.js
gkatsev: Looks like we're mostly consistent with that but I see we have a `handleTechTimeUpdate_` method, lol.

src/js/clickable-component.js
```diff
@@ -155,6 +167,7 @@ class ClickableComponent extends Component {
       }
       this.on(['tap', 'click'], this.handleClick);
       this.on('keydown', this.handleKeyDown);
+      this.on('languagechange', this.handleLanguageChange);
```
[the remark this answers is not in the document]
gkatsev: I don't think that we need a check for the method because it inherits from Component which has an abstract method for it.
However, why not have this line in Component itself? That way all components are automatically hooked up and just need to implement the `handleLanguagechange` method?

src/js/player.js
```diff
@@ -4132,6 +4133,15 @@ class Player extends Component {
...
+    if (this.language_ !== String(code).toLowerCase() && this.language_ !== undefined) {
+      /**
+      * fires when the player language change
+      *
+      * @fires Player#languagechange
+      * @type {EventTarget~Event}
+      */
+      this.trigger('languagechange');
+    }
     this.language_ = String(code).toLowerCase();
```
gkatsev, line 4145: might as well bring this inside the if statement above.

PR #6851, opened 2020-09-22
docs: Fix simple typo, intial -> initial

timgates42:
There is a small typo in src/js/player.js, src/js/tech/html5.js, test/unit/sourceset.test.js.

Should read `initial` rather than `intial`.


gkatsev (approved):
haha, good find, thanks!

PR #6871, opened 2020-10-03
fix: always set tabIndex to restore keydown a11y

### LINKED ISSUE #6870: skipped tabindex property breaks ModalDialog keydown event
## Description
`ModalDialog` modals can no longer be closed by pressing Escape. This happens because the `<div class=".vjs-modal-dialog">` element no longer gets an (explicit) `tabIndex` property after #6145 and it is thus not focusable.

https://codepen.io/zmousm/pen/bGpXror

## Steps to reproduce

1. Load and open any `ModalDialog`.
2. Press Escape.

### Expected
The modal should close.

### Actual
The modal does not close.

#### videojs
7.7.0 up to latest
#### browsers
any
#### OSes
any

zmousm:
Fixes #6870

## Description
Skipping the `tabIndex` property on created elements due to #6145 optimizations blocks them from receiving keyboard events, due to not
being focus-able; for example this breaks closing `ModalDialog` elements by pressing Escape.
Fix this by always setting `tabIndex`, as the element may return the same value even though the property has not been explicitly set.

Fixes #6870

## Specific Changes proposed
* Always set a `tabIndex` property passed to `Dom.createEl()`.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -165,7 +165,7 @@ export function createEl(tagName = 'div', properties = {}, attributes = {}, cont
     // method for it.
     } else if (propName === 'textContent') {
       textContent(el, val);
-    } else if (el[propName] !== val) {
+    } else if (el[propName] !== val || propName === 'tabIndex') {
       el[propName] = val;
     }
   });


gkatsev (approved):
Thanks! That makes sense. The alternative would be to check whether a tabIndex is actually set but yeah, easier to just always copy it.

brandonocasey (approved):

PR #6891, opened 2020-10-13
feat: trigger languagechange event on a language change

marcodeltorob:
## Description
This change will trigger `languagechange` event each time `player.language()` is used as a setter. Also, `Component` will call `handleLanguageChange` method to manage the localization behavior.

### DIFF
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -178,6 +178,18 @@ class ClickableComponent extends Component {
   }
 
   /**
+   * Handles language change in ClickableComponent for the player in components
+   *
+   *
+   */
+  handleLanguagechange() {
+    if (!this.controlText_) {
+      return;
+    }
+    this.controlText(this.controlText_);
+  }
+
+  /**
    * Event handler that is called when a `ClickableComponent` receives a
    * `click` or `tap` event.
    *
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -183,9 +183,6 @@ class ClickableComponent extends Component {
    *
    */
   handleLanguagechange() {
-    if (!this.controlText_) {
-      return;
-    }
     this.controlText(this.controlText_);
   }
 
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -95,6 +95,9 @@ class Component {
     if (options.evented !== false) {
       // Make this an evented object and use `el_`, if available, as its event bus
       evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+      if (this.player_) {
+        this.on(this.player_, 'languagechange', this.handleLanguagechange);
+      }
     }
     stateful(this, this.constructor.defaultState);
 
@@ -120,6 +123,7 @@ class Component {
     if (options.reportTouchActivity !== false) {
       this.enableTouchActivity();
     }
+
   }
 
   /**
@@ -317,6 +321,13 @@ class Component {
 
     return localizedString;
   }
+
+  /**
+   * Handles language change for the player in components. Should be overriden by sub-components.
+   *
+   * @abstract
+   */
+  handleLanguagechange() {}
 
   /**
    * Return the `Component`s DOM element. This is where children get inserted.
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -95,7 +95,9 @@ class Component {
     if (options.evented !== false) {
       // Make this an evented object and use `el_`, if available, as its event bus
       evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+
       if (this.player_) {
+        this.handleLanguagechange = this.handleLanguagechange.bind(this);
         this.on(this.player_, 'languagechange', this.handleLanguagechange);
       }
     }
@@ -292,6 +294,7 @@ class Component {
    *         The localized string or if no localization exists the english string.
    */
   localize(string, tokens, defaultValue = string) {
+
     const code = this.player_.language && this.player_.language();
     const languages = this.player_.languages && this.player_.languages();
     const language = languages && languages[code];
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -96,10 +96,8 @@ class Component {
       // Make this an evented object and use `el_`, if available, as its event bus
       evented(this, {eventBusKey: this.el_ ? 'el_' : null});
 
-      if (this.player_) {
-        this.handleLanguagechange = this.handleLanguagechange.bind(this);
-        this.on(this.player_, 'languagechange', this.handleLanguagechange);
-      }
+      this.handleLanguagechange = this.handleLanguagechange.bind(this);
+      this.on(this.player_, 'languagechange', this.handleLanguagechange);
     }
     stateful(this, this.constructor.defaultState);
 
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -399,7 +399,7 @@ class Player extends Component {
     this.tagAttributes = tag && Dom.getAttributes(tag);
 
     // Update current language
-    this.language(this.options_.language, true);
+    this.language(this.options_.language);
 
     // Update Supported Languages
     if (options.languages) {
@@ -4280,10 +4280,15 @@ class Player extends Component {
   }
 
   /**
-   * The player's language code
-   * NOTE: The language should be set in the player options if you want the
-   * the controls to be built with a specific language. Changing the language
-   * later will not update controls text.
+   * The player's language code.
+   *
+   * Changing the langauge will trigger
+   * [languagechange]{@link Player#event:languagechange}
+   * which Components can use to update control text.
+   * ClickableComponent will update its control text by default on
+   * [languagechange]{@link Player#event:languagechange}.
+   *
+   * @fires Player#languagechange
    *
    * @param {string} [code]
    *        the language code to set the player to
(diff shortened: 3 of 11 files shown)


src/js/clickable-component.js
```diff
@@ -177,6 +177,18 @@ class ClickableComponent extends Component {
...
 
+  /**
+   * Handles language change in ClickableComponent for the player in components
+   *
+   *
+   */
+  handleLanguagechange() {
+    if (!this.controlText_) {
+      return;
+    }
```
gkatsev, line 188: `controlText()` does a null check for before proceeding, we can probably just call it unconditionally.

src/js/player.js
```diff
@@ -4295,7 +4299,19 @@ class Player extends Component {
       return this.language_;
     }
 
-    this.language_ = String(code).toLowerCase();
```
gkatsev, line 4298: we can always set this unconditionally, since it's the same line in both clauses of the if/else.
However, I think that instead of settng a property, we can add an optional argument to the language method `skipListeners` (like in the dimension method in component https://github.com/videojs/video.js/blob/b6c356d2942b3025b4acf0be0418738da9731a1b/src/js/component.js#L998). Then, in the constructor (line 405) we can call it as `this.language(this.options_.language, true)` and just have an `if (skipListeners)` around the trigger.

test/unit/component.test.js
```diff
@@ -43,15 +45,15 @@ QUnit.module('Component', {
   }
 });
 
-const getFakePlayer = function() {
```
gkatsev, line 46: are the changes in this file necessary?

test/unit/tracks/text-tracks.test.js
```diff
@@ -58,18 +58,7 @@ QUnit.test('Player track methods call the tech', function(assert) {
...
-      calls--;
-    },
-    getChild() {
-      calls--;
-    },
-    ready() {
-      calls++;
-    }
-  }, {});
+  const ttd = new TextTrackDisplay(TestHelpers.makePlayer(), {}, calls++);
```
gkatsev, line 61: By making this change, the test actually becomes a no-op since `calls` never updates as needed via `ready()`, `getChild()` and `addTextTracks()`, instead, it will *always* get incremented here because it isn't wrapped in a function call.
If there's just one new method necessary, we should stub it in the object passed to TextTrackDisplay, otherwise, after making the player, we should replace the methods with new methods that increment and decrement `calls`.

test/unit/poster.test.js
```diff
@@ -10,20 +10,23 @@ QUnit.module('PosterImage', {
...
+    //   },
+    //   handler_: null,
+    //   off() {},
+    //   on(type, handler) {
+    //     this.handler_ = handler;
+    //   },
+    //   trigger(type) {
+    //     this.handler_.call();
+    //   }
+    // };
```
gkatsev, line 26:
```suggestion
```
There's no need to keep old code around if it's no longer needed.

test/unit/player.test.js
```diff
@@ -1385,6 +1385,34 @@ QUnit.test('sets lang attribute on player el', function(assert) {
   }
 });
 
+QUnit.test('language changed should trigger languagechange event', function(assert) {
```
gkatsev, line 1388: I think another test where we change the language multiple times would be good to have. Twice should suffice. Just to make sure that we don't get stuck in the new state.

test/unit/clickable-component.test.js
```diff
@@ -111,3 +111,32 @@ QUnit.test('handleClick should use handler from options', function(assert) {
...
+    }
+  });
+
+  const testClickableComponent = new ClickableComponent(player);
+
+  testClickableComponent.controlText_ = 'Play';
+  const el = testClickableComponent.createEl();
+
+  player.language('en');
+  assert.ok(el.innerHTML.match(/vjs-control-text"?[^<>]*>Play/));
```
gkatsev, line 135: instead of string matching, we should use DOM methods to get the text of the control text.
For example:
```js
el.querySelector('.vjs-control-text').textContent
```
Then we can use `assert.equal` for greater visibility into what this check is doing.

test/unit/component.test.js
marcodeltorob: I did these changes since `new Component()`  is expecting a DOM because of

```
if (this.player_) {
    this.on(this.player_, 'languagechange', this.handleLanguagechange);
}
```

in the constructor and  if we use the previews `getFakePlayer()` function  I'm getting
 `Uncaught (in promise) Error: Invalid target; must be a DOM node or evented object`

src/js/player.js
```diff
@@ -398,8 +398,11 @@ class Player extends Component {
     // Store the tag attributes used to restore html5 element
     this.tagAttributes = tag && Dom.getAttributes(tag);
 
+    //  true if it is the firts time that we set language before evented
+    this.initLang_ = true;
+
```
gkatsev, line 403:
```suggestion
```
this is no longer used.

gkatsev (commented):
code looks good, I'll take it for a spin before we merge it in.

gkatsev:
so, in testing this, I found that the items in the control bar that are added by default, like the PlayToggle, don't update when the language changes. To test I did the following:
```js
videojs.addLanguage('foo', { 'Play': 'BarFoo'});
player.language('foo');
```
Then, when I hover over the PlayToggle, I can see that it will always show `Play`. However, if I add a new play toggle via `player.controlBar.addChild('playToggle')` it will show the new language and will toggle(!) between the `Play` and `BarFoo` as I switch the language between `en` and `foo`.

I think we need to either figure that out in this PR or split this PR up into two with the changes to Component and ClickableComponent going into the other PR and this PR just being the triggering of the event.

marcodeltorob:
@gkatsev  I did the same test and if I  click pause and hover over PlayTogle I will see `play` then If I add the language and change it to `foo`

```js
videojs.addLanguage('foo', { 'Play': 'BarFoo'});
player.language('foo');
```
and then click on play and pause I will see the new `BarFoo` I did not add the play toggle via `addChild` so my question is this is the expected or should I need re-render the PlayToggle, since with your test I needed to do the same click `play` and `pause` to see the new `title`

gkatsev:
The component should update without requiring us to click on it. The PlayToggle has a click action but it's possible for some components to not include such an action and therefore will never get updated.

misteroneill:
I see basically what Gary sees. Here's what I did:

1. Hit BPB to start playback
1. Pause
1. Run `videojs.addLanguage('foo', { 'Play': 'BarFoo'}); player.language('foo');`
1. *Hover over the Play button, tooltip says "Play".* This is the issue, it should update to "Play" immediately upon changing the language.
1. Click the Play button
1. Click the Pause button
1. Note that the Play button will now be updated to say "BarFoo" in its tooltip.

src/js/player.js
```diff
@@ -4290,12 +4291,24 @@ class Player extends Component {
...
+  language(code, skipListeners) {
     if (code === undefined) {
       return this.language_;
     }
 
-    this.language_ = String(code).toLowerCase();
+    if (this.language_ !== String(code).toLowerCase()) {
+      this.language_ = String(code).toLowerCase();
+      // skipListeners helps to avoid trigger on initialization  when component is not evented
+      if (!skipListeners) {
```
misteroneill, line 4302: I don't think I understand the need for this... if the component is not evented, it will not have a `languagechange` listener, so there should be no harm done by always triggering the event.

misteroneill (changes requested):

src/js/player.js
marcodeltorob: We want this because in the first call when we initialized the language the component is not evented when this happens and will throw a `trigger` function is not defined.

src/js/component.js
```diff
@@ -95,7 +95,9 @@ class Component {
     if (options.evented !== false) {
       // Make this an evented object and use `el_`, if available, as its event bus
       evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+
       if (this.player_) {
+        this.handleLanguagechange = this.handleLanguagechange.bind(this);
```
marcodeltorob, line 100: With this change, we create a new function signature for the component that overloads `handleLanguagechange` and we bind it to it; The problem was that every component shares the handler stored in the player and if one removes the handler it will be removed for every component.

gkatsev: Classic 😆

src/js/player.js
misteroneill: Gotcha, I get it now! 👍

misteroneill (approved):

src/js/component.js
```diff
@@ -95,6 +95,11 @@ class Component {
     if (options.evented !== false) {
       // Make this an evented object and use `el_`, if available, as its event bus
       evented(this, {eventBusKey: this.el_ ? 'el_' : null});
+
+      if (this.player_) {
+        this.handleLanguagechange = this.handleLanguagechange.bind(this);
+        this.on(this.player_, 'languagechange', this.handleLanguagechange);
+      }
```
gkatsev, line 102: I don't think that this check is necessary. We make sure that a `this.player_` property is always set on lines 57-61. Removing it locally, it works and tests pass.
```suggestion
      this.handleLanguagechange = this.handleLanguagechange.bind(this);
      this.on(this.player_, 'languagechange', this.handleLanguagechange);
```

gkatsev (commented):
Nice find with the change handler. That's actually a common issue we see with these types of things.
Tested it and looks good, just realized that the `player_` check in Component.js isn't necessary.

gkatsev (approved):

gkatsev:
Weird that tests failed. They passed locally, I'll take a look at it if you don't get around to it. Likely tests need to be modified a bit for the handleLanguageChange change.

src/js/player.js
```diff
@@ -4290,12 +4291,24 @@ class Player extends Component {
...
+        /**
+        * fires when the player language change
+        *
+        * @fires Player#languagechange
+        * @type {EventTarget~Event}
+        */
+        this.trigger('languagechange');
```
brandonocasey, line 4309: Would it make more sense to check if the component is evented or has a trigger function at this point rather than having a `skipListeners` argument?

gkatsev: yeah, that makes sense. Didn't even think of that. We have `skipListeners` elsewhere, which is why I brought it up but `isEvented` definitely makes sense here.

gkatsev: PR for it https://github.com/marcodeltorob/video.js/pull/2

PR #6914, opened 2020-10-26
fix(menu): focus correct MenuItem on keyboard open

### LINKED ISSUE #6912: Keyboard focus isn't set to the first MenuItem if the Menu has a title
## Description
If a Menu component has a title, keyboard focus doesn't get set to the first MenuItem when the menu is triggered via keyboard.

Reduced test case: https://codepen.io/acmertz/pen/oNLZyzK

## Steps to reproduce
1. Start playing a video that has chapters (the default MenuButton for chapters includes a title in its Menu).
2. Tab over to the chapters MenuButton and trigger it via Space, Enter, or the Up/Down arrow keys.
3. Try to navigate through the MenuItems via the Up/Down arrow keys, and see that focus doesn't move between menu items.

### Expected
Up/Down arrow keys move focus between different menu items.

### Actual
Nothing happens - focus cannot be moved onto any of the MenuItems via either the Up/Down arrow keys or the Tab key.

### Error output
None

## Additional Information
When a Menu _doesn't_ have a title, everything works as expected and focus is set to the first MenuItem when the Menu is invoked via the keyboard. The issue only occurs when the Menu has a title.

The code checks for the presence of an HTML element representing the title by getting the first MenuItem and comparing its `className` to the one used for a title element. However, the check always resolves to `false` because all the MenuItem children are Video.js components and therefore don't have a `className` property.

https://github.com/videojs/video.js/blob/1cbea93b3ef16d5a900e165f3fa153771d0c3cdf/src/js/menu/menu.js#L262-L263

This appears to be left over from #3612 (#5722), when title switched from using a plain HTML element to using a Video.js component.

#### videojs
7.10.1

#### browsers
Chrome, Firefox

#### OSes
Windows

### plugins
None

acmertz:
Closes #6912

## Description
Updates the `focus()` method of the Menu component to fix a bug with a CSS `className` comparison that expected the title element to be a standard DOM element instead of a Video.js Component.

## Specific Changes proposed
When `focus()` is called on a Menu, use `Component.hasClass()` to determine if the first child is the title.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/menu/menu.js
+++ b/src/js/menu/menu.js
@@ -259,8 +259,7 @@ class Menu extends Component {
    */
   focus(item = 0) {
     const children = this.children().slice();
-    const haveTitle = children.length && children[0].className &&
-      (/vjs-menu-title/).test(children[0].className);
+    const haveTitle = children.length && children[0].hasClass('vjs-menu-title');
 
     if (haveTitle) {
       children.shift();


gkatsev (approved):
Thanks!
I think the reason it was like this is because a long time ago the title element wasn't a Component.

brandonocasey (approved):
Good Catch @acmertz

PR #6928, opened 2020-11-04
feat(track): make label property mutable and fire a labelchange event when the label is changed

claudiah12:
## Description
Allows the editing of a track's label after it's creation. Track buttons will listen for the labelchange event and update their content accordingly. This aligns videojs behavior with the [HTML label spec](https://html.spec.whatwg.org/multipage/media.html#text-track-label) linked in the docs, which says that "The label of a track can change dynamically".

## Specific Changes proposed
- Add a setter for label property of the Track & fire labelchange event when the label is changed
- When tracks are added to TrackList, add a listener for a labelchange event on the track to trigger a labelchange event on the parent TrackList
- TrackButton listens for labelchange event on the TrackList, and will call the existing update handler to update the menu when the event is triggered.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/track.js
+++ b/src/js/tracks/track.js
@@ -42,9 +42,10 @@ class Track extends EventTarget {
     const trackProps = {
       id: options.id || 'vjs_track_' + Guid.newGUID(),
       kind: options.kind || '',
-      label: options.label || '',
       language: options.language || ''
     };
+
+    let label = options.label || '';
 
     /**
      * @memberof Track
@@ -59,15 +60,6 @@ class Track extends EventTarget {
      * @memberof Track
      * @member {string} kind
      *         The kind of track that this is. Cannot be changed after creation.
-     * @instance
-     *
-     * @readonly
-     */
-
-    /**
-     * @memberof Track
-     * @member {string} label
-     *         The label of this track. Cannot be changed after creation.
      * @instance
      *
      * @readonly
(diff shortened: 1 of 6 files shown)


gkatsev:
So, I took a look at the spec against and did notice that it calls for the `label` to be [readonly](https://html.spec.whatwg.org/multipage/media.html#texttrack). When I previously checked I didn't see that. That said, I think this is a reasonable change.

One thing I noticed when testing is that there was an error being logged:
```
Uncaught TypeError: track.addEventListener is not a function
    at VideoTrackList.addTrack (video.js:6619)
    at VideoTrackList.addTrack (video.js:6935)
    at VideoTrackList.addtrack (video.js:21634)
```
It was coming from this line https://github.com/videojs/video.js/pull/6928/files#diff-35e16d0093cf9dc3d26d160d4d0d8e132309a1223f4ea06633e274c259871690R96. The reason is that Chrome now exposes Audio and Video tracks on the media element, however, AudioTrack and VideoTrack are not EventTargets according to the spec (https://html.spec.whatwg.org/multipage/media.html#audiotracklist-and-videotracklist-objects) while a TextTrack *is*. I made a PR that fixes that issue: https://github.com/claudiah12/video.js/pull/1

claudiah12:
@gkatsev Thanks so much for reviewing and adding that fix! I merged it in to the pr.

src/js/tracks/track.js
```diff
@@ -91,6 +83,33 @@ class Track extends EventTarget {
...
+
+        /**
+         * An event that fires when label changes on this track.
+         *
+         * > Note: This is not part of the spec!
+         *
+         * @event Track#labelchange
+         * @type {EventTarget~Event}
+         */
+        this.trigger('labelchange');
```
brandonocasey, line 110: Should we only trigger this label actually changed? ie:
```js
if (label !== newLabel) { 
  label = newLabel;
  this.trigger('labelchange')
}
```

brandonocasey (commented):
one question code lgtm other than that.

src/js/tracks/track.js
gkatsev: Yeah, this makes sense.

claudiah12: Good call- added a check for this!

gkatsev (approved):
Thanks for the quick response @claudiah12!

brandonocasey (approved):

PR #6943, opened 2020-11-11
fix: cast TOUCH_ENABLED to boolean

third774:
## Description
Ran into an issue where we were using a configuration that looks something like:

```
{
  nativeControlsForTouch: true,
  html5: {
    nativeTextTracks: browser.TOUCH_ENABLED
  }
}
```

and we were not seeing captions appear in the native controls in Chrome on Microsoft Surface devices. After some digging, we found that with that browser/device combination, `TOUCH_ENABLED` is a number, not a boolean. And [the `nativeTextTracks` logic specifically does strict equality checks against `true` and `false` values](https://github.com/videojs/video.js/blob/694fe0f787f6b9b4edb06e6c87b91c50a8429237/src/js/tech/tech.js#L134-L138).

## Specific Changes proposed
`TOUCH_ENABLED` should ALWAYS be a boolean. This PR casts the expression.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -169,10 +169,10 @@ export const IS_WINDOWS = (/Windows/i).test(USER_AGENT);
  * @const
  * @type {Boolean}
  */
-export const TOUCH_ENABLED = Dom.isReal() && (
+export const TOUCH_ENABLED = Boolean(Dom.isReal() && (
   'ontouchstart' in window ||
   window.navigator.maxTouchPoints ||
-  window.DocumentTouch && window.document instanceof window.DocumentTouch);
+  window.DocumentTouch && window.document instanceof window.DocumentTouch));
 
 /**
  * Whether or not this device is an iPad.


src/js/utils/browser.js
```diff
@@ -169,10 +169,10 @@ export const IS_WINDOWS = (/Windows/i).test(USER_AGENT);
  * @const
  * @type {Boolean}
  */
-export const TOUCH_ENABLED = Dom.isReal() && (
+export const TOUCH_ENABLED = !!(Dom.isReal() && (
   'ontouchstart' in window ||
   window.navigator.maxTouchPoints ||
```
third774, line 174: I suspect `window.navigator.maxTouchPoints` is where the number comes from.

src/js/utils/browser.js
```diff
@@ -169,10 +169,10 @@ export const IS_WINDOWS = (/Windows/i).test(USER_AGENT);
  * @const
  * @type {Boolean}
  */
-export const TOUCH_ENABLED = Dom.isReal() && (
+export const TOUCH_ENABLED = !!(Dom.isReal() && (
```
brandonocasey, line 172: Can you change this to `Boolean(...)`. I know its the same thing, but it's much easier to understand.

gkatsev (approved):

brandonocasey (approved):

PR #6948, opened 2020-11-18
feat: migrate async timeout/interval/animation fns to a mixin

brandonocasey:
## Description
This moves `setTimeout`, `setInterval`, `requestAnimationFrame`, and `requestNamedAnimationFrame` functions into a mixin so that we can re-use them in an `async-event-target` class. This would be very beneficial for VHS as we re-implement a lot of this code on our own after extending event target and we cannot just extend component.


src/js/async-event-target.js
```diff
@@ -0,0 +1,33 @@
+import asyncTimers from './mixins/async-timers.js';
+import EventTarget from './event-target.js';
+
+class AsyncEventTarget extends EventTarget {
+  constructor(options = {}) {
+    super();
+    asyncTimers(this);
+
+    if (options.disposeWith) {
```
brandonocasey, line 9: I am unsure wether we should have something like this, but we do have automatic disposal on player dispose for a lot of our regular components. This seemed like a good idea as it allows an AsyncEventTarget to be bound to the lifecycle of something else.

src/js/component.js
```diff
@@ -178,9 +174,6 @@ class Component {
         this.el_.parentNode.removeChild(this.el_);
       }
 
-      if (DomData.has(this.el_)) {
```
brandonocasey, line 181: This is now a part of the evented mixin dispose handler.

src/js/mixins/evented.js
```diff
@@ -499,6 +500,11 @@ function evented(target, options = {}) {
   // When any evented object is disposed, it removes all its listeners.
   target.on('dispose', () => {
     target.off();
+    [target, target.el_, target.eventBusEl_].forEach(function(val) {
```
brandonocasey, line 503: dispose of any and all dom data here, previously we did this in component.js, but I think we need it for any evented mixin object.

PR #6967, opened 2020-12-01
fix: clear readyQueue with dispose

brandonocasey:
## Description
Seems like we never clear the internal `readyQueue_` when calling `dispose`  which may cause the issues seen in https://github.com/videojs/video.js/pull/6701#issuecomment-736046505 and #5706

### DIFF
--- a/src/js/mixins/evented.js
+++ b/src/js/mixins/evented.js
@@ -9,26 +9,6 @@ import * as Fn from '../utils/fn';
 import * as Obj from '../utils/obj';
 import EventTarget from '../event-target';
 
-const objName = (obj) => {
-  if (typeof obj.name === 'function') {
-    return obj.name();
-  }
-
-  if (typeof obj.name === 'string') {
-    return obj.name;
-  }
-
-  if (obj.name_) {
-    return obj.name_;
-  }
-
-  if (obj.constructor && obj.constructor.name) {
-    return obj.constructor.name;
-  }
-
-  return typeof obj;
-};
-
 /**
  * Returns whether or not an object has had the evented mixin applied.
  *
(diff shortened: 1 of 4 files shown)


gkatsev (approved):

brandonocasey:
I updated this pull request slightly when adding a unit test I realized that this case will also throw an error when it happens about `nodeName` since we don't check for null in `validateTarget`. The error was also too vague for my liking but we can back that part of the changes out if we think that it's too much for this pull request.

src/js/mixins/evented.js
```diff
@@ -68,9 +88,9 @@ const isValidEventType = (type) =>
  * @param  {Object} target
  *         The object to test.
  */
-const validateTarget = (target) => {
-  if (!target.nodeName && !isEvented(target)) {
-    throw new Error('Invalid target; must be a DOM node or evented object.');
+const validateTarget = (target, obj, fnName) => {
```
brandonocasey, line 91: Made these errors more specific.

src/js/mixins/evented.js
```diff
@@ -68,9 +88,9 @@ const isValidEventType = (type) =>
...
+  if (!target || (!target.nodeName && !isEvented(target))) {
```
brandonocasey, line 92: Check if target is null before trying to check a property on it.

src/js/mixins/evented.js
```diff
@@ -398,6 +418,8 @@ const EventedMixin = {
    *          Whether or not the default behavior was prevented.
    */
   trigger(event, hash) {
+    validateTarget(this.eventBusEl_, this, 'trigger');
```
brandonocasey, line 421: validate on trigger, this is likely the cause of the issue we see in `dom`

gkatsev:
If it can be split out into two PRs, that would be nice. The changes look good.

brandonocasey:
Moved the evented changes to https://github.com/videojs/video.js/pull/6982

gkatsev (approved):

Issue #6949, opened 2020-11-19
Focus Returned to Captions Menu instead of Caption Settings Menu Item when closing Caption Settings


josephshanak:
## Description
During an accessibility audit of a project I work on that uses VideoJS we received the following feedback about how the focus is handled after closing the Caption Settings menu.

> Users will expect to be returned to the item that activated the dialog when they close the dialog.
> When a user closes the dialog, focus should be moved back to the menu item for "caption settings".

It seems like the behavior expected should have been implemented as part of #2746

> Keyboard focus does not return to the control which opened the dialog when it is closed

## Steps to reproduce
1. Go to https://videojs.com/advanced/?video=elephantsdream
2. Tab to Captions menu
3. Enter Caption Settings
4. Tab to Close button and activate

## Results
Focus is returned to Captions Menu on the control bar

### Expected
Focus is returned to "Caption Settings" menu item.

### Actual
Focus is returned to Captions Menu on the control bar

### Error output
N/A

## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
Happens on 7.7.5 and 7.10.2 at least

#### browsers
Firefox, Chrome. Possibly others but untested.

#### OSes
Verified on MacBook Pro running MacOS Catalina 10.15.7

### plugins
Unsure

gkatsev:
I think we return the captions to the menu because the menu ends up getting closed. It does seem like maybe the menu should get re-opened and focus brought back into the menu.
@OwenEdwards any thoughts on this?

kannapples:
I created a PR for this issue. The fix I came up with isn't as modular as I'd like it to be - its specific to the caption settings modal. This means if another modal is added, it wont automatically include this accessibility fix. I couldn't figure out a way to handle this gracefully, but I'm happy to make updates if you have any suggestions.

OwenEdwards:
@gkatsev there is some disagreement in accessibility circles about the expected behavior in this situation. I'm getting some clarification, but I don't think that #6971 should be merged until I have a more definite answer (also, #6971 isn't robust for different cases when other menus can open the Captions Settings Dialog)

kannapples:
@OwenEdwards were you able to dig up any clarification about the expected behavior here?

OwenEdwards:
Confirmed that this is acceptable behavior for focus movement when closing the Captions Setting Dialog.

PR #6971, opened 2020-12-02
Re-select the caps button after closing modal

kannapples:
## Description
During an accessibility audit of a project I work on that uses VideoJS we received the following feedback about how the focus is handled after closing the Caption Settings menu.

> Users will expect to be returned to the item that activated the dialog when they close the dialog.
> When a user closes the dialog, focus should be moved back to the menu item for "caption settings".
> Keyboard focus does not return to the control which opened the dialog when it is closed.

Fixes [#6949](https://github.com/videojs/video.js/issues/6949)

## Specific Changes proposed
When the TextTrackSettings modal is closed, the SubsCapsButton will be re-selected, and the captions menu list will be re-opened. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/text-track-controls/subs-caps-button.js
+++ b/src/js/control-bar/text-track-controls/subs-caps-button.js
@@ -59,6 +59,12 @@ class SubsCapsButton extends TextTrackButton {
     return items;
   }
 
+  /**
+   * Re-open the caption list after modal close to promote accessibility
+   */
+  reopenCaptionListAfterModalClose() {
+    this.pressButton();
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


brandonocasey (changes requested):
I see what you mean by this and what it fixes. I don't think we want to do it this way though. I think we would be better suited adding a `modalclose` listener in `handleClick` in `CaptionSettingsMenuItem` that calls `pressButton` on it's parent. We would need to do it this way because there are many types of captions menus.

gkatsev:
@OwenEdwards is looking into whether this is something we want. We'll keep this PR open until we have more info from him.
https://github.com/videojs/video.js/issues/6949#issuecomment-738906559

OwenEdwards:
See #6949 - the current behavior is acceptable, so this fix is not needed.

PR #6700, opened 2020-06-09
fix: change time-display to not crash after dispose

travisbader:
## Description
Please describe the change as necessary.
If it's a feature or enhancement please be as detailed as possible.
If it's a bug fix, please link the issue that it fixes or describe the bug in as much detail.
https://github.com/videojs/video.js/issues/6699

## Specific Changes proposed
Check the view hierarchy before replacing to avoid the crash

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [x] Change has been verified in an actual browser (Chome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
- [ ] Reviewed by Two Core Contributors

PR #6977, opened 2020-12-06
fix(control-bar): dont crash after dispose

kontrollanten:
## Description
Fixed the change requested in #6700. Thanks to @travisbader

### DIFF
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -5,6 +5,7 @@ import document from 'global/document';
 import Component from '../../component.js';
 import * as Dom from '../../utils/dom.js';
 import formatTime from '../../utils/format-time.js';
+import log from '../../utils/log.js';
 
 /**
  * Displays time information about the video
@@ -86,7 +87,13 @@ class TimeDisplay extends Component {
         return;
       }
 
-      const oldNode = this.textNode_;
+      let oldNode = this.textNode_;
+
+      if (oldNode && !this.contentEl_.contains(oldNode)) {
+        oldNode = null;
+
+        log.warn('Prevented to execute \'replaceChild\' on \'Node\' since the node to be replaced was not a child of this node.');
+      }
 
       this.textNode_ = document.createTextNode(this.formattedTime_);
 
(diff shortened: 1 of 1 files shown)


src/js/control-bar/time-controls/time-display.js
```diff
@@ -86,7 +87,13 @@ class TimeDisplay extends Component {
         return;
       }
 
-      const oldNode = this.textNode_;
+      let oldNode = this.textNode_;
+
+      if (oldNode && !this.contentEl_.contains(oldNode)) {
+        oldNode = null;
+
+        log.warn('Prevented to execute \'replaceChild\' on \'Node\' since the node to be replaced was not a child of this node.');
```
brandonocasey, line 95: perhaps this should instead be: "Prevented TimeDisplay#updateTextnode_ from running as the content el no longer contains the text node element."

kontrollanten: That was a bit clearer, but it's not prevented from running, it's just changing its behavior. What do you say about the following?

```suggestion
        log.warn('TimeDisplay#updateTextnode_: Prevented to replace text node element since it was no longer a child of this node. Appending a new node instead.');
```

mister-ben: "Prevented replacement of" in that case

kontrollanten:
```suggestion
        log.warn('TimeDisplay#updateTextnode_: Prevented replacement of text node element since it was no longer a child of this node. Appending a new node instead.');
```

brandonocasey (approved):

gkatsev (approved):

gkatsev:
Looks like this broke IE11. The TimeDisplay ends up appending rather replacing. It looks like IE11 doesn't handle text nodes with contains
image

I'll make a PR to switch to looking at comparing the first child 
image

gkatsev:
MDN confirms that IE11 only supports HTMLElement for contains: https://developer.mozilla.org/en-US/docs/Web/API/Node/contains#browser_compatibility. Fix here https://github.com/videojs/video.js/pull/7059

PR #6982, opened 2020-12-07
fix: better evented validation and error messages

brandonocasey:
Our current error messages and validation are not enough as some errors can still seek through. I improved our error handling and reporting in many places and will make code comments to show why a change was made below. Originally this was a part of #6967

### DIFF
--- a/src/js/mixins/evented.js
+++ b/src/js/mixins/evented.js
@@ -9,6 +9,26 @@ import * as Fn from '../utils/fn';
 import * as Obj from '../utils/obj';
 import EventTarget from '../event-target';
 
+const objName = (obj) => {
+  if (typeof obj.name === 'function') {
+    return obj.name();
+  }
+
+  if (typeof obj.name === 'string') {
+    return obj.name;
+  }
+
+  if (obj.name_) {
+    return obj.name_;
+  }
+
+  if (obj.constructor && obj.constructor.name) {
+    return obj.constructor.name;
+  }
+
+  return typeof obj;
+};
+
 /**
  * Returns whether or not an object has had the evented mixin applied.
  *
@@ -68,9 +88,9 @@ const isValidEventType = (type) =>
  * @param  {Object} target
  *         The object to test.
  */
-const validateTarget = (target) => {
-  if (!target.nodeName && !isEvented(target)) {
-    throw new Error('Invalid target; must be a DOM node or evented object.');
+const validateTarget = (target, obj, fnName) => {
+  if (!target || (!target.nodeName && !isEvented(target))) {
+    throw new Error(`Invalid target for ${objName(obj)}#${fnName}; must be a DOM node or evented object.`);
   }
 };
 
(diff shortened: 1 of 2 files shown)


src/js/mixins/evented.js
```diff
@@ -68,9 +88,9 @@ const isValidEventType = (type) =>
  * @param  {Object} target
  *         The object to test.
  */
-const validateTarget = (target) => {
-  if (!target.nodeName && !isEvented(target)) {
-    throw new Error('Invalid target; must be a DOM node or evented object.');
+const validateTarget = (target, obj, fnName) => {
```
brandonocasey, line 91: report the object name and the evented function it was trying to call when the error occurred. Should make understanding issues with this code easier to follow.

src/js/mixins/evented.js
```diff
@@ -171,7 +191,7 @@ const normalizeListenArgs = (self, args) => {
  *         A listener function.
  */
 const listen = (target, method, type, listener) => {
-  validateTarget(target);
+  validateTarget(target, target, method);
```
brandonocasey, line 194: We never validated listen, which means a trigger on a disposed object or an invalid trigger would go through and give some strange errors further down in our `dom` helpers.

src/js/mixins/evented.js
```diff
@@ -398,6 +418,14 @@ const EventedMixin = {
    *          Whether or not the default behavior was prevented.
    */
   trigger(event, hash) {
+    validateTarget(this.eventBusEl_, this, 'trigger');
+
+    const type = event && typeof event !== 'string' ? event.type : event;
+
+    if (!isValidEventType(type)) {
```
brandonocasey, line 425: We never validated trigger which could get us into all kinds of strange errors in lower helper objects. Usually in `Events` or `Dom`. Validation here, along with the new errors, will make debugging issues much easier.

test/unit/mixins/evented.test.js
```diff
@@ -61,33 +62,76 @@ QUnit.test('evented() with custom element', function(assert) {
   );
 });
 
+QUnit.test('trigger() errors', function(assert) {
```
brandonocasey, line 65: We never had a test for this before (since we didn't have an errors). I added it.

src/js/mixins/evented.js
brandonocasey: I think that chances of breaking or almost zero because an error would be thrown anyway if `isValidEventType` returned `false`. The resulting error would just be less useful.

gkatsev (approved):

thijstriemstra:
This broke something in my plugin. Investigating..

gkatsev:
In 7.11.3, we had it throw in all error cases, in 7.11.4 we only throw in a few cases https://github.com/videojs/video.js/pull/7067.
Can you share a snippet of you using the event system? We definitely didn't mean to break anyone, and we probably would want to change it never throw again. I guess that's the risk of these changes. Would just be nice to know exactly what you're doing.

thijstriemstra:
@gkatsev it actually uncovered an issue in my tests: there was no player passed into constructor of custom component. So this is a good change for video.js so far.

I fixed it here: https://github.com/collab-project/videojs-record/commit/493440ba38535776558e99bd4f811c8be2cdf383#diff-fd3e579b7fcc86745372fdfc089275daa8bc05c106d2ae363004114511d04481

gkatsev:
@thijstriemstra ah, cool! So, no further action needed on our end right now?

thijstriemstra:
nope!

PR #6984, opened 2020-12-07
fix: prevent dispose error and text track duplicate listeners

brandonocasey:
## Description
Noticed an error when disposing a player with text tracks on and was able to replicate it on our standard simple demo by turning on throttling to 6x in chrome and disposing the player with the default text track on. I also noticed that we do two other things:
1. It's possible to configure the player such that we trigger two updates for a text track on every timeupdate, as we don't remove timeupdate handlers before adding them in modechange.
2. If mode is set to `showing` more then once in a text track we add another update handler, without a limit to how many we can add.

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -201,10 +201,9 @@ class TextTrack extends Track {
 
     const disposeHandler = () => {
       this.tech_.off('timeupdate', timeupdateHandler);
-      this.tech_.off('dispose', disposeHandler);
     };
 
-    this.tech_.on('dispose', disposeHandler);
+    this.tech_.one('dispose', disposeHandler);
     if (mode !== 'disabled') {
       this.tech_.on('timeupdate', timeupdateHandler);
     }
@@ -243,6 +242,10 @@ class TextTrack extends Track {
           if (!TextTrackMode[newMode]) {
             return;
           }
+          if (mode === newMode) {
+            return;
+          }
+
           mode = newMode;
           if (!this.preload_ && mode !== 'disabled' && this.cues.length === 0) {
             // On-demand load.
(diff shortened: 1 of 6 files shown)


src/js/tracks/text-track.js
```diff
@@ -241,12 +248,10 @@ class TextTrack extends Track {
             // On-demand load.
             loadTrack(this.src, this);
           }
+          this.tech_.off('timeupdate', timeupdateHandler);
```
brandonocasey, line 251: Should we also return early in this function if the mode is the same as the current mode? Is there a reason we trigger `modechange` when a track goes from `disabled` to `disabled` again?

gkatsev: Making this change makes sense. I think we'd want to see whether doing that causes any issues for the captions menu, though, I doubt it.

gkatsev:
Is there a specific timing for adding the throttling? I've tried it and cannot reproduce.

I think this hasn't really come up previously because the menu itself prevents changing the mode to the same item already. It's only an issue programmatically.

src/js/tracks/text-track.js
```diff
@@ -196,10 +199,14 @@ class TextTrack extends Track {
       }
     });
 
+    const disposeHandler = () => {
+      this.tech_.off('timeupdate', timeupdateHandler);
+      this.tech_.off('dispose', disposeHandler);
+    };
+
+    this.tech_.on('dispose', disposeHandler);
```
gkatsev, line 207: alternatively, use .one here?

gkatsev (commented):
Changes here make sense, though, I wasn't able to reproduce.

gkatsev (approved):

PR #6993, opened 2020-12-11
fix: seek to live until seekable stops changing or user seeks

### LINKED ISSUE #6975: Live hls stream with discontinuities starts behind live
## Description
On a live stream with discontinuities video.js starts playing behind live by about 1 minute.
With the same stream other players like Clappr or native safari hls player the video starts at live edge.
This can be reproduced constantly [here](https://videojs-http-streaming.netlify.app/?debug=true&autoplay=true&muted=false&minified=true&liveui=true&partial=false&url=https%3A%2F%2Fstorage.googleapis.com%2Fhls-test%2Fandrei-test-gcloud%2Findex.m3u8&type=application%2Fvnd.apple.mpegurl&keysystems=&buffer-water=false).

## Results
image

On streams without discontinuities the video starts playing at live edge as expected.

### Expected
Video should autoplay from live edge.

### Actual
Video starts behind the live edge by approximately 1 minute

brandonocasey:
## Description
Continue trying to seek to live edge until we get to it, but do not do it if the user seeks.
Fixes #6975

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -155,6 +155,7 @@ class LiveTracker extends Component {
       this.one(this.player_, 'timeupdate', this.handleFirstTimeupdate);
     } else {
       this.on(this.player_, 'seeked', this.handleSeeked);
+      this.seekToLiveEdge();
     }
   }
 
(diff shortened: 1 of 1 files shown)


gkatsev:
I think I'm going to close this for now. We can revisit this at a later date.

PR #7021, opened 2020-12-29
fix: do not preload default text track if preloadTextTracks is false

### LINKED ISSUE #7019: Non-preloaded, default text track is loaded and parsed twice
## Description

When not preloading text tracks and setting a default, the default text track is being loaded and parsed twice, resulting in a "double caption"

Reduced test case: https://codepen.io/scarlton/pen/ZEpvaEo

## Steps to reproduce
1. Set `preloadTextTracks` to `false`
2. Set one of your text tracks as `default`
3. Load the player

### Expected
A single network request for the `default` text track, the `default` track to be selected, the selected track cues to be shown.
image

### Actual
Two network requests for the `default` text track, the `default` track is selected, the selected track cues are showing double.
image

## Additional Information
On track load, if we are preloading OR this track is default, then `loadTrack` is called. However, when we are NOT preloading, `loadTrack` is also called for each track when its mode changes. Since a track set as `default` fires a mode change, `loadTrack` ends up being called twice.

As far as I can tell, the issue could be resolved by removing `|| default_` from [line 344 of text-track.js](https://github.com/videojs/video.js/blob/affc0611d978b1a6d02e443a46d14a58f2b1441c/src/js/tracks/text-track.js/#L344). When preloading, all tracks would still be loaded here while non preloaded default tracks would load when its mode changes.

#### videojs
7.10.2
#### browsers
Chrome (probably others)
#### OSes
macOS (probably Windows)

isabelleingato:
When preloadTextTracks option is set to false, it still preloads the default text track. This leads to duplicate tracks once the mode changes to showing and the track is loaded a second time. This includes the default text track in the behavior defined by the preloadTextTracks option.

Fixes #7019

## Description
#7019

## Specific Changes proposed
Removes default text tracks from exclusion of behavior defined by `preloadTextTracks` option

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -341,7 +341,7 @@ class TextTrack extends Track {
         // Act like we're loaded for other purposes.
         this.loaded_ = true;
       }
-      if (this.preload_ || default_ || (settings.kind !== 'subtitles' && settings.kind !== 'captions')) {
+      if (this.preload_ || (settings.kind !== 'subtitles' && settings.kind !== 'captions')) {
         loadTrack(this.src, this);
       }
     } else {
--- a/test/unit/tracks/text-track.test.js
+++ b/test/unit/tracks/text-track.test.js
@@ -419,6 +419,52 @@ QUnit.test('enabled and disabled cuechange handler when changing mode to showing
   player.dispose();
 });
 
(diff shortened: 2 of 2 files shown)


gkatsev (approved):
Thanks!
I can also verify that without the fix the test fails.

brandonocasey (approved):

PR #7022, opened 2020-12-30
docs: Add note to legacy notes 

mister-ben:
## Description
People keep finding the v4 legacy docs at docs.videojs.com/docs and Google keeps positioning them highly in search results. This attempts to lessen that.

## Specific Changes proposed
- Adds a note with a link to the top of each page.
- Adds a `link` with `rel="canonical"` to each page.
- The URL linked is the equivalent doc in current docs if my script could figure it out otherwise just docs.videojs.com

Putting a nofollow rule for /docs in a robots.text wouldn't hurt too.


gkatsev (approved):
Thanks @mister-ben! It's something I've wanted for a long time, thanks for actually getting to it.

gkatsev:
If we want a robots.txt, we can just add it to the docs repo as a file

PR #7034, opened 2021-01-11
fix: set liveWindow to 0 liveCurrentTime is Infinity

brandonocasey:
## Fixes
* Prevents the live UI from showing when we don't have a seekable range
* Prevents the live ui control bar and tooltips from looking invalid when the trackingThreshold is turned off. (everything will display 0 now).

## Description
Fixes an issue in native Safari and Android HLS playback where liveCurrentTime returns Infinity (as we don't have a seekableEnd or seekableEnd is actually Infinity). Which causes the new live ui to show up when we don't really have a live window. 

Instead of returning Infinity when liveCurrentTime is Infinity, return 0. So that everything knows that we do not have a seekable window of live playback.

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -103,7 +103,7 @@ class LiveTracker extends Component {
 
     // we cannot be behind if
     // 1. until we have not seen a timeupdate yet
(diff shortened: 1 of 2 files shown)


gkatsev (approved):
Tested it and verified it's now working as expected.

PR #7035, opened 2021-01-12
fix(fs): make sure handlers are unique per player

### LINKED ISSUE #7013: IE11 - Fullscreen doesn't work when disposing of another player on the same page
## Steps to reproduce
1. I have two videos on the same page, suppose that video1 and video2
2. Dispose of video1 by calling videojs("video1").dispose()
3. Call videojs("video2").requestFullscreen()

## Results
Video2 is not full screen and lost control bar
image

### Expected
fullscreen is normal

#### videojs
 videojs 7.8.4
#### browsers
IE 11 OS  18363.1256

gkatsev:
Using Video.js's .bind still makes us attach prototype methods as
handlers. Then when one is removed, all handlers are removed.

Instead, use arrow methods to make these methods unique.

Fixes #7013.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -351,8 +351,8 @@ class Player extends Component {
     super(null, options, ready);
 
     // Create bound methods for document listeners.
-    this.boundDocumentFullscreenChange_ = Fn.bind(this, this.documentFullscreenChange_);
-    this.boundFullWindowOnEscKey_ = Fn.bind(this, this.fullWindowOnEscKey);
+    this.boundDocumentFullscreenChange_ = (e) => this.documentFullscreenChange_(e);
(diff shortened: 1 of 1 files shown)


brandonocasey:
Code looks good we probably want to do this for visiblity change handles that we have: https://github.com/videojs/video.js/blob/330c82c88fa3d85de25f9d4fb0118f3b6a801a1c/src/js/live-tracker.js#L55
https://github.com/videojs/video.js/blob/330c82c88fa3d85de25f9d4fb0118f3b6a801a1c/src/js/control-bar/progress-control/seek-bar.js#L71

gkatsev:
That makes sense, I'll make a separate PR for that.

brandonocasey (approved):

PR #7041, opened 2021-01-14
test: update ci workflow to prevent install failures

brandonocasey:
Fixes the issue seen in the tests that causes dependency installs to fail. Such as in this pull request: https://github.com/videojs/video.js/pull/7038 

This is just taking the updates from https://github.com/videojs/.github/blob/main/workflow-templates/ci.yml that were implemented in @videojs/http-streaming


gkatsev (approved):

gkatsev:
Thanks!

PR #7045, opened 2021-01-15
chore: netlify ci demo and docs

brandonocasey:
Add a ci target for the video.js demo page so that we can view docs and the demo page easier in pull requests.


build/netlify-docs.js
```diff
@@ -1,26 +0,0 @@
...
-const GIT_LOG = `git log --format=%B -n 1 ${process.env.COMMIT_REF}`;
-const output = sh.exec(GIT_LOG, {async: false, silent: true}).stdout;
-
-// if we're on master branch and not on a tagged commit,
-// error the build so it doesn't redeploy the docs
-if (process.env.BRANCH === 'master' && semver.valid(output.trim()) === null) {
-  process.exit(1);
-
-// if we're on any other branch, we can regenerate docs
-} else {
```
gkatsev, line 14: we still need this piece so that the site doesn't get updated in main until there's a tag. Also, this probably needs to change to be `main` instead now.

gkatsev (approved):

PR #7050, opened 2021-01-19
docs: change master to main

brandonocasey:
Noticed that a lot of our docs point to the master branch, which is out of date and unused now. Along with this pull request we will have to update https://github.com/videojs/conventional-changelog-videojs to use `main`.


gkatsev (approved):
Updated that repo to use main.

PR #7059, opened 2021-01-22
fix(time-display): fix IE11 appending times instead of replacing

gkatsev:
IE11 only supports HTMLElement and not text nodes in the contains
method. See https://developer.mozilla.org/en-US/docs/Web/API/Node/contains
Instead, compare firstChild.

### DIFF
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -89,7 +89,7 @@ class TimeDisplay extends Component {
 
       let oldNode = this.textNode_;
 
-      if (oldNode && !this.contentEl_.contains(oldNode)) {
+      if (oldNode && this.contentEl_.firstChild !== oldNode) {
         oldNode = null;
 
         log.warn('TimeDisplay#updateTextnode_: Prevented replacement of text node element since it was no longer a child of this node. Appending a new node instead.');


gkatsev:
Compare the remaining time here https://videojs-preview.netlify.app/sandbox/index.html (broken):
image
vs the fixed here https://deploy-preview-7059--videojs-preview.netlify.app/sandbox/index.html:
image

src/js/control-bar/time-controls/time-display.js
```diff
@@ -89,7 +89,7 @@ class TimeDisplay extends Component {
 
       let oldNode = this.textNode_;
 
-      if (oldNode && !this.contentEl_.contains(oldNode)) {
+      if (oldNode && this.contentEl_.firstChild !== oldNode) {
```
brandonocasey, line 92: Should we check `oldNode.parentElement` and `oldNode.parentElement === this.contentEl_` that way we don't get into a situation where `contentEl_` has multiple elements and the first one is not oldNode, causing us to insert nodes over and over.

gkatsev: Maaaybe. Time displays should only have the one text node in them.

brandonocasey (approved):

PR #7063, opened 2021-01-25
Fix up preventDefault error on Android

ckybonist:
## Description
Focus on video player and scroll will raise the error
```
Ignored attempt to cancel a touchend event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.
```

image

To reproduce the error:
1. Put your finger on video player (do not lift up the finger)
2. Scroll up and down

## Specific Changes proposed
Check `event.cancelable` before executing `event.preventDefault`

see: https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault#notes

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2023,7 +2023,9 @@ class Player extends Component {
    */
   handleTechTouchEnd_(event) {
     // Stop the mouse events from also happening
-    event.preventDefault();
+    if (event.cancelable) {
+      event.preventDefault();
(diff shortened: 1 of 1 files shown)


gkatsev (approved):
Looks like cancelable should be supported most places, so, LGTM

brandonocasey (approved):

PR #7067, opened 2021-01-26
fix(evented): log an error on invalid type

gkatsev:
Follow up from #6982. We previously threw an error, but we've seen it
happen unexpectedly. Instead, we should log an error.

Here, if we have a `log` object on the current object, we should use it,
otherwise, we use a default `log` object.

### DIFF
--- a/src/js/mixins/evented.js
+++ b/src/js/mixins/evented.js
@@ -8,6 +8,7 @@ import * as Events from '../utils/events';
 import * as Fn from '../utils/fn';
 import * as Obj from '../utils/obj';
 import EventTarget from '../event-target';
+import log from '../utils/log';
 
 const objName = (obj) => {
   if (typeof obj.name === 'function') {
@@ -444,7 +445,7 @@ const EventedMixin = {
     const type = event && typeof event !== 'string' ? event.type : event;
 
     if (!isValidEventType(type)) {
-      throw new Error(`Invalid event type for ${objName(this)}#trigger; ` +
+      (this.log || log).error(`Invalid event type for ${objName(this)}#trigger; ` +
         'must be a non-empty string or object with a type key that has a non-empty value.');
     }
     return Events.trigger(this.eventBusEl_, event, hash);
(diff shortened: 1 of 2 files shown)


src/js/mixins/evented.js
```diff
@@ -444,7 +445,7 @@ const EventedMixin = {
     const type = event && typeof event !== 'string' ? event.type : event;
 
     if (!isValidEventType(type)) {
-      throw new Error(`Invalid event type for ${objName(this)}#trigger; ` +
+      (this.log || log).error(`Invalid event type for ${objName(this)}#trigger; ` +
```
gkatsev, line 448: This allows us to use the player/plugin/etc log if it exists, for example:
```
> player.controlBar.trigger({})
< VIDEOJS: ERROR: Invalid event type for ControlBar#trigger; must be a non-empty string or object with a type key that has a non-empty value.
> player.trigger({})
< VIDEOJS: vjs_video_3: ERROR: Invalid event type for null#trigger; must be a non-empty string or object with a type key that has a non-empty value.
```

brandonocasey (approved):

src/js/mixins/evented.js
```diff
@@ -444,8 +445,14 @@ const EventedMixin = {
     const type = event && typeof event !== 'string' ? event.type : event;
 
     if (!isValidEventType(type)) {
-      throw new Error(`Invalid event type for ${objName(this)}#trigger; ` +
-        'must be a non-empty string or object with a type key that has a non-empty value.');
+      const error = `Invalid event type for ${objName(this)}#trigger; ` +
+        'must be a non-empty string or object with a type key that has a non-empty value.';
+
+      if (event) {
+        (this.log || log).error(error);
```
brandonocasey, line 452: should we return no matter what if we don't get a valid event type?

gkatsev: I think to reduce potential breakage, we want to try and continue through to the next part here and not return early.
Though, if they got here there's probably something wrong, but it could still be working in some way, for better and worse.

PR #7074, opened 2021-02-02
fix(dom): stop findPosition at a fullscreenElement

gkatsev:
On iPads, if the player is inside a floating div, when fullscreen, we'll have an incorrect offset value. Instead, we should stop counting our offsets once we get to our fullscreen element (i.e. the player) because in fullscreen it's outside the regular flow, and we don't actually want to add offset the float gives us inline.

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -4,6 +4,7 @@
  */
 import document from 'global/document';
 import window from 'global/window';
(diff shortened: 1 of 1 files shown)


brandonocasey (approved):
I haven't tested but the changes make logical sense to me based on the description of the issue.

PR #7079, opened 2021-02-03
fix(dom): account for translated parent in pointer position

gkatsev:
Since we switched to using adding up offsets to calculate the pointer position relative to the current element, we've had some issues, particularly on iOS, where a translated parent would cause us to miscalculate the position. This is currently a quickfix for the issue, and I'd like to spend some time to figure out a better solution that hopefully won't require us to iterate through the DOM and add up the transform matrix.

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -8,6 +8,7 @@ import fs from '../fullscreen-api';
 import log from './log.js';
 import {isObject} from './obj';
 import computedStyle from './computed-style';
+import * as browser from './browser';
 
 /**
  * Detect if a value is a string with any non-whitespace characters.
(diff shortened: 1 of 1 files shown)


src/js/utils/dom.js
```diff
@@ -607,6 +609,33 @@ export function findPosition(el) {
  *
  */
 export function getPointerPosition(el, event) {
+  const translated = {
+    x: 0,
+    y: 0
+  };
+
+  if (browser.IS_IOS) {
```
gkatsev, line 617: on iOS, iterate up the DOM tree and add up all the translation transforms so that we can modify the pageX and pageY coordinates for changedTouches to account for it.

gkatsev:
If I can't find a better solution, I could at least improve this solution so that we don't iterate up the DOM tree each time that getPointerPosition is called, but I think it's good enough for a quickfix.

misteroneill (approved):
Seems good for a quick fix.

brandonocasey (approved):

PR #7098, opened 2021-02-17
feat: enable responsive controls on fullscreen 

ipadilla4:
A client reported that control buttons are not positioned correctly when player is full-screened and screen is small.

Currently we are excluding our responsive controls on fullscreen and changes include removing that. This solves the issue for small screens. Also added _!important_ rule to override BC skin properties and hide controls when necessary.

### DIFF
--- a/src/css/components/_adaptive.scss
+++ b/src/css/components/_adaptive.scss
@@ -18,7 +18,7 @@
 // - Volume Mute button
 // - Fullscreen Button
 //
-.video-js:not(.vjs-fullscreen) {
+.video-js {
 
   &.vjs-layout-small,
   &.vjs-layout-x-small,
@@ -34,7 +34,7 @@
     .vjs-subtitles-button,
     .vjs-audio-button,
     .vjs-volume-control {
-      display: none;
+      display: none !important;
     }
 
     // Reset the size of the volume panel to the default so we don't see a big


gkatsev (approved):

misteroneill (approved):
This is good, I think, but we should look at our internal BC skin to see if we can change the specificity such that the `!important` is not necessary.

amtins:
@gkatsev, @misteroneill we recently updated to the latest videojs version which contains this fix. However, this fix forces us to introduce the use of !important, which means we will have to do something like @infinite-persistence did [here](https://github.com/lbryio/lbry-desktop/commit/17a21e59dc4eb71b22875a99d562ee472cf4d5cb). In our case, this is not a good solution because we provide a library in which integrators can override some rules by increasing the specificity.

I don't know if I missed anything and what the BC means, but I think the !important is not necessary, based on my tests. @ipadilla4, maybe a reduced test case will help to understand why the !important is so important.

gkatsev:
You're definitely right. We should look into and remove `!important`

gkatsev:
@amtis this is being removed by https://github.com/videojs/video.js/pull/7449 and we'll get a patch released soon!

gkatsev:
@amtins the change is in Video.js 7.15.7

PR #7113, opened 2021-03-01
fix: ModalDialog, PosterImage :focus styles

### LINKED ISSUE #7112: ModalDialog and PosterImage components lose background styles if a non-focusable child element is clicked
## Description
If any element other than a focusable control is clicked within a ModalDialog or PosterImage, the component loses its background styles. For ModalDialog, this results in the background becoming transparent and the player surface displaying through the modal content. For PosterImage, this causes the background image to be positioned and sized incorrectly.

Reduced test case: https://codepen.io/acmertz/pen/WNoxRwq

### ModalDialog
1. Open a ModalDialog, either by calling `player.createModal()` or by creating a new ModalDialog component, adding it to the player, and calling `open()` on it
2. Click on a non-focusable element within the contents of the ModalDialog once it appears
3. The background color of the ModalDialog disappears, and the video surface is no longer dimmed
4. If a control within the ModalDialog or an element outside the modal is focused, the dialog's background is restored.

### PosterImage
1. In any player that displays a poster image, click anywhere on the poster image's surface EXCEPT for the actual play button
2. The poster image will momentarily be displayed with an incorrect size and position before the poster image is hidden for playback
3. To more easily see the effect, click and hold anywhere in the PosterImage except for the play button. As long as your mouse is depressed (and the click event therefore hasn't fired yet), the image will be displayed incorrectly.

### Expected
ModalDialog and PosterImage components' backgrounds do not change when the component is clicked.

### Actual
The ModalDialog's background becomes transparent and the PosterImage's image is resized strangely whenever the component is clicked.

### Error output
None

## Additional Information

This appears to have been caused by #6871, which causes `tabindex="-1"` to be set on ModalDialog and PosterImage components where previously it was not. As a result, it's now possible for CSS `:focus` styles to be applied to these components, and as such, their own styles are being overridden by a rule elsewhere in the CSS intended to manage focus styles on other controls when clicking vs. tabbing through the interface.

#### videojs
7.11.0&ndash;7.11.5

What version of videojs does this occur with?
#### Browsers
Chrome, Firefox

#### OSes
Windows 10

### Plugins
None

acmertz:
Fixes #7112

## Description
Fixes a styling issue that occurred after #6871 where the background styles would be removed from ModalDialog and PosterImage components when they received focus due to a click.

## Specific Changes proposed
Use `:not([tabindex="-1"])` to prevent the global styles for mouse/keyboard focus on a control from applying to ModalDialog and PosterImage components' backgrounds.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/video-js.scss
+++ b/src/css/video-js.scss
@@ -60,8 +60,8 @@
   background: none;
 }
 
-.video-js *:focus:not(:focus-visible),
-.video-js .vjs-menu *:focus:not(:focus-visible) {
+.video-js *:focus:not(:focus-visible):not([tabindex="-1"]),
+.video-js .vjs-menu *:focus:not(:focus-visible):not([tabindex="-1"]) {
   outline: none;
   background: none;
 }


acmertz:
@gkatsev No problem. I've updated the PR - have a look and let me know if that's better.

gkatsev (approved):

gkatsev:
Thanks, this is great. Targeting it specifically is helpful. I verified it, took me a second to figure out what #5888 was trying to fix.

misteroneill (approved):
Nice, thanks!

PR #7114, opened 2021-03-02
fix: try enabling liveui on canplay

gkatsev:
On native playback engines, like Safari, for some streams, seekable and
currentTime won't have correct values until video data has started
loading. Previously, we only tried to turn toggle tracking on
durationchange, now we also do it on canplay.

This is a followup to #7034 which helped prevent the live tracker from
showing up in some cases.

### DIFF
--- a/src/js/live-tracker.js
+++ b/src/js/live-tracker.js
@@ -47,6 +47,9 @@ class LiveTracker extends Component {
     this.reset_();
 
     this.on(this.player_, 'durationchange', this.handleDurationchange);
+    // we should try to toggle tracking on canplay as native playback engines, like Safari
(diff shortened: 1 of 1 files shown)


src/js/live-tracker.js
```diff
@@ -47,6 +47,9 @@ class LiveTracker extends Component {
     this.reset_();
 
     this.on(this.player_, 'durationchange', this.handleDurationchange);
```
gkatsev, line 49: this handler will get wrapped in an arrow function in this PR https://github.com/videojs/video.js/pull/7060

brandonocasey (approved):
code looks good and I tested to confirm that this does fix the issue.

PR #7134, opened 2021-03-11
Add a player option `noUITitleAttributes` to prevent title attributes in the UI (fixes #6767).

### LINKED ISSUE #6767: Configurable title attribute strings for player controls to improve accessibility
## Description
The current means of configuring player control strings assigns the same string for both the button's name and its `title` attribute. While this behavior is well-intentioned, it is ultimately [harmful for accessibility](https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/rl). As background, the general purpose of the `title` attribute is to provide supplementary information to help clarify or further describe the purpose of a control. Simply duplicating a control's existing string label is misuse, and results in redundant output when using screen reading software.

As noted in the HTML specification under [https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute):
> Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

While it is common practice to utilize the `title` attribute to display native tooltips, this in itself is also problematic. For one, native browser tooltips only render while utilizing a pointer device and are thus completely inaccessible to keyboard-only users. Behavior is also inconsistent between various browser implementations. In this case, a custom accessible tooltip is a much better approach. There are several other points discussed under https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/ that dissuade its use.

For these stated reasons, I propose providing the ability to customize/override and prevent control strings from being automatically applied to their respective `title` attributes. This can likely be accomplished in backward compatible means so that existing implementations aren't affected.

Our organization discourages and is in the process of replacing similar patterns of redundant `title`s. However, this is something we are blocked on with our media players which utilize video.js and is also observed on BigPlayButton. We also leverage custom accessible tooltips which ultimately render the button titles unnecessary, but have no straight forward means of preventing them from rendering.

OwenEdwards:
## Description
Some designers and developers feel that the presence of (redundant) `title` attribute values on UI controls makes the accessibility of those controls worse. The title attributes can also conflict with additions to make truly accessible tooltips on the controls.

This PR allows developers to easily disable the addition of `title` attributes on all UI controls with a single player configuration option.

## Specific Changes proposed
Add a player option `noUITitleAttributes` which, if set to `true`, prevents video.js from adding any `title` attributes to UI controls.

## Requirements Checklist
- [x] Feature implemented
  - [ ] Docs/guides updated
  - [x] Example created


gkatsev (approved):
Makes sense to me.

OwenEdwards:
I need to add documentation in docs/guides/options.md before this is ready to merge.

docs/guides/options.md
```diff
@@ -28,8 +28,11 @@
   * [language](#language)
   * [languages](#languages)
   * [liveui](#liveui)
+  * [liveTracker.trackingThreshold](#livetrackertrackingthreshold)
+  * [liveTracker.liveTolerance](#livetrackerlivetolerance)
```
OwenEdwards, line 32: I don't know where these two lines came from?! I can remove them?

gkatsev: If you've run `npm run docs:fix`, it probably added these two items that were added to the doc but we missed updating the TOC with them. https://github.com/videojs/video.js/pull/7097

OwenEdwards: @gkatsev so it would be okay to merge this change and let those two lines get added in `options.md`? Or would you prefer to have them in a separate PR?

gkatsev: They can go in here.

PR #7142, opened 2021-03-23
fix: Remove extra timeupdate event when progress controls disabled

alex-barstow:
## Description
`SeekBar.handleMouseUp()` is a side-effect of disabling progress controls via `ProgressControls.disable()`. This is an issue because `SeekBar.handleMouseUp()` [manually triggers](https://github.com/videojs/video.js/blob/main/src/js/control-bar/progress-control/seek-bar.js#L352) a `'timeupdate'` event (because it assumes a seek has just taken place). If someone attempts to disable the progress controls before playback starts, a `'timeupdate'` will fire potentially before we even have `'loadedmetadata'`, which can cause problems.

## Specific Additions
- Move the relevant mouse and touch event listener removal that is needed by the `disable()` method from `handleMouseUp()` into its own function.
- Preserve the behavior where `ProgressControls.disable()` will resume a normal playback state if it is invoked while the player is scrubbing.

### DIFF
--- a/src/js/control-bar/progress-control/progress-control.js
+++ b/src/js/control-bar/progress-control/progress-control.js
@@ -5,6 +5,7 @@ import Component from '../../component.js';
 import * as Dom from '../../utils/dom.js';
 import clamp from '../../utils/clamp.js';
 import {bind, throttle, UPDATE_REFRESH_INTERVAL} from '../../utils/fn.js';
+import {silencePromise} from '../../utils/promise';
 
 import './seek-bar.js';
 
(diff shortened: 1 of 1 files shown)


gkatsev:
> a `'timeupdate'` will fire potentially before we even have `'loadedmetadata'`, which can cause problems.

Should we make sure that we have started before we try to trigger `timeupdate`?

gkatsev (approved):
Kind of surprised we don't have any tests for this but I'm not going to make sure write them.

brandonocasey (approved):

PR #7163, opened 2021-04-06
fix: always have an enabled audio track when switching

brandonocasey:
## Description
Only change enabled on one track, the others will automatically be turned off. This prevents us from getting two  `AudioTrackList` `change` events where the first will have no tracks enabled, and the second has a track enabled.


gkatsev (approved):
Works on iOS, so, good to go.

PR #7172, opened 2021-04-08
add MenuRadioButton to unselect other items when selected

AwokeKnowing:
## Description
Most time, when you select an item in a menu, the other items should be unselected.
Currently, each custom Menu or MenuItem must figure out how to manage this, such as by iterating through items on some event, and removing the vjs-selected class or calling unselected.

Many plugins and derived menus implement some kind of special logic just to unselect the other items

## Specific Changes proposed
In order to not conflict with existing menus and plugins, the MenuRadioItem is a new more specialized MenuItem that by default will call select(false) on the other items when one item is clicked on.  

The main change of the proposal is a default check to unselect other items:
```javascript
  handleClick(event) {
    for (const item of this.parentComponent_.parentComponent_.items) {
      item.selected(item.id_ === this.id_);
    }
  }
```

With this change, making a specialized menu item, only involves handleClick.

```javascript
class MyCoolMenuItem extends MenuRadioItem
{
  handleClick(event) {
    super.handleClick(event)
    this.player().trigger('MyItemSelected', this.options_)
  }
}
```

furthermore, I'm suggesting to add a clickHandler parameter so entire menus can be created **without needing to subclass the menuitem** (i mean really it's just an item in a menu, associated to an object value) like this:

```javascript
var MenuButton = videojs.getComponent('MenuButton');
var MenuRadioItem = videojs.getComponent('MenuRadioItem');

class MyBasicMenuButton extends MenuButton
{
  createItems() {
    return this.options_.items.map(item => 
      new MenuRadioItem(this.player_, item, this.options_.clickHandler)
    );
  }
}
videojs.registerComponent('MyBasicMenuButton', MyBasicMenuButton);

player = videojs('my-player', null, function() {
  let menuOptions = {
      items: [{"label":"item 1", val: 1}, {"label":"item 2", val: 2}, {"label":"item 3", val: 3}], 
      clickHandler: (e, item) => alert(item.val)
  };
  this.getChild('controlBar').addChild('MyBasicMenuButton', menuOptions);
});



```

The big difference is that the code above seems 'normal' and does what you think.  But without the MenuRadioItem you must add code to handle unselecting other items, you end up with a weird mix of very unexpected low-level code handling html or reflecting to find the parent and iterate through the children all for 'nothing' (just unselect other items).

The modified MenuItem makes creating menus clean and simple.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/menu/menu-radio-item.js
+++ b/src/js/menu/menu-radio-item.js
@@ -11,12 +11,14 @@ import Component from '../../component.js';
  * @extends MenuItem
  */
 class MenuRadioItem extends MenuItem {
-  constructor(player, options) {
-    // only one item should be selectable in menu
+  constructor(player, options, clickHandler) {
+    // one and only one item should be selectable in menu at a time
+    options.selectable = true;
     options.selectable = true;
     options.multiSelectable = false;
     super(player, options);
 
+    this.clickHandler = clickHandler;
     this.selected(options.defaultSelection);
   }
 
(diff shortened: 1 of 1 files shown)


mister-ben:
Since `MenuItem` already knows whether it's supposed to be single or multi-selectable in order to add accessibility text, it might make sense for this to be optional behaviour in `selected()` in `MenuItem`. Although off by default.

AwokeKnowing:
> Since `MenuItem` already knows whether it's supposed to be single or multi-selectable in order to add accessibility text, it might make sense for this to be optional behaviour in `selected()` in `MenuItem`. Although off by default.

so like "radioSelection=true" to enable automatically unselecting other items?

PR #7177, opened 2021-04-13
fix(package): remove remove

### LINKED ISSUE #7176: Spurious dependency to 'remove'
## Description

While upgrading video.js I had errors in my license check script (using `license-checker`) because `video.js` is now dependent on `seq`, `hashish` and `traverse` that don't have a correct SPDX identifier via `remove` (All are very old packages with a last publish 10 years ago and predate the SPDX identifier unification in package.json).

This is not a problem as all are MIT and I could whitelist them but I wondered why and noticed that the dependency seem unused.

I suspect it's a spurious dependency introduced in https://github.com/videojs/video.js/pull/7128 that doesn't even seem to mention it.

https://github.com/videojs/video.js/blob/main/package.json#L90

#### videojs
7.11.8

gkatsev:
Fixes #7176

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6837,14 +6837,6 @@
         "pinkie-promise": "^2.0.0"
       }
     },
-    "hashish": {
-      "version": "0.0.4",
-      "resolved": "https://registry.npmjs.org/hashish/-/hashish-0.0.4.tgz",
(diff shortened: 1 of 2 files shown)


brandonocasey (approved):
🤣

misteroneill (approved):
remove!

heyboi78 (approved):

PR #7178, opened 2021-04-13
fix(package): update vtt.js to allow server-side-rendering

gkatsev:
See https://github.com/videojs/vtt.js/pull/56


brandonocasey (approved):

gkatsev:
Double-checked bipbop and oceans and captions still show up.

Issue #7068, opened 2021-01-26
Tracking for a Major version of Video.js


gkatsev:
This issue is to track things that should happen whenever we plan or decide it's time to make a major version of Video.js. There's no timeline right now.

- [ ] Convert https://github.com/videojs/video.js/pull/7067 to *always* throw as something is wrong if you got there.

mister-ben:
Some thoughts

- Use keys like PLAY_BUTTON for translation and import a en.json by default, to allow flexibility to change translations without further changing the key
- Review text and translations generally
- Icons - are fonts still the best way to go now? Would inline SVG be more flexible?
- Standard styling options for buttons - icon only, icon and text and text only.
- Use CSS variables - could be easier for a plugin to follow player style 
- Normalise ID3 cue data (`cue.value.data` can be at least two types string, or a Uint8Array depending on browser)
- Settings menu - rather than having multiple menu buttons in the control bar

gkatsev:
For ID3 cues, the Media and Entertainment Interest Group at the W3C is working on updating DataCue and normalizing ID3 and emsg from DASH, I think we probably want to hold off updating ours but also give input to them.

rtritto:
- Remove hardcoded *node_modules* in `video-js.scss` #7208

heff:
> when seeking, set playbackrate to zero instead of pausing and playing, consider hiding events from player in these cases

Interesting idea, is there somewhere I can read up on the background? Wondering if there's any downsides.

gkatsev:
@heff I think that setting the playback rate to zero is closer to what should actually be happening when seeking. Since pausing and playing is kind of a workaround to the fact that if you scrub and stop moving the mouse, it'll start playing from that place, which isn't what we want. Plus, a lot of folks currently get confused when they get extra play/pause events from seeking.

misteroneill:
A few items here are not complete, but we may push them out. They are:

- Update menus on change, rather than recreating a whole new menu (performance). Adding a text track can take >100ms because we have to dispose the menu and recreate.
  - We have one related PR here (in `main` not `next`), but there may be more to do. I don't think this specifically has to be in a major release, though.
- Remove hardcoded node_modules in video-js.scss
  - Didn't seem like a high enough priority, but if someone wants to take it on in the next week or so, that would be great!
- turn retryOnError by default and remove the option
  - I think this one was missed.

gkatsev:
Created PRs for the sass path (#7867) and the retryOnError (#7868).

[...]

PR #7183, opened 2021-04-14
fix: Better text for exit fullscreen

mister-ben:
Improves the tooltip/accessibility text for the fullscreen button so it's clearer the action is to leave fullscreen mode.

Adds an English translation within src/video.js rather than make it the default text which would change the translation key.
Updates English and German translations, some existing translations already translate it more like "exit fullscreen".

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -569,5 +569,9 @@ videojs.url = Url;
 
 videojs.defineLazyProperty = defineLazyProperty;
 
+// Adding less ambiguous text for fullscreen button.
+// In a major update this could become the default text and key.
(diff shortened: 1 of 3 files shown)


gkatsev (approved):
Thanks, this seems like a great way of fixing this.

src/js/video.js
```diff
@@ -569,5 +569,9 @@ videojs.url = Url;
 
 videojs.defineLazyProperty = defineLazyProperty;
 
+// Adding less ambiguous text for fullscreen button.
+// In a major update this could become the default text and key.
```
gkatsev, line 573: added to our major version tracking list https://github.com/videojs/video.js/issues/7068

brandonocasey (approved):

PR #7190, opened 2021-04-16
feat: Add normalizeAutoplay option to treat autoplay: true as autoplay: "play"

alex-barstow:
## Description
This PR adds a new option to treat `autoplay: true` the same as `autoplay: 'play'`. In general we want video.js to handle as much of the autoplay logic itself as possible, since the browser's treatment of the video element's `autoplay` attribute is less predictable. For now the default option value is `false`, but it may be made the default behavior in a future major version.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3677,24 +3677,16 @@ class Player extends Component {
 
     let techAutoplay;
 
-    // if the value is a valid string set it to that
-    if (typeof value === 'string' && (/(any|play|muted)/).test(value)) {
+    // if the value is a valid string set it to that, or normalize `true` to 'play if need be'
+    if (typeof value === 'string' && (/(any|play|muted)/).test(value) || value === true && this.options_.normalizeAutoplay) {
       this.options_.autoplay = value;
-      this.manualAutoplay_(value);
+      this.manualAutoplay_(typeof value === 'string' ? value : 'play');
       techAutoplay = false;
 
     // any falsy value sets autoplay to false in the browser,
     // lets do the same
     } else if (!value) {
       this.options_.autoplay = false;
-
-    // normalize `true` as 'play' if need be
-    } else if (value === true && this.options_.normalizeAutoplay) {
-      // we still want player.autoplay() to return the provided setting,
-      // even though it will be treated as a 'play' behind the scenes
-      this.options_.autoplay = value;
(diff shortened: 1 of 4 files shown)


src/js/player.js
```diff
@@ -3684,6 +3688,14 @@ class Player extends Component {
     } else if (!value) {
       this.options_.autoplay = false;
 
+    // normalize `true` as 'play' if need be
+    } else if (value === true && this.options_.normalizeAutoplay) {
```
gkatsev, line 3692: Should this check be combined into the if on line 3681 above? This way we aren't duplicated the lines below.

gkatsev (commented):
We should document `normalizeAutoplay` in https://github.com/videojs/video.js/blob/main/docs/guides/options.md
Also, worth looking into writing a test (probably in player.test.js because that's were some of this already exists) where we create a player based on a `<video>` element with an `autoplay` attribute but that has the attribute removed on init, I think this part is relatively straight forward. The trickier part, if testable, is whether adding a source will autoplay via our manual autoplay or via the video element autoplaying.

alex-barstow:
@gkatsev Pushed those changes, including a new test. Regarding your question about testing "whether adding a source will autoplay via our manual autoplay or via the video element autoplaying," isn't this sort of [tested already](https://github.com/videojs/video.js/blob/main/test/unit/autoplay.test.js#L116-L133) by making sure `play()` isn't called when `autoplay: true`?

gkatsev:
> isn't this sort of [tested already](https://github.com/videojs/video.js/blob/main/test/unit/autoplay.test.js#L116-L133) by making sure `play()` isn't called when `autoplay: true`?

No, that only verifies that `play()` wasn't called. I specifically want to know whether the video element is will autoplay natively.
For example:
embed code:
```html
<video autoplay id=vid>
</video>
```
At this point, the video element is primed for autoplay. If you were to put a source in, it will start playing it.
```js
const player = videojs('vid', {
  normalizeAutoplay: true
});
player.src('video.mp4');
// playback should be initiated via Video.js calling `play()`
// rather than the video element getting a source starting playback because it was already primed to do so
```

It's probably tricky to get a test for this. Let me know whether this makes sense, happy to talk it about it some more.

gkatsev (approved):
👍🏻

PR #7193, opened 2021-04-19
Make Playback Rate control work with screen readers - fixes #7121

### LINKED ISSUE #7121: Screen Readers do not announce new playback rate after switching to new rate using keyboard only commands
## Description

When I have the playback rate option as outlined above, screen readers do not properly announce the change in playback rate after the rate changes.

You can test the above in this codepen: https://codepen.io/feliperobledo/pen/bGBjXXQ

## Steps to reproduce

1. Open the screen reader software of your choice
2. Open the codepen above
3. Tab or click on the playback rate button to switch the rate.

At this point the screen reader will simply announce, `"Playback rate"`, but it will not announce the current play speed.

### Expected

Consider that the playback rate button has options to go `1`, `1.2`, `1.5`, or `2`.
If the label on the playback rate button is `1x` and I click on the playback rate button, the screen reader software should announce `"Playback rate 1 point 2 x"`, meaning it announces the new label selected rate.

### Actual

The screen reader simply announce `"Playback rate"`

### Error output
N/A

### Proposed solution

Screen Shot 2021-02-23 at 12 06 32 PM (1)

That we add `id`s to the elements that hold the information we want announced, and we tell screen readers that the playback rate button's label is a combination of those elements, as shown above.

I've personally tried it on:
- macOS - Chrome - Screen Reader extension
- macOS - Chrome - VoiceOver

And it works.

#### videojs
`7.9.7`

#### browsers
All of which that have a playback rate button that is not a menu but just a button that changes rate

#### OSes
All

### plugins
Self hosted `videojs-offset`
`videojs-contrib-quality-levels`

OwenEdwards:
## Description
The Playback Rate control is a MenuButton, but has been modified from the standard MenuButton operation so that clicking on the button just switches to the next playback rate without displaying the menu. Also, the label of the button visually includes the current playback rate but that visual indication is not included in the accessible name or description of the button. As a result, screen reader users cannot perceive the current setting of the control, and do not realize that they actually can change the setting (by activating the button) since activating the button does not open a menu but also does not announce the rate change.

This temporary fix associates the current rate selection display in the button with the button itself using `aria-describedby`, so that it is announced by screen readers when the button receives focus, but is not included as part of the name/label of the button (which would be confusing). The new rate is also announced by the screen reader when the setting is changed by activating the button.

This fixes #7121.

Long term, the button needs to be fixed to correctly operate as a menu button for screen reader users, but for now, this change doesn't change any of the DOM structure or any of the behavior for sighted mouse users or sighted keyboard-only users.

## Specific Changes proposed
Add a unique HTML ID to the visible rate label in the Playback Rate menu button control, and associate it with the button via `aria-describedby`.

## Requirements Checklist
  - [x] Change has been verified in an actual browser (Chrome + JAWS)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
+++ b/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
@@ -26,6 +26,8 @@ class PlaybackRateMenuButton extends MenuButton {
   constructor(player, options) {
     super(player, options);
 
+    this.menuButton_.el_.setAttribute('aria-describedby', this.labelElId_);
+
     this.updateVisibility();
     this.updateLabel();
 
@@ -42,8 +44,11 @@ class PlaybackRateMenuButton extends MenuButton {
   createEl() {
     const el = super.createEl();
 
+    this.labelElId_ = 'vjs-playback-rate-value-label-' + this.id_;
+
     this.labelEl_ = Dom.createEl('div', {
       className: 'vjs-playback-rate-value',
+      id: this.labelElId_,
       innerHTML: '1x'
     });
 


gkatsev (approved):

gkatsev:
Thanks, Owen, looks like a great workaround for now.

gkatsev:
Leaving #7121 open to track replacement/updates for it.

PR #7202, opened 2021-04-22
fix: don't reset background styles of selected MenuItem when focused

### LINKED ISSUE #7200: Selected MenuItem loses background when focused and becomes unreadable
## Description

When you select a MenuItem, its text and background colors change to clearly indicate which item is selected (as expected). However, if the MenuItem receives mouse focus while in this state, its background color reverts to transparent but the dark text remains, making it very difficult to read.

https://codepen.io/acmertz/pen/gOgQWZV

## Steps to reproduce
1. Hover over a MenuButton in the control bar that has a menu associated with it (for example, the Captions button)
2. Click on the MenuItem that is currently selected. Notice how the background briefly reverts back to transparent before the Menu disappears, while the text color remains dark.
3. Hover over the MenuButton again. The selected item still has dark text on a dark background.

If the currently selected MenuItem is also the first item in the Menu, then this issue is also easily reproducible by simply clicking the MenuButton so the Menu is locked in its "opened" state. This immediately sets focus to the first (and currently selected) MenuItem and thus reverts the background styles.

### Expected
The selected MenuItem always has a white background and dark text.

### Actual
If a MenuItem is selected and also has focus, it has dark text on a dark background.

## Additional Information
The bug does _not_ occur if you trigger the Menu via your keyboard. The `:focus:not(:focus-visible)` selector used in the Menu styles appears to be conflicting with the MenuItem's "selected" styles. This looks related to #5558 and #7712.

#### Video.js
v7.11.8

#### Browsers
Chrome/Edge, Firefox

#### OS
Windows

### Plugins
None

acmertz:
Fixes #7200

## Description
Fixes a styling issue with the MenuItem component that caused it to display dark text over a dark background when focused, if it was the currently selected MenuItem (`.vjs-selected`).

## Specific Changes proposed
Use `:not(.vjs-selected)` to prevent the menu styles from resetting the background of a selected MenuItem on focus.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/menu/_menu.scss
+++ b/src/css/components/menu/_menu.scss
@@ -56,8 +56,8 @@
   color: $primary-background-color;
 }
 
-.video-js .vjs-menu *:focus:not(:focus-visible),
-.js-focus-visible .vjs-menu *:focus:not(.focus-visible) {
+.video-js .vjs-menu *:not(.vjs-selected):focus:not(:focus-visible),
+.js-focus-visible .vjs-menu *:not(.vjs-selected):focus:not(.focus-visible) {
   background: none;
 }
 


gkatsev (approved):
The easiest way I've found to verify this is to add `vjs-hover` class to a menu button (like subs caps button or the playback rate button), then, find the selected item under the `<vjs-menu>` element, and force the `:focus` state on it. Without this change, you end up no longer being able to read it the text. With this, you can continue seeing the item selected and readable.

Thanks @acmertz!

PR #7214, opened 2021-04-29
fix: add try and catch for computedStyle

weiz18:
## Description
With shadow root enable component, computedStyle gets called on dragging the progress bar. and since we are traversing the path back to the top,  if you have a shadow root component in the path, it will try to call computeStyle for shadow root (document fragment) which will then cause  the following error
`Uncaught TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.`

https://github.com/videojs/video.js/blob/4238f5c1d88890547153e7e1de7bd0d1d8e0b236/src/js/utils/dom.js#L620

## Specific Changes proposed

Adding try and catch so the code can continue to run despite the error

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/computed-style.js
+++ b/src/js/utils/computed-style.js
@@ -26,7 +26,13 @@ function computedStyle(el, prop) {
   }
 
   if (typeof window.getComputedStyle === 'function') {
-    const computedStyleValue = window.getComputedStyle(el);
+    let computedStyleValue;
+
+    try {
+      computedStyleValue = window.getComputedStyle(el);
+    } catch (_) {
+      return '';
+    }
 
     return computedStyleValue ? computedStyleValue.getPropertyValue(prop) || computedStyleValue[prop] : '';
   }


src/js/utils/computed-style.js
```diff
@@ -26,7 +26,13 @@ function computedStyle(el, prop) {
   }
 
   if (typeof window.getComputedStyle === 'function') {
-    const computedStyleValue = window.getComputedStyle(el);
+    let computedStyleValue;
+
+    try {
+      computedStyleValue = window.getComputedStyle(el);
+    } catch (_) {
```
gkatsev, line 33:
```suggestion
    } catch (e) {
```
We generally use `e` as the catch placeholder, even when we don't use it directly.

gkatsev (approved):
Thanks. LGTM, just one stylistic nit.

src/js/utils/computed-style.js
weiz18: yea, makes sense. just updated.

brandonocasey (approved):

PR #7215, opened 2021-04-29
fix: Don't hide menus with one item and a title

mister-ben:
## Description
Menus with one item are hidden if they have a title, because adding the title increments the hide threshold, but the title is not counted as an item. e.g. the chapters menu if there is a single chapter - `player.addRemoteTextTrack({src: 'data:text/vtt;base64,' + btoa('WEBVTT\n\n00:00:00.000 --> 00:00:05.000\nChapter 1\n'), kind: 'chapters'}, true);`

## Specific Changes proposed
Don't increment `hideThreshold_` for the title. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/menu/menu-button.js
+++ b/src/js/menu/menu-button.js
@@ -120,8 +120,6 @@ class MenuButton extends Component {
         tabIndex: -1
       });
 
-      this.hideThreshold_ += 1;
-
       const titleComponent = new Component(this.player_, {el: titleEl});
 
       menu.addItem(titleComponent);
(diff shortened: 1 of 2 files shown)


gkatsev (approved):
At first, I wasn't sure this is the fix we want, but it makes sense now that I look at it more closely. Since, as you said, the title isn't counted as an item and only items are checked for the hide threshold.
Other menus, like the subs caps button have their own hide threshold since they have the off item and the caption settings item.

PR #7218, opened 2021-05-03
feat: Add option to use full window mode instead of using tech's fullscreen

mister-ben:
## Description
iPhone's native fullscreen isn't always desirable as you can't overlay controls or anything else. Adds an option to prefer "full window" mode over the video el's fullscreen.

## Specific Changes proposed
If the `preferFullWindow` option is set to true, on a browser that does not support the proper fullscreen API but does support fullscreen on the video element (i.e. iPhone), then `requestFullscreenHelper_()` should call `enterFullWindow()` instead of fullscreen on the tech.

## Requirements Checklist
  - [x] Docs/guides updated


gkatsev:
Would be it useful to also have an option that forces full window mode?

mister-ben:
Maybe, I was only considering the case of overlays over "full screen" iOS really but we could cover that scenario too.

test/unit/player-fullscreen.test.js
```diff
@@ -208,6 +208,26 @@ QUnit.test('fullscreenOptions from function args should override player options'
...
+
+  player.requestFullscreen();
+
+  assert.strictEqual(player.isFullscreen(), true, 'player considered fullscreen');
+  assert.strictEqual(player.isFullWindow, true, 'player is full window');
+
+  player.exitFullscreen();
+  assert.strictEqual(player.isFullWindow, false, 'full window is exited');
+  
+    player.dispose();
```
gkatsev, line 228:
```suggestion

  player.dispose();
```

gkatsev (approved):

hasanbuba:
8 Haz 2021 Sal, saat 18:55 tarihinde Gary Katsevman <
***@***.***> şunu yazdı:

PR #7228, opened 2021-05-10
feat(player): Add playbackRates() method

### LINKED ISSUE #7198: Add `playbackRates()` method
Add a new `playbackRates(rates)` method to the player which will update the playback rates used by the playback rate menu. It should also turn the menu button on if it is off.

Currently, the only way to turn it on is by providing a `playbackRates` option during player initialization. This will allow users to update the menu and turn it on after a player has already been created.

gkatsev:
Adds a new playbackRates() method that takes an Array of numbers
representing the rates that are wanted to show up in the playback rates
menu. When new rates are given, a rateschange event will trigger, which
will be used by the PlaybackRatesMenuButton to update itself.

An empty array will hide the menu. No value will return the currently
set playback rates. Other values will be ignored.

Fixes #7198

### DIFF
--- a/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
+++ b/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
@@ -2,7 +2,6 @@
  * @file playback-rate-menu-button.js
  */
 import MenuButton from '../../menu/menu-button.js';
-import Menu from '../../menu/menu.js';
 import PlaybackRateMenuItem from './playback-rate-menu-item.js';
 import Component from '../../component.js';
 import * as Dom from '../../utils/dom.js';
@@ -33,6 +32,7 @@ class PlaybackRateMenuButton extends MenuButton {
 
     this.on(player, 'loadstart', (e) => this.updateVisibility(e));
     this.on(player, 'ratechange', (e) => this.updateLabel(e));
+    this.on(player, 'rateschange', (e) => this.handleRateschange(e));
   }
 
   /**
@@ -78,22 +78,18 @@ class PlaybackRateMenuButton extends MenuButton {
   }
 
   /**
-   * Create the playback rate menu
+   * Create the list of menu items. Specific to each subclass.
    *
-   * @return {Menu}
-   *         Menu object populated with {@link PlaybackRateMenuItem}s
    */
-  createMenu() {
-    const menu = new Menu(this.player());
+  createItems() {
     const rates = this.playbackRates();
+    const items = [];
 
-    if (rates) {
-      for (let i = rates.length - 1; i >= 0; i--) {
-        menu.addChild(new PlaybackRateMenuItem(this.player(), {rate: rates[i] + 'x'}));
-      }
+    for (let i = rates.length - 1; i >= 0; i--) {
+      items.push(new PlaybackRateMenuItem(this.player(), {rate: rates[i] + 'x'}));
     }
 
-    return menu;
+    return items;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -2218,6 +2220,7 @@ class Player extends Component {
       src: '',
       source: {},
       sources: [],
+      playbackRates: [],
```
gkatsev, line 2223: default cache

src/js/player.js
```diff
@@ -4856,6 +4859,45 @@ class Player extends Component {
...
+      return this.cache_.playbackRates;
+    }
+
+    // ignore any value that isn't an array
+    if (!Array.isArray(newRates)) {
+      return;
+    }
+
+    // ignore any arrays that don't only contain numbers
+    if (!newRates.every((rate) => typeof rate === 'number')) {
```
brandonocasey, line 4888: should we log something in these error cases?

src/js/player.js
```diff
@@ -4856,6 +4859,45 @@ class Player extends Component {
...
+
+    this.cache_.playbackRates = newRates;
+
+    /**
+    * fires when the player language change
+    *
+    * @event Player#rateschange
+    * @type {EventTarget~Event}
+    */
+    this.trigger('rateschange');
```
brandonocasey, line 4900: perhaps this should be `playbackratesupdated` as this doesn't signify that a new value was selected. It just shows that the list of values was updated. Also this comment is incorrect.

gkatsev: Oops, forgot to update the jsdoc for it.
Given that `ratechange` is already a thing for when the playback rate changes, I was using that as a template for the new event name. Since the _rates changed_ as part of calling this method. I think this fits with the rest of the DOM naming of things. Maybe `playbackrateschange` to make it clear it's different from `ratechange`? Just the single extra `s` is hard to notice.

src/js/player.js
gkatsev: We can log a warning, do you think it's worth doing here?

brandonocasey: I took another look at the code and It seems for invalid values we may: log an error, throw an error, or do nothing. Might be something to address another time to make it all consistent, but since it seems like doing nothing is the most consistent I guess we should do that.

src/js/player.js
brandonocasey: `playbackrateschange` works for me.

gkatsev: Updated.

src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
```diff
@@ -78,22 +78,18 @@ class PlaybackRateMenuButton extends MenuButton {
...
+  createItems() {
     const rates = this.playbackRates();
+    const items = [];
 
-    if (rates) {
-      for (let i = rates.length - 1; i >= 0; i--) {
-        menu.addChild(new PlaybackRateMenuItem(this.player(), {rate: rates[i] + 'x'}));
-      }
+    for (let i = rates.length - 1; i >= 0; i--) {
+      items.push(new PlaybackRateMenuItem(this.player(), {rate: rates[i] + 'x'}));
```
mister-ben, line 89: This should probably be localised with `rates[i].toLocaleString()`, because the decimal separator in many locales is a comma.  That does complicate things, `PlaybackRateMenuItem` will use `parseFloat()` on this string but that only works with a decimal point regardless of locale. Is it better to pass the float here and have `PlaybackRateMenuItem` set its label to something localised instead? Perhaps the 'x' also should be localisable, e.g. 1.5倍.

gkatsev: Making this more accessible does make sense. I doubt it will affect anyone but changing `rate` from a string to a number would be a technically breaking change, but might be an acceptable soft-breakage. It definitely makes sense for the MenuItem to handle how it wants to do things itself.
Probably should have an issue to discuss this further, I don't think it should be part of this PR.

mister-ben: It could be passed through as it currently is to not be breaking, and the Menu item could still parse and localise it. But yes makes sense to be a separate change.

brandonocasey (approved):

PR #7237, opened 2021-05-16
updated index

AhmedFathy-cyber:
changed the iframe size


gkatsev:
Is there a particular reason to change the iframe size? It looks like the current size fits the default player size well.

PR #7258, opened 2021-06-02
Fix attaching instead of detaching an event listener

boris-petrov:
## Description
I believe this is a bug. In `dispose` event listeners should be detached instead of reattached.

## Specific Changes proposed
Changed `on` to `off`.

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev (approved):
Good catch, thanks!

misteroneill (approved):

PR #7282, opened 2021-06-22
chore: add a code coverage ci workflow

brandonocasey:
## Description
Get code coverage working again.


gkatsev (approved):

gkatsev:
Pleasantly surprised, our code coverage isn't a lot lower than ~80%.

gkatsev:
Enabled codecov's app, so, should have nicer reporting. Enabled it for all repos, so, we don't need to do that when/if we add coverage to other repos.

PR #7287, opened 2021-06-25
feat: Add ended middleware

alex-barstow:
## Description
Pretty straightforward. As far as I can tell, this is the only change needed to add support for ended() getter middleware. One use case for this is for client-side postroll ads, when the ended state of the tech is `true` but the player is still displaying ad content.

### DIFF
--- a/src/js/tech/middleware.js
+++ b/src/js/tech/middleware.js
@@ -188,7 +188,8 @@ export const allowedGetters = {
   played: 1,
   paused: 1,
   seekable: 1,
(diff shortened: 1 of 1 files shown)


gkatsev (approved):

gkatsev:
One thing to note is that this won't affect the play toggle from changing to the replay button on ended because it only listens to the ended event. Otherwise, this works fine.

mrttpz (approved):

PR #7293, opened 2021-06-29
fix: throw error on muted resolution rejection during autoplay

roman-bc-dev:
## Description
When autoplay is set to `any` or `muted` and muted() returns `false`, video.js attempts the following workflow:
1. Call play(). If resolved, trigger `autoplay-success`
2. On rejection, save original muted() value and set it to `true`
3. Call play() again. If resolved, trigger `autoplay-success`
4. On a second rejection, we reset muted() back to the saved value
5. Ideally, the following .then method in the promise chain would be skipped and the `autoplay-failure` event would be triggered in the final .catch.

Since we .catch at multiple points along the promise chain, it's possible for step 4 to occur while still resolving to the following .then method, incorrectly triggering an `autoplay-success` event, instead of the expected `autoplay-failure`.

## Specific Changes proposed
1. Rethrow an error after the muted() value is reset in step 4, causing the following .next method to be skipped and the `autoplay-failure` event to be triggered in the final .catch in the promise chain.
2. Add a try-catch to the mock "rejected" promise used in the autoplay.test suite to synchronize promise chains. This fixed an issue where the new error was being swallowed up by QUnit, which interrupted the final .then().catch() methods and generated false failures.
3. Amended a function name and added some comments for clarity.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -1430,7 +1430,9 @@ class Player extends Component {
       return;
     }
 
-    const muted = () => {
+    // Save original muted() value, set muted to true, and attempt to play().
+    // On promise rejection, restore muted from saved value
+    const resolveMuted = () => {
       const previouslyMuted = this.muted();
 
       this.muted(true);
@@ -1448,21 +1450,24 @@ class Player extends Component {
         return;
       }
 
-      return mutedPromise.catch(restoreMuted);
+      return mutedPromise.catch(err => {
+        restoreMuted();
+        throw new Error(`Rejection at manualAutoplay. Restoring muted value. ${err ? err : ''}`);
+      });
     };
 
     let promise;
 
     // if muted defaults to true
     // the only thing we can do is call play
-    if (type === 'any' && this.muted() !== true) {
+    if (type === 'any' && !this.muted()) {
       promise = this.play();
 
       if (isPromise(promise)) {
-        promise = promise.catch(muted);
+        promise = promise.catch(resolveMuted);
       }
-    } else if (type === 'muted' && this.muted() !== true) {
-      promise = muted();
+    } else if (type === 'muted' && !this.muted()) {
+      promise = resolveMuted();
     } else {
       promise = this.play();
     }
@@ -1473,7 +1478,7 @@ class Player extends Component {
 
     return promise.then(() => {
       this.trigger({type: 'autoplay-success', autoplay: type});
-    }).catch((e) => {
+    }).catch(() => {
       this.trigger({type: 'autoplay-failure', autoplay: type});
     });
   }
--- a/test/unit/autoplay.test.js
+++ b/test/unit/autoplay.test.js
@@ -29,18 +29,25 @@ QUnit.module('autoplay', {
 
     fixture.appendChild(videoTag);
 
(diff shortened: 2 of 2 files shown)


gkatsev (approved):

misteroneill (approved):

gkatsev:
The test does catch this. Without wrapping the `fn()` call in our rejectPromise fake, it fails spectacularly.

brandonocasey (approved):

test/unit/autoplay.test.js
```diff
@@ -29,18 +29,25 @@ QUnit.module('autoplay', {
...
         return this;
       },
       catch(fn) {
-        fn();
+        try {
+          fn();
+        } catch (err) {
+          return this;
+        }
         return this;
```
alex-barstow, line 46: Maybe this (line 46) should return `this.resolvePromise` here instead of `this`? I believe that way if a `catch()` throws it will trigger the next `catch`, and if it doesn't throw it will trigger the next `then()`.

src/js/player.js
```diff
@@ -1448,21 +1450,24 @@ class Player extends Component {
         return;
       }
 
-      return mutedPromise.catch(restoreMuted);
+      return mutedPromise.catch(err => {
+        restoreMuted();
+        throw new Error(`Rejection at manualAutoplay. Restoring muted value. ${err ? err : ''}`);
```
alex-barstow, line 1455: If I remove this line that throws the error, I would expect [these tests](https://github.com/videojs/video.js/blob/d2241ca64f7498f71608d3ec36739f62e593f199/test/unit/autoplay.test.js#L278-L319) to fail since the promise on line [1479](https://github.com/videojs/video.js/pull/7293/files#diff-f92dc431786299ba52ba41d6901e029cc1eeb2f5603653bd03eb4d88a27f5020R1479) should resolve and result in an `'autoplay-success'` event and increment `this.counts.success` rather than `this.counts.failure`, however the test still passes which suggests it isn't capturing the test case exactly. It might be necessary to modify `this.rejectPromise` for these tests. I've shared one idea below but I'm not positive it will work.

test/unit/autoplay.test.js
gkatsev: Returning `this.resolvePromise` here changes the expectations of the `rejectPromise`, this may end up requiring a lot of other test changes that may not be worth it right now.

src/js/player.js
gkatsev: Maybe what we need is a separate promise fake for these two tests?

gkatsev: I tried it locally, and I think it might be too complicated to get it in now.

alex-barstow: In that case how about adding a `TODO` comment to fix these two tests at some point?

gkatsev: Spent some time playing around with sinon's promises, and probably needs a complete rewrite of the test to use that.
I'll create an issue for this.

gkatsev: #7295

alex-barstow (approved):

PR #7299, opened 2021-07-01
fix: properly return promise from requestFullscreen and exitFullscreen

### LINKED ISSUE #7298: calling player.exitFullscreen produces an error and is not caught by promise.catch
Getting `Uncaught (in promise) TypeError: Not in fullscreen mode` when `player.exitFullscreen()` is called.
`exitFullscreen` returns a promise, but attaching `.catch()` to it still keeps the error uncaught.
Ideally, with `player.exitFullscreen().catch((err)=>(console.log(err)))`, we should be able to catch this error.

gkatsev:
In addition to mapping the promise to the executor's resolve, and reject
methods, we should also catch the error in the promise chain that's
created inside exitFullscreenHelper_.

Fixes #7298.


FredZeng:
LGTM

gkatsev:
That's a good question and I think the answer is yes.

brandonocasey (approved):

test/sinon.js
```diff
@@ -0,0 +1,12 @@
...
+const s = document.createElement('script');
+
+// on IE11 and Safari 9, load the last supported sinon version
+if (/(?:MSIE|Trident\/7.0|Version\/9.*Safari)/.test(navigator.userAgent)) {
+  s.src = 'https://unpkg.com/sinon@9.2.4/pkg/sinon-no-sourcemaps.js';
+} else {
+  s.src = '/test/base/node_modules/sinon/pkg/sinon.js';
+}
+
+document.write(s.outerHTML);
```
gkatsev, line 12: evil stuff until we get rid of IE11

brandonocasey: should we only do this in the ie 11/safari 9 if statement?

gkatsev: no, we have to write out the sinon URL in all cases. Once we drop IE11/Safari 9, we can remove this file and only include sinon in karma, since right now, the current version is only served and not included by default (https://github.com/videojs/video.js/pull/7299/files#diff-8cbe9d03b9fa5e04ef3fa3f8be9763779e57a2b026d5c70cb2ed7afcef0d6df5R43). Trying to limit scripts erroring out in IE11.

test/unit/player-fullscreen.test.js
```diff
@@ -280,3 +281,70 @@ QUnit.test('fullscreenerror event from Html5 should pass through player', functi
...
+  assert.rejects(
+    player.requestFullscreen(),
+    new Error('Cannot go fullscreen'),
+    'our promise was rejected'
+  );
+
+  stub.restore();
+});
+
+QUnit[skipOrTest]('requestFullscreen returns a resovled promise if we were fullscreen', function(assert) {
```
brandonocasey, line 306:
```suggestion
QUnit[skipOrTest]('requestFullscreen returns a resolved promise if we were fullscreen', function(assert) {
```

test/unit/player-fullscreen.test.js
```diff
@@ -280,3 +281,70 @@ QUnit.test('fullscreenerror event from Html5 should pass through player', functi
...
+  assert.rejects(
+    player.exitFullscreen(),
+    new Error('Document not active'),
+    'our promise was rejected'
+  );
+
+  stub.restore();
+});
+
+QUnit[skipOrTest]('exitFullscreen returns a resovled promise if we were fullscreen', function(assert) {
```
brandonocasey, line 338:
```suggestion
QUnit[skipOrTest]('exitFullscreen returns a resolved promise if we were fullscreen', function(assert) {
```

brandonocasey (approved):
approved with two spelling suggestions.

gkatsev:
Those typos keep coming back, lol

Issue #6726, opened 2020-06-24
 getPointerPosition returns wrong position if the container is transformed


vakarelov:
## Description
Briefly describe the issue.
The algorithm of getPointerPosition assumes that video container is not subject to scaling and rotation transformations. As a result, it produces the wrong value what such transformations exist. 

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. Add the video container in a div that has a transformation.  See attached file
[video_test.zip](https://github.com/videojs/video.js/files/4827875/video_test.zip)

2. The slider does not provide correct position

#version: 7.7.5

gkatsev:
Yeah, there's definitely an issue and there is a PR that needs to be fixed up that addresses this https://github.com/videojs/video.js/pull/5773

Issue #7248, opened 2021-05-26
Progress bar time tooltip out of bounds on the right


nochev:
## Description
Hello all, I updated `video.js` from `5.20.5` to `7.12.3` and having issue with the time tooltip.

When I moving mouse at the end of progress bar time tooltip going out of bounds. Seems the issue has come from this PR https://github.com/videojs/video.js/pull/5773/. In `TimeTooltip.update` method invoked by the mouse move event `seekBarRect` doesn't contain `right` property and  this produce wrong calculation. When I add the following debug https://screenrec.com/share/Unp5mABVWo I'm seeing this https://screenrec.com/share/IRwJPp8nZ0. On the player `ready` event, object contains all properties https://screenrec.com/share/1Vl4iWBcFq.

## Steps to reproduce
1. Move the mouse to the end of progress bar

### Expected
The time tooltip fitting the bounds as at the left side https://screenrec.com/share/cY6PQmxRs4

### Actual
The time tooltip going out of bounds on the right side https://screenrec.com/share/LXBeKnQm27

## Additional Information
If I replace `findPosition(seekBarEl)` with `getBoundingClientRect(seekBarEl)` in `ProgressControl.handleMouseMove` method problem is solving https://screenrec.com/share/tKp1VvlcG4.
#### videojs
`7.12.3`
#### browsers
Tested on `Version 90.0.4430.212 (Official Build) (64-bit)`
#### OSes
`Windows 7`
### plugins
`videojs-contrib-ads` and `videojs-ima`

efb4f5ff-1298-471a-8973-3d47447115dc:
+1


https://user-images.githubusercontent.com/73130443/123659303-0e93ff80-d822-11eb-8557-58da4a91a62f.mp4

PikachuEXE:
The calculation of `spaceRightOfPoint` is broken by the `undefined` value in `seekBarRect.right` on
https://github.com/videojs/video.js/blob/v7.14.0/src/js/control-bar/progress-control/time-tooltip.js#L75-L76

The `seekBarRect` is from
https://github.com/videojs/video.js/blob/v7.14.0/src/js/control-bar/progress-control/progress-control.js#L75

And it's using [`findPosition`](https://github.com/videojs/video.js/blob/v7.14.0/src/js/utils/dom.js)
Which would never return `right`

By updating the calculation of `spaceRightOfPoint` to `spaceRightOfPoint = seekBarRect.width - seekBarPointPx`
[...]

Issue #1102, opened 2014-03-21
css:transform and scrubber


commonpike:
See a JSBin of the issue here http://jsbin.com/qovayule/1

If transform:scale is applied to the video or any of its parent elements, the scrubber (or in fact any slider elements) don't correctly respond to mouse interaction. In the above jsbin, clicking halfway the scrubbers tracker positions the playhead at the end.

Its not an easy issue and its not limited to videojs. The problem is that getBoundingClientRect returns .. well ... values that don't correspond to the mouse events pageX and pageY. jQueryUI has the same problems, ao.

As an intermediate solution for users that run into this problem, I would suggest to just hide the sliders with css.  If anyone knows how to disable mouse interaction on the progress bar, so we can show it non-interactive, please let us know.

makker:
@heff I also have a same issue as @commonpike. We also have a 'transform: scale(n, n)' style in one of the video elements parents. And this causes the MouseTimeDisplay component, and the slider heads in SeekBar and VolumeBar to position incorrectly in relation to mouse pointer position.  

I have found a way to fix this and working on implementing it to videojs. The thing is that when mouse is over the SeekBar it's position is calculated continuously. And my fix loops through video element parents and gets and multiplies their scale values from the computed styles. I feel this may affect performance and should probably be enabled with a setting. 

In our case, scale is changed (to one of the video's parent elements) when window is resized. So the scale doesn't need to be calculated every time, but only after window size changes. And then it could be saved to a variable. In some other cases it only needs to be calculated once and in some other cases continuously. 

So my suggestion is that there is a new option added to videojs, options.html5.calculateTransformScale with option values 'never' (default value) | 'init' | 'once' | 'windowresize' | 'always' ). Is this ok. Is the option place and name OK? How about the option values?

My fix without options can be found here:
https://github.com/makker/video.js/commit/b90ef1a0f518ae80aec4b6dc3cf4afa7871391f8
https://github.com/makker/video.js/tree/fix-css-transform-issue-1102

makker:
Improved my code so that it now saves the calculated scale on first time it's needed and updates it every time after window is resized.

PR #7308, opened 2021-07-06
Fix mouse time tooltip when moved to far right of progress bar

PikachuEXE:
## Description
Fix mouse time tooltip when moved to far right of progress bar

Linked issue: #7248

## Specific Changes proposed

## Requirements Checklist
- [x] Bug fixed
  - [x] Change has been verified in an actual browser (FreeTube dev version)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/progress-control.js
+++ b/src/js/control-bar/progress-control/progress-control.js
@@ -72,7 +72,7 @@ class ProgressControl extends Component {
     }
 
     const seekBarEl = seekBar.el();
-    const seekBarRect = Dom.findPosition(seekBarEl);
+    const seekBarRect = Dom.getBoundingClientRect(seekBarEl);
     let seekBarPoint = Dom.getPointerPosition(seekBarEl, event).x;
 
     // The default skin has a gap on either side of the `SeekBar`. This means


peepopoggers:
I have this problem too.

gkatsev:
Unfortunately, this change basically reverts the change from #5773 which fixes #6726 and #1102.
With this change, if you set `transform: scale(1.5)` for example, the mouse tooltip positions don't align properly.
We'd love to get #7248 fixed but we wouldn't want it to break other usecases.

phloxic:
@mister-ben, @gkatsev imho this is still needed.

amtins:
@phloxic to help me understand, is the `transform: scale(1.5)` applied to the whole player?

```javascript
.video-js {
  transform: scale(1.5);
}
```

phloxic:
@amtins, thanks for insisting. Indeed I applied the `scale` to the tooltip instead of the container, duh. It did not help that with video.js in 9.3 I saw misaligned tooltips when scaling them. These do not appear in more recent versions, "fix" or no "fix".
So, yes, something is needed when the _container_ is transformed, but also something is needed to avoid the tooltip sticking out on the right side, the same way as on the left side. This backout change helps with symmetric behaviour, but apparently brings back problem with transformed containers.
As an aside: When the container is scaled, the tooltip(s) may stick out both sides.

PR #7329, opened 2021-07-19
fix: prevent control bar clicks/taps with while hidden

brandonocasey:
## Description
1. On mobile tap a menu (such as the captions menu)
2. Wait for the menu to timeout and hide, but remember where the menu's position was on the player
3. Attempt to select one of the menu options, even though they are not visible
4. Notice that the menu item will be selected.

## Specific Changes proposed
Prevent `pointer-events` from being triggered on the control bar while it is hidden or going into a hidden state.

## Requirements Checklist


gkatsev (approved):
Verified that seems to not cause issues for screen readers

PR #7330, opened 2021-07-19
feat(time-ranges): make iteratable if Symbol.iterator exists

gkatsev:
This allows us to call `Array.from()` on TimeRange objects created by Video.js. For example:
```js
var input = [[0, 10], [20, 30]];
var timeRangeObj = videojs.createTimeRanges(input);
var output = Array.from(timeRangeObj);
assert.deepEqual(input, output, "they're the same!");
```

### DIFF
--- a/src/js/utils/time-ranges.js
+++ b/src/js/utils/time-ranges.js
@@ -2,6 +2,7 @@
  * @file time-ranges.js
  * @module time-ranges
  */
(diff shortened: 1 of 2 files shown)


brandonocasey (commented):
seems good to me, could use a  test though.

brandonocasey (approved):

PR #7334, opened 2021-07-22
fix: remove IE8 url parsing workaround

gkatsev:


gkatsev:
Fun, IE11 fails without this in some cases. Making changing how the removed code works causes it to fail to work as well. So, needs more thought here.

src/js/utils/url.js
```diff
@@ -86,8 +73,8 @@ export const parseUrl = function(url) {
     details.protocol = window.location.protocol;
   }
 
-  if (addToBody) {
-    document.body.removeChild(div);
+  if (!details.host) {
+    details.host = window.location.host;
```
gkatsev, line 77: codecov thinks this isn't covered, but it gets covered in IE11. We just only run coverage in chrome.

brandonocasey: I think there is a special comment you can use to skip this if statement.

I think it's `/* istanbul ignore if  */` from https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

brandonocasey (approved):

brandonocasey (approved):

PR #7337, opened 2021-07-22
Fix/remove innerhtml

brandonocasey:
## Description
Remove our usage of innerHTML across the board.

TODO: text-track-settings-menu which requires more changes than all of the rest of the code together. I think that will need it's own pull request.

### DIFF
--- a/src/js/control-bar/audio-track-controls/audio-track-menu-item.js
+++ b/src/js/control-bar/audio-track-controls/audio-track-menu-item.js
@@ -3,7 +3,6 @@
  */
 import MenuItem from '../../menu/menu-item.js';
 import Component from '../../component.js';
-import {assign} from '../../utils/obj';
 
 /**
  * An {@link AudioTrack} {@link MenuItem}
@@ -46,20 +45,28 @@ class AudioTrackMenuItem extends MenuItem {
   }
 
   createEl(type, props, attrs) {
-    let innerHTML = `<span class="vjs-menu-item-text">${this.localize(this.options_.label)}`;
+    const el = super.createEl(type, props, attrs);
+    const parentSpan = super.createEl('span', {
+      className: 'vjs-menu-item-text',
+      textContent: this.localize(this.options_.label)
+    });
 
     if (this.options_.track.kind === 'main-desc') {
-      innerHTML += `
-        <span aria-hidden="true" class="vjs-icon-placeholder"></span>
-        <span class="vjs-control-text"> ${this.localize('Descriptions')}</span>
-      `;
+      const icon = super.createEl('span', {
+        className: 'vjs-icon-placeholder'
+      }, {
+        'aria-hidden': true
+      });
+      const controlText = super.createEl('span', {
+        className: 'vjs-control-text',
+        textContent: this.localize('Descriptions')
+      });
+
+      parentSpan.appendChild(icon);
+      parentSpan.appendChild(controlText);
     }
 
-    innerHTML += '</span>';
-
-    const el = super.createEl(type, assign({
-      innerHTML
-    }, props), attrs);
+    el.appendChild(parentSpan);
 
     return el;
   }
--- a/src/js/control-bar/audio-track-controls/audio-track-menu-item.js
+++ b/src/js/control-bar/audio-track-controls/audio-track-menu-item.js
@@ -47,6 +47,11 @@ class AudioTrackMenuItem extends MenuItem {
 
   createEl(type, props, attrs) {
     const el = super.createEl(type, props, attrs);
+
+    while (el.firstChild) {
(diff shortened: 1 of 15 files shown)


gkatsev:
> TODO: text-track-settings-menu which requires more changes than all of the rest of the code together. I think that will need it's own pull request.

also, that one doesn't use any user-generated content, so, there aren't any issues related to that.

gkatsev (commented):
Reviewed most of these, still have audio track and subs caps menu items to review.

src/js/control-bar/spacer-controls/custom-control-spacer.js
```diff
@@ -28,14 +28,12 @@ class CustomControlSpacer extends Spacer {
    *         The element that was created.
    */
   createEl() {
-    const el = super.createEl({
-      className: this.buildCSSClass()
+    return super.createEl({
+      className: this.buildCSSClass(),
+      // No-flex/table-cell mode requires there be some content
+      // in the cell to fill the remaining space of the table.
+      textContent: '\u00a0'
```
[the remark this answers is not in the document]
brandonocasey: Ok I found the issue here, the `Spacer` class that this inherits from did not pass along props or attributes. After correcting that this works. textContent says that it recommends not using html escape codes and instead using hex/js character codes.

src/js/control-bar/spacer-controls/spacer.js
```diff
@@ -27,10 +27,12 @@ class Spacer extends Component {
    * @return {Element}
    *         The element that was created.
    */
-  createEl() {
-    return super.createEl('div', {
-      className: this.buildCSSClass()
-    });
+  createEl(tag = 'div', props = {}, attributes = {}) {
+    if (!props.className) {
```
brandonocasey, line 31: pass props along

src/js/control-bar/audio-track-controls/audio-track-menu-item.js
```diff
@@ -46,20 +46,31 @@ class AudioTrackMenuItem extends MenuItem {
...
-      innerHTML += `
-        <span aria-hidden="true" class="vjs-icon-placeholder"></span>
-        <span class="vjs-control-text"> ${this.localize('Descriptions')}</span>
-      `;
+    while (el.firstChild) {
+      el.removeChild(el.firstChild);
     }
 
-    innerHTML += '</span>';
+    const parentSpan = super.createEl('span', {
```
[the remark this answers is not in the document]
gkatsev: Actually, looking at `el`, maybe what we need to do here is to get the span that gets created and append to it?
So, not removing stuff from el on line 51/52, then `parentSpan = el.querySelector('.vjs-menu-item-text');`

src/js/control-bar/text-track-controls/subs-caps-menu-item.js
```diff
@@ -14,20 +15,33 @@ import {assign} from '../../utils/obj';
...
+    while (el.firstChild) {
+      el.removeChild(el.firstChild);
     }
 
-    innerHTML += '</span>';
+    const parentSpan = createEl('span', {
+      className: 'vjs-menu-item-text'
+    });
+
+    parentSpan.appendChild(document.createTextNode(this.localize(this.options_.label)));
```
gkatsev, line 28: Seems like there we can do the same as for the audio track menu item. Instead of removing all children, we grab the span that's created and append to it, since it seems like it gets created as we want.

src/js/menu/menu-item.js
```diff
@@ -63,11 +64,21 @@ class MenuItem extends ClickableComponent {
...
     }, props), attrs);
+
+    while (el.firstChild) {
+      el.removeChild(el.firstChild);
+    }
+
+    el.appendChild(createEl('span', {
+      className: 'vjs-menu-item-text',
+      textContent: this.localize(this.options_.label)
+    }));
```
gkatsev, line 79: to do less work, we potentially can replace the first child with this new span and potentially do less work
```js
el.replaceChild(createEl('span', {
  className: 'vjs-menu-item-text',
  textContent: this.localize(this.options_.label)
}, el.firstChild)
```
Based on the output of ClickableComponent and the expected output of MenuItem, that's the only change.

gkatsev: Alternatively, we leave as is, but we should create an issue for this to do a follow-up refactor where we look and see where we can reduce the amount of work we're doing when creating components.

brandonocasey:
going to do some testing with your suggestions.

brandonocasey:
OK @gkatsev I updated this pull request with element replacements where possible.

gkatsev (approved):

gkatsev:
Thanks, seems good to me!

PR #7347, opened 2021-07-26
fix: Correctly dispose of html5 media element for chrome 92

brandonocasey:
## Description
Chrome 92 requires media elements to be torn down completely before they are removed from memory. This includes unsetting the src property and removing the src attribute.

This is only an issue when a lot of players are created and disposed on a page. For us this mostly happens in tests.

The error in question:
Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1322,8 +1322,7 @@ Html5.disposeMediaElement = function(el) {
     el.removeChild(el.firstChild);
   }
 
-  // remove any src reference. not setting `src=''` because that causes a warning
-  // in firefox
+  el.src = '';
   el.removeAttribute('src');
 
   // force the media element to update its loading state by calling load()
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -1322,8 +1322,7 @@ Html5.disposeMediaElement = function(el) {
     el.removeChild(el.firstChild);
   }
 
-  // remove any src reference. not setting `src=''` because that causes a warning
-  // in firefox
+  el.src = '';
```
gkatsev, line 1325: seems like firefox no longer shows a warning here.

Should we add this to `resetMediaElement` below as well? Line 1356

brandonocasey:
Fairly certain at this point the Video.js does not need this change. Dispose appears to work correctly and does not limit our media element creation ability in any cases that I can see.

PR #7349, opened 2021-07-27
feat(hooks): Error hooks

gkatsev:
Adding `beforeerror` and `error` hooks that make it easier to know when errors occurred on all players and allows intercepting and modifying errors.

### DIFF
--- a/docs/guides/hooks.md
+++ b/docs/guides/hooks.md
@@ -106,7 +106,7 @@ videojs('some-id', {autoplay: true, controls: true});
 ### beforeerror
 
 `beforeerror` occurs just as we get an error on the player. This allows plugins or other custom code to intercept the error and modify it to be something else.
-`error` is can be an object such as `{code: 4}` or something with a message. It potentially includes a `cause` property too. Alternatively, it could be `null` which means that the current error will be getting cleared.
+`error` is an object such with a `code` property, an optional `message` property, and potentially a `cause` property too. Alternatively, it could be `null` which means that the current error should be cleared.
 
 `beforeerror` hook functions:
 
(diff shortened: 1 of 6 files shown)


docs/guides/hooks.md
```diff
@@ -99,6 +103,53 @@ videojs.hook('setup', function(player) {
 videojs('some-id', {autoplay: true, controls: true});
 ```
 
+### beforeerror
+
+`beforeerror` occurs just as we get an error on the player. This allows plugins or other custom code to intercept the error and modify it to be something else.
+`error` is can be an object such as `{code: 4}` or something with a message. It potentially includes a `cause` property too. Alternatively, it could be `null` which means that the current error will be getting cleared.
```
brandonocasey, line 109:
```suggestion
`error` can be a string with a message, an object or `MediaError`, a javascript `Error`, a string, or null indicating that error should be cleared.

Examples:
MediaError or Object:
`{code: 4, message: 'something', cause: 'xyz'}`. 

String
`something bad happened`

No error:
`null`

Comment (brandonocasey) [inline docs/guides/hooks.md:109]:
```suggestion
`error` is an object that has a `code` and `message` property and potentially a `cause` property too. Alternatively, it could be `null` which means that the current error should be cleared.
```

src/js/player.js
```diff
@@ -3991,6 +4005,9 @@ class Player extends Component {
      */
     this.trigger('error');
 
+    // notify hooks of the per player error
+    hooks('error').forEach((hookFunction) => hookFunction(this, this.error_));
```
brandonocasey, line 4009: we should use the `err` here so that a hook function does not modify player.error_ and cause weird things to happen.

docs/guides/hooks.md
gkatsev: Looks like error can be a string or just a number too, so, I'll update this and the code based on expectations.

src/js/player.js
gkatsev: `player.error()` returns `player.error_` directly, so, I don't see an issue with returning it here too.

brandonocasey (approved):

PR #7350, opened 2021-07-28
fix: evented should cleanup dom data

brandonocasey:
## Description
Since `evented` is a mixin, it should cleanup after itself. I migrated the DomData cleanup from Component to the mixin dispose event listener.


src/js/mixins/evented.js
```diff
@@ -499,6 +500,11 @@ function evented(target, options = {}) {
   // When any evented object is disposed, it removes all its listeners.
   target.on('dispose', () => {
     target.off();
+    [target, target.el_, target.eventBusEl_].forEach(function(val) {
```
brandonocasey, line 503: handle disposal of any DomData for the evented object.

gkatsev (approved):

Issue #7313, opened 2021-07-08
Video Overlay and Controls are not getting removed when changing quality:


bhargavavarma:
## Description
When changing quality using `videojs-http-source-selector`, the overlay and control bar components are not getting removed.
As per our observation, `listenForUserActivity_` is not getting called sometimes, but the video is playing in the background.
You can check this out in [this](https://codesandbox.io/s/pztxb?file=/index.js) codesandbox link

You can check this in the video attached [here](https://drive.google.com/file/d/1WLFmM7FcSsV8sDp1eBQ8Bh7500JiQcZq/view?usp=sharing) too

## Steps to reproduce
1. Change the video quality and keep the mouse pointer on the video itself.
2. Try to repeat step-1, 2-3 times to reproduce the issue.

### Expected
The controls should be removed after some time automatically.

### Actual
The controls are not getting removed and are staying infinitely until reload.

#### browsers
Chrome 90
#### OSes
Linux and MAC
### plugins
videojs-contrib-quality-levels 2.1.0
videojs-http-source-selector 1.1.6

mister-ben:
Hm, I could reproduce that on the captions menu on your sample too.

gkatsev:
I can reproduce it when I click on the menu button, but not when I just hover to select a choice. Probably something funky related to lock showing.

gkatsev:
Also, I did notice that the codesandbox is depending on `videojs` rather than our package name `video.js`. Though, getting a newer version doesn't seem to help.
Interestingly, I cannot reproduce via https://videojs-preview.netlify.app/sandbox/index.html, on https://videojs-http-streaming.netlify.app/, I cannot reproduce via the captions menu, but I can via the http-source-selector menu. My guess is that it's something related to that plugin, because until I cause it to happen with the source selector, everything seems to be working fine.

Ridham177:
@gkatsev When you reproduce the issue using http-source-selector menu, do you use [vod.dev](https://vod.dev/)? I am able to recreate the issue in the given sandbox but not on the aforementioned site which is a demo for http-source-selector.

Ridham177:
This may be primitive information to mention but `player.userActive()` always returns true after recreating the bug. Should we be explicitly setting state to false? This is my first time working with video.js library and my second OSS issue so any directions are much appreciated.

gkatsev:
I was testing with https://videojs-http-streaming.netlify.app/.
I can't reproduce on vod.dev. Don't really know what the difference is.

[...]

PR #7383, opened 2021-08-18
fix: prevent cached inactivityTimeout from being overwritten

jgcaruso:
Fixes https://github.com/videojs/video.js/issues/7313

## Description
If multiple `mouseenter` events fire on the controlBar before a `mouseleave` event occurs, the cached inactivityTimeout value will get overwitten with a 0. This prevents the control bar from being hidden when the inactivity timeout is reached because the hide call is skipped when inactivityTimeout <= 0.

The circumstances around multiple `mouseenter` events firing before a `mouseleave` event are when a menubutton on the toolbar has its `update()` method called: the menu popup is disposed but a `mouseleave` event isn't fired for it. Once any mouse activity is detected the control bar's `mouseenter` fires again, wiping out the cached value.

An example of this occurring on a plugin https://github.com/FoxCouncil/videojs-max-quality-selector/issues/8

## Specific Changes proposed
This PR adds a check to the caching of the inactivityTimeout value to not set it if the current value is `0`. This is safe since the value is about to be set to `0` below, and the cache will already have the initial value stored in it (if it happened to already be `0`). So no negative change in behaviour should occur.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4141,7 +4141,9 @@ class Player extends Component {
     if (controlBar && !browser.IS_IOS && !browser.IS_ANDROID) {
 
       controlBar.on('mouseenter', function(event) {
-        this.player().cache_.inactivityTimeout = this.player().options_.inactivityTimeout;
+        if (this.player().options_.inactivityTimeout !== 0) {
+          this.player().cache_.inactivityTimeout = this.player().options_.inactivityTimeout;
+        }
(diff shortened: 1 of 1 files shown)


gkatsev (approved):
Thanks, LGTM.
Longer comment here https://github.com/videojs/video.js/issues/7313#issuecomment-901602707

brandonocasey (approved):

PR #7385, opened 2021-08-18
chore: add a release and deploy Github Action

gkatsev:
The release job will do an npm publish and a github-release.
It's based on videojs/.github/github-release.yml.
The new version is published as `next` and requires to be promoted to
latest manually.

The deploy job, should do an aws s3 sync to upload the files to S3 for
us. It uses a Github Environment secrets. This is so that we can require
a user to approve the CDN upload.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   release:
     env:
-      NETLIFY_BASE: 'videojs-preview.netlify.app/sandbox/'
+      NETLIFY_BASE: 'videojs-preview.netlify.app'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -29,6 +29,7 @@ jobs:
         with:
           node-version: '${{steps.nvm.outputs.NVMRC}}'
           cache: npm
+          # this line is required for the setup-node action to be able to run the npm publish below.
           registry-url: 'https://registry.npmjs.org'
 
       - name: npm install
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,92 @@
+name: release
+on:
+  push:
+    tags:
+    # match semver versions
+    - "v[0-9]+.[0-9]+.[0-9]+"
+    # match semver pre-releases
+    - "v[0-9]+.[0-9]+.[0-9]+-*"
+jobs:
+  release:
+    env:
+      NETLIFY_BASE: 'videojs-preview.netlify.app/sandbox/'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        # We neeed to fetch the entire history as conventional-changelog needs
+        # access to any number of git commits to build the changelog.
+        with:
+          fetch-depth: 0
+
+      - name: read node version from .nvmrc
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        shell: bash
+        id: nvm
+
+      - name: setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '${{steps.nvm.outputs.NVMRC}}'
+          cache: npm
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: npm install
+        run: npm i --prefer-offline --no-audit
+
+      # publish runs build for us via a prepublishOnly script
+      - name: npm release
+        run: npm publish --tag next
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: get current changelog
+        run: node -e "console.log(require('./build/current-changelog.js')())" > CHANGELOG-LATEST.md
+
+      - name: get dashed package version for netlify
+        run: echo ::set-output name=VERSION::$(node -e "process.stdout.write(require('./package.json').version.split('.').join('-'))")
+        id: get-version
+        shell: bash
+        if: env.NETLIFY_BASE != ''
+
+      - name: add netlify preview to release notes
+        run: |
+          echo "" >> CHANGELOG-LATEST.md
+          echo "[netlify preview for this version](https://v${{steps.get-version.outputs.VERSION}}--${{env.NETLIFY_BASE}})" >> CHANGELOG-LATEST.md
+        if: env.NETLIFY_BASE != ''
+
+      - name: Create Github release
+        uses: softprops/action-gh-release@v1
+        with:
+          body_path: CHANGELOG-LATEST.md
+          token: ${{github.token}}
+          prerelease: true
+          files: dist/*.zip
+          discussion_category_name: Releases
+
+  deploy:
+    needs: release
+    runs-on: ubuntu-latest
+    environment: Deploy
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+        run: npm i --prefer-offline --no-audit
+
+      # publish runs build for us via a prepublishOnly script
+      - name: npm release
+        run: npm publish --tag next
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: get current changelog
+        run: node -e "console.log(require('./build/current-changelog.js')())" > CHANGELOG-LATEST.md
```
gkatsev, line 44: Decided to continue using what we have locally for now until we deprecate releasing locally with the gh-release or update it to use the same thing that's in videojs/.github/github-release.yml

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: npm install
+        run: npm i --prefer-offline --no-audit
+
+      # publish runs build for us via a prepublishOnly script
+      - name: npm release
+        run: npm publish --tag next
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
gkatsev, line 41: we should add this to VHS too.

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+          node-version: '${{steps.nvm.outputs.NVMRC}}'
+          cache: npm
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: npm install
+        run: npm i --prefer-offline --no-audit
+
+      # publish runs build for us via a prepublishOnly script
+      - name: npm release
+        run: npm publish --tag next
```
gkatsev, line 39: we always want to publish to the next tag by default and then later on someone can manually upgrade the version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_S3_SECRET }}
+          aws-region: us-east-1
+
+      - name: Copy files to the CDN with the AWS CLI
+        run: |
+          aws s3 sync node_modules/video.js/dist/ s3://${S3_KEY}/${VJS_VERSION}/ --acl ${S3_ACCESS}
```
gkatsev, line 92: this should upload to a full version specific folder, i.e. `7.15.0/video.js`. I may create a `promote` workflow that can set the npm tag to latest and also upload it to s3 under the minor, i.e. `7.15/video.js` (and maybe major too, i.e. `7/video.js`).

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+          prerelease: true
+          files: dist/*.zip
+          discussion_category_name: Releases
+
+  deploy:
+    needs: release
+    runs-on: ubuntu-latest
+    environment: Deploy
+    env:
+      VJS_VERSION: ${{ github.event.head_commit.message }}
```
gkatsev, line 72: We have it set up so that the commit message here is the full version of Video.js without any preceding `v`, so, we can use it directly as the VJS_VERSION.

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+          body_path: CHANGELOG-LATEST.md
+          token: ${{github.token}}
+          prerelease: true
+          files: dist/*.zip
+          discussion_category_name: Releases
+
+  deploy:
+    needs: release
+    runs-on: ubuntu-latest
+    environment: Deploy
```
gkatsev, line 70: This env is what makes the AWS secrets available. It also requires a manual approval from me before it'll run.

.github/workflows/release.yml
gkatsev: While this should work, I wasn't able to test it directly, so, our first release with this will likely be also our test. I was able to test the gh and npm release in a test project and those worked flawlessly.

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+        if: env.NETLIFY_BASE != ''
+
+      - name: Create Github release
+        uses: softprops/action-gh-release@v1
+        with:
+          body_path: CHANGELOG-LATEST.md
+          token: ${{github.token}}
+          prerelease: true
+          files: dist/*.zip
+          discussion_category_name: Releases
```
gkatsev, line 65: Create a discussion from the release under the Releases category.

.github/workflows/release.yml
gkatsev: Creating a new task would be cool, because we could potentially automate updating the GH Release to no longer be a pre-release either.

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        shell: bash
+        id: nvm
+
+      - name: setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '${{steps.nvm.outputs.NVMRC}}'
+          cache: npm
+          registry-url: 'https://registry.npmjs.org'
```
gkatsev, line 32: this line is required for the setup-node action to be able to run the npm publish below.

.github/workflows/release.yml
```diff
@@ -0,0 +1,92 @@
...
+  push:
+    tags:
+    # match semver versions
+    - "v[0-9]+.[0-9]+.[0-9]+"
+    # match semver pre-releases
+    - "v[0-9]+.[0-9]+.[0-9]+-*"
+jobs:
+  release:
+    env:
+      NETLIFY_BASE: 'videojs-preview.netlify.app/sandbox/'
```
brandonocasey, line 12: I think we actually want to link to the index with all the other demos here?
```suggestion
      NETLIFY_BASE: 'videojs-preview.netlify.app'
```

.github/workflows/release.yml
brandonocasey:
```suggestion
          # this line is required for the setup-node action to be able to run the npm publish below.
          registry-url: 'https://registry.npmjs.org'
```

.github/workflows/release.yml
brandonocasey: Yeah I could see benefits to us publishing in ci, perhaps we even make the build depend upon the ci integration passing?

.github/workflows/release.yml
brandonocasey: Do we want `deploy` to run on semver prereleases?

gkatsev: We haven't previously, but I don't see an issue with doing so. We also haven't done semver pre-releases in Video.js in a long time.

.github/workflows/release.yml
gkatsev: _I guess so_

brandonocasey (approved):

PR #7449, opened 2021-09-28
fix: remove rule on small layout

ipadilla4:
Previously we introduced _!important_ rule to override BC skin properties and fix the layout for small screens. I've done several tests in Chrome, Firefox and this is not longer needed, so we can safely remove it.


gkatsev (approved):
Thanks!

PR #7456, opened 2021-09-29
fix(package): update to VHS 2.11.0

harvister:
update to VHS 2.11.0
fix: Only check for bad seeks after seeking and an append

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1312,18 +1312,49 @@
       }
     },
     "@videojs/http-streaming": {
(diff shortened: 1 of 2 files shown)


gkatsev:
Thanks! Unfortunately, when updating VHS we need to also update VHS's dependencies in Video.js now too (like mux.js and mpd-parser). I've created https://github.com/videojs/video.js/pull/7459 to do the update.

Issue #6699, opened 2020-06-09
Failed to execute 'replaceChild' on 'Node': The node to be replaced is not a child of this node. 


travisbader:
## Description
Briefly describe the issue.
Include a [reduced test case](https://css-tricks.com/reduced-test-cases/), we have a [starter template](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0) on JSBin you can use.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1. I am not sure the exact steps to repro, but I theorize this crash is surrounding disposing of the video player and the updating of the remaining time

### Expected
Don't throw errors

### Actual
The error is thrown

### Error output
![Screen Shot 2020-06-09 at 1 58 57 PM](https://user-images.githubusercontent.com/7252621/84199068-6a45ec00-aa59-11ea-917d-a3972bbc5b5c.png)
## Additional Information
Please include any additional information necessary here. Including the following:
#### videojs
7.7.6
#### browsers
Chrome for sure, haven't seen in other browsers yet
#### OSes
unknown
### plugins
    "@silvermine/videojs-airplay": "^1.0.0",
    "@silvermine/videojs-chromecast": "^1.2.1",
    "videojs-contrib-quality-levels": "^2.0.9",
    "videojs-hotkeys": "^0.2.25",
    "videojs-vtt.js": "^0.12.6"

travisbader:
PR submitted

gkatsev:
Like https://github.com/videojs/video.js/issues/5706, it would be good to figure out the root cause. Often times these issues and fixes with null checks just hide a worse problem that would be better if it were fixed.

dansgithub:
Hey everyone, is there any update on this issue? Currently we have many error logs coming from this one and there seems to be no fix!

JustinMGaiam:
This is also our highest error rate in our NewRelic Browser monitoring (we use VideoJS through the Brightcove player) so anything that moves towards fixing this error would be very helpful.

travisbader:
Now three users all reporting high quantity of errors.  Im ok if you don't merge my naive fix for reasons of wanting to find the root cause, but can you please find the root cause! We still have an enormous number of errors emitting out of this and would appreciate any attention you can give it

gkatsev:
Are all of you using newrlic for error tracking?

Also, can you give me some more info about your environments? I've tried to reproduce it a bit but I've been unable to.

travisbader:
[...]

PR #7474, opened 2021-10-14
fix: don't let the player be translated except captions

gkatsev:
This is another follow-up to #6699.

Potentially, it means we could get rid of #6977

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -803,6 +803,8 @@ class Player extends Component {
     // if it's been set to something different to the doc
     this.el_.setAttribute('lang', this.language_);
 
+    this.el_.setAttribute('translate', 'no');
+
     this.el_ = el;
 
     return el;
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -218,6 +218,7 @@ class TextTrackDisplay extends Component {
     return super.createEl('div', {
       className: 'vjs-text-track-display'
     }, {
+      'translate': 'yes',
       'aria-live': 'off',
       'aria-atomic': 'true'
     });


gkatsev:
@OwenEdwards yeah, I was thinking the scenario of only Spanish captions without English captions. Theoretically, English->English in that case would be a no-op.
https://github.com/videojs/video.js/pull/7493 makes sense to have.

The items in the text track menu won't get translated because they're covered by the main `translate=no`.

OwenEdwards:
@gkatsev you wrote:
> The items in the text track menu won't get translated because they're covered by the main `translate=no`.

Wouldn't it make sense to put `translate="yes"` on those items, and on the content of the Captions Settings dialog?

Broadly, I understand the #6699 identifies a specific case where using Google Translate to translate a page that includes video.js causes a problem, and applying `translate="no"` to the whole player is a quick fix. Longer-term, could we be more targeted about what gets translated and what doesn't?

gkatsev:
> Wouldn't it make sense to put `translate="yes"` on those items, and on the content of the Captions Settings dialog?

Maybe, if we can get just the text isolated enough. Since, we wouldn't want to introduce issues where translate replaces the elements and using the menu doesn't work anymore.

Yup, applying `translate=no` to the entire player is definitely the quicker fix, and we can definitely enable more elements to be available as `translate=yes`, though, I think time is better spent on making our translation support better and more usable so that people don't need to go through this translation thing.

If there's no objections with proceeding with this PR, I'd like to merge it in.

gkatsev:
Also, I guess the alternative is to only do `translate=no` on the player right now and then audit the player to see where we can/should do `translate=yes`

OwenEdwards:
LGTM.

Zyhazeeb:
Hi friends

PR #7482, opened 2021-10-27
Update preset env, drop IE11 and older browser support

gkatsev:


misteroneill:
Replaced by #7701 and #7708

PR #7510, opened 2021-11-10
fix: improve enabling liveui when switching sources

gkatsev:
We try and enable the liveui on `canplay`, however, we only do it the first time after the `LiveTracker` is enabled. This means that if you change sources, we may not catch that the liveui should be enabled. This is particularly important for browsers where native playback is used, like Safari, as the metadata may not be available until `canplay`.

This is a follow-up from https://github.com/videojs/video.js/pull/7114 which enabled listening to `canplay` but apparently forgot to account for switching videos in the same player.


gkatsev:
Currently, have a failing test that should then pass once I push the actual change once the build fails.

misteroneill (approved):

PR #7514, opened 2021-11-12
fix: try again on volume feature detection on iOS

### LINKED ISSUE #7040: Can't set volume when using iPhone without full screen
## Description
When I use **iPhone** to watch videojs player on all browsers like chrome, safari, etc.  I **can** set volume when player is full screen, but I **can't** set volume when player is **NOT** full screen. I can't drag and drop the dot on volume bar to set volume. It always stay at default volume. **Android** phone can set volume smoothly. Whether it is full screen or not full screen

## Steps to reproduce
1. use iPhone and go to videojs official demo: https://videojs.com/advanced/?video=disneys-oceans
2. try to set volume, and you can't set volume.

gkatsev:
On latest iOS, we are seeing times when the volume feature detection is
showing that we are able to change the volume, though, that is not the
case. Instead, on iOS, when we detect that we can control the volume, we
set a short timer to retest and reset the featuresVolumeControl property.

Fixes #7040

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -1032,7 +1032,20 @@ Html5.canControlVolume = function() {
     const volume = Html5.TEST_VID.volume;
 
     Html5.TEST_VID.volume = (volume / 2) + 0.1;
-    return volume !== Html5.TEST_VID.volume;
+
+    const canControl = volume !== Html5.TEST_VID.volume;
+
+    // on latest iOS, there are cases where we read the volume as changed.
+    // In those cases, we should add a timeout and check again later.
+    // Since features doesn't currently work asynchronously,
+    // we then have to manually set the new value.
+    if (canControl && browser.IS_IOS) {
+      setTimeout(() => {
+        Html5.prototype.featuresVolumeControl = volume !== Html5.TEST_VID.volume;
+      });
+    }
+
+    return canControl;
   } catch (e) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


src/js/tech/html5.js
```diff
@@ -1032,7 +1032,20 @@ Html5.canControlVolume = function() {
...
-    return volume !== Html5.TEST_VID.volume;
+
+    const canControl = volume !== Html5.TEST_VID.volume;
+
+    // on latest iOS, there are cases where we read the volume as changed.
+    // In those cases, we should add a timeout and check again later.
+    // Since features doesn't currently work asynchronously,
+    // we then have to manually set the new value.
+    if (canControl && browser.IS_IOS) {
+      setTimeout(() => {
```
misteroneill, line 1043: I don't have a strong opinion on this, but should this use `window.setTimeout` instead?

gkatsev: oops, technically yes.

gkatsev: updated

src/js/tech/html5.js
```diff
@@ -1032,7 +1032,20 @@ Html5.canControlVolume = function() {
...
 
     Html5.TEST_VID.volume = (volume / 2) + 0.1;
-    return volume !== Html5.TEST_VID.volume;
+
+    const canControl = volume !== Html5.TEST_VID.volume;
+
+    // on latest iOS, there are cases where we read the volume as changed.
+    // In those cases, we should add a timeout and check again later.
+    // Since features doesn't currently work asynchronously,
+    // we then have to manually set the new value.
```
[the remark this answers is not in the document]
gkatsev: Sounds good to me.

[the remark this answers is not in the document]
gkatsev: Updated it manually to adjust positioning

gesinger (approved):

PR #7527, opened 2021-11-15
fix: don't always use fastSeek when available.

### LINKED ISSUE #6988: Seeking via tab on seekBar does not work properly on iOS
## Description
When seeking via tab on the seekBar, currentTime is not set correctly.
Player must be in fullscreen mode to see the problem. The problem occurs from version 7.9.0 and above.

video-js-7 9 0-not-working-properly

## Steps to reproduce

Tested on iOS iPad Simulator with iOS 13.7

Created two code pens:

### Working properly:
video.js v7.8.4
https://codepen.io/nimoatwoodway/full/NWRRRpo

### Working not properly:
video.js v7.11.0
https://codepen.io/nimoatwoodway/full/WNGGGEW

Open in fullscreen and tab on the seek bar. Then you will see the offset. In the codepen the offset is much smaller than in my example in the description but you can see it.

### LINKED ISSUE #7360: Dash fastSeek issue in safari
## Description
When in safari, we are encountering an issue where the progress bar does not work properly when using Dash. While playing the video, when we click in the progress bar and then click before the current time, the video playback stops and shows strange behaviour. I included a potential fix and reproduction steps here: https://codepen.io/MarcoTradecast/pen/ZEKoJYb. For the fix I disabled fastSeek only for safari, since fastSeek is also possible for Firefox.

## Steps to reproduce
1. Open https://codepen.io/MarcoTradecast/pen/ZEKoJYb in Safari.
2. Change/revert the code as described in the comment.
3. Play the video and click (no dragging) to around 70% percent of the video in the progress bar.
4. Now click near the beginning of the video in the progress bar (once again no dragging).

### Expected
The video should go back to the desired timestamp.

### Actual
Video playback stops and currentTime is not being updated. Play/pause button is also not responding anymore.

### Error output
There are no errors

#### videojs
7.13.3
#### browsers
Safari (in my case 14.1.1)
#### OSes
macOS Big Sur 11.4
### plugins
No additional plugins are being used in the example.

I hope this issue is clear enough to reproduce :-)

gkatsev:
We were always setting `scrubbing(true)` on mouse down. This means, that
we'd use `fastSeek` even when seeking while clicking, rather than only
when scrubbing.

The main fix involves knowing in `handleMouseMove` whether we were
called directly as a `mousemove` handler or whether it was called from
`handleMouseDown`. This means that when `handleMouseMove` is called via
`handleMouseDown` we can skip setting `scrubbing(true)` and only do it
when we are scrubbing directly.

Fixes #6988, fixes #7360

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -256,7 +256,6 @@ class SeekBar extends Slider {
 
     // Stop event propagation to prevent double fire in progress-control.js
     event.stopPropagation();
-    this.player_.scrubbing(true);
 
     this.videoWasPlaying = !this.player_.paused();
     this.player_.pause();
@@ -269,13 +268,19 @@ class SeekBar extends Slider {
    *
    * @param {EventTarget~Event} event
    *        The `mousemove` event that caused this to run.
+   * @param {boolean} mouseDown this is a flag that should be set to true if `handleMouseMove` is called directly. It allows us to skip things that should not happen if coming from mouse down but should happen on regular mouse move handler.
    *
    * @listens mousemove
    */
-  handleMouseMove(event) {
+  handleMouseMove(event, mouseDown) {
     if (!Dom.isSingleLeftClick(event)) {
       return;
     }
+
+    if (!mouseDown && !this.player_.scrubbing()) {
+      this.player_.scrubbing(true);
+    }
+
     let newTime;
     const distance = this.calculateDistance(event);
     const liveTracker = this.player_.liveTracker;
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -268,11 +268,11 @@ class SeekBar extends Slider {
    *
    * @param {EventTarget~Event} event
    *        The `mousemove` event that caused this to run.
-   * @param {boolean} mouseDown this is a flag that should be set to true if `handleMouseMove` is called directly. It allows us to skip things that should not happen if coming from mouse down but should happen on regular mouse move handler.
+   * @param {boolean} mouseDown this is a flag that should be set to true if `handleMouseMove` is called directly. It allows us to skip things that should not happen if coming from mouse down but should happen on regular mouse move handler. Defaults to false
    *
    * @listens mousemove
    */
-  handleMouseMove(event, mouseDown) {
+  handleMouseMove(event, mouseDown = false) {
     if (!Dom.isSingleLeftClick(event)) {
       return;
     }
--- a/src/js/slider/slider.js
+++ b/src/js/slider/slider.js
@@ -180,7 +180,7 @@ class Slider extends Component {
     this.on(doc, 'touchmove', this.handleMouseMove_);
     this.on(doc, 'touchend', this.handleMouseUp_);
 
-    this.handleMouseMove(event);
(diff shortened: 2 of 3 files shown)


src/js/slider/slider.js
```diff
@@ -180,7 +180,7 @@ class Slider extends Component {
     this.on(doc, 'touchmove', this.handleMouseMove_);
     this.on(doc, 'touchend', this.handleMouseUp_);
 
-    this.handleMouseMove(event);
+    this.handleMouseMove(event, true);
```
gesinger, line 183: If it's something we may consider for the future, it may be worth adding a TODO for splitting the seeking logic out of the mouse move handler and calling it separately from here.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -269,13 +268,19 @@ class SeekBar extends Slider {
...
    * @listens mousemove
    */
-  handleMouseMove(event) {
+  handleMouseMove(event, mouseDown) {
     if (!Dom.isSingleLeftClick(event)) {
       return;
     }
+
+    if (!mouseDown && !this.player_.scrubbing()) {
+      this.player_.scrubbing(true);
```
gesinger, line 281: It would be worth adding a test for this behavior.

gkatsev: At first, I thought it would be too complicated to do, which it wasn't, but it was a bit tricker to get a good working test for this since we listen to mousemove on the document.

gkatsev: tests now passing.

src/js/slider/slider.js
```diff
@@ -192,6 +192,7 @@ class Slider extends Component {
    * @param {EventTarget~Event} event
    *        `mousedown`, `mousemove`, `touchstart`, or `touchmove` event that triggered
    *        this function
+   * @param {boolean} mouseDown this is a flag that should be set to true if `handleMouseMove` is called directly. It allows us to skip things that should not happen if coming from mouse down but should happen on regular mouse move handler.
```
gesinger, line 195:
```suggestion
   * @param {boolean} mouseDown this is a flag that should be set to true if `handleMouseMove` is called directly. It allows us to skip things that should not happen if coming from mouse down but should happen on regular mouse move handler. Defaults to false.
```

test/unit/controls.test.js
```diff
@@ -141,6 +143,35 @@ QUnit.test('calculateDistance should use changedTouches, if available', function
...
 
+QUnit.test("SeekBar doesn't set scrubbing on mouse down, only on mouse move", function(assert) {
+  const player = TestHelpers.makePlayer();
+  const scrubbingSpy = sinon.spy(player, 'scrubbing');
+  const seekBar = new SeekBar(player);
+  const doc = new EventTarget();
+
+  // mousemove is listened to on the document.
+  // Specifically, we check the ownerDocument of the seekBar's bar.
+  // Therefore, we want to mock it out to be able to trigger mosuemove
```
gesinger, line 154:
```suggestion
  // Therefore, we want to mock it out to be able to trigger mousemove
```

src/js/slider/slider.js
gkatsev: Created https://github.com/videojs/video.js/issues/7531 to follow up on this.

gesinger (approved):

PR #7578, opened 2021-12-23
Introduce an option to allow disabling of extra event data being added to events triggered by advanced plugins

evanfarina:
## Description
Today, custom events that are triggered by advanced plugins via `Plugin.trigger` include three additional fields, populated by video.js: `name`, `instance`, `plugin`. In some cases, having these fields auto-included can be confusing, and in other cases problematic (e.g name collision). I provide a library that allows consumers to create plugins and often have to explain to these consumers why there are unexpected fields in their event payloads. This can be solved by making these fields optional.

## Specific Changes proposed
I've introduced a new option, `includeExtraPluginEventData`, which is `false` by default for backward compatibility. When enabled, the `name`, `plugin` and `instance` fields will not be auto-included in event payloads emitted by calls to `Plugin.trigger`.

A couple of notes/design decisions that I'd like feedback on from the community:
1. This change only applies to advanced plugins. I made this decision since consumers likely are not interacting with the default events emitted by basic plugins and, if they are, should be fine with whatever system information is included in those payloads. Custom events, on the other hand, should provide a payload that matches that which the user created.
2. This option is implemented at the player level, as opposed to being enabled by each plugin. I felt that this API was easier to understand and simpler to implement. While it does add overhead to the global options object, it will make configuration simpler for the end-user.

## Requirements Checklist
  - [ x ] Change has been verified in an actual browser (Chrome, Firefox, IE)
  - [  x ] Unit Tests updated or fixed
  - [ x ] Docs/guides updated

### DIFF
--- a/docs/guides/options.md
+++ b/docs/guides/options.md
@@ -45,6 +45,7 @@
   * [suppressNotSupportedError](#suppressnotsupportederror)
   * [techCanOverridePoster](#techcanoverrideposter)
   * [techOrder](#techorder)
+  * [includeExtraPluginEventData](#includeextraplugineventdata)
   * [userActions](#useractions)
   * [userActions.click](#useractionsclick)
   * [userActions.doubleClick](#useractionsdoubleclick)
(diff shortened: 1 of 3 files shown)


misteroneill:
Thanks for taking the time to open a PR!

I don't think we can merge it, though, because the two argument signature is universal for all Video.js events. Most do not include the second argument, but it is available in all events. See: [EventListener](https://docs.videojs.com/eventtarget#~EventListener)

It doesn't feel right to implement an exception just for plugins.

PR #7588, opened 2022-01-02
feat: Change addRemoteTextTrack's manualCleanup option default value to false

alex-barstow:
## Description
This changes the default behavior of `player.addRemoteTextTrack()` so text tracks will now automatically be removed on source changes without needing to pass a `manualCleanup: false` option. This also removes a deprecation warning and updates documentation and tests.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4319,23 +4319,20 @@ class Player extends Component {
 
   /**
    * Create a remote {@link TextTrack} and an {@link HTMLTrackElement}.
-   * When manualCleanup is set to false, the track will be automatically removed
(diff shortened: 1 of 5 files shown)


misteroneill (changes requested):
Overall, looks fine, but does look like code coverage went down.

Also, this will be backward incompatible, so will have to go in Video.js 8 - we'll need to add the `BREAKING CHANGE` flag to the commit message.

misteroneill (approved):
I think we're good to merge this. I don't see why the coverage went down and it doesn't look like there's any code paths here that aren't tested. :shipit:

PR #7609, opened 2022-01-14
docs: Add some FAQ entries

mister-ben:
Minor doc additions for things commonly encountered in Slack/SO etc

### DIFF
--- a/docs/guides/faq.md
+++ b/docs/guides/faq.md
@@ -41,7 +41,7 @@
 * [Q: Can Video.js be required in node.js?](#q-can-videojs-be-required-in-nodejs)
 * [Q: Does Video.js work with webpack?](#q-does-videojs-work-with-webpack)
 * [Q: Does Video.js work with react?](#q-does-videojs-work-with-react)
-* [Q: Can the big play button be centred?](#q-can-the-big-play-button-be-centred)
+* [Q: Can the big play button be centered?](#q-can-the-big-play-button-be-centered)
 * [Q: Can the big play button be shown when paused?](#q-can-the-big-play-button-be-shown-when-paused)
 * [Q: Why is the picture-in-picture button different in Firefox?](#q-why-is-the-picture-in-picture-button-different-in-firefox)
 
@@ -276,7 +276,7 @@ Yes! See the [Webpack and Video.js configuration guide][webpack-guide].
 
 Yes! See [ReactJS integration example][react-guide].
 
-## Q: Can the big play button be centred?
+## Q: Can the big play button be centered?
 
 The default skin offsets the button to not obscure the poster image, but just add a `vjs-big-play-centered` class to the player to have it centred.
 
--- a/docs/guides/faq.md
+++ b/docs/guides/faq.md
@@ -41,6 +41,8 @@
 * [Q: Can Video.js be required in node.js?](#q-can-videojs-be-required-in-nodejs)
 * [Q: Does Video.js work with webpack?](#q-does-videojs-work-with-webpack)
 * [Q: Does Video.js work with react?](#q-does-videojs-work-with-react)
+* [Q: Can the big play button be centred?](#q-can-the-big-play-button-be-centred)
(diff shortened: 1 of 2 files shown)


docs/guides/troubleshooting.md
```diff
@@ -89,6 +90,24 @@ a component.
...
+  const myPlayer = videojs('my_video');
+</script>
+```
+
+Here you might expect the player to be intitialised with the `autoplay` option set, but it will not be. While Video.js sets up player embeds that have a `data-setup` attribute automatically when the page is loaded, in this case the player will have _already been setup_ in the script tag. The `data-setup` option will never be applied.
+
+This can be more confusing if `videojs('my_video')` is used in an async script as different behaviour will occur def=pending on when the script is executed.
+
+It's better to not use `data-setup` and use the `videojs()` constructor once to set up the player, and/or only use the explicit getter `videojs.getPlayer('my_video')` to get a plyer reference.
+
```
OwenEdwards, line 110: Typo: should be "... to get a _player_ reference."

docs/guides/faq.md
```diff
@@ -273,6 +276,18 @@ Yes! See the [Webpack and Video.js configuration guide][webpack-guide].
 
 Yes! See [ReactJS integration example][react-guide].
 
+## Q: Can the big play button be centred?
```
misteroneill, line 279:
```suggestion
## Q: Can the big play button be centered?
```

docs/guides/faq.md
```diff
@@ -41,6 +41,9 @@
 * [Q: Can Video.js be required in node.js?](#q-can-videojs-be-required-in-nodejs)
 * [Q: Does Video.js work with webpack?](#q-does-videojs-work-with-webpack)
 * [Q: Does Video.js work with react?](#q-does-videojs-work-with-react)
+* [Q: Can the big play button be centred?](#q-can-the-big-play-button-be-centred)
```
misteroneill, line 44:
```suggestion
* [Q: Can the big play button be centered?](#q-can-the-big-play-button-be-centered)
```

docs/guides/troubleshooting.md
```diff
@@ -89,6 +90,24 @@ a component.
 
 To fix this issue please make sure that all event listeners are cleaned up on dispose.
 
+## Problems with setup options
+
+If a player is inititalised without the expected setup options, it's usually because of using both the `data-setup` attriubte and the `videojs()` constructor. Take this example:
```
misteroneill, line 95:
```suggestion
If a player is inititalized without the expected setup options, it's usually because of using both the `data-setup` attribute and the `videojs()` constructor. Take this example:
```

docs/guides/troubleshooting.md
```diff
@@ -89,6 +90,24 @@ a component.
...
+
+```html
+<video-js id="my_video" data-setup='{"autoplay": "any"}'></video-js>
+
+<script>
+  const myPlayer = videojs('my_video');
+</script>
+```
+
+Here you might expect the player to be intitialised with the `autoplay` option set, but it will not be. While Video.js sets up player embeds that have a `data-setup` attribute automatically when the page is loaded, in this case the player will have _already been setup_ in the script tag. The `data-setup` option will never be applied.
```
misteroneill, line 105:
```suggestion
Here you might expect the player to be initialized with the `autoplay` option set, but it will not be. While Video.js sets up player embeds that have a `data-setup` attribute automatically when the page is loaded, in this case the player will have _already been setup_ in the script tag. The `data-setup` option will never be applied.
```

docs/guides/troubleshooting.md
```diff
@@ -89,6 +90,24 @@ a component.
...
+
+This can be more confusing if `videojs('my_video')` is used in an async script as different behaviour will occur def=pending on when the script is executed.
```
misteroneill, line 107:
```suggestion
This can be more confusing if `videojs('my_video')` is used in an async script as different behavior will occur depending on when the script is executed.
```

docs/guides/troubleshooting.md
gkatsev: should probably be `initialized`

PR #7616, opened 2022-01-20
play promise error prevention

renebaudisch:
## Description
If the vast plugin starts a vast ad it pauses the player.
Because the player before has applied a play promise the browser will throw the error:
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().

## Specific Changes proposed
To not having an error displayed in the console I added a catch pipe to the play promise to handle the thrown error

## Requirements Checklist

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2345,7 +2345,7 @@ class Player extends Component {
     if (PromiseClass) {
       return new PromiseClass((resolve) => {
         this.play_(resolve);
-      });
+      }).catch();
     }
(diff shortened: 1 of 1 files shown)


misteroneill (commented):
I'm not sure we want to silence all play promise warnings. Is this something that might be better addressed in the VAST plugin?

misteroneill:
Thanks for the PR, but this is something that belongs in a plugin or app, not the core library.

Issue #7577, opened 2021-12-22
userAction.click doesn't work VideoJS v7.17.0


DJ-PD:
Using VideoJS as an audio player with use of an mp3 live stream source via the <audio> element.
Trying to inactivate the player window play/pause toggle function with use of the new function "userAction.click"
Shouldn't I be able to set this to false via the player options "userActions" as below example?
Tried several different ways but I can't get it to work.
`                            
options = {	                       
controls: true,                         
autoplay: false,                          
muted: false,														
**userActions: { click: false },**                         
preload: 'none',                          
poster: [ 'img/pd-stream1920x1080.jpg' ],	                        
techOrder: [ 'chromecast', 'html5' ],`

mister-ben:
That should work, have you a live example where it is not?

DJ-PD:
Hmm, interesting, I tried again both on local as well on the public server and everything works without the userAction false function. I don't want to share the link public here but if you mail me at neccen@gmail.com I will share it with you. We will go live over the Christmas 24 and 25th.

DJ-PD:
Made a new audio test and regardless if I use videoJS version 7.16, 7.17 or 7.18 via a mp3 live stream or just a mp3 file with or without the option `userActions: { click: false }` it seems like the poster option `poster: [ 'img/pd-stream1920x1080.jpg' ]` takes over, see below print screen...
![image](https://user-images.githubusercontent.com/46725109/147291475-9628abb3-5a13-4590-81a4-f784d9d993cc.png)

mister-ben:
Are you using an `<audio>` element? `userActions.click` affects clicks on the playback el, but you're clicking on the `posterImage` component instead.

You can either use a `<video>` instead (which can play audio), or remove the click handler from the poster with `player.posterImage.off('click')`

DJ-PD:
Yes, I use the `<audio>` element. Made a new test via the `<video>` element as you mentioned and that works good in combination of the option `userActions: { click: false }` even with a poster. Downside is that I lose the all time activated control bar which I (users) see as a benefit when just playing audio. Can I trigg the control bar to be activated all the time even with use of the `<video>` element?
I can also see that the control bar icons toggle the pause/play function with use of a mobile device (Android tested right now) which is extremely enjoying i.e tap on the "full screen" or "chromecast" icon pause the audio where the user do need to tap again on the same icon to activate the icon function. See below print screen.
[...]

PR #7629, opened 2022-02-01
feat: Add audioPosterMode option

harisha-swaminathan:
## Description
Adds audioPosterMode option that is false by default. Turning the audioPosterMode on will enable the Poster Viewer experience by hiding the video element and persistently displaying the poster image. If no poster is defined, the video player will be black.

### DIFF
--- a/src/css/components/_poster.scss
+++ b/src/css/components/_poster.scss
@@ -16,22 +16,14 @@
   height: 100%;
 }
 
-// Hide the poster after the video has started playing
-.vjs-has-started .vjs-poster {
-  display: none;
-}
-
-// Don't hide the poster if we're playing audio
-.vjs-audio.vjs-has-started .vjs-poster {
-  display: block;
-}
-
-// Hide the poster when native controls are used otherwise it covers them
+// Hide the poster after the video has started playing and when native controls are used
+.vjs-has-started .vjs-poster,
 .vjs-using-native-controls .vjs-poster {
   display: none;
 }
 
-// Don't hide the poster when audio-poster-mode is true
-.vjs-has-started.vjs-audio-poster-mode .vjs-poster {
+// Don't hide the poster if we're playing audio or when audio-poster-mode is true
+.vjs-audio.vjs-has-started .vjs-poster, 
+.vjs-has-started.vjs-audio-poster-mode  .vjs-poster {
   display: block;
 }
--- a/src/css/components/_poster.scss
+++ b/src/css/components/_poster.scss
@@ -30,3 +30,8 @@
 .vjs-using-native-controls .vjs-poster {
   display: none;
 }
+
+// Don't hide the poster when audio-poster-mode is true
+.vjs-has-started.vjs-audio-poster-mode .vjs-poster {
+  display: block;
+}
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -349,8 +349,6 @@ class Player extends Component {
       }
     }
 
-    options.audioPosterMode = false;
-
     // Run base component initializing with new options
     super(null, options, ready);
 
@@ -4303,10 +4301,13 @@ class Player extends Component {
    */
   audioPosterMode(value) {
 
-    if (value === undefined) {
+    value = value && Boolean(value);
+
+    if (value === undefined || value === this.options_.audioPosterMode) {
       return this.options_.audioPosterMode;
     }
-    this.options_.audioPosterMode = !!value && !this.options_.audioOnlyMode;
+
+    this.options_.audioPosterMode = value;
 
     if (this.options_.audioPosterMode) {
       this.tech_.hide();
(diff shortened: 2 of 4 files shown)


src/js/player.js
```diff
@@ -349,6 +349,8 @@ class Player extends Component {
       }
     }
 
+    options.audioPosterMode = false;
```
misteroneill, line 352: suggestion: It seems like this would be better set at the bottom of this file in `Player.prototype.options_`.

src/js/player.js
```diff
@@ -4290,6 +4292,32 @@ class Player extends Component {
...
+   *
+   * @return {boolean}
+   *         True if audioPosterMode is on, false otherwise.
+   */
+  audioPosterMode(value) {
+
+    if (value === undefined) {
+      return this.options_.audioPosterMode;
+    }
+    this.options_.audioPosterMode = !!value && !this.options_.audioOnlyMode;
```
misteroneill, line 4309: I'm having trouble following the logic here. Do you think this would be clearer?
```suggestion
    value = Boolean(value);

    // If no value is provided or it is unchanged, acts as a getter.
    if (value === undefined || value === this.options_.audioOnlyMode) {
      return this.options_.audioOnlyMode;
    }
```

test/unit/player.test.js
```diff
@@ -1570,6 +1570,21 @@ QUnit.test('should add an audio player region if an audio el is used', function(
   player.dispose();
 });
 
+QUnit.test('should turn on/off audioPosterMode when audioPosterMode is called with true/false', function(assert) {
```
misteroneill, line 1573: suggestion: I think we should separate this into a few distinct tests. Also, in general, I would suggest verifying things by calling `audioPosterMode()` as a getter rather than checking `options_` as we want to test the public API.

- A test that asserts the option defaults to `false`, but can be set to `true` at player creation. 
- A test that asserts calling `audioPosterMode()` as a setter changes the values returned by calling `audioPosterMode()` as a getter.
- A test that asserts that the `vjs-audio-poster-mode` class and tech visibility is changed appropriately when `audioPosterMode()` is called as a setter.

You could do all this in a single test with many assertions, but separating it into multiple tests with a few assertions each is arguably more readable and maintainable. 👍

src/css/components/_poster.scss
```diff
@@ -30,3 +30,8 @@
 .vjs-using-native-controls .vjs-poster {
   display: none;
 }
+
+// Don't hide the poster when audio-poster-mode is true
+.vjs-has-started.vjs-audio-poster-mode .vjs-poster {
+  display: block;
+}
```
misteroneill, line 37: thought: Could this be combined with the above `.vjs-audio.vjs-has-started .vjs-poster` block? Seems like it'd be cleaner that way. Could probably also combine the `.vjs-has-started .vjs-poster` and `.vjs-using-native-controls .vjs-poster` blocks, too.

misteroneill (changes requested):
Looks great, overall! Some thoughts and suggestions follow...

mister-ben:
Either as part of this PR or as a followup, we should consider implementing `userAction.click` and `doubleClick` on the poster instead of the tech in audio poster mode - #7577

misteroneill:
@mister-ben Yeah, we should address that, but let's do it in a separate PR.

mister-ben:
We should have some documentation in docs/guides/options.md until docs are moved elsewhere, otherwise looks good.

src/js/player.js
```diff
@@ -4290,6 +4290,35 @@ class Player extends Component {
...
+    value = value && Boolean(value);
+
+    if (value === undefined || value === this.options_.audioPosterMode) {
+      return this.options_.audioPosterMode;
+    }
+
+    this.options_.audioPosterMode = value;
+
+    if (this.options_.audioPosterMode) {
+      this.tech_.hide();
```
gkatsev, line 4313: thought: hiding the tech explicitly could potentially cause playback issues, particularly with other techs like youtube. Though, I'm not positive. Probably not worth handling now, but might need another solution if we do run into such a problem in the future.

src/js/player.js
```diff
@@ -4290,6 +4290,35 @@ class Player extends Component {
...
+   * @return {boolean}
+   *         True if audioPosterMode is on, false otherwise.
+   */
+  audioPosterMode(value) {
+
+    value = value && Boolean(value);
+
+    if (value === undefined || value === this.options_.audioPosterMode) {
+      return this.options_.audioPosterMode;
+    }
```
gkatsev, line 4308: if `value` is falsey, this will maintain value as that value, for example `"" || Boolean("")` will return `""` rather than `false`.

Might be better to do a type check below. Something like

```suggestion
    value = value && Boolean(value);

    if (typeof value === 'undefined' || typeof value !== "boolean" || value === this.options_.audioPosterMode) {
      return this.options_.audioPosterMode;
    }

```

PR #7633, opened 2022-02-03
feat: Greater text track precision

### LINKED ISSUE #7026: Develop timed data processing pipeline around meta data text track
## Description
I have some data (around 10,000 records and growing) against a video playing in video js which is timed (having start and end time). I need to process that data when video is playing and need to draw bounding boxes against faces appearing in video player. My data contains record for every frame of video, and a record exists at 30ms interval.

Earlier i tried to utilize timeUpdate event of player, and on every timeUpdate event i was filtering the whole data set. But as precision is very important in my use case, and timeUpdate event has latency of around 250ms (it fires after every 250ms) so i had to drop this implementation.

Later on, i created a timer which fires player time after every 30 milliseconds which served my purpose. But as my data is increasing, i am looking for more optimized way of implementing this.

One way which i thought of was to create a text track of type metadata and add all my data as VTTCue inside that text track and let player perform all filter logic for me. I thought that this will run outside of my main JavaScript thread (kindly share any resource if you have for me to understand it better) and will be more accurate. I posted a [question](https://stackoverflow.com/questions/60917608/how-html5-video-process-text-tracks) about VTT processing a while ago on stack overflow but never got any response. I ran into problem with this approach too as cueChange event of text track is fired after every 100-400 ms.

## Steps to reproduce
I have created a small [demo](https://stackblitz.com/edit/js-gxfnym?file=index.html) to observe behavior of cueChange event. I have added a VTTCue with interval of 50ms but player is firing cueChange event after around 100-400 ms.

## Results
i was hoping that metadata text track will provide me required precision.

## Expectations

1. i am hoping to understand and read more about internals of text tracks processing and how this all actually work internally.
2. Can anyone else share their experience like how did they handled this sort of data or may be share their use case.

mister-ben:
## Description
Text tracks currently update on `timeupdate`, which can can be too slow and/or unpredictable for some (edge) cases. This PR switches to using [`requestVideoFrameCallback()`](https://web.dev/requestvideoframecallback-rvfc/) instead, with a fallback to `requestAnimationFrame()`. There was some previous work along these lines with rAF in #2985.

Closes #2985, closes #7026.

## Specific Changes proposed
* Implements `Tech.requestVideoFrameCallback()` as a fallback to `requestAnimationFrame()` whilst playing.
* Implements `Html5.requestVideoFrameCallback()` if the HTML5 element supports it.
* Switches text track to use this method instead of the `timeupdate` event.

## Requirements Checklist
- x ] Feature implemented / Bug fixed
  - [ ] Docs/guides updated
  - [x] Example created https://codepen.io/mister-ben/pen/jOarxro?editors=1011

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -739,6 +739,32 @@ class Html5 extends Tech {
   }
 
   /**
+   * Native requestVideoFrameCallback if supported by browser/tech, or fallback
+   *
+   * @param {function} cb function to call
+   * @return {number} id of request
+   */
+  requestVideoFrameCallback(cb) {
+    if (this.featuresVideoFrameCallback) {
+      return this.el_.requestVideoFrameCallback(cb);
+    }
+    return super.requestVideoFrameCallback(cb);
+  }
+
+  /**
+   * Native or fallback requestVideoFrameCallback
+   *
+   * @param {number} id request id to cancel
+   */
+  cancelVideoFrameCallback(id) {
+    if (this.featuresVideoFrameCallback) {
+      this.el_.cancelVideoFrameCallback(id);
+    } else {
+      super.cancelVideoFrameCallback(id);
+    }
+  }
+
+  /**
    * A getter/setter for the `Html5` Tech's source object.
    * > Note: Please use {@link Html5#setSource}
    *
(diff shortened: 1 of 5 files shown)


gkatsev:
With this change, it might be worth updating https://github.com/videojs/video.js/blob/main/src/js/tracks/text-track.js#L313 to ignore equal end-times, as that's what the spec says. Though, I think there might be some implications in VHS related to that, I think.

mister-ben:
What's the concern with VHS? Changing to ignore equal end time doesn't seem to cause an issue with the big bop test stream.

gkatsev:
I think potentially with the segment metadata track? If you don't see any issues, then it's probably not an issue. Unfortunately, I don't remember the specifics anymore.

mister-ben:
Segment metadata seems to not be any different. I'm not getting any gaps with no active cue if that was the concern.

gkatsev:
@mister-ben cool! I can't remember or find anything concrete, so, I say this is good to go and if we hit that potential issue we can fix it forward at that point.

asadmalik3:
This was a much-needed feature. Thanks @mister-ben

PR #7647, opened 2022-02-16
feat: Audio Only Mode

alex-barstow:
## Description
This adds support for `audioOnlyMode`, which will hide all player children except the control bar, and within the control bar any controls that are only needed for video.

This can be enabled vis either the new `player.audioOnlyMode()` getter/setter or a `audioOnlyMode` option, which defaults to `false`

## Specific Changes proposed
When `audioOnlyMode` is enabled:
- Hide the tech and other player children, except the control bar
- Lock the control bar in a visible state so it does not hide when the user is inactive
- Hide the `CaptionsButton`, `DescriptionButton`, `SubsCapsButton`, `FullscreenToggle`, `PictureInPictureToggle` (this is done with CSS rather than JS to avoid unneeded complexity since some of these controls show/hide themselves dynamically depending on available support and/or the presence of certain text tracks.
- Leave certain controls potentially useful for audio visible (ex. `PlaybackRateMenuButton`, `ChaptersButton`)
- Set the player height equal to the control bar height
- Trigger a `'audioonlymodechange'` event


src/js/player.js
```diff
@@ -4290,6 +4297,88 @@ class Player extends Component {
...
+    }
+
+    if (value === this.audioOnlyMode_) {
+      return;
+    }
+
+    if (typeof value !== 'boolean') {
+      log.warn('audioOnlyMode can only be enabled or disabled using a boolean value');
+      return;
+    }
```
misteroneill, line 4324: I think this whole thing could be consolidated down to something similar to what's in #7629. And I'm not sure logging is particularly useful here. Maybe this:

```suggestion
    if (typeof value !== 'boolean' || value === this.audioOnlyMode_) {
      return this.audioOnlyMode_;
    }
```

src/js/player.js
```diff
@@ -4290,6 +4297,88 @@ class Player extends Component {
...
+      this.addClass('vjs-audio-only-mode');
+
+      const playerChildren = this.children();
+      const controlBar = this.getChild('ControlBar');
+      const controlBarHeight = controlBar && controlBar.height();
+
+      // Hide all player components except the control bar. Control bar components
+      // needed only for video are hidden with CSS
+      playerChildren.forEach(child => {
+        if (child.name_ === 'ControlBar') {
```
misteroneill, line 4337: This might be cleaner checking the objects by reference:

```suggestion
        if (child === controlBar) {
```

src/js/player.js
```diff
@@ -395,6 +395,12 @@ class Player extends Component {
     // Init debugEnabled_
     this.debugEnabled_ = false;
 
+    // Init state audioOnlyMode_
+    this.audioOnlyMode_ = false;
+
+    // Init state hiddenPlayerChildren_
+    this.hiddenPlayerChildren_ = [];
```
misteroneill, line 402: Since this property specifically relates to audio-only mode, I'd suggest changing the name to something like `audioOnlyHiddenChildren_`.

However, given my comment about `height` below, maybe an `audioOnlyCache_` object that can store arbitrary values for reverting out of audio-only mode?

src/js/player.js
```diff
@@ -4290,6 +4297,88 @@ class Player extends Component {
...
+      // Show player components that were previously hidden
+      if (this.hiddenPlayerChildren_.length > 0) {
+        this.hiddenPlayerChildren_.forEach(child => child.show());
+      }
+
+      // Reset player height
+      this.height(this.options_.height);
+
+      // Show control bar
+      this.userActive(true);
```
misteroneill, line 4374: Shouldn't this be `false`?

src/js/player.js
```diff
@@ -4290,6 +4297,88 @@ class Player extends Component {
...
+    } else {
+      this.removeClass('vjs-audio-only-mode');
+
+      // Show player components that were previously hidden
+      if (this.hiddenPlayerChildren_.length > 0) {
+        this.hiddenPlayerChildren_.forEach(child => child.show());
+      }
+
+      // Reset player height
+      this.height(this.options_.height);
```
misteroneill, line 4371: Are there cases where `this.options_.height` is not set? What will happen in those cases?

Might make sense to cache the value.

src/js/player.js
```diff
@@ -4290,6 +4297,88 @@ class Player extends Component {
...
+
+      // Show control bar
+      this.userActive(true);
+
+      this.audioOnlyMode_ = false;
+
+      this.trigger('audioonlymodechange');
+    }
```
misteroneill, line 4379: A number of things in this block are repeated and they could be consolidated after the block (or before it, depending).

For example, `this.trigger('audioonlymodechange')` is in two places. Or `this.audioOnlyMode_ = [true|false];` could be replaced with `this.audioOnlyMode_ = value;`. There are other opportunities for consolidation, too.

src/js/player.js
```diff
@@ -4290,6 +4297,88 @@ class Player extends Component {
...
+      }
+
+      this.audioOnlyMode_ = true;
+
+      this.trigger('audioonlymodechange');
+    } else {
+      this.removeClass('vjs-audio-only-mode');
+
+      // Show player components that were previously hidden
+      if (this.hiddenPlayerChildren_.length > 0) {
```
misteroneill, line 4366: I think we can safely assume this is an array, so not bother testing `length`. An empty array will simply have no effect.

misteroneill (commented):
Miscellaneous thoughts!

src/js/player.js
alex-barstow: I just removed these `userActive()` calls entirely. The intention was to make sure controls were shown following transitions in and out of audioOnlyMode, but I don't think it's necessary. Now the behavior is:

Enter audioOnlyMode - controls fade in and stay locked
Exit audioOnlyMode - controls fade out and start responding to user activity again

misteroneill (dismissed):
@alex-barstow I've merged #7629 - let's update this with `main` and unify `audioPosterMode` and `audioOnlyMode` in this PR:

* Generally, ensure that both options have a similar implementation/API
* When `audioOnlyMode` is set to `true`, set `audioPosterMode` to `false`
* When `audioPosterMode` is set to `true`, set `audioOnlyMode` to `false`

I believe we were planning on doing this in a third PR, but let's just do that here.

src/js/player.js
```diff
@@ -4290,6 +4300,69 @@ class Player extends Component {
...
+      return this.audioOnlyMode_;
+    }
+
+    if (value === true) {
+      // Update styling immediately so we can get the control bar's height
+      this.addClass('vjs-audio-only-mode');
+
+      const playerChildren = this.children();
+      const controlBar = this.getChild('ControlBar');
+      const controlBarHeight = controlBar && controlBar.height();
```
gkatsev, line 4326: non-blocking: We potentially want `currentHeight()` here which should give us the computed height and not the height that it was set to, which may not be the true height.
```suggestion
      const controlBarHeight = controlBar && controlBar.currentHeight();
```

In practice and based on how things are set up, this may not really matter here.

src/js/player.js
```diff
@@ -4290,6 +4300,69 @@ class Player extends Component {
...
+
+      this.audioOnlyCache_.playerHeight = this.height();
+
+      // Set the player height the same as the control bar
+      this.height(controlBarHeight);
+
+      // Fullscreen is not supported in audioOnlyMode, so exit if we need to
+      if (this.isFullscreen()) {
+        this.exitFullscreen();
+      }
```
gkatsev, line 4350: should we also exit PiP?

gkatsev: should we exit fs/pip before getting the control bar height, in case it's different in fullscreen? At least exitFullscreen should have a promise for knowing when exiting has finished.

gkatsev (commented):
A couple of tiny comments

gkatsev:
also, if you set `font-size: 3em` on the player div and the control bar is huge, it doesn't like this mode much. For example, hovering on the volume panel to open it up makes the whole player grow.
Aside from that, works great and the `inline: false` volume panel works fine too.

test/unit/player.test.js
```diff
@@ -2161,6 +2162,7 @@ QUnit.test('When VIDEOJS_NO_DYNAMIC_STYLE is set, apply sizing directly to the t
   assert.equal(player.tech_.el().width, 600, 'the width is equal to 600');
   assert.equal(player.tech_.el().height, 300, 'the height is equal 300');
   player.dispose();
+  window.VIDEOJS_NO_DYNAMIC_STYLE = originalVjsNoDynamicStyling;
```
alex-barstow, line 2165: This state was leaking, causing the below `'audioOnlyMode(true/false) changes player height'` test to fail.

misteroneill (approved):

thijstriemstra:
Really cool feature @alex-barstow , but I was surprised to see it in a >= 7.18.1 release, I would've preferred to see it in a major release.. 

Fortunately it didn't break anything in my 'audio only plugin' for video.js which I've been maintaining since video.js 4.x: https://github.com/collab-project/videojs-wavesurfer

Now that video.js finally supports an audio-only mode, I guess I'll have to revisit my plugin at some point and remove all the nasty hacks in order to support wavesurfer.js/audio-only.

gkatsev:
Video.js 8 is coming. Since this is a new feature, with theoretically no breakages, it should come in a minor release (it was released in 7.19.0).

Would love to see an updated version of your plugin that interfaces with the audio-only mode.

PR #7654, opened 2022-02-22
feat: Add big pause button

ulwlu:
## Description
I add the big pause button along with big play button which is already implemented in video.js.

Big pause button is commonly used in any video platform, such as youtube. My project also wants this feature so I implemented it. Any advice is appreciated.

## Specific Changes proposed

- Add Big Pause Button component in Player components.
- Refactor Big Play Button css along with Big Pause Button css.

## Requirements Checklist
  - [ ] Docs/guides updated : if this proposal is agreed, I will add a doc.
  - [x] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0)) : `sandbox/index.html` can show the example.


misteroneill:
Thank you for the pull request!

Unfortunately, we've discussed and don't think this feature is right for the Video.js core. The player already toggles playback based on a click/tap on the stage. Additionally, this would be a big change for existing users of Video.js.

We suggest implementing this as a plugin and, instead of adding a "big pause button," we suggest turning the big play button into a play toggle that will play the video if paused, or pause the video if playing.

Thanks again for taking the time to open a PR!

ulwlu:
@misteroneill 
thanks for your reviewing.

i have a question. 
iOS can't tap to stop the video currently because of its limitation, can it?
our team is thinking that this decreases user experiences.

with this feature, iOS can stop the video just like big play button.

how about adding this not by default but still optional to turn on?

PR #7673, opened 2022-03-11
fix: async audio only tests

alex-barstow:
## Description
Quick fix for failing tests on #main.


misteroneill (approved):

gkatsev:
@alex-barstow fyi, you can push to upstream now, which will run your PRs on browserstack 😁

PR #7692, opened 2022-03-24
fix: tight controlbar coupling (#7689)

### LINKED ISSUE #7689: Tight coupling with controlBar component
## Description
We have tight coupling in player.js Class Player method resetControlBarUI_(), when using reset() method. 
ControlBar component may be not exists or exists on another nesting level..
image

## Steps to reproduce
Use player.reset() method, when in player object there is no controlBar child/component at property "children"  

### Expected
Reset player instance 

### Actual
We have error

### Error output
image

#### videojs
7.17.1
#### browsers
all
#### OSes
all

(Sorry for my bad english)

try2beth3b3st:
## Description
We have tight coupling in player.js Class Player method resetControlBarUI_(), when using reset() method.
ControlBar component may be not exists or exists on another nesting level.
Fixes #7689

## Specific Changes proposed
Added condition

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -2073,6 +2073,7 @@ QUnit.test('player#reset progress bar', function(assert) {
   const player = TestHelpers.makePlayer();
 
   player.removeChild('controlBar');
+  player.controlBar = null;
 
   try {
     player.resetProgressBar_();
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -2066,6 +2066,23 @@ QUnit.test('player#reset removes remote text tracks', function(assert) {
   log.warn.restore();
 });
 
+QUnit.test('player#reset progress bar', function(assert) {
+
+  let error;
+
+  const player = TestHelpers.makePlayer();
+
+  player.removeChild('controlBar');
+
+  try {
+    player.resetProgressBar_();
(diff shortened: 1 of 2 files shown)


try2beth3b3st:
> Nice, these missing null checks are too common! Would you mind adding a test for this case?

Done..

test/unit/player.test.js
```diff
@@ -2066,6 +2066,23 @@ QUnit.test('player#reset removes remote text tracks', function(assert) {
   log.warn.restore();
 });
 
+QUnit.test('player#reset progress bar', function(assert) {
+
+  let error;
+
+  const player = TestHelpers.makePlayer();
+
+  player.removeChild('controlBar');
```
misteroneill, line 2075: This is a reasonable test, but the `player.controlBar` property appears to still exist even when `removeChild` is called.

```suggestion
  player.removeChild('controlBar');
  player.controlBar = null;
```

misteroneill (changes requested):
One minor quibble, otherwise looks good!

gkatsev (approved):

try2beth3b3st:
@misteroneill Excuse me, when to expect patch with this fix?)

PR #7706, opened 2022-04-05
docs: Redirect guides to videojs.com

misteroneill:
## Description

With the recent changes to move guides from this repo to [the main website repo](https://github.com/videojs/videojs.com), guides are no longer maintained in this repository. However, I've left them here so that the docs site still shows something there and existing links don't break.

Related: https://github.com/videojs/videojs.com/pull/134


roman-bc-dev (approved):

mister-ben:
Changes are good for the repo itself, I'm sure I've shared links to docs here rather than the docs site, rightly or wrongly. On docs.videojs.com we'd need to have Netlify send 301 redirects instead for Google etc to index the pages properly.

misteroneill:
Nice, I didn't even think to look into doing actual redirects. 👍

build/fix-api-docs.js
```diff
@@ -41,3 +42,9 @@ replacements.forEach(function(obj) {
     silent: true
   });
 });
+
+// move docs/_redirects into the root of the docs site
+//
+// this is needed because the root of the docs site is docs/api, which is not
+// in version control.
+sh.mv(path.join(apiPath, '..', '_redirects'), path.join(apiPath, '_redirects'));
```
gkatsev, line 50: this should move to `build/netlify-docs.js` right?

misteroneill:
It seems like Netlify's "splats" thing doesn't work in the case where they're inside a filename - e.g. `/tutorials-*.html` - so we need to fall back to using full paths. It's not the worst thing in the world as there are less than 30 of them.

Also, redirects won't work on Netlify previews.

docs/_redirects
```diff
@@ -0,0 +1,29 @@
+# Netlify redirects to redirect old tutorial pages to new guide pages
+# These will not work on netlify preview pages
```
mister-ben, line 2:
```suggestion
```

docs/_redirects
```diff
@@ -0,0 +1,29 @@
...
+/tutorial-modal-dialog.html       https://videojs.com/guides/modal-dialog       301!
+/tutorial-options.html            https://videojs.com/guides/options            301!
+/tutorial-player-workflows.html   https://videojs.com/guides/player-workflows   301!
+/tutorial-plugins.html            https://videojs.com/guides/plugins            301!
+/tutorial-react.html              https://videojs.com/guides/react              301!
+/tutorial-setup.html              https://videojs.com/guides/setup              301!
+/tutorial-skins.html              https://videojs.com/guides/skins              301!
+/tutorial-tech.html               https://videojs.com/guides/tech               301!
+/tutorial-text-tracks.html        https://videojs.com/guides/text-tracks        301!
+/tutorial-tracks.html             https://videojs.com/guides/tracks             301!
```
mister-ben, line 23: There's no https://videojs.com/guides/text-tracks. https://videojs.com/guides/text-tracks instead?

gkatsev: Probably either remove or port over tutorials-tracks as well (it just links out to video, audio, and text track pages)

misteroneill: Oh! That's one that didn't get carried over. We can probably leave it in place because that guide still has content in this repo.

mister-ben (approved):

gkatsev (approved):

PR #7707, opened 2022-04-05
feat: remove use of firstplay event

hugorogz:
## Description
Removed the use of existing 'firstplay' events as it's deprecated in: https://github.com/videojs/video.js/pull/4353

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated


mister-ben (changes requested):
We wouldn't need to trigger `play` instead of `firstplay` -- `play` is already fired so we'll end up with duplicate events.

src/js/player.js
```diff
@@ -1883,7 +1880,6 @@ class Player extends Component {
      * @deprecated As of 6.0 firstplay event is deprecated.
```
mister-ben, line 1880: This and other comments relating to `firstplay` can be removed now too.

docs/legacy-docs/api/js/doc-data-full.js
```diff
@@ -6648,24 +6648,6 @@ var docData =
         "longname": "event:seeked",
         "scope": "global"
     },
-    {
```
gkatsev, line 6651: this piece should stay as it relates to legacy docs.

gkatsev (changes requested):

misteroneill (approved):
Looking good to me!

mister-ben (approved):

amtins:
Just for my understanding, I see that this PR has the **major** label. Does this mean that the `firstplay` event will be removed in version 8?

mister-ben:
That's right

PR #7717, opened 2022-04-08
feat: export more helpers in videojs object

hugorogz:
## Description
- Export all of the Video.js helpers (from utils) on the Video.js object. 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -23,20 +23,15 @@ import * as Fn from './utils/fn.js';
 import TextTrack from './tracks/text-track.js';
 import AudioTrack from './tracks/audio-track.js';
 import VideoTrack from './tracks/video-track.js';
-import * as TimeRanges from './utils/time-ranges.js';
+import { createTimeRanges } from './utils/time-ranges.js';
 import formatTime, { setFormatTime, resetFormatTime } from './utils/format-time.js';
 import log, { createLogger } from './utils/log.js';
 import * as Dom from './utils/dom.js';
 import * as browser from './utils/browser.js';
 import * as Url from './utils/url.js';
 import * as Obj from './utils/obj';
-import * as Buffer from './utils/buffer';
 import clamp from './utils/clamp';
-import filterSource from './utils/filter-source';
-import map from './utils/map';
-import * as Promise from './utils/promise';
-import * as MimeTypes from './utils/mimetypes';
-import set from './utils/set';
+import { isPromise, silencePromise } from './utils/promise';
 import * as StringCases from './utils/string-cases';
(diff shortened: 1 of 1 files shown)


src/js/video.js
```diff
@@ -500,5 +509,36 @@ videojs.defineLazyProperty = defineLazyProperty;
...
 videojs.addLanguage('en', {'Non-Fullscreen': 'Exit Fullscreen'});
 
+const helpers = {
+  browser,
+  buffer: Buffer,
+  clamp,
+  computedStyle,
+  createLogger,
+  defineLazyProperty,
+  domData: DomData,
```
gkatsev, line 519: this property is specifically not exported because it's contains internal proprties

gkatsev (changes requested):
I don't think that the same thing should be exported in multiple locations, it increases the API surface unnecessary.

I would also argue that some have no real utility being exported like guid.

Each item should be considered whether it should be exported, rather than exporting it because it exists.

misteroneill (changes requested):
I agree, I think we need to be more deliberate with this one. But in fairness to @hugorogz, the requirements were... limited...

> Export all of the Video.js helpers (from utils) on the Video.js object. Video.js is a library and it has a lot of useful and well tested functions that we often need in other projects. I think this currently means that we duplicate the code in places and we should instead just export all utils.

misteroneill:
After talking with @hugorogz, I went ahead and updated the PR a bit since the requirements were unclear and that's not on him to resolve! Instead of writing detailed requirements, this was the more expedient path. Thanks, Hugo!

brandonocasey (commented):
I think we might also want to add `getAbsoluteURL` and `getFileExtension` from `src/js/utils/url.js` as they are not exported right now. I also wonder if we should add all of these under a `videojs.utils` or `videojs.helpers` and deprecate any helpers at the top level. imo having these on the top level object could be a bit weird for users and makes it hard to tell util functions from actual API functions.

gkatsev:
looks like all url stuff is being exported via videojs.url.

My main worry with exposing all these utils is that it expands the API surface and now if we wanted to change/remove a helper because Video.js itself no longer uses it. That said, I don't have any major objections to this.

misteroneill:
@brandonocasey @gkatsev What do you think about merging this as-is for now and then having a follow-up discussion about what the final exposed API surface is for utils/helpers?

brandonocasey:
Yeah I think this pull request is good to go for now and we can talk about future API stuff another time

brandonocasey (approved):

gkatsev (approved):

PR #7721, opened 2022-04-09
feat: Merge `data-setup` options on player init

mister-ben:
## Description
Plenty of people get confused where they've included `data-setup` options but use `videojs()` to init the player, not realising that `data-setup` options only apply to auto setup.

This merges any `data-setup` options on the el when a player is initialised.

Also makes the warning a bit clearer for the other case, where `videojs()` is used with options on an already initialised player.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -136,7 +136,7 @@ function videojs(id, options, ready) {
 
   if (player) {
     if (options) {
-      log.warn(`Player "${id}" is already initialised. Options will not be applied.`);
+      log.warn(`Player "${id}" is already initialised. New options were not applied.`);
     }
     if (ready) {
       player.ready(ready);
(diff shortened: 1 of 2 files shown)


src/js/video.js
```diff
@@ -162,6 +162,10 @@ function videojs(id, options, ready) {
 
   options = options || {};
 
+  if (el.hasAttribute('data-setup')) {
+    options = videojs.mergeOptions(JSON.parse(el.getAttribute('data-setup')), options);
```
gkatsev, line 166: should we log a warning saying that this isn't an ideal scenario?

mister-ben:
This is redundant, there _is_ already a merge in player.js although no tests for it. Need to verify apparent inconsistencies, and add tests separately.

PR #7722, opened 2022-04-11
feat: Player can be replaced with original el after dispose()

mister-ben:
## Description
Adds an option to replace player with a clone of the original element on dispose, which has been asked for several times.

## Specific Changes proposed
* Component accepts an options object. If provided, the component element will be replaced by the element provided at `options.restoreEl`.
* Player passes a `restoreEl` to `super.dispose()` if provided as a player option
* The `videojs()` constructor accepts a `restoreEl` option, which can be any element, or `true` in which case the element the player is initialised on is used.

Docs will need to be added at videojs.com/guides

## Requirements Checklist
- [*] Feature implemented / Bug fixed
  - [*] Change has been verified in an actual browser (Chrome, Firefox, IE)
  - [*] Unit Tests updated or fixed
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -162,6 +162,11 @@ function videojs(id, options, ready) {
 
   options = options || {};
 
+  // Store a copy of the el before modification, if it is to be restored in destroy()
+  if (options.restoreEl === true) {
+    options.restoreEl = el.cloneNode(true);
+  }
+
   hooks('beforesetup').forEach((hookFunction) => {
     const opts = hookFunction(el, mergeOptions(options));
 
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -163,8 +163,9 @@ function videojs(id, options, ready) {
   options = options || {};
 
   // Store a copy of the el before modification, if it is to be restored in destroy()
+  // If div ingest, store the parent div
   if (options.restoreEl === true) {
-    options.restoreEl = el.cloneNode(true);
+    options.restoreEl = (el.parentNode && el.parentNode.hasAttribute('data-vjs-player') ? el.parentNode : el).cloneNode(true);
   }
 
   hooks('beforesetup').forEach((hookFunction) => {
--- a/test/unit/video.test.js
+++ b/test/unit/video.test.js
@@ -579,3 +579,45 @@ QUnit.test('adds video-js class name with the video-js embed', function(assert) 
   assert.ok(player.hasClass('video-js'), 'video-js class was added to the first embed');
   assert.ok(player2.hasClass('video-js'), 'video-js class was preserved to the second embed');
 });
+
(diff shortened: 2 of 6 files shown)


src/js/video.js
```diff
@@ -162,6 +162,11 @@ function videojs(id, options, ready) {
 
   options = options || {};
 
+  // Store a copy of the el before modification, if it is to be restored in destroy()
+  if (options.restoreEl === true) {
+    options.restoreEl = el.cloneNode(true);
```
gkatsev, line 167: Not a block: Clever to re-use the option to store the clone element, though, that might be a bit unexpected.

gkatsev, line 167: to get this line covered by tests, we probably want to add a test to https://github.com/videojs/video.js/blob/main/test/unit/video.test.js

Ideally, we'd have a test per embed type https://videojs.com/guides/embeds/

gkatsev (approved):
Confirmed with:
```js
var d= document.createElement('div')
d.innerHTML = '<span style="color: red; font-size: 30px;">RESTORED</span>'
player.options_.restoreEl = d
player.dispose()
```

If you're able to add the other tests, that would be awesome, but I think it's probably ok as is.

src/js/video.js
mister-ben: I'll add some more tests. And sine this can be a minor I'll rebase onto `main`. Before starting I was thinking this would be more impactful a change than it is.

test/unit/video.test.js
```diff
@@ -579,3 +579,45 @@ QUnit.test('adds video-js class name with the video-js embed', function(assert)
...
+    {
+      type: 'div ingest',
+      src: '<div data-vjs-player><video id="test3"><source src="http://example.com/video.mp4" type="video/mp4"></source></video></div>',
+      initSelector: 'test3',
+      testSelector: 'div[data-vjs-player]'
+    }
+  ];
+
+  embeds.forEach(embed => {
+    const comparisonEl = document.createRange().createContextualFragment(embed.src).children[0];
```
gkatsev, line 608: huh, never seen `createContextualFragment` before

test/unit/video.test.js
```diff
@@ -579,3 +579,45 @@ QUnit.test('adds video-js class name with the video-js embed', function(assert)
...
+
+    fixture.innerHTML += embed.src;
+
+    const player = videojs(embed.initSelector, {restoreEl: true});
+
+    assert.ok(comparisonEl.isEqualNode(player.options_.restoreEl), `${embed.type}: restoreEl option replaced by element`);
```
gkatsev, line 614: non-blocking: should we check that `player.el()` isn't equal to `comparisonEl`?

gkatsev (approved):

gkatsev (approved):

PR #7724, opened 2022-04-13
fix: Audio only mode styling conflicts with fluid mode

alex-barstow:
## Description
Disable all fluid and fill mode CSS when in audio only mode, except for that which adjusts the player's width responsively. This fixes a styling conflict that broke the player UI when both fluid and audio only mode were simultaneously enabled.

### DIFF
--- a/src/css/components/_layout.scss
+++ b/src/css/components/_layout.scss
@@ -70,13 +70,20 @@
 
 // Not including a default AR in vjs-fluid because it would override
 // the user set AR injected into the header.
+.video-js.vjs-fluid,
+.video-js.vjs-16-9,
+.video-js.vjs-4-3,
+.video-js.vjs-9-16,
+.video-js.vjs-1-1 {
+  width: 100%;
+  max-width: 100%;
+}
+
 .video-js.vjs-fluid:not(.vjs-audio-only-mode),
 .video-js.vjs-16-9:not(.vjs-audio-only-mode),
 .video-js.vjs-4-3:not(.vjs-audio-only-mode),
 .video-js.vjs-9-16:not(.vjs-audio-only-mode),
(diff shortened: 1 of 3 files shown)


misteroneill (approved):

src/css/components/_layout.scss
```diff
@@ -77,26 +77,33 @@
...
   @include apply-aspect-ratio(9, 16);
 }
 
-.video-js.vjs-1-1 {
+.video-js.vjs-1-1:not(.vjs-audio-only-mode) {
   @include apply-aspect-ratio(1, 1);
 }
 
-.video-js.vjs-fill {
+.video-js.vjs-fill:not(.vjs-audio-only-mode) {
```
roman-bc-dev, line 106: Is there a reason why we wouldn't be able to add a rule with a target like `.video-js.vjs-audio-only-mode` after these rules? That would be of equal hierarchy as the `.video-js.<ratio-class-name>` targets and should override them just because it's further down in the stylesheet. That way you wouldn't need to specify `:not()`.

roman-bc-dev: And same thought for the targeting on line 82. It might just be easier a bit more terse to have a small `.video-js.vjs-audio-only-mode` rule after the ones modified here. That way it's more declarative and less conditional.

alex-barstow: Summary following offline discussion:

This is a good suggestion and it would probably be ideal to do it this way, but overriding the player's `height` via CSS in these places disables the `player.height()` setter method, on which `audioOnlyMode()` currently depends. So I think we should leave this as is for now, and if we decide to rework `audioOnlyMode()` later not to use `player.height()` we can do that, but it's out of scope for this PR.

PR #7735, opened 2022-04-25
chore: Update issue template to a form

mister-ben:
* Replaces the issues template a [form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) to ensure issues are populated with structured information, and more guidance can be given to help full it out.
* Disables the ability to create an issue without the template - text in the template direects users to discussions if it's not an bug report.

We might want to consider some other issue types — would separate types for playback issues and general implementation issues be useful? Do we want a "feature request" template? Is no blank issue too strict?

Live example at https://github.com/mister-ben/issues-test-abc/issues/new/choose


gkatsev:
Should we add fields for expected behavior and actual behavior?

mister-ben:
> Should we add fields for expected behavior and actual behavior?

The thought was that a good description would already include that. But we can add explicit fields too.

misteroneill (commented):
> Should we add fields for expected behavior and actual behavior?

We did discuss this and as Ben says, the thinking was that those fields ought to be optional because most of the time it's pretty obvious. That said, if you have a counter-point, there's no reason we couldn't change that!

gkatsev:
Mostly, that folks need to be pushed into putting in as much detail as possible, so, extra fields would be nice. That said, we can always add those in later on if we felt like this template isn't good enough.

PR #7745, opened 2022-05-04
feat(lang): add Estonian (et) translations

Pikse:
## Description
New translations.


mister-ben (approved):
I don't speak Estonian, but looks good from a quick Google Translate sanity check. Thanks!

misteroneill (approved):

PR #7754, opened 2022-05-10
fix(accessibility): frame must have a title attribute

gjanblaszczyk:
## Description

According to some HTML validators, the resize manager frame doesn't pass validation because the title attribute is not present.
Please check the below validators for details
1. https://rocketvalidator.com/accessibility-validation/frame-title-unique
2. https://equalizedigital.com/accessibility-checker/iframe-missing-title/

## Specific Changes proposed
Add 'title' attribute.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -85,5 +85,6 @@
   "End of dialog window.": "End of dialog window.",
   "{1} is loading.": "{1} is loading.",
   "Exit Picture-in-Picture": "Exit Picture-in-Picture",
-  "Picture-in-Picture": "Picture-in-Picture"
+  "Picture-in-Picture": "Picture-in-Picture",
+  "No content": "No content"
 }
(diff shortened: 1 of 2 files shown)


gkatsev (approved):
Validators are weird. Though, definitely no harm in having the title.

mister-ben:
If it matters for accessibility it should probably be localised...

gkatsev:
Given that the iframe should only be used on IE/older browsers (https://caniuse.com/resizeobserver), I'm not sure it's quite as necessary nowadays

gjanblaszczyk:
@gkatsev Please, feel free to close this PR if you decide it is not needed nowadays.

Btw maybe it is a good time to remove a `& Flash video player` text from an About text. Rest in peace Flash player... I will never forget you :)

image

misteroneill:
Good call! Updated the "About" section to remove the mention of Flash

PR #7788, opened 2022-06-03
feat: add a new title bar component

misteroneill:
## Description
This PR mostly adopts [the Brightcove videojs-dock plugin](https://github.com/brightcove/videojs-dock) as a first-class feature of Video.js.

One key difference is the lack of a "shelf" element as an empty `<div>` without a clear purpose felt a bit unnecessary in the context of Video.js itself (and may not survive in the Brightcove Player either).

### BREAKING CHANGE

This removes the `vjs-big-play-centered` class and moves the Big Play Button to the center of the player by default to make room for the title bar. This is a more common UI convention in 2022.

## Specific Changes proposed
- Add `TitleBar` component, which is a new default child of the player.
- The component is empty (and, therefore, hidden) by default.
- Providing the `Player` with video metadata via `loadMedia` will automatically populate the `TitleBar` with a `title` and `description` (both optional).
  - The `TitleBar` can be independently updated using `player.titleBar.update()`.
    - To unset a previously set value, set it to a falsy value like `''` or `null`.
  - If neither is provided in `loadMedia()` or `titleBar.update()`, the `TitleBar` component is hidden.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [x] Example created (in repo)

### DIFF
--- a/sandbox/load-media.html.example
+++ b/sandbox/load-media.html.example
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>Video.js loadMedia Demo</title>
+  <link href="../dist/video-js.css" rel="stylesheet" type="text/css">
+  <script src="../dist/video.js"></script>
+</head>
+<body>
+  <h1>Video.js <code>loadMedia</code> Demo</h1>
+  <p>This shows how the <code>loadMedia</code> method is used and the effect it has on the player.</p>
+
+  <video-js
+    id="vid1"
+    controls
+    width="640"
+    height="264">
+  </video-js>
+
+  <script>
+    var vid = document.getElementById('vid1');
+    var player = videojs(vid);
+
+    player.log('window.player created', player);
+
+    player.loadMedia({
+      artist: 'Disney',
+      album: 'Oceans',
+      title: 'Oceans',
+      description: 'Journey in to the depths of a wonderland filled with mystery, beauty and power. Oceans is a spectacular story, narrated by Pierce Brosnan, about remarkable creatures under the sea. It\'s an unprecedented look at the lives of these elusive deepwater creatures through their own eyes. Incredible state-of-the-art-underwater filmmaking will take your breath away as you migrate with whales, swim alongside a great white shark and race with dolphins at play.',
+      poster: 'https://vjs.zencdn.net/v/oceans.png',
+      src: [{
+        src: 'https://vjs.zencdn.net/v/oceans.mp4',
+        type: 'video/mp4'
+      }, {
+        src: 'https://vjs.zencdn.net/v/oceans.webm',
+        type: 'video/webm'
+      }, {
+        src: 'https://vjs.zencdn.net/v/oceans.ogv',
+        type: 'video/ogv'
+      }],
+      textTracks: [{
+        kind: 'captions',
+        label: 'English',
+        src: '../docs/examples/shared/example-captions.vtt',
+        srclang: 'en'
+      }]
+    })
+  </script>
+
+</body>
+</html>
--- a/src/css/components/_title-bar.scss
+++ b/src/css/components/_title-bar.scss
@@ -1,0 +1,3 @@
+.vjs-title-bar {
+
+}
--- a/src/css/components/_title-bar.scss
+++ b/src/css/components/_title-bar.scss
@@ -1,3 +1,36 @@
 .vjs-title-bar {
 
+  // At a base inherited font-size of 10px, the title bar overall height should
+  // be 96px with the area of text occupying the first 48px and the rest being
+  // padding. This leaves plenty of room for the gradient to fade to
+  // transparent while maintaining an WCAG AA-compliant contrast ratio (tested
+  // using the TPGi Color Contrast Analyzer application) even on top of a solid
+  // white background.
+  @include linear-gradient(
+    180deg,
+    rgba(0, 0, 0, 0.9) 0%,
+    rgba(0, 0, 0, 0.7) 60%,
+    rgba(0, 0, 0, 0) 100%
+  );
+  font-size: 1.2em; // 12px
+  line-height: 1.5; // 18px
+  @include transition(opacity 0.1s);
+  padding: 0.666em 1.333em 4em; // 8px 16px 48px
+  pointer-events: none;
+  position: absolute;
+  top: 0;
+  width: 100%;
+}
+
+.vjs-title-bar-title,
+.vjs-title-bar-desc {
+  margin: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.vjs-title-bar-title {
+  font-weight: bold;
+  margin-bottom: 0.333em; // 4px
 }
--- a/src/css/components/_title-bar.scss
+++ b/src/css/components/_title-bar.scss
@@ -23,7 +23,7 @@
 }
 
 .vjs-title-bar-title,
-.vjs-title-bar-desc {
+.vjs-title-bar-description {
   margin: 0;
   overflow: hidden;
   text-overflow: ellipsis;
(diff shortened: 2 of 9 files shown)


src/css/components/_title-bar.scss
```diff
@@ -0,0 +1,41 @@
...
+  white-space: nowrap;
+}
+
+.vjs-title-bar-title {
+  font-weight: bold;
+  margin-bottom: 0.333em; // 4px
+}
+
+.vjs-playing.vjs-user-inactive .vjs-title-bar {
+  opacity: 0;
```
mister-ben, line 39: Not specific to just the title bar, but this is a good time to think about adding it, if on fullscreen we set `display: none` on anything we overlay and `background-color: transparent` on the fullscreen element then [Safari's low power mode](https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari) can kick in. `opacity: 0` or `visibility: hidden` aren't invisible enough of course.

```css
.vjs-playing.vjs-fullscreen,
.vjs-playing.vjs-fullscreen .vjs-tech {
  background-color: transparent;
}
.vjs-playing.vjs-fullscreen.vjs-user-inactive .vjs-title-bar
.vjs-playing.vjs-fullscreen.vjs-user-inactive .vjs-control-bar {
  display: none;
}
```

mister-ben, line 39: Not specific to just the title bar, but this is a good time to think about adding it, if on fullscreen we set `display: none` on anything we overlay and `background-color: transparent` on the fullscreen element then [Safari's low power mode](https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari) can kick in. `opacity: 0` or `visibility: hidden` aren't invisible enough of course.

```css
.vjs-playing.vjs-fullscreen,
.vjs-playing.vjs-fullscreen .vjs-tech {
  background-color: transparent;
}
.vjs-playing.vjs-fullscreen.vjs-user-inactive .vjs-title-bar
.vjs-playing.vjs-fullscreen.vjs-user-inactive .vjs-control-bar {
  display: none;
}
```

misteroneill: That's a good thing to discuss. That doc says that the background should be black, not transparent. And it says `Set the CSS property display:none for all non-video elements on your webpage while the video is playing`... does that mean _all_ elements or just those that are part of the fullscreen experience?

Either way, setting `display:none` on all elements except the `video` element seems like a non-starter because we lose all our UI and any other implementations like overlays or ads.

misteroneill: Also, is this something that we can only do in a major version?

src/css/components/_title-bar.scss
```diff
@@ -0,0 +1,41 @@
+.vjs-title-bar {
+
+  // At a base inherited font-size of 10px, the title bar overall height should
+  // be 96px with the area of text occupying the first 48px and the rest being
+  // padding. This leaves plenty of room for the gradient to fade to
+  // transparent while maintaining an WCAG AA-compliant contrast ratio (tested
+  // using the TPGi Color Contrast Analyzer application) even on top of a solid
+  // white background.
+  @include linear-gradient(
```
gkatsev, line 9: Confirmed that worst-case scenario is still above the WCAG AA threshold of 4.5:1 contrast ratio.

src/css/components/_title-bar.scss
mister-ben: [This blog](https://squaredi.blogspot.com/2021/05/) explains it better than Apple's documention, they're right that `background-color: transparent` rather than `black` is necessary for low power mode when teting with the quart debug tool.

`display: none` would only be while the user is inactive. if they tab, then they're not inactive and it would not longer be `display: none`

src/js/player.js
```diff
@@ -4905,6 +4906,10 @@ class Player extends Component {
       textTracks.forEach(tt => this.addRemoteTextTrack(tt, false));
     }
 
+    if (this.titleBar) {
+      this.titleBar.update({title, description});
```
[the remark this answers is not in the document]
mister-ben: maybe we want a configuration of which field(s) to use from `getMedia()`. i've known people to want a third line in the existing dock with the duration, or something like a category

src/css/components/_title-bar.scss
gkatsev: The issue is that with a screen reader, you can still be interacting with a control when the controls visually hide. Since our compromise was hiding the controls even if it has focus but keeping it accessible for Screen Readers.
For regular keyboard control, I think it would've been fine, but we specifically want SRs to maintain access to the controls even if the controls are visually hidden.

gkatsev: heh, yeah. It would be cool to investigate how if we can take advantage of low power mode. Maybe a carve out for fullscreen on safari? Not sure.

src/js/player.js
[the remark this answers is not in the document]
misteroneill: That can probably be an enhancement. I don't want to overload this PR.

src/js/title-bar.js
```diff
@@ -0,0 +1,113 @@
...
+   * @param  {Object} [options={}]
+   *         An options object. When empty, the title bar will be hidden.
+   *
+   * @param  {string} [options.title]
+   *         A title to display in the title bar.
+   *
+   * @param  {string} [options.description]
+   *         A description to display in the title bar.
+   */
+  update({title = '', description = ''} = {}) {
```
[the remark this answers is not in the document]
misteroneill: Updated this to use `setState()` internally, so it behaves in a more predictable/consistent way.

sandbox/load-media.html.example
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
misteroneill: Done

src/css/components/_title-bar.scss
misteroneill: Nice!

misteroneill:
> > ### BREAKING CHANGE
> > This removes the `vjs-big-play-centered` class and moves the Big Play Button to the center of the player by default to make room for the title bar. This is a more common UI convention in 2022.
> 
> Why not have the corner button the default and then toggle the centered button if the title bar is visible? This would make it backwards compatible.
That's sensible, _but_ this is being included as a new feature for 8.0 specifically. Plus a centered button is the more common UX in general, I think.

There's no _technical_ reason why it couldn't be backward compatible.

roman-bc-dev (approved):

src/js/title-bar.js
```diff
@@ -0,0 +1,113 @@
...
+    // according to whether a title and/or description have been provided.
+    if (techEl) {
+      techEl.removeAttribute('aria-labelledby');
+      techEl.removeAttribute('aria-describedby');
+
+      if (title) {
+        techEl.setAttribute('aria-labelledby', titleEl.id);
+      }
+      if (description) {
+        techEl.setAttribute('aria-describedby', descriptionEl.id);
```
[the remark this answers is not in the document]
gkatsev: Ok, let's leave it as is then, and we can update it later on if we find it's better another way or something.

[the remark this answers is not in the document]
gkatsev: I did a tiny bit more testing, still not able to focus using VO directly onto the player element with I add the aria label, but adding the attribute and focusing programmatically does say the label.
I think there are improvements that can be made here, but we probably can come in separately.

gkatsev:
I think it's a bit unexpected that updating the title only won't remove the description too.
```js
player.titleBar.update({title: 'foo'});
```
I'd expect that I would only get a title then, but I can see keeping the description.
We'd want to document that to hide one while displaying the other pass in `''` or `null` (or any falsey value really) to hide that line.

gkatsev:
I guess I mentioned my previous point previously 😅. I guess the new thing is documentation :)

misteroneill:
Yeah, I'll document it.

gkatsev (approved):

PR #7808, opened 2022-06-21
feat: add an option to suppress out of dom warning

joeflateau:
## Description
Adds an option to suppress the "The element supplied is not included in the DOM" warning for intentionally detached video/audio elements. 

Use case: video.js wrapping html5 audio element but not using video.js UI and video.js wrapping html5 video rendered into a WebGL scene. 

## Specific Changes proposed
add a player option that, when set to true, suppresses the warning

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/video.js
+++ b/src/js/video.js
@@ -150,17 +150,17 @@ function videojs(id, options, ready) {
     throw new TypeError('The element or ID supplied is not valid. (videojs)');
   }
 
+  options = options || {};
+
   // document.body.contains(el) will only check if el is contained within that one document.
   // This causes problems for elements in iframes.
   // Instead, use the element's ownerDocument instead of the global document.
   // This will make sure that the element is indeed in the dom of that document.
   // Additionally, check that the document in question has a default view.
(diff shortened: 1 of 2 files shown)


misteroneill (commented):
Thanks for the PR and sorry about the delay in addressing it!

I don't have anything strongly against it, but I wonder if it's needed. Your use case makes sense, but isn't logging a warning the right thing to do?

A warning doesn't mean action needs to be taken, it means there is a condition which may deserve attention, but may also be ignorable.

joeflateau:
I don't disagree that the warning should be generated by default if the element is not in the dom, but there should also be a way to suppress the warning if you have the element out of the dom on purpose. Once you start ignoring warnings because you can't suppress them then you're going to start ignoring warnings that you should actually be correcting. Perhaps instead of introducing a new option here we can instead change the logic so if there is a VIDEO element out of dom then the warning is generated but if there is an AUDIO element out of the dom then no warning is generated?

PR #7832, opened 2022-07-07
chore(lock-threads): run only daily at 1:00 am, and skip in forks

amtins:
## Description

The github action introduced by #7777 in charge of locking old issues runs every hour and is followed by an e-mail. 

github-lock-action

## Specific Changes proposed

The [Lock Threads](https://github.com/dessant/lock-threads) does not seem to allow restricting its execution to a named repo. The proposal would be to execute the action only once a day to avoid spamming those who have a fork of the project.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -13,6 +13,7 @@ concurrency:
 
 jobs:
   action:
+    if: ${{ github.repository_owner == 'videojs' }}
     runs-on: ubuntu-latest
     steps:
       - uses: dessant/lock-threads@v3
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -2,7 +2,7 @@ name: 'Lock Threads'
 
 on:
   schedule:
-    - cron: '0 * * * *'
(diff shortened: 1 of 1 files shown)


mister-ben:
Thanks, this doesn't need to run often now the backlog is cleared. Could have `if: ${{ github.repository_owner == 'videojs' }}` under `action` too.

gkatsev:
I wish actions didn't run in forked repos by default. Can definitely be annoying.

.github/workflows/lock.yml
```diff
@@ -13,6 +13,7 @@ concurrency:
 
 jobs:
   action:
+    if: ${{ github.repository_owner == 'videojs' }}
```
mister-ben, line 16: This will cause the jobs to be skipped on forks. They'll still show in the actions list.

amtins: @mister-ben thank you, that's even better !

mister-ben (approved):

gjanblaszczyk (approved):

PR #7853, opened 2022-07-26
feat: Remove fallback for Element.closest()

mister-ben:
## Description
Removes the fallback for `Element.closest()` which was only needed for IE.


gkatsev (approved):

misteroneill (approved):
Nice!

PR #7856, opened 2022-07-27
feat(lang): use less ambiguous text for the fullscreen button when in fullscreen mode

misteroneill:
## Description
This solves a small piece of technical debt introduced by #7183 by removing the code that replaced the poorly-worded "Non-Fullscreen" with "Exit Fullscreen"

## Specific Changes proposed
There are two commits here. The second removes all "Non-Fullscreen" keys from language files and is one I think we could consider reverting because it would mean the "Exit Fullscreen" text falls back to English in all cases in 8.0.

## Requirements Checklist


mister-ben (approved):
LGTM. It would be useful for testing, but not essential for this PR, to also update /sandbox/language.html to use ../dist/lang/es.js.

mister-ben (approved):

PR #7865, opened 2022-08-05
feat: Use picture el for poster

mister-ben:
## Description
Use a picture element instead of background image for the poster. Resolves issues some have had with the PosterImage and video el loading the poster differently.

Leaves scope to add support for source set for multiple image formats later, which could be done as a backwards-compatible minor.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/poster-image.js
```diff
@@ -46,12 +46,16 @@ class PosterImage extends ClickableComponent {
    *         The element that gets created.
    */
   createEl() {
-    const el = Dom.createEl('div', {
+    const crossOrigin = this.player_.options_.crossOrigin || this.player_.options_.crossorigin || null;
```
gkatsev, line 49: Should we use the `crossOrigin` getter?
```suggestion
    const crossOrigin = this.player_.crossOrigin() ?? null;
```

src/js/player.js
```diff
@@ -839,6 +839,9 @@ class Player extends Component {
     }
 
     this.techCall_('setCrossOrigin', value);
+    if (this.posterImage) {
+      this.posterImage.$('img').crossOrigin = value;
+    }
```
gkatsev, line 844: thought: maybe PosterImage should have a crossOrigin getter/setter so that this piece can be moved into it?
```
if (this.posterImage) {
  this.posterImage.crossOrigin(value);
}
```
If we do, we'll probably want to update the createEl below to use this option.

src/js/poster-image.js
mister-ben: The poster can be set up before the getter will return a correct value, as it's dependent on the tech

gkatsev: Ah, that's right. However, we should probably still use the getter first because otherwise, the value may be out of sync with the rest of the system.

mister-ben:
Updated to add a getter/setter to posterImage. The getter gets posterImage's value if the el has been created, otherwise the player/tech's vlaue if the tech is ready, otherwise from the player options.
Updated the player setter to accept null as a value, which is needed to unset the value.

misteroneill (approved):

gkatsev (approved):

PR #7868, opened 2022-08-08
feat: make retryOnError be the default

gkatsev:
This means that a retryOnError is removed and is no longer needed for
this behavior, which means that during source selection, if a source
fails, it'll try the next source that's available, to match the video
element.

BREAKING CHANGE: remove retryOnError option, turn it on by default


misteroneill (approved):
Nice, glad it was simple.

PR #7879, opened 2022-08-17
feat: Enable sourceset by default

gkatsev:
Can still be disabled with `enableSourceset: false`


misteroneill (approved):
Tests are "failing" because Codecov _lies!_

gkatsev:
It's also funny that it went down, given that I clearly added extra tests. But I guess I technically added a tiny bit of code to the player.

PR #7899, opened 2022-08-28
fix(picture-in-picture-control): removes the component in non-compatible browsers

### LINKED ISSUE #7898: PictureInPictureToggle is greyed out in browsers that do not support the Picture-in-Picture API
### Description

The PictureInPictureToggle component should not be displayed when changing the order of components in the ControlBar in browsers that do not support the Picture-in-Picture API.

Prior to v7.19.0, it was possible to change this behavior by doing the following
```JavaScript
class PictureInPictureToggle extends (videojs.getComponent('PictureInPictureToggle')) {
    enable() {
        super.enable();
        this.show();
    }

    disable() {
        super.disable();
        this.hide();
    }
}
videojs.registerComponent('PictureInPictureToggle', PictureInPictureToggle);
```

Although this solution was not ideal if the `requestPictureInPicture` failed.

### Reduced test case

_No response_

### Steps to reproduce

```JavaScript
const video = document.createElement('video');
video.id = 'video';
video.classList.add('video-js');
document.body.append(video);
const p1 = videojs(video, {
    muted: true,
    controls: true,
    controlBar: {
        children: [
            'playToggle',
            'volumePanel',
            'pictureInPictureToggle',
            'currentTimeDisplay',
            'timeDivider',
            'durationDisplay',
            'progressControl',
            'liveDisplay',
            'seekToLive',
            'remainingTimeDisplay',
            'customControlSpacer',
            'playbackRateMenuButton',
            'chaptersButton',
            'descriptionsButton',
            'subsCapsButton',
            'audioTrackButton',
            'fullscreenToggle'
        ]
    }
}, () => {
    p1.src('https://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8')
});
```

### Errors

_No response_

7.20.2

### Video.js plugins used.

_No response_

Firefox

All

amtins:
## Description

Fix  #7898

## Specific Changes proposed

This change allows the PictureInPictureToggle component to no longer be displayed when the web browser does not support the Picture-in-Picture API and the display order is changed in the ControlBar. 

Finally, commit e703ebda71f40eddafd628e5e134824ca9729a5e extracts a listener for `audio mode change` and removes an indentation level.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/control-bar/picture-in-picture-toggle.js
```diff
@@ -26,23 +26,17 @@ class PictureInPictureToggle extends Button {
...
     super(player, options);
-    this.on(player, ['enterpictureinpicture', 'leavepictureinpicture'], (e) => this.handlePictureInPictureChange(e));
-    this.on(player, ['disablepictureinpicturechanged', 'loadedmetadata'], (e) => this.handlePictureInPictureEnabledChange(e));
 
-    this.on(player, ['loadedmetadata', 'audioonlymodechange', 'audiopostermodechange'], () => {
-      // This audio detection will not detect HLS or DASH audio-only streams because there was no reliable way to detect them at the time
-      const isSourceAudio = player.currentType().substring(0, 5) === 'audio';
+    // Removes the pictureInPictureToggle in browsers that do not support the Picture-in-Picture API.
+    if (!('exitPictureInPicture' in document)) {
+      this.dispose();
```
amtins, line 32: An alternative solution would be to
```javaScript
      this.player_.ready(() => {
        this.parentComponent_.removeChild(this);
        this.dispose();
      });
```

This would avoid having a component not accessible in the `controlBar.children`

src/js/control-bar/picture-in-picture-toggle.js
```diff
@@ -56,7 +43,30 @@ class PictureInPictureToggle extends Button {
...
+    return `vjs-picture-in-picture-control vjs-hidden ${super.buildCSSClass()}`;
+  }
+
+  /**
+   * Displays or hides the button depending on the audio mode detection.
+   * Exits picture-in-picture if it is enabled when switching to audio mode.
+   */
+  handlePictureInPictureAudioModeChange() {
+    // This audio detection will not detect HLS or DASH audio-only streams because there was no reliable way to detect them at the time
+    const isSourceAudio = this.player_.currentType().substring(0, 5) === 'audio';
```
mister-ben, line 55: We should probably update `isAudio_` when this matches as the source is set, then use `isAudio()` here.

amtins: @mister-ben it make sens. Your suggestion is to set the audio flag somewhere in 
https://github.com/videojs/video.js/blob/f1558c660bbc132b23dbd63b2f3cdf7b6e60d00b/src/js/player.js#L1582-L1629

Do you recommend doing it in another PR? If yes with what kind of scope? feat, refactor or fix?

I would tend to create another PR to keep modification more atomic. This would allow to try to better handle the HLS and Dash audio.

mister-ben (approved):
Codecov is being weird again, it's flagging a coverage change in an unrelated file that doesn't make any sense.

PR #7913, opened 2022-09-03
Add a player option keepTimeTooltipInSeekBar to prevent time tooltip overflow

Ami-OS:
Screenshot from 2022-09-04 02-31-24

## Description
Originally, the time tooltip was allowed to overflow into the seek bar, which is fine in general but for those who want to customize their skins, allowing overflow is not necessarily a good idea.

This PR allows developers to easily keep the time tooltip inside the seek bar with a single player configuration option.

## Specific Changes proposed
Add a player option `keepTimeTooltipInSeekBar`  which, if set to `true`, prevents the time tooltip overflow the seek bar

## Requirements Checklist
  - [ ] Docs/guides updated


phloxic:
For what it's worth, the behaviour proposed by this option is [now the default](https://github.com/videojs/video.js/pull/8527#pullrequestreview-1830635165).

PR #7939, opened 2022-09-24
fix: enter fullscreen from fullWindow does not trigger exitFullWindow

### LINKED ISSUE #7705: Going fullscreen from full window does not trigger exitFullWindow
## Description
Going from full window mode to fullscreen does not trigger `exitFullWindow()` 

## Steps to reproduce
1. Play the video 
2. going full window `enterFullWindow()`
3. going fullscreen `requestFullscreen()`

### Expected
```env
isFullWindow=false
```

### Actual
```env
isFullWindow=true
```

tapornanan:
## Description
Fix #7705

## Specific Changes proposed
When the player is playing in full window mode and then calls `requestFullscreen()` but the full window state not get reset

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2862,6 +2862,11 @@ class Player extends Component {
   requestFullscreen(fullscreenOptions) {
     const PromiseClass = this.options_.Promise || window.Promise;
 
+    // In case of requestFullscreen is happend when player is in fullWindow mode
+    if (this.isFullWindow) {
+      this.exitFullWindow();
+    }
+
     if (PromiseClass) {
       const self = this;
 


gkatsev:
This does make sense on the surface, but fullWindow was originally created as a fallback for lack of native fullscreen support and it's still there at the end of a bunch of if/else statements https://github.com/videojs/video.js/blob/b58a2201c4e4af76f2d754aa633f21456f60cbdc/src/js/player.js#L2929-L2932
So, a bit weird to exit it only to maybe enter it immediately after.

tapornanan:
@gkatsev Thanks you

In my case, There was a requirement that we need to **load videojs** and **play** the video in fullscreen whenever the user clicked on UI. It works fine on high-end/fast devices but in some case, there is some low-end/slow device that is failed to run `requestFullscreen()` within valid user gesture time therefore I add an extra check to go full window if the fullscreen is failed. Then we wait for the user to click on the play button in the video we will again trigger `requestFullscreen` which this time is usually successful because all scripts are already loaded. But the state of videojs is still shown that `isFullWindow` still true 😢

PR #7940, opened 2022-09-29
refactor: rename fn.bind to fn.bind_ to strongly indicate it should not be used externally

misteroneill:
## Description
The `utils/fn.bind` function will be exposed in Video.js 8 because `fn` is exposed as `videojs.fn`. We don't want people using it, but it cannot be removed without breaking the Video.js event system.

This PR adds a `_` at the end of its name to indicate that it is private.

No external updates should be needed as it was not _directly_ exposed. The previously-supported `videojs.bind` has been deprecated.

## Specific Changes proposed
- Rename `utils/fn.bind` to `utils/fn.bind_`
- Update usages

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -1321,7 +1321,7 @@ class Component {
     }
 
     // listener for reporting that the user is active
-    const report = Fn.bind(this.player(), this.player().reportUserActivity);
+    const report = Fn.bind_(this.player(), this.player().reportUserActivity);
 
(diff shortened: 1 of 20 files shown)


gkatsev:
Rename to `Fn.bind_or_you_will_be_fired`

gkatsev (approved):

PR #7946, opened 2022-10-03
chore: fix mixed content warnings from netlify

gkatsev:
Netlify will email me if it detects mixed content in html it hosts. We fixed the stuff in sandbox/ previously, but the legacy docs apparently still had some. Mostly does some google fonts and vjs cdn http->https.


Essk (approved):
> Netlify will email me if it detects mixed content in html it hosts.
RIP your inbox 😬 

Changes all LGTM

gkatsev:
Thankfully, it only sent one email per PR and not one email per item 😆

gkatsev:
Thanks for the review!

PR #7949, opened 2022-10-03
feat: Deprecate extend() in favor of ES6 classes

alex-barstow:
## Description
`extend()` will be [removed](https://github.com/videojs/video.js/pull/7950) in Video.js 8 in favor of ES6 classes, so we should deprecate it in 7.x.


misteroneill:
Closed in favor of #7944

PR #7956, opened 2022-10-12
fix: remove unnecessary handling of invalid cues

mister-ben:
## Description
We had some special handling of cues which have equal start and end times, so they would be considered active 0.5 after they start/end. Cues with equal start and end times are not valid; the end time must be greater. However some painton style VTT in the wild contain many such cues, resulting in all such with a 0.5 seconds being displayed.

We thought this handling might still be needed for ID3 cues as ID3 only has a start time, but it's not: on adding to the track [the `endTime` is set](https://github.com/videojs/http-streaming/blob/9db9e99f1568049502058ce8312aff2dae77beab/src/util/text-tracks.js#L213) to the next cue's `startTime` or the video duration.

## Specific Changes proposed
Removes the handling of these cues. A cue with equal start and end times is not strictly ignored, but it will only render if active cues are updated at the precise millisecond they are to be active.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track.js
+++ b/src/js/tracks/text-track.js
@@ -326,10 +326,6 @@ class TextTrack extends Track {
 
             if (cue.startTime <= ct && cue.endTime >= ct) {
               active.push(cue);
-            } else if (cue.startTime === cue.endTime &&
-                       cue.startTime <= ct &&
-                       cue.startTime + 0.5 >= ct) {
-              active.push(cue);
             }
           }
 
--- a/test/unit/tracks/text-track.test.js
+++ b/test/unit/tracks/text-track.test.js
@@ -253,6 +253,43 @@ QUnit.test('can only remove one cue at a time', function(assert) {
   assert.equal(tt.cues.length, 0, 'we have removed the other instance of cue1');
 });
 
+QUnit.test('does not include past cues in activeCues', function(assert) {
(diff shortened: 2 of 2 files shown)


gkatsev (approved):

OwenEdwards:
Hmm, interesting. I hadn't thought about this before, and it's a separate (but related) issue - does the current way of handling text track cues work correctly for cues whose duration is less than the rate that this display method is called?

For example, if the timerupdate event is triggered every 1/4th of a second (I think that's what it is, or at least it used to be), and if a cue's end time is, say, 150mS after its start time, then it'll be entirely unpredictable whether that cue is displayed for 250mS or not at all, depending on where the timerupdate ticks occur in relation to the cue start time. Is that the right behavior? Is there anything in the text track spec about that? Just wondering 🤔

gkatsev:
Given that @mister-ben added requestVideoFrameCallback/requestAnimationFrame support (with fallback), it's likely going to get called enough in most cases.
That said, the HTML spec says to skip missed cues.

Actually, while we're here, I realized we should make the end time exclusive.

mister-ben:
Anything like adding time to caption durations to ensure they are displayed longer would only cause issues. There has to be an onus on the vtt authoring to ensure timings are long enough with a margin of error - no implemention could be more precise than the monitor refresh rate.

PR #7979, opened 2022-10-28
feat: Use userAgentData in favour of userAgent

### LINKED ISSUE #7572: Google Chrome User-Agent Deprecation
## Description
Google Chrome is changing the way it shares user-agents for increased privacy of users. You can read more about it here: https://www.chromestatus.com/feature/5704553745874944. Google Chrome has released Client Hints to provide device information.

1. videojs uses navigator.userAgent for detection
2. chrome will deprecate making this test break/unreliable
3. videojs thus needs to adapt their tests to take this change into account.

mister-ben:
## Description
Although we're not doing anything nefarious, switch to `userAgentData` where available so Chrome doesn't pester users about accessing `userAgent`. 

Deprecates `IE_VERSION` and `IS_CHROME`. 

`IS_CHROME` isn't useful as present, as matches any Chromium browser except Edge. The intention was to exclude legacy Edge from the test

Adds `IS_CHROMIUM` as explicitly any Chromium browser.

BREAKING CHANGE: `ANDROID_VERSION` is `null` on browsers supporting `userAgentData` becuase the version is not accessible from `userAgentData`

`ANDROID_VERSION` was only used in one place in video.js, where `IS_ANDROID` is sufficent anyway.

Only use of `IE_VERSION` in non-archived repos in the videojs org is in VHS. `ANDROID_VERSION` was not used 
* VHS: https://github.com/videojs/http-streaming/pull/1343

Fixes #7572

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/utils/browser.js
```diff
@@ -5,196 +5,198 @@
...
-
 /**
  * Whether or not this device is an iPod.
  *
  * @static
- * @const
  * @type {Boolean}
  */
-export const IS_IPOD = (/iPod/i).test(USER_AGENT);
+export let IS_IPOD = false;
```
mister-ben, line 14: If we'd rather these remain as `@const` in jsdoc, these could be temp vars and exported as consts later instead.

src/js/utils/browser.js
```diff
@@ -5,196 +5,198 @@
...
+
+  IS_ANDROID = UAD.platform === 'Android';
+  IS_CHROME = Boolean(UAD.brands.find(b => b.brand === 'Google Chrome'));
+  IS_EDGE = Boolean(UAD.brands.find(b => b.brand === 'Microsoft Edge'));
+  CHROME_VERSION = (UAD.brands.find(b => b.brand === 'Chromium') || {}).version || null;
+  IS_WINDOWS = UAD.platform === 'Windows';
+}
+
+if (!CHROME_VERSION) {
+  // no userAgentData, or a browser that has added userAgentData since this was added
```
mister-ben, line 140: i.e. if Firefox adds `userAgentData`, it might specify its brand as `Firefox` or `Mozilla Firefox` so we can't really add in advance.

src/js/utils/browser.js
```diff
@@ -5,196 +5,198 @@
...
  * @static
- * @const
  * @type {Boolean}
  */
-export const IS_CHROME = !IS_EDGE && ((/Chrome/i).test(USER_AGENT) || (/CriOS/i).test(USER_AGENT));
+export let IS_CHROME = false;
 
 /**
  * The detected Google Chrome version - or `null`.
+ * This has always been the _Chromium_ version, i.e. would return on Chromium Edge
```
mister-ben, line 71: Maybe we should have separate `IS_CHROMIUM` and an `IS_CHROME` that is just Chrome. Is "Chromium but not Edge" useful when it captures things like Opera which may have its own quirks?

src/js/tech/html5.js
```diff
@@ -1332,7 +1329,7 @@ Html5.patchCanPlayType = function() {
 
   // Android 4.0 and above can play HLS to some extent but it reports being unable to do so
```
gkatsev, line 1330: I wonder if this is still the case or if this block can be removed.

mister-ben: Yeah, this was for the browser before chrome.

src/js/utils/browser.js
gkatsev: I think that's reasonable. We can probably also cheat and declare it a const in the jsdoc.

src/js/tech/html5.js
```diff
@@ -1325,38 +1322,6 @@ Html5.prototype.featuresTimeupdateEvents = true;
  */
 Html5.prototype.featuresVideoFrameCallback = !!(Html5.TEST_VID && Html5.TEST_VID.requestVideoFrameCallback);
 
-// HTML5 Feature detection and Device Fixes --------------------------------- //
-let canPlayType;
-
-Html5.patchCanPlayType = function() {
```
misteroneill, line 1331: Nice catch.

src/js/utils/browser.js
misteroneill: I would expect that when people do `IS_CHROME` they are generally looking for `IS_CHROMIUM`.

src/js/utils/browser.js
```diff
@@ -5,196 +5,233 @@
...
+
+  IS_ANDROID = UAD.platform === 'Android';
+  IS_EDGE = Boolean(UAD.brands.find(b => b.brand === 'Microsoft Edge'));
+  IS_CHROMIUM = Boolean(UAD.brands.find(b => b.brand === 'Chromium'));
+  IS_CHROME = !IS_EDGE && IS_CHROMIUM;
+  CHROME_VERSION = (UAD.brands.find(b => b.brand === 'Chromium') || {}).version || null;
+  IS_WINDOWS = UAD.platform === 'Windows';
+}
+
+if (!CHROME_VERSION) {
```
misteroneill, line 160: I'm a bit confused about why this block is so long and how `!CHROME_VERSION` acts as a proxy for all the things inside here. Should it be `CHROMIUM_VERSION` instead?

misteroneill (commented):
One question.

src/js/utils/browser.js
mister-ben: Yes, and if anything I'm surprised "Chrome but not Edge" doesn't cause problems somewhere on Edge. Added `IS_CHROMIUM` and deprecated `IS_CHROME`

src/js/utils/browser.js
mister-ben: Only Chromium browsers support `userAgentData` now. The tests in this block are repeating the same tests with userAgent (for Chromium < 90) or for things that can't be true if `userAgentData` is set and it's Chromium. We can use `IS_CHROMIUM` here as it's a bit clearer.

If and when other browsers add `userAgentData` and we know the values e.g. if Firefox's brand string would be "Mozilla Firefox" or "Firefox" then this will need to be updated.

misteroneill (approved):
Thanks for the explanation!

Issue #7989, opened 2022-11-03
 The bar is not centered in video control section


KangXinzhi:
### Description

<img width="546" alt="image" src="https://user-images.githubusercontent.com/50164727/199654693-acde1ac4-2a77-40a0-a572-a1b3fc761f48.png">

<img width="744" alt="image" src="https://user-images.githubusercontent.com/50164727/199654592-525ba725-d905-4a66-a9b8-549c6c444e0e.png">

### Steps to reproduce

1.
2.
3.

### What version of Video.js are you using?

^7.20.3

### What browser(s) including version(s) does this occur with?

chrome 104.0.5112.81

### What OS(es) and version(s) does this occur with?

mac

KangXinzhi:
i try fixed
<img width="622" alt="image" src="https://user-images.githubusercontent.com/50164727/199654831-f93034e2-e285-4a7b-8c3d-83dee2e5acb6.png">

hover：
<img width="679" alt="image" src="https://user-images.githubusercontent.com/50164727/199654892-d553d019-9e4d-41e7-8588-b755b0bb64de.png">

mister-ben:
Fixed in #7990, which will be in the next v8.x release.

CSS like below could be used for v7.x players.

```css
    .video-js .vjs-volume-level:before,
    .video-js .vjs-play-progress:before {
      top: unset !important;
      line-height: .35em;
    }
```

Issue #7004, opened 2020-12-21
The progress-dot's position is not right


SongJuXin:
## Description
The progress-dot's position is not right,and the same phenomenon is occurs in volume-bar. 
![image](https://user-images.githubusercontent.com/20874344/102781500-bc85a380-43d2-11eb-84e8-7288b4e7f96f.png)

My archive on the base of videojs as follows.
![image](https://user-images.githubusercontent.com/20874344/102780051-3e280200-43d0-11eb-9a15-140ada1159d0.png)
https://codepen.io/SongJuXin/pen/WNGEMRX is the online show.

## Steps to reproduce
Explain in detail the exact steps necessary to reproduce the issue.
1.play video
2.observe progress-dot 

### Expected
 Progress-dot  is vertical middle.

### Actual
Progress-dot  is dropdown.

### resolve method
reason:`\f111` character has white gap.
resolve method:update the  font file is used  of the section.
 ```css
.vjs-icon-circle:before, .video-js .vjs-mouse-display:before, .video-js .vjs-play-progress:before, .video-js .vjs-volume-level:before {
    content: "\f111";
}
```

gkatsev:
Looks like you're testing with Video.js 5 which is a really old version at this point. Latest Video.js does not seem to have this problem since the dot is made purely using CSS, without any font icons.
I would recommend upgrading because there have been many improves since the last v5 release, almost 3 years ago.

SongJuXin:
@gkatsev Thanks reply.
Now,I replace with the lastest v7.10.2.
![image](https://user-images.githubusercontent.com/20874344/102845770-3bb5be80-4449-11eb-8f5d-7ed3c7ca07b1.png)
The same problem occured again.
![image](https://user-images.githubusercontent.com/20874344/102845934-a6ff9080-4449-11eb-9cb7-4d82ec996b46.png)

gkatsev:
The way you create the large progress bar is not equivalent to how it is done currently.
If we set the player's font-size really large you can see that it is nicely centered on top of the progress control https://codepen.io/gkatsev/pen/wvzrdxq

SongJuXin:
Ok,my previous achieve have some flaw.But the small size control actully has position problem.
This is origin image.
![normal](https://user-images.githubusercontent.com/20874344/102962731-69bb0180-4522-11eb-9785-f32addbfd27f.png)
This is scale 10x image.
[...]

PR #7990, opened 2022-11-03
fix(control-bar): align center

KangXinzhi:
## Description
[Please describe the change as necessary.
The bar is not centered in video control section
[issues7989](https://github.com/videojs/video.js/issues/7989)

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -75,7 +75,7 @@
     font-size: 0.9em;
     position: absolute;
     right: -0.5em;
-    top: -0.333333333333333em;
+    line-height: .35em;
     z-index: 1;
   }
 }
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -139,7 +139,7 @@
 
   // Volume handle
   &:before {
-    top: -0.3em;
+    line-height: .35em;
     right: -0.5em;
   }
 }


mister-ben (changes requested):
This causes the opposite issue for me: the handle is too high. There's been a few reports of this before #7004 which users have fixed with some CSS. It seems specific to Chinese OS/page. I wonder if there's something different about font rendering with Chinese characters which affects em units.

KangXinzhi:
> too high. There's been a few reports of th

When I open the videoJS official website, I have the same problem.  and it has not any Chinese.


image

mister-ben:
I can replicate what you see if I run Chrome in Chinese Traditional on MacOS - the application rather than the page language is apparently what matters. I note the handle is slightly larger.

Removing those two `top` rules and replacing them with `line-height: .35em;` seems to work well with the browser in English, Chinese, Arabic, Hindi, Japanese on Mac Chrome, Safari, Firefox and Edge. Would need some testing on Windows - if it works there two, it looks like a solution. It should be merged into `next` rather than `main` - it could be a breaking change for custom styles.

mister-ben:
In this codepen, the first player uses the styling we currently have, and the second uses `line-height` instead of `top`. On all MacOS browsrs I've tried, the slider handles are in the correct place regarless of OS/app language for the second player.

I haven't tried on Windows as the standard edition of Windows doesn't let you change the OS/app language. If @KangXinzhi  or anyone else could try this on Chinese (or other non-Latin script) Windows browsers it would be useful.

https://codepen.io/mister-ben/full/MWXeOKr

KangXinzhi:
> 
It is great when I try this on Chinese Windows browsers！
image

image

src/css/components/_progress.scss
```diff
@@ -49,6 +49,10 @@
   font-size: 1.666666666666666666em;
 }
 
+.video-js .vjs-progress-control:hover .vjs-play-progress:before{
+  top: -0.333333333333333em;
+}
+
```
mister-ben, line 55:
```suggestion
```

src/css/components/_progress.scss
```diff
@@ -79,8 +83,7 @@
     font-size: 0.9em;
     position: absolute;
     right: -0.5em;
-    top: -0.333333333333333em;
-    z-index: 1;
+    top: -0.45em;
```
mister-ben, line 86:
```suggestion
    line-height: .35em;
    z-index: 1;
```

src/css/components/_volume.scss
```diff
@@ -157,7 +157,7 @@
 
   // Volume handle
   &:before {
-    top: -0.3em;
+    top: -0.45em;
```
mister-ben, line 160:
```suggestion
    line-height: .35em;
```

mister-ben (changes requested):
These suggested changes are as my codpen example.

mister-ben:
This could be a breaking change for custom skins, so should be in a major release - v8.x should be very soon now. Could you rebase agains the next branch?

KangXinzhi:
> This could be a breaking change for custom skins, so should be in a major release - v8.x should be very soon now. Could you rebase agains the next branch?

that's ok!

mister-ben (approved):

misteroneill (approved):
Took a quick look and seems to work nicely. Thank you @KangXinzhi!

PR #7991, opened 2022-11-03
fix(lang): Improve Persian translation

ebraminio:
## Description
Minor typography related improvements.

## Specific Changes proposed
In Persian compound words are joined with Zero-width non-joiner character (\u200c) and this change uses it wherever needed. Also Persian uses a different comma than the English keyboard one and this uses that instead.

Also used the missing diacritics according to Persian Academy's orthography guidelines.

Feel free to read more about the added character here, https://en.wikipedia.org/wiki/Zero-width_non-joiner


mister-ben (approved):

ebraminio:
Great, thanks 😊

PR #8009, opened 2022-11-20
feat: text track display overlays a video

gjanblaszczyk:
## Description
Text track display does not always overlay a video which may generate a different text track size and position across different browsers and devices.

## Specific Changes proposed
The text track display always overlays a video.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -78,6 +78,21 @@ function tryUpdateStyle(el, style, rule) {
 }
 
 /**
+ * Converts the CSS top/right/bottom/left property numeric value to string in pixels.
+ *
+ * @param {number} position
+ *        The CSS top/right/bottom/left property value.
+ *
+ * @return {string}
+ *          The CSS property value that was created, like '10px'.
+ *
+ * @private
+ */
+function getCSSPositionValue(position) {
+  return position ? `${position}px` : '';
+}
+
+/**
  * The component for displaying text track cues.
  *
  * @extends Component
@@ -99,11 +114,18 @@ class TextTrackDisplay extends Component {
   constructor(player, options, ready) {
     super(player, options, ready);
 
-    const updateDisplayHandler = (e) => this.updateDisplay(e);
+    const updateDisplayTextHandler = (e) => this.updateDisplay(e);
+    const updateDisplayHandler = (e) => {
+      this.updateDisplayOverlay();
+      this.updateDisplay(e);
+    };
 
     player.on('loadstart', (e) => this.toggleDisplay(e));
-    player.on('texttrackchange', updateDisplayHandler);
-    player.on('loadedmetadata', (e) => this.preselectTrack(e));
+    player.on('texttrackchange', updateDisplayTextHandler);
+    player.on('loadedmetadata', (e) => {
+      this.updateDisplayOverlay();
+      this.preselectTrack(e);
+    });
 
     // This used to be called during player init, but was causing an error
     // if a track should show by default and the display hadn't loaded yet.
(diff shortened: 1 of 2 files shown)


src/js/tracks/text-track-display.js
```diff
@@ -294,6 +316,36 @@ class TextTrackDisplay extends Component {
...
+    }
+
+    const playerWidth = this.player_.currentWidth();
+    const playerHeight = this.player_.currentHeight();
+    const playerAspectRatio = playerWidth / playerHeight;
+    const videoAspectRatio = this.player_.videoWidth() / this.player_.videoHeight();
+    let leftRight = 0;
+    let topBottom = 0;
+
+    if (Math.abs(playerAspectRatio - videoAspectRatio) > 0.01) {
```
gkatsev, line 335: I think this can probably be 0.1.

gkatsev: Specifically, I think there should be a lot more leeway in being a bit off from the exact aspect ratio of the video but still taking up the entire player size.
For example, on the [sandbox/index page](https://deploy-preview-8009--videojs-preview.netlify.app/sandbox/index.html), the player aspect ratio is 2.42 compared to the video's 2.4, which squishes the text track display by 3 pixels on each side, but I think this should be more than reasonable to be the full size.

src/js/tracks/text-track-display.js
```diff
@@ -294,6 +316,36 @@ class TextTrackDisplay extends Component {
...
+        leftRight = Math.round((playerWidth - playerHeight * videoAspectRatio) / 2);
+      } else {
+        topBottom = Math.round((playerHeight - playerWidth / videoAspectRatio) / 2);
+      }
+    }
+
+    tryUpdateStyle(this.el_, 'top', getCSSPositionValue(topBottom));
+    tryUpdateStyle(this.el_, 'bottom', getCSSPositionValue(topBottom));
+    tryUpdateStyle(this.el_, 'left', getCSSPositionValue(leftRight));
+    tryUpdateStyle(this.el_, 'right', getCSSPositionValue(leftRight));
```
gkatsev, line 346: can use `insetInline` instead of `left` and `right`.

In CSS, it's:
```
.vjs-text-track-display: {
  inset-inline: 5px;
}
```

src/js/tracks/text-track-display.js
```diff
@@ -294,6 +316,36 @@ class TextTrackDisplay extends Component {
...
+    if (Math.abs(playerAspectRatio - videoAspectRatio) > 0.01) {
+      if (playerAspectRatio > videoAspectRatio) {
+        leftRight = Math.round((playerWidth - playerHeight * videoAspectRatio) / 2);
+      } else {
+        topBottom = Math.round((playerHeight - playerWidth / videoAspectRatio) / 2);
+      }
+    }
+
+    tryUpdateStyle(this.el_, 'top', getCSSPositionValue(topBottom));
+    tryUpdateStyle(this.el_, 'bottom', getCSSPositionValue(topBottom));
```
gkatsev, line 344: can use `insetBlock` instead of separate `top` and `bottom`.

In CSS, it's:
```
.vjs-text-track-display: {
  inset-block: 5px;
}
```

gkatsev: I guess the `topBottom` var can also be renamed to match.

src/js/tracks/text-track-display.js
gkatsev: I guess the `leftRight` var can also be renamed to match.

gkatsev (commented):
Awesome, this works well!
Had a couple of minor comments.

src/js/tracks/text-track-display.js
gjanblaszczyk: It is a good idea. I have already changed that.

src/js/tracks/text-track-display.js
gjanblaszczyk: Thanks, It is a great idea. I have already changed that.

src/js/tracks/text-track-display.js
gjanblaszczyk: Yes, it sounds reasonable. I have already changed 0.01 to 0.1.

gkatsev (approved):

PR #8015, opened 2022-11-23
chore: re-generate package-lock.json to fix merge issues with main

misteroneill:
## Description
After rebasing `next` on `main`, there were issues with the `package-lock.json` (not too surprising), so this re-generates it afresh using npm 6.14.17 (the default version that ships with Node 14).

## Specific Changes proposed
- Re-generate `package-lock.json` to put us in a good place for Video.js 8.0


gkatsev (approved):
Locally, I get the same package-lock, though, with sha1 instead of sha 512 hashes. But my npm is 6.14.15.

PR #8043, opened 2022-12-12
fix: Set alt attr on poster img

mister-ben:
## Description
As the poster now uses an img element, it should have an alt attribute.

## Specific Changes proposed
Sets a null alt attribute on the poster image. This conveys to screen readers (and accessibility validators) that the image is decorative and does not need descriptive text.

## To do in a future update (needs discussion)
- Allow an arbitrary string to be set, so that an image can have a meaningful alt text where appropriate. Requires update to poster/player options.

## Requirements Checklist
  - [ ] Docs/guides updated


OwenEdwards:
The `` element should definitely have an `alt` attribute, but as far as whether it should have text or just a "*null alt*" (`alt=""`), that's a little complicated.

There was a lot of discussion at W3C about whether the `<video>` element should accept an `alt` attribute for the `poster` image, but the decision was that the poster image is an alternate for the video itself, so the title of the video (either as a `title` attribute or an `aria-label`/`aria-labelledby` on the `video` element) was effectively the text alternative for the image. That was under the specific expectation that the `poster` image is always a still from the video and not another image. In that case, I personally think the term "poster" is misleading because many people think of the movie theater poster when they hear that term related to a video. A movie theater poster is rarely just a still from the video and often includes text which isn't available to non-sighted users anywhere else. For example (from https://en.wikipedia.org/wiki/Gone_with_the_Wind_(film)):

The theatrical pre-release poster for the film Gone With the Wind, from Wikipedia

So, I think the *default* should be to make it a *null alt*, but there should be an option to change that to a description of the poster image.

mister-ben:
Thanks @OwenEdwards, very useful feedback. Let's keep this simple as a quick fix with the null alt and figure out how best to allow a more meaningful value to be set as a feature.

src/js/poster-image.js
```diff
@@ -51,9 +51,13 @@ class PosterImage extends ClickableComponent {
...
       // Don't want poster to be tabbable.
       tabIndex: -1
-    }, {}, Dom.createEl('img', {
+    }, {
+      alt: 'some alt text'
+    }, Dom.createEl('img', {
       loading: 'lazy',
       crossOrigin: this.crossOrigin()
+    }, {
+      alt: ''
```
mister-ben, line 60: This will insert `alt` rather than `alt=""`. Do we know if all validators would treat this as `null alt`?

mister-ben:
The codecov failures are codecov hitting rate limits. https://github.com/codecov/codecov-action/issues/837

src/js/poster-image.js
OwenEdwards: I don't think it will pass validation; I _think_ just inserting the attribute `alt` on an element sets alt to `true`, rather than an empty string, which isn't correct.

Would `alt: '""'` work?

mister-ben: That ends up as `alt=""""` or `alt='""'` depending on the browser. On Chrome `el.setAttribute('alt', '')` will reset back to `alt` even if a string was previously set. Firefox has `alt=""`. I think as far as the browser is concerned `alt` and `alt=""` are equal. Lighthouse is happy with just `alt`.

mister-ben: Even on Chome `el.getAttribute('alt')` returns `''` and `el.outerHTML` is `` so it ought to be ok... not sure if they're another way a validator could be testing that.

gkatsev: With
```html
<!DOCTYPE html>
<html lang="en-US">
<head>
<title>foo</title>
<body>
<img alt src="https://example.com/foo.png">
</body>
</html>
```
the w3c validator passes https://validator.w3.org/nu/

I'd things to be fine as long as we're using `el.setAttribute('alt', '')`.

mister-ben: And `Dom.createEl()` is using that.

gkatsev (approved):

OwenEdwards (approved):

PR #8048, opened 2022-12-16
docs: Update jsdoc template for better usability on mobile

mister-ben:
The current jsodc template isn't very usable on mobile, and scores poorly in Lighthouse/Google Search Console in particular because of the small text. The TUI theme is unmaintained, so I've switched to https://github.com/ankitskvmdam/clean-jsdoc-theme which is a more responsive/accessible.

Removed guides as they live on videojs.com now.

To do:
- [x] Display version number
- [x] Get rid of js file listing from front page
- [x] Compare paths, in case any redirects needed
- [x] Add logo

Preview: https://deploy-preview-8048--videojs-docs.netlify.app/


.jsdoc.json
```diff
@@ -1,16 +1,24 @@
...
+    "include": ["src/js/"],
     "includePattern": ".js$"
   },
   "opts": {
     "destination": "docs/api",
     "readme": "docs/index.md",
-    "template": "node_modules/tui-jsdoc-template",
-    "package": "package.json",
+    "template": "node_modules/clean-jsdoc-theme",
+    "package": "",
```
mister-ben, line 10: with this theme, this keeps the output in the root of the destination dir and not at package/version/

gkatsev:
I'm not a fan of the main page (specifically, the list of all files in the project), but otherwise, looks good 👍🏻

mister-ben:
Yeah, I was looking for an option to remove that big list.

misteroneill (commented):
I agree, the list is a bit too "in your face" for the entry point, but overall looks fantastic!

mister-ben:
The theme author is looking into why the file list is there, it's not intentional.

mister-ben:
The file list is gone. I've added a script to update the jsdoc config file so it invludes the current version number in the link bakc to github, since this tempalte doesn't otherwise display the number.

Essk (approved):

PR #8052, opened 2022-12-19
(WIP) feat: Allow picture sources to be set on poster

mister-ben:
Now that the PosterImage uses a poster, we can make use of it to support multiple image formats. Very much a work in progress, needs tests, and there may well be edge cases yet to be considered.

## Changes:

- The argument to the `poster()` setter may now be a poster options object. A string is still supported to be non-breaking.
- The poster options object should specify an array of objects as `sources`. The object properties will be used as attributes of `<source>` elements of the `PosterImage`'s `<picture>`.
- The poster options object may include a string `img`, which is the `src` used for the `<picture>`'s ``. If `sources` is set and `img` is not, `img` is set to the last item in `sources`.
- The poster options object may include a string `alt`, so alt text can be set on the ``.
- The poster options object may also be passed as a player setup option, `posterOptions`.
- Adds a `posterOpts()` getter to fetch the poster options. However this is _not_ a setter, since the options are set with `poster()`.
- if sources are used, `poster()` as a getter will return the image actually used.
- If sources are used, the poster attribute is not set on the tech, so the tech and poster do not load different images.
- If sources are not used, the poster attriute is optionally not set if a new player option `noTechPoster` is set to true. This is useful since the poster now uses laszy loading, so the video el doesn't load the image.

## Examples

```js
// Works as currently
player.poster('https://example.com/image.jpeg');
// Would have the same effect
player.poster({img: 'https://example.com/image.jpeg'});
// Would set a poster, and its alt text
player.poster({img: 'https://example.com/image.jpeg', alt: 'Descriptive text for a screen reader'});
```

```js
// The browser would use webp if it suppports it
player.poster({
  sources: [{
    srcset: 'https://example.com/image.webp',
    type: 'image/webp'
  },
  {
    srcset: 'https://example.com/image.jpeg',
    type: 'image/jpeg'
  }],
  img: 'https://example.com/image.jpeg'
});
```

```js
// Where the fallback image is omited, the last source's image is used. This would be
// equivalent to the last example.
player.poster({
  sources: [{
    srcset: 'https://example.com/image.webp',
    type: 'image/webp'
  },
  {
    srcset: 'https://example.com/image.jpeg',
    type: 'image/jpeg'
  }]
});
```

To do

- [ ] Tests
- [ ] jsdoc/types

### DIFF
--- a/sandbox/poster.html.example
+++ b/sandbox/poster.html.example
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <title>Video.js Sandbox</title>
+  <link href="../dist/video-js.css" rel="stylesheet" type="text/css">
+  <script src="../dist/video.js"></script>
+</head>
+<body>
+  <div style="background-color:#eee; border: 1px solid #777; padding: 10px; margin-bottom: 20px; font-size: .8em; line-height: 1.5em; font-family: Verdana, sans-serif;">
+    <p>You can use /sandbox/ for writing and testing your own code. Nothing in /sandbox/ will get checked into the repo, except files that end in .example (so don't edit or add those files). To get started run `npm start` and open the index.html</p>
+    <pre>npm start</pre>
+    <pre>open http://localhost:9999/sandbox/index.html</pre>
+  </div>
+
+  <video-js
+    id="vid1"
+    controls
+    preload="auto"
+    width="640"
+    height="264">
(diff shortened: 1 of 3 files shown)


gkatsev:
> If sources are used, the poster attribute is not set on the tech, so the tech and poster do not load different images.

What about on iPhones? Might be worth special-casing it and still setting it in that case.

Additionally, if the PosterImage component is turned off, we should set poster on the tech.

phloxic:
@mister-ben - what's the reason for closing this?

PR #8075, opened 2023-01-17
feat: Add middleware method to override currentSource()

Essk:
## Description
We've come across a use case whereby an advanced user would want to to substitute the `src` attribute of the video using `setSource` middleware, but also requires `player.currentSource()` to continue to report the original src value, globally. 

Because `setSource` is used to effect the substitution it follows that the user should also have a way to control what is reported by the corresponding "getter", i.e. `player.currentSource()` and middleware seems to be a natural place to implement this.

## Specific Changes proposed
* implement `middleware.currentSource` to run through user-provided middleware and run a reduce on whatever currentSource implementations exist
  * because `player.currentSource` doesn't delegate to the tech, this needs it's own middleware function. 
* `player.currentSource()` will return `middleware.currentSource()` if possible
  * noting that this will have knock-on effects to the values returned by `player.currentSrc` & `player.currentType`
* Add unit test

## Requirements Checklist
  - [x ] Change has been verified in an actual browser (Chrome, Firefox, IE)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3305,7 +3305,7 @@ class Player extends Component {
   handleSrc_(source, isRetry) {
     // getter usage
     if (typeof source === 'undefined') {
-      return this.cache_.src || '';
+      return this.currentSrc();
     }
 
     // Reset retry behavior for new source
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3569,7 +3569,9 @@ class Player extends Component {
    *         The current source object
    */
   currentSource() {
-    return this.cache_.source || {};
+    const mw = this.middleware_;
+
+    return middleware.currentSource(this.cache_.source, mw) || this.cache_.source || {};
   }
 
   /**
--- a/src/js/tech/middleware.js
+++ b/src/js/tech/middleware.js
@@ -82,6 +82,10 @@ export function setSource(player, src, next) {
   player.setTimeout(() => setSourceHelper(src, middlewares[src.type], next, player), 1);
 }
 
+export function currentSource(src, middleware) {
+  return middleware.reduceRight(middlewareIterator('currentSource'), src);
+}
+
 /**
  * When the tech is set, passes the tech to each middleware's `setTech` method.
  *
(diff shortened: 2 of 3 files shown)


alex-barstow:
Should we do anything to address the fact that `player.src()` will bypass this middleware, since it returns `cache_.src`? My concern is that it could be unintuitive to have `player.src()` return a different value than `player.currentSource()` and `player.currentSrc()`

Essk:
@alex-barstow , That's a really good point. I think I initially skipped over this because of the "setter" element of `player.src(source)` but looking at it again, I don't see any obvious reason the early return in `handleSrc_` is reading player.cache_.src directly rather than going via `player.currentSrc()`. I'll make the change and see if anything breaks 🤞

src/js/tech/middleware.js
```diff
@@ -82,6 +82,20 @@ export function setSource(player, src, next) {
...
+ * Allows user to override the returned value of player.currentSource() using middleware.
+ *
+ * @param {Tech~SourceObject} src
+ *         A source object.
+ * @param {Object[]} middleware
+ *        An array of middleware instances.
+ * @return {Tech~SourceObject}
+ *         A source object.
+ */
+export function currentSource(src, middleware) {
```
gkatsev, line 95: nit: the method signature doesn't match the other middleware methods, should be swapped
```suggestion
export function currentSource(middleware, src) {
```

src/js/player.js
```diff
@@ -3305,7 +3305,7 @@ class Player extends Component {
   handleSrc_(source, isRetry) {
     // getter usage
     if (typeof source === 'undefined') {
-      return this.cache_.src || '';
```
gkatsev, line 3308: I would be careful with this change. I don't remember all the details, but I believe `src` and `source` are specifically defined to be allowed to be different in some cases. I believe that maybe contrib-ads relies on that?

src/js/tech/middleware.js
gkatsev, line 95: thought: I was thinking about whether this should be a separate method or if `middleware.get` should be able to handle this. Though, I guess the difference is here we don't want it to actually get to the tech? Though, maybe that's that it gets it from the tech, but potentially we'd want a more generic `get` that can also accept a default value?

Essk:
This needs some more thorough analysis of the knock-on effects in modules that may rely on specific behaviour in this regard, closing

Issue #8068, opened 2023-01-11
Why do captions and subtitles not display in Safari browsers on macOS and iOS when using Video.js in this code?


rainentoyama:
### Description

The code provided sets up a Video.js player using the **<video>** element with the id "player" and uses the **videoOptions** object to configure the player. Within the **videoOptions** object, a subtitle track is defined using the **tracks** option, and a source for the video is defined using the **sources** option. Despite this configuration, captions and subtitles are not displaying in Safari browsers on macOS and iOS but it works fine in Chrome (macOS only), Firefox(macOS only). and Opera (macOS only).
What could be the cause of this issue and what steps could be taken to resolve this problem?

### Steps to reproduce

```
<link href="https://vjs.zencdn.net/7.20.3/video-js.css" rel="stylesheet" />
<script src="https://vjs.zencdn.net/7.20.3/video.js"></script>
<video id="player" class="video-js"></video>
<script>
var videoOptions = videojs("player", {
  poster: "https://example.com/poster.jpg",
  controls: true,
  responsive: true,
  autoplay: false,
  preload: "auto",
  tracks: [
    {
      src: "https://example.com/caption.vtt",
      default: "true",
      kind: "captions",
      srclang: "en",
      label: "English",
    },
  ],
  sources: [
    { src: "https://example.com/hls.m3u8", type: "application/x-mpegURL" },
  ],
});
</script>
```

### What version of Video.js are you using?

7.20.3

### Video.js plugins used.

none

### What browser(s) including version(s) does this occur with?

Safari

### What OS(es) and version(s) does this occur with?

macOS High Sierra, iOS 15.7.2

amtins:
@rainentoyama  have you tried adding `crossorigin: 'anonymous'` in the configuration and does it fix your problem? Please see https://videojs.com/guides/text-tracks/#text-tracks-from-another-domain

```JavaScript
var videoOptions = videojs("player", {
  poster: "https://example.com/poster.jpg",
  controls: true,
  crossorigin: 'anonymous', // add this to you configuration
  responsive: true,
  autoplay: false,
  preload: "auto",
  tracks: [
    {
      src: "https://example.com/caption.vtt",
      default: "true",
      kind: "captions",
      srclang: "en",
      label: "English",
    },
  ],
  sources: [
    { src: "https://example.com/hls.m3u8", type: "application/x-mpegURL" },
  ],
});
```

[...]

PR #8085, opened 2023-01-22
fix: loss of crossOrigin value when loadMedia is called

amtins:
## Description

This PR fixes the deletion of the `crossOrigin` value when `loadMedia` is called.

Fixes: https://github.com/videojs/video.js/issues/8086

## Specific Changes proposed
- stores the value of `crossOrigin` before `reset` deletes it
- restores the value of `crossOrigin` after `reset` has been called
- unit test to verify that the value of crossOrigin is restored :
  - when the `crossOrigin` method is called before `loadMedia`
  - when the `crossOrigin` value is passed via the options then `loadMedia` is called
  - when the video tag contains the `crossOrigin` property then `loadMedia` is called 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -4899,6 +4899,8 @@ class Player extends Component {
       return;
     }
 
+    const crossOrigin = this.crossOrigin();
+
     this.reset();
 
     // Clone the media object so it cannot be mutated from outside.
@@ -4912,6 +4914,10 @@ class Player extends Component {
         src: poster,
         type: getMimetype(poster)
       }];
+    }
+
+    if (crossOrigin) {
+      this.crossOrigin(crossOrigin);
     }
 
     if (src) {
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -3224,3 +3224,48 @@ QUnit.test('turning on audioPosterMode when audioOnlyMode is already on will tur
       assert.notOk(player.audioOnlyMode(), 'audioOnlyMode is false');
     });
 });
+
+QUnit.test('crossOrigin value should be maintained after loadMedia is called', function(assert) {
+  const fixture = document.getElementById('qunit-fixture');
+
+  const example1 = '<video id="example_1" class="video-js" preload="none"></video>';
+  const example2 = '<video id="example_2" class="video-js" preload="none"></video>';
+  const example3 = '<video id="example_3" class="video-js" crossorigin="anonymous" preload="none"></video>';
+
+  fixture.innerHTML += example1;
+  fixture.innerHTML += example2;
+  fixture.innerHTML += example3;
+
+  const tagExample1 = document.getElementById('example_1');
+  const tagExample2 = document.getElementById('example_2');
+  const tagExample3 = document.getElementById('example_3');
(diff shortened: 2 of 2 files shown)


amtins:
@misteroneill thank you for taking the time to review this PR 👍🏽 

> Perhaps an alternative would be calling player.crossOrigin() in the ready callback?

TL;DR Quick answer, yes it will work.

Maybe I can add a little more context regarding this PR. This PR is the result of this issue #8068 and the comments https://github.com/videojs/video.js/issues/8068#issuecomment-1378767568 and https://github.com/videojs/video.js/issues/8068#issuecomment-1379428897

According to the `loadMedia` documentation this function should `Populate the player using a MediaObject`.

However, this function does much more, it resets the player, but also deletes and rebuilds the video element. The consequence is the creation of a new playback session losing properties that are important for the developer and _interesting for the user playback experience_ (not detailed as it's out of the scope of this PR).

## The important property for the developer

`videojs` allows the `crossOrigin` parameter to be supplied as an instance `option` even though it's not [documented](https://videojs.com/guides/options/). A developer can expect this option to remain unchanged for the life of the player instance or until it is intentionally changed. Refer to 

https://github.com/videojs/video.js/blob/1491d71b26a407111104dde7dbbbc0ee49b4c0c8/src/js/player.js#L776-L777


### Demonstration

`src` and `loadMedia` produce two different behaviors depending on which one is used:


```javascript
const player videojs('player', {crossorigin : 'anonymous'});

player.el().firstChild.crossOrigin; // -> anonymous 
player.options().crossorigin; // -> anonymous 
player.crossorigin(); // -> anonymous

player.src('...')

player.el().firstChild.crossOrigin; // -> anonymous 
player.options().crossorigin; // -> anonymous 
player.crossorigin(); // -> anonymous

player.loadMedia('...')

player.el().firstChild.crossOrigin; // -> null 
player.options().crossorigin; // -> anonymous 
player.crossorigin(); // -> null

```

This means that these two functions behave very differently. `src` keeps the current playback session while `loadmedia` recreates a new playback session almost as if it were a new instance of the player, but without the `crossOrigin`, which leans more towards to a bug.

### Code references

The workflow leading to the loss of this property:

https://github.com/videojs/video.js/blob/1491d71b26a407111104dde7dbbbc0ee49b4c0c8/src/js/player.js#L1274

https://github.com/videojs/video.js/blob/866ef24b797320c397893acef37219a53aa7496d/src/js/player.js#L1125

The fix would have been more appropriate in `techOptions`, but that also involves modifying `html5.js` and in the end would not handle as many cases as the proposed fix.
https://github.com/videojs/video.js/blob/866ef24b797320c397893acef37219a53aa7496d/src/js/player.js#L1152-L1169

https://github.com/videojs/video.js/blob/866ef24b797320c397893acef37219a53aa7496d/src/js/player.js#L1196

gkatsev:
While we're at it, should the object that loadMedia takes accept a crossOrigin property to force the value to something specific? This would be an additional feature and can come in a separate PR.

I think this change is fine as is.

misteroneill:
I'm going to accept your comment as a second approval @gkatsev 😄

PR #8101, opened 2023-01-31
feat: improved text tracks settings labels

wseymour15:
## Description
Based off of an audit that a Brightcove client received, we are updating the text tracks settings menu with improved labels. This change is regarding Web Content Accessibility Guidelines (WCAG). 

Screen Shot 2023-01-31 at 3 08 32 PM

## Specific Changes proposed
- Labels added/improved to fields in the settings menu
- CSS so labels are displayed next to inputs
- Translations for labels
- Updates docs regarding missing translations.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -86,6 +86,10 @@
   "Seek to live, currently playing live": "Zur Live-Übertragung wechseln. Es wird aktuell live abgespielt.",
   "Exit Picture-in-Picture": "Bild-im-Bild-Modus beenden",
   "Picture-in-Picture": "Bild-im-Bild-Modus",
-  "No content": "Kein Inhalt"
+  "No content": "Kein Inhalt",
+  "Color": "Farbe",
+  "Opacity": "Opazität",
+  "Text Background": "Text-Hintergrund",
+  "Caption Area Background": "Hintergrund des Untertitelbereichs"
 }
 
--- a/lang/de.json
+++ b/lang/de.json
@@ -88,8 +88,8 @@
   "Picture-in-Picture": "Bild-im-Bild-Modus",
   "No content": "Kein Inhalt",
   "Color": "Farbe",
-  "Opacity": "Opazität",
-  "Text Background": "Text-Hintergrund",
+  "Opacity": "Deckkraft",
+  "Text Background": "Texthintergrund",
   "Caption Area Background": "Hintergrund des Untertitelbereichs"
 }
 
--- a/lang/en.json
+++ b/lang/en.json
@@ -87,5 +87,9 @@
   "{1} is loading.": "{1} is loading.",
   "Exit Picture-in-Picture": "Exit Picture-in-Picture",
   "Picture-in-Picture": "Picture-in-Picture",
-  "No content": "No content"
+  "No content": "No content",
+  "Color": "Color",
+  "Opacity": "Opacity",
+  "Text Background": "Text Background",
+  "Caption Area Background": "Caption Area Background"
 }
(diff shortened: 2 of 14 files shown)


gkatsev:
Not that it shouldn't change, but the names are taken from the CVAA, which has these user settings requirements.

OwenEdwards:
Personally, I think the text changes are fine. But my concern with the layout is that the "Font Size", "Text Edge Style", and "Font Family" controls are so close to the three "Opacity" controls that it isn't clear visually that the "Opacity" fields are each part of a fieldset with the legend "Text", "Text Background", and "Caption Area Background". That could be fixed by increasing the spacing, or adding an outline for each fieldset.

wseymour15:
> That could be fixed by increasing the spacing, or adding an outline for each fieldset.

Good suggestion, I updated the CSS a bit to increase the spacing between some of these fields:

Screen Shot 2023-02-01 at 10 17 25 AM

mister-ben:
It looks quite cramped at a smallish size. It might look better with color and opacity to be slightly indented at this size. Maybe text/text background in as bold heading?

grafik

lang/de.json
```diff
@@ -86,6 +86,10 @@
   "Seek to live, currently playing live": "Zur Live-Übertragung wechseln. Es wird aktuell live abgespielt.",
   "Exit Picture-in-Picture": "Bild-im-Bild-Modus beenden",
   "Picture-in-Picture": "Bild-im-Bild-Modus",
-  "No content": "Kein Inhalt"
+  "No content": "Kein Inhalt",
+  "Color": "Farbe",
+  "Opacity": "Opazität",
+  "Text Background": "Text-Hintergrund",
```
mister-ben, line 92:
```suggestion
  "Opacity": "Deckkraft",
  "Text Background": "Texthintergrund",
```

lang/en.json
```diff
@@ -87,5 +87,9 @@
   "{1} is loading.": "{1} is loading.",
   "Exit Picture-in-Picture": "Exit Picture-in-Picture",
   "Picture-in-Picture": "Picture-in-Picture",
-  "No content": "No content"
+  "No content": "No content",
+  "Color": "Color",
```
mister-ben, line 91: We're going to need an en-GB :)

wseymour15: Good call, created this issue here: https://github.com/videojs/video.js/issues/8105

wseymour15:
> It looks quite cramped at a smallish size. It might look better with color and opacity to be slightly indented at this size. Maybe text/text background in as bold heading?
You're right, that does not look as expected, and nor did a lot of mobile devices when the menu was opened full screen. I updated this with bolded headers and a few small styling changes to make it look better in most cases. The only flaw I found is that in some medium sized players, the user is now forced to scroll to get to the bottom of the menu. If this is an issue let me know, but I believe in most cases this will look much better.
Screen Shot 2023-02-02 at 1 14 20 PM

lang/en.json
gkatsev: I think mastodon using british naming balances things out 😉

gkatsev:
I think it's fine for now, even with the scrolling. I built it out 10 years and 2 months ago, and it needs to be re-designed from scratch at this point.

roman-bc-dev:
I agree with @mister-ben that a small indent to the left of the labeled settings (color, opacity) would also improve at-a-glance legibility. Overall, it looks much cleaner!

wseymour15:
> I agree with @mister-ben that a small indent to the left of the labeled settings (color, opacity) would also improve at-a-glance legibility. Overall, it looks much cleaner!

@roman-bc-dev This was more work then I expected 😅 One of these label/select elements was wrapped in a span, the other was not.. so I updated the HTML to ensure that both are wrapped in a span element, updated the css, and also a few test fixes :)

roman-bc-dev (approved):

PR #8102, opened 2023-02-01
fix(lang): Add missing comma in turkish

mister-ben:


gkatsev (approved):

mister-ben:
Failing test is pushing to codecov, and this change does not affect coverage.

PR #8113, opened 2023-02-08
feat: Add document picture-in-picture support

### LINKED ISSUE #8108: Feature Request: Use Document Picture-in-Picture Web API
Users should be able to use the Video.js Player in a Picture-in-Picture window, not just the HTML video inside the web player. 

image

Before the Document Picture-in-Picture API, it was only possible to put an HTML `<video>` element into a Picture-in-Picture window. This new API makes it possible to open an always-on-top window that can be populated with arbitrary HTML content. It is available as an origin trial starting in Chrome 111 on desktop.

```js
async function togglePictureInPicture() {
  // Close Picture-in-Picture window if any.
  if (documentPictureInPicture.window) {
    documentPictureInPicture.window.close();
    return;
  }

  // Open a Picture-in-Picture window.
  const pipWindow = await documentPictureInPicture.requestWindow({
    initialAspectRatio: 640 / 360,
    copyStyleSheets: true,
  });

  // Move video to the Picture-in-Picture window and make it full page.
  const video = document.querySelector("#video");
  pipWindow.document.body.append(video);
  video.classList.toggle("fullpage", true);

  // Listen for the PiP closing event to move the video back.
  pipWindow.addEventListener("unload", (event) => {
    const videoContainer = document.querySelector("#videoContainer");
    const pipVideo = event.target.querySelector("#video");
    pipVideo.classList.toggle("fullpage", false);
    videoContainer.append(pipVideo);
  });
}

// Replace the built-in PiP button with our own button in the Video.js player.
window.onload = () => {
  const controls = videojs("video", { controls: true }).getChild("ControlBar");
  controls.removeChild("FullscreenToggle");
  controls.removeChild("PictureInPictureToggle");
  controls.addChild("button", {
    controlText: "Toggle Picture-in-Picture",
    className: "vjs-visible-text",
    clickHandler: togglePictureInPicture,
  });
};
```

Documentation:
- https://developer.chrome.com/blog/the-future-of-picture-in-picture/
- https://developer.chrome.com/docs/web-platform/document-picture-in-picture/
- https://github.com/WICG/document-picture-in-picture/issues

mister-ben:
## Description
Implements [Document Picture in Picture API](https://developer.chrome.com/docs/web-platform/document-picture-in-picture/), fixes #8108

Preview at https://deploy-preview-8113--videojs-preview.netlify.app/sandbox/docpip.html

## Specific Changes proposed
Adds a `enableDocumentPictureInPicture` player option, to use the doc PIP API where available.
Inserts an el with the player el's classes as a placeholder. Adds poster and titlebar to the placeholder if present.

## To do / to discuss
- [x] Tests
- [x] Docs - videojs/videojs.com/pull/168
- [x] Aspect ratio not respected by PIP window _Tracked in Chromium https://crbug.com/1407629_
- [x] Events
- [x] Fullscreen button from PIP window - can we trigger fuullscreen from here, otherwise disable
- [x] Responsive layout styles needed to resize plaeholder text on small players?
- [x] Exit PIP button needed in PIP window, as the window UI already has two buttons
- [x] User expectations that `disablePictureInPicture` will also disable this? 
- [x] Translations

This is implemented so that the current API / control triggers either `window.documentPictureInPicture` (docPiP) or `video.requestPictureInPicure` (elPiP). The same events will fire. Promises will be returned when entering and exiting.

- docPiP is opt-in with the `enableDocumentPictureInPicture` option.
- elPiP remains opt-out with the `disablePictureInPicture` option.
- if docPiP is enabled and available, and elPiP is not disabled but available, docPiP is used.

The assumption is an integrator will generally prefer docPiP, so controls, ads, etc can be used. Therefore `disablePictureInPicture` **does not** disable docPiP, even if that may initially seem unintuitive.


src/js/player.js
```diff
@@ -3031,6 +3031,42 @@ class Player extends Component {
...
+      if (this.posterImage) {
+        pipContainer.appendChild(this.posterImage.el().cloneNode(true));
+      }
+      if (this.titleBar) {
+        pipContainer.appendChild(this.titleBar.el().cloneNode(true));
+      }
+      pipContainer.appendChild(Dom.createEl('p', { className: 'vjs-pip-text' }, {}, this.localize('Playing in picture-in-picture')));
+
+      window.documentPictureInPicture.requestWindow({
+        // This aspect ratio doesn't seem to be respected
```
beaufortfrancois, line 3048: Do you see this on macOS?

mister-ben: Yes at https://deploy-preview-8113--videojs-preview.netlify.app/sandbox/docpip.html on Canary 112

In the original page `videoWidth / videoHeight` is 2.4. in the PiP window `innerWidth / innerHeight` is 2.6666666666666665.

A 16/9 video's window is also too wide. I suppsose the aspect ratio is for the outer size of the window, including its title bar?

beaufortfrancois: I think you're hitting https://bugs.chromium.org/p/chromium/issues/detail?id=1407629 then. I hope it will be fixed soon.

FYI @liberato-at-chromium

mister-ben: The PiP window at https://document-picture-in-picture-api.glitch.me/ is also too wide. It would be more useful to be able to set the inner size / aspect ratio. The title bar height will differ by OS, font settings, locale, and eventually browser.

beaufortfrancois: That is good feedback! I'll add it to the Chromium bug.

mister-ben: Thanks. I was going to add an issue at WICG/document-picture-in-picture but won't then if it's tracked elsewhere.

beaufortfrancois: Filing an issue to the spec would help as well actually. Feel free to do so.

beaufortfrancois: Thanks for filing https://github.com/WICG/document-picture-in-picture/issues/25

src/js/player.js
```diff
@@ -3031,6 +3031,42 @@ class Player extends Component {
    *         A promise with a Picture-in-Picture window.
```
beaufortfrancois, line 3031: Comments may need to be updated as well there. Does the `Player#enterpictureinpicture` fire?

mister-ben: 👍  It does now

src/js/player.js
```diff
@@ -3023,14 +3023,62 @@ class Player extends Component {
...
    *
    * @return {Promise}
    *         A promise with a Picture-in-Picture window.
    */
   requestPictureInPicture() {
+    if (this.options_.enableDocumentPictureInPicture && 'documentPictureInPicture' in window) {
+      const pipContainer = document.createElement(this.el().tagName);
+
+      pipContainer.classList = this.el().classList;
+      pipContainer.classList.add('vjs-pip-container');
```
misteroneill, line 3045: curious: Does the `pipContainer.classList` get assigned by reference and, so, does this modification update both `this.el()` and `pipContainer`?

src/js/player.js
```diff
@@ -3023,14 +3023,62 @@ class Player extends Component {
...
+          pipWindow.addEventListener('unload', (event) => {
+            const pipVideo = event.target.querySelector('.video-js');
+
+            pipContainer.replaceWith(pipVideo);
+            this.player_.isInPictureInPicture(false);
+            this.player_.trigger('leavepictureinpicture');
+          });
+
+          resolve(pipWindow);
+        }).catch(reject);
```
misteroneill, line 3079: question: Rather than `resolve` and `reject` here, would it make sense to return the promise returned by `window.documentPictureInPicture.requestWindow` above? I guess it's not clear to me what the purpose is of this outer promise.

src/js/player.js
mister-ben: `this.el().classlist` is not modifed

src/js/player.js
mister-ben: Ah yes I should return the window here instead.

src/js/player.js
```diff
@@ -3053,7 +3099,12 @@ class Player extends Component {
    *         A promise.
    */
   exitPictureInPicture() {
-    if ('pictureInPictureEnabled' in document) {
+    if (window.documentPictureInPicture && window.documentPictureInPicture.window) {
+      return new Promise((resolve) => {
+        window.documentPictureInPicture.window.close();
+        resolve();
+      });
+    } else if ('pictureInPictureEnabled' in document) {
```
beaufortfrancois, line 3107:
```suggestion
      window.documentPictureInPicture.window.close();
      return Promise.resolve();
    } 
    if ('pictureInPictureEnabled' in document) {

```

src/js/control-bar/picture-in-picture-toggle.js
```diff
@@ -63,11 +64,19 @@ class PictureInPictureToggle extends Button {
...
 
   /**
-   * Enables or disables button based on document.pictureInPictureEnabled property value
-   * or on value returned by player.disablePictureInPicture() method.
+   * Enables or disables button based on availablity of a Picture-In-Picture mode.
+   *
+   * Enabled if
+   * - `player.options().enableDocumentPictureInPicture` is true and
+   *   window.documentPictureInPicture is available; or
+   * - `player.disablePictureInPicture()` is not `true and
```
beaufortfrancois, line 72:
```suggestion
   * - `player.disablePictureInPicture()` is not true and
```

src/js/control-bar/picture-in-picture-toggle.js
```diff
@@ -63,11 +64,19 @@ class PictureInPictureToggle extends Button {
   }
 
   /**
-   * Enables or disables button based on document.pictureInPictureEnabled property value
-   * or on value returned by player.disablePictureInPicture() method.
+   * Enables or disables button based on availablity of a Picture-In-Picture mode.
```
beaufortfrancois, line 67:
```suggestion
   * Enables or disables button based on availability of a Picture-In-Picture mode.
```

src/js/control-bar/picture-in-picture-toggle.js
beaufortfrancois: or better 
```suggestion
   * - `player.disablePictureInPicture()` is false and
```

src/js/player.js
```diff
@@ -3037,14 +3037,59 @@ class Player extends Component {
    * continue consuming media while they interact with other content sites, or
    * applications on their device.
    *
+   * This can use document picture-in-picture or element picture in picture
+   *
+   * Set `enableDocumentPictureInPicture` to `true` to use docPiP on a supported browser
+   * Else set `disablePictureInPicture` to `false` to disable elPiP on a supported browser
+   *
+   *
    * @see [Spec]{@link https://wicg.github.io/picture-in-picture}
```
beaufortfrancois, line 3046:
```suggestion
   * @see [Spec]{@link https://w3c.github.io/picture-in-picture/}
```

src/js/player.js
```diff
@@ -3037,14 +3037,59 @@ class Player extends Component {
...
+   * @see [Spec]{@link https://wicg.github.io/document-picture-in-picture/}
    *
    * @fires Player#enterpictureinpicture
    *
    * @return {Promise}
    *         A promise with a Picture-in-Picture window.
    */
   requestPictureInPicture() {
+    if (this.options_.enableDocumentPictureInPicture && 'documentPictureInPicture' in window) {
```
beaufortfrancois, line 3055: Nit: Be consistent and check `window.documentPictureInPicture` everywhere

src/js/player.js
```diff
@@ -3067,7 +3113,12 @@ class Player extends Component {
    *         A promise.
    */
   exitPictureInPicture() {
+    if (window.documentPictureInPicture && window.documentPictureInPicture.window) {
+      window.documentPictureInPicture.window.close();
```
beaufortfrancois, line 3117: You may want to add a comment saying the `"unload"` event handler takes care of firing the `Player#leavepictureinpicture` event.

mister-ben: Thanks for the feedback, added.

beaufortfrancois: Happy to help!

beaufortfrancois:
It seems like all TODOs have been checked. Is there anything left before merging this PR?

src/css/components/menu/_menu-popup.scss
```diff
@@ -9,6 +9,11 @@
   border-top-color: rgba($primary-background-color, $primary-background-transparency); // Same as ul background
 }
 
+.vjs-pip-window .vjs-menu-button-popup .vjs-menu {
+  left: unset;
+  right: 1em; // Extra offset for last menu button in pip window, as fullscreen button not present
+}
```
mister-ben, line 15: Consider this temporary. This needs a better solution longer term that only modifies the last _visible_ menu, but this situation could apply in situations other than PiP so deserves tackling separately to thi PR.

PR #8114, opened 2023-02-08
fix: add padding to title bar text for icon space

usmanonazim:
## Description
With the deprecation of the [videojs-dock plugin](https://github.com/brightcove/videojs-dock) and the creation of the `TitleBar` to replace it, the [`shelf`](https://github.com/brightcove/videojs-dock/blob/8309e7f28bdced765ca0421b0326967e0c870172/src/plugin.js#L61) component in the dock plugin was removed. The `shelf` component took up 25% of the dock and ensured that icons added to the dock would not overlap with the title/description text. This change is adding the padding that the `shelf` used to have so that any icons in the TitleBar do not overlap with the title/description text.

## Specific Changes proposed
- CSS to add padding to the right of the title and description text to ensure no overlap with any icons added to the `TitleBar`

### Example of the TitleBar before the proposed change
image

### Example after the proposed change
image

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_title-bar.scss
+++ b/src/css/components/_title-bar.scss
@@ -24,10 +24,17 @@
 
 .vjs-title-bar-title,
 .vjs-title-bar-description {
+  // The width of the 'shelf' component that existed in the deprecated 'videojs-dock' plugin.
+  // Adding this amount of padding to the right of the title bar title/descriptions allows for
+  // icons to be placed in the top right of the title bar without overlapping on the title/description
+  // text.
+  $shelf-width: 25%;
+
   margin: 0;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


misteroneill (commented):
I'm good with this in principle, but I think it should probably be either moved into the Brightcove Player as it's specifically related to our use case _or_ associated with an option of some kind for the `TitleBar` component.

usmanonazim:
> I'm good with this in principle (the code change is fine), but I think it should be either moved into the Brightcove Player as it's specifically related to our use case _or_ associated with an option of some kind for the `TitleBar` component.

Good point - going to add this css to the relevant plugin and close this PR

PR #8129, opened 2023-02-13
Update progress bar slider translation

sophiaonion:
## Description
Our app uses videoJS and there's an a11y report on the progress bar aria text is not specific enough. The string `progress bar timing: currentTime={1} duration={2}` should provide enough specificity. However, looks like videoJS is falling back to use the default translation string because we didn't add the `progress bar timing: currentTime={1} duration={2}` to the translation file 

## Specific Changes proposed
- update translation in english

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -14,7 +14,7 @@
   "Loaded": "Loaded",
   "Progress": "Progress",
   "Progress Bar": "Progress Bar",
-  "progress bar timing: currentTime={1} duration={2}": "{1} of {2}",
+  "progress bar timing: currentTime={1} duration={2}": "progress bar timing: currentTime={1} duration={2}",
   "Fullscreen": "Fullscreen",
   "Exit Fullscreen": "Exit Fullscreen",
   "Mute": "Mute",


mister-ben (dismissed):
It's not intended that you include en.json, as English is the default. However even as a template for translations, which is the purpose of this file, it does make sense to have the resultant string as the value. Thanks.

gkatsev:
> it does make sense to have the resultant string as the value

I don't think this will work well since the left hand side isn't intended to be exactly what's in the labels for ARIA. Plus, we already have enough trouble with folks translating the left hand side instead of the right hand side.

mister-ben:
> > it does make sense to have the resultant string as the value
> 
> I don't think this will work well since the left hand side isn't intended to be exactly what's in the labels for ARIA. Plus, we already have enough trouble with folks translating the left hand side instead of the right hand side.

Ah, I'd misread what the change was. It's correct as it is.

wseymour15:
Hi @sophiaonion, this appears to be something we will not change in video.js, however. if you would like to update the translation for your own player, you can do it by using the [Adding languages via the API](https://videojs.com/guides/languages/#adding-languages-via-the-api).

Please feel tree to close this review if this resolves the issue.

PR #8130, opened 2023-02-13
fix: Remove img el when there's no poster source

### LINKED ISSUE #8122: [v8] Empty `src` attribute of the `img` element when no poster is added
### Description

[The poster is set to an empty string](https://github.com/videojs/video.js/blob/v8.0.4/src/js/player.js#L433) if no poster is initially configured. Which causes that empty string to be set for `src` of the `img` element of the poster. Which is invalid HTML according to HTML-validators.

### Reduced test case

_No response_

### Steps to reproduce

Check the description for an explanation.

### Errors

_No response_

8.0.4

### Video.js plugins used.

_No response_

110

Linux

mister-ben:
## Description
If there is no poster source, an img el can't be present as it would be invalid.

## Specific Changes proposed
Use a div (again) as the top level element of the PosterImage, to keep position in the DOM. Insert/remove the picture/img els when a poster source is set/unset. 
Fixes #8122 

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/poster-image.js
+++ b/src/js/poster-image.js
@@ -101,43 +101,49 @@ class PosterImage extends ClickableComponent {
   update(event) {
     const url = this.player().poster();
 
+    this.setSrc(url);
+
+    // If there's no poster source we should display:none on this component
+    // so it's not still clickable or right-clickable
     if (url) {
-      // As there's a poster source, add a picture/img if not already present
-
-      if (!this.$('img')) {
-        this.el_.appendChild(Dom.createEl(
-          'picture', {
-            className: 'vjs-poster',
-
-            // Don't want poster to be tabbable.
-            tabIndex: -1
-          },
-          {},
-          Dom.createEl('img', {
-            loading: 'lazy',
-            crossOrigin: this.crossOrigin()
-          }, {
-            alt: ''
-          })
-        ));
-      }
-      this.setSrc(url);
       this.show();
     } else {
-      // With no source, remove child elements to not end up with an invalid img
-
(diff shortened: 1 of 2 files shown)


src/js/poster-image.js
```diff
@@ -131,7 +138,7 @@ class PosterImage extends ClickableComponent {
    *        The URL to the source for the `PosterImage`.
    */
   setSrc(url) {
-    this.el_.querySelector('img').src = url;
+    this.$('img').src = url;
```
gkatsev, line 141: this will throw if someone tries to set a src and the img element isn't present.

src/js/poster-image.js
```diff
@@ -113,13 +101,32 @@ class PosterImage extends ClickableComponent {
...
     const url = this.player().poster();
 
-    this.setSrc(url);
-
-    // If there's no poster source we should display:none on this component
-    // so it's not still clickable or right-clickable
     if (url) {
+      // As there's a poster source, add a picture/img if not already present
+
+      if (!this.$('img')) {
```
gkatsev, line 107: this probably should move into `setSrc` so that either update or setSrc can be called without any errors. And it'll keep this method the same.

src/js/poster-image.js
mister-ben: I was thinking a poster is only set with `player.poster('...')`, but you're right of course; `setSrc()` isn't "private" so does need to work independently of `update()`.

gkatsev (approved):

PR #8142, opened 2023-02-20
fix: replay button broken for native playback

adrums86:
## Description
During native playback in Safari (desktop or iOS) if the user reaches the end of content in fullscreen, the replay button will no longer function with a single click. Meaning, the button will work and change state to `playing` however the video does not restart. It seems to be an issue with the content not ending correctly as the `ended` event is firing, however `video.ended` is returning `false`. Since it's native, we don't have the option of calling `endOfStream` on the mediaSource.

## Specific Changes proposed
If we detect native playback from the `tech` and the play toggle is still in the replay state, we can manually do a seek to 0 and call play.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -58,12 +58,6 @@ class PlayToggle extends Button {
    */
   handleClick(event) {
     if (this.player_.paused()) {
-      // Safari will not restart playback on a play click when playing back natively, so we seek to 0, then play.
-      const endedNativePlayback = this.hasClass('vjs-ended') && !this.player_.lastSource_.tech.startsWith('blob');
-
-      if (endedNativePlayback) {
-        this.player_.currentTime(0);
-      }
       silencePromise(this.player_.play());
     } else {
       this.player_.pause();
--- a/src/js/control-bar/play-toggle.js
+++ b/src/js/control-bar/play-toggle.js
@@ -132,9 +132,7 @@ class PlayToggle extends Button {
     this.addClass('vjs-ended');
     // change the button text to "Replay"
     this.controlText('Replay');
-    // console.dir(event);
-    // console.dir(this.player_.tech_);
-    // is not getting fired in Safari
+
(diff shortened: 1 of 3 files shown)


Essk (approved):
Looks good!

gkatsev:
This seems like it probably should happen in [`player.play()`](https://github.com/videojs/video.js/blob/main/src/js/player.js#L2312) since I'd expect the programmatic method would also fail in the same way.

gkatsev:
Also, I'm really sad that it's needed. I think Video.js used to have it a long time ago but was removed because we thought it wasn't needed anymore.

src/js/control-bar/play-toggle.js
```diff
@@ -58,6 +58,12 @@ class PlayToggle extends Button {
    */
   handleClick(event) {
     if (this.player_.paused()) {
+      // Safari will not restart playback on a play click when playing back natively, so we seek to 0, then play.
+      const endedNativePlayback = this.hasClass('vjs-ended') && !this.player_.lastSource_.tech.startsWith('blob');
```
mister-ben, line 62: I don't think `lastSource_` will always be present. With the YouTube tech for example.

adrums86:
@gkatsev It actually works like this, Safari doesn't seem to have a problem with the play call, tested on mac and iOS. That said, I agree it would fit better in the actual play function. This will need some refactoring anyway due to what @mister-ben mentioned. 
I agree it is sad to see, luckily it seems isolated to native _fullscreen only_ (ending in fullscreen, returning from fullscreen, hitting replay).

src/js/control-bar/play-toggle.js
adrums86: Moved everything to `player.js` per @gkatsev's suggestion and used the `browser` constants to check for native playback rather than the `lastSource_`

adrums86:
I moved everything over to `player.js` and added a unit test.

alex-barstow (approved):

Issue #8143, opened 2023-02-20
Time display for player causes prevention role conflict accessibility issue


bwood614:
### Description

The time display elements in the player have an accessibility issue. The <span> elements within the various time display elements (e.g. spans with classes like `vjs-current-time-display`, `vjs-duration-display`, `vjs-remaining-time-display`, etc.) have the attributes `role="presentation"` and `aria-live="off"`. It looks like `role="presentation"` was added some time after the aria-live attribute and the aria-live attribute was never removed.
 
Having global aria attributes and `role="presentation"` in the same element is redundant and can lead to some unexpected behaviors across browsers. See [https://dequeuniversity.com/rules/axe/4.1/presentation-role-conflict](https://dequeuniversity.com/rules/axe/4.1/presentation-role-conflict) for more information about this issue.

### Steps to reproduce

1. Inspect the time display element(s) in the player with browser dev tools

### What version of Video.js are you using?

8.0.4

### Video.js plugins used.

None

### What browser(s) including version(s) does this occur with?

Chrome 109

### What OS(es) and version(s) does this occur with?

Mac OS Monterey v12.4

brayden-wood:
I am picking up this issue and submitting a PR for it

PR #8144, opened 2023-02-20
fix: improved accessibility for time display

bwood614:
This change is to remove the aria-live attribute from time display elements. aria-live is no longer needed since the presentation role was added. Both attributes being present can lead to unexpected screen reader behavior.

## Description
Bug issue: https://github.com/videojs/video.js/issues/8143

## Specific Changes proposed

- Remove aria-live="off" from all time display elements within the player since they already have role="presentation"

## Requirements Checklist
  - [ ] Docs/guides updated


Yie3:
# @

PR #8147, opened 2023-02-21
feat: add skip forward/backward buttons

usmanonazim:
## Description
With the upgrade to videojs font 4.0.0, font icons for skipping forward and backward have been introduced. This PR adds the option to add a skip forward or skip backward button to the control bar. These buttons are placed after the `PlayToggle` and before the `VolumeControl`, as shown in the image below. 

image

The user also has the option to enable only one of the two skip buttons, as shown in the image below.

image

Code pen example with all icons: https://codepen.io/usmanomar/pen/MWqjWjj
Relevant docs: https://github.com/videojs/videojs.com/pull/167

### Options Description

`controlBar.skipButtons.forward`
> Type: `number`

The number of seconds forward that should be skipped on button click (can be 5, 10, or 30)

`controlBar.skipButtons.backward`
> Type: `number`

The number of seconds backward that should be skipped on button click (can be 5, 10 or 30)

## Specific Changes proposed
- Implement a `SkipForward` button that skips forward through a video a configurable amount of seconds
- Implement a `SkipBackward` button skips backward through a video a configurable amount of seconds
- Use the `vjs-icon-forward-x` and `vjs-icon-replay-x` for the `SkipForward` and `SkipBackward` buttons respectively

## Requirements Checklist
  - [x] Docs/guides updated
  - [x] Example created (`sandbox/skip-buttons.html.example`)


mister-ben:
The numbers on the icons are pretty illegible at the default size of the controls. That may well be flagged in an acessiblity audit.

src/js/control-bar/skip-buttons/skip-forward.js
```diff
@@ -0,0 +1,67 @@
...
+   *        The `click` event that caused this function
+   *        to be called
+   */
+  handleClick(event) {
+    const currentVideoTime = this.player_.currentTime();
+    const duration = this.player_.duration();
+
+    let newTime;
+
+    if (currentVideoTime + this.skipTime <= duration) {
```
mister-ben, line 53: Live videos will have a duration of infinity. We shouldn't try to seek beyond the seekable range.

usmanonazim: Ahh I see, will make that change to use seekable range instead of duration and update the corresponding tests :)

usmanonazim:
> The numbers on the icons are pretty illegible at the default size of the controls. That may well be flagged in an acessiblity audit.

Do you know if there's a good way to fix that?

mister-ben:
> > The numbers on the icons are pretty illegible at the default size of the controls. That may well be flagged in an acessiblity audit.
> 
> Do you know if there's a good way to fix that?

I think it needs to be different icons that are readble at that size, or plain arrows. The control text could (and for accessibility probably should) state the increment to be skipped by.

roman-bc-dev:
> > The numbers on the icons are pretty illegible at the default size of the controls. That may well be flagged in an acessiblity audit.
> 
> Do you know if there's a good way to fix that?

The icons used here are from the standard Material Design set. Unless we jerry-rig a solution with existing videojs font icons, we'd need to decide on brand new ones and then update the font repo with them to then be used here, which seems outside the scope of the changes described.

@mister-ben The empty round arrow icon is already used for replay UI and plain arrows are not descriptive of the intended action here. Are there any other icons from the Material Design or Font Awesome collections that you would recommend for legibility?

src/js/control-bar/skip-buttons/skip-forward.js
usmanonazim: I've now added logic to check for seekable range if it is a live video (for both skip forward and skip backward buttons), and updated tests

mister-ben:
I've seen the rewind / fast forward icons used for this in OTT apps, but they are also problematic as they can be understood as changing the playback rate.

Some also use a similar looped arrow with a double arrowhead, but there's no such icon in Material Icons or Font Awesome. Netflix is like this. (They also show numbers, but have the icons large enough for this to work.)

[Here](https://github.com/mister-ben/videojs-seek-buttons/blob/master/src/plugin.css#L10-L11) I used rotated arrows to distinguish them from replay.

Maybe showing the number, larger, when the button has focus could work? If we really want numbers they need to be readable, I might try tweaking the SVG icons.

It might be easier if we could consider using "restart alt" from Material icons for replay with a fuller, broken loop.
grafik

mister-ben:
[Here's one idea,](https://codepen.io/mister-ben/pen/gOdMRVP?editors=1100) modifying Material Icons's SVG

grafik

usmanonazim:
> [Here's one idea,](https://codepen.io/mister-ben/pen/gOdMRVP?editors=1100) modifying Material Icons's SVG
> 
> grafik

That looks great - thank you for the suggestion! Is it ok if I use that?

roman-bc-dev:
> [Here's one idea,](https://codepen.io/mister-ben/pen/gOdMRVP?editors=1100) modifying Material Icons's SVG
> 
> grafik

_Love_ that icon idea.

@usmanonazim Are you comfy with modifying SVGs? You have the tools to do so?

mister-ben:
Thinking about it, the icon could just be that partial arrow, and the number can be text. That's a lot less SVG work, we don't have to impose the increments that can be used, and the number can be localised for languages that don't use "arabic numerals", like actual arabic numbers.

usmanonazim:
> Thinking about it, the icon could just be that partial arrow, and the number can be text. That's a lot less SVG work, we don't have to impose the increments that can be used, and the number can be localised for languages that don't use "arabic numerals", like actual arabic numbers.

Just to clarify, do you mean that the icon used would just be the partial arrow, and then the control text would specify the number, or that the number would go underneath the arrow icon? If it's the second one, setting a number that's too large (ie 1000) could ruin the look of the icon right?

mister-ben:
The control text should be something useful for screenreaders, like "skip back 10 seconds".

I've updated the [codepen example](https://codepen.io/mister-ben/pen/gOdMRVP?editors=1100) with a version that adds a span for the number, with aria-hidden so it's not read out, and can be localised to match the player language.

grafik

roman-bc-dev:
You'd also have to consider dynamic typeface sizing/interactions within an SVG icon, which gets finicky, especially with hover effects, etc. For the purposes of the intended work here, it may be best to implement the expected icons with the 5, 10, and 30 increments as SVGs and then build a more expanded solution later down the line.

usmanonazim:
> You'd also have to consider dynamic typeface sizing/interactions within an SVG icon, which gets finicky, especially with hover effects, etc. For the purposes of the intended work here, it may be best to implement the expected icons with the 5, 10, and 30 increments as SVGs and then build a more expanded solution later down the line.

I agree - I've created a [codepen example](https://codepen.io/usmanomar/pen/MWqjWjj) with the new icons. @mister-ben @roman-bc-dev please let me know what you think of them 😄

mister-ben:
Nice. Are we sticking with SVG or getting these into the font? (In general I'd prefer to move to SVG directly).

usmanonazim:
> Nice. Are we sticking with SVG or getting these into the font? (In general I'd prefer to move to SVG directly).

I was planning to replace the existing icons for replay-x and forward-x in [font](https://github.com/videojs/font) with the SVGs that I've created.

usmanonazim:
I've now created a PR in videojs/font to update the existing icons: https://github.com/videojs/font/pull/49

src/js/control-bar/skip-buttons/skip-backward.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+class SkipBackward extends Button {
+  constructor(player, options) {
+    super(player, options);
+
+    this.validOptions = [5, 10, 30];
+    this.skipTime = this.getSkipBackwardTime();
+
+    if (this.skipTime && this.validOptions.includes(this.skipTime)) {
+      this.controlText(`Skip backward ${this.skipTime} seconds`);
```
mister-ben, line 20: This way means the translation file need a translation for each allowed increment. It's relatively manageable with 3 permitted values, but it is [possible to use a variable](https://github.com/videojs/video.js/blob/e1a7ac376a2b0cb69f92bff45f6ad2b0fbc1a55d/src/js/loading-spinner.js#L25) in these translations: `this.localize('Skip backward {seconds} seconds', [this.skipTime])`

Either way there should be at least an entry for each string in en.json, so there's a template to translate from.

mister-ben (changes requested):
The deploy preview is looking good! Just the note about translations. At the least there should be an en.json. If swithcing to variables is complicated we can tackle that later (and there's an issue with variables and languagechange event that is out of scope for this PR). 

Could you also add a link to the sandbox page in /index.html?

src/js/control-bar/skip-buttons/skip-backward.js
misteroneill: I agree.

It should be safe to pass a pre-localized string to `controlText`, like:
```js
this.controlText(this.localize('Skip backward {seconds} seconds', [this.skipTime]));
```

src/js/control-bar/skip-buttons/skip-forward.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+class SkipForward extends Button {
+  constructor(player, options) {
+    super(player, options);
+
+    this.validOptions = [5, 10, 30];
+    this.skipTime = this.getSkipForwardTime();
+
+    if (this.skipTime && this.validOptions.includes(this.skipTime)) {
+      this.controlText(`Skip forward ${this.skipTime} seconds`);
```
misteroneill, line 20: Same comment here about localization.

misteroneill (changes requested):
Looks great, Usman! I think just addressing the localization thing should be all that's needed.

src/js/control-bar/skip-buttons/skip-forward.js
usmanonazim: I've now moved the control text strings from `skip-forward` and `skip-backward` to `en.json` for localization.

usmanonazim:
> The deploy preview is looking good! Just the note about translations. At the least there should be an en.json. If swithcing to variables is complicated we can tackle that later (and there's an issue with variables and languagechange event that is out of scope for this PR).
> 
> Could you also add a link to the sandbox page in /index.html?

I've added the link to the sandbox page now 😁

misteroneill (approved):
LGTM, thanks Usman!

mister-ben (approved):

Issue #8141, opened 2023-02-18
Unable to install via NPM in Angular application


LexanRed:
### Description

When installing vidoe.js via NPM and trying to include it in an Angular 15 application I get all kind of strange errors. 

I assume the guide on https://videojs.com/guides/angular/ is outdated. 

Also when trying to import the player like this: import Player from "video.js/dist/types/player"; 
The errors persist. 

I also tried installing the video.js types. Unfortunately there are no types for v8 yet. 

In anyway, it is not possible to install it via NPM into an Angular 15 application. 

`Error: node_modules/video.js/dist/types/component.d.ts:68:25 - error TS2304: Cannot find name 'Player'.

68     constructor(player: Player, options?: {
                           ~~~~~~


Error: node_modules/video.js/dist/types/component.d.ts:71:8 - error TS1016: A required parameter cannot follow an optional parameter.

71     }, ready: any);
          ~~~~~


Error: node_modules/video.js/dist/types/component.d.ts:117:15 - error TS2304: Cannot find name 'Player'.

117     player(): Player;
                  ~~~~~~

Error: node_modules/video.js/dist/types/event-target.d.ts:9:13 - error TS2395: Individual declarations in merged declaration 'EventTarget' must be all exported or all local.

9 export type EventTarget = () => any;
              ~~~~~~~~~~~


Error: node_modules/video.js/dist/types/event-target.d.ts:18:15 - error TS2395: Individual declarations in merged declaration 'EventTarget' must be all exported or all local.

18 declare class EventTarget {
                 ~~~~~~~~~~~

Error: node_modules/video.js/dist/types/modal-dialog.d.ts:48:25 - error TS2304: Cannot find name 'Player'.

48     constructor(player: Player, options?: {
                           ~~~~~~


Error: node_modules/video.js/dist/types/modal-dialog.d.ts:49:19 - error TS2304: Cannot find name 'Mixed'.

49         content?: Mixed;
                     ~~~~~


Error: node_modules/video.js/dist/types/modal-dialog.d.ts:141:24 - error TS2304: Cannot find name 'Mixed'.

141     fillWith(content?: Mixed): void;
                           ~~~~~


Error: node_modules/video.js/dist/types/modal-dialog.d.ts:164:21 - error TS2304: Cannot find name 'Mixed'.

164     content(value?: Mixed): Mixed;
                        ~~~~~


[...]

Issue #8109, opened 2023-02-05
TypeScript: Property 'on' does not exist on type 'Player'


jdufresne:
### Description

The TypeScript definitions are missing the methods added by the EventedMixin on the Player class.

### Reduced test case

https://github.com/jdufresne/video-js-test

### Steps to reproduce

Files: `test.ts`

```ts
import videojs from "video.js";

const videoEl = document.createElement("video");
const player = videojs(videoEl, {
  sources: [{ src: "https://vjs.zencdn.net/v/oceans.mp4", type: "video/mp4" }],
});
player.on("ended", () => {
  console.log("video ended");
});
```

Compiling this TypeScript results in the following error:

```
$ npx tsc
test.ts:7:8 - error TS2339: Property 'on' does not exist on type 'Player'.

7 player.on("ended", () => {
         ~~
```

### Errors

```
test.ts:7:8 - error TS2339: Property 'on' does not exist on type 'Player'.
```

### What version of Video.js are you using?

8.0.3

### Video.js plugins used.

none

### What browser(s) including version(s) does this occur with?

n/a

### What OS(es) and version(s) does this occur with?

n/a

boris-petrov:
These are the things that I found missing from the v8 types compared to the v7 ones:

On the `Player` type: `controlBar`, `readyState`, `bigPlayButton`, `remoteTextTracks`, `audioTracks`, `on`, `one`, `off`, `trigger`

`player.ready` should set the `this` of the function in the first argument to be the `player` instance.

On `Tech`: `on`, `off`

`AudioTrackList` should extend `AudioTrack[]`.

`TextTrackList` should extend `TextTrack[]`.

insilications:
The v8 types/typing is very incomplete too.

Quimy:
In my case, there was a change in typings from separate package `@types/video.js` to types in videojs itself and it lacks many of the types:

* there is no Player class exposed, previously it was `VideoJsPlayer`
* custom components show errors:
```
const VjsButton = videojs.getComponent('Button');
class FrameControlButton extends VjsButton {
  private frameSkipTime: number;

  constructor(player: VideoJsPlayer, options: { [key: string]: never } = {}) {
    super(player, options);
    this.frameSkipTime = options.frameSkipTime;
  }

  handleClick() {
    this.player().currentTime(this.player().currentTime() + this.frameSkipTime);
  }
}
```
`Type 'Component' is not a constructor function type.`
[...]

PR #8148, opened 2023-02-21
fix(types): Improve Typescript coverage

mister-ben:
Updates to JSDOC, and to the Component constructor to improve the typescript definitions output. Hopefully addresses #8109 and #8141.

- Some fixes to incorrect / incomplete JSDOC
- Typescript doesn't understand JSDOC's `{EventTarget~Event}` syntax and produces broken output when it's encountered (type would end up as `{EventTarget}` but it also unexpectedly turns optional params into required params). These have had to be simplified to unscoped types.
- Imports typedefs into a lot of files, player and tech especially.
- Adds placeholder event methods (`on` etc) to Component. Typescript doesn't support Typescript's `@mixes`/`@mixin` and I've not found another way to get these included in the output other than having the methods present by default. These will be replaced or set to undefined. This approach doesn't seem great. I'm hoping there's a better alternative I'm missing here.

Unresolved
- This does not resolve a couple of typescript errors with tracks (about how classes have been extended) and with `videojs.VERSION`.
- This does not address adding types for `videojs.Vhs`
- Typescript does not believe that a Component retrieved by `getComponent()` can be used as a contructor.

Reference
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -6,6 +6,10 @@ import Component from './component.js';
 import {isPromise, silencePromise} from './utils/promise';
 
 /**
+ *  @typedef {import('../event-target').Event} Event
+ */
+
+/**
  * The initial play button that shows before the video has played. The hiding of the
  * `BigPlayButton` get done via CSS and `Player` states.
  *
@@ -34,7 +38,7 @@ class BigPlayButton extends Button {
    * This gets called when a `BigPlayButton` "clicked". See {@link ClickableComponent}
    * for more detailed information on what a click can be.
    *
-   * @param {EventTarget~Event} event
+   * @param {Event} event
    *        The `keydown`, `tap`, or `click` event that caused this function to be
    *        called.
    *
--- a/src/js/button.js
+++ b/src/js/button.js
@@ -104,7 +104,7 @@ class Button extends ClickableComponent {
    * This gets called when a `Button` has focus and `keydown` is triggered via a key
    * press.
    *
-   * @param {EventTarget~Event} event
+   * @param {Event} event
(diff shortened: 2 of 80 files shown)


gjanblaszczyk (approved):

jdufresne:
I tested the PR locally. It resolves the original TypeScript issue form me that I reported in #8109. :+1: 

Thanks!

gkatsev:
Thanks for confirming @jdufresne!

LeeWhite187:
Is there a viable example of this library (video.js) in Angular 15, yet?
I'm having the same problems as this: [8141](https://github.com/videojs/video.js/issues/8141).
I walked through the issue links, and did not find a resolution.

Thanks in advance.

LeeWhite187:
Yeah. I did that workaround...
But, setting that "skipLibCheck" flag also removes some of the benefits of working in Typescript.

Any idea when the Angular-friendly update will release?

gkatsev:
> But, setting that "skipLibCheck" flag also removes some of the benefits of working in Typescript.

Yup, totally get it. It's not an ideal workaround.
One of the issues is that we thought that using types generation based on our pretty good jsdocs would be sufficient, but it seems it isn't. Especially with this change, most of the type errors go away, but there's still some stuff missing like exporting the types as expected from the main package.

We're considering the next steps for improved types support.

LeeWhite187:
No worries. Thanks for the workaround in the meantime.

lfuelling:
Is there an ETA when this fix will be released?
It seems to be not included in v8.3.0, even if it was merged before that release.

PR #8160, opened 2023-02-26
Fix/reset progress bar

amtins:
## Description

When `player.reset` is called the `load-progress-bar`, `seek-bar` and `current-time-display` components are not correctly reset.

Screenshot from 2023-02-26 20-21-47

## Specific Changes proposed

Adds the missing UI components to the `resetProgressBar_` method so that they are properly reset.

## Requirements Checklist
  - [ ] Docs/guides updated


amtins:
I also tried to use a unit test to add the test coverage, but it seems that the test below does not do what it says it does 😅
https://github.com/videojs/video.js/blob/54830cdd0e347d0d362dde11f2d1cebbb5b7b664/test/unit/reset-ui.test.js#L5-L20

From PR #5684 release in version [v7.5.0](https://github.com/videojs/video.js/releases/tag/v7.5.0)

src/js/player.js
```diff
@@ -3530,6 +3540,14 @@ class Player extends Component {
     if (remainingTimeDisplay) {
       remainingTimeDisplay.updateContent();
     }
+
+    if (seekBar) {
+      seekBar.update();
+    }
+
+    if (seekBar && seekBar.loadProgressBar) {
```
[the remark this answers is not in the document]
amtins: I just pushed the modification

amtins:
> Definitely a good catch! It would be worth either rewriting this unit test or writing new tests for all the functionality you've added

@usmanonazim Thank you for taking the time to review this PR. I will try to add the test coverage this weekend.

test/unit/reset-ui.test.js
```diff
@@ -1,21 +1,81 @@
...
   player.resetProgressBar_();
-  assert.equal(
-    player.controlBar.progressControl.seekBar.playProgressBar.el().offsetWidth, 0,
-    'progress bar is reset to width 0%'
-  );
-  assert.equal(
-    player.currentTime(), 0,
-    'player current time is 0'
-  );
+  player.duration(0);
```
usmanonazim, line 55: question: is there a reason that duration is set to 0 here? 

I think it would make more sense if the player had a duration of more than 0  so that the tests show `resetProgressBar_` works with a duration of more than 0.

usmanonazim (commented):
Looks good! Just had a question about the test 😁

test/unit/reset-ui.test.js
amtins: @usmanonazim thank you for your feedback 👍🏽 

> question: is there a reason that duration is set to 0 here?

Yes, the main reason is that we test directly the function `resetProgressBar_`. The time reset is done by the function `reset`, that will eventually call [`Html5.resetMediaElement`](https://github.com/videojs/video.js/blob/2a99a7824ea175836249a4044c1d530cce174e08/src/js/tech/html5.js#L1356-L1382) which will cause a reset of `durationDisplay`. This means that in this test case the time must be reset manually otherwise the duration will be `40` see:

https://github.com/amtins/video.js/blob/9d58ef7d8b9bb445b38d6f626b7c613d602a730e/test/unit/reset-ui.test.js#L16

> I think it would make more sense if the player had a duration of more than 0 so that the tests show resetProgressBar_ works with a duration of more than 0.

I'm not quite sure I understand. The `resetProgressBar_` function calling the `durationDisplay.updateContent` if the duration value is not zero then the component will not be reset. 
https://github.com/videojs/video.js/blob/2a99a7824ea175836249a4044c1d530cce174e08/src/js/control-bar/time-controls/duration-display.js#L69-L74

src/js/player.js
```diff
@@ -3521,7 +3521,17 @@ class Player extends Component {
   resetProgressBar_() {
     this.currentTime(0);
```
amtins, line 3522: Do we really need to do this since the call to `el.load()` will reset everything ?
https://github.com/videojs/video.js/blob/2a99a7824ea175836249a4044c1d530cce174e08/src/js/tech/html5.js#L1356-L1382

test/unit/reset-ui.test.js
usmanonazim: Ahh I see thank you for the explanation! Ignore the second part of my comment I was getting a little confused with the way the reset worked.

src/js/player.js
usmanonazim: It might not be necessary to call `this.currentTime(0)` but I'm not 100% sure about that so I'd be happy still keeping that in the `resetProgressBar_` fn

usmanonazim (approved):
Looks good to me! 🥳

amtins:
@usmanonazim thanks again for your review. 👍🏽

PR #8164, opened 2023-02-27
fix: use next-7 instead of next for 7x during npm publish

dzianis-dashkevich:
change publish tag to next-7 for 7x


gkatsev (approved):
Should use `chore:` rather than `fix`, since this is build-related stuff

PR #8174, opened 2023-03-06
fix(lang): Add strings for skip buttons

mister-ben:
## Description
Adds translations for skip buttons, from [mister-ben/videojs-seek-buttons](https://github.com/mister-ben/videojs-seek-buttons)


usmanonazim (commented):
Is there a specific tool we rely on to get the translations? or are there translators that help and give us the translations?

mister-ben:
We just rely on the contributions we get organically from Video.js users willing to contribute currently. We do need to look into this proposal from Wikipmedia #8004

usmanonazim (approved):
LGTM - thank you 😁

PR #8182, opened 2023-03-08
fix: improved accessibility for time display

brayden-wood:
This change is to remove the aria-live attribute from time display elements. The aria-live attribute is no longer needed since the presentation role was added. Both attributes being present can lead to unexpected screen reader behavior. Fixes issue #8143

## Description
This PR fixes the accessibility bug described in https://github.com/videojs/video.js/issues/8143

## Specific Changes proposed
Remove aira-live attribute from time display elements in the player. This attribute was originally added to prevent screen readers from announcing every time change, but later, role="presentation" was added. The presentation role will hide the element from the screen reader all together, so aria-live is no longer needed.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -57,13 +57,11 @@ class TimeDisplay extends Component {
     this.contentEl_ = Dom.createEl('span', {
       className: `${className}-display`
     }, {
-      // tell screen readers not to automatically read the time as it changes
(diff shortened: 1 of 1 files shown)


gkatsev (approved):
Makes sense to me. @OwenEdwards did you have anything to add on this?

misteroneill (approved):
This checks out, thanks for submitting a PR!

PR #8203, opened 2023-03-21
fix: hide menu items on click after hover

usmanonazim:
## Description
This change fixes an issue where hovering over the menu button and then clicking it would not hide the menu items.

## Specific Changes proposed
- Set button pressed to true on `mouseenter` for a `MenuButton` so that clicking has the expected functionality 
- Add tests around `mouseenter` and `mouseleave` fns for a `MenuButton`

## Requirements Checklist
  - [ ] Docs/guides updated


gkatsev:
This breaks the "lock showing" feature when you press the menu-button itself, and then the menu stays visible even as you no longer hover over it. The menu will show on hover, and then the click will now hide the menu.

misteroneill (changes requested):
Gary's correct, the behavior is by design. Was this reported as a bug somewhere?

usmanonazim:
> Gary's correct, the behavior is by design. Was this reported as a bug somewhere?

Yeah in a BC ticket, but looks it was BC css causing the bug - closing this PR :)

PR #8220, opened 2023-03-30
7.8.x

Codinplus31:
I don't know if you have notice but order video player like YouTube,vlc and more if you click on fullscreen on phone and if the video width is above the the video height in size the orientation should be landscape but instead is portrait then in a case where is the video height is above the video width it should be portrait, I have fix this problem in my video player I think you should do the same as well or add it as a new feature.
This is the code:
  
if(video.requestFullscreen){
    video.requestFullscreen();

if (video.videoHeight < video.videoWidth){
 
    if (screen.orientation.lock && screen.orientation) {
      screen.orientation.lock("landscape")
    }
 
  
}
  
  }

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+<a name="7.8.4"></a>
+## [7.8.4](https://github.com/videojs/video.js/compare/v7.8.3...v7.8.4) (2020-07-08)
+
+### Bug Fixes
+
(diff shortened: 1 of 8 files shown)


misteroneill:
I think this was maybe meant as an issue report?

Codinplus31:
Yes it was

Codinplus31:
I will know how to fix it I just need to understand your code

PR #8224, opened 2023-03-31
doc: add GOVERNANCE.md

misteroneill:
This adds a governance document based on discussions from within Brigthcove and current TSC members.

Reviews should be by TSC members specifically, please.


misteroneill:
Moving to another repo.

PR #8225, opened 2023-04-04
Fix ts imports

mister-ben:
## Description
The typedef import syntax we had to introduce for tsc output isn't compatible with jsdoc itself, see e.g. https://github.com/jsdoc/jsdoc/issues/1645 . This is breaking the `docs:api` task. 

## Specific Changes proposed
Adds a jsdoc plugin to transform the import syntax to be understood by jsdoc.
Fixes a few other errors in jsdoc.

## Requirements Checklist

### DIFF
--- a/.jsdoc.json
+++ b/.jsdoc.json
@@ -21,7 +21,7 @@
           "class": "link-vjs"
         },
         {
-          "title": "v8.0.2 source",
+          "title": "v8.2.1 source",
           "link": "https://github.com/videojs/video.js",
           "class": "link-gh"
         },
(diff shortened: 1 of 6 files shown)


gkatsev (approved):

mister-ben:
Hmm, it's outputting filenames like `module-videojs-videojs.html#.AudioTrack`, `global.html#TextTrackMenuItem`

mister-ben:
Moved all the type imports inline, has resolved the build errors and netlify output.

misteroneill (approved):

PR #8232, opened 2023-04-07
fix: audio description track doesn't have icon on Safari browser

gjanblaszczyk:
On Firefox/Chrome browser the HLS stream that contains an audio track with a description show AD icon in the menu list.
image
However, The same stream on Safari doesn't have the AD icon.

## Description
Safari without VHS (in native mode) creates an audio track with a `kind` attribute value equal to "description". 
Unfortunately, video.js handles track kind attribute value equal to "main-desc" only. 

## Specific Changes proposed
Add support for `kind: "description"`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_audio.scss
+++ b/src/css/components/_audio.scss
@@ -2,13 +2,15 @@
   @extend .vjs-icon-audio;
 }
 
+.video-js .vjs-audio-button + .vjs-menu .vjs-description-menu-item .vjs-menu-item-text .vjs-icon-placeholder,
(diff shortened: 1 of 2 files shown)


Sindhuja365 (approved):

Sindhuja365 (approved):

mister-ben (approved):
Not sure why the mixed content in the legacy docs is flagged for this PR in particular, but it's unrelated to the change.

Issue #8092, opened 2023-01-24
about volume location not at centre  at controlBar: {volumePanel: {inline: false}}, at 8.0.3 in chinese


friendtasy:
### Description

volume location not at centre  at controlBar: {volumePanel: {inline: false}}, at 8.0.3 in chinese
<img width="299" alt="屏幕截图 2023-01-24 152316" src="https://user-images.githubusercontent.com/17196859/214234052-e05b1eb0-dbb8-4f03-ad3d-042a09c718b2.png">
[Uploading test.txt…]()

### Steps to reproduce

1.play video in chinese system and explorer
2.
3.

### What version of Video.js are you using?

8.0.3

### What browser(s) including version(s) does this occur with?

chrome 109 zh-cn

### What OS(es) and version(s) does this occur with?

windows11

friendtasy:
> It probably needs this to be changed to line height as was done for the horizontal bar in #7990
> 
> https://github.com/videojs/video.js/blob/509b3d075750be689ded67ca887d70af578c7804/src/css/components/_volume.scss#L133
> 
> Would you be able to test that?

i don't know how to do this way.you can provide js command change  that,let me try it...use up test.html file

mister-ben:
Manually changing styles on the element in developer tools is probably the easiest way to work it out. I had a try with Chrome Canary in Chinese, but I haven’t found a value that works for both Chinese and non-Chinese browsers.

gkatsev:
> Manually changing styles on the element in developer tools is probably the easiest way to work it out. I had a try with Chrome Canary in Chinese, but I haven’t found a value that works for both Chinese and non-Chinese browsers.

@mister-ben what about targeting css based on locale like we do for the subscaps button?

mister-ben:
No, we can't target it with `:lang()` because it's not the language of the page or part of the DOM that matters; it's the language the browser is runnning in. I can replicate only by [running Chrome in Chinese](https://support.apple.com/en-gb/guide/mac-help/mh26684/mac#mchlf6cd6fbc). It's weird.

mister-ben:
I see the reason it's different now. If Chrome is running in Chinese, it sets the minimum font size to 12px, (in preferences > appearance > customise font). In other languages it's unset. Our CSS sets the slider handle to .9em, which we expect to be 9px, but on the Chinese browser it's 12px becuase of that min value.

friendtasy:
[...]

Issue #8066, opened 2023-01-10
How to adjust the size of the controlBar?


YorkeSun:
### Description

I just started learning to use video.js and I use "simple-embed" from the official example. After I adjusted the width to 100%, I found that the controlbar is very small. How can I make the controlbar adaptively adjust the size?

https://raw.githubusercontent.com/YorkeSun/photo/main/videojs-menu.png

### Steps to reproduce

I just changed these codes:
` <video id="example_video_1" webkit-playsinline="true" style="width:100%;" class="video-jsn vjs-big-play-centered" controls preload="auto" poster="http://vjs.zencdn.net/v/oceans.png" data-setup="{}">`

### What version of Video.js are you using?

7.20.3

### Video.js plugins used.

none

### What browser(s) including version(s) does this occur with?

chrome

### What OS(es) and version(s) does this occur with?

mac

amtins:
You can try something like :
Please refer to https://videojs.com/guides/options/#responsive

**JavaScript** 
```JavaScript
const player = new videojs('player', {
    controls: true,
    responsive: true,
    //other options...
});
```
**CSS**
```css
.vjs-layout-tiny {
    font-size:  8px;
}
.vjs-layout-x-small {
    font-size: 10px;
}
.vjs-layout-small {
    font-size: 12px;
}
.vjs-layout-medium {
    font-size: 14px;
}
.vjs-layout-large {
    font-size: 16px;
}
.vjs-layout-x-large {
    font-size: 18px;
}
.vjs-layout-huge {
    font-size: 20px;
}
```
Please refer to https://videojs.com/guides/options/#breakpoints, you can also check https://videojs.com/guides/layout/

If you want to be more flexible, you can imagine something like the code below that allows for easy theming:
```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Breakpoints</title>
    <link href="https://vjs.zencdn.net/7.20.3/video-js.css" rel="stylesheet" />
    <style>
        :root {
            /* Defines the default font sizes */
            --layout-default-font-size: 10px;
            --layout-tiny-font-size: calc(var(--layout-default-font-size) * 0.8);
            --layout-x-small-font-size: var(--layout-default-font-size);
            --layout-small-font-size: calc(var(--layout-default-font-size) * 1.2);
[...]

PR #8260, opened 2023-05-02
feat: Add useSVGIcons option

wseymour15:
## Description
This change adds a player option titled `useSVGIcons` which will replace the font icons from `videojs/font` with SVGs that are stored in the VJS repo's SCSS. Eventually, we will want to remove all `videojs/font` styling, but for now we give the user the option to replace it.

To test: run this repo locally, use `npm start`, and go to `http://localhost:9999/sandbox/svg-icons.html` to see the new icons.

## Specific Changes proposed
- `useSVGIcons` player option is added, which adds a high level `vjs-svg-icons-enabled` class to the `video` element.
- SCSS with all of the icons as SVG data URLs.
- SCSS to overwrite font icons when the option is available.
- Test page to display all of the svg icons that are available.

## Requirements Checklist
  - [x] Docs/guides updated

### DIFF
--- a/src/css/_icons.scss
+++ b/src/css/_icons.scss
@@ -1,0 +1,224 @@
+@import "private-variables";
+
+// A CSS implementation of the icons used by video.js.
+//
+// The goal is to replace all icons from the font family pulled from videojs/font entirely.
+// This project currently uses fonts. We want to replace this with SVG data
+// URLs. This will ensure consitency between versions, as well as simplified
+// and straight-forward customization.
+
+// Default styling for all SVG icons
+.vjs-svg-icons-enabled {
+  .vjs-icon-placeholder {
+    // Default height and width of icons. This can be easily changed in custom CSS.
+    height: 16px;
+    width: 16px;
+  
+    // This determines the color of the SVGs.
+    // Here is a great tool to convert HEX color values to CSS filter if the color needs to be changed:
+    // https://isotropic.co/tool/hex-color-to-css-filter/
+    filter: $primary-foreground-color-svg;
+    -webkit-filter: $primary-foreground-color-svg;
+  
+    display: inline-block !important;
+    background-repeat: no-repeat;
+    background-position: center;
+  
+    // Overwrite any font content
+    &:before {
+      content: none !important;
+    }
+  }
+  
+  // SVG shadow on hover and focus
+  .vjs-icon-placeholder:hover,
+  .vjs-control:focus .vjs-icon-placeholder {
+    -webkit-filter: $primary-foreground-color-svg drop-shadow(0 0 0.25em #fff);
+    filter: $primary-foreground-color-svg drop-shadow(0 0 0.25em #fff);
+  }
+}
+
+.vjs-icon-captions-svg {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 576 512' %3E%3C!--! Font Awesome Pro 6.4.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc. --%3E%3Cpath d='M0 96C0 60.7 28.7 32 64 32H512c35.3 0 64 28.7 64 64V416c0 35.3-28.7 64-64 64H64c-35.3 0-64-28.7-64-64V96zM200 208c14.2 0 27 6.1 35.8 16c8.8 9.9 24 10.7 33.9 1.9s10.7-24 1.9-33.9c-17.5-19.6-43.1-32-71.5-32c-53 0-96 43-96 96s43 96 96 96c28.4 0 54-12.4 71.5-32c8.8-9.9 8-25-1.9-33.9s-25-8-33.9 1.9c-8.8 9.9-21.6 16-35.8 16c-26.5 0-48-21.5-48-48s21.5-48 48-48zm144 48c0-26.5 21.5-48 48-48c14.2 0 27 6.1 35.8 16c8.8 9.9 24 10.7 33.9 1.9s10.7-24 1.9-33.9c-17.5-19.6-43.1-32-71.5-32c-53 0-96 43-96 96s43 96 96 96c28.4 0 54-12.4 71.5-32c8.8-9.9 8-25-1.9-33.9s-25-8-33.9 1.9c-8.8 9.9-21.6 16-35.8 16c-26.5 0-48-21.5-48-48z'/%3E%3C/svg%3E");
+}
+
+.vjs-icon-fullscreen-enter-svg {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 448 512' %3E%3C!-- Font Awesome Pro 5.15.4 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) --%3E%3Cpath d='M0 180V56c0-13.3 10.7-24 24-24h124c6.6 0 12 5.4 12 12v40c0 6.6-5.4 12-12 12H64v84c0 6.6-5.4 12-12 12H12c-6.6 0-12-5.4-12-12zM288 44v40c0 6.6 5.4 12 12 12h84v84c0 6.6 5.4 12 12 12h40c6.6 0 12-5.4 12-12V56c0-13.3-10.7-24-24-24H300c-6.6 0-12 5.4-12 12zm148 276h-40c-6.6 0-12 5.4-12 12v84h-84c-6.6 0-12 5.4-12 12v40c0 6.6 5.4 12 12 12h124c13.3 0 24-10.7 24-24V332c0-6.6-5.4-12-12-12zM160 468v-40c0-6.6-5.4-12-12-12H64v-84c0-6.6-5.4-12-12-12H12c-6.6 0-12 5.4-12 12v124c0 13.3 10.7 24 24 24h124c6.6 0 12-5.4 12-12z'/%3E%3C/svg%3E");
+}
+
+.vjs-icon-fullscreen-exit-svg {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 448 512' %3E%3C!--! Font Awesome Pro 6.4.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc. --%3E%3Cpath d='M160 64c0-17.7-14.3-32-32-32s-32 14.3-32 32v64H32c-17.7 0-32 14.3-32 32s14.3 32 32 32h96c17.7 0 32-14.3 32-32V64zM32 320c-17.7 0-32 14.3-32 32s14.3 32 32 32H96v64c0 17.7 14.3 32 32 32s32-14.3 32-32V352c0-17.7-14.3-32-32-32H32zM352 64c0-17.7-14.3-32-32-32s-32 14.3-32 32v96c0 17.7 14.3 32 32 32h96c17.7 0 32-14.3 32-32s-14.3-32-32-32H352V64zM320 320c-17.7 0-32 14.3-32 32v96c0 17.7 14.3 32 32 32s32-14.3 32-32V384h64c17.7 0 32-14.3 32-32s-14.3-32-32-32H320z'/%3E%3C/svg%3E");
+}
+
+.vjs-icon-play-svg {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512' %3E%3C!--! Font Awesome Pro 6.4.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc. --%3E%3Cpath d='M73 39c-14.8-9.1-33.4-9.4-48.5-.9S0 62.6 0 80V432c0 17.4 9.4 33.4 24.5 41.9s33.7 8.1 48.5-.9L361 297c14.3-8.7 23-24.2 23-41s-8.7-32.2-23-41L73 39z'/%3E%3C/svg%3E");
+}
+
+.vjs-icon-play-circle-svg {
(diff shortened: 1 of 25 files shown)


src/js/component.js
```diff
@@ -524,6 +525,52 @@ class Component {
...
+   * @param {Component|Element} [el=this.el()]
+   *        Component or element to set the title on. Defaults
+   *
+   * @return {Element}
+   *        The newly created icon element.
+   */
+  setIcon(iconName, el = this.el()) {
+    if (!this.player_.options_.useSVGIcons) {
+      return;
+    }
```
misteroneill, line 543: Putting this here means it's easy to remove in Video.js 9 - good thinking. Maybe a TODO comment here to remove this in Video.js 9?

src/js/control-bar/audio-track-controls/audio-track-button.js
```diff
@@ -25,6 +25,8 @@ class AudioTrackButton extends TrackButton {
     options.tracks = player.audioTracks();
 
     super(player, options);
+
+    this.setIcon('vjs-icon-audio');
```
misteroneill, line 29: If all the icons are prefixed with `vjs-icon-`, perhaps this should be automatically prepended, so we can simply call `this.setIcon('audio')`?

misteroneill (commented):
I like the direction! Two minor comments/thoughts.

src/js/component.js
```diff
@@ -524,6 +525,52 @@ class Component {
...
+
+    const xmlnsURL = 'http://www.w3.org/2000/svg';
+
+    // The below creates an element in thw format of:
```
mister-ben, line 547:
```suggestion
    // The below creates an element in the format of:
```

test/unit/player.test.js
```diff
@@ -967,6 +967,16 @@ QUnit.test('should add a touch-enabled classname when touch is supported', funct
   player.dispose();
 });
 
+QUnit.test('should add a svg-icons-enabled classname when svg icons are supported', function(assert) {
+  assert.expect(1);
+
+  const player = TestHelpers.makePlayer({useSVGIcons: true});
+
+  assert.notEqual(player.el().className.indexOf('vjs-svg-icons-enabled'), -1, 'svg-icons-enabled classname added');
```
gkatsev, line 975: I was originally going to say to use `classList.contains()` but then I realized we have a helper for this already
```suggestion
  assert.ok(player.hasClass('vjs-svg-icons-enabled'), 'svg-icons-enabled classname added');
```

build/images.js
gkatsev, line 1: what does this file do and when is it run?

images/icons.svg
gkatsev, line 1: is this a file that we should check in or something that should be generated on build/publish?

gkatsev, line 1: need to add the `images/` folder to here for the netlify build to properly display https://deploy-preview-8260--videojs-preview.netlify.app/sandbox/svg-icons.html

https://github.com/videojs/video.js/blob/main/build/netlify.js#L32

build/images.js
gkatsev: looks similar to the build/sandbox.js file

images/icons.svg
wseymour15: With the above comment, I _believe_ this should be taken care of, but let me know if that is not the case.

src/js/component.js
```diff
@@ -524,6 +525,57 @@ class Component {
...
+      className: 'vjs-icon-placeholder vjs-svg-icon'
+    }, {'aria-hidden': 'true'});
+
+    const svgEl = document.createElementNS(xmlnsURL, 'svg');
+
+    svgEl.setAttributeNS(null, 'viewBox', '0 0 512 512');
+    const useEl = document.createElementNS(xmlnsURL, 'use');
+
+    svgEl.appendChild(useEl);
+    useEl.setAttributeNS(null, 'href', `../images/icons.svg#vjs-icon-${iconName}`);
```
misteroneill, line 564: I know we talked about this path being sufficient, but the more I think about it, the more it makes sense for there to be an additional Video.js option for the URL for the SVG sprite.

It could be `../images/icons.svg` by default and that should work for the Video.js CDN or unpkg or similar, but I can see use-cases where a Video.js user will want to point that to a different location.

This would also allow people to create their own icons.

```js
const player = videojs('example', {
  iconsUrl: '//my-cdn.com/my-icons.svg'
});
```

misteroneill (commented):
Looks great, but I had a thought about the SVG sprite URL.

src/css/_icons.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  fill: #FFFFFF;
+  height: 16px;
+  width: 16px;
+
+  // Overwrite any font content
+  &:before {
+    content: none !important;
+  }
+}
```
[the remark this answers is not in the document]
wseymour15: I'm trying to think of a better way.. without it, we basically have 2 icons overlapping each other because the font icons are added through css content. A different option would be to go through everywhere those fonts are added and do a `(:not)` check to ensure that the content is only being added when svgs are not supposed to be used, but not sure if that is a better solution.

We can remove this once we remove the font icons entirely, but this might be the best path for now.

images/icons.svg
gkatsev: How was this file created? I'm wondering whether the svg should be checked in or the thing that was used to generate it should be checked in or at least documented somewhere?

Also, potentially should go into the `src/images/` directory instead of at the root?

wseymour15: This file was created by downloading the raw SVGs from either Font Awesome or Material UI, and I basically added that SVG code manually to this file here. If we need more icons, we'll have to insert it in a similar fashion into this file.

I'm definitely open to opinions on where the best place for this file is, and `src/images` seems like a good place. I updated that here: https://github.com/videojs/video.js/pull/8260/commits/4997916be56ce9990abc5bc45e075d98a91335e0

src/js/component.js
wseymour15: This is a very good idea! My only question is, do we expect these custom SVG sprites to have the exact same definitions of SVGs as our SVG sprite?

For icons that already exist in VJS, we expect them to have a certain id on the SVG element. If the custom sprite were to work with this implementation, it would need to have those same ids on the `defs` in their sprite, or else the preexisting icons won't show up.

Would it make more sense to be able to just pass that URL into the `setIcon` function as a param? This way, if a user wanted to overwrite an icon, they could get whatever component they want and call the function to update the icon from their sprite url. This would also allow them to avoid sprites entirely and call a file with a single SVG.

Definitely open to all ideas!

amtins:
@wseymour15 this feature is absolutely awesome and is a fantastic addition to video.js! 😍 👍🏽 

I have a little feedback on the `responsive` mode of the player which seems to be not properly supported anymore.

By setting the default font size per `breakpoint` it is possible to have buttons bigger or smaller depending on the screen size, see use case described here https://github.com/videojs/video.js/issues/8066#issuecomment-1378863842

wseymour15:
> Perhaps this comment is better formulated

Thanks @amtins for the clarification, the videos were extremely helpful. I went ahead and updated the styling on the icons to use em instead of px so it is truly responsive to those style changes. Feel free to take a look and let me know if that appears to fix the issue!

dzianis-dashkevich (approved):

mister-ben:
Nit, but the different volume icons don't seem quite in line with each other, the speaker is larger in the medium and low states.

The CC icon seems much more legible at a smaller size now compared to the font.

wseymour15:
> Nit, but the different volume icons don't seem quite in line with each other, the speaker is larger in the medium and low states.

Agreed, those did look a bit consistent. This is because these were a mix of the Font Awesome and Material UI Icons. Ideally, I wanted to go with all of the Font Awesome Icons here, but some of them required a PRO account.

With my most recent commit, I decided to use Material UI Icons for all of the volume SVGs. I had to add a little extra CSS because those icons filled less of the viewBox then the other icons and appeared a bit too small.

BONUS: These icons require less data then the Font Awesome ones :)

mister-ben:
This brings back an issue on Chrome in Chinese that the slider handles are misplaced. That was fixed with the font icons in #7990.
It's to do with the browser setting a minimum font size valuewhen the OS or [app langauge](https://support.apple.com/en-gb/guide/mac-help/mh26684/mac#mchlf6cd6fbc) is Chinese. This complicates using em units for positioning.  Conceivably users who set a minium font size for accessibility reasons would also be affected. Setting the page to Chinese or user language preference doesn't change that minium setting, so does not reproduce.
grafik

On the other hand it fixes the placement of the slider on the veritcal slider, which is unresolved with the font. #8092 
grafik

src/css/components/_progress.scss
```diff
@@ -80,6 +80,23 @@
...
+// Remove content from play-progress when using SVGs.
+.vjs-svg-icons-enabled .vjs-play-progress {
+  &:before {
+    content: none !important;
+  }
+}
+
+.vjs-play-progress .vjs-svg-icon {
+  position: absolute;
+  top: -0.3em;
```
mister-ben, line 92:
```suggestion
  top: -0.3.5em;
```
This looks better to me on Chinese and non-Chinese Chrome

src/css/components/_volume.scss
```diff
@@ -144,6 +166,17 @@
   }
 }
 
+// Update placement of circle icon when using SVG icons
+.vjs-slider-horizontal .vjs-volume-level .vjs-svg-icon {
+  top: -0.2em;
+  right: -0.3em;
+  line-height: 0.35em;
```
mister-ben, line 173:
```suggestion
  line-height: 0.15em;
```
And this seems to work on both better too

mister-ben:
Your last changes still didn't look right on my machine. Do my suggestions look right to you? Interesting and frustrating if we're getting different results.

src/css/components/_volume.scss
wseymour15: > Your last changes still didn't look right on my machine. Do my suggestions look right to you? Interesting and frustrating if we're getting different results.

The changes you suggested do look right, but the line height adjustment did not appear to make much of a difference in the progress slider. However, changing the smaller line-height in the volume slider did seem to make a bit of a difference. The sliders look good to me in both Chinese and non-Chinese in Chrome here: https://deploy-preview-8260--videojs-preview.netlify.app/sandbox/svg-icons-enabled.html

src/css/_icons.scss
```diff
@@ -0,0 +1,29 @@
+// CSS styles for SVG icons used throughout video.js.
+//
+// The goal is to replace all icons from the font family pulled from videojs/font entirely.
+// This project currently uses fonts. We want to replace this with SVGs from
+// images/icons.svg. This will ensure consitency between versions, as well as simplified
+// and straight-forward customization.
+
+// Default styling for all SVG icons
+.vjs-svg-icon {
+  display: inline-block !important;
```
misteroneill, line 10: Is the `!important` necessary here?

src/js/player.js
```diff
@@ -549,6 +550,25 @@ class Player extends Component {
...
+    if (options.useSVGIcons) {
+      // Add SVG Sprite to the DOM
+      const parser = new window.DOMParser();
+      const parsedSVG = parser.parseFromString(icons, 'image/svg+xml');
+      const errorNode = parsedSVG.querySelector('parsererror');
+
+      if (errorNode) {
+        log.warn('Failed to load SVG Icons. Falling back to Font Icons.');
+        return;
+      }
```
misteroneill, line 562: Good idea here. Should this also set the player option to `false` so that other code doesn't think SVG icons are disabled?

src/css/_icons.scss
wseymour15: No it is not! I was using important to overwrite the button css for icons, but I just added some new styling in that file instead. https://github.com/videojs/video.js/pull/8260/commits/357adf1b267515534e4253af49a46a59777fdc1d#diff-db2130389e768f54f29c1ef67a128d854ea929e98f23d561de7f5b534747cd91R31

src/js/player.js
wseymour15: Great call! I reset the option, updated the logic to not return if the failure happened, as well as make sure this all happens before the children components are initialized: https://github.com/videojs/video.js/commit/357adf1b267515534e4253af49a46a59777fdc1d#diff-f92dc431786299ba52ba41d6901e029cc1eeb2f5603653bd03eb4d88a27f5020R521

misteroneill (approved):

mister-ben:
The inconsistencies in these icons are bothering me:
- The fullscreen exit icon is more rounded than fullscreen enter
- The new CC icon is better than the old as it's much more legible, is there an matching icon for HD?
- The play and pause icons are rounded, but most other icons are not, e.g. the arrows on replay, next item, download. I'd be minded to keep the simple, straight material icons as the _default_ set.

Could we reduce the number of icons? I know why they're in the font, but I think the Facebook etc social icons in particular ought not to be included as they are not used by the core player. Instead, we should have a way for a plugin to supplement icons, some sort of `registerIcon()`.

The size of some icons in the source seem inefficient. Are we happy with how rollup-plugin-svg is optimising them in the build?

My major concern is how we will allow users to change the icons and how easily, for example to a rounded set. As this is experimental and opt-in, that could be addressed in a later PR as long as the current approach isn't locking that out too much.

wseymour15:
Thanks @mister-ben, I greatly appreciate this feedback.

> The fullscreen exit icon is more rounded than fullscreen enter

I do see that this is slightly off.. I could not find these icons at all in material UI, and the matching fullscreen exit icon required a Font Awesome pro account to download.

> The new CC icon is better than the old as it's much more legible, is there an matching icon for HD?

Same here.. there IS a better HD icon in Font Awesome.. that requires a pro account.

> The play and pause icons are rounded, but most other icons are not.

I made an update to make these icons align more closely.

I believe that making the icons more consistent can probably come with future work. I think it would be to our benefit to have access to a Font Awesome Pro account. Material UI doesn't quite have everything we need, and many of the "ideal" icons are locked in Font Awesome behind a pro account.

> Could we reduce the number of icons?

Great point. I know we do use a few of these social icons on the Brightcove side, but they are not a part of video.js anywhere. I'm leaning towards keeping the same icons as the past for now. However I am all for taking the time later to determine which icons would and would not be valuable to keep.

> The size of some icons in the source seem inefficient. Are we happy with how rollup-plugin-svg is optimising them in the build?

Most of these icons come straight from material UI or Font Awesome, I am sure there are ways to make some of them smaller, but I think for now we should probably just keep them as is.

For the build, this was a talking point on the Brightcove side initially, and we did some testing. We are satisfied with how much larger the build has become after the change, and assume once we remove the fonts entirely, it should be smaller than before. Here is some comparison between the builds before and after.
screen_shot_2023-05-30_at_1 35 42_pm

> My major concern is how we will allow users to change the icons and how easily..

This sounds related to your idea of a `registerIcon()` function, which I think would bring a lot of value to VJS. It could be a bit tricky especially if we are trying to avoid things like using `innerHTML` or trying to avoid Data URIs. However, it would probably be possible to pass in a URL to an SVG file, manipulate it and add it to the DOM.

I also think that there will currently be ways the user could add icons by customizing their own player, it would just not be as simple as a function call.

All in all, I think we are at a pretty good spot to move forward with what we have, and then iterate with future improvements.

wseymour15:
> Great points, @mister-ben. I propose that we change the name to `experimentalSvgIcons` to indicate that it's a work in progress. We can review the specific icons used and customization paths in the future (even if it's just a new sample/sandbox page).

Totally agree. Updated the current implementation to change the option name, as well as updated the docs in my other PR.

PR #8270, opened 2023-05-11
fix: Document Picture-in-Picture: Use width/height instead of initialAspectRatio

beaufortfrancois:
The `initialAspectRatio` option was removed from the Document Picture-in-Picture spec in favour of `width` and `height`, so we should use them. See https://github.com/WICG/document-picture-in-picture/pull/36


mister-ben:
Thanks. Seems setting width and height has the same issue as with aspect ratio of sizing the outer pip window rather than setting the content size.

mister-ben (approved):

adrums86 (approved):

beaufortfrancois:
Shall we merge?

PR #8277, opened 2023-05-14
fix(jsdoc): Corrections to jsdoc

mister-ben:
## Description
Corrects some jsodc in Player, especially
- getter/setters not specifying that they won't return when setting
- getter/setters that stated they could return the player, very old behaviour

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -2209,6 +2209,29 @@ QUnit.test('should not allow to register custom player when any player has been 
   videojs.registerComponent('Player', Player);
 });
 
+QUnit.test('setters getters passed to tech', function(assert) {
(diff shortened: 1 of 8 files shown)


test/unit/player.test.js
```diff
@@ -2209,6 +2209,29 @@ QUnit.test('should not allow to register custom player when any player has been
   videojs.registerComponent('Player', Player);
 });
 
+QUnit.test('setters getters passed to tech', function(assert) {
```
mister-ben, line 2212: This test is unrelated to the changes but makes Codecov happy as it increases coverage...

misteroneill (approved):

PR #8287, opened 2023-05-21
feat: Seek bar smooth seeking

### LINKED ISSUE #8283: Slider does not follow the mouse.
### Description

When dragging the progress bar to an unloaded progress while the video is paused, there is an issue where the slider does not follow the mouse.

### Reduced test case

https://codesandbox.io/s/suspicious-https-x4burl?file=/src/App.js

### Steps to reproduce

1. play video
2. quickly pausing the video.
3. dragging the progress bar to the end when the later part of the video has not been loaded yet.
4. I dragged the progress bar to 31 seconds, but the slider didn't follow, and when I finished dragging, the time indicated by the slider and the current video time were inconsistent.

I would like the slider to follow immediately like in YouTube videos when dragging, and then to load the video.

### Errors

image

8.3.0

### Video.js plugins used.

_No response_

chrome  111.0.5563.146

Mac OS 12.3

amtins:
## Description

This PR adds a player option called `enableSmoothSeeking`, which is `false` by default, to provide a smoother seeking experience on mobile and desktop devices.

[player-smooth-seek-browserstack.webm](https://github.com/videojs/video.js/assets/34163393/0f1fc92b-58bf-4303-9b00-c5c92585adb7)

This PR includes the fix #8285 and should fix #8283.

### How to use it?

```javascript
// Enables the smooth seeking
const player = videojs('player', {enableSmoothSeeking: true});

// Disable the smooth seeking
player.options({enableSmoothSeeking: false});

```

## Specific Changes proposed

- **player.js** add an `option` called `enableSmoothSeeking`
- **time-display.js** add a listener to the `seeking` event if `enableSmoothSeeking` is `true` allowing to update the `CurrentTimeDisplay` and `RemainingTimeDisplay` in real time
- **seek-bar.js** `update` the seek bar on `mousemove` event  if `enableSmoothSeeking` is `true`
- add test cases
- refer to #8285 for the `player.currentTime` modification

### Benchmark

The number of calls to the `update` method when the player is `playing` versus when `seeking` with smooth seeking enabled. The test was performed 10 times in a row and consisted of a counter that incremented by one on each call. It was inserted in the anonymous function provided to `requestNamedAnimationFrame` of the `update` method. 

The duration of each round was 5 seconds. The `enableInterval_` it started when the `play` button was clicked and stopped at the end of the `setTimeout`. The `handleMouseMove` it started on the `mousedown`, then was followed by frantic movement on the seek bar and stopped at the end of the `setTimeout`.

|            | enableInterval_ | handleMouseMove |
|----------|-----------------|-----------------|
| Round 1  | 142             | 114             |
| Round 2  | 144             | 133             |
| Round 3  | 145             | 126             |
| Round 4  | 152             | 114             |
| Round 5  | 144             | 122             |
| Round 6  | 150             | 119             |
| Round 7  | 146             | 112             |
| Round 8  | 155             | 133             |
| Round 9  | 144             | 131             |
| Round 10 | 133             | 121             |

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -324,6 +324,10 @@ class SeekBar extends Slider {
 
     // Set new time (tell player to seek to new time)
     this.userSeek_(newTime);
+
+    if (this.player_.options_.enableSmoothSeeking) {
+      this.update();
+    }
   }
 
   enable() {
--- a/src/js/control-bar/time-controls/time-display.js
+++ b/src/js/control-bar/time-controls/time-display.js
@@ -26,7 +26,7 @@ class TimeDisplay extends Component {
   constructor(player, options) {
     super(player, options);
 
-    this.on(player, ['timeupdate', 'ended'], (e) => this.updateContent(e));
+    this.on(player, ['timeupdate', 'ended', 'seeking'], (e) => this.update(e));
     this.updateTextNode_();
   }
 
@@ -69,6 +69,20 @@ class TimeDisplay extends Component {
     this.textNode_ = null;
 
     super.dispose();
+  }
+
+  /**
+   * Updates the displayed time according to the `updateContent` function which is defined in the child class.
+   *
+   * @param {Event} [event]
+   *          The `timeupdate`, `ended` or `seeking` (if enableSmoothSeeking is true) event that caused this function to be called.
+   */
+  update(event) {
+    if (!this.player_.options_.enableSmoothSeeking && event.type === 'seeking') {
+      return;
+    }
+
+    this.updateContent(event);
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -5316,7 +5316,9 @@ Player.prototype.options_ = {
   breakpoints: {},
   responsive: false,
   audioOnlyMode: false,
-  audioPosterMode: false
+  audioPosterMode: false,
+  // Default smooth seeking to false
+  enableSmoothSeeking: false
 };
 
 [
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2509,10 +2509,6 @@ class Player extends Component {
 
     this.techCall_('setCurrentTime', seconds);
     this.cache_.initTime = 0;
-
-    if (isFinite(seconds)) {
-      this.cache_.currentTime = Number(seconds);
-    }
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -2509,6 +2509,10 @@ class Player extends Component {
 
     this.techCall_('setCurrentTime', seconds);
     this.cache_.initTime = 0;
+
+    if (isFinite(seconds)) {
+      this.cache_.currentTime = Number(seconds);
+    }
   }
 
   /**
--- a/test/unit/player.test.js
+++ b/test/unit/player.test.js
@@ -3419,3 +3419,66 @@ QUnit.test('should not reset the error when the tech triggers an error that is n
   errorStub.restore();
   log.error.restore();
 });
+
+QUnit.test('smooth seeking set to false should not update the display time components or the seek bar', function(assert) {
+  const player = TestHelpers.makePlayer({});
+  const {
+    currentTimeDisplay,
+    remainingTimeDisplay,
+    progressControl: {
+      seekBar
+    }
+  } = player.controlBar;
+  const currentTimeDisplayUpdateContent = sinon.spy(currentTimeDisplay, 'updateContent');
+  const remainingTimeDisplayUpdateContent = sinon.spy(remainingTimeDisplay, 'updateContent');
+  const seekBarUpdate = sinon.spy(seekBar, 'update');
+
+  assert.false(player.options().enableSmoothSeeking, 'enableSmoothSeeking is false by default');
+
+  player.trigger('seeking');
+
+  assert.ok(currentTimeDisplayUpdateContent.notCalled, 'currentTimeDisplay updateContent was not called');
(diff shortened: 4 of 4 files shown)


amtins:
## Observation

- It would be interesting to rename `this.currentTime_` to `this.previousCurrentTime_` and `this.duration_` to `this.previousDuration_` to improve readability and better define their purpose
https://github.com/videojs/video.js/blob/f1558c660bbc132b23dbd63b2f3cdf7b6e60d00b/src/js/control-bar/progress-control/seek-bar.js#L164-L178

- Thanks to the fix #8285 it is now possible to rely only on `this.player_.getCache().currentTime` and thus remove the following logic
 https://github.com/videojs/video.js/blob/f1558c660bbc132b23dbd63b2f3cdf7b6e60d00b/src/js/control-bar/progress-control/seek-bar.js#L213-L217
 https://github.com/videojs/video.js/blob/f1558c660bbc132b23dbd63b2f3cdf7b6e60d00b/src/js/control-bar/progress-control/play-progress-bar.js#L64-L66
 https://github.com/videojs/video.js/blob/f1558c660bbc132b23dbd63b2f3cdf7b6e60d00b/src/js/control-bar/time-controls/current-time-display.js#L39

mister-ben (approved):
Looks good and is more in line with other players' behaviour. Thanks!. One observation is that you no longer have an indication of where you were in the video when you started seeking, if that matters.

src/js/control-bar/progress-control/seek-bar.js
```diff
@@ -39,7 +40,13 @@ class SeekBar extends Slider {
...
    */
   constructor(player, options) {
-    super(player, options);
+    const options_ = merge({
+      playerOptions: {
+        enableSmoothSeeking: false
+      }
+    }, options);
+
+    super(player, options_);
```
gkatsev, line 49: while it makes sense for this to be a player option, I think that rather than setting a default player option here, we should have it be translated into a seek-bar option called `enableSmoothSeeking`.

Maybe something like this (I didn't verify this works)

```suggestion
    const opts = merge({
      enableSmoothSeeking: options.playerOptions.enableSmoothSeeking ?? false
    }, options);

    super(player, opts);
```

Then below, it could use `this.options_.enableSmoothSeeking`

gkatsev: I guess the alternative would be to always do `this.player().options().enableSmoothSeeking` (then someone could more easily update this on the fly and won't need to drill down to the seek bar via `player.controlBar.progressControl.seekBar.options_.playerOptions.enableSmoothSeeking = true`, which I had to do to test this out. However, I think making it a seek-bar option is fine and makes sense.

gkatsev (commented):
At first, I was worried about this change because the original reason we didn't update this is that internally we are still seeking directly and basically relying on it for the preview and we were worried about the potential jump it a seek failed and the mismatch between the preview and the seek bar. However, at least on desktop, it seems to be working well enough and I actually think this behavior feels a bit better. Keeping it as an option (and perhaps enabling it by default later on) makes sense to me, though, still.

Overall, this looks good to me, but I had one suggestion about the option.

gkatsev:
Also, appreciate all the PRs you've been opening, thanks!

src/js/control-bar/progress-control/seek-bar.js
amtins: @gkatsev thanks for the suggestion. 

> I guess the alternative would be to always do `this.player().options().enableSmoothSeeking `(then someone could more easily update this on the fly and won't need to drill down to the seek bar

Your suggestion is much better because it offers a nicer developer experience. 

Here's what the changes could look like:

**seek-bar.js**
```javascript
    super(player, options);
    this.setEventHandlers_();
  }

  handleMouseMove(event, mouseDown = false) {
     // ...
     if (this.player_.options_.enableSmoothSeeking) {
      this.update();
    }
  }
```

**time-display.js**

```javascript
    super(player, options);

    this.on(player, ['timeupdate', 'ended', 'seeking'], (e) => this.update(e));
    this.updateTextNode_();
  }

  update(event) {
    if (!this.player_.options_.enableSmoothSeeking && event.type === 'seeking') {
      return;
    }

    this.updateContent(event);
  }
```

These changes remove `merge` `options`, reducing the number of lines of code to maintain. As a result, a developer will be able to toggle the smooth seeking on the fly:


```javascript
player.options({enableSmoothSeeking : true});
player.options({enableSmoothSeeking : false});
```

Finally, relying directly on player options is something already done for other components: `clickable-component`, `live-tracker`, `poster-image` and `picture-in-picture-toggle`.

What do you think about these changes?

misteroneill: Sounds good to me!

misteroneill:
Echoing what @gkatsev said, it's a huge help and great to see such active contribution from a community member! Thanks!

src/js/control-bar/progress-control/seek-bar.js
amtins: @mister-ben thank you for the feedback, I'll try to make the changes by the end of the week.

amtins: I pushed the changes. However, I don't see where I can add the documentation for this new option.

mister-ben:
Should this still have the needs-updates tag?

amtins:
@mister-ben I've implemented @gkatsev's suggestions (thanks again), which improves the developer experience. From my point of view, it looks like everything has been addressed.

mister-ben:
Great, let's get this merged before the next minor release then. We'll need to add it to the options guide in the videos.com repo too.

PR #8300, opened 2023-06-01
test: fix Safari test failures

misteroneill:
## Description

It looks like after #8274 was merged, tests began failing consistently in Safari. Upon investigation, it seems that code in the test harness was hitting `load()` before a fake tech had been created, causing it to throw and then the test to time out.

Oddly enough, the tests didn't fail in the PR... only after it was merged! 🤷 

## Specific Changes proposed

- Do a `null` check for the `tech_` before inspecting it.

## Requirements Checklist


adrums86 (approved):
:shipit:

PR #8309, opened 2023-06-04
feat: Add mediasession support

mister-ben:
## Description
Adds built-in, opt-in support for populating mediasession metadata and setting action handlers

## Specific Changes proposed
- Enabled with a truthy `mediaSession` player setup option. Since implementors may be using the API, not on by default. Possibly this could be made default in a major version.
- `mediaSession.skipTime` (seconds) will override the default skip interval (if the browser/os doesn't request a time) 
- Supports getting metadata from `getMedia()` (if `loadMedia()` has been used) or from [playlist](https://github.com/videojs/videojs-playlist) metadata.
- Adds a player event `updatemetadata` with the resolved metadata object as its second argument. The metadata can be modified synchronously if needed, i.e. there is another source of truth for metadata.
- Guards against actions during ads, assuming a [contrib-ads](https://github.com/videojs/videojs-contrib-ads)-based implementation.

This will drop overall test coverage slightly, as the action handlers can't be called in a test.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/mediasession.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+  };
+
+  this.on('playing', () => {
+    // Called on each `playing` rather than `sourceset`, in case of multiple players on a page
+    updateMediaSession();
+    ms.playbackState = 'playing';
+  });
+
+  this.on('paused', () => {
```
amtins, line 125: Shouldn't it be `pause` instead of `paused` ?

```suggestion
  this.on('pause', () => {
```

test/unit/mediasession.test.js
```diff
@@ -0,0 +1,411 @@
...
+  this.player.playbackRate = () => 1.5;
+
+  this.player.trigger('timeupdate');
+  assert.true(window.navigator.mediaSession.setPositionState.calledWith({
+    duration: 10,
+    playbackRate: 1.5,
+    position: 5
+  }), 'setPositionState called');
+
+  this.player.trigger('paused');
```
amtins, line 284: Referring to my comment above:
> Shouldn't it be `pause` instead of `paused` ?

src/js/mediasession.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // This allows the metadata to be updated before being set, e.g. if loadMedia() is not used.
+    this.trigger('updatemediasession', mediaSessionData);
+
+    ms.metadata = new window.MediaMetadata(mediaSessionData);
+  };
+
+  const updatePositionState = () => {
+    const dur = parseFloat(this.duration());
```
amtins, line 108: Question, is there a case that requires a `parseFloat`?

src/js/mediasession.js
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * @method initMediaSession Sets up media session if supported and configured
+ * @this { import('./player').default } Player
+ */
+export const initMediaSession = function() {
+  if (!this.options_.mediaSession || !('mediaSession' in window.navigator)) {
+    return;
+  }
+  const ms = window.navigator.mediaSession;
+  const defaultSkipTime = 15;
```
amtins, line 13: Question, do you think it would be a good idea to use `skipButtons` values if they are available? This would allow a unified experience between the `player` and the `mediaSession`.

amtins:
@mister-ben everything works well, except for the minor issue related to the `pause` event. 

I found a small issue in [videojs-playlist](https://github.com/videojs/videojs-playlist/blob/main/src/playlist-maker.js#L316-L324) that prevents the artwork to be set, this means that when the next media is played, the poster is not displayed.

src/js/mediasession.js
mister-ben: That can probably go

src/js/mediasession.js
mister-ben: I think that makes sense, yes.

mister-ben:
> @mister-ben everything works well, except for the minor issue related to the `pause` event. 
> 
> I found a small issue in [videojs-playlist](https://github.com/videojs/videojs-playlist/blob/main/src/playlist-maker.js#L316-L324) that prevents the artwork to be set, this means that when the next media is played, the poster is not displayed.

Thanks for reviewing! I think there may be a PR that would fix that issue, need to check.

src/js/mediasession.js
amtins: From a specification perspective ([w3c](https://www.w3.org/TR/2011/WD-html5-20110113/video.html#dom-media-duration), [whatwg](https://html.spec.whatwg.org/multipage/media.html#dom-media-duration-dev)), it seems safe to remove the `parseFloat`, especially as the condition immediately following it prevents any `NaN` or `Infinity`.  However, if you prefer to keep it, that's fine too.

src/js/mediasession.js
amtins: What do you think about keeping this property as it is for now and adding this modification via another PR? Because it would be nice to have the `titleBar` values as well.

PR #8314, opened 2023-06-08
fix: Don't use copyStyleSheets with documentPIP

beaufortfrancois:
Document Picture-in-Picture API owners are [removing `copyStyleSheets` option](https://github.com/WICG/document-picture-in-picture/pull/79) as it can be done in JavaScript. This PR removes this option and copy all style sheets internally to the PiP window.

@mister-ben Please review.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3074,9 +3074,9 @@ class Player extends Component {
       return window.documentPictureInPicture.requestWindow({
         // The aspect ratio won't be correct, Chrome bug https://crbug.com/1407629
         width: this.videoWidth(),
-        height: this.videoHeight(),
-        copyStyleSheets: true
+        height: this.videoHeight()
       }).then(pipWindow => {
+        this.copyStyleSheetsToWindow_(pipWindow);
         this.el_.parentNode.insertBefore(pipContainer, this.el_);
 
         pipWindow.document.body.append(this.el_);
@@ -3107,6 +3107,36 @@ class Player extends Component {
       return this.techGet_('requestPictureInPicture');
     }
     return Promise.reject('No PiP mode is available');
+  }
+
+  /**
+   * Copy document style sheets to another window.
+   *
+   *  @param {Window} win
+   *
+   * @private
+   */
+  copyStyleSheetsToWindow_(win) {
+    const allCSS = [...document.styleSheets]
+      .map((styleSheet) => {
+        try {
+          return [...styleSheet.cssRules].map((rule) => rule.cssText).join('');
+        } catch (e) {
+          const link = document.createElement('link');
+
+          link.rel = 'stylesheet';
+          link.type = styleSheet.type;
+          link.media = styleSheet.media;
+          link.href = styleSheet.href;
+          win.document.head.appendChild(link);
+        }
+      })
+      .filter(Boolean)
+      .join('\n');
+    const style = document.createElement('style');
+
+    style.textContent = allCSS;
+    win.document.head.appendChild(style);
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3076,7 +3076,7 @@ class Player extends Component {
         width: this.videoWidth(),
         height: this.videoHeight()
       }).then(pipWindow => {
-        this.copyStyleSheetsToWindow_(pipWindow);
+        Dom.copyStyleSheetsToWindow_(pipWindow);
         this.el_.parentNode.insertBefore(pipContainer, this.el_);
 
         pipWindow.document.body.append(this.el_);
@@ -3107,36 +3107,6 @@ class Player extends Component {
       return this.techGet_('requestPictureInPicture');
     }
     return Promise.reject('No PiP mode is available');
-  }
-
-  /**
-   * Copy document style sheets to another window.
-   *
-   *  @param {Window} win
-   *
-   * @private
-   */
-  copyStyleSheetsToWindow_(win) {
-    const allCSS = [...document.styleSheets]
-      .map((styleSheet) => {
-        try {
-          return [...styleSheet.cssRules].map((rule) => rule.cssText).join('');
-        } catch (e) {
-          const link = document.createElement('link');
-
-          link.rel = 'stylesheet';
-          link.type = styleSheet.type;
-          link.media = styleSheet.media;
-          link.href = styleSheet.href;
-          win.document.head.appendChild(link);
-        }
-      })
-      .filter(Boolean)
-      .join('\n');
-    const style = document.createElement('style');
-
-    style.textContent = allCSS;
-    win.document.head.appendChild(style);
   }
 
   /**
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3076,7 +3076,7 @@ class Player extends Component {
         width: this.videoWidth(),
         height: this.videoHeight()
       }).then(pipWindow => {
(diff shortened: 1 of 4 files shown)


src/js/player.js
```diff
@@ -3109,6 +3109,36 @@ class Player extends Component {
...
   }
 
+  /**
+   * Copy document style sheets to another window.
+   *
+   *  @param {Window} win
+   *
+   * @private
+   */
+  copyStyleSheetsToWindow_(win) {
```
mister-ben, line 3119: This is probably better in src/js/utils/dom.js than exposed as a method on the player.

mister-ben (changes requested):
Thanks for the update. I'm not really sure I agree that it makes any more sense for each player library to manage copying stylesheets, but if that's how it is...

The copy method would ideally not be exposed on the player, as the inline comment. We'll need to address test coverage too.

src/js/player.js
beaufortfrancois: Done.

beaufortfrancois:
> The copy method would ideally not be exposed on the player, as the inline comment. We'll need to address test coverage too.

I've moved it and added some basic tests. Let me know if that looks good to you.

beaufortfrancois:
(gentle ping)

mister-ben:
@beaufortfrancois Tahnks for the updates. I've been out for a couple of weeks, back now and will check this week.

src/js/utils/dom.js
```diff
@@ -857,3 +857,33 @@ export function computedStyle(el, prop) {
...
+ *           The window element you want to copy the document style sheets to.
+ *
+ */
+export function copyStyleSheetsToWindow(win) {
+  const allCSS = [...document.styleSheets]
+    .map((styleSheet) => {
+      try {
+        return [...styleSheet.cssRules].map((rule) => rule.cssText).join('');
+      } catch (e) {
+        const link = document.createElement('link');
```
mister-ben, line 874: When would it be expected for code in the try block to fail so we reach here? Running locally and in CI the concatenation of styles works for both `<style>` and `<link>` els, hence the test failures - no `<link>` el is added to the doc.

If `<links>` did need to be added here, they would being added in the order they occur but the content of all `<style>`s would be in a `<style>` after those `<link>`s, potentially changing the original order of rules.

beaufortfrancois: Accessing `cssRules` from cross-origin `<link>` elements such as `<link href="https://cdn.com/styles.css" rel="stylesheet">` raise an error: "Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules". That's why we try/catch here and append them manually instead.

FWIW, tests pass locally.

mister-ben: The reason the tests fail is that adding the the `<link>` stylesheet is async and it's not present in `document.stylesSheets` in time for the test.

src/js/utils/dom.js
```diff
@@ -857,3 +857,33 @@ export function computedStyle(el, prop) {
...
+        link.href = styleSheet.href;
+        win.document.head.appendChild(link);
+      }
+    })
+    .filter(Boolean)
+    .join('\n');
+  const style = document.createElement('style');
+
+  style.textContent = allCSS;
+  win.document.head.appendChild(style);
```
mister-ben, line 888: Adding each stylesheet in the original order would avoid changing the preceence of rules.

```suggestion
  [...document.styleSheets].forEach((styleSheet) => {
    try {
      const cssRules = [...styleSheet.cssRules].map((rule) => rule.cssText).join('');
      const style = document.createElement('style');

      style.textContent = cssRules;
      win.document.head.appendChild(style);
    } catch (e) {
      const link = document.createElement('link');

      link.rel = 'stylesheet';
      link.type = styleSheet.type;
      link.media = styleSheet.media;
      link.href = styleSheet.href;
      win.document.head.appendChild(link);
    }
  });
```

test/unit/utils/dom.test.js
```diff
@@ -686,3 +686,48 @@ QUnit.test('isSingleLeftClick() checks return values for mousedown event', funct
 
   assert.ok(Dom.isSingleLeftClick(mouseEvent), 'a touch event on simulated mobiles is a single left click');
 });
+
+QUnit.test('Dom.copyStyleSheetsToWindow() copies all style sheets to a window', function(assert) {
+  const fakeWindow = document.createElement('div');
```
mister-ben, line 691: This tests would need to use async to wait for the <link> el.

```suggestion
  const fakeWindow = document.createElement('div');
  const done = assert.async();

  assert.expect(7);
```

test/unit/utils/dom.test.js
```diff
@@ -686,3 +686,48 @@ QUnit.test('isSingleLeftClick() checks return values for mousedown event', funct
...
+  assert.true(fakeWindowStyle.textContent.includes(style1.textContent), 'the <style> in the fake window contains content from first <style> element');
+  assert.true(fakeWindowStyle.textContent.includes(style2.textContent), 'the <style> in the fake window contains content from second <style> element');
+
+  assert.strictEqual(fakeWindow.document.head.querySelectorAll('link[rel=stylesheet]').length, 1, 'the fake window has one <link> stylesheet element');
+
+  const fakeWindowLink = fakeWindow.document.head.querySelectorAll('link[rel=stylesheet]')[0];
+
+  assert.strictEqual(fakeWindowLink.type, link.type, 'the <style> type attribute in the fake window is the one from <link> element');
+  assert.strictEqual(fakeWindowLink.href, link.href, 'the <style> href attribute in the fake window is the one from <link> element');
+  assert.strictEqual(fakeWindowLink.media, link.media, 'the <style> media attribute in the fake window is the one from <link> element');
```
mister-ben, line 732: Then the tests can be run after the link stylesheet exists in `document.styleSheets`

```suggestion
  const containsRulesFromStyle = (el) => {
    return Boolean([...fakeWindow.document.head.querySelectorAll('style')].find(s => {
      return s.textContent.includes(el.textContent);
    }));
  };

  link.onload = link.onerror = () => {
    Dom.copyStyleSheetsToWindow(fakeWindow);

    assert.strictEqual(fakeWindow.document.head.querySelectorAll('style, link').length, document.styleSheets.length, 'the fake window has as many <style> or <link> elements as document.styleSheets');

    assert.true(containsRulesFromStyle(style1), 'a <style> in the fake window contains content from first <style> element');
    assert.true(containsRulesFromStyle(style2), 'a <style> in the fake window contains content from second <style> element');

    assert.strictEqual(fakeWindow.document.head.querySelectorAll('link[rel=stylesheet]').length, 1, 'the fake window has one <link> stylesheet element');

    const fakeWindowLink = fakeWindow.document.head.querySelectorAll('link[rel=stylesheet]')[0];

    assert.strictEqual(fakeWindowLink.type, link.type, 'the <style> type attribute in the fake window is the one from <link> element');
    assert.strictEqual(fakeWindowLink.href, link.href, 'the <style> href attribute in the fake window is the one from <link> element');
    assert.strictEqual(fakeWindowLink.media, link.media, 'the <style> media attribute in the fake window is the one from <link> element');

    done();
  };

  document.head.appendChild(link);
```

mister-ben:
We'll need to do something with this as Chrome Canary already remove `copyStylesheets`. @beaufortfrancois any comments re the suggested changes to make sure the order of stylesheets is preserved?

amtins (approved):

PR #8320, opened 2023-06-13
fix(volume): vertical slider causes viewport overflow in rtl layout

### LINKED ISSUE #7743: videojs default skin with vertical volumeslider overflows viewport in rtl layout
## Description
The viewport overflows because of the use of `left:-3000em` within a rtl html layout

## Steps to reproduce
Using the default skin. Set:
```
	controlBar: {
		volumePanel: {
			vertical: true,
			inline: false
		}
	}
```
Open a videojs player within a `<html dir="rtl">`-layout

### Expected
The viewport should not horizontally overflow

### Actual
The viewport overflows on the left to -3000 em, as the overflow of viewport in rtl is always on the left.

This is a common problem with trying to push text offscreen. See also:
* https://www.zeldman.com/2012/03/01/replacing-the-9999px-hack-new-image-replacement/
* https://snook.ca/archives/html_and_css/hiding-content-for-accessibility

Screenshot 2022-05-02 at 20 58 55

A quick way around this is using top:-negative (though might still have performance problems) or to clip the surface to invisible.

## Additional Information
#### videojs
#### browsers
what browser are affected?
#### OSes
### plugins

amtins:
## Description

This PR fixes #7743 where a scroll bar appears when the text direction is RTL.

[Screencast from 13. 06. 23 19:02:52.webm](https://github.com/videojs/video.js/assets/34163393/85343a12-8193-487a-8f35-39c19f3f0b0d)

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_volume.scss
+++ b/src/css/components/_volume.scss
@@ -51,7 +51,11 @@
     position: relative;
 
     &.vjs-volume-vertical {
-      left: -3.5em;
+      height: 8em;
+      width: 3em;
+
+      pointer-events: initial;
+
       @include transition(left 0s);
     }
     $transition-property: visibility 0.1s, opacity 0.1s, height 0.1s, width 0.1s, left 0s, top 0s;
@@ -75,10 +79,6 @@
 }
 
 .video-js .vjs-volume-panel .vjs-volume-control.vjs-volume-vertical {
-  height: 8em;
-  width: 3em;
-  left: -3000em;
-
   $transition-property: visibility 1s, opacity 1s, height 1s 1s, width 1s 1s, left 1s 1s, top 1s 1s;
   @include transition($transition-property)
 }
@@ -192,9 +192,10 @@
 }
 
 .video-js .vjs-volume-vertical {
-  width: 3em;
-  height: 8em;
+  left: -3.5em;
   bottom: 8em;
+
+  pointer-events: none;
 
   @include background-color-with-alpha($primary-background-color, $primary-background-transparency);
 }


amtins:
## Observation

The current behavior allows you to hover over the volume slider when it's disappearing to make it reappear.

[actual-behavior.webm](https://github.com/videojs/video.js/assets/34163393/1df60fb3-2da1-4139-a71d-aef6893dfa05)

With this fix, the behavior doesn't allow you to hover over the volume slider when it's disappearing to make it reappear, which is a regression.

[behavior-with-fix.webm](https://github.com/videojs/video.js/assets/34163393/831e37f4-6439-4759-a822-71727fae8fe0)

amtins:
## Observation

This is outside the scope of this fix. On touchscreen devices, a long press on the volume button displays the volume slider. Usually, on this type of device, a user would tend to use the device's physical buttons to adjust the volume.

https://github.com/videojs/video.js/blob/2b0df25df332dceaab375327887f0721ca8d21d0/src/css/components/_volume.scss#L39-L41

amtins:
I'm putting this RP up for review to get opinions and maybe find a better approach that doesn't introduce regression. Please see comment above https://github.com/videojs/video.js/pull/8320#issuecomment-1606114528

misteroneill (approved):

misteroneill:
> ## Observation
> This is outside the scope of this fix. On touchscreen devices, a long press on the volume button displays the volume slider. Usually, on this type of device, a user would tend to use the device's physical buttons to adjust the volume.
> 
> https://github.com/videojs/video.js/blob/2b0df25df332dceaab375327887f0721ca8d21d0/src/css/components/_volume.scss#L39-L41

Yes, my expectation would be that on a phone/tablet the user would use the physical buttons for volume instead of the web UI.

amtins:
> Yes, my expectation would be that on a phone/tablet the user would use the physical buttons for volume instead of the web UI.

Thank you for the feedback. I'll try to create a PR to solve this problem this week-end.

PR #8332, opened 2023-06-26
fix(seek-bar): don't update when user is inactive (#8331)

### LINKED ISSUE #8331: High CPU usage due to the seek bar even when it's hidden
### Description

I'm seeing high CPU usage from video.js due to the seek bar being updated even when it's hidden.

The high CPU usage comes from the need to:
- handle events
- handle intervals
- update the DOM
- reflow

I'm using video.js through Invidious (on a PinePhone Pro).

I'll put a public invidious instance in the URL, but I'm using a private invidious instance.

A PR is coming.

### Reduced test case

https://yewtu.be/

### Steps to reproduce

1. Play an audio or a video on Invidious
2. wait for the seek-bar hide itself (due to "user inactivity")
3. notice the CPU usage is still higher than it could. On Firefox, you can also capture a performance profile to see that most of the time is spent on updating the seek bar.

### Errors

None

master

### Video.js plugins used.

_No response_

Firefox 114.0.2

Linux

raphj:
## Description

Fixes #8331.

## Specific Changes proposed

In the seek-bar,
- block updates when the user is considered inactive by the player. When this is the case, the seek-bar is hidden
- listen to the `timeupdate` event only when the user is considered active by the player
- listen to the useractive and userinactive events from the player to do this
- extend toggleVisibility_ to it also handles user activity, in addition to the document visibility. I didn't change the name to limit the diff, and also because I think "visibility" is suitable. Only, it does not qualify only the document anymore, but also the seek-bar

## Requirements Checklist
  - [ ] Unit Tests updated or fixed - *no test updated*
  - [ ] Docs/guides updated - *no docs / guides updated*
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0)) - *no docs / guides updated*

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -52,7 +52,7 @@ class SeekBar extends Slider {
     this.update_ = Fn.bind_(this, this.update);
     this.update = Fn.throttle(this.update_, Fn.UPDATE_REFRESH_INTERVAL);
 
-    this.on(this.player_, ['ended', 'durationchange', 'timeupdate'], this.update);
+    this.on(this.player_, ['ended', 'durationchange'], this.update);
     if (this.player_.liveTracker) {
       this.on(this.player_.liveTracker, 'liveedgechange', this.update);
     }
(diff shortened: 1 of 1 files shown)


raphj:
The fix also applies cleanly and works well on Invidious on version v7.12.4

gkatsev:
Unfortunately, this removes an accessibility need. If a screen reader user is focused on the seekbar, the expectation is that the seekbar continues announcing time changing. This can happen even after the user has become “inactive”. There is no way to know whether a screen reader is focused on the seek bar or not.

raphj:
Indeed. Would it be a solution to:
1. only update aria attributes to avoid reflow?
2. if not, only update aria attributes and text content, but not size (width) of elements? I don't know if it would cause reflow, but it's probably less CPU intensive to not update the widths.

Could it be a setting (disabled by default, as we probably want accessibility by default)?

gkatsev:
Not updating the width in this case is probably fine, but probably need an option. Some designs keep the seek bar visible even when the user is inactive.

raphj:
Okay, I'll think about what I can propose. Thanks for the quick replies :-)

I might not have time to work on this in the coming days, so if someone wants to take over, please do.

As a user of an app that uses video.js and who does not use a screen reader, but has a device that gets warm easily, I'd be interested in a solution that runs as little code as possible / necessary.

I have a temporary solution that works for me because I can just drop my patched video.js on my private Invidious instance and the Invidious project are not currently updating video.js (they seem stuck on version 7.12). It'd be nice if it could benefit others as well, hence the PR :-)

amtins:
In addition to what @gkatsev said, this also breaks the `audioOnlyMode`.

raphj:
Marking as draft so this does not get merged by mistake.

raphj:
I'm going to close this, it's broken and I'm unlikely going to find time to fix this properly.

PR #8356, opened 2023-07-12
fix(tests): Skip a test on old Safari

mister-ben:
## Description
Safari versions less than 14 have a broken `document.styleSheets` - not all style elements are included, which breaks a recent test. We're currently running tests on Brpwserstack on Safari 12 on main.

## Specific Changes proposed
Since we use the `copyStyleSheets` function only in browsers that support documentPictureInPicture, this simply skips the test on Safari < 14. Should Safari implement documentPictureInPicture in a future update this isn't a concern as `document.styleSheets` behaves properly on 14+.

## Requirements Checklist

### DIFF
--- a/test/unit/utils/dom.test.js
+++ b/test/unit/utils/dom.test.js
@@ -1,7 +1,9 @@
 /* eslint-env qunit */
 import document from 'global/document';
+import window from 'global/window';
 import sinon from 'sinon';
 import * as Dom from '../../../src/js/utils/dom.js';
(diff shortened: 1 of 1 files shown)


amtins (approved):
LGTM! 

@mister-ben would there be a way to run all the tests when a PR has reached its review quota? This would make it possible to see that some tests don't pass before reaching main.

mister-ben:
@amtins agree we need to have PR tests running on browserstack

PR #8360, opened 2023-07-15
Refactor utils

ahmed0saber:


src/js/utils/buffer.js
```diff
@@ -30,13 +28,11 @@ export function bufferedPercent(buffered, duration) {
...
-    start = buffered.start(i);
-    end = buffered.end(i);
+    const start = buffered.start(i);
+    const end = buffered.end(i);
 
     // buffered end can be bigger than duration by a very small fraction
-    if (end > duration) {
-      end = duration;
-    }
+    end = Math.min(end, duration);
```
mister-ben, line 35: This tries to redefine a const

src/js/utils/fn.js
```diff
@@ -60,19 +60,15 @@ export const bind_ = function(context, fn, uid) {
  *
  * @return   {Function}
  */
-export const throttle = function(fn, wait) {
-  let last = window.performance.now();
```
mister-ben, line 64: This was changed to use `window.performance.now()` in #5870. Why do you propose changing it to use `Date.now()`?

src/js/utils/buffer.js
ahmed0saber: That's unbelievable mistake, I've fixed it.

src/js/utils/fn.js
ahmed0saber: That's a good question, we use `window.performance.now()` when we need a high level of accuracy as it is a measurement of floating point milliseconds, but do these `0.153184` milliseconds really matter in our situation? I don't think they matters because we are using 30 milliseconds as a delay for the throttle which is so huge compared to `% 1`, so we can simply neglect these neglectable numbers which have bad small effect on the performance, as `window.performance.now()` is `3 to 4` times slower than `Date.now()`, as shown in the following picture:
Screenshot 2023-08-11 162527
Finally, I have initialized `lastCallTime` with 0, to avoid calling any function that requires a non-zero milliseconds to run.

gkatsev: The main reason to use performance.now() here isn't the higher resolution but the guarantee of monotonic time updates. I would be inclined to leave it as is due to that.

I'm not sure that the performance difference between Date.now() and performance.now() in this usage is applicable.

src/js/utils/time.js
```diff
@@ -141,7 +141,7 @@ function createTimeRangesObj(ranges) {
 export function createTimeRanges(start, end) {
   if (Array.isArray(start)) {
     return createTimeRangesObj(start);
-  } else if (start === undefined || end === undefined) {
+  } if (start === undefined || end === undefined) {
```
gkatsev, line 144:
```suggestion
  }
  if (start === undefined || end === undefined) {
```

src/js/utils/fn.js
thijstriemstra: Perhaps the method used to determine the start time should be configurable, I can imagine `performance.now()` being superceded by some other better native JS alternative at some point (it never stops, does it) so a method that allows the user to use `Date.now()` or `performance.now()` or even `nextnewthing.now()`, is a reasonable idea, one I should've thought of when I submitted #5870

PR #8368, opened 2023-07-19
fix(tests): Fixes for old Safari

mister-ben:
## Description
Followup to #8356. The issue with older Safari with `document.styleSheets` is a little different than thought. The problem is that Safari doesn't necessarily return rules verbatim, so a text comparison wasn't working. A rule with a shorthand property like `background: white;` is returned as `background-color: white;`. 

## Specific Changes proposed
Changes the test to use a rule that won't be changed.
Revert skipping the test on Safari from #8356 as we can make that test work after all.
Change copyStyleSheets to use `styleSheet.media.mediaText` when setting media. On old Safari this has to be a string, other browsers allow the MediaList object to be set.

Test ran on browserstack passed here:
https://github.com/mister-ben/video.js/actions/runs/5600546784/jobs/10243092934

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -878,7 +878,8 @@ export function copyStyleSheetsToWindow(win) {
 
       link.rel = 'stylesheet';
       link.type = styleSheet.type;
-      link.media = styleSheet.media;
+      // For older Safari this has to be the string; on other browsers setting the MediaList works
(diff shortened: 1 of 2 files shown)


wseymour15 (approved):
+1, good comment describing the issue with Safari!

misteroneill (approved):
Nice, was just looking at these test failures.

PR #8370, opened 2023-07-19
fix: Check for VTTCue

wseymour15:
This was causing a bug when it came to VTT Cue positioning. This check was never true becasue the `window` instances of the VTTCue being checked, and the `window` instance here are never the same.

## Specific Changes proposed
A different way to check if the property being used is indeed a VTTCue.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15300,6 +15300,14 @@
           "requires": {
             "global": "^4.4.0"
           }
+        },
+        "videojs-vtt.js": {
+          "version": "0.15.4",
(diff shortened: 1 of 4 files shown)


misteroneill (approved):
LGTM, but converted to a draft in case it's still blocked. Feel free to mark it ready if that's erroneous.

wseymour15:
> we do not need to update the version of VHS here, since technically VHS depends on Video.js in this instance.

Thanks @misteroneill! In this case, we actually have all the dependencies we need, as VHS actually depends on the changes here. We will need to update the video.js version in VHS once this is deployed. I'll reopen this!

PR #8382, opened 2023-08-01
fix(svg-icons): default icons color

amtins:
## Description

Uses the same color as defined by the `color` property of the `video-js` class to apply to `svg icons` for easy customization.

```css
.video-js {
  color: red; /* Also redefines svg icons color  */
}
```

## Specific Changes proposed

- set `fill` property to `currentColor`

## Requirements Checklist
  - [ ] Docs/guides updated


src/css/_icons.scss
```diff
@@ -11,7 +11,7 @@
   background-repeat: no-repeat;
   background-position: center;
 
-  fill: #FFFFFF;
+  fill: currentColor;
```
wseymour15, line 14: I could be wrong here and I haven't tested, but should this be [currentcolor](https://www.w3schools.com/colors/colors_currentcolor.asp)?

gkatsev: Looks like it https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword
though, I think it's case-insensitive

amtins: @wseymour15, @gkatsev thanks for the review! 

Would you like me to adjust the value ? If it helps to make a decision, here's what the W3C says https://www.w3.org/TR/css-color-3/#currentcolor

wseymour15: It's always fun when docs conflict 😄 You can leave as is. No strong feelings after reading that.

gkatsev: yeah, interesting that mdn doesn't document it like the spec. I would always err on the side of the spec, if possible.

dzianis-dashkevich (approved):

PR #8386, opened 2023-08-03
test: add tests for chrome v53

wseymour15:
## Description
This, along with the change in the karma config generator, adds an additional test suite for Chrome v53. We need this to ensure we test a viable version of chrome for SMART TVs.

## Specific Changes proposed
- Logic to use the older version of sinon when testing with Chrome v53
- Adding that older sinon package to the test modules with an alias.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
     "shelljs": "^0.8.5",
     "shx": "^0.3.2",
     "sinon": "^11.1.1",
+    "sinon8": "npm:sinon@^8",
     "typescript": "^4.8.4",
     "uglify-js": "^3.6.0",
     "unified": "^7.0.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -162,6 +162,7 @@
     "shelljs": "^0.8.5",
     "shx": "^0.3.2",
     "sinon": "^11.1.1",
+    "sinon8": "npm:sinon@^8",
     "typescript": "^4.8.4",
     "uglify-js": "^3.6.0",
     "unified": "^7.0.2",
```
wseymour15, line 168: Note that we need to update the `videojs-generate-karma-config` version before this will work.

wseymour15:
The plan is to reopen this PR once the Brightcove team has the time to finish up this work.

PR #8395, opened 2023-08-11
fix(text): caption settings typo

usmanonazim:
## Description
In caption settings, under 'Text Edge Style', the 'Dropshadow' option should be spelled 'Drop shadow'

## Specific Changes proposed
Changed 'Dropshadow' in caption settings to 'Drop shadow', and updated all language files.

## Requirements Checklist
  - [ ] Docs/guides updated


misteroneill (approved):
In general, I would say we should avoid changing the keys of `lang` files without changing the values. But in this case, because it's a pretty minor difference, I think it's okay. Many of the translations include a space, even.

PR #8435, opened 2023-09-10
fix(control-bar): incorrect display when control bar display is locked

amtins:
## Description

The control bar is not displayed correctly when the display is locked because `.vjs-lock-showing` uses `display` `block` instead of `flex`.

Screenshot from 2023-09-10 16-03-40

## Specific Changes proposed

- use `display` `flex` instead of `display` `block` when the control bar has `.vjs-lock-showing` class
- ensure that the control bar is not displayed if the `player` has classes:
  - `.vjs-controls-disabled`
  - `.vjs-using-native-controls`
  - `.vjs-error`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_control-bar.scss
+++ b/src/css/components/_control-bar.scss
@@ -10,6 +10,14 @@
   @include background-color-with-alpha($primary-background-color, $primary-background-transparency);
 }
 
+// Locks the display only if:
+//  - controls are not disabled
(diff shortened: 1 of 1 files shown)


mister-ben (approved):
I think lock showing had only really been considered in terms of menu buttons before. Good to implement it properly here too.

Issue #8425, opened 2023-08-31
Caption zoom bug in min js


j77h:
### Video JS versions tested

8.5.2
8.5.3
7.14.3 from the wordpress plugin

Faulty behaviour happens only with **minified** js of **version 8**.

### Browser versions tested

Recent Chromium and Firefox, on Manjaro Linux.
(updated within the last two weeks)

### Steps to reproduce

Put video, html, vtt files in one folder, with **video.min.js** and (video-js.css or video-js.min.css).

This is my video element, but I think most of the settings are not relevant,
as I first observed the behaviour before I added most of them.
```html
  <video id="videojs1" class="video-js" muted playsinline controls preload="metadata" 
  		width="768" height="432" data-setup='{"persistTextTrackSettings": true, "preferFullWindow": true}'>
    <source src="./test.mp4" type="video/mp4">
    <track kind="subtitles" src="./test.vtt" srclang="en" label="English" default>
  </video>
```

The behaviour is the same with remote web server or local,
so for testing just use `python -m http.server` locally,
and in a browser open `http://localhost:8000/test.html`

Play the video, pause **while a caption is visible**, 
and go into the caption settings dialog.

Select a font size that's not 100%, observe the change in caption size.
No need to press Reset or Done button.
Select another font size and observe the change in caption size.

### Observed behaviour

Selected 150%, and caption size increased as expected;
then selected 125%, and caption size increased a further 125% (to nearly 200%).

Pressed 'Done' and played the video; the caption size remained near 200%.
At the end of the video, pressed 'Replay'; the caption size got another large increase.

Reloaded the page, played the video: caption size was at 125%, as intended.

What seems to be happening is that 
any change of caption zoom is relative to the **current** size;
if you alternate selecting 50% and 75%, the captions become progressively smaller.

If you edit the minified js file and make the default size 150%,
then in the caption settings click 'Reset' repeatedly,
the caption size increases by 150% on each click.

### Expected behaviour

The minified js file should produce the same behaviour as the readable js file, which is that 
changes to the caption zoom level are always relative to the pre-determined 100% size.

mister-ben:
That's strange. I can replicate what you're seeing.

psiservices-ben-voytko:
I also seee this behavior on minified version of videojs 7.21.5 with videojs-vtt 0.15.5.

mister-ben:
[...]

PR #8442, opened 2023-09-19
fix: Resolves captions sizing issue when minified

mister-ben:
## Description
Fixes issue with sizing captions in #8425.

## Specific Changes proposed
Changes test to check for presence of `getCueAsHTML` instead of the constructor name, as the constructor name is changed when minified.
Open to suggested for a test for this, since the issue only manifested when minified. 

## Requirements Checklist


wseymour15 (approved):
Looks good! Thanks for updating this.

PR #8444, opened 2023-09-20
Enhanced logger

dzianis-dashkevich:
## Description
- Added the `createNewLogger` method, to create a logger without a naming chain.
- Added optional custom `delimiter` and custom `styles`.
- Several improvements in jsDoc for proper types(d.ts) generation.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/utils/create-logger.js
+++ b/src/js/utils/create-logger.js
@@ -24,6 +24,8 @@ const LogByTypeFactory = (name, log, styles) => (type, level, args) => {
   const lvl = log.levels[level];
   const lvlRegExp = new RegExp(`^(${lvl})$`);
 
+  let resultName = name;
+
   if (type !== 'log') {
 
     // Add the type to the front of the message when it's not "log".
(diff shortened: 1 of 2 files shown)


misteroneill (approved):
LGTM

src/js/utils/create-logger.js
```diff
@@ -99,22 +109,47 @@ export default function createLogger(name) {
...
+   * @param {string} newName
+   *        The name for the new logger
+   * @param {string} [newDelimiter]
+   *        Optional delimiter
+   * @param {string} [newStyles]
+   *        Optional styles
    * @return {Object}
    */
-  log.createLogger = (subname) => createLogger(name + ': ' + subname);
+  log.createNewLogger = (newName, newDelimiter, newStyles) => {
```
gkatsev, line 150: An alternative to creating a new function is to switch to an options object. If it's a string use the old method signature, if it's an object use the new one.

dzianis-dashkevich: That was my initial implementation, and I realized that a separate function looks cleaner... 

Ideally, we should re-name `createLogger` to `createSubLogger`, but this is a breaking change, so I came up with this `newLogger` naming.

Issue #8393, opened 2023-08-09
Player controls not keyboard accessible with JAWS 2022 using Firefox


robfentress:
### Description

In Windows 11, when using JAWS 2022 with the latest version of Firefox, attempting to activate any of the player controls by pressing the <kbd>Enter</kbd> or <kbd>Space Bar</kbd> keys while they have focus does not trigger the expected action.

### Reduced test case

https://codepen.io/gkatsev/full/GwZegv

### Steps to reproduce

1. Turn on JAWS Version 2022.2307.1 ILM
2. Open Firefox 116.0.2 (64-Bit)
3. Press <kbd>Alt</kbd>+<kbd>D</kbd> to set focus to address bar.
4. Type "https://codepen.io/gkatsev/full/GwZegv" in the address bar and press the <kbd>Enter</kbd> key.
5. Press the <kbd>Tab</kbd> key until the **Play Video** button has focus. JAWS says "To activate, press Enter."
6. Press the <kbd>Enter</kbd> key, which begins playing the video and sets focus to the **Pause** button.  JAWS says "To activate, press Enter."
7. Press the <kbd>Enter</kbd> key, which should pause the video, but does not.
8. If properly coded, pressing the <kbd>Space Bar</kbd> key should also activate the button and pause the the video.  Press the <kbd>Space Bar</kbd> key.  The video does not pause.
9. Press the <kbd>Tab</kbd> key to move focus to the **Mute** button.
10. Press the <kbd>Enter</kbd> key.  This should mute the video, but does not.
11. Press the <kbd>Space Bar</kbd> key.  This should mute the video, but does not.
12. Attempt to activate any of the other player controls using the keyboard.  They will not work.

### Errors

None of the player controls can be activated using the standard keys for doing so.

### What version of Video.js are you using?

8.3.0

### What browser(s) including version(s) does this occur with?

Firefox 116.0.2 (64-Bit)

### What OS(es) and version(s) does this occur with?

Windows 11 Enterprise Version 22H2

robfentress:
I encountered similar issues with the NVDA screen reader and Firefox.  With NVDA, pressing <kbd>Insert</kbd>+<kbd>Space Bar</kbd> to switch modes allowed me to activate the controls using standard key commands, but this should not be necessary.  See the behavior of Able Player for reference.

Essk:
[...]

PR #8450, opened 2023-09-28
fix: throttle clickable component

Essk:
## Description
See [#8393](https://github.com/videojs/video.js/issues/8393)

We need to throttle click events on "toggle" UI components or interactions from screen readers in the firefox browser will toggle the control on and off instantly, effectively causing the control to do nothing.

## Specific Changes proposed
These changes add the ability to specify a `throttle` option to any component that extends `ClickableComponent`. The value can be `true` (which will set a default wait time of 50ms on the throttle) or a number >= 1, which will be passed as a custom wait time in ms. 

Toggle components in the standard UI have been updated to have the option `{ toggle: true}`.

### Further thoughts
Given that the need for a throttle is governed by external factors that we cannot predict,  I think it may be beneficial to throttle _all_ clickable UI by default with an option to turn throttling  _off_. However in practice, as things stand, this breaks a lot of videojs tests which were never written with throttled components in mind. Those internal tests could be updated* but users that have implemented custom components with their own testing are likely to suddenly find their tests failing and I think this would be considered a breaking change. 

* Updating the tests is not straightforward because as the current design of imports from `utils/fn` prevents stubbing the throttle utility that is imported into component classes. 

Perhaps these considerations could be taken into account for VJS 9?

## Requirements Checklist
     - Will verify when I next have access to a Windows environment
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -5,6 +5,7 @@ import Component from './component';
 import * as Dom from './utils/dom.js';
 import log from './utils/log.js';
 import keycode from 'keycode';
+import { throttle } from './utils/fn';
 
 /**
  * Component which is clickable or keyboard actionable, but is not a
@@ -41,9 +42,11 @@ class ClickableComponent extends Component {
       this.controlText(this.options_.controlText);
     }
 
+    const throttledClick = throttle(this.handleClick.bind(this), 50);
+
     this.handleMouseOver_ = (e) => this.handleMouseOver(e);
     this.handleMouseOut_ = (e) => this.handleMouseOut(e);
-    this.handleClick_ = (e) => this.handleClick(e);
+    this.handleClick_ = (e) => throttledClick(e);
     this.handleKeyDown_ = (e) => this.handleKeyDown(e);
 
     this.emitTapEvents();
--- a/src/js/clickable-component.js
+++ b/src/js/clickable-component.js
@@ -34,7 +34,7 @@ class ClickableComponent extends Component {
    *         A class or space separated list of classes to add the component
    *
    * @param  {number | boolean} [options.throttle]
-   *         A throttle will be applied to the clickHander if the number is >= 1 or the value is `true`
+   *         A throttle will be applied to the clickHandler if the number is >= 1 or the value is `true`
    *         A number specifies the desired wait time in ms or a default wait of 50ms will be applied
    *
    */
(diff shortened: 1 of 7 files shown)


src/js/clickable-component.js
```diff
@@ -32,6 +33,10 @@ class ClickableComponent extends Component {
    * @param  {string} [options.className]
    *         A class or space separated list of classes to add the component
    *
+   * @param  {number | boolean} [options.throttle]
+   *         A throttle will be applied to the clickHander if the number is >= 1 or the value is `true`
```
mister-ben, line 37:
```suggestion
   *         A throttle will be applied to the clickHandler if the number is >= 1 or the value is `true`
```

src/js/clickable-component.js
```diff
@@ -41,9 +46,20 @@ class ClickableComponent extends Component {
       this.controlText(this.options_.controlText);
     }
 
+    const selectClickHandler = () => {
+      if (typeof this.options_.throttle === 'number' || this.options_.throttle === true) {
+        const wait = typeof this.options_.throttle === 'number' ? parseInt(this.options_.throttle, 10) : 50;
```
gkatsev, line 51: Also, should this be `parseFloat` instead of `parseInt`?

src/js/clickable-component.js
```diff
@@ -41,9 +46,20 @@ class ClickableComponent extends Component {
...
+      }
+      return this.handleClick.bind(this);
+    };
+
+    const selectedClickHander = selectClickHandler();
+
     this.handleMouseOver_ = (e) => this.handleMouseOver(e);
     this.handleMouseOut_ = (e) => this.handleMouseOut(e);
-    this.handleClick_ = (e) => this.handleClick(e);
+    this.handleClick_ = (e) => selectedClickHander(e);
```
gkatsev, line 62:
```suggestion
    this.handleClick_ = (e) => selectedClickHandler(e);
```

src/js/clickable-component.js
```diff
@@ -32,6 +33,10 @@ class ClickableComponent extends Component {
    * @param  {string} [options.className]
    *         A class or space separated list of classes to add the component
    *
+   * @param  {number | boolean} [options.throttle]
+   *         A throttle will be applied to the clickHander if the number is >= 1 or the value is `true`
+   *         A number specifies the desired wait time in ms or a default wait of 50ms will be applied
```
gkatsev, line 38: any details on how 50 was chosen as the default?

src/js/clickable-component.js
```diff
@@ -41,9 +46,20 @@ class ClickableComponent extends Component {
...
+        return throttle(this.handleClick.bind(this), wait);
+      }
+      return this.handleClick.bind(this);
+    };
+
+    const selectedClickHander = selectClickHandler();
+
     this.handleMouseOver_ = (e) => this.handleMouseOver(e);
     this.handleMouseOut_ = (e) => this.handleMouseOut(e);
-    this.handleClick_ = (e) => this.handleClick(e);
```
gkatsev, line 46: since we still want a bound handleClick inside of selectedClickHandler, should we keep this around somewhere and use that rather than rebinding inside selectedClickHandler?

gkatsev (commented):
Some minor points.

src/js/clickable-component.js
Essk: I figured since throttle's key condition is `(now - last >= wait) ` where `now - last` will always be a whole number of milliseconds, `parseInt` was most appropriate. 

Looking with fresh eyes, maybe the `parseX(arg-that-might-be-a-number`) should be removed entirely as I really just put it in out of habit and I see `throttle` itself doesn't enforce anything.

src/js/clickable-component.js
Essk: Nothing empirical, just a guess on my part based on previous similar solutions I've worked on. However, this is a good point so I measured the gap on my fairly modern Windows PC, and got 1-2 ms. Sadlt FF doesn't have the same performance spoofing options as Chrome (which doen't receive double input)  but maybe I can get more data from Browserstack or similar.

Essk:
Thanks everyone for the feedback - sorry about all the typos! 😳

src/js/clickable-component.js
Essk: My bad -  I had hoped Browserstack might have options to set hardware limitations, but it does not.

PR #8451, opened 2023-09-30
fix(types): use typeof for registerComponent and registerPlugin

mister-ben:
## Description
These accept and return the component/plugin class rather than an instance of the class. Currently, typescript will throw  `TS2345: Argument of type 'typeof MyComponent' is not assignable to parameter of type 'Component'.` when registering a component.

## Specific Changes proposed
jsdoc updates

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -582,7 +582,6 @@ class Component {
   /**
    * Add a child `Component` inside the current `Component`.
    *
-   *
    * @param {string|Component} child
    *        The name or instance of a child to add.
    *
@@ -593,7 +592,8 @@ class Component {
    * @param {number} [index=this.children_.length]
    *        The index to attempt to add a child into.
    *
-   * @return {typeof Component}
+   *
+   * @return {Component}
    *         The `Component` that gets added as a child. When using a string the
    *         `Component` will get created by this process.
    */
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -593,7 +593,7 @@ class Component {
    * @param {number} [index=this.children_.length]
    *        The index to attempt to add a child into.
    *
-   * @return {Component}
+   * @return {typeof Component}
    *         The `Component` that gets added as a child. When using a string the
    *         `Component` will get created by this process.
    */
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -286,8 +286,8 @@ const DEFAULT_BREAKPOINTS = {
  *
  * After an instance has been created it can be accessed globally in three ways:
  * 1. By calling `videojs.getPlayer('example_video_1');`
- * 2. By calling `videojs('example_video_1');` (not recomended)
- * 2. By using it directly via  `videojs.players.example_video_1;`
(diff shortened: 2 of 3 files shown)


src/js/component.js
```diff
@@ -593,6 +592,7 @@ class Component {
    * @param {number} [index=this.children_.length]
    *        The index to attempt to add a child into.
    *
+   *
    * @return {Component}
    *         The `Component` that gets added as a child. When using a string the
    *         `Component` will get created by this process.
```
amtins, line 598: Question, would it be a good idea to do something like:
```javascript
  /**
   * Add a child `Component` inside the current `Component`.
   *
   * @template {Component} T
   *
   * @param {string|T} child
   *        The name or instance of a child to add.
   *
   * @param {Object} [options={}]
   *        The key/value store of options that will get passed to children of
   *        the child.
   *
   * @param {number} [index=this.children_.length]
   *        The index to attempt to add a child into.
   *
   * @return {T}
   *         The `Component` that gets added as a child. When using a string the
   *         `Component` will get created by this process.
   */
```

This generates:
```javascript
    /**
     * Add a child `Component` inside the current `Component`.
     *
     * @template {Component} T
     *
     * @param {string|T} child
     *        The name or instance of a child to add.
     *
     * @param {Object} [options={}]
     *        The key/value store of options that will get passed to children of
     *        the child.
     *
     * @param {number} [index=this.children_.length]
     *        The index to attempt to add a child into.
     *
     * @return {T}
     *         The `Component` that gets added as a child. When using a string the
     *         `Component` will get created by this process.
     */
    addChild<T extends Component>(child: string | T, options?: any, index?: number): T;
```

The result is:
Screenshot from 2023-10-01 13-24-56

Instead of:
Screenshot from 2023-10-01 13-30-43

amtins (approved):
LGTM! 👍🏽

src/js/component.js
mister-ben: Thanks, I was experimenting along those lines for `addChild` but hadn't got it quite working yet. The trouble is jsdoc won't understand it, so we will need another jsdoc plugin to transform this.

amtins: I've just tested the following plugin [jsdoc-plugin-typescript](https://www.npmjs.com/package/jsdoc-plugin-typescript), but I didn't understand what result I should expect 😅

mister-ben: We're using [jsdoc-tsimport-plugin](https://www.npmjs.com/package/jsdoc-tsimport-plugin) now which is similar, it doesn't look like either handles that template syntax. Let's keep the `addChild` update, which is needed, in a separate PR because it will be more impactful than this one if it needs tooling changes.

mister-ben: This also doesn't cover the case where the component is registered, and added by name

```js
videojs.registerComponent('ComponentX', X);
const x = player.addChild('ComponentX');
x.__y();
// Property '__y' does not exist on type 'Component'.ts(2339)
```

PR #8462, opened 2023-10-12
chore: update VHS and mux.js versions

wseymour15:
## Description
update VHS and mux.js versions

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.12.5",
-    "@videojs/http-streaming": "3.6.0",
+    "@videojs/http-streaming": "3.7.0",
     "@videojs/vhs-utils": "^4.0.0",
     "@videojs/xhr": "2.6.0",
     "aes-decrypter": "^4.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -86,15 +86,15 @@
...
-    "@videojs/http-streaming": "3.6.0",
+    "@videojs/http-streaming": "3.7.0",
     "@videojs/vhs-utils": "^4.0.0",
     "@videojs/xhr": "2.6.0",
     "aes-decrypter": "^4.0.1",
     "global": "4.4.0",
     "keycode": "2.2.0",
     "m3u8-parser": "^6.0.0",
     "mpd-parser": "^1.0.1",
-    "mux.js": "^6.2.0",
```
dzianis-dashkevich, line 97: Do we need it?

adrums86 (approved):
This LGTM, but I second @dzianis-dashkevich's question. Do we need the mux.js dependency here at all? I think maybe we should re-evaluate that in a separate issue/PR.

dzianis-dashkevich (approved):

package.json
wseymour15: Unsure, we need to determine if this is actually necessary, but non-blocking for now

PR #8466, opened 2023-10-12
fix(types): Minor fix for types

andreifilip123:
## Description
Hello! We're using videojs but since the upgrade to v8, we've ran into some typing issues that we had to patch with @ts-ignore and such.
This PR fixes the issues we had with the .off() function always requiring a second param for function and the issues we have with the videojs.getComponent('Component') not returning the right type.
I'm not sure how to build from the jsdoc into the .d.ts files to check that the generated types are correct but I think these changes might work.
I'd also like to help fixing the missing ".remoteTextTracks()" function on the Player type and the missing export for video player options but I don't know how. If you give me some pointers, I'd be glad to help with this

## Specific Changes proposed
- remove the fn argument as required on the off function
- change the returned type of the getComponent function

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -160,7 +160,7 @@ class Component {
    * @param {string|string[]} type
    *        An event name or an array of event names.
    *
-   * @param {Function} fn
+   * @param {Function} [fn]
    *        The function to remove.
    */
   off(type, fn) {}
@@ -1843,7 +1843,7 @@ class Component {
    * @param {string} name
    *        The Name of the component to get.
    *
-   * @return {Component}
+   * @return {typeof Component}
    *         The `Component` that got registered under the given name.
    */
   static getComponent(name) {
--- a/src/js/component.js
+++ b/src/js/component.js
@@ -161,7 +161,7 @@ class Component {
    *        An event name or an array of event names.
    *
    * @param {Function} [fn]
-   *        The function to remove.
+   *        The function to remove. If not specified, all listeners managed by Video.js will be removed.
    */
   off(type, fn) {}
 


gkatsev:
While providing a function to `off` is technically optional, I would highly suggest not using it that way unless you're doing so for an object you've created. Calling off without a function param, will remove *all* handlers for that type, whether you added the handler or Video.js did it internally.

andreifilip123:
> While providing a function to `off` is technically optional, I would highly suggest not using it that way unless you're doing so for an object you've created. Calling off without a function param, will remove *all* handlers for that type, whether you added the handler or Video.js did it internally.

The place we're using this is in the react unmount hook. Therefore there can't be any other events after that. If I recall correctly, the fn argument was optional in the past. That's why I made it optional now too. If you want to keep it required I can remove it and now I understand the reason.
What about the other parts of the PR? I want to help with the other parts too if you give me some hints 😄

gkatsev:
It probably should be optional, as that is the proper method signature. Just wanting to make sure you're using it properly because otherwise, it could cause problems :)

src/js/component.js
```diff
@@ -160,7 +160,7 @@ class Component {
    * @param {string|string[]} type
    *        An event name or an array of event names.
    *
-   * @param {Function} fn
+   * @param {Function} [fn]
    *        The function to remove.
```
mister-ben, line 164:
```suggestion
   * @param {Function} [fn]
   *        The function to remove. If not specified, all listeners managed by Video.js will be removed.
```
I think the thing to do it here is to mark it as optional as it is, but warn of the consequences when omitting it.

mister-ben (approved):

mister-ben:
Re `remoteTextTracks()`, this [is documented in jsdoc](https://github.com/videojs/video.js/blob/d535e163da3f1df5d87f7f7a1a2e8ff628361afb/src/js/player.js#L5200), but [tsc ignores](https://github.com/Microsoft/TypeScript/issues/28730) it 😡. The methods are added in [tech.js](https://github.com/videojs/video.js/blob/d535e163da3f1df5d87f7f7a1a2e8ff628361afb/src/js/tech/tech.js#L1099). One way around it without a bigger refactor (but needs discussion and a separate PR) is to stub these methods in player.js with a placeholder that gets replaced instead of the `@method` jsdoc.

andreifilip123:
Thank you for looking into this.

jolson88:
> Re `remoteTextTracks()`, this [is documented in jsdoc](https://github.com/videojs/video.js/blob/d535e163da3f1df5d87f7f7a1a2e8ff628361afb/src/js/player.js#L5200), but [tsc ignores](https://github.com/Microsoft/TypeScript/issues/28730) it 😡. The methods are added [here](https://github.com/videojs/video.js/blob/d535e163da3f1df5d87f7f7a1a2e8ff628361afb/src/js/player.js#L5212-L5225). One way around it without a bigger refactor (but needs discussion and a separate PR) is to stub these methods in player.js with a placeholder that gets replaced instead of the `@method` jsdoc.

Yup, you were right. That's what was needed. I have a PR I'm polishing up now that I hope to get created today. My company just needed audioTracks and videoTracks, but I'll add the textTracks and remote track stuff too so those on this thread needing it can get it in the same PR. I'll tag this PR.

jolson88:
Just wanted to follow-up. Looks like I'll be getting this up tomorrow. I found some changes that were caused to the generated JS docs for the website that aren't good. So I need to do a little more polishing before submitting the PR.

jolson88:
> Just wanted to follow-up. Looks like I'll be getting this up tomorrow. I found some changes that were caused to the generated JS docs for the website that aren't good. So I need to do a little more polishing before submitting the PR.

Apologies for the delay. An initial PR has been submitted. This doesn't address all the issues (just a minor subset for now). I wanted to get the low-hanging fruit done first and drive some conversation before more invasive changes can be explored. I hope the PR that I submitted and linked to this helps other TypeScript devs at least unblock themselves for the next steps though.

dzianis-dashkevich (approved):

PR #8487, opened 2023-11-10
fix(player): reset CSS classes at player.reset

amtins:
## Description

Allows CSS classes to be reset when `player.reset` is called, so that the player is close to its initial state.

## Specific Changes proposed

- remove `vjs-playing`
- add `vjs-paused`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3599,6 +3599,10 @@ class Player extends Component {
     if (this.tech_) {
       this.tech_.clearTracks('text');
     }
+
+    this.removeClass('vjs-playing');
+    this.addClass('vjs-paused');
+
     this.resetCache_();
     this.poster('');
     this.loadTech_(this.options_.techOrder[0], null);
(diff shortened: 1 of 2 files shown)


src/js/player.js
```diff
@@ -3599,6 +3599,11 @@ class Player extends Component {
     if (this.tech_) {
       this.tech_.clearTracks('text');
     }
+
+    this.removeClass('vjs-playing');
+    this.removeClass('vjs-has-started');
```
gkatsev, line 3604: the others make sense but I don't think we want to remove this one. Yes, the player is reset, but it should still be in started state, could be a breaking change, potentially as well.

amtins: @gkatsev thank you for taking the time to review this PR and for the feedback. I removed `this.removeClass('vjs-has-started');`

mister-ben (approved):

PR #8488, opened 2023-11-12
feat(svg-icons): handle custom svg icons

amtins:
## Description

Replace the player's default icons with custom icons.

The `experimentalSvgIcons` option accepts a `boolean` or a `string`. If the string represents a valid svg, it will replace the default icons.

**Usage**

```javascript
import playerIcons from './asset/vjs-sprite-icons.svg';

const player = videojs('my-player', {
  experimentalSvgIcons: playerIcons,
  // ...
});
```

Or:

```javascript
const player = videojs('my-player', {
  experimentalSvgIcons: `
  <svg>
  ...
  </svg>
  `,
  // ...
});
```

## Specific Changes proposed

- extract a `initSvgIcons_` function to activate svg icons
- add test cases

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -518,25 +518,7 @@ class Player extends Component {
 
     this.playbackRates(options.playbackRates);
 
-    if (options.experimentalSvgIcons) {
-      // Add SVG Sprite to the DOM
-      const parser = new window.DOMParser();
-      const parsedSVG = parser.parseFromString(icons, 'image/svg+xml');
-      const errorNode = parsedSVG.querySelector('parsererror');
-
-      if (errorNode) {
-        log.warn('Failed to load SVG Icons. Falling back to Font Icons.');
-        this.options_.experimentalSvgIcons = null;
-      } else {
-        const sprite = parsedSVG.documentElement;
-
-        sprite.style.display = 'none';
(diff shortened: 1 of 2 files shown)


wseymour15 (commented):
I think this is a great idea @amtins!

My only real concern is that this leaves the door open for human error. If the `id`s in the sprite string do not match the ones in our default sprite, we will be missing icons. I don't really view this as a problem, but I think we need to ensure that it is well documented.

On that note, I think we could add to [the options guide](https://videojs.com/guides/options/#experimentalsvgicons) with this change in functionality, as well as describing (and maybe linking) to the current sprite.

amtins:
Thanks for your feedback and for taking the time to review this PR.

> My only real concern is that this leaves the door open for human error

I couldn't agree more. That's why I created a [project](https://github.com/amtins/vjs-iconator) to reduce potential human errors (wip). However, I think this project deserves a few improvements in order to enhance the developer experience and have something more like [videojs/font](https://github.com/videojs/font), [icons.json](https://github.com/videojs/font/blob/master/icons.json) and [material icons](https://github.com/videojs/font/blob/f999fb75cd6012acc7408732b23a60560786d382/package.json#L30)

PR #8495, opened 2023-11-23
fix(i18n): new italian labels

astagi:
## Description
Fix Italian labels according to ones used by Bootstrap Italia (Italian government official library)

## Requirements Checklist
  - [ ] Docs/guides updated


mister-ben:
Grazie.
A question about "filmata" in these translations (not new in these changes). Does that convey specifically "video" or "video and/or audio" as is the intention of "media" in the English?

mister-ben:
Thanks, I mean more is it also valid if it's an audio source, that has no video.

astagi:
> Thanks, I mean more is it also valid if it's an audio source, that has no video.

oh ok! no, in that case "filmato" is not the right term. You can use "L'elemento media" (media element) in Italian to refer both video and/or only audio. I'll look for other terms and let you know, I update this PR in case.

astagi:
Found it @mister-ben ! We should use "contenuto multimediale" (media content) instead of "filmato", it's ok both for video and audio. If it's ok for you I'll update this PR.

mister-ben:
Seems good to me, thanks!

dzianis-dashkevich (approved):

Issue #8273, opened 2023-05-13
time tooltip out of bounds on the right


phloxic:
### Description

Since https://github.com/videojs/video.js/commit/c4c8fc1caf780fb2351c6f9d662c845dc1b9f0d4 progress time tooltip and mouse display time tooltip can move out of bounds on the right side of the player.

A fix was proposed in https://github.com/videojs/video.js/pull/7308. However, it was [declined](https://github.com/videojs/video.js/pull/7308#issuecomment-888622689) because it broke the fixes done by the above commit.

However, the alignment issues pointed out in https://github.com/videojs/video.js/pull/7308#issuecomment-888622689 cannot be reproduced with the current Video.js v8.x and v7.x versions after the patch is applied:
- [example vanilla v8.3.0 without patch](https://phloxic.productions/test/videojs/ttt/v8.html): progress tooltip in its rightmost position sticks out on the right
- [example v8.3.0 with patch applied](https://phloxic.productions/test/videojs/ttt/v8-fix.html): progress tooltip stays inside player

Note that I _can_ reproduce the misalignment when I apply the patch to v7.9.4 (the reason why the patch was declined at the time).

### Reduced test case

https://phloxic.productions/test/videojs/ttt/v8.html

### Steps to reproduce

1. Set up player with a wide time tooltip
2. Observe behaviour when the time tooltip is in its rightmost position

### Errors

No errors.

### What version of Video.js are you using?

v8.3.0, v7.21.4 -- introduced in 7.9.4

### Video.js plugins used.

none

### What browser(s) including version(s) does this occur with?

all

### What OS(es) and version(s) does this occur with?

all

phloxic:
Since https://github.com/videojs/video.js/commit/50f14bd66d422a0a23637fc30b2c72e616596076 the time tooltip does not out of bounds of the _control bar_ by default.
Closing.

PR #8527, opened 2023-12-14
fix: time tooltip truncated

harisha-swaminathan:
## Description
 [spaceRightOfPoint](https://github.com/videojs/video.js/blob/main/src/js/control-bar/progress-control/time-tooltip.js#L75) is always `NaN` for [mouse time display](https://github.com/videojs/video.js/blob/main/src/js/control-bar/progress-control/progress-control.js#L84) (unlike playProgressBar's time display) because the [seekbarRect](https://github.com/videojs/video.js/blob/main/src/js/control-bar/progress-control/progress-control.js#L75) ([dom.findPosition(seekbar)](https://github.com/videojs/video.js/blob/main/src/js/utils/dom.js#L503)) does not have a `right` property. 
 
 And since spaceRightOfPosition is always `NaN`, when the mouse tool tip is moved close to edge of the right side player, it's position is not adjusted as it is on the left and so it get's cut off.
 
truncated

## Specific Changes proposed
The proposed change is to ignore the gap between the right edge of the `SeekBar` and the player (`playerRect.right - seekBarRect.right`) in such cases.

fixed

## Requirements Checklist

### DIFF
--- a/src/js/control-bar/progress-control/time-tooltip.js
+++ b/src/js/control-bar/progress-control/time-tooltip.js
@@ -66,7 +66,8 @@ class TimeTooltip extends Component {
     // of the player. We calculate any gap between the left edge of the player
     // and the left edge of the `SeekBar` and add the number of pixels in the
     // `SeekBar` before hitting the `seekBarPoint`
-    const spaceLeftOfPoint = (seekBarRect.left - playerRect.left) + seekBarPointPx;
+    let spaceLeftOfPoint = (seekBarRect.left - playerRect.left) + seekBarPointPx;
+    // (seekBarRect.left - playerRect.left) + seekBarPointPx;
 
     // This is the space right of the `seekBarPoint` available within the bounds
     // of the player. We calculate the number of pixels from the `seekBarPoint`
@@ -78,10 +79,11 @@ class TimeTooltip extends Component {
     // spaceRightOfPoint is always NaN for mouse time display
     // because the seekbarRect does not have a right property. This causes
     // the mouse tool tip to be truncated when it's close to the right edge of the player.
-    // In such cases, we ignore the `playerRect.right - seekBarRect.right` value when calculating
-    // spaceRightofPoint.
+    // In such cases, we ignore the `playerRect.right - seekBarRect.right` value when calculating.
+    // For the sake of consistency, we ignore seekBarRect.left - playerRect.left for the left edge.
     if (!spaceRightOfPoint) {
       spaceRightOfPoint = seekBarRect.width - seekBarPointPx;
+      spaceLeftOfPoint = seekBarPointPx;
     }
     // This is the number of pixels by which the tooltip will need to be pulled
     // further to the right to center it over the `seekBarPoint`.
--- a/src/js/control-bar/progress-control/time-tooltip.js
+++ b/src/js/control-bar/progress-control/time-tooltip.js
@@ -75,6 +75,11 @@ class TimeTooltip extends Component {
     let spaceRightOfPoint = (seekBarRect.width - seekBarPointPx) +
       (playerRect.right - seekBarRect.right);
 
+    // spaceRightOfPoint is always NaN for mouse time display
+    // because the seekbarRect does not have a right property. This causes
+    // the mouse tool tip to be truncated when it's close to the right edge of the player.
+    // In such cases, we ignore the `playerRect.right - seekBarRect.right` value when calculating
+    // spaceRightofPoint.
     if (!spaceRightOfPoint) {
       spaceRightOfPoint = seekBarRect.width - seekBarPointPx;
     }
--- a/src/js/control-bar/progress-control/time-tooltip.js
+++ b/src/js/control-bar/progress-control/time-tooltip.js
@@ -72,9 +72,12 @@ class TimeTooltip extends Component {
     // of the player. We calculate the number of pixels from the `seekBarPoint`
     // to the right edge of the `SeekBar` and add to that any gap between the
     // right edge of the `SeekBar` and the player.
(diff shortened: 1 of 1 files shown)


phloxic:
@harisha-swaminathan - I believe the solution at https://github.com/videojs/video.js/pull/7308 is better, as I tried to explain here: https://github.com/videojs/video.js/issues/8273. The reasons why that change was reverted imo do not apply any longer.

amtins:
@harisha-swaminathan, @phloxic I'm not quite sure I understand the problem as a whole.

What I do understand (but maybe I'm wrong) is that we want to prevent the `timeTooltip` from overflowing. If that's the case, perhaps we need to change the approach to calculating the positioning of the `timeTooltip` component.

To do this, the only parameters we need are:
- seek bar width
- timeTooltip width
- The position of the timeTooltip in the seek bar

With these 3 parameters, we should be able to keep the timeTooltip inside the seek bar.

In this case, we could do something like https://github.com/videojs/video.js/pull/8530

phloxic:
@amtins - I just rebased what I am using onto main [here](https://github.com/phloxic/video.js/commit/72d4c528b9fccd23e94bc9022f46b5aadb1113d6).
Basically, `findPosition` does the correct thing for the left side, but not for the right side. `getBoundingClientRect` provides data for left and right and works for me[tm].

Demo v8.3.0 without the change [here](https://phloxic.productions/test/videojs/ttt/v8.html)
Demo v8.3.0 with the change [here](https://phloxic.productions/test/videojs/ttt/v8-fix.html)
afaics the demo with the fix (2nd example) also shows that issues leading to using `findPosition` do not exist anymore.

amtins:
@phloxic thank you for these clarifications. I'm currently going through all the issues (and there are a lot) to try and understand the initial problem.

phloxic:
> @phloxic thank you for these clarifications. I'm currently going through all the issues (and there are a lot) to try and understand the initial problem.

Certainly a good idea, see https://github.com/videojs/video.js/pull/7308#issuecomment-1858511864. I was too focused on the out of bounds on the right problem (which does still exist though), sorry.

harisha-swaminathan:
@amtins @phloxic 
I agree that a solution like #7308 is a more ideal permanent solution but since it reverts the change from #5773 (which fixes #6726 and #1102) I think this PR could work as a good solution for now as it'a a smaller change and causes no regression.
The mouse tooltip position still aligns as it should (even when transform: scale is applied to the video container).

I set up a demo player to show
- that there is no regression in the player behavior. Specifically that the fix for #6726 and #1102 is not reverted - https://codepen.io/hswaminathan/pen/xxBOVLZ
- how the fix works when the progress bar is extended to the full width of the player  and transform is applied to the container - [with fix](https://codepen.io/hswaminathan/pen/VwRjmBW ) vs [without fix](https://codepen.io/hswaminathan/pen/XWGjMKV )

roman-bc-dev (approved):
I think this is a simple fix that addresses the issue, helps define expected UI behavior, and doesn't prevent a deeper refactor in the future (if that's something we'd like to invest effort into).

harisha-swaminathan:
> Both in https://codepen.io/hswaminathan/pen/xxBOVLZ and https://codepen.io/hswaminathan/pen/VwRjmBW you can see the tooltip staying within the left boundary of the _control bar_ or the player(which imo is the correct behavior), but stuck on the right by the boundary of the _progress bar_.
> 
> It might still worth looking into #8530
> 
> Also, when the cursor is dragged the current time tooltip does weird jumps on both ends. In my tests this could be remedied by setting [enableSmoothSeeking](https://github.com/videojs/video.js/pull/8287).



> Both in https://codepen.io/hswaminathan/pen/xxBOVLZ and https://codepen.io/hswaminathan/pen/VwRjmBW you can see the tooltip staying within the left boundary of the _control bar_ or the player(which imo is the correct behavior), but stuck on the right by the boundary of the _progress bar_.
> 
> It might still worth looking into #8530
> 
> Also, when the cursor is dragged the current time tooltip does weird jumps on both ends. In my tests this could be remedied by setting [enableSmoothSeeking](https://github.com/videojs/video.js/pull/8287).


I've updated the PR to set the tooltip boundary within the progress bar on both sides for the sake of consistency. Looks like #8530 too limits the tooltip within the progress bar instead of the control bar. 

Also, I observed that in https://phloxic.productions/test/videojs/ttt/v8-dev.html, the tooltip overflows in the left side.

phloxic:
> @phloxic I've updated the PR to set the tooltip boundary within the progress bar on both sides for the sake of consistency.

Yes, that's consistent. But isn't it kind of ironic (or inconsistent) from a viewer's perspective to _reduce_ the room for tooltip display specifically for the use cases of tooltips which are _enlarged_, e.g., thumbnail previews etc. or, indeed, scaling the player?

It's technically correct, but in my personal opinion a step backwards from the audience's point of view. I  don't think there are a lot of complaints about the tooltip sticking out of bounds of the progress bar. I'm reasonably certain that I would notice the general behaviour change (and complain). But then I might be the only one ;-)

Having said that I'm obviously fine with conceding that @roman-bc-dev's [argument](https://github.com/videojs/video.js/pull/8527#pullrequestreview-1830635165) may be the realistic way forward for the project.

> Looks like #8530 too limits the tooltip within the progress bar instead of the control bar.

Yes, in the second example.

> Also, I observed that in https://phloxic.productions/test/videojs/ttt/v8-dev.html, the tooltip overflows in the left side.

Yes, scaling the player is still an issue.

roman-bc-dev (approved):

misteroneill (approved):
I agree with @roman-bc-dev that this is an overall improvement even with existing limitations. Great discussion and thanks for the feedback @phloxic!

PR #8529, opened 2023-12-15
fix: error-display

harisha-swaminathan:
## Description
This recent change closes the error modal and opens it again inorder to  ensure that the error display component is updated with the new error message when consecutive errors occur.

But when this is used with the videojs-errors plugin, no error is displayed after the second error occurs. This is because [when the error display is closed ](https://github.com/videojs/video.js/blob/main/src/js/error-display.js#L27)after the second error, error is cleared and set to null [here](https://github.com/videojs/videojs-errors/blob/main/src/plugin.js#L302).

## Specific Changes proposed
The suggested change is to just update the modal's content with the latest error message instead of closing and opening the error display.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/error-display.js
+++ b/src/js/error-display.js
@@ -24,7 +24,6 @@ class ErrorDisplay extends ModalDialog {
   constructor(player, options) {
     super(player, options);
     this.on(player, 'error', (e) => {
-      this.close();
       this.open(e);
     });
   }
--- a/src/js/modal-dialog.js
+++ b/src/js/modal-dialog.js
@@ -156,10 +156,8 @@ class ModalDialog extends Component {
    * @fires ModalDialog#modalopen
    */
   open() {
-    if (this.opened_) {
-      if (this.options_.fillAlways) {
-        this.fill();
-      }
+    if (this.opened_ && this.options_.fillAlways) {
+      this.fill();
       return;
     }
 
--- a/src/js/modal-dialog.js
+++ b/src/js/modal-dialog.js
@@ -156,51 +156,56 @@ class ModalDialog extends Component {
    * @fires ModalDialog#modalopen
    */
   open() {
-    if (!this.opened_) {
-      const player = this.player();
-
-      /**
-        * Fired just before a `ModalDialog` is opened.
-        *
-        * @event ModalDialog#beforemodalopen
-        * @type {Event}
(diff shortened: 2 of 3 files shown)


amtins:
@harisha-swaminathan visually, this modification has the same effect. However, I have a question about accessibility. Shouldn't there be a narrated feedback after each error?

Before this modification, if the player had focus and several errors occurred, there was a narrated feedback. Even if the feedback wasn't very descriptive, the user was informed that something had happened.

This case can be tested as follows:
1. open Chrome
2. install the screen reader extension
3. open a page with the player
4. open the developer console
5. paste the script
  ```javascript
  setTimeout(() => {
      player.src('https://media-url.com/404.m3u8')

      setTimeout(() => {
          player.src('https://media-url.com/not-compatible.rtmp')
      }, 15_000);
  }, 10_000);
  ```
6. return to the page with the player and click play then pause
7. wait for errors to occur and listen to narrated feedback


It is possible to provide a little more context on the error by modifying the script as follows:
```javascript
videojs.dom.$('.vjs-error-display .vjs-modal-dialog-content').id = 'vjs-modal-content-id';

videojs.dom.$('.vjs-error-display').setAttribute('aria-describedby', `${videojs.dom.$('.vjs-error-display').getAttribute('aria-describedby')}, vjs-modal-content-id`);

setTimeout(() => {
    player.src('https://media-url.com/404.m3u8')

    setTimeout(() => {
        player.src('https://media-url.com/not-compatible.rtmp')
    }, 15_000);
}, 10_000);
```

harisha-swaminathan:
Thank you @amtins

This, somehow, isn't a problem when voiceOver accessibility on Mac is used. However, I'm able to reproduce it with a screen reader extension.

Do you think "aria-live: polite" can be user here instead? Added that in the latest commit and it seems to work as expected.

roman-bc-dev (approved):
- This change seems to be inline with the existing visible UX. We may want to consider a solution in the future that would allow the user to see both errors in the player modal and not just the last one to be generated. 
- Because we can't guarantee the timing between errors, it's quite possible that `aria-live: polite` here could result in both errors being announced by a screen reader but I think that might actually be preferable to only hearing about the latest error emitted.

PR #8542, opened 2024-01-03
feat: add an option to render a thumbnail when hovering the progress bar

juloStud:
## Description
add "thumbnailPreview" option to the control bar to render a thumbnail when hovering the progress bar

## Specific Changes proposed
thumbnailPreview is a boolean. If set to false or not set, the thumbnail won't be rendered below the progress bar. If set to yes, it will be.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [x] Example created : thumbnail-preview.html.example

### DIFF
--- a/sandbox/thumbnail-preview.html.example
+++ b/sandbox/thumbnail-preview.html.example
@@ -1,0 +1,113 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Video.js Sandbox</title>
+    <link href="../dist/video-js.css" rel="stylesheet" type="text/css" />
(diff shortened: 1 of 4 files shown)


mister-ben:
Thank you for the contribution, we appreciate it. I think however this would be more appropriate as a plugin rather than added to the core player, mainly because it would only work with simple sources e.g. MP4, but not HLS or DASH played with MSE. Those formats are usually used for longer videos where this UX adds most value.

There are plugins which add similar functionality, such as https://github.com/phloxic/videojs-sprite-thumbnails which uses image sprites as the thumbnail source. This doesn't suffer the latency that seeking through the second video element has in this video element approach.

PR #8564, opened 2024-01-19
feat: improved error interface

wseymour15:
## Description

Changes to the `MediaError` class so we can expect a `metadata` property on this class, as well as creating a list of error types.

## Specific Changes proposed

- Update of the MediaError class so jsdoc generates TS types for error types, as well as adding a new `metadata` property so we can give greater specificity to errors.
- Create a list of error types set on the `video.js` object to be reused throughout video.js and for developer usage.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/video.js
```diff
@@ -617,4 +617,20 @@ videojs.str = Str;
...
+// TODO: Move this to another file and import it.
+videojs.Errors = {
+  DashManifestRequestError: 'dash-manifest-request-error',
+  DashManifestSidxContainerError: 'dash-sidx-container-error',
+  DashManifestSidxParsingError: 'dash-manifest-sidx-parsing-error',
+  HlsPlaylistRequestError: 'hls-playlist-request-error',
+  SegmentUnsupportedMediaFormat: 'segment-unsupported-media-format',
+  SegmentUnsupportedContainer: 'segment-unsupported-container',
+  SegmentSwitchError: 'segment-switch-error',
+  SegmentExceedsBufferQuota: 'segment-exceeds-buffer-quota',
```
dzianis-dashkevich, line 629: I suggest using `*-error` format

src/js/player.js
```diff
@@ -4200,6 +4200,7 @@ class Player extends Component {
       return;
     }
 
+    // error should contain eventtype, and be passed in
```
dzianis-dashkevich, line 4203: you mean errorType?

src/js/media-error.js
```diff
@@ -75,6 +76,15 @@ MediaError.prototype.message = '';
...
 MediaError.prototype.status = null;
 
+/**
+ * An optional error type string to give more detail about the error. This can be used to give
+ * a higher level of specificity to an error versus the more generic MediaError codes.
+ * This value should align with the values from videojs.Errors
+ *
+ * @type {string}
+ */
+MediaError.prototype.errorType = null;
```
dzianis-dashkevich, line 86: I would suggest something more robust,

interface VideoJsError {
  errorType: ErrorType
}

interface ErrorMetadata {
  error: VideoJsError // other errors may have additional properties other than just errorType
}

MediaError.metadata: ErrorMetadata

src/js/player.js
wseymour15: Yup, this is a typo. This comment will be removed, and was just meant to show the process of how the `errorType` should be passed through.

src/js/media-error.js
wseymour15: Good call, it would probably be nice to pass more metadata than just the type if we want. The issue here is that `interface` is not allowed in JS.

We could change `errorType` to `metadata` here, and then just allow any properties to be tied to it. I am trying to think of a way in JS to be more strict about what `metadata` could contain though, without using TS.

dzianis-dashkevich: You can use typescipt interfaces using typedef in jsdoc

wseymour15: Makes sense, I had to toggle with it a bit, but I made some updates to the JSDoc to ensure that is all working. Let me know if it looks correct to you.

src/js/utils/errors.js
```diff
@@ -0,0 +1,13 @@
...
+  HlsPlaylistRequestError: 'hls-playlist-request-error',
+  SegmentUnsupportedMediaFormat: 'segment-unsupported-media-format-error',
+  UnsupportedMediaInitialization: 'unsupported-media-initialization-error',
+  SegmentSwitchError: 'segment-switch-error',
+  SegmentExceedsSourceBufferQuota: 'segment-exceeds-source-buffer-quota-error',
+  SegmentAppendError: 'segment-append-error',
+  VttLoadError: 'vtt-load-error',
+  VttCueParsingError: 'vtt-cue-parsing-error',
+  EMEKeySessionCreationError: 'eme-key-session-creation-error'
+};
```
dzianis-dashkevich, line 13: I think it should be in `consts` instead of `utils`

dzianis-dashkevich (approved):

PR #8369, opened 2023-07-19
fix(loading-spinner): border size customization

amtins:
## Description

Allows to change the border size of the `loading-spinner` component without having to reflect the border size in pseudo-element's margin.

[Screencast from 19. 07. 23 18:08:10.webm](https://github.com/videojs/video.js/assets/34163393/c8432886-ff23-4988-b887-23f3ca1b2c6c)

## Specific Changes proposed

- uses display `flex` to facilitate centering of pseudo-elements
- deletes the pseudo-elements `margin`

## Requirements Checklist
- [X] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [X] Change has been verified in an actual browser (Chrome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
- [ ] Reviewed by Two Core Contributors

PR #8595, opened 2024-02-16
fix: Loading spinner animation misalignment

roman-bc-dev:
## Description
Add top/left positioning rules to the `.vjs-loading-spinner:before` and `.vjs-loading-spinner:after` to fix the loading spinner animated segment misalignment.

Screenshot 2024-02-16 at 10 38 21 AM

This seems to occur because the `border: 0.6em` rule on the `.vjs-loading-spinner` parent effectively pushes content inward with the current box-model interpretation, against a set height/width, like padding. Unlike padding, however, the edge of the element box is interpreted as _inside_ the border, not outside, so the spinner segment pseudo-elements inheriting its positioning rules are visually nudged downward and to the right. 

Adding negative top/left positioning rules to account for the border fixes the misalignment. It's most likely why there was a negative margin in place for the segments in the past.

Screenshot 2024-02-16 at 10 39 28 AM

## Specific Changes proposed
Add top/left positioning rules targeting `.vjs-loading-spinner::before, .vjs-loading-spinner::after`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_loading.scss
+++ b/src/css/components/_loading.scss
@@ -39,6 +39,8 @@
 .vjs-loading-spinner:after {
   content: "";
   position: absolute;
+  left: -0.6em;
+  top: -0.6em;
   box-sizing: inherit;
   width: inherit;
   height: inherit;


amtins:
@roman-bc-dev this should no longer be the case since this #8369. Is it possible to have an example to reproduce the issue?

I did a quick test of this PR and it seems to bring the issue back.

[Screencast from 16. 02. 24 19:16:13.webm](https://github.com/videojs/video.js/assets/34163393/ec35be49-beaa-4268-89c9-6c0ddaf800e6)

amtins:
I was able to reproduce the issue by doing:

```javascript
player.loadingSpinner.lockShowing()
```

If it's the case I would do something like #8435 instead.

roman-bc-dev:
Hi @amtins. I'm able to reproduce this issue consistently in local videojs samples by updating visibility on the `.vjs-loading-spinner` element with:
```
  display: block;
  visibility: visible;
```

I might be misinterpreting how `.vjs-lock-showing` is applied generally  but changing that display rule up top to `flex` instead of `block` did not alleviate the issue. I wouldn't expect it to because neither block nor flex rules modify the box model of an element and they wouldn't change how the border affects descendant elements.

Furthermore, because the pseudo-elements are affected by `position: absolute` rules, the correct fix here is to add top and left positioning. Not just to account for the border in this case but because not stipulating those rule can cause unexpected behavior.

amtins:
@usmanonazim could you share in which sandbox example you have the problem, I can't reproduce the problem unless I use `player.loadingSpinner.lockShowing()`? 😢

roman-bc-dev:
@amtins This is a general issue that is visible outside of `lockShowing()`.

1. We're not changing `.vjs-lock-showing` behavior with this change and are not interested in doing so.
2. Manually modifying an individual style rule out of context for a single element will negatively affect any part of the spinner animation solution. That is especially true for the border width because it _must match_ the `.vjs-loading-spinner` element and its pseudo-element descendants. I don't believe it's reasonable to expect any single solution to cover all possible permutations and combinations of sizing in this instance. However, it absolutely should at least accommodate the existing `0.6em` border width (which adding left/top rules does).

amtins:
Thanks for the clarification 👍🏽. Since change #8369 the loading-spinner uses centering with `flex` positioning. This change makes it easier to customize the thickness of the component border without having to worry about adapting margins to stay aligned with the parent, which improves the developer experience.

There are several solutions, depending on the experience you wish to offer developers:
- Removing entirely the change #8369, which repositions the element in `block` and adds `margins` back to pseudo-elements. This avoids having a `top` and a `left`, which makes customization even more complex than before change #8369.
- Moving `justify-content`, `align-items` directly into the `.vjs-loading-spinner` class, so that if for any reason a developer wants to display the loading spinner, all the developer has to do is a `display: flex`. https://github.com/videojs/video.js/blob/395d608761d2c145975c11bb6e965f2b2f7c7cb8/src/css/components/_loading.scss#L26-L27 can be moved here https://github.com/videojs/video.js/blob/395d608761d2c145975c11bb6e965f2b2f7c7cb8/src/css/components/_loading.scss#L1-L21

To illustrate my point, here's an example:

- With the fix 8369
  ```css
  .vjs-loading-spinner {
    border: 0.9em solid green;
    display: flex;
  }
  ```

- Before the fix 8369
  ```css
  .vjs-loading-spinner {
    border: 0.9em solid green;
    display: block;
  }
  .vjs-loading-spinner::before,
  .vjs-loading-spinner::after {
    margin: -0.9em;
  }
  ```

- With this PR
  ```css
  .vjs-loading-spinner {
    border: 0.9em solid green;
    display: block;
  }
  .vjs-loading-spinner::before,
  .vjs-loading-spinner::after {
    left: -0.9em;
    top: -0.9em;
  }
  ```

- Moving `justify-content`, `align-items` directly into the `.vjs-loading-spinner`
  ```css
  .vjs-loading-spinner {
    border: 0.9em solid green;
    display: flex;
    visibility: visible;
  }
  ```

PR #8614, opened 2024-02-27
chore: Add action to validate PR titles

mister-ben:
## Description
Adds a GitHub action to validate that PR titles follow conventional commit, to make sure we don't merge a PR that wouldn't show up in the change log.

## Specific Changes proposed
Adds a simple action to validate the title with a regex. There are GitHub marketplace actions which will do this, but they have a lot of dependencies and configuration options we don't need for now.

### DIFF
--- a/.github/actions/pr-titles.js
+++ b/.github/actions/pr-titles.js
@@ -1,0 +1,13 @@
+const core = require('@actions/core');
+const github = require('@actions/github');
+
+(async function run() {
+  const title = github.context.payload.pull_request?.title;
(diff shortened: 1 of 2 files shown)


amtins (approved):
LGTM, I have a question, though: why not introduce a tool like `commitlint`?

mister-ben:
> LGTM, I have a question, though: why not introduce a tool like `commitlint`?

That would enforce the format on each individual commit message, which isn't too important as we squash and merge, and wouldn't stop the PR title from being changes to something that doesn't fit.

gkatsev (approved):

dzianis-dashkevich (approved):

PR #8615, opened 2024-02-27
feat(lang): added arabic seek button translations

silevitas:
## Description
Added Arabic translations for seek forward/back buttons.

## Specific Changes proposed
Added two Arabic translations: `Skip backward {1} seconds` and `Skip forward {1} seconds`.


mister-ben:
Thanks for the contribution! package-lock.json shouldn't be updated though, could you remove that please?

PR #8616, opened 2024-02-27
feat(lang): added arabic seek button translations

silevitas:
## Description
Added two Arabic new seek button translations.

## Specific Changes proposed
Arabic translations for `Skip backward {1} seconds` and `Skip forward {1} seconds`.

### DIFF
--- a/docs/translations-needed.md
+++ b/docs/translations-needed.md
@@ -16,9 +16,8 @@ This default value is hardcoded as a default to the localize method in the SeekB
 
 | Language file           | Missing translations                                                                |
 | ----------------------- | ----------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


mister-ben:
This line and the corresponding one in the forward buttons should use `this.skipTime.toLocaleString(player.language())` so actual Arabic numbers rather than what we call Arabic numbers are used.

https://github.com/videojs/video.js/blob/a57b07a418636daf9723a28715f96f2e3152ddc3/src/js/control-bar/skip-buttons/skip-backward.js#L21

silevitas:
Added the localisation support to the skipButtons as recommended!

mister-ben (approved):

adrums86 (approved):

PR #8625, opened 2024-03-01
feat(emulated-tracks): add class to force cues to be center aligned

gkatsev:
https://github.com/videojs/http-streaming/pull/1408 updated 608 captions to default to be left aligned. This may be unwanted by some folks and we should provide an easier way to force them to be centered. This PR adds a player level class that will override the text alignment to be `center`. It also overrides the `width` to `80%` because otherwise the cue box isn't set up correctly to be 10% from the right of the display area (a side effect of hardcoding a width value and using inset in the generation of the cues).

## Description

## Specific Changes proposed

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -40,3 +40,9 @@ video::-webkit-media-text-track-display {
 .video-js.vjs-user-inactive.vjs-playing video::-webkit-media-text-track-display {
   @include transform(translateY(-1.5em));
 }
+
+// force cues to be center aligned
+.video-js.vjs-force-center-align-cues .vjs-text-track-cue {
+  text-align: center !important;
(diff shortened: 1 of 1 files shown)


wseymour15:
This looks like a fine change to me, assuming that this class setting will be optional. We made this change because according to [the spec](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-608): "CEA-608 does not distinguish between different alignment, so all captions are left aligned and require a WebVTT align cue setting to override the default middle alignment".

We interpreted this to mean by default, 608 captions should be left aligned.

wseymour15 (approved):

adrums86 (approved):

PR #8629, opened 2024-03-05
feat: Adds a transient button component

mister-ben:
## Description
Adds a `TransientButton` component for the types of button that are shown on top of the video briefly during playback and reappear when there is user activity. e.g. Unmute buttons, skip intro. It aims is to be a generic button type to be extended. Some basic styles are provided but kept light to not complicate customisation.
It's important to insert a transient button before the control bar for the tab order to make sense.

_Optionally_ takes focus when shown.

## Specific Changes proposed
Adds `TransientButton` component.

## Requirements Checklist
  - [ ] Docs/guides updated
  - [x] Example: https://deploy-preview-8629--videojs-preview.netlify.app/sandbox/transient-button.html


src/css/components/_transient-button.scss
```diff
@@ -0,0 +1,62 @@
...
+  right: 5px;
+}
+
+.video-js .vjs-transient-button.top {
+  top: 5px;
+}
+
+.video-js .vjs-transient-button.bottom {
+  bottom: 35px;
+}
```
amtins, line 35: Using em instead of px allows greater flexibility when changing player size. What do you think ?

With PX
[Screencast from 06. 03. 24 15:56:09.webm](https://github.com/videojs/video.js/assets/34163393/bcfc6c89-c0cf-48a2-b42e-868dab4003dc)

With EM
[Screencast from 06. 03. 24 15:57:20.webm](https://github.com/videojs/video.js/assets/34163393/bcda9e95-8855-4b24-b905-e923ff09846c)

src/css/components/_transient-button.scss
```diff
@@ -0,0 +1,62 @@
...
+  font-size: 120%;
+}
+
+.video-js.vjs-layout-x-large .vjs-transient-button,
+.video-js.vjs-layout-huge .vjs-transient-button {
+  height: 8em;
+  padding: 0 1.5em;
+
+  font-size: 150%;
+}
```
amtins, line 49: I would have been inclined to do nothing in this case, in order to leave freedom to the developer, and keep the same visual consistency as the control bar.

What was the intention behind it? Maybe I'm missing a bit of context, hence my comment.

src/js/transient-button.js
```diff
@@ -0,0 +1,77 @@
...
+
+    // When shown, the float button will be visible even if the user is inactive.
+    // Clear this if there is any interaction.
+    player.on(['useractive', 'userinactive'], () => {
+      this.removeClass('force-display');
+    });
+  }
+
+  buildCSSClass() {
+    return `vjs-transient-button ${this.options_.position}`;
```
amtins, line 39: How about prefixing the CSS classes `top`, `right`, `bottom`, `left` with `vjs`? The reasoning is that these terms are so generic that they could potentially create a conflict.

src/css/components/_transient-button.scss
```diff
@@ -0,0 +1,62 @@
...
+
+.video-js .vjs-transient-button.top.avoid-title {
+  top: 80px;
+}
```
amtins, line 53: Does the developer have to specify this class manually? Do you think it would be possible to use `em`s here too?

src/js/transient-button.js
mister-ben: Very good point

src/css/components/_transient-button.scss
mister-ben: It's to avoid small buttons on large players. Maybe we could leave this out, I do want to keep the styling functional but relatively light so they can be customised.

src/css/components/_transient-button.scss
mister-ben: Changed this to a `near-top` position options

amtins (approved):
LGTM!

dzianis-dashkevich (approved):

PR #8634, opened 2024-03-08
chore: add contrib-eme errors

adrums86:
## Description
Add error constants for contrib-eme specific errors.

## Specific Changes proposed
Add constants that are representative of the separate keysystems for error reporting out of videojs-contrib-eme, also add errors specific to each step in the EME specific process.
Two notable omissions:
No errors for `remove` or `load` on the mediakeysession as they're not implemented in contrib-eme.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/consts/errors.js
+++ b/src/js/consts/errors.js
@@ -16,5 +16,9 @@ export default {
   AdsMidrollError: 'ads-midroll-error',
   AdsPostrollError: 'ads-postroll-error',
   AdsMacroReplacementFailed: 'ads-macro-replacement-failed',
-  AdsResumeContentFailed: 'ads-resume-content-failed'
+  AdsResumeContentFailed: 'ads-resume-content-failed',
+  // Errors used in contrib-eme:
+  EMEEncryptedError: 'eme-encrypted-error',
+  MSKeyError: 'ms-key-error',
+  WebkitKeyError: 'webkit-key-error'
 };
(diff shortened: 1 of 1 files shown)


wseymour15 (approved):
🚀

src/js/consts/errors.js
```diff
@@ -9,12 +9,23 @@ export default {
...
   AdsPrerollError: 'ads-preroll-error',
   AdsMidrollError: 'ads-midroll-error',
   AdsPostrollError: 'ads-postroll-error',
   AdsMacroReplacementFailed: 'ads-macro-replacement-failed',
-  AdsResumeContentFailed: 'ads-resume-content-failed'
+  AdsResumeContentFailed: 'ads-resume-content-failed',
+  // Errors used in contrib-eme:
+  EMEEncryptedError: 'eme-encrypted-error',
+  MSKeyError: 'ms-key-error',
+  WebkitKeyError: 'webkit-key-error',
```
dzianis-dashkevich, line 22: Could you elaborate more on this 3:

What are they?
Why do we need them?

```js
  EMEEncryptedError: 'eme-encrypted-error',
  MSKeyError: 'ms-key-error',
  WebkitKeyError: 'webkit-key-error',
```

adrums86: The idea behind these 3 was to be a catch-all for the outermost promise to the respective `handle` events. For `MSKeyError` and `WebkitKeyError` they specifically cover errors from legacy workflows ('msneedey' and 'webkitneedkey' events). Since this is not _really_ in-spec EME, I figured they were worthy of their own specific errors. For `EMEEncryptedError` it might not be necessary and was only added incase any other type of unknown error occurs and bubbles up to the `handleEncryptedEvent` promise.

PR #8658, opened 2024-03-25
chore: Remove unused type.js file

SimonAllen0901:
## Description
The type.js file appears to have been added in #8148   but has never been used.
Perhaps it can be removed.

## Specific Changes proposed
Remove type.js


mister-ben (approved):

adrums86 (approved):
:shipit:

Issue #8653, opened 2024-03-22
Potential bug/leak: middleware cache grows even if no middleware created


BrainCrumbz:
### Description

TL;DR: when videojs players are instantiated and disposed repeatedly on page, `middlewareInstances` keeps growing with new keys and null values.

We are hunting potential memory leaks in our video-based app. Comparing heap snapshots on Windows Chrome and Android Chrome, we noticed a (small?) amount of property arrays due to `middlewareInstances`. Snapshot were taken after disposing a player once, and then after instantiating/disposing e.g. 10 times. 

Adding logpoints in `tech/middleware.js`, we noticed that:
1. [`getOrCreateFactory`](https://github.com/videojs/video.js/blob/2f1cc6c2b38182bfb9a4dd77a2a27cd8dd38687c/src/js/tech/middleware.js#L257) is never called, so no new `middlewareInstances` is set for a new player
2. [`clearCacheForPlayer`](https://github.com/videojs/video.js/blob/2f1cc6c2b38182bfb9a4dd77a2a27cd8dd38687c/src/js/tech/middleware.js#L246) is called every time player is disposed, so a new null value is add to `middlewareInstances`

This happened for videojs v7.20.3. Comparing file with [latest revision](https://github.com/videojs/video.js/compare/v7.20.3...main#diff-166dcdb09385d794b7e633d8b85f67ef61f04f0bb4bf3fc7367fbf37025a66bb) showed no significant change, so maybe the behaviour is the same in latest code.

If everything is correct, would you consider a tiny PR that addresses this? E.g. `clearCacheForPlayer' could check if key exists in the first place?

Thanks

### Reduced test case

None. Willing to create one if issue is of interest

### Steps to reproduce

1. After page load, create a player with `videoPlayer = videojs(videoElem, videoPlayerOptions);` and set a source on it with `videoPlayer.src({src: ..., type: ...})`
2. After a timeout, dispose player with `videoPlayer.dispose();`
3. After a timeout, repeat the process e.g. twice

### Errors

None

### What version of Video.js are you using?

7.20.3

### Video.js plugins used.

None

### What browser(s) including version(s) does this occur with?

Chrome 123

### What OS(es) and version(s) does this occur with?

Windows 11

mister-ben:
A PR would be welcome, thanks.

BrainCrumbz:
PR can be found at https://github.com/videojs/video.js/pull/8674

Waiting for comments etc. Bye!

Issue #8149, opened 2023-02-21
(Still) Wrong path for font scss in video-js.scss on v8.0.4


BrainCrumbz:
### Description

Apparently we met the issue mentioned at #7863 . 

### Steps to reproduce

1. Clone the git repository
2. rum `npm install`
3. run `npm test`. Also, one can run `npm run build` afterwards.

Expected result:  
All should be fine.

Actual result:
scripts show errors.

### Errors

```
[... omitted ...]
> video.js@8.0.4 build:css
> npm-run-all build:css:*


> video.js@8.0.4 build:css:cdn
> sass --load-path='./' --load-path='./node_modules/' --no-source-map src/css/vjs-cdn.scss dist/alt/video-js-cdn.css

Error: Can't find stylesheet to import.
  ╷
5 │ @import "videojs-font/scss/icons";
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  src\css\video-js.scss 5:9  @import
  src\css\vjs-cdn.scss 3:9   root stylesheet
ERROR: "build:css:cdn" exited with 65.
ERROR: "build:css" exited with 1.
```

The errors seem the same for test or build.

### What version of Video.js are you using?

8.0.4 up to commit a27ee05

### Video.js plugins used.

None

### What browser(s) including version(s) does this occur with?

None. This is from command line

### What OS(es) and version(s) does this occur with?

Windows 11  
npm 9.3.1  
node v18.14.0

Both from windows command line as well as WSL terminal.

BrainCrumbz:
Is there any workaround (even temporary) for this one? So that we can run tests and give a go to #8136

amtins:
@BrainCrumbz I just tested to see if we could possibly run into the same error. However, I did not encounter this issue. But maybe I didn't follow the procedure as described.

Here is what I did:
- I've `node v18.15.0`, `npm v9.5.0`
- I `clone` de repo `git@github.com:videojs/video.js.git`, with the latest version 8.2.0
- I CD into the folder
- I ran `npm install`
- I ran `npm test`
- I ran `npm run build`

Everything went smoothly, see the logs:
```bash
> video.js@8.2.0 build:css
> npm-run-all build:css:*


> video.js@8.2.0 build:css:cdn
> sass --load-path='./' --load-path='./node_modules/' --no-source-map src/css/vjs-cdn.scss dist/alt/video-js-cdn.css


> video.js@8.2.0 postbuild:css:cdn
[...]

PR #8674, opened 2024-04-08
fix(middleware): cache grows even if no middleware created

BrainCrumbz:
## Description
See issue #8653 

## Specific Changes proposed
When in `middleware.js` the function `clearCacheForPlayer` runs, before setting a value to null in middlware caches, it checks if the key exists in the first place.

## Requirements Checklist
  - [ ] Docs/guides updated


BrainCrumbz:
BTW, we're contributing to this on Windows, and still suffer from issue #8149. To make that work, we temporarily modified package.json

BrainCrumbz:
We noticed from e-mails some repeated failure on that CI task. I guess, that is not related to the specific changes in code, is it?

src/js/tech/middleware.js
```diff
@@ -244,7 +244,9 @@ function executeRight(mws, method, value, terminated) {
  *         A {@link Player} instance.
  */
 export function clearCacheForPlayer(player) {
-  middlewareInstances[player.id()] = null;
+  if (middlewareInstances.hasOwnProperty(player.id())) {
+    middlewareInstances[player.id()] = null;
```
mister-ben, line 248: I wonder if `delete` is preferable, if the object having null properties was the issue being addressed

mister-ben:
> We noticed from e-mails some repeated failure on that CI task. I guess, that is not related to the specific changes in code, is it?

No, I tried re-running the failing test a few times. Upload to codecov is inexplicably flaky, even when a `CODECOV_TOKEN` token is used.

BrainCrumbz:
Hi there. Just to let you know: if there's something more to do on our side, we might be able to. 
Kind regards

src/js/tech/middleware.js
mister-ben: Let's go with `delete` here

BrainCrumbz: ok, commit is pushed

BrainCrumbz: not sure if - at this point - code should just delete without checking whether `middlewareInstances` has that property or not.

mister-ben (approved):

dzianis-dashkevich (approved):

PR #8685, opened 2024-04-14
fix(playback-rate-menu-button): ensure proper menu behavior on double click

amtins:
## Description

When the playback rate button is clicked twice, the menu does not close. This has an impact when a developer has disabled menu display on mouse-over, forcing the user to click elsewhere to close the menu.

### How to reproduce

- open [codepen](https://codepen.io/amtins/pen/KKYBogr)
- click the playback rate button

## Specific Changes proposed

- moves the `label` inside the button
- deletes function `createEl`

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
+++ b/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js
@@ -25,6 +25,13 @@ class PlaybackRateMenuButton extends MenuButton {
   constructor(player, options) {
     super(player, options);
 
+    this.labelElId_ = 'vjs-playback-rate-value-label-' + this.id_;
+    this.labelEl_ = Dom.createEl('div', {
+      className: 'vjs-playback-rate-value',
+      id: this.labelElId_,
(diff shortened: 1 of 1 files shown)


mister-ben:
Not sure I understand the issue. On the codepen, the behaviour of the rates button within each player is the same as the captions/subtitiles and audio buttons.

amtins:
@mister-ben Thank you for taking the time to review this PR. 

In the audio or subtitle menu, the first click will open the menu and the second will close it. Whereas in the playback rate menu, the first click opens the menu but the second does not close it.

mister-ben:
Hmm it varies for browser. For me, Chrome and Firefox are as you say. On Safari all the buttons behave the same (second click doesn't close).

EchoChen012:
The videojs version I use is ^8.12.0, and this problem still exists.

EchoChen012:
The video.js version I use is ^8.12.0, and the same problem still exists on the PC. When the playback rate button is clicked twice, the menu does not close.

PR #8711, opened 2024-04-25
fix: Ensure aria-labelledby values in track settings are valid

### LINKED ISSUE #8708: [8.13.0] HTML-validation errors with the new version
### Description

I upgraded `video.js` to 8.13.0 and my HTML validator started giving the following warnings:

Error: Duplicate ID “captions-background-vjs_video_826_component_1399”.
Error: The “aria-labelledby” attribute must point to an element in the same document.

Indeed in the document I see multiple:
```html
<legend id="captions-background-vjs_video_826_component_1399">
```
And
```html
<select id="vjs_select_1441" aria-labelledby="captions-background-vjs_video_826_component_1399 captions-background-color-vjs_video_826_component_1399">
```

### Reduced test case

None

### Steps to reproduce

Just add `video.js` to any page.

### Errors

_No response_

8.13.0

### Video.js plugins used.

_No response_

Chromium

Linux

cc @bzizmo, @wseymour15, @CarlosVillasenor

mister-ben:
## Description
Fixes new text track settings, so that `aria-labelledby` attributes contain only valid attributes. Currently there is one invalid id, and two els with duplicate ids.
Fixes #8708 

## Specific Changes proposed
Updates text-track-settings.js to set the `id` property to an empty string where there is no `label`, only a `legend` element.
Adds tests to validate aria-labelled by values.

## Requirements Checklist

### DIFF
--- a/src/js/tracks/text-track-settings.js
+++ b/src/js/tracks/text-track-settings.js
@@ -86,7 +86,7 @@ const selectConfigs = {
 
   edgeStyle: {
     selector: '.vjs-edge-style > select',
-    id: '%s',
+    id: '',
     label: 'Text Edge Style',
     options: [
       ['none', 'None'],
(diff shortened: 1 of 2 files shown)


adrums86 (approved):
Makes sense to me. :shipit:

PR #8714, opened 2024-04-29
chore: Add file size PR check

mister-ben:
Adds [compressed-size-action](https://github.com/preactjs/compressed-size-action) on PRs to view the impact on file size.


mister-ben:
Will revisit another time. The action doesn't seem to be able to add a comment to a PR from a fork, so the information is buried in the actions run log.

PR #8716, opened 2024-04-30
refactor: use URL API

mister-ben:
## Description
Replaces some now unnecessary fallbacks with the standard `URL` API.
`parseUrl` new returns an actual `URL` instead of a `URL`-like object.
Updates tests slightly: `getAbsoluteURL()` to match `URL`'s behaviour with trailing slashes, and use a `URL` object in the cross origin tests. 
Adds a test for `getAbsoluteURL()` with a relative URL

## Requirements Checklist

### DIFF
--- a/test/unit/utils/url.test.js
+++ b/test/unit/utils/url.test.js
@@ -43,7 +43,7 @@ QUnit.test('should strip port from hosts using http or https', function(assert) 
 QUnit.test('should get an absolute URL', function(assert) {
   // Errors on compiled tests that don't use unit.html. Need a better solution.
   // assert.ok(Url.getAbsoluteURL('unit.html') === window.location.href);
-  assert.ok(Url.getAbsoluteURL('http://asdf.com') === 'http://asdf.com');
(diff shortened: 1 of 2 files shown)


test/unit/utils/url.test.js
```diff
@@ -81,10 +80,7 @@ QUnit.test('isCrossOrigin can identify cross origin urls', function(assert) {
   // we cannot test that relative urls work on https, though
   assert.ok(!Url.isCrossOrigin('example.vtt'), 'relative url is not cross origin');
 
-  const location = {
-    protocol: 'https:',
-    host: 'google.com'
-  };
+  const location = new URL('https:/google.com');
 
   assert.ok(!Url.isCrossOrigin('https://google.com/example.vtt', location), 'http://google.com from https://google.com is not cross origin');
```
amtins, line 85: LTGM. But while you're at it, would it be possible to modify the description of this test? If not, that's fine too.

```suggestion
  assert.ok(!Url.isCrossOrigin('https://google.com/example.vtt', location), 'https://google.com is not cross origin');
```

amtins (approved):

PR #8717, opened 2024-04-30
fix(spatialNavigation): change method name

bzizmo:
## Description
Change method name `updateFocusableComponents `to more appropriate name `getFocusableComponents `, since it is a getter.

## Specific Changes proposed
Change method name `updateFocusableComponents `to `getFocusableComponents `

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/spatial-navigation.js
+++ b/src/js/spatial-navigation.js
@@ -49,7 +49,7 @@ class SpatialNavigation extends EventTarget {
     this.player_.on('modalKeydown', this.onKeyDown_);
     // Listen for source change events
     this.player_.on('loadedmetadata', () => {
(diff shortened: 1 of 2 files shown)


src/js/spatial-navigation.js
```diff
@@ -223,7 +223,7 @@ class SpatialNavigation extends EventTarget {
    * @return {Array}
    *         Returns an array of focusable components.
    */
-  updateFocusableComponents() {
+  getFocusableComponents() {
```
amtins, line 226: Quick question, isn't this an accidental breaking change? If so, shouldn't we remove the `npm` package of the version `8.13.0` to avoid making a breaking change?

PR #8735, opened 2024-05-17
refactor: Replace keycode dependency with event.key

mister-ben:
## Description
Removes the [keycode](https://www.npmjs.com/package/keycode) dependency. That relies on now deprecated event properties. `KeyboardEvent.key` is now [ubiquitous](https://caniuse.com/keyboardevent-key), so can be used directly instead.

Reduces min.js slightly, 646.9 kB => 645.22 kB

The uses of deprecated `keyCode` in utils/spatial-navigation-keycode.js ought to be reviewed too.

## Requirements Checklist
  - [ ] Docs/guides updated


src/js/menu/menu-item.js
```diff
@@ -96,7 +94,7 @@ class MenuItem extends ClickableComponent {
    * @listens keydown
    */
   handleKeyDown(event) {
-    if (!MenuKeys.some((key) => keycode.isEventKey(event, key))) {
+    if (['Tab', 'Escape', 'ArrowUp', 'ArrowLeft', 'ArrowRight', 'ArrowDown'].includes(event.key)) {
```
gkatsev, line 97: I guess menukeys was only ever used here?

gkatsev (approved):

src/js/menu/menu-item.js
mister-ben: Yes, or at least it is now.

src/js/menu/menu-button.js
```diff
@@ -298,19 +297,19 @@ class MenuButton extends Component {
   handleKeyDown(event) {
 
     // Escape or Tab unpress the 'button'
-    if (keycode.isEventKey(event, 'Esc') || keycode.isEventKey(event, 'Tab')) {
+    if (event.key === 'Esc' || event.key === 'Tab') {
```
gkatsev, line 300: looks like during review, we missed that these should be `Escape` rather than `Esc`, which keycode likely helped with. Someone noticed this issue in slack.

PR #8793, opened 2024-07-10
fix(menu-item): Replace innerHTML of menu item to fix string bug on player

CarlosVillasenor:
## Description
Replace innerHTML of menu item to fix string bug on player.

## Specific Changes proposed
With a condition using a regex replace innerHTML of menu item to fix string bug on player when content of the string contains HexCode.   

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/menu/menu-item.js
+++ b/src/js/menu/menu-item.js
@@ -74,6 +74,15 @@ class MenuItem extends ClickableComponent {
       textContent: this.localize(this.options_.label)
     });
 
+    const containsHexCode = (s) => {
+      return /\w*(&#x...)\w*/.test(s);
+    };
+
+    if (containsHexCode(menuItemEl.textContent)) {
+      // Replacement that allows innerHTML to be render properly.
+      menuItemEl.innerHTML = menuItemEl.textContent;
+    }
+
     // If using SVG icons, the element with vjs-icon-placeholder will be added separately.
     if (this.player_.options_.experimentalSvgIcons) {
       el.appendChild(menuItemEl);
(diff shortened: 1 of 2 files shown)


src/js/menu/menu-item.js
```diff
@@ -74,6 +74,15 @@ class MenuItem extends ClickableComponent {
       textContent: this.localize(this.options_.label)
     });
 
+    const containsHexCode = (s) => {
```
roman-bc-dev, line 77: This feels like a utility function that pertains to strings in particular so I'm wondering if it might be better to move it [here](https://github.com/videojs/video.js/blob/main/src/js/utils/str.js).

Since we're proposing using innerHTML as part of this solution, and this function is essentially guarding that step, maybe it would be beneficial to add some tighter conditions to make it more difficult to use other problematic characters (i.e.: `/'"`, etc) that are typically escaped for safety? We should also probably add a check for a semicolon `;` as part of the regex test.

mister-ben:
Does the menu item label absolutely need to accept an HTML string? Using `innerText` doesn't seem ideal. Could this be introducing an xss vector, say if an HLS manifest had a captions track with a `NAME` attribute containing HTML? Can this be fixed closer to the source of the problem, by (re)encoding the string properly before it's passed to the component?

CarlosVillasenor:
> Does the menu item label absolutely need to accept an HTML string? Using `innerText` doesn't seem ideal. Could this be introducing an xss vector, say if an HLS manifest had a captions track with a `NAME` attribute containing HTML? Can this be fixed closer to the source of the problem, by (re)encoding the string properly before it's passed to the component?



> HTML string



> Does the menu item label absolutely need to accept an HTML string? Using `innerText` doesn't seem ideal. Could this be introducing an xss vector, say if an HLS manifest had a captions track with a `NAME` attribute containing HTML? Can this be fixed closer to the source of the problem, by (re)encoding the string properly before it's passed to the component?

In terms of security the regex that is used in the new function 'containsHexCode' was updated and will not accept any problematic characters like (i.e.: /'", etc)

PR #8811, opened 2024-07-21
fix(types): Add has|usingPlugin to typedef by adding stubs which are removed from builds

mister-ben:
## Description
tsc doesn't understand mixins and ignores jsdoc not followed by code. The jsdoc for the plugin methods `usingPlugin()` and `hasPlugin()` in Player are being ignored. To get them included in type outputs we need to have otherwise unnecessary stubs codes, as we already have for `on()` etc, which adds unnecessary, even if a small amount of, code to the outputs.

## Specific Changes proposed
* Slight refactor of Player to include those stubs.
* Adds a rollup plugin to delete lines between certain comments, so those stubs are deleted from the outputs.
* Applies those comments to the `on()` etc stubs in Component and the new plugin stubs in Player.

Any code surrounded by these comments, and the comments themselves, is deleted from the dist and test builds:

```js
/* start-delete-from-build */
  console.log('hi');
/* start-delete-from-build */
```

Compared to main, video.min.js is 53 bytes smaller.

## Requirements Checklist
  - [ ] Docs/guides updated


misteroneill (approved):
Not a huge fan of needing to do this, but if it's the only thing that works...

PR #8826, opened 2024-08-09
fix: allow captions in devices that use old chrome to be shown

CarlosVillasenor:
## Description
In the specific case of using an old chrome version this minor changes will allow the captions to be shown.

## Specific Changes proposed
Small fix when using videojs in an old browser that does not support the css 'inset' property.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -47,9 +47,11 @@ video::-webkit-media-text-track-display {
   width: 80% !important;
 }
 
-.video-js .vjs-text-track-display-inset {
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+@supports (inset: 10px) {
+  .video-js .vjs-text-track-display {
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+  }
 }
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -47,7 +47,7 @@ video::-webkit-media-text-track-display {
   width: 80% !important;
 }
 
-.video-js .inset-alternative {
+.video-js .vjs-text-track-display-inset {
   top: 0;
   right: 0;
   bottom: 0;
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -46,3 +46,10 @@ video::-webkit-media-text-track-display {
   text-align: center !important;
   width: 80% !important;
 }
+
+.video-js .vjs-text-track-display > div {
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+}
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -48,7 +48,7 @@ video::-webkit-media-text-track-display {
 }
 
 @supports (inset: 10px) {
-  .video-js .vjs-text-track-display {
+  .video-js .vjs-text-track-display > div {
     top: 0;
     right: 0;
     bottom: 0;
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -47,8 +47,7 @@ video::-webkit-media-text-track-display {
   width: 80% !important;
 }
 
-.video-js .vjs-text-track-display > div {
-  position: absolute;
+.video-js .inset-alternative {
   top: 0;
   right: 0;
   bottom: 0;
--- a/src/css/components/_text-track.scss
+++ b/src/css/components/_text-track.scss
@@ -48,6 +48,7 @@ video::-webkit-media-text-track-display {
 }
 
 .video-js .vjs-text-track-display > div {
+  position: absolute;
   top: 0;
   right: 0;
   bottom: 0;
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -330,6 +330,9 @@ class TextTrackDisplay extends Component {
       // Clear inline style before getting actual height of textTrackDisplay
       textTrackDisplay.style = '';
 
+      // Add custom class to textTrackDisplay div child for not inset support styles
+      textTrackDisplay.firstChild.classList.add('inset-alternative');
+
       // textrack style updates, this styles are required to be inline
       tryUpdateStyle(textTrackDisplay, 'position', 'relative');
       tryUpdateStyle(textTrackDisplay, 'height', (playerHeight - controlBarHeight) + 'px');
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -330,9 +330,6 @@ class TextTrackDisplay extends Component {
       // Clear inline style before getting actual height of textTrackDisplay
       textTrackDisplay.style = '';
 
-      // Add custom class to textTrackDisplay div child for not inset support styles
-      textTrackDisplay.firstChild.classList.add('vjs-text-track-display-inset');
-
       // textrack style updates, this styles are required to be inline
       tryUpdateStyle(textTrackDisplay, 'position', 'relative');
       tryUpdateStyle(textTrackDisplay, 'height', (playerHeight - controlBarHeight) + 'px');
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -322,14 +322,27 @@ class TextTrackDisplay extends Component {
     }
 
     if (browser.IS_SMART_TV && !window.CSS.supports('inset', '10px')) {
-      const textTrack = window.document.querySelector('.vjs-text-track-display');
-      const textTrackHeight = textTrack.getBoundingClientRect().height;
+      const textTrackDisplay = this.el_;
+      const textTrackDisplayHeight = textTrackDisplay.getBoundingClientRect().height;
+      const vjsTextTrackCues = textTrackDisplay.querySelectorAll('.vjs-text-track-cue');
 
-      // This styles are required to be inline
-      textTrack.style.position = 'relative';
-      textTrack.style.height = textTrackHeight + 'px';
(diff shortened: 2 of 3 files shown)


mister-ben:
Would it work to put the changes to `.vjs-text-track-display` in `updateDisplayOverlay()` in this block before the return? Just to keep the alternate styles for the same element close to each other for maintainability. 

https://github.com/videojs/video.js/blob/5192a6758ff92177eaa1134502c032ee22d73018/src/js/tracks/text-track-display.js#L343-L345

test/unit/tracks/text-track-display.test.js
```diff
@@ -500,4 +500,32 @@ if (!Html5.supportsNativeTextTracks()) {
 
     player.dispose();
   });
+
+  QUnit.test('should use relative position for vjs-text-track-display element if browser is does not support inset property', function(assert) {
```
wseymour15, line 504: nit:
```suggestion
  QUnit.test('should use relative position for vjs-text-track-display element if browser does not support inset property', function(assert) {
```

src/js/tracks/text-track-display.js
```diff
@@ -319,6 +320,17 @@ class TextTrackDisplay extends Component {
       }
       this.updateForTrack(descriptionsTrack);
     }
+
+    if (browser.IS_SMART_TV && !window.CSS.supports('inset', '10px')) {
+      const textTrack = window.document.querySelector('.vjs-text-track-display');
```
wseymour15, line 325: Can you just use:
```suggestion
      const textTrack = this.el_;
```

src/js/tracks/text-track-display.js
```diff
@@ -484,6 +496,19 @@ class TextTrackDisplay extends Component {
         this.updateDisplayState(track);
       }
     }
+
+    if (browser.IS_SMART_TV && !window.CSS.supports('inset', '10px')) {
```
wseymour15, line 500: Should this logic be in the loop above? It appears that there can be many text-track-cues. Should this only be done on the first of them?

src/js/tracks/text-track-display.js
CarlosVillasenor: this was added in the last commit, can this be re-reviewed?

src/js/tracks/text-track-display.js
```diff
@@ -319,6 +320,23 @@ class TextTrackDisplay extends Component {
       }
       this.updateForTrack(descriptionsTrack);
     }
+
+    if (browser.IS_SMART_TV && !window.CSS.supports('inset', '10px')) {
```
mister-ben, line 324: Should this be any browser doesn't support `inset`, and not just smart TVs? Smart TVs are the only real reason to be using such old browser versions, but it's not inconceivable that people might test things destined for a smart TV on an old browser on desktop/browserstack etc.

src/js/tracks/text-track-display.js
CarlosVillasenor: thanks Walter, I updated the PR, I took into account that there can be many text-track-cues in the updated code

src/js/tracks/text-track-display.js
```diff
@@ -329,6 +347,17 @@ class TextTrackDisplay extends Component {
...
     if (!this.player_.videoHeight() || !window.CSS.supports('inset-inline: 10px')) {
+      if (browser.IS_SMART_TV && !window.CSS.supports('inset', '10px')) {
+        const textTrackDisplay = this.el_;
+        const textTrackDisplayHeight = textTrackDisplay.getBoundingClientRect().height;
+
+        // textrack style updates, this styles are required to be inline
+        textTrackDisplay.style.position = 'relative';
+        textTrackDisplay.style.height = textTrackDisplayHeight + 'px';
+        textTrackDisplay.style.top = 'unset';
+        textTrackDisplay.style.bottom = '0px';
```
gkatsev, line 358:
```suggestion
        tryUpdateStyle(textTrackDisplay, 'position', 'relative');
        tryUpdateStyle(textTrackDisplay, 'height', textTrackDisplayHeight + 'px');
        tryUpdateStyle(textTrackDisplay, 'top', 'unset');
        tryUpdateStyle(textTrackDisplay, 'bottom', '0px');
```

this updates to use tryUpdateStyle for consistency with the rest of the code, but maybe that isn't really neccessary as the comment for tryUpdateStyle metions IE8 🤔

src/js/tracks/text-track-display.js
gkatsev: Yeah, should be a feature test only, unless there's a good reason to only target TVs?
Below should be updated to match as well

src/js/tracks/text-track-display.js
CarlosVillasenor: add this change in the last commit, can this be re-reviewed?

src/js/tracks/text-track-display.js
CarlosVillasenor: Updated the PR to support this changes on any browser, I got to move the update of the styles of textTrackDisplay to add support when user switches to full screen, can this be re-reviewed?

src/js/tracks/text-track-display.js
```diff
@@ -319,6 +319,33 @@ class TextTrackDisplay extends Component {
...
+      const vjsTextTrackCues = textTrackDisplay.querySelectorAll('.vjs-text-track-cue');
+
+      // Clear inline style before getting actual height of textTrackDisplay
+      textTrackDisplay.style = '';
+      const textTrackDisplayHeight = textTrackDisplay.getBoundingClientRect().height;
+
+      // vjsTextTrackCue style updates
+      if (vjsTextTrackCues.length > 0) {
+        vjsTextTrackCues.forEach((vjsTextTrackCue) => {
+          const insetStyles = vjsTextTrackCue.style.inset.split(' ');
```
wseymour15, line 334: Is it possible for the `inset` style to not be set on the cue? If so, we might want a check to isure it exists before we move forward.

CarlosVillasenor: In every case I have tried those are always there, nevertheless is a good thing what you mentioned I updated the PR to ensure those exists before we use those to avoid any error

src/css/components/_text-track.scss
```diff
@@ -46,3 +46,11 @@ video::-webkit-media-text-track-display {
   text-align: center !important;
   width: 80% !important;
 }
+
+.video-js .vjs-text-track-display > div {
```
wseymour15, line 50: What does the `div` here represent? Do we want this CSS set always? Or maybe we could add a class name to this `div` in the scenario we want so we know it is SMART TV related..

I just want to be sure this does not have any unexpected side effects..

CarlosVillasenor: It is a div that is dynamically generated without a classname it is the parent of the vjs-text-track-cue, all of this styles are just backup styles, are all overwritten by inline styles except in old chrome devices.

Screenshot 2024-08-21 at 13 35 50

CarlosVillasenor: But I updated the PR to add those styles only when required, can this be re-reviewed?

wseymour15: Cool, I would maybe make that name more descriptive "vjs-text-track-display-inset" or something like that. Otherwise, nice work!

wseymour15 (approved):

src/css/components/_text-track.scss
```diff
@@ -46,3 +46,10 @@ video::-webkit-media-text-track-display {
   text-align: center !important;
   width: 80% !important;
 }
+
+.video-js .vjs-text-track-display-inset {
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+}
```
gkatsev, line 55:
```suggestion
@supports (inset: 10px) {
  .video-js .vjs-text-track-display {
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
  }
}
```

src/js/tracks/text-track-display.js
```diff
@@ -319,6 +320,45 @@ class TextTrackDisplay extends Component {
...
+      const textTrackDisplay = this.el_;
+      const vjsTextTrackCues = textTrackDisplay.querySelectorAll('.vjs-text-track-cue');
+      const controlBarHeight = this.player_.controlBar.el_.getBoundingClientRect().height;
+      const playerHeight = this.player_.el_.getBoundingClientRect().height;
+
+      // Clear inline style before getting actual height of textTrackDisplay
+      textTrackDisplay.style = '';
+
+      // Add custom class to textTrackDisplay div child for not inset support styles
+      textTrackDisplay.firstChild.classList.add('vjs-text-track-display-inset');
```
gkatsev, line 334: an alternative to adding this class is using css `@supports`. Then, this line won't be necessary and it could all be done in css.
```suggestion
```

src/js/tracks/text-track-display.js
```diff
@@ -319,6 +320,45 @@ class TextTrackDisplay extends Component {
       }
       this.updateForTrack(descriptionsTrack);
     }
+
+    if (!window.CSS.supports('inset', '10px')) {
```
gkatsev, line 324: since the current place where inset is used is `updateDisplayOverlay`, should this block go in there? There's already a check there for inset support.

CarlosVillasenor: This condition is here because **updateDisplay** runs for every time the subtitles change unlike  **updateDisplayOverlay** & that is required since the subtitles change position a little bit every time they are generated.

gkatsev: ah, i see. I guess the comment for it is incorrect, probably a copypasta from this method.

src/js/tracks/text-track-display.js
CarlosVillasenor: updated this in the last commit, can this be re-reviewed?

gkatsev (approved):

PR #8886, opened 2024-10-08
feat: Add methods to add and remove <source> elements

alex-barstow:
## Description
It is useful to have methods for appending and removing `<source>` elements to the `<video>` element, as they are sometimes required to enable certain playback features, for example, using [Airplay with MSE](https://webkit.org/blog/15036/how-to-use-media-source-extensions-with-airplay).

## Specific Changes proposed
Add new methods-- `addSourceElement()` and `removeSourceElement()` to the player and tech. The former will take a source object and create and append a new `<source>` element to the `<video>` element, and the latter will take a source url and remove any `<source>` element with a matching `src`.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tech/html5.js
+++ b/src/js/tech/html5.js
@@ -786,16 +786,25 @@ class Html5 extends Tech {
   /**
    * Add a <source> element to the <video> element.
    *
-   * @param {Tech~SourceObject} [srcObj]
-   *        The source object to append as a <source> element. Must have a `src` and it is recommended to have a `type` property.
+   * @param {string} srcUrl
+   *        The URL of the video source.
+   *
+   * @param {string} [mimeType]
+   *        The MIME type of the video source. Optional but recommended.
    */
-  addSrcAsSourceElement(srcObj) {
-    if (!srcObj || !srcObj.src) {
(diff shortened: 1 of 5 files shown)


src/js/tech/html5.js
```diff
@@ -783,6 +783,62 @@ class Html5 extends Tech {
...
+   * @param {string} srcUrl
+   *        The URL of the source to remove.
+   *
+   * @return {boolean}
+   *         Returns true if the source element was successfully removed, false otherwise.
+   */
+  removeSourceElement(srcUrl) {
+    if (!srcUrl) {
+      log.error('Source URL is required to remove the source element.');
+      return false;
```
wseymour15, line 824: Is there a reason this function returns a boolean but `addSourceElement` does not?

alex-barstow: Good call, both probably should. The reason I didn't have `addSourceElement()` return a boolean is because originally I hadn't added that early return so a source element would always be added.

alex-barstow: Updated

wseymour15 (approved):

PR #8904, opened 2024-10-29
fix(text-track-settings): localization not correctly applied

amtins:
## Description

Localization is not applied correctly in fieldset labels and select options. As a result, the text track setting modal dialog is only half translated.

Screenshot from 2024-10-29 19-35-34

## Specific Changes proposed

- add `localize` at `label` level in `TextTrackFieldset`
- add `localize` at `option` level in `TextTrackSelect`
- add test cases

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-fieldset.js
+++ b/src/js/tracks/text-track-fieldset.js
@@ -87,7 +87,7 @@ class TextTrackFieldset extends Component {
         const label = Dom.createEl('label', {
           id,
           className: 'vjs-label',
-          textContent: selectConfig.label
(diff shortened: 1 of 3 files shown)


mister-ben (approved):
LGTM. And it reminds me, we really ought to show a placeholder cue when that dialog is open at least when there isn't one showing, otherwise the truncated dialog seems strange.

gkatsev (approved):

dzianis-dashkevich (approved):

PR #8906, opened 2024-11-02
Introduce Global hotkey support

sinaru:
## Description
The video player does not currently support global hotkeys. Hotkeys only work when we have focus on the player. This PR is to introduce global hotkey support. 

With this change, for example, if you land on the page with a video player, and does not have focus on any specific fields like form fields or the video player, you can perform a hotkey event like space bar key press which will play/pause video.

## Specific Changes proposed

Global hotkeys are not turned on by default. To turn on, need to set `userActions.globalHotkeys` option to `true`. As below:

```js
var player = videojs(vid, {
  userActions: {
    globalHotkeys: true,
    hotkeys: true
  }
});
```

## Requirements Checklist
  - [ ] Docs/guides updated


sinaru:
Hi @mister-ben sorry for getting back to you late. Yes I agree. I tried to add a test and wanted to run locally to check the test but I couldn't figure out how to run the specific test in isolation. Is there a specific command we can use so we can just tell test file path and line number? The only thing I have is a way to run the whole test suite. This is not practical when writing the test.

mister-ben:
If you change `QUnit.test()` to `QUnit.only()`, only that test runs.

sinaru:
@mister-ben I have added couple of tests.

adrums86:
Thanks for submitting a PR @sinaru ! A couple of thoughts here before merging this:
1. I think we want to add some documentation for this feature.
2. Have we considered how this could fit with accessibility standards/requirements for users?

sinaru:
Hi @adrums86 . I'm sorry for the late reply. Unfortunately I won't be able to find anymore time to work on this. In fact the only reason I proposed this feature is that we needed this feature in our company so thought it would be nice to have it available in the main repo.

Having said that, feel free to use the changes as you wish or do anything with this PR. 

Thanks.

sinaru:
@adrums86 I got some time to update the docs. I have added an example HTML in sandbox. Also created a PR to update the guides (https://github.com/videojs/videojs.com/pull/195).

Regarding accessibility standards/requirements, I have tested with a screen reader and I compared the behavior with a website like YouTube (that has global hotkey support). The behavior appears to be identical. I'm not sure what else I can check for this. Please let me know if you have suggestions. 

Thanks.

PR #8932, opened 2024-12-01
fix: registering new player component

### LINKED ISSUE #8925: Possible incorrect logic with registering new Player
I'm not sure, but shouldn't there be `some` instead of `every`?

"If at least one player is not disposed yet"

https://github.com/videojs/video.js/blob/62f38446a5f01e85739bdb7207c7e992b57c1c86/src/js/component.js#L2024-L2032

victordidenko:
Do not allow to register new Player component, if any instance of the current component still exists.

Fixes #8925


dzianis-dashkevich (approved):

gkatsev (approved):
LGTM. Thanks for the PR and your patience on getting it reviewed!

PR #8945, opened 2024-12-22
fix: hide mouse tooltip on touch devices when not scrubbing

phloxic:
Problem: the vjs-user-inactive class triggering the hiding of the tooltip is added much later on touch devices, therefore the mouse tooltip stays visible on its last seek target position whenever the user taps the screen to access the controls.

This change ensures that the mouse tooltip is only visible when the user is actually scrubbing.

### DIFF
--- a/src/css/components/_progress.scss
+++ b/src/css/components/_progress.scss
@@ -178,7 +178,8 @@
   display: block;
 }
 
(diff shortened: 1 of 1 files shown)


dzianis-dashkevich (approved):

adrums86 (approved):
Thanks for another great contribution @phloxic !

PR #8968, opened 2025-01-23
[WIP] Type enhancements

mister-ben:
Several changes for Typescript.

* Adds a jsdoc plugin to not break on `@template {xyz} T`. Replaces uses of T with `Class.<xyz>`
* Adds definitions for `ComponentOptions`, `SourceObject`, `PlayerOptions`, `MediaObject`
* Removes further uses of ~ namespacing in jsdoc, which tsc fails to understand 
* Changes track kinds to simple union types. Change the existing definitions to enums so we can use them as object lookups without conflicting with the standard definitions. 
* Use native `TextTrackCue`, `TextTrackCue` definitions.
* Exports `Player`, `PlayerOptions` and `Plugin` from `video.js`.
  * This makes it easier to add to the object than at present e.g. in a plugin to `declare module 'video.js' { interface Player {`
  * This sets a precedent - what else could/should be exported at this level? Should the names be more verbose?


src/js/video.js
```diff
@@ -617,3 +620,7 @@ videojs.url = Url;
 videojs.Error = VjsErrors;
 
 export default videojs;
+export {
+  Player,
```
mister-ben, line 624: This allows adding to the Player interface in plugins:

```ts
declare module 'video.js' {
  interface Player {
    touchOverlay: TouchOverlay;
    touchOverlayDirect: TouchOverlay;
    mobileUi: {
      (options?: MobileUiPluginOptions): void;
      VERSION: string;
    }
  }
}
```

src/js/component.js
```diff
@@ -589,22 +593,22 @@ class Component {
...
+   * @param {ComponentOptions} [options={}]
    *        The key/value store of options that will get passed to children of
    *        the child.
    *
    * @param {number} [index=this.children_.length]
    *        The index to attempt to add a child into.
    *
    *
-   * @return {Component}
+   * @return {T}
```
mister-ben, line 611: Renders like this in the API docs with the changes to jsdoc-typeof-plugin:
image

build/jsdoc-typeof-plugin.js
```diff
@@ -3,6 +3,16 @@
...
+    const templateMatch = /@template {(.*)} ([A-Z])\n/.exec(event.comment || '');
+
+    if (templateMatch) {
+      const className = templateMatch[1];
+      const templateName = templateMatch[2];
+
+      event.comment = event.comment.replace(new RegExp(`{(.*\\b)(${templateName})(\\b.*)}`, 'g'), `{$1typeof ${className}$3}`);
+    }
+
+    // \{.*\bT\b.*\}
```
amtins, line 15: Do you think it's possible to modify the regex so that this case is also handled?

```javascript
    class CustomComponent extends Component{
      __z(){}
    }
    const Custom = Component.registerComponent('CustomComponent', CustomComponent);
    const custom = new Custom(player);
```

Instead of:
Screenshot from 2025-02-16 19-57-07

We would have:
Screenshot from 2025-02-16 19-58-05

src/js/component.js
```diff
@@ -1979,10 +1983,10 @@ class Component {
    * @param {string} name
    *        The name of the `Component` to register.
    *
-   * @param {Component} ComponentToRegister
+   * @param {typeof Component} ComponentToRegister
    *        The `Component` class to register.
    *
-   * @return {Component}
+   * @return {typeof Component}
```
amtins, line 1989: Based on my previous comment, if the regex could handle `@template {typeof Component} T` we could have:

```javascript
  /**
   * 
   * @template {typeof Component} T
   * 
   * Register a `Component` with `videojs` given the name and the component.
   *
   * > NOTE: {@link Tech}s should not be registered as a `Component`. {@link Tech}s
   *         should be registered using {@link Tech.registerTech} or
   *         {@link videojs:videojs.registerTech}.
   *
   * > NOTE: This function can also be seen on videojs as
   *         {@link videojs:videojs.registerComponent}.
   *
   * @param {string} name
   *        The name of the `Component` to register.
   *
   * @param {T} ComponentToRegister
   *        The `Component` class to register.
   *
   * @return {T}
   *         The `Component` that was registered.
   */
```

build/jsdoc-typeof-plugin.js
mister-ben: Thanks! The latest commit should cover that.

dasa:
Is it possible that this PR will also fix #9059?

src/js/player.js
```diff
@@ -62,13 +62,154 @@ import './tech/html5.js';
...
+ * @property  {number} [breakpoints.xlarge=800]
+ * @property  {number} [breakpoints.huge=900]
+ * @property  {Array|Object} [children]
+ * @property  {boolean} [disablePictureInPicture]
+ *              Prevents picture-in-picture mode
+ * @property  {boolean} [enableDocumentPictureInPicture]
+ *              Enable picture-in-picture using the document picture-in-picture API if available
+ * @property  {boolean} [enableSmoothSeeking]
+ *              Smoother seeking experience
+ *              TODO: Why not?
```
amtins: Maybe this comment can be removed.

PR #8989, opened 2025-02-24
fix: controlText for text track modal

adrums86:
## Description
Fixes text track modal controlText.

## Specific Changes proposed
Only apply reset text to the `Reset` button and done text to the `Done` button.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-settings-controls.js
+++ b/src/js/tracks/text-track-settings-controls.js
@@ -16,10 +16,8 @@ class TrackSettingsControls extends Component {
     super(player, options);
 
     // Create DOM elements
-    const defaultsDescription = this.localize('restore all settings to the default values');
-
     const resetButton = new Button(player, {
-      controlText: defaultsDescription,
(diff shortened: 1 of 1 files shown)


src/js/tracks/text-track-settings-controls.js
```diff
@@ -28,14 +26,15 @@ class TrackSettingsControls extends Component {
 
     this.addChild(resetButton);
 
+    const doneText = this.localize('Done');
     const doneButton = new Button(player, {
-      controlText: defaultsDescription,
+      controlText: doneText,
```
gkatsev, line 31: if it's repeating the text content, we shouldn't need control text (though, worth verifying with a screen reader).

Essk: I've tested these changes in NVDA & Voiceover, there's no repeat, the changes here just ensure the done and reset buttons have the correct text. Previously the done button had the same text as the reset button, which was misleading.

Essk (approved):

PR #8998, opened 2025-02-28
feat: toJSON methods for text track serialization

wseymour15:
## Description

Adds `toJSON` method to both the `TextTrack` and `TextTrackList` classes. The `toJSON` method creates a serializable JSON object from the track or track list.

There was an issue when attempting to serialize (JSON.stringify()) `player.textTracks()` due to circular references in `textTracks.tech_`. These methods create serializable copies of textTracks so a user has a way to serialize this data.

Since JSON.stringify calls `toJSON` natively, it will use this new overidden `toJSON` method instead of the native method. `JSON.stringify(player.textTracks())` will now be a valid serialized string. Previously, it threw an error.

## Specific Changes proposed

- Add `toJSON` methods to `TextTrack` and `TextTrackList` classes
- Tests for the above methods

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-list.js
+++ b/src/js/tracks/text-track-list.js
@@ -58,8 +58,8 @@ class TextTrackList extends TrackList {
   }
 
   /**
-   * Makes the text track list serializable.
-   * This transforms each track into a serializable object.
+   * Creates a serializable array of objects that contains serialized copies
+   * of each text track.
    *
    * @return {Object[]} A serializable list of objects for the text track list
    */
--- a/src/js/tracks/text-track-list.js
+++ b/src/js/tracks/text-track-list.js
@@ -56,5 +56,32 @@ class TextTrackList extends TrackList {
       }
     }
   }
+
+  /**
+   * Makes the text track list serializable.
+   * This transforms each track into a serializable object.
+   *
+   * @return {Object[]} A serializable list of objects for the text track list
+   */
+  toJSON() {
+    const trackCopies = [];
+
+    this.tracks_.forEach((track) => {
+      const copy = track.toJSON();
+
+      trackCopies.push(copy);
+    });
+
+    return trackCopies;
+  }
+
+  /**
+   * A method to serialize the text track list.
+   *
+   * @return {string} The serialized list of text tracks
+   */
+  serialize() {
+    return JSON.stringify(this.toJSON());
+  }
 }
 export default TextTrackList;
--- a/src/js/tracks/text-track-list.js
+++ b/src/js/tracks/text-track-list.js
@@ -66,15 +66,5 @@ class TextTrackList extends TrackList {
   toJSON() {
     return this.tracks_.map((track) => track.toJSON());
   }
-
-  /**
-   * A method to serialize the text track list.
-   *
-   * @return {string} The serialized list of text tracks
-   */
-  serialize() {
-    // Stringify calls toJSON if it exists, so the method above will be used.
-    return JSON.stringify(this);
-  }
 }
 export default TextTrackList;
--- a/src/js/tracks/text-track-list.js
+++ b/src/js/tracks/text-track-list.js
@@ -64,15 +64,7 @@ class TextTrackList extends TrackList {
    * @return {Object[]} A serializable list of objects for the text track list
    */
   toJSON() {
-    const trackCopies = [];
-
-    this.tracks_.forEach((track) => {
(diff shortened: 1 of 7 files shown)


src/js/tracks/text-track.js
```diff
@@ -420,6 +420,36 @@ class TextTrack extends Track {
     this.cues.setCues_(this.cues_);
   }
 
+  /**
+   * Makes the text track serializable by removing circular dependencies.
+   *
+   * @return {Object} The track information as a serializable object
+   */
+  toJSON() {
```
gkatsev, line 428: we already have en existing implementation here: https://github.com/videojs/video.js/blob/8842d376ed3d834ad3ce0a5e662fe446513ea0bd/src/js/tracks/text-track-list-converter.js#L21
should it be used instead? Are there properties we'd want to keep that isn't in that list?

test/unit/tracks/text-track-list.test.js
```diff
@@ -5,6 +5,8 @@ import EventTarget from '../../../src/js/event-target.js';
 import TechFaker from '../tech/tech-faker';
 import sinon from 'sinon';
 
+// here
+
```
gkatsev, line 9: nit
```suggestion
```

gkatsev (commented):
Thanks! Some thoughts in comments

src/js/tracks/text-track-list.js
```diff
@@ -56,5 +56,32 @@ class TextTrackList extends TrackList {
...
+
+    return trackCopies;
+  }
+
+  /**
+   * A method to serialize the text track list.
+   *
+   * @return {string} The serialized list of text tracks
+   */
+  serialize() {
```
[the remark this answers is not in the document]
wseymour15: You're totally right. Calling `JSON.stringify(player.textTracks())` would have the same result as `player.textTracks().serialize()` for `TextTrackLists`.

I do not have a strong preference whether to keep this or not. This is basically just an alias to do the same thing as of now.

src/js/tracks/text-track.js
wseymour15: Great question (and this was not a function I realized existed 😅). The newer implementation does contain more properties (activeCues, cues), but not sure how vital this information is for the serialization.

I do not have a strong preference on which functionality to use to determine the serialized data (I lean towards having the extra properties), but I do think if we decide to use the old method, we should move it to the `TextTrack` class so it can be public.

src/js/tracks/text-track-list.js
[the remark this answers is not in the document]
gkatsev: I would be for limiting the api as much as possible :)

src/js/tracks/text-track.js
gkatsev: I think the best bet is to have this call out to `trackToJson_`, since usage of text track list converter can end up working with the native text track object as well.

I don't think `activeCues` matters, but that could be added similarly to how trackToJson_ currently handles `cues`

wseymour15: Makes sense, I updated to use the `trackToJson` method. I did update the name from `trackToJson_` as _I think_ that was trying to signal a private method, when it is not.

src/js/tracks/text-track.js
```diff
@@ -420,6 +421,16 @@ class TextTrack extends Track {
...
```
mister-ben, line 429: Should we be calling this `toJSON` when it's returning an object rather than a JSON string?

Noting we have this here too:
https://github.com/videojs/video.js/blob/8842d376ed3d834ad3ce0a5e662fe446513ea0bd/src/js/player.js#L4884-L4891

gkatsev: it should be fine. From what i see from native behavior, the `toJSON()` method returns a serializable object rather than strictly a json string.

src/js/tracks/text-track.js
```diff
@@ -420,6 +421,16 @@ class TextTrack extends Track {
     this.cues.setCues_(this.cues_);
   }
 
+  /**
+   * Makes the text track serializable by removing circular dependencies.
```
alex-barstow, line 425: I think this description should make clear that it returns a new object rather than mutating the track in place

alex-barstow (approved):

gkatsev (approved):

Issue #9006, opened 2025-03-13
[Mobile] Disable focus on click of Big Play Button


rtritto:
### Description

As default Video.js displays _Big Play_ (`vjs-big-play-button` CSS class).

When `vjs-big-play-button` is clicked, the `vjs-play-control` is focused:

https://github.com/user-attachments/assets/5a8cd8b2-e0bd-4150-8763-8dbac4bd093e

Related:
- https://github.com/videojs/video.js/discussions/8954
- https://github.com/mister-ben/videojs-mobile-ui/issues/185

### Reduced test case

https://stackblitz.com/edit/vercel-next-js-dykavrxx

### Steps to reproduce

1.
2.
3.

### What version of Video.js are you using?

8.22.0

### What browser(s) including version(s) does this occur with?

Firefox 135

### What OS(es) and version(s) does this occur with?

Windows 11

gkatsev:
Looks like the last relevant change (https://github.com/videojs/video.js/pull/4497) was specifically to skip changing focus for mouse events. Maybe, it should be updated to instead only change focus on keyboard events instead of skipping mouse events.

gkatsev:
Doing a stricter check for keyboard events is _probably_ fine, but we'd need to confirm that it continues to function for Screen Readers. So, potentially, the easiest is to exclude touch events, unless VoiceOver on iOS, for example, uses touch events. In which case, it gets a bit trickier.

May be easier to have stricter focus styles.

Frenzie:
After a quick peek I thought this might suffice, but it turns out the eventual event that reaches `handleClick` doesn't have clientX and clientY.

```diff
diff --git a/src/js/big-play-button.js b/src/js/big-play-button.js
index 2bc7b4a1..4cd0d799 100644
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -20,6 +20,7 @@ class BigPlayButton extends Button {
     this.setIcon('play');
 
     this.on('mousedown', (e) => this.handleMouseDown(e));
+    this.on('touchstart', (e) => this.handleMouseDown(e));
   }
 
   /**
```

https://github.com/videojs/video.js/blob/c7298d40a4632a6e9dfcd5a2f5cc3bbe92a78744/src/js/big-play-button.js#L49-L58

There's some documentation about how to work that out here: https://developer.mozilla.org/en-US/docs/Web/API/Touch/clientX

mister-ben:
Checking for the "`tap`" event should suffice.

[...]

PR #9015, opened 2025-03-20
fix: only change focus from BPB if not tap or mouse click

Frenzie:
References #9006. Minor extension of #4497.

## Specific Changes proposed
Don't set focus after tapping on the big play button

## Requirements Checklist
  - [x] Change has been verified in an actual browser (Chrome, Firefox, ~IE~)
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -46,8 +46,8 @@ class BigPlayButton extends Button {
   handleClick(event) {
     const playPromise = this.player_.play();
 
-    // exit early if clicked via the mouse
-    if (this.mouseused_ && 'clientX' in event && 'clientY' in event) {
+    // exit early if tapped or clicked via the mouse
+    if (event.type === 'tap' || this.mouseused_ && 'clientX' in event && 'clientY' in event) {
       silencePromise(playPromise);
 
       if (this.player_.tech(true)) {


mister-ben (approved):

OwenEdwards:
@Frenzie and @mister-ben - where _does_ focus go when the big play button is activated, if it doesn't move to the play/pause button in the control bar? For accessibility, it can't be left in an indeterminate state, and since the BPB is hidden when it is activated, moving focus to the other play/pause button made the most sense. I'm not clear what the expected outcome of this fix is if the user clicks or taps the BPB, but this idea of *not* moving focus to a control has been used in media players in the past as a workaround for the focus outline appearing on a control when some designers don't like the focus outline.

If that's what this fix is trying to achieve, then it would be _much better_ (and more accessible) to adopt the [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) approach, so that focus _does_ move to the play/pause button but the focus outline is only shown if it is needed (by users who are using a mouse or touch). See also the https://github.com/videojs/video.js/blob/main/sandbox/noUITitleAttributes.html.example example I created 4 years ago, back when focus-visible was polyfilled.

Frenzie:
Focus shifts to the video element.

PR #7699, opened 2022-03-29
fix(accessibility): fix broken aria menu

Noemite:
## Description
Fix for issue : https://github.com/videojs/video.js/issues/7688 by @dave105010
Remove role="menu" in html on hidden player options if they are not used (no chapters added or no CC added etc.).

## Specific Changes proposed
Add/remove role="menu"

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [x] Change has been verified in an actual browser (Chrome, Firefox, IE)
image
image

  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
- [ ] Reviewed by Two Core Contributors

PR #9040, opened 2025-05-09
fix: invalid HTML

ahosgood:
## Description
Fixes HTML validation

## Specific Changes proposed
Maintain the role of "menu" on all MenuButton components which removes the HTML validation issue of:

> An element with role=menuitemradio must be contained in, or owned by, an element with role=menubar or role=menu or role=group

...as MenuItem components can have a role.

Instead, add `aria-hidden="true"` on the MenuButton components without items to hide them from assistive technologies.

Tested in https://validator.w3.org/ with the included sandbox example.

## Requirements Checklist
  - [ ] Docs/guides updated


opa26033 (commented):
Op bro,

mister-ben:
Menus that are not disabled do have `role="menu"`. This would recreate the problem fixed in https://github.com/videojs/video.js/pull/7699

ahosgood:
Removing `role="menu"` when the menu is not used doesn't feel like a robust solution because it causes this invalid HTML.

I feel better options would be either to add `aria-hidden="true"` to any of the menus that are disabled or just to remove the menu element from the HTML entirely.

ahosgood:
@mister-ben Hopefully this is now a better approach that avoids recreating #7699 but fixes the invalid HTML 🤞🏼

src/js/menu/menu-button.js
```diff
@@ -31,6 +31,8 @@ class MenuButton extends Component {
   constructor(player, options = {}) {
     super(player, options);
 
+    this.el_.setAttribute('role', 'menu');
+
```
mister-ben:
```suggestion

```
Leaving `role="menu"` on the menu rather than the button ought to be more semantic.

src/js/menu/menu-button.js
```diff
@@ -89,11 +91,11 @@ class MenuButton extends Component {
 
     if (this.items && this.items.length <= this.hideThreshold_) {
       this.hide();
-      this.menu.contentEl_.removeAttribute('role');
+      this.menu.contentEl_.setAttribute('aria-hidden', 'true');
```
mister-ben:
```suggestion
     this.menu.contentEl_.removeAttribute('role');
      this.menu.contentEl_.setAttribute('hidden', 'hidden');
      this.menu.children().forEach((menuItem) => {
        menuItem.dispose();
      });
```
aria-hidden doesn't prevent errors in common accessibility tools, unfortunately. This is removing the menu role, and also removing the children that won't be displayed to avoid the validation errors for having menu items without a menu.

src/js/menu/menu-button.js
```diff
@@ -89,11 +91,11 @@ class MenuButton extends Component {
...
 
     } else {
       this.show();
-      this.menu.contentEl_.setAttribute('role', 'menu');
+      this.menu.contentEl_.removeAttribute('aria-hidden');
```
mister-ben:
```suggestion
      this.menu.contentEl_.setAttribute('role', 'menu');
      this.menu.contentEl_.removeAttribute('hidden');
```

mister-ben (changes requested):
The current version still shows errors in the [WAVE web accessibility tool](https://wave.webaim.org/extension/), which we know video.js users use from other issue reports: `An ARIA menu does not contain required menu items.`
The changes I've suggested seem to resolve both the W3C validation and WAVE errors.
(I've not updated tests in these suggestions).

PR #8988, opened 2025-02-24
feat: Improve SmartTV scrubbing behavior

bzizmo:
## Description
This PR refactors the handling of pendingSeekTime to improve scrubbing behavior on SmartTVs. It introduces a more structured approach for handling seek updates, ensuring better UX for SmartTV users who navigate via remote controls.

## Specific Changes proposed

- Refactored pendingSeekTime handling in SeekBar, replacing direct property access with a getter/setter method.
- Introduced disableSeekWhileScrubbingOnSTV option to prevent immediate seeks while scrubbing on SmartTVs.
- Updated PlayProgressBar and CurrentTimeDisplay to respect pendingSeekTime.
- Adjusted stepForward and stepBack to modify pendingSeekTime without immediately seeking, unless explicitly triggered.
- Ensured seek confirmation occurs when the user presses Enter or completes scrubbing.
- Added unit tests covering new SmartTV scrubbing behavior.

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [ ] Change has been verified in an actual browser (Chrome, Firefox, IE)
  - [x] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))
  - [ ] Has no DOM changes which impact accessiblilty or trigger warnings (e.g. Chrome issues tab)
  - [ ] Has no changes to JSDoc which cause `npm run docs:api` to error
- [ ] Reviewed by Two Core Contributors

PR #9110, opened 2025-10-18
fix: currentTimeDisplay shows 0:00 after seek

phloxic:
## Description

Steps to reproduce originally posted by @johanberonius at https://github.com/videojs/video.js/issues/8988#issuecomment-3402464579

- Pause player
- Click progress bar and hold mouse button down
- Current time display is updated correctly
- Release mouse button
- Current time display is reset to zero
- Play video
- Current time display is updated again

This patch integrates to proposed workaround by @johanberonius.

The issue was not caused by adding the option
`disableSeekWhileScrubbingOnMobile` per se, but is a side-effect of the improvements added in 77c99d2829f56d1037361c33ccd96c6ea966ecb7.

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/control-bar/progress-control/seek-bar.js
+++ b/src/js/control-bar/progress-control/seek-bar.js
@@ -56,7 +56,7 @@ class SeekBar extends Slider {
     super(player, options);
 
     this.shouldDisableSeekWhileScrubbing_ = shouldDisableSeekWhileScrubbing;
-    this.pendingSeekTime_ = null;
+    this.pendingSeekTime = this.getCurrentTime_;
 
     this.setEventHandlers_();
   }


phloxic:
Closing this. Somehow managed to overlook that this prevents seeking on mobile, or at least Android.

Sorry for the noise.

johanberonius:
My workaround was assuming the new feature was disabled, probably wont work with the feature enabled.

Also, I don't think the changes here in the constructor is doing exactly the same thing as the workaround either.

phloxic:
> My workaround was assuming the new feature was disabled, probably wont work with the feature enabled.

If by new feature you mean the `disableSeekWhileScrubbingOnSTV` option, I don't know, I'm not familiar with the STV player.

> Also, I don't think the changes here in the constructor is doing exactly the same thing as the workaround either.

I think they pretty much do, once one gets the conditionals right. I tweaked [the patch](https://github.com/phloxic/video.js/commit/2353f7f25959ae6aee1c0dbd771b9ee48fc08987) and it works, let's say better, but see the caveats. Bisecting it's something in [this change](https://github.com/videojs/video.js/commit/77c99d2829f56d1037361c33ccd96c6ea966ecb7) which triggers the problems. But that's too deep a dive to take for me atm.

PR #9160, opened 2026-03-04
fix: force text track updateDisplay to recalculate on playerresize

Frenzie:
Otherwise processCues reuses the old position and size.

## Description
1. Go to https://legacy.videojs.org/advanced/?video=bipbop-advanced
2. Activate the cc1 caption track
3. Pause to more easily see the effect
4. Activate fullscreen or resize the window such that the player resizes
5. The caption position and size isn't recalculated

<kbd>Screenshot_20260304_163748</kbd>

## Specific Changes proposed
Force recalculation in vtt.js in this function:

```js
  // Determine if we need to compute the display states of the cues. This could
  // be the case if a cue's state has been changed since the last computation or
  // if it has not been computed yet.
  function shouldCompute(cues) {
    for (var i = 0; i < cues.length; i++) {
      if (cues[i].hasBeenReset || !cues[i].displayState) {
        return true;
      }
    }
    return false;
  }
```

## Requirements Checklist
  - [ ] Docs/guides updated

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -273,9 +273,10 @@ class TextTrackDisplay extends Component {
    * @listens Player#useractive
    * @listens Player#userinactive
    */
-  updateDisplay() {
+  updateDisplay(event) {
     const tracks = this.player_.textTracks();
     const allowMultipleShowingTracks = this.options_.allowMultipleShowingTracks;
+    const forceCueRecompute = event && event.type === 'playerresize';
 
     this.clearDisplay();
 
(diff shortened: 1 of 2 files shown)


mister-ben:
I can't reproduce the issue on Chrome or Firefox. It's resizing without your change. Is there anything more to know to reproduce that?

Frenzie:
That means you selected any of the subtitle tracks, not the cc1 caption track. ;-)

It happens in all browsers because the issue is that it purposefully skips computing the size/position when resizing.

mister-ben:
Sorry, I see now!

mister-ben (approved):

Frenzie:
No problem and thanks! While I have your attention, would you mind taking a look at my PRs against vtt.js?

* https://github.com/videojs/vtt.js/pull/69
* https://github.com/videojs/vtt.js/pull/71

PR #9166, opened 2026-04-03
Security Audit Fix

danrossi:
## Description
There was security audit errors that were critical related to installing a vulnerable Axios package via the access-sniff package. The browserify package security issue is low but unfixable and won't override the set packages. It is needed only for a test file so should be removed maybe.

There is alot of deprecated warnings that may be hard to fix. "videojs-standard" uses an old eslint which has an api change in 8.0.0. 

## Specific Changes proposed
This updates packages with overrides to fix the security audit issues

The access-sniff accessibility test has been changed to a pa11y accessibility test report. So that old outdated package can be removed. Some overrides wouldn't fix it's problem so safe to remove.


danrossi:
I cleaned up the overrides better. However there is no fix for browserify because the elliptic package is still vulnerable. If browserify is just used for a test file it might be safe to remove for now. 

```
elliptic  *
Elliptic Uses a Cryptographic Primitive with a Risky Implementation - https://github.com/advisories/GHSA-848j-6mx2-7j84
fix available via `npm audit fix`
node_modules/elliptic
  browserify-sign  >=2.4.0
  Depends on vulnerable versions of elliptic
  node_modules/browserify-sign
    crypto-browserify  >=3.4.0
    Depends on vulnerable versions of browserify-sign
    Depends on vulnerable versions of create-ecdh
    node_modules/crypto-browserify
      browserify
      Depends on vulnerable versions of crypto-browserify
      node_modules/browserify
  create-ecdh  *
  Depends on vulnerable versions of elliptic
  ````

danrossi:
There is two Axios flaws now along with the other security audit fixes. Video.JS pulls this in for an old accessibility tester package. This migrates to a different tester.

danrossi:
a lock file update fixes the xml vulnerability issue. There is just browserify which is used for a test. Might be easier to just remove it.

PR #9173, opened 2026-04-20
feat(poster): add support for marking player as maincontent

### LINKED ISSUE #9171: Option to make player as main content (eager/prioritized loading)
### Description

Currently the poster image is marked as `loading="lazy"`, which isn't desirable all the times. In cases where the video is the main content of the page it should be marked as `loading="eager"` and `fetchpriority="high"` to make a good user experience and SEO.

https://github.com/videojs/video.js/issues/8038

### Reduced test case

https://legacy.videojs.org/

### Steps to reproduce

1. Go to https://legacy.videojs.org/
2. Check the `loading` property of the posters `img` element.

### Errors

_No response_

8.23.4

### Video.js plugins used.

_No response_

Brave 1.89.137 Chromium: 147.0.7727.102 

Ubuntu 22

kontrollanten:
## Description
When maincontent is set to true the poster img element will get [loading=eager](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#eager) and [fetchpriority=high](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/fetchpriority#high) to decrease [LCP](https://web.dev/articles/lcp) time.

#### maincontent: true

```
    var player = videojs(vid, { maincontent: true });
```

image

#### Default mode

```
    var player = videojs(vid);
```

image

## Requirements Checklist
  - [ ] Docs/guides updated

I'm not sure if any example should be created, and if so where to put it. The same goes with the documentation.

closes #9171

### DIFF
--- a/src/js/poster-image.js
+++ b/src/js/poster-image.js
@@ -27,9 +27,9 @@ class PosterImage extends ClickableComponent {
   constructor(player, options) {
     super(player, options);
 
-    const { maincontent } = (options && options.playerOptions) || {};
+    const { mainContent } = (options && options.playerOptions) || {};
 
-    this.isMainContent = maincontent;
+    this.isMainContent = mainContent;
 
     this.update();
 
--- a/src/js/poster-image.js
+++ b/src/js/poster-image.js
@@ -27,6 +27,10 @@ class PosterImage extends ClickableComponent {
   constructor(player, options) {
     super(player, options);
 
+    const { maincontent } = (options && options.playerOptions) || {};
+
+    this.isMainContent = maincontent;
+
     this.update();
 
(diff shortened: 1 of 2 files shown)


mister-ben:
At the time `loading="lazy"` was added, Lighthouse would complain if it were absent, even on a prominent player. Iirc it was said that browsers would disregard the lazy directive on elements it deemed to have priority. Since the recommendation has changed, this change is more than reasonable.

Nit: The option name should be camel case like other options.

kontrollanten:
Thanks for reviewing! I've now changed `maincontent` to `mainContent`.

mister-ben (approved):

PR #9177, opened 2026-05-06
fix: handle Mac trackpad tap-to-click in isSingleLeftClick

### LINKED ISSUE #7736: Missing case in Dom.isSingleLeftClick
## Description
On mac, tapping the touchpad may be interpreted in two ways and thus, passed down to the browser in different ways (briefly discussed [here](https://stackoverflow.com/questions/72004625/inconsistent-html-mouseevent-with-mac-trackpad?noredirect=1)), but one of them is not accounted for in `Dom.isSingleLeftClick`.
More specifically, when `button === 0 && buttons === 1`.
Therefore, `handleMouseDown` of `SeekBar` does not recognize such an event as click and the event gets dismissed by `Dom.isSingleLeftClick(event)` (line 798 [here](https://docs.videojs.com/utils_dom.js.html)) called in the body of `handleMouseDown` (line 252 [here](https://docs.videojs.com/control-bar_progress-control_seek-bar.js.html)).

It then may take 2-3 (up to 8, sometimes) taps to actually handle mouse down (and to change volume or current time, as consequence).

Returning true if `button === 0 && buttons === 1` should do the trick 😊

## Steps to reproduce
1. On a mac, try tapping (not clicking) around the video search bar or the volume bar to change volume or current time. Can be observed also on the official website's player.

### Expected
On each tap, the slider, be it sound or time control, changes its position accordingly

### Actual
Due to the nature of the hardware (briefly discussed [here](https://stackoverflow.com/questions/72004625/inconsistent-html-mouseevent-with-mac-trackpad?noredirect=1)), 2-3 taps are needed to actually change the position

### Error output
None

#### videojs
7.3.33
#### browsers
Tried on latest desktop versions of Chrome, Opera, Firefox, Safari and Edge
#### OSes
Mac 12.3.1
### plugins
None

okxint:
## Summary

Fixes #7736.

Mac trackpad tap-to-click fires `mousedown` with `{button: 0, buttons: 0}` instead of the expected `{button: 0, buttons: 1}`. The current `isSingleLeftClick` has separate blocks for `mouseup` and `mousedown` handling this `{0, 0}` case — this PR consolidates them into a single `event.type !== 'mousemove'` guard.

The `{0, 0}` rejection is only meaningful for `mousemove` events (detecting button release during a drag). For all other event types (`mousedown`, `mouseup`, etc.), `{button: 0, buttons: 0}` is a valid single left click.

This approach was suggested by @gkatsev in https://github.com/videojs/video.js/issues/7736#issuecomment-1115386139.

## Changes

- `src/js/utils/dom.js`: Replace `event.type === 'mouseup'` with `event.type !== 'mousemove'` for the `{0, 0}` check, removing the separate `mousedown` block

## Test plan

- [x] Existing tests cover all three event types: `mousemove {0,0}` → false, `mouseup {0,0}` → true, `mousedown {0,0}` → true
- [ ] Manual verification on Mac with tap-to-click enabled — seek bar and volume slider respond on first tap

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -775,15 +775,12 @@ export function isSingleLeftClick(event) {
     return true;
   }
 
-  // `mouseup` event on a single left click has
-  // `button` and `buttons` equal to 0
-  if (event.type === 'mouseup' && event.button === 0 &&
+  // `button` and `buttons` are both 0 on `mouseup` (release) and on
+  // `mousedown` when a Mac trackpad tap-to-click fires {button:0, buttons:0}.
+  // The only event type where {0,0} should be rejected is `mousemove`,
+  // because it signals that the button was released during a drag.
+  if (event.type !== 'mousemove' && event.button === 0 &&
       event.buttons === 0) {
-    return true;
-  }
-
-  // MacOS Sonoma trackpad when "tap to click enabled"
-  if (event.type === 'mousedown' && event.button === 0 && event.buttons === 0) {
     return true;
   }
 


okxint:
Still valid and relevant — bumping to keep this open. Happy to rebase or adjust if needed.

PR #9178, opened 2026-05-10
fix: use aria-live="polite" for descriptions text track display

### LINKED ISSUE #7843: Text Track Display Accessibility Issues
### Description

Videojs auto-generates the following div that is always updating to contain the current video caption:

texttrackdisplay

This has the following issues when viewing a video using a screen reader, for instance JAWS or NVDA on a browser like Firefox or Chrome:

1. The div is aria-live="assertive" which means whenever this div is updated, all other announcements being made by the screen reader will be interrupted for the announcement of what's in the updated div. Also, if a notification pops up on the page or the user navigates to or interacts with another part of the page like a menu, whatever announcements are being made by the screen reader will be interrupted by the playing video's captions.
2. The screen reader wants to say “text track display {caption}”, but before it can read the caption it’ll often be interrupted by the next “text track display {caption}” so the result is it saying “text track display” over and over instead of the captions.

I believe this is a violation of [WCAG 4.1.3 - Status Messages](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html) because assistive technology can't read the captions completely, and other important announcements that may be happening on the page will be interrupted by the video's captions. For those relying on assistive technology, important messages about the page and the video's captions get interrupted or even missed entirely.

The suggested fix here is making the div aria-live="polite" and removing the aria-label from this generated div.

P.S. This was using the azure media player

### Reduced test case

_No response_

### Steps to reproduce

1. Open a page in Chrome or Firefox containing a Videojs video that has captions.
2. Open a screen reader like JAWS or NVDA
3. Play the video with captions on
4. Notice that instead of reading the caption sometimes it just reads "Text Track Display" or not the whole caption

### Errors

_No response_

latest

### Video.js plugins used.

_No response_

Chrome 103.0.5060.114

Windows 11

manichandra:
Fixes #7843                                                                                                                             
                                                                  
  Using `aria-live="assertive"` on the text track display container caused
  screen readers to interrupt all other announcements every time a descriptions                                                           
  cue updated. This made audio descriptions inaccessible in practice — users   
  lost context from other page announcements.                                                                                             
                                                                  
  Changed to `aria-live="polite"` so descriptions are queued and announced
  without interrupting ongoing screen reader output, per WCAG 4.1.3.                                                                      
  Captions and subtitles tracks correctly remain at `aria-live="off"` since
  users relying on those tracks do not need screen reader announcement of                                                                 
  caption text.                                                                                                                           
   
  Specific Changes proposed:                                                                                                              
  - `src/js/tracks/text-track-display.js`: Changed `aria-live="assertive"`
    to `aria-live="polite"` in `updateDisplay()` for descriptions tracks  
  - `test/unit/tracks/text-track-display.test.js`: Added QUnit test verifying                                                             
    aria-live is "off" for captions tracks and "polite" for descriptions tracks
                                                                                                                                          
  Requirements Checklist:                                                                                                                 
    - [ ] Docs/guides updated                                                                                                             
    - [ ] Example created    
    - [x] Has no DOM changes which impact accessibility or trigger warnings                                                               

### DIFF
--- a/src/js/tracks/text-track-display.js
+++ b/src/js/tracks/text-track-display.js
@@ -320,8 +320,8 @@ class TextTrackDisplay extends Component {
       }
       this.updateForTrack(captionsSubtitlesTrack);
     } else if (descriptionsTrack) {
-      if (this.getAttribute('aria-live') !== 'assertive') {
-        this.setAttribute('aria-live', 'assertive');
+      if (this.getAttribute('aria-live') !== 'polite') {
+        this.setAttribute('aria-live', 'polite');
       }
       this.updateForTrack(descriptionsTrack);
     }
--- a/test/unit/tracks/text-track-display.test.js
+++ b/test/unit/tracks/text-track-display.test.js
@@ -620,3 +620,39 @@ if (!Html5.supportsNativeTextTracks()) {
     player.dispose();
   });
 }
+
+QUnit.test('aria-live is set to "off" for captions/subtitles tracks and "polite" for descriptions tracks', function(assert) {
+  const player = TestHelpers.makePlayer();
+  const display = player.textTrackDisplay;
+
+  // Stub clearDisplay and updateForTrack so no WebVTT processing occurs
+  display.clearDisplay = () => {};
+  display.updateForTrack = () => {};
+
+  const tracks = player.textTracks();
+
+  // Add a captions track and make it active
+  player.addRemoteTextTrack({ kind: 'captions', language: 'en', label: 'English' }, true);
+  tracks[0].mode = 'showing';
+
+  display.updateDisplay();
+  assert.strictEqual(
+    display.el_.getAttribute('aria-live'),
+    'off',
+    'aria-live should be "off" when a captions track is active to avoid interrupting screen reader announcements'
+  );
+
+  // Switch to a descriptions track
+  tracks[0].mode = 'disabled';
+  player.addRemoteTextTrack({ kind: 'descriptions', language: 'en', label: 'English Descriptions' }, true);
+  tracks[1].mode = 'showing';
+
+  display.updateDisplay();
+  assert.strictEqual(
+    display.el_.getAttribute('aria-live'),
+    'polite',
+    'aria-live should be "polite" for descriptions tracks so announcements queue without interrupting other screen reader output'
+  );
+
+  player.dispose();
+});


manichandra:
Can we pleas get this merged ?

Essk:
Thanks for picking this up, and for including a test. Descriptions accessibility gets very little attention and the effort is appreciated. Having dug into the history here, though, I don't think we should make this change as it stands.

A few things came out of reviewing the history and prior art:

**The `assertive` level for descriptions appears to be deliberate.** It survived the fix that created the current split (685404d0), where `aria-live` was explicitly removed for captions/subtitles and kept only for descriptions. #7843, which this PR cites, describes caption announcements interrupting each other; captions have been `aria-live="off"` since that change, so the situation reported there doesn't apply to the current code.

**Description cues are time-sensitive, opted-in content.** They describe what is on screen right now, and the user chose to turn them on, so being interrupted by them is the consented-to behaviour. Under `polite`, cues queue behind other speech: an overrunning cue delays the next one, drift compounds, and in some configurations polite regions are skipped entirely (VoiceOver read-all, some Braille output). A description arriving after the moment it describes can be worse than none. The queue-clearing behaviour of `assertive` is arguably the desirable one in this specific case.

**Prior art agrees.** Able Player, the reference implementation for text-based audio description, uses `aria-live="assertive"` with `aria-atomic="true"` for its description region, and goes further: it removes `aria-live` from its own status region while a description is being spoken, specifically "to prevent description from being interrupted". Its priority ordering is the opposite of this PR's premise.

One note on the justification: WCAG 4.1.3 covers status messages (results of an action, waiting states, progress, errors). Description cues aren't status messages, they're primary media content, and 4.1.3's caution against `assertive` is scoped to content that is "not important and time-sensitive", which description cues are.

That said, `assertive` isn't a perfect answer either. Screen reader handling of politeness is inconsistent (JAWS treats assertive regions as polite, TalkBack applies its own behaviour regardless), and the real gap identified by the W3C Media Accessibility User Requirements is user-controllable overflow handling: options to continue, stop, or pause the video when a description overruns ([TVD-4], [EVD-2]/[EVD-3]). Able Player implements pause-during-description for exactly this reason. A PR making the politeness level a user-facing option, or adding pause/resume for overrunning cues, would be a genuinely valuable improvement and I'd be happy to see this evolve in that direction.


@gkatsev @mister-ben you may have more context or a different read on the history here.

PR #9196, opened 2026-05-24
fix(player): add null check for this.tech_ in fullscreen methods

xxiaoxiong:
## Problem

When `exitFullscreen()` is called after the player has been disposed, `this.tech_` is null and the code crashes with `TypeError: Cannot read properties of null`.

## Root cause

The fullscreen methods access `this.tech_` without a null guard. During rapid dispose + fullscreen interaction, the tech is destroyed before the fullscreen handler runs.

## Changes

- `src/js/player.js`: Added `if (!this.tech_) return;` guard at the top of fullscreen-related methods, matching the defensive pattern used elsewhere in video.js.

## Why

Null-guarding follows the existing defensive programming pattern throughout video.js. The alternative — ensuring cleanup order — would require impractical async coordination.

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -3035,7 +3035,7 @@ class Player extends Component {
       }
 
       return promise;
-    } else if (this.tech_.supportsFullScreen() && !this.options_.preferFullWindow === true) {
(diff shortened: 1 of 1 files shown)


amtins (approved):
@xxiaoxiong thanks for the PR.

PR #9217, opened 2026-07-31
fix: focus the play toggle instead of the tech element on Edge to avoid a black frame with hardware-accelerated protected playback

BCovePW:
On Microsoft Edge, `BigPlayButton` and `PosterImage` focused the raw `<video>` (tech) element as playback started (the mouse/tap path). With hardware-accelerated protected (DRM/EME) playback this prevents the protected video surface from being presented: audio plays over a black frame until a later repaint or focus change. Focus the control-bar play toggle instead on Edge, mirroring the existing keyboard path.

Pointer-path counterpart to the fix for #6270; #6318/#6508 redirected only the keyboard path, so the mouse/tap `tech.focus()` survived and regressed on Chromium Edge with hardware-accelerated DRM.

## Description

When a user starts playback by clicking (mouse) or tapping the big play button or the poster, `handleClick()` calls `this.player_.tech(true).focus()`, moving focus to the raw `<video>` element. On Microsoft Edge, focusing the `<video>` at the moment a hardware-accelerated **protected** (EME/DRM) surface is being composited stops that surface from being presented — the user gets audio over a black frame until something forces a repaint (e.g. a resize, a later focus change, entering picture-in-picture, or clicking elsewhere).

This is the same failure mode as #6270. #6318 originally guarded this exact mouse/tap call site in big-play-button.js (skipping tech.focus() when (IE_VERSION || IS_EDGE) && sourceIsEncrypted), and #6508 added the same guard to poster-image.js. Both guards were later removed in #7701 ("remove ie-specific code") — that removal is the regression this PR fixes. Rather than reinstating the old sourceIsEncrypted-gated skip — which dropped focus (hurting keyboard accessibility) and likely wouldn't fire in a click-to-play flow anyway, since EME sessions are typically created after playback starts — this restores the protection in a better form by redirecting focus to the control-bar play toggle. The keyboard-activation path's play-toggle focus is a separate lineage and is untouched.

The fix makes the mouse/tap path behave like the keyboard path **on Edge only**: focus the control-bar play toggle (falling back to the player root) instead of the tech element. All other browsers are unaffected — the branch is guarded by `browser.IS_EDGE`, so on Chrome/Firefox/Safari the existing `tech.focus()` behavior is unchanged.

**Reproduction:** on Microsoft Edge with a hardware-accelerated protected source, start playback with a mouse click → audio plays over a black frame. Observed on both the legacy EdgeHTML engine (the environment #6270 was originally reported on) and current Chromium Edge. The black frame clears on autoplay, keyboard start, clicking elsewhere, pause/resume from the control bar, PIP, or disabling hardware acceleration — all of which either avoid the `<video>` receiving focus at play-start or force a recomposite. Redirecting play-start focus to the play toggle resolves it while keeping playback on the hardware/DRM path.

## Specific Changes proposed

- `src/js/big-play-button.js` — in `handleClick()`, the mouse/tap branch now focuses the control-bar play toggle (or the player root as a fallback) instead of `this.player_.tech(true).focus()` when `browser.IS_EDGE`. Non-Edge browsers keep the existing `tech.focus()`.
- `src/js/poster-image.js` — same guarded change in `handleClick()`.
- Keyboard-activation path is unchanged (it already targets the play toggle via #6318 / #6508).
- Uses the existing `browser.IS_EDGE` export from `src/js/utils/browser.js`; no new detection logic.
- Optional fallback (behind the same Edge guard): force a single-frame repaint on the first `play` (toggle a `transform` on the next `requestAnimationFrame`) for cases where redirecting focus alone is insufficient.
- Unit tests added covering the Edge branch (focus lands on the play toggle, not the tech element) and confirming non-Edge behavior is unchanged.

## Accessibility

The change **redirects** focus rather than dropping it: on Edge, play-start focus moves to the control-bar play toggle — an existing interactive control — instead of the `<video>`. Keyboard operability and focus visibility are preserved, and the behavior now matches the keyboard-activation path. No focus is left detached and no elements are added or removed from the DOM.

## Requirements Checklist

- [x] Bug fixed
  - [x] Change has been verified in an actual browser — Microsoft Edge (Chromium Edge 150 and legacy EdgeHTML 18); Chrome / Firefox / Safari behavior is unchanged by construction (guarded on `browser.IS_EDGE`)
  - [ ] Docs/guides updated (not applicable — no public API change)
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0)) — can add on request
  - [x] Has no DOM changes which impact accessibility or trigger warnings — focus is redirected to an existing control; keyboard a11y preserved

### DIFF
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -3,6 +3,7 @@
  */
 import Button from './button.js';
 import Component from './component.js';
+import * as browser from './utils/browser.js';
 import {isPromise, silencePromise} from './utils/promise';
 
 /**
@@ -45,20 +46,29 @@ class BigPlayButton extends Button {
    */
   handleClick(event) {
     const playPromise = this.player_.play();
+    const cb = this.player_.getChild('controlBar');
+    const playToggle = cb && cb.getChild('playToggle');
 
     // exit early if tapped or clicked via the mouse
     if (event.type === 'tap' || this.mouseused_ && 'clientX' in event && 'clientY' in event) {
       silencePromise(playPromise);
 
-      if (this.player_.tech(true)) {
+      // On Microsoft Edge, moving focus to the <video> (tech) element as playback
+      // starts prevents a protected (DRM/EME) video surface from being presented:
+      // audio plays but the frame stays black until a later repaint or focus change.
+      // Focus the control-bar play toggle instead of the tech on Edge, mirroring the
+      // keyboard branch below. This is the pointer-path counterpart to the fix for
+      // https://github.com/videojs/video.js/issues/6270 (#6318/#6508 redirected only
+      // the keyboard path; the mouse/tap tech.focus() survived and regressed on
+      // Chromium Edge with hardware-accelerated DRM).
+      if (browser.IS_EDGE && playToggle) {
+        playToggle.focus();
+      } else if (this.player_.tech(true)) {
         this.player_.tech(true).focus();
       }
 
       return;
     }
-
-    const cb = this.player_.getChild('controlBar');
-    const playToggle = cb && cb.getChild('playToggle');
 
     if (!playToggle) {
       this.player_.tech(true).focus();
--- a/src/js/big-play-button.js
+++ b/src/js/big-play-button.js
@@ -61,8 +61,8 @@ class BigPlayButton extends Button {
       // https://github.com/videojs/video.js/issues/6270 (#6318/#6508 redirected only
       // the keyboard path; the mouse/tap tech.focus() survived and regressed on
       // Chromium Edge with hardware-accelerated DRM).
-      if (browser.IS_EDGE && playToggle) {
-        playToggle.focus();
+      if (browser.IS_EDGE) {
+        (playToggle || this.player_).focus();
       } else if (this.player_.tech(true)) {
         this.player_.tech(true).focus();
       }
--- a/src/js/poster-image.js
+++ b/src/js/poster-image.js
@@ -4,6 +4,7 @@
 import ClickableComponent from './clickable-component.js';
 import Component from './component.js';
 import * as Dom from './utils/dom.js';
+import * as browser from './utils/browser.js';
 import {silencePromise} from './utils/promise';
 
 /** @import Player from './player' */
@@ -166,7 +167,17 @@ class PosterImage extends ClickableComponent {
       return;
     }
 
-    if (this.player_.tech(true)) {
+    // On Microsoft Edge, moving focus to the <video> (tech) element as playback
+    // starts prevents a protected (DRM/EME) video surface from being presented:
+    // audio plays but the frame stays black until a later repaint or focus change.
+    // Focus the control-bar play toggle instead of the tech on Edge. See
+    // https://github.com/videojs/video.js/issues/6270.
+    const cb = this.player_.getChild('controlBar');
+    const playToggle = cb && cb.getChild('playToggle');
+
+    if (browser.IS_EDGE && playToggle) {
+      playToggle.focus();
+    } else if (this.player_.tech(true)) {
       this.player_.tech(true).focus();
     }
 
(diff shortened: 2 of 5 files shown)


Essk (changes requested):
Thanks @BCovePW, this is a well-targeted fix and CI is green. The approach is right: redirecting focus rather than dropping it keeps the keyboard accessibility that the original workaround lost. Two change requests and one suggested correction to the description.

**1. Use the existing `stub_IS_EDGE` test seam instead of adding `stubIsEdge` to `browser.js`**

The test bundle is built with `rollup-plugin-stub` (see `rollup.config.js`), which generates a `stub_<EXPORT>` setter for every mutable export. Tests already use this for browser flags, e.g. `browser.stub_TOUCH_ENABLED(...)` and `browser.stub_IS_IOS(...)` in `test/unit/player.test.js`. Your new tests can call `browser.stub_IS_EDGE(true)` with no source change.

As written, `stubIsEdge` ships in the production bundle and is reachable as `videojs.browser.stubIsEdge(...)`, since `videojs.browser` exposes the whole module. The `@private` tag doesn't prevent that. Please drop the `browser.js` change and switch the tests to `stub_IS_EDGE`.

**2. The fallback doesn't match the description**

The description says focus falls back to the player root, but the code falls back to the tech when there is no play toggle on Edge:

```js
if (browser.IS_EDGE && playToggle) {
  playToggle.focus();
} else if (this.player_.tech(true)) {
  this.player_.tech(true).focus();
}
```

To be fair this is very much a corner case: a layout with the play toggle removed is rare in practice, so few if any real implementations would hit it. Still, as written that path would re-trigger the black frame on Edge, and the fix is a one-liner. Suggest implementing the described behaviour in both components:

```js
if (browser.IS_EDGE) {
  (playToggle || this.player_).focus();
} else if (this.player_.tech(true)) {
  this.player_.tech(true).focus();
}
```

**3. Suggested correction to the history in the description**

The description says #6318/#6508 redirected only the keyboard path. Checking the history, that isn't quite what happened:

- #6318 guarded exactly this mouse/tap call site in `big-play-button.js`: it skipped `tech.focus()` when `(browser.IE_VERSION || browser.IS_EDGE) && sourceIsEncrypted`.
- #6508 added the same guard to `poster-image.js`.
- Both guards were removed in #7701 ("remove ie-specific code"), which is the regression this PR fixes.
- The keyboard path's play-toggle focus is a separate lineage and is untouched here.

Worth restating in the description to keep the record straight for anyone tracing this history later: this restores protection that #7701 removed, in a better form. Redirecting focus preserves keyboard accessibility where the old guard dropped it, and the old `sourceIsEncrypted` gate would likely not fire in a click-to-play flow anyway, since EME sessions are typically created after playback starts.

Two notes where I think the PR has it right, no change needed:

- Guarding on `IS_EDGE` alone rather than also gating on Windows seems fine: the redirect is harmless on platforms where the composition issue can't occur, and it keeps the branch simple.
- The `PosterImage` change also affects keyboard activation, since `ClickableComponent` routes keydown through `handleClick`. That makes Edge behaviour more consistent with the `BigPlayButton` keyboard path, which is an improvement.

BCovePW:
Thanks @Essk — really appreciate the thorough review, and especially digging
up the #7701 history. All three addressed in b57efc3 (plus the description edit):

1. Dropped the `stubIsEdge` export from `browser.js`; the tests now use the
   rollup-plugin-stub auto-generated `browser.stub_IS_EDGE(...)` seam, so
   there's no production surface. Good catch on `videojs.browser` exposing it.

2. Both components now fall back to the player root on Edge —
   `(playToggle || this.player_).focus()` — so a toggle-less layout won't
   re-trigger the black frame.

3. Corrected the description: it now frames #6318/#6508 as guarding this
   mouse/tap call site and #7701 as removing those guards (the regression this
   restores, in a focus-preserving form rather than the old
   `sourceIsEncrypted` skip).

Re-requesting review. Happy to add a test for the `playToggle`-absent fallback
path too if you think it's worth covering.

Essk (approved):

PR #9220, opened 2026-08-15
Fix JSDoc param names that do not match the signatures

darkdi:
Left two that the same check flags. `Slider#handleMouseMove` documents `mouseDown` because subclasses take it and the base is an empty stub, and `removeRemoteTextTrack` names its parameter `obj` but falls back to treating the whole argument as the track, so `track` reads closer to what callers actually pass.


lbesecker195 (approved):
Each JSDoc `@param` now matches the actual parameter name in the signature, and the removed `ready` param on `SubsCapsButton` is not in the constructor. LGTM.

PR #9228, opened 2026-08-29
fix(Menu): only set role=menu on the content list once it has an item

### LINKED ISSUE #7688: A11Y - Broken ARIA menu on Chapters and Closed Captions (Web Accessibility)
## Description
[WAVE Accessibility Tester](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh?hl=en-US)
and
[This example official video.js page: https://videojs.com/city](https://videojs.com/city) or wherever video.js is loaded

brings up a Broken ARIA menu "Error" on the following hidden player options if they are not used (no chapters added or no CC added etc.).

1. Chapters
2. Closed Captions

`role=menu` attribute shouldn't be added to the "vjs-menu" div's "vjs-menu-content" ul tag if the "unordered list(ul tag)" doesn't have any children(when there are no captions or no chapters added)

Considering 90% of the time people don't add Chapters or Closed Captions. This accessibility error shows up almost all the time on all video.js instances.

## Steps to reproduce
1. Go to https://videojs.com/city
3. Activate WAVE Accessibility Tester ([WAVE Chrome Plugin can be used to test this error](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh?hl=en-US))
4. Broken ARIA Menus are visible and can be seen/reproducible after turning the "style" off on the WAVE accessibility tool. (see screenshots below)

### Expected
There shouldn't be any ARIA errors.

### Actual
There are Accessibility errors due to the "role=menu" div having zero children.

### Error output

Screen Shot 2022-03-21 at 6 01 13 PM

Screen Shot 2022-03-21 at 6 01 49 PM
Screen Shot 2022-03-21 at 6 02 03 PM
Screen Shot 2022-03-21 at 6 02 30 PM
Screen Shot 2022-03-21 at 6 03 00 PM

## Additional Information
#### videojs

- All versions starting from 7.0.8 to 7.18.1(The Latest Version)

#### browsers
what browser are affected?

- All browsers

#### OSes

- All OSes

### plugins

- Not applicable (N/A)

nityanand123gupta:
## Problem

`Menu.createEl()` sets `role="menu"` on the content list unconditionally at element creation, before any items exist. Accessibility tools (e.g. WAVE) flag this as a broken ARIA menu whenever a menu button has nothing to show — most commonly the Captions/Chapters buttons when a video has no caption tracks or chapters, which per the issue happens on the large majority of real-world videojs instances.

## Fix

Moved the `role="menu"` assignment out of `createEl()` and into a new `updateMenuRole_()` helper, called from both `addItem` and `removeChild`, which sets the role once there's at least one real (non-title) item and removes it when the menu is empty again. It reuses the existing "is this child just the title placeholder" check already present in `focus()`, since `MenuButton` always adds its title via `addItem` too — a title-only menu (the exact real-world scenario in this issue) must not get `role="menu"` either.

## Testing

I don't have a full karma/browser test cycle completing in this environment (video.js's suite needs a full webpack/browserify/rollup build cycle that didn't finish in the time I had), so I want to be upfront about that rather than claim more than I verified. What I did confirm:

- `npx rollup -c` builds clean with this change — no syntax/bundling errors.
- There's already a maintainer-authored test in `test/unit/menu.test.js`, `"should add or remove role menu for accessibility purpose"`, that asserts exactly this behavior (no `role` attribute with zero items, `role="menu"` once an item exists). On the current unmodified `main`, that test should fail, since `createEl()` sets the attribute unconditionally regardless of item count — this fix is what makes it pass.
- The pre-commit `vjsstandard --fix` linter (this repo's own style checker) ran clean against the change.

I'd appreciate CI (or a maintainer) confirming the full suite, especially that existing test, since I could only verify the logic and build in isolation here.

Fixes #7688

### DIFF
--- a/src/js/menu/menu.js
+++ b/src/js/menu/menu.js
@@ -90,6 +90,8 @@ class Menu extends Component {
 
     this.removeEventListenerForItem(component);
     super.removeChild(component);
+
+    this.updateMenuRole_();
   }
 
   /**
@@ -105,6 +107,26 @@ class Menu extends Component {
     if (childComponent) {
       this.addEventListenerForItem(childComponent);
     }
+
+    this.updateMenuRole_();
+  }
+
+  /**
+   * A menu with no selectable items -- an empty captions/chapters menu, for
+   * example, still shows its title -- has no need for `role="menu"`; leaving
+   * it in place is an ARIA accessibility error when the underlying list has
+   * no children a screen reader could navigate to.
+   */
+  updateMenuRole_() {
+    const children = this.children();
+    const hasTitle = children.length > 0 && children[0].hasClass('vjs-menu-title');
+    const itemCount = hasTitle ? children.length - 1 : children.length;
+
+    if (itemCount > 0) {
+      this.contentEl_.setAttribute('role', 'menu');
+    } else {
+      this.contentEl_.removeAttribute('role');
+    }
   }
 
   /**
@@ -119,8 +141,6 @@ class Menu extends Component {
     this.contentEl_ = Dom.createEl(contentElType, {
       className: 'vjs-menu-content'
     });
-
-    this.contentEl_.setAttribute('role', 'menu');
 
     const el = super.createEl('div', {
       append: this.contentEl_,


nityanand123gupta:
Update: the full karma suite did finish after all (just took longer than expected here). Final result: **4169 SUCCESS, 2 FAILED** — both failures on Edge only; Chrome Headless passed all 1391 tests with zero failures, including `should add or remove role menu for accessibility purpose` in `menu.test.js`. No mention of "menu" appears in the failure output at all, so the 2 Edge failures look unrelated to this change (pre-existing Edge-specific flakiness, most likely) — but flagging for a maintainer to confirm on CI since I can't fully rule that out myself.

